PR #2, opened 2015-03-31
replaced ifusio.com by ${domain}

askz:
Your domain was in the start-mailserver.sh script, so the domain folder was never created with ENV vars.

### DIFF
--- a/start-mailserver.sh
+++ b/start-mailserver.sh
@@ -18,7 +18,7 @@ do
 	userdb ${login} set uid=5000 gid=5000 home=/var/mail/${domain}/${user} mail=/var/mail/${domain}/${user}
 	echo "${pass}" | userdbpw -md5 | userdb ${login} set systempw
 	echo "${pass}" | saslpasswd2 -p -c -u ${domain} ${login}
-	mkdir -p /var/mail/ifusio.com
+	mkdir -p /var/mail/${domain}
 	maildirmake /var/mail/${domain}/${user}
 
 done < /tmp/docker_mail_users


tomav:
Thank you @askz.

askz:
There is still errors at the creation of the container, I'm working on it.

askz:
For instance, I have this : 
`mail_1 | Regenerating 'vmailbox' for given users
mail_1 | docker_mail_users => ['max@xxx.fr|xxxxxx']
mail_1 | user 'max' for domain 'zkp.fr' with password 'xxxxxxx'
mail_1 | maildirmake: /var/mail/zkp.fr/max: No such file or directory
mail_1 | Postmap configurations`

tomav:
I know why. At this time, we try to postmap a file that doesn't exists in the container.
It's provided by a volume containing the full /etc/postfix directory.
I'm fixing it right now.

tomav:
@askz I just updated the repository. Please have a look to the changes.

askz:
Thanks, the vmail error is fixed. 
but now : 
```
mail_1 | Postfix configurations
mail_1 | sed: can't read /etc/postfix/main.cf: No such file or directory
mail_1 | postmap: fatal: open /etc/postfix/main.cf: No such file or directory
```

tomav:
Hmm have you updated your local git repository? I tried locally and it works.
Or are you using the docker pre-built images tvial/docker-mailserver? (I just had a look and build has failed but it doesn't semmed to be related to the changes but a Docker infrastructure issue.)

tomav:
The image on the registry is up to date. You can `docker pull tvial/docker-mailserver`.
Don't forget to remove existing container before you run the new one.

askz:
Okay, thank you for this, the docker-compose is now building, I can log into imap server, but smtp's refusing login.
I'm very interested in this kind of container, I'll give it a shoot when I'll have some time !

Issue #10, opened 2015-07-31
Improvement on virtual


dy0nysos:
Hi,

Once again, thanks for your job. It's pretty amazing to get a full working postfix + clamav + spamassassin server.

Is it possible to separate virtual and vmailbox configuration?

I need to have a contact email that relays its email on several other adresses (such as gmail, outlook etc.). To do this, I didn't find any other way than connect me on the container and modify the virtual file.
Then postmap it and reload postfix configuration.

The problem is that the virtual file is recreated each time I start the container.

Is there a way to make a new conf file (added in the postfix folder on the host machine) to set the virtual emails ?

tomav:
@dy0nysos, a new file will work.
But perhaps we can improve `accounts.cf` llike this: 

    # real account
    user@domain.tld|mypassword|alias1,alias2
    # relay account
    user@domain.tld|RELAY|user@external1.com,user@external2.com

With this solution, `accounts.cf` will remain the same.
You just have to improve `start-mailserver.sh` to add the parsing of `RELAY` accounts correctly.

It's an open discussion, let me know what you think.

tomav:
Hmm, an advantage of your solution with a new file is that it will separate `accounts.cf` and `redirects.cf` (or other name) and could avoid errors.

dy0nysos:
The question is, do we need to remove aliases from `accounts.cf` and set up all aliases in `redirects.cf` (or whatever is its name) or mixing both aliases config.

My opinion is that we should separate accounts and aliases config. With this split, you could set only virtuals or only accounts, and an error in one of these file does not break the entire container.

Moreover, it could be a first step to mailiing lists or this kind of advanced mail features.

tomav:
So `accounts.cf` will only contain mail accounts and passwords.
Both aliases and redirect will be managed by `redirect.cf`.
This is ok for me.

Note that it will break backward compatibility.

dy0nysos:
Seems good to me too.

PR #11, opened 2015-08-05
Accounts and virtual conf split

dy0nysos:
#10 Separate virtuals and accounts.

### DIFF
--- a/postfix/accounts.cf
+++ b/postfix/accounts.cf
@@ -1,1 +1,1 @@
-user@domain.tld|mypassword|alias1,alias2
+user@domain.tld|mypassword
--- a/postfix/redirects.cf
+++ b/postfix/redirects.cf
@@ -1,1 +1,0 @@
-user@domain.tld | otheruser@domain.tld,otheruser@otherdomain.tld
--- a/postfix/redirects.cf
+++ b/postfix/redirects.cf
@@ -1,1 +1,1 @@
-user@domain.tld | otheruser@domain.tld otheruser@otherdomain.tld
+user@domain.tld | otheruser@domain.tld,otheruser@otherdomain.tld
--- a/postfix/redirects.cf
+++ b/postfix/redirects.cf
@@ -1,0 +1,1 @@
+user@domain.tld | otheruser@domain.tld otheruser@otherdomain.tld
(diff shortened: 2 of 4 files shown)


tomav:
Good job @dy0nysos!
But I'm thinking that `aliases` and `redirects` are 2 different things (and we need both).

`user@domain.tld|otheruser@domain.tld,otheruser@otherdomain.tld` is nice but with this, we have to manage 1 entry per alias, no?
I mean 1 real account, with 50 aliases will have 1 entry or 50 entries?

dy0nysos:
Sorry I thought I answered your question.

Here I'm managing redirects only. user@domain.tld to otheruser@domain.tld,anotherone@domain.tld etc.

In facts, if you need one mail to have several aliases, you'll have one line per alias.

Can you see a ( new ? ) file structure that will allows us to have several aliases for one email on a single line ?

Maybe could we add another conf file, `aliases.cf` with that kind of structure : 
`mailtobealiased|alias1,alias2,alias3` or something of this kind ?

tomav:
Yes, this kind of format existed before the PR, aliases were managed, not redirects.

The point is that `accounts.cf` remains mandatory and must be parsed when the container starts because there are commands to execute to create these accounts.

We can add a new file format and parse it, but I think it's not really necessary.
Aliases and redirects may be static files in postfix format, just copied from `/tmp`to `/etc/postfix/virtual`.

Aliases and redirects are managed the same way.
http://www.postfix.org/VIRTUAL_README.html#forwarding

Sorry, I'm on vacation and can't do more than answering at this time.
I'll be back on monday.

dy0nysos:
Okay, I'll change my code to change the behavior according to your proposition.

Enjoy your vacation ! :+1:

tomav:
Thank you.

tomav:
I don't know why `build` fails on the docker registry. I'm re-trying locally.

tomav:
I'm not blind, it works locally. Waiting for the registry.

tomav:
Latest docker registry build is `done`.
https://hub.docker.com/r/tvial/docker-mailserver/builds/bjkcm6jaq4bzfaizu5mqhrp/

dy0nysos:
Nice !

PR #26, opened 2015-10-14
forward only domains should be added to vhosts

### LINKED ISSUE #25: forward only domains are not added to vhosts
When I add a domain in /postfix/virtual like this

test@test.com blah@example.com

The domain test.com should appear in vhosts, but it only appears if there is a user @test.com in accounts.cf.

tbutter:
should fix #25


tomav:
Thank you @tbutter!

PR #27, opened 2015-10-16
Added docker subnet to avoid relay access denied

dy0nysos:


tomav:
Thank you @dy0nysos!

dy0nysos:
You're welcome

tomav:
@dy0nysos I get some `warning: non-null host address bits in "172.17.42.1/16", perhaps you should use "172.17.0.0/16" instead` and don't receive mail anymore. Do you experience the same behaviour?

tomav:
Yeah, changing the range to `172.17.0.0/16` fixes the problem. I'm gonna push it soon.

tomav:
Let me know if it still fits your needs.

dy0nysos:
I encountered the problem only when I pulled the new image. 
I tried changing with  ```172.17.0.0/16``` and it works so it's okay for me

tomav:
The new image is building. I'll let you know when everything is ok.

dy0nysos:
Image pulled.
Seems to be okay

tomav:
Same here.

tomav:
I'd like to add unit / integration testing to this image but I don't know really how... I'll open an issue as a reminder.

gegere:
Greetings @tomav I have been building to using build tests on my docker images and thought I would share this with you. At the bottom of [a mail service readme](https://github.com/htmlgraphic/postfix) you will see **Shippable** and **CircleCI**

This might be something I can help you improve on your docker build so functions can be tested with ease after each contributed commit.

tomav:
Thank you @gegere, I'll have a look to shunit2.

tomav:
Added tests using `assert.sh` in #29 to fix #28

PR #37, opened 2015-11-20
added rm -rf /var/lib/apt/lists/*

kopax:
- cleanup lists /var/lib/apt/lists/*

For a smaller image


tomav:
Thank you @kopax

tomav:
Went from 565.7 MB to 528.5 MB

PR #43, opened 2015-12-06
Fixed #42 - No more weak ciphers.

### LINKED ISSUE #42: Disable weak SSL ciphers
Can be checked using https://ssl-tools.net/mailservers

tomav:


tomav:
You can check ciphers with https://ssl-tools.net/mailservers

PR #48, opened 2016-01-08
Added SMTP and IMAP ports

sirgantrithon:


tomav:
Thank you @sirgantrithon

PR #50, opened 2016-01-12
Add plain and login mech

tomav:


tomav:
You can `docker pull tvial/docker-mailserver:add-plain-and-login-mech` to test it before I merge into master.

PR #52, opened 2016-01-15
Do not create dir if it already exists

spacecowboy:


start-mailserver.sh
```diff
@@ -25,7 +25,12 @@ if [ -f /tmp/postfix/accounts.cf ]; then
     echo "${pass}" | userdbpw -md5 | userdb ${login} set systempw
     echo "${pass}" | saslpasswd2 -p -c -u ${domain} ${login}
     mkdir -p /var/mail/${domain}
-    maildirmake /var/mail/${domain}/${user}
+    if [ -d "/var/mail/${domain}/${user}" ]; then
```
tomav: Why not only:

    if ! [ -d "/var/mail/${domain}/${user}" ]; then
       maildirmake "/var/mail/${domain}/${user}"
    fi

I think we don't need the comment.
I probably should remove some outputs but it's another story.

spacecowboy: Sure. It was nice to have during development is all. Can remove it

spacecowboy:
@tomav fixed your comment

tomav:
Thank you!

Issue #55, opened 2016-01-19
FeatureRequest: add dkim support


bilak:
What about adding [dkim](https://www.digitalocean.com/community/tutorials/how-to-install-and-configure-dkim-with-postfix-on-debian-wheezy) support?

tomav:
:+1: 

I think @spacecowboy worked on it:
https://github.com/spacecowboy/docker-mailserver/commit/0704ff1bf64f5a48ec680705b495e1233ec2688a

spacecowboy:
I never finished it though. Instead I implemented it and other things from scratch in https://github.com/spacecowboy/docker-post because I wanted to integrate with dovecot and use a database.

You can probably take most of the logic though. Some key points:

Setup of keys is done here: https://github.com/spacecowboy/docker-post/blob/master/postfix/run.sh#L3

Outgoing mail is signed with dkim via a milter: https://github.com/spacecowboy/docker-post/blob/master/postfix/master.cf#L42

Incoming mail signatures are verified by Amavis: https://github.com/spacecowboy/docker-post/blob/master/amavis/50-user#L23

rubytastic:
Looks great; great repo only a bit to complex like the simplicity of this one.
Going to run this container soon and if time allows take a look at this also.

tomav:
@bilak just PR this feature. Code is being reviewed.
https://github.com/tomav/docker-mailserver/pull/56

tomav:
Merged.

tomav:
@bilak does the server sign outgoing emails?

When I do `amavisd-new testkeys` I have:
`No DKIM private keys declared in a config file.`

Are you sure there's nothing to do in Amavis configuration?

Thank you.

bilak:
Hi @tomav.
Yes it's definitely signing outgoing requests, but I'm testing this from java application in my laptop  and server with docker is another computer. Try to remove entries from `/etc/opendkim/TrustedHosts` reload the config and send mail.

tomav:
Yes, you're right.

    BC0BB9F9: DKIM-Signature field added (s=mail, d=domain.com)

I was asking that because there are DKIM options in Amavis.

PR #56, opened 2016-01-20
#55 - added DKIM support -i 55

bilak:
Added support for DKIM
Maybe something can be refactored, but it's working correctly now.


tomav:
Thank you @bilak!

My questions:
- does it work with multi-domains because I see a `domainname` ENV added to the project.
- I'm not a DKIM expert but is there a way to add integration tests?

Thank you.

postfix/TrustedHosts
```diff
@@ -0,0 +1,2 @@
+127.0.0.1
+localhost
```
tomav: Missing new line.

postfix/default-opendkim
```diff
@@ -0,0 +1,12 @@
...
+#DAEMON_OPTS=""
+#
+# Uncomment to specify an alternate socket
+# Note that setting this will override any Socket value in opendkim.conf
+#SOCKET="local:/var/run/opendkim/opendkim.sock" # default
+#SOCKET="inet:54321" # listen on all interfaces on port 54321
+#SOCKET="inet:12345@localhost" # listen on loopback on port 12345
+#SOCKET="inet:12345@192.0.2.1" # listen on 192.0.2.1 on port 12345
+
+SOCKET="inet:12301@localhost"
```
tomav: Missing new line.

postfix/opendkim.conf
```diff
@@ -0,0 +1,21 @@
...
+KeyTable                refile:/etc/opendkim/KeyTable
+SigningTable            refile:/etc/opendkim/SigningTable
+
+Mode                    sv
+PidFile                 /var/run/opendkim/opendkim.pid
+SignatureAlgorithm      rsa-sha256
+
+UserID                  opendkim:opendkim
+
+Socket                  inet:12301@localhost
```
tomav: Missing new line.

start-mailserver.sh
```diff
@@ -1,10 +1,55 @@
...
+  echo "Creating DKIM SigningTable"
+  echo "*@$domainname mail._domainkey.$domainname" > /etc/opendkim/SigningTable
+fi
+echo "Changing permissions on /etc/opendkim"
+# chown entire directory
+chown -R opendkim:opendkim /etc/opendkim/
+# And make sure permissions are right
+chmod -R 0700 /etc/opendkim/keys/
+
+# Opendkim:
```
tomav: These next lines are needed for debug but not on a production server.
(I have to clean the current debug by the way..)

bilak:
- It's now working only for one domain. We can try to convert it to multi-domain in future, but there must be something with what we can decide there's more domains (maybe domainname values separated by comma)
- About integration tests - only thing which we can check is information in log `mail opendkim[579]: 359D4A1245: DKIM-Signature field added (s=mail, d=domain.com)`. I don't know much about postfix neither about docker so your help will be needed.

tomav:
- You probably don't have to provide a domain list since it already exists and is parsed to create users, aliases and postfix configuration. https://github.com/tomav/docker-mailserver/blob/master/start-mailserver.sh#L20
- For the tests, I will do it, no problem.

bilak:
- So if I understand it correctly, there could be function created from my script which will accept only domain name (think about how to check duplicated processing - for example ten users with same domain and another 10 users with another domain). There's need to create (if not exists) DKIM (privatet, txt), add entry to KeyTab and to SigningTable.

tomav:
The `postfix/vhost` file is generated by startup script and contains unique domains.
(I just found a strange behaviour on my server and trying to fix it but it does the trick).
I think you can use this file as reference and paste your DKIM startup lines after its generation.

The point is, for these domains, enable DKIM if necessary information are provided.

tomav:
(about the bug mentioned above, it was not the container but my `virtual` alias list, so no worry.)

tomav:
Thank you @bilak

Issue #54, opened 2016-01-18
About pop3 support


crash7:
Hi,

Is there any special reason about not installing pop3 (and pop3-ssl)? Seems that with the courier modules and a couple of lines in the config (TLS certs and port), it should work.

I'm asking because if you want to use the mail server with gmail pop3 check, it may not work since this only supports imap.

Great work anyway, these docker images is really helpful.

UPDATE: just saw this issue https://github.com/tomav/docker-mailserver/issues/32, sorry about the duplication.

tomav:
Hey @crash7, thank you for your message.

If you want to add `pop3` support in this public image:
- `pop3` should be disabled by default
- enabling `pop3` could be done using an ENV variable
- don't forget to add integration tests

If you PR I'll merge. I can help you if you have any question.

tomav:
@crash7 Let me know if you work on it or not.
Thank you.

crash7:
Hi! 
Yes, I will try to submit this during the weekend.

tomav:
Thanks!

awb99:
I have ENV variable set for POP. But I cannot telnet into the pop ports. Why is this?
In docker ps it shows that the pop ports 110 and 993 arenot showing 0.0.0.0    but instead
no IP address.

PR #63, opened 2016-01-23
Add pop3 support

crash7:
Add pop3 support with optional flag as per #54 

Note:
self-signed certificates won't work with gmail since they don't support them anymore


tomav:
Thank you @crash7

PR #69, opened 2016-01-28
 - fixed crash of opendmarc

bilak:
opendmarc startup failed due to wrong variable and missing new line in opendmarc.conf


tomav:
Thanks. Could you add a test on this?

PR #78, opened 2016-02-12
added fail2ban

dominikwinter:
To prevent brute force attacks like:

`Feb 12 18:17:25 mail postfix/smtpd[2595]: warning: unknown[xxx.xxx.xxx.xxx]: SASL LOGIN authentication failed: authentication failure
`


docker-compose.yml.dist
```diff
@@ -2,6 +2,7 @@ mail:
   image: tvial/docker-mailserver
   hostname: mail
   domainname: domain.com
+  privileged: true
```
tomav: Why do we need privileged here?

start-mailserver.sh
```diff
@@ -188,6 +188,22 @@ echo "required_score 5" >> /etc/mail/spamassassin/local.cf
 echo "rewrite_header Subject ***SPAM***" >> /etc/mail/spamassassin/local.cf
 cp /tmp/spamassassin/rules.cf /etc/spamassassin/
 
+
+echo "Configuring fail2ban"
+# enable filters
+perl -i -0pe 's/(\[postfix\]\n\n).*\n/\1enabled  = true\n/'     /etc/fail2ban/jail.conf
```
tomav: I would have used sed to like the other ones to simplify maintenance.

tomav:
Good idea!

docker-compose.yml.dist
dominikwinter: To ban ip addresses fail2ban uses iptables. Running in privileged mode will allow to run without the permission denied message: "permission denied (you must be root)"

start-mailserver.sh
dominikwinter: At first I also tried sed. But to replace multiline strings with perl is much easier :)

See https://unix.stackexchange.com/questions/26284/how-can-i-use-sed-to-replace-a-multi-line-string for example.

docker-compose.yml.dist
tomav: Ah ok. And since `privileged` is global, can we use `--cap-add NET_ADMIN` or something like that?

start-mailserver.sh
tomav: Let's keep like this for now.

docker-compose.yml.dist
dominikwinter: Yes, you are right. This is much better. I just read it on:
- https://docs.docker.com/compose/compose-file/#cap-add-cap-drop
- https://groups.google.com/forum/#!topic/docker-user/0e_zsN61Uk0

tomav:
Could you add tests?
One regarding [running process](https://github.com/tomav/docker-mailserver/blob/master/test/test.sh#L7) and perhaps we can simulate a brute force attack.

Let me know if I can help.

tomav:
BTW, test that you don't get `permission denied` using docker capability instead of `privileged` is a good test :-)

tomav:
Thank you @dominikwinter!

PR #81, opened 2016-02-18
fix pop3 certificate path

espitall:
Pop3 certificate path is broken when "letsencrypt mode" is selected.


tomav:
Thank you @espitall!

Issue #98, opened 2016-03-05
Reducing docker image size


Brandl:
Hello,

there are not many solutions for a docker based mailserver out there and from what I've looked at this is a very promising project. Starting at feature-completness to active maintenance this features a lot of the things I'm looking for to use in production.

One thing, that would totally win me over is to reduce the docker image size to a minimum. Currently the image size is around 500mb, which is okay, but could be improved a lot:
https://imagelayers.io/?images=tvial%2Fdocker-mailserver:latest

Here is a checklist of things that worked well:
- [ ] move to a minimal base image like docker alpine (100-200mb reduction)
- [ ] do installs and cleanup in a single RUN-command 
- [ ] reduce number of installed packages (do you really need vim in a container?)

This might include some compatibility breaking changes and requires some fine tuning, like looking for equivalent packages in the docker alpine package index. On the plus side, this would:

- Decrease download size and increase download/build speed
- Increase start/general performance
- Decrease resource usage and footprint
- Improve security by minimising the attack surface

The docker fluentd image, is a excellent showcase, where this worked well:
https://github.com/fluent/fluentd-docker-image/pull/9

tomav:
Hi @Brandl and thanks for your support.
These tasks can be done very quickly:
* do installs and cleanup in a single RUN-command
* reduce number of installed packages (do you really need vim in a container?)

If you wanna PR this, it can be merged very fast.

Regarding the base image, why not! I'll have to have a look to these images. I mostly work with Ubuntu and Debian so I don't know the other.

Gymnae:
I just replied to a closed request going in the same direction.
It would be superb to see this project with an Alpine base and it's smaller attack surface.

PhilippeChepy:
I have removed vim, and merged lots of commands in Dockerfil in #102

tomav:
The move to Alpine will break a lot of changed. 
We can plan that later in a specific branch to test things before merge.
Anyone can work on that?

The rest is being discussed in #102 with @PhilippeChepy.

tomav:
An image can be tested at `tvial/docker-mailserver:lightweight `

tomav:
tvial/docker-mailserver           lightweight                fe2efbb8c8e8        47 hours ago        432.4 MB
    tvial/docker-mailserver           latest                     294dcb629180        2 weeks ago         540 MB

tomav:
Anyone tried?

jmparra:
i have tested with debian, but no improvements in image size
[...]

PR #102, opened 2016-03-11
Reduce image size - #98

PhilippeChepy:
- Reduced image size from 550,5 MB to 386 MB.


Dockerfile
```diff
@@ -1,74 +1,71 @@
 FROM ubuntu:14.04
```
jmparra: maybe move to [Debian:8.3](https://imagelayers.io/?images=ubuntu:14.04,debian:8.3) can help

PhilippeChepy: I have tested with Debian:jessie and curiously size of image increased. But it produce a working image as well.

jmparra: works for me, but i use other [cleanup](https://hub.docker.com/r/cgswong/min-jessie/) process
```
apt-get -y clean && apt-get -y autoclean && apt-get -y autoremove && \
cp -R /usr/share/locale/en\@* /tmp/ && rm -rf /usr/share/locale/* && mv /tmp/en\@* /usr/share/locale/ && \
rm -rf /var/cache/debconf/*-old && rm -rf /var/lib/apt/lists/* && rm -rf /usr/share/doc/*
```

tomav:
@PhilippeChepy thanks for this work!
Is there a reason you moved a lot of things into `target/` folder?
Are you sure it won't break something on running docker mail servers?

tomav:
Hi @PhilippeChepy, let me know. Note that the branch has conflicts.

PhilippeChepy:
I moved all files that needs to be in image in the `target/` folder just for practical reasons. I prefer doing one `COPY` instead of multiple `ADD`s in my Dockerfile. If you want to add a file in your image, you just have to add that file in the `target/` folder.
I guess it's not very important for the size reduction of the image.

tomav:
I'm merging manually and checking things.

tomav:
This PR is conflicting again. @PhilippeChepy could you have a look?
Thank you.

tomav:
Closed due to inactivity.

PR #107, opened 2016-03-23
[wip] Feature to avoid clear text pwd

00angus:
@tomav 
This is a possible approach to the problem of having to handle clear text passwords for user accounts. 
My proposal is : 

-  start with an account.cf (and clear text passwords)
-  run _generate-user-databases_ inside the docker: it will export two DBs (for courier and sasl)
-  run the docker mailserver providing userdb and sasldb2 instead of accounts.cf

That way passwords are kept secret even on the host.

The project buils and runs as for my own tests.
Part of the code that handles users in _start-mailserver.sh_ is redundant wrt that of the new script _generate-user-databases_ (probably it would need a better approach).

The docker command that could be used to generate user dbs is like the following:
docker run --rm **-v "$(pwd)"/postfix:/tmp/postfix** -h mail.fobbar.it -t angus/docker-mailserver **generate-user-databases**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,7 +49,9 @@ ADD postfix/main.cf /etc/postfix/main.cf
 ADD postfix/master.cf /etc/postfix/master.cf
 ADD postfix/sasl/smtpd.conf /etc/postfix/sasl/smtpd.conf
 ADD bin/generate-ssl-certificate /usr/local/bin/generate-ssl-certificate
+ADD bin/generate-user-databases /usr/local/bin/generate-user-databases
 RUN chmod +x /usr/local/bin/generate-ssl-certificate
+RUN chmod +x /usr/local/bin/generate-user-databases
 
 # Get LetsEncrypt signed certificate
 RUN curl https://letsencrypt.org/certs/lets-encrypt-x1-cross-signed.pem > /etc/ssl/certs/lets-encrypt-x1-cross-signed.pem
--- a/bin/generate-user-databases
+++ b/bin/generate-user-databases
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+# Generate the courier and sasl databases
+if [ -f /tmp/postfix/accounts.cf ]; then
+  # Checking that /tmp/postfix/accounts.cf ends with a newline
+  sed -i -e '$a\' /tmp/postfix/accounts.cf
+
+  # Creating users
+  while IFS=$'|' read login pass
+  do
+    # Setting variables for better readability
+    user=$(echo ${login} | cut -d @ -f1)
+    domain=$(echo ${login} | cut -d @ -f2)
+    # Let's go!
+    echo "user '${user}' for domain '${domain}' with password '********'"
+    /usr/sbin/userdb ${login} set uid=5000 gid=5000 home=/var/mail/${domain}/${user} mail=/var/mail/${domain}/${user}
+    echo "${pass}" | userdbpw -md5 | userdb ${login} set systempw
+    echo "${pass}" | saslpasswd2 -p -c -u ${domain} ${login}
+  done < /tmp/postfix/accounts.cf
+  cp /etc/courier/userdb /tmp/postfix/userdb
+  cp /etc/sasldb2 /tmp/postfix/sasldb2
+  echo "Courier and Sasl databases populated with user accounts"
+fi
--- a/start-mailserver.sh
+++ b/start-mailserver.sh
@@ -5,6 +5,22 @@ die () {
   exit 1
 }
 
+mkpaths () {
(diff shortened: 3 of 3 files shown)


tomav:
Build timeout, I restart manually.

tomav:
Green. I'll check in detail tomorrow or friday. Thank you @00angus

tomav:
@00angus, I like the approach.

So the process is:
* `docker run --rm -v "$(pwd)"/postfix:/tmp/postfix -h host.domain.tld tvial/docker-mailserver generate-user-databases` (now crypted db is on the host)
* `docker-compose up mail` (to simplify)

Questions:
* how do you choose between current `accounts.cf` and crypted db method?
* could you add integration tests on the crypted version?

Thank you.

00angus:
On Mar 24, 2016 12:56, "Thomas VIAL" <notifications@github.com> wrote:
>
> @00angus, I like the approach.

Thanks! :)

>
> So the process is:
>
> docker run --rm -v "$(pwd)"/postfix:/tmp/postfix -h host.domain.tld
tvial/docker-mailserver generate-user-databases (now crypted db is on the
host)
> docker-compose up mail (to simplify)
>

Correct.

> Questions:
>
> how do you choose between current accounts.cf and crypted db method?

Currently:
1) first look for: accounts.cf
2) if missing, look for the 2 dbs
3) else no user defined

> could you add integration tests on the crypted version?

I'll look into it :)

-- Marco

00angus:
@tomav I've looked at the tests. From my understanding the current ones are still valid if we run the mailserver by providing "accounts.cf" or  userdbs. In the latter case I don't see special cases needed to be tested apart from those already implemented.
What do you think ?

tomav:
@00angus can we keep this feature for `v2`only?

tomav:
> Currently:
> 1) first look for: accounts.cf
> 2) if missing, look for the 2 dbs
> 3) else no user defined

I think you should do:
1. Check if encrypted db exists and use it if true
2. Use `accounts.cf` if no crypted db is provided

This could be the test case.

00angus:
@tomav if I look at the tests, accounts tests in particular, those ones are ok, do you agree? 
What you cited above is the startup logic: are you suggesting to change it? 

As for keeping this PR, I think it's useful only for V1 (courier version).

tomav:
Well I think that encrypted files are more secure than `accounts.cf` so it should be default.
But it should also work for existing users that don't use encrypted database.

So yes, I think you should change it but perhaps I am missing something...

00angus:
Of course I can. Let me think a little bit more on the possibilities and I'll come back :)

00angus:
I think we should talk about the expected "workflow" of a user to understand what good choices we have. Example:
she world start with a populated accounts.cf, then generate encrypted dbs, then start the mailserver. In this workflow I find keeping clear text passwords in accounts.cf dangerous. Should we *remove* (!!!) it? Dangerous. Should we exit with an error message? Probably better.

Or she world start with a populated accounts.cf, then encrypted dbs are generated from clear text passwords and that dbs used from now on? At the next startup encrypted dbs will automatically be used. 

I previously suggested to let user choose between encrypted or not pwd... But probably I was wrong and a better solution should be provided. 

For me 2nd workflow hypothesis could be a choice.

Comments?

tomav:
I know that a lot of users chose this image to version accounts in their favorite (and private) version control system. This means that the original source of data for accounts creation remains `accounts.cf`.

Of course, and it's the purpose of this enhancement, this file must be deleted on the running instance of this image.

So yes, the expected workflow is:

1.  `docker run --rm -v /host/config:/tmp/docker-mailserver tvial/docker-mailserver command-to-generate-crypted-db`, so the db will be sent to host (in a different dir of `accounts.cf`).
2. `docker-compose up -d mail` (with all needed options) mounting the folder containing only the crypted db.

If no database is provided on start, we should exit with an appropriate and explicit message.

Question: the crypted db only contains accounts, right? `virtual` remain a static file copied from host to postfix folder? (I'll have to dig into dovecot...). Because we also have to think how to deploy a new alias (this should be very simple).

00angus:
I agree with and will work to implement this workflow. The use of a different path to store encrypted dbs is a good choice.

Regarding your question @tomav : here we are talking about docker-mailserver version using Courier, right? The encrypted DBs we are talking about are those of Cyrus Sasl and Courier and those only contain accounts.

The same apply anyway to the userdb I've implemented in `v2`: it only contains users and passwords. `virtual `hasn't changed from `v1`. As a general rule of thumb I've maintened full compatibility with v1 setup and configuration, the only addition I've not (yet) PRed to master being a new feature on opendkim keys handling.

PR #108, opened 2016-03-26
fixed log rotate 

dominikbechstein:
syslog and mail (postfix?) daemon were not continue to write the logs in the newly created file after rotating the old log file. so after a while fail2ban daemon was watching the new (but empty) file and did not work anymore.

see `ls -l /var/log` and `lsof | grep /var/log` for example.

adding "copytruncate" in the configuration file is probably a common solution.

by the way: i also switched on the file compress feature (.gz) for rotated files.


tomav:
Thank you @dominikbechstein

Issue #110, opened 2016-03-29
Issue with combined.pem from LE certificates


DominikTrenz:
I had many problems getting the courier server running with LE certificates. After a while i found the problem:

The .pem files from LE are merged into "combined.pem" and when there is no newline at the end of the LE files this happens:
`-----END CERTIFICATE----------BEGIN CERTIFICATE-----`
Which causes the imap server to fail.

Changing Line 129 in start-mailserver.sh to:
`( cat "/etc/letsencrypt/live/$(hostname)/cert.pem"; echo""; cat "/etc/letsencrypt/live/$(hostname)/chain.pem"; echo ""; cat "/etc/letsencrypt/live/$(hostname)/privkey.pem"; echo ) > "/etc/letsencrypt/live/$(hostname)/combined.pem"`
solves the problem

tomav:
Hi @DominikTrenz, do you want to submit a PR?

chadyred:
How do you create the LE certificate ? 

On many case I use the URL /well-known URL when I have a web server running/ I don't have this possibility to do that, so I don't know how generate a certificate with LetEncrypt with a mail server.

DominikTrenz:
@tomav Maybe i will do so the next days. Thanks for this awesome image btw :)
@chadyred I use following docker images for automatic LE certificates:
```
jwilder/nginx-proxy
jrcs/letsencrypt-nginx-proxy-companion  
```
The proxy-companion sets up the well-known urls in nginx and renews the certificates which are shared the email server.

chadyred:
Yes, I alse use them, but to generate certificate jrcs/letsencrypt-nginx-proxy-companion and particulary lets encrypt, use an URL for ACME challenge is needed and mail server container don't have one. 

My idea : 

Create a container with nginx which could receive /well-known link to mail server
So the certificate is generate and I shared this one to my mail container in the /etc/letsencrypt/live
folder.

DominikTrenz:
The ACME challenge should be received just fine if you mounted the /usr/share/nginx/html volume from nginx in the letsencrypt container. At least it works fine with my setup. I also had to modify the letsencrypt image to generate all the necessary certificate files for the mailserver.

chadyred:
Thanks for help, I will re-open an issue if I have more problem more developed.

Reldeis:
PR #111 fixes the Problem

[...]

PR #111, opened 2016-03-30
Add eol to letsencrypt provided files [Issue #110]

Reldeis:
Some LE tools like simp_le create the cert files with no end-of-line. This change adds an eol if non exists.

If a file does not exist, sed throws an error on STDERR. It might be a good idea to check for file existence first like in the "self-signed" and "custom" case.

This fixes [Issue #110 ](https://github.com/tomav/docker-mailserver/issues/110).

### DIFF
--- a/start-mailserver.sh
+++ b/start-mailserver.sh
@@ -120,6 +120,10 @@ fi
 case $DMS_SSL in
   "letsencrypt" )
     # letsencrypt folders and files mounted in /etc/letsencrypt
+      # add eol to all files before concatenation
+      sed -i -e '$a\' /etc/letsencrypt/live/$(hostname)/cert.pem
+      sed -i -e '$a\' /etc/letsencrypt/live/$(hostname)/chain.pem
+      sed -i -e '$a\' /etc/letsencrypt/live/$(hostname)/privkey.pem
 
       # Postfix configuration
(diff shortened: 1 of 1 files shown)


tomav:
Thank you @Reldeis

00angus:
I've just discovered that this patch unlink the files in the 'live' directory of LE config dir, in particular:

- under the 'live' path there are symlinks to the archive directory, for continuity between updates
- the patch `sed -i -e '$a\' /etc/letsencrypt/live/$(hostname)/cert.pem` cause the unlink in the 'live' directory and the creation of a new file not linked to the one in the archive directory

@Reldeis : does this may breaks future updates of letsencrypt certificates ?

Reldeis:
I can't recreate this on my machine in a quick test. The files created with the standard LE binary already have the eol, so the  files are not changed. 
I'll look into it.

@tomav I will change it the way Dominik suggested in #110. That way the LE created files are never touched and no unlinking should happen

00angus:
Thanks for the fast update. Anyway for future reference the problem arise even in case EOLs are presents in the files, the problem being simply the use of the switch '-i' to sed itself.

Reldeis:
yes i see where the problem was. On my host (Ubuntu 14.04) the links stay in tact.
However i fixed it in #118 by simply not changing the LE created *.pem files

Issue #112, opened 2016-03-30
fail2ban and iptables question


00angus:
It would be useful to have Fail2ban not started by default, IMHO. In my case, for example, I prefer to filter on the host by looking at the container logs. And usually  I don't like privileged containers :o

What do you think?

gegere:
I agree with not having Fail2ban start by default. There is often valid IMAP requests with the incorrect password often enough and Fail2ban would block it.

KyleOndy:
Could be have this configurable via an environmental variable? That way its easy to have one way or the other.

00angus:
Yes, something like ENABLE_FAIL2BAN. I've implemented it for my own purposes: just a few more tests and I'll pull a request to see also @tomav opinion.

BTW we could also implement it as DISABLE_* if this make more sense to you, but for general concordance with the rest of the project ENABLE_*  is a better solution IMHO.

tomav:
I agree, use an env to enable it.

00angus:
BTW - by looking at the integration tests for fail2ban I've found the following:

     @test "checking fail2ban: localhost is not banned" {
      run docker exec mail_fail2ban /bin/sh -c "fail2ban-client status sasl | grep 'IP list:\s*127.0.0.1'"
      [ "$status" -eq 1 ]
    }

Here the grep will match '127.0.0.1' only when it's the first IP listed. So for example here will not match:

     - IP list:       1.2.3.4 188.188.188.188 127.0.0.1

I think this is not the intended behaviour, right? It should match even in this case because we want to be sure that localhost is not banned. 
I'm including a change for this in my next PR. @tomav have a look at it to check it is ok.

tomav:
Fixed by #114

PR #114, opened 2016-03-31
Add ENABLE_FAIL2BAN environment variable to enable fail2ban service

00angus:
Implement #112
Fail2ban is not enabled by default, now.

* Changed/updated Makefile and integration tests
* Changed some grep expressions in the fail2ban tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,12 @@ run:
 		-v "`pwd`/test":/tmp/test \
 		-e SMTP_ONLY=1 \
 		-h mail.my-domain.com -t $(NAME)
+	docker run -d --name mail_fail2ban \
+		-v "`pwd`/postfix":/tmp/postfix \
+		-v "`pwd`/spamassassin":/tmp/spamassassin \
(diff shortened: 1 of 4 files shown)


KyleOndy:
The build timed out.

@tomav care to give it a restart?

tomav:
Restarted.

00angus:
Hmm... stil time out problems:
> Downloading main.cvd [ 50%]
Keeps stopping while downloading updates.

tomav:
Restarted again. I can remove this step in the build but a newly started container will have to `freshclam` manually or wait for the cron to be executed.

00angus:
I'm experiencing me too big delays while downloading updates for clamav, anyway...
I would prefer to have an image freshly updated - IMHO, if it's possible.
Is it possible to raise the amount of the timeout?

tomav:
It's related to Clamav or Travis. I don't think inscreasing timeout would really solve the problem.

Thanks for your PR.

Issue #113, opened 2016-03-31
loggings


00angus:
I'd like to suggest to put mail.* logs inside a specific folder instead of the more general /var/log. In this way it's easy to provide let's say a -v mail-log:/var/log/mailserver where only mail logs are made available on the host.
I use that approach together with a fail2ban modification to filter incoming connections directly on the host, for example.

What do you think?

tomav:
I agree.
`/var/log/mail` could be a good place to be (not `mailserver`, project could change in v2 milestone)

00angus:
@tomav Ok, `mailserver` is not ok. I'm thinking of putting inside this directory also the `clamav` logs, so all relevant server logs (apart from fail2ban) could be made easily available on the host. 
In this case `mail` could still be adopted for that directory, I think. Ok?

00angus:
I'm going to do a PR for that issue. I've changed the way fail2ban jail.con get updated (from perl to awk) because there are now several changes to be managed and awk was simple IMHO,  but if you think perl is better please let me know.
Moreover there was an annoying message when no spamassassin rule file was not provided l, so I've added a simple check.

gegere:
I have been using a external logging service called LogEntries.com would this be something to consider? I have been using their service for about a year and have been quite happy even with the free tier. I have about 10 containers logging at any given time and can easily search across all logs. You simply pass the API call into the containers that are built and you don't need to worry about storing or managing the logs via the container.

Here is more details on this: https://logentries.com/doc/docker/

00angus:
@gegere I personally prefer to locally manage (on the host) the logs, but probably logentries.com is a good service. Anyway I've not configured such a service so I cannot be of help.

gegere:
We all have our tools for the job. It has been really challenging to merge
logs locally for 5 containers spread across 5 datacenter locations, so
external logging was a proper solution and I can choose to keep logs for a
short or long duration per the storage provider.

Using this service is very easy and if the idea grows I will lend my
support.

tomav:
Out of the box, it should work locally. Using an external tool can't be a default but could be an option.
@00angus do you want to PR the folder modification (we should add test on it)

00angus:
@tomas as for the tests I'll add a check.
Unfortunately yesterday evening I've mistaken the commit comment
(referencing ENABLE_FAIL2BAN when instead I was committing changes on log
path changes!). So yes, in the next PR I'll also add the reference to the
path modification.

[...]

PR #116, opened 2016-04-01
All email-related services now logs in /var/log/mail/

00angus:
Implements  #113 

Fail2ban config has been updated accordingly.
Added also a check before copying spamassassin rule file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,18 +21,21 @@ run:
 		-e SA_KILL=3.0 \
 		-e SASL_PASSWD=testing \
 		-h mail.my-domain.com -t $(NAME)
+	sleep 15
 	docker run -d --name mail_pop3 \
 		-v "`pwd`/postfix":/tmp/postfix \
 		-v "`pwd`/spamassassin":/tmp/spamassassin \
 		-v "`pwd`/test":/tmp/test \
 		-e ENABLE_POP3=1 \
 		-h mail.my-domain.com -t $(NAME)
+	sleep 15
 	docker run -d --name mail_smtponly \
 		-v "`pwd`/postfix":/tmp/postfix \
 		-v "`pwd`/spamassassin":/tmp/spamassassin \
 		-v "`pwd`/test":/tmp/test \
 		-e SMTP_ONLY=1 \
 		-h mail.my-domain.com -t $(NAME)
+	sleep 15
 	docker run -d --name mail_fail2ban \
 		-v "`pwd`/postfix":/tmp/postfix \
 		-v "`pwd`/spamassassin":/tmp/spamassassin \
@@ -40,7 +43,7 @@ run:
 		-e ENABLE_FAIL2BAN=1 \
 		-h mail.my-domain.com -t $(NAME)
 	# Wait for containers to fully start
-	sleep 60
+	sleep 15
 
 fixtures:
 	# Sending test mails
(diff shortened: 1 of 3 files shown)


00angus:
Tests are failing!
That's strange as it has passed on my host all the checks.

I've repeated now my tests and : 

docker exec mail_fail2ban fail2ban-client status sasl
ERROR  Unable to contact server. Is it running?
And a few seconds later:
 docker exec mail_fail2ban fail2ban-client status sasl
Status for the jail: sasl
|- filter
|  |- File list:        /var/log/mail/mail.log
|  |- Currently failed: 0
|  `- Total failed:     0
`- action
   |- Currently banned: 0
   |  `- IP list:
   `- Total banned:     0

I'll look at it later.

00angus:
@tomav I've repeated the tests in my own environment (via Makefile) and it seems that it may happen that during some test not all services are already online. I've expanded the sleep time used in the makefile and now everything is working (on my own server): 
> 52 tests, 0 failures

I don't know if that could be the kind of problem we have on the Travis CI build. Maybe having another container to start up (mail_fail2ban) requires more time for the services to be online.

tomav:
```
not ok 47 checking fail2ban: ban ip on multiple failed login
# (in test file test/tests.bats, line 324)
#   `docker exec mail_fail2ban fail2ban-client status sasl' failed with status 255
# ERROR  NOK: ('sasl',)
# Sorry but the jail 'sasl' does not exist
not ok 48 checking fail2ban: unban ip works
# (in test file test/tests.bats, line 335)
#   `docker exec mail_fail2ban fail2ban-client set sasl addignoreip 127.0.0.1/8' failed with status 255
# ERROR  NOK: ('sasl',)
# Sorry but the jail 'sasl' does not exist
```

00angus:
@tomav the jail is there and is configured for sasl.
From my own experience - on my server - the problem is that when the tests are run the new `mail_fail2ban` container _hasn't fully started _all the services, fail2ban included. What do you think about it?
On my server I've increased the sleep time in the makefile between each rules and it complete the tests, now, as I've reported on April 2nd.

tomav:
Increasing time is not a good solution, we already have timeouts because of freshclam.
We should use a matrix of `env` and play dedicated tests.
https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix

That's probably what @gegere suggested.

00angus:
That's a possibility.
Do you have any suggestion? Because we could test fail2ban togheter with pop3 for example? What do you think?

tomav:
Never played with Travis matrix build and I'm currently out of time. Let's see if a contributor can help.

00angus:
@tomav I've seen that today's Reldeis patch has passed the CI tests. Would you like to try to restart this one? Probably now the Travis-CI resources are less busy!

tomav:
Restared.

00angus:
@tomav I've another idea

00angus:
@tomav sorry, I've another idea: I've tested on my server and make the job reach the end (on my server at least!).
In the makefile I've splitted the single `sleep 60` at the end of the **run** rule in 4 distinct `sleep 15` at the end of each _docker run_ command.
This way each container has some time to startup and reach an almost stable state (where almost all if not all the processes are up).

In fact the problem with the current version is that starting up all the processes togheter causes a lot of contention on the resources (iowait and so on). As an example, 60s after the startup of all the containers on the container `mail` only 20 processes out of 30 are ready!

By splitting 60s in 4x15s we don't add more time to the job :)
Is it ok for you?

tomav:
Ok, let's try!

00angus:
Ok - I'm going to submit another PR.

00angus:
@tomav ... sort of ... eureka!
Finally all the tests are ok :)

00angus:
@tomav I've done some changes and added a PR.
You will also find a check for the integration tests. Please consider that some of the tests have been modified to look at the new `/var/log/mail/mail.log` path so - even if indirectly - we are already checking that logs are now in the new location. Anyway, I've added a specific test to check that.
Moreover the commit message is - I hope - more clear.

PR #118, opened 2016-04-05
check .pem existence and concat to combined.pem with added eol

Reldeis:
Checks existince of cert, chain and privkey.pem.
Concats these files to combined.pem and adds a linebreak after each file
if there were none. This commit doesn't change the LE created .pem
files, so no unlinking on the host should happen (see
https://github.com/tomav/docker-mailserver/pull/111)

### DIFF
--- a/start-mailserver.sh
+++ b/start-mailserver.sh
@@ -120,17 +120,18 @@ fi
 case $DMS_SSL in
   "letsencrypt" )
     # letsencrypt folders and files mounted in /etc/letsencrypt
-      # add eol to all files before concatenation
-      sed -i -e '$a\' /etc/letsencrypt/live/$(hostname)/cert.pem
-      sed -i -e '$a\' /etc/letsencrypt/live/$(hostname)/chain.pem
-      sed -i -e '$a\' /etc/letsencrypt/live/$(hostname)/privkey.pem
+    if [ -e "/etc/letsencrypt/live/$(hostname)/cert.pem" ] \
+    && [ -e "/etc/letsencrypt/live/$(hostname)/chain.pem" ] \
+    && [ -e "/etc/letsencrypt/live/$(hostname)/privkey.pem" ]; then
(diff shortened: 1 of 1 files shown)


tomav:
Thank you @Reldeis

00angus:
@Reldeis Sorry, me again!
While trying your patch (I'm in v2 branch currently but the code is the same for letsencrypt) I've found a bug, at least it doesn't work on my server.
Here are the details.
The command you have introduced to combine certs is as follows:

     sed -e '$a\' -s "/etc/letsencrypt/live/$(hostname)/{cert,chain,privkey}.pem" > ...

The problem is that those `".." `around the command don't let the in-line expansion take place:

> root@mail:/# ls -1 "/etc/letsencrypt/live/$(hostname)/{cert,chain,privkey}.pem"
> ls: cannot access /etc/letsencrypt/live/mail.printk.it/{cert,chain,privkey}.pem: No such file or directory
> root@mail:/# ls -1 /etc/letsencrypt/live/$(hostname)/{cert,chain,privkey}.pem
> /etc/letsencrypt/live/mail.printk.it/cert.pem
> /etc/letsencrypt/live/mail.printk.it/chain.pem
> /etc/letsencrypt/live/mail.printk.it/privkey.pem

So even the sed command fails with:

      sed: can't read /etc/letsencrypt/live/mail.printk.it/{cert,chain,privkey}.pem: No such file or directory

I think we have to remove the `".."`.

Anyway, as for branch v2, it seems that for dovecot we don't need any more the combined .pem. I'm running my server with this version and the only certs I'm using right now are:

- fullchain.pem
- privkey.pem

It's the same as for Postfix.

Reldeis:
I'm on vacation right now, so i have no access to my PC. Could you PR your fix, @00angus? 

The curious thing is: it worked (and actually works right now) on my system.

00angus:
@reldeis PR submitted right now

Issue #120, opened 2016-04-06
Question on start-mailserver.sh and SASL_PASSWORD


00angus:
I've seen that in the code of start-mailserver.sh it has been introduced the use of SASL_PASSWORD env var. However when no passwd is being provided the script outputs a warning:

> Loaded '/tmp/postfix/main.cf'
> ==> Warning: 'SASL_PASSWORD' is not provided. /etc/postfix/sasl_passwd not created.

I find this warning not necessary. Could we leave simply the message without the warning?
Or it has been left there for a specific reason? For the parsing of main.cf the warning has been removed, for example, by the PR #104 .

tomav:
I agree. If optional, it should be an `Info`.

tomav:
Feel free to PR.

tomav:
Fixed.

PR #122, opened 2016-04-07
Changed $SASL_PASSWD to $SASL_PASSWORD

aforty:
The README lists the environmental variable as `SASL_PASSWORD` while the startup script was actually looking for `SASL_PASSWD`. The script now looks for `SASL_PASSWORD` which is consistent with documentation.

Also changed warning to info, as per discussion in #120.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ run:
 		-e SA_TAG=1.0 \
 		-e SA_TAG2=2.0 \
 		-e SA_KILL=3.0 \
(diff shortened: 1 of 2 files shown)


tomav:
Hi @aforty and thank you.
The easiest thing would have been to change de documentation no? If users have configured their mailserver, they will get a BC break.

aforty:
I liked the variable name being spelled out better. I can change it back and change doc instead. Your call, just my 2 cents.

Issue #109, opened 2016-03-26
Docker MailServer V2


tomav:
Here is the list of what should be done in version 2 of this image

- [x] Reduce image size
- [ ] Move to Alpine or other lightweight image
- [X] Use encrypted user database + generator from clear password text file (#123) 
- [x] Comment `start-mailserver.sh` to help maintenance
- [x] Move from `/tmp` folder to `/var/docker-mailserver` or equivalent
- [x] Rename files (some are `.cf`, some are not)
- [x] Rename env using convention
- [x] Move static files to `target` folder
- [x] `git clone` + `docker-compose up` should work _out of the box_ 
- [x] Add tests on SSL certificates, at least for letsencrypt
- [x] Clean Fail2ban configuration
- [ ] Refactor integration tests workflow in Travis
- [X] Add Sieve
- [x] Review documentation (`README.md` and Wiki)
- [ ] Find another project name :blush: 
- [ ] Create Github organization
- [ ] Find 1 or 2 core contributors to help project maintenance

mgansler:
Hi!

Take a look at https://gitlab.martingansler.de/docker/mail
I basically took your postfix configuration and integrated it in an alpine/s6 based image.
Currently it doesn't result in a working server but it should help you figuring out how to start the services etc.
Unfortunately I don't have much time at the moment, so that all I could hack together in the last 30 minutes or so.

tomav:
The challenge for us if to find a lightweight and secure distro with all our needed packages.
After a first look with @KyleOndy it seems that some packages are not available on Alpine.

mgansler:
Do you remember which ones are missing?

00angus:
First of all ... what a great project @tomav :) 
Really! I'm enjoying using it, while trying to contribute some changes.
One of the key issue is the 'encrypted database': well, I'm currently running a modified docker-mailserver image where all _cyrus-sasl_ and _courier-*_ has been replaced with **dovecot**: it builds and it WORKS :)

My current implementation:

- postfix auth for sending out emails is now based on dovecot (sasl)
- imap, pop3 are all served by dovecot, and SSL is configured accordingly to DMS_SSL env
- imap, pop3, postfix all use (thanks to dovecot) the same user-database with passwords encrypted
- users are still managed via `accounts.cf` but now the passwords are encrypted
[...]

PR #123, opened 2016-04-08
Dovecot based mailserver (#109)

00angus:
This is the Dovecot-based version of the mailserver.
In the README you will find enough information (I hope!) to configure and startup the server (and for the generation of the passwords).
I'll leave the description of this version to one of the commit messages :)

    Dovecot based version of the mailserver.
    Courier and Cyrus Sasl have been removed and substituted with Dovecot which now
    handle authentication for Postfix, Imap and Pop3, with support for SSL.
    This allow the use of several encryption schemes for the password as well as a
    single user db.
    OpenDKIM keys can now be provided at the startup and will be used instead of
    generating new ones (so that you don't have to change your DNS configuration).

    This version builds correctly on Docker but no integration tests have been
    reworked to accommodate Dovecot instead of Courier and Cyrus Sasl. As such
    at present no automatic tests can be executed.

As I've already written somewhere else It will be a good idea to take a look at the dovecot configs and see where things can be optimized/secured. I've also checked connections to/from popular email services and I've also tested `checktls.com`, `mail-tester.com` (score 9/10 because of DMARC record missing on my dns) etc.
One thing to note: as of this version, dovecot config files are copied into the image at build-time. Probably in the future we could setup it in the startup-mailserver.sh.
Take a look at the different commits.
Enjoy!


tomav:
Thank you @00angus.
I'm gonna merge this into `v2`.
We'll have to add tests on top of this.

Issue #125, opened 2016-04-09
imap mail access


awb99:
I have a container running;  and smtp sending works, but getting data via imap does not work. There are no folders displayed at  all. I just added the accounts.cf file .

My computer is: backoffice.newstrading.org[104.222.96.51]:
This is the log I get via  docker logs mymail_mail_1
http://pastebin.com/QXN3pwVW

I believe it misses all the imap log entries.

```
cat docker-compose.yml 

maildata:
  image: ubuntu
  volumes:
    - /var/mail
  command: /bin/true

mail:
  #restart: always
  #build: ./docker-mailserver
  image: tvial/docker-mailserver
  hostname: backoffice
  domainname: newstrading.org
  volumes_from:
   - maildata
  ports:
  - "25:25"
  - "143:143"
  - "587:587"
  - "993:993"
  volumes:
  - ./spamassassin:/tmp/spamassassin/
  - ./postfix:/tmp/postfix/
  environment:
  - ENABLE_POP3=1
```

andrealinux1:
I also experienced this problem yesterday after having pulled the latest version from dockerhub. I thought that the issue was due to a misconfiguration (I moved some folders lately after having refactored my deploy method of the services running on my server, and i thought I broke something). I tried building a old version (I cloned the repo and did a checkout of 5-6 commits ago, built it locally) and it returned to work. So I think that the issue has been introduced with some of the last commits/merges of pull request. When I'll have some times (hopefully tomorrow) I'll try to build adding one commit at a time to see if I can spot the commit (and the modification) that is causing this issue. Anyone else that has this problem after having updated and has an idea of what goes wrong? Or converserly  anyone that has updated and do not have the problem?

awb99:
What I noticed is that the complete filesystem of the host appears mounted in the guest mail container... Perhaps this is related?

00angus:
@awb99 you are right, the **logs are missing entries for imaps connections**.
Some suggestions:

- are the services up and running in the container? **The log say they have been started**, but pls check it out
- if the services are up and running from the logs it's clear the container isn't receiving any request on  imap/s ports: are those ports available from outside your host (firewall issues?).

awb99:
@00angus I have been able to telnet  into the imap server, and I see all the normal messages. Also my client app can connect, just authentication fails.

auguster:
The new image seems to have broken the SSL Certificates, I have this line that has appear in my logs:

[...]

PR #129, opened 2016-04-11
Fix bug when combining certs for Letsencrypt.

00angus:
See also Issue #125 
Patches for @Reldeis PR #118


tomav:
Restarted job.

Issue #121, opened 2016-04-07
wrong vhost from virtual with side effect


auguster:
I'm trying to configure postfix to accept multiple domains, so that say all map to the same corresponding mailboxes. For example, I want the mails for userA@domainB.tld and userA@domainA.tld to go into the same mailbox. I haven't find how to do that yet.

Anyway, during my tests I noticed there is something wrong with the way vhost is generated from virtual. In the following example the server's hostname is mail.domainA.tld the given path are from within the docker container:

/etc/postfix/virtual:
```userA@domainB.tld userA@domainA.tld```

/etc/postfix/vhost:
```
domainA.tld
domainB.tld userA
```

This has the side effect of causing opendkim to regenerate the key for domainB.tld at each launch. This is due to the fact that start-mailserver.sh is looking for the file `/etc/opendkim/keys/domainB.tld userA/mail.private`, doesn't find it, ask opendkim to generate one for "domainB.tld userA". Opendkim uses only the "domainB.tld" argument and so overwrites the previous key for "domainB.tld".

**Comment by tomav** [comment]
Perhaps @bilak may help on this.

**Comment by bilak** [comment]
Hi, try to mount you opendkim configuration from volume and it should be ok.

**Comment by tomav** [comment]
Hmm, the problem is that `domainB.tld userA` is wrong. It should only be the domain name.

@auguster could you post a link to a repo that we can clone and start to see what happens?

Thank you.

**Comment by auguster** [comment]
@bilak I already did that, the issue I describe overwrites the existing files anyway.
@tomav you want me to commit the machine and post the docker image somewhere or you just want a copy of /etc/postix ?

Normally you can easily reproduce the issue:
accounts.cf
```
userA@domainA.tld|password
```
for the two other files, the content is in my initial post.

**Comment by tomav** [comment]
I'll try with the information provided in this post.

**Comment by tomav** [comment]
Just tried from latest image and here what I have:

```
➜  docker-mailserver git:(master) ✗ docker-compose up -d
Creating dockermailserver_mail_1

➜  docker-mailserver git:(master) ✗ docker exec dockermailserver_mail_1 cat /tmp/postfix/accounts.cf
userA@domainA.tld|password

➜  docker-mailserver git:(master) ✗ docker exec dockermailserver_mail_1 cat /tmp/postfix/virtual
userA@domainB.tld userA@domainA.tld

[...]

Issue #57, opened 2016-01-20
Does it allow multiple domains?


citec:
Hi, I have a question. Does this stack allow multiple domains? As it open the common ports, I wonder if it allows multiple domains, so I can deploy just one instance of it without the need for a proxy to separate domains.

citec:
When I say "multiple domains" I mean: can I set account for multiple domains in the postfix/accounts.cf file? What happend then with the ssl certificate? should I put just one certificate for the main domainname configured in the docker-compose.yml or do I need a ssl cert for each account domain?
thanks

tomav:
Hi @citec, this is more a `postfix` question than a `docker`one.

The answer is yes, postfix is multidomain.
You can manage as many domains as you want using Postfix virtual domains that is pre-configured on this docker image.
The `accounts.cf` is the key.

You also should have a look to Postfix configuration: http://www.postfix.org/VIRTUAL_README.html

Regarding your SSL question, these answer will help you: http://serverfault.com/questions/344038/virtual-domains-with-postfix-and-ssl

citec:
Ok, I understand now: I will set the autoconfig for my clients, or tell them, to use mail.mycompany.com
Thanks

PR #131, opened 2016-04-11
Removed IFS + Added vmailbox config

auguster:
Here is a fix for issue #121 and I went a bit further and added support for custom vmailbox settings (usefull for issue #57).

Now, you can define a `/tmp/postfix/vmailbox` (from within the container) that specifies extra virtual mail boxes. It is describe here: http://www.postfix.org/VIRTUAL_README.html#virtual_mailbox

Custom vmailbox settings are appended to the end of `/etc/postfix/vmailbox`, so after the generated content from `accounts.cf`.

What it allows is the support of routing all the other mail from a domain to a mail box.

For example:
accounts.cf:
```
user@domain.tld|password
contact@domain.tld|password2
```

vmailbox:
`@domain.tld domain.tld/contact/`

In this example, all the mails **other** than `user@domain.tld` and `contact@domain.tld` will be placed in `contact@domain.tld`'s mailbox. Be aware that the trailing "/" is required !

### DIFF
--- a/start-mailserver.sh
+++ b/start-mailserver.sh
@@ -58,6 +58,9 @@ if [ -f /tmp/vhost.tmp ]; then
   cat /tmp/vhost.tmp | sort | uniq > /etc/postfix/vhost && rm /tmp/vhost.tmp
 fi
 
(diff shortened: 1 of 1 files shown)


tomav:
Thank you @auguster

PR #134, opened 2016-04-15
Makefile cleanup

KyleOndy:
As I start digging intot this branch these two makefile tweaks jumped out at me.


tomav:
Right! Thank you.

PR #135, opened 2016-04-16
Feat userdbs : an update 

00angus:
I've changed the way the previous #107 worked and implemented received suggestions.
The startup logic for user configuration is now as follows:

     * if `/tmp/postfix/accounts.cf` available: interrupt and stop immediately the container 
       with an error message;
     * if `/tmp/postfix/accounts-db/{userdb,sasldb2}` available: DBs copied to the image,
       users setup completed;
     * otherwise no mail account will be setup.

For security reason accounts.cf file must not be provided anymore to the container or it will stop with an error message! User configuration is only allowed via pre-configured user databases. See README.md for instructions.


tomav:
That's great @00angus!

Could you add a test regarding the fact that the container should stop if `accounts.cf` is provided.
I suppose that copying `test/accounts.cf`and `test/virtual`is not needed anymore.

Also, the generation of your db should be tested.

Otherwise, really nice work!

00angus:
By generation...  do you mean the utility script?
In that case I'll have some questions on BATS tests for you :)

tomav:
Feel free to ask 😄

tomav:
But there was something I misunderstood regarding the user db.
The generated db is already tested using authentication tests so it's probably fine.
But if you find useful to add some tests, please do.

00angus:
The one I've added here I think is sufficient. I've introduced a temporary container in tests : I'm not sure this is the best overall solution, the main problem being:

1. I've someway hardcoded the image to be used (or bettere a dependency on the name of another container: if `mail `container in `Makefile `change name, a test needs to be modified). I haven't found a better solution;
2. if the container doesn't exit on its own it would require a manual intervention

If someone finds a better solution pls let me know :)

test/tests.bats
```diff
@@ -195,7 +195,15 @@
 # accounts
 #
 
+@test "checking accounts.cf: if present shouldn't start the container" {
+  cp `pwd`/postfix/disab.accounts.cf `pwd`/postfix/accounts.cf
+  run docker run --rm --name test_on_accounts -v "$(pwd)/postfix":/tmp/postfix `docker inspect --format '{{ .Config.Image }}' mail`
+  [ "$status" -eq 1 ]
+  [ "${lines[1]}" = "SECURITY WARNING ==> ABORTED startup !" ]
```
tomav: I'll be more explicit in the exit message.
Something like

```
#
# ERROR:
# Container startup has been aborted because password file containing clear entries has been provided
# Please have a look to the documentation to provide a crypted mail account database.
#
```

tomav: Sorry, haven't read code above. It's Ok!

tomav:
The implementation is correct.

PR #136, opened 2016-04-17
V2 CI-tests

00angus:
Adapted the user account setup for the test environment to the new v2 approach (encrypted passwords).
Modified integration tests to adapt to dovecot.

Need to be worked out:
- user accounts checks: dir layout under dovecot probably include dynamic filenames.
  How can we handle that ?
- fail2ban container cannot ban via iptables if we do not run it with special
  privileges.

Please note @tomav : tests involving the use of "nc -w 1" on dovecot services may need a longer timeout, in particular when testing failures as dovecot do not answer immediately (by design? who know!). It may even depends on my busy host.

### DIFF
--- a/test/config/clear.postfix-accounts.cf
+++ b/test/config/clear.postfix-accounts.cf
@@ -1,0 +1,2 @@
+user1@localhost.localdomain|mypassword
+user2@otherdomain.tld|mypassword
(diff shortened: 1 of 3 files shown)


tomav:
#109

tomav:
Yes, test failed on dynamic folder. I think we should only grep the expected folders (= not all output lines)

tomav:
I let you fix this, I'll merge, and and I'll get your code to improve fail2ban tests (I have something working locally)

00angus:
No. 44 and 45 (fail2ban) are failing :/
All the others are working now :)

tomav:
I'm gonna check for `fail2ban`. We're close :-)

PR #138, opened 2016-04-18
Fixes paths for SSL setup.

00angus:
Fixes error from logrotate when fail2ban is disabled (failure
when reloading the service).
BTW - `/tmp/docker-mailserver/spamassassin/rules.cf` is a valid path for spamassassin rules? Because I don't have a supposed path layout for all the use case :)

### DIFF
--- a/target/start-mailserver.sh
+++ b/target/start-mailserver.sh
@@ -176,10 +176,10 @@ case $DMS_SSL in
 
   "custom" )
     # Adding CA signed SSL certificate if provided in 'postfix/ssl' folder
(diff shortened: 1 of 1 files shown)


tomav:
No, it should be in `config/`. I'm open to suggestion :-)

00angus:
accounts are now looked at the following position: `/tmp/docker-mailserver/postfix-accounts.cf`
So if I understand, spamassassin should be looked at: `/tmp/docker-mailserver/spamassassin-rules.cf`

If this is correct I've to change my PR, so please do not merge this one. I'll push another PR.

Issue #132, opened 2016-04-13
opendkim howto ?


auguster:
I mounted the `/etc/opendkim` folder on the host to ensure persistence of the DKIM signature, I set the corresponding public key in the DNS using content from `/etc/opendkim/keys/<my domain>/mail.txt`. But while testing it I noticed the mail weren't signed.

In fact, I noticed this in my logs:
`opendkim: /etc/opendkim.conf: refile:/etc/opendkim/TrustedHosts: dkimf_db_open(): No such file or directory`

It coming from:
`/etc/init.d/opendkim start` ([L292 of start-mailserver.sh](https://github.com/tomav/docker-mailserver/blob/master/start-mailserver.sh#L292))

Is there anything specific to do to enable DKIM ?

EDIT: BTW, I'm not familiar with DKIM but shouldn't there be an `/etc/opendkim/TrustedHosts` file ? It seems to be missing.

auguster:
Ok, I got it to work by manually adding the `/etc/opendkim/TrustedHosts` with the correct content. It was easier than I thought.

I noticed it's not part of the opendkim initialization process of the script I guess it should be. I will make a fix and a PR soon.

00angus:
@auguster sorry for the late reply... anyway...
I've that same setup.
I've copied the content from a first run of the container and made it persistent, that way all the needed file are ok :) 
I've added a commit to the v2 branch that handles host based opendkim configurations.
If interested I can merge it on master, take a look and let me know :)

auguster:
@00angus  Don't worry, I was supposed to make a PR soon anyway ^^

I think the file should be generated by the start script if missing. It should just contain `127.0.0.1` and `localhost` anyway. Manually copying files feels `manual` and error proned if not well documented (or if the documentation gets ignored).

I still hope on making a quick fix soon if my schedule allows.

00angus:
[Here](https://github.com/tomav/docker-mailserver/blob/v2/target/start-mailserver.sh#L94) is the patch take a look if you want.

The only manual part is the initial creation of the config that should be made persistent. It's sufficient to add a /usr/local/bin/ script and call it like we call the script for the generation of ssl certs.

auguster:
From what I read that script doesn't create `/etc/opendkim/TrustedHosts` if the file is missing.

[...]

PR #140, opened 2016-04-19
/etc/opendkim/TrustedHosts creation if missing

auguster:
Following issue #132, the script `start-mailserver.sh` now creates and populates `/etc/opendkim/TrustedHosts` with localhost information.

This allow for mounting an empty directory from the host for DKIM data persistence.


tomav:
Thank you!

auguster:
You're welcome, it wasn't hard at all ^^

aguerrave:
Hello for all, Excellent work, you beginning to test, you have a howto?

auguster:
@aguerrave I did some local test, it seems to work fine. As you can see the code is rather simple so I think it's safe ;)

For the howto, just mount an empty directory from host for `/etc/opendkim` and you'll see that the servers create a basic `/etc/opendkim/TrustedHosts` with localhost informations:

```
localhost
127.0.0.1
```
The discussion open this is mainly in issue #132 for extended details.

aguerrave:
Ok 
i have a domain valid for test

Issue #137, opened 2016-04-18
cron.daily for spamassassin


00angus:
I'm experiencing problems with the distro-provided cron job for spamassassin in `cron.daily`.
It seems that the implemented `reload function` is not working as is trying to reload spamassassin via `invoke-rc.d` instead /etc/init.d/script 

Ubuntu 14.04 is supporting both methods ... but invoke-rc.d isn't working for me:

     # invoke-rc.d spamassassin status
     invoke-rc.d: policy-rc.d denied execution of status.
     invoke-rc.d: emulating initscript action "status", returning "unknown"
     # /etc/init.d/spamassassin status
     * spamd is running

The problem is the policy-rc.d script. 
Since policy-rc.d on docker images is setup to 

    101 - action forbidden by policy

and [here](http://askubuntu.com/questions/365911/why-the-services-do-not-start-at-installation) you will find why this is good, I propose to modify the spamassassin cron.daily script to only invoke the script from init.d :)

Do you see other solutions?

tomav:
Reproduced. I agree on the proposed solution.

PR #144, opened 2016-04-20
Fix a problem in the reload function of spamassassin cron.daily script.

00angus:
(see issue #137 on github).


tomav:
Thank you @00angus

PR #147, opened 2016-04-21
Dockerfile optimization to reduce image size

auguster:
I noticed the Dockerfile wasn't well written: files have to be removed in the same layer (for example  same RUN) it was added to save on image size.

I therefor merged the apt-get lines.

The result is an image size down to 489.7 MB instead of 559.7 MB (-12,5%). It's not much but it's still something ;-)


tomav:
Thank you @auguster

PR #151, opened 2016-04-22
Modify the Dockerfile attempting to use alpine

KyleOndy:
# BROKEN & INCOMPLETE

So this is the core of what I had when working towards alpine.

I didn't commit my changes to `start-mailserver.sh` as my images never got that far. My main concern was getting dovecot to start.

Trying to start it fails because alpines [init systen](http://wiki.alpinelinux.org/wiki/Alpine_Linux_Init_System) `OpenRC` [does not exist](https://github.com/gliderlabs/docker-alpine/issues/42) in the alpine's docker image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,7 +29,7 @@ RUN apk add shadow --update-cache --repository http://dl-3.alpinelinux.org/alpin
 # Configures Dovecot
 RUN sed -i -e 's/include_try \/usr\/share\/dovecot\/protocols\.d/include_try \/etc\/dovecot\/protocols\.d/g' /etc/dovecot/dovecot.conf
 RUN cp -a /etc/ssl/dovecot/server.pem /etc/dovecot/dovecot.pem
-RUN mkdir /etc/dovecot/private && cp -a /etc/ssl/dovecot/server.pem /etc/dovecot/private/dovecot.pem
+RUN mkdir /etc/dovecot/private && cp -a /etc/ssl/dovecot/server.key /etc/dovecot/private/dovecot.pem
 ADD target/dovecot/auth-passwdfile.inc /etc/dovecot/conf.d/
 ADD target/dovecot/10-*.conf /etc/dovecot/conf.d/
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,27 +1,50 @@
-FROM ubuntu:14.04
+FROM gliderlabs/alpine:3.3
 MAINTAINER Thomas VIAL
 
 # Packages
-RUN apt-get update -q --fix-missing
-RUN apt-get -y upgrade
-RUN DEBIAN_FRONTEND=noninteractive apt-get -y install --no-install-recommends \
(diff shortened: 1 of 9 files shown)


KyleOndy:
@mgansler Here is what I was working on.

Discussion relates to #109

00angus:
@KyleOndy that one you posted is an important step: from here we can better understand what we can obtain with alpine.

To dig further I've a question on your quote:
 
> My main concern was getting dovecot to start. [snip] Trying to start it fails because alpines init systen OpenRC does not exist in the alpine's docker image.

From my experience, on a docker image we don't need an init system. If you take a look to a running instance of docker-mailserver:master o v2 you will not find an init process.
In fact, to startup dovecot it's sufficient to run:

`/usr/sbin/dovecot -c /etc/dovecot/dovecot.conf`

Given a proper setup/config, it will work without an init system. The same apply to postfix.

So, you have already included dovecot in the dockerfile with apline (which I've not tested right now). What's wrong with running it if you apply docker-mailserver:v2 setup ?

KyleOndy:
I am having difficulty reproducing the issues I were having. 

I may dig into this tonight again.

KyleOndy:
I think I might be making slow progress. By removing everything else, I got dovecot to start. Test still fail all over, but I'm going to see what I can do.

KyleOndy:
So this is where I am getting hung up things.

Dovecot stats fine
```
Regenerating postfix 'vmailbox' and 'virtual' for given users
user 'user1' for domain 'localhost.localdomain' with password '********'
user 'user2' for domain 'otherdomain.tld' with password '********'
Fixing permissions
Creating /etc/mailname
 * Starting dovecot services
Listing users
user1@localhost.localdomain
user2@otherdomain.tld
Starting...
Apr 22 22:56:57 master: Info: Dovecot v2.2.20 (46a35dcdb936) starting up for imap, pop3, lmtp
Apr 22 22:56:58 ssl-params: Info: Generating SSL parameters
Apr 22 22:56:59 ssl-params: Info: SSL parameters regeneration completed
```

When I run the test suite it hangs.
```
checking imap: server is ready with STARTTLS
```

Looking at the test file the command for this test is
```
docker exec mail /bin/bash -c "nc -w 5 -i 10 0.0.0.0 143 | grep '* OK' | grep 'STARTTLS' | grep 'ready'"
```

which when run manually yields 
```
* OK [CAPABILITY IMAP4rev1 LITERAL+ SASL-IR LOGIN-REFERRALS ID ENABLE IDLE STARTTLS AUTH=PLAIN AUTH=LOGIN] Dovecot ready.
```

So the test should pass! Yay. However netcat never terminates the connection. I don't know if this is because of the version alpine is using or what.

KyleOndy:
It also seems the text fixtures are returning an error code of `1`, causing the make command to fail.

Could it be the busybox version of `nc` causing these problems?

KyleOndy:
`50 tests, 32 failures` I think I'm done for the night.

KyleOndy:
`50 tests, 27 failures`

PR #163, opened 2016-04-26
Changed few docker commands used in tests:

00angus:
Removed hardcoded image name in tests.

Tests have not been changed, BTW.
Possible problem: locally the current v2 version fails CI tests, so I don't know if even this branch will build correctly on Travis. 
Anyway I've not changed the tests. Let see.


tomav:
Oh yes, thank you @00angus.

PR #165, opened 2016-04-27
Adapted Postfix configuration to block typical spam sending

### LINKED ISSUE #161: Discussion: more restrictive Postfix configuration to block spam early
I am using an issue here to discuss the possibility to have a more restrictive ***Postfix*** configuration to block spam before it is even looked at by Spamassasin etc.

I recently switched to **docker-mailserver** and I observed an increase in daily spam mails that actually land in my INBOX. Looking at the diff between the container's postfix configuration I noticed that the container's configuration doesn't impose many restrictions on the sending server. But there are some configurations option which block many typical spam sending server (dynamic IPs etc.) using some more restrictive configurations options as described here: https://www.webstershome.co.uk/2014/04/07/postfix-blocking-spam-enters-server/.

In production I noticed that configurations described as those in the blog post actually the number of spam mails that hit the server and thus don't need to be processed by Spamassasin (and eventually getting in the INBOX).

Does something like that makes sense as a default configuration? I think we could figure out some settings which should work for most mail servers. And optionally this could be made an startup option?

Thanks!

stonemaster:
mail servers using an enhanced client, sender and helo
restriction configuration.

The configuration has been adapted using this blog post:
https://www.webstershome.co.uk/2014/04/07/postfix-blocking-spam-enters-server/

Basically mail servers having invalid configuration (as e.g. sending
from and dynamic IP or a misconfigured hostname) will have their
mails rejected.

Additionnally three RBL servers are used to detect spam sending
IPs: dnsbl.sorbs.net, zen.spamhaus.org and bl.spamcop.net.

The results of a 12h test drive using a 100+ daily spam
mail account (SpamAssasin was always enabled, just counting
delivered mails to inbox not counting what SA detected):
- Before: 34 incoming mails
- Afer change: 6 incoming mails (82% reduction)

Fixes #161.

### DIFF
--- a/target/postfix/main.cf
+++ b/target/postfix/main.cf
@@ -24,10 +24,6 @@ smtpd_tls_key_file=/etc/ssl/private/ssl-cert-snakeoil.key
 #smtp_tls_CAfile=
 smtpd_tls_security_level = may
 smtpd_use_tls=yes
-smtpd_relay_restrictions = permit_mynetworks permit_sasl_authenticated defer_unauth_destination
-smtpd_recipient_restrictions = permit_sasl_authenticated, permit_mynetworks, reject_unauth_destination
-smtpd_client_restrictions = permit_mynetworks, permit_sasl_authenticated, reject_unauth_destination
-smtpd_sender_restrictions = permit_mynetworks
 smtp_tls_security_level = may
 smtp_tls_loglevel = 1
 tls_ssl_options = NO_COMPRESSION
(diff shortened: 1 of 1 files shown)


target/postfix/main.cf
```diff
@@ -38,6 +34,17 @@ smtpd_tls_mandatory_ciphers = high
 smtpd_tls_mandatory_protocols = !SSLv2, !SSLv3
 smtpd_tls_exclude_ciphers = aNULL, LOW, EXP, MEDIUM, ADH, AECDH, MD5, DSS, ECDSA, CAMELLIA128, 3DES, CAMELLIA256, RSA+AES, eNULL
 
+# Settings to prevent SPAM early
+smtpd_helo_required = yes
+smtpd_delay_reject = yes
+smtpd_helo_restrictions = permit_mynetworks, reject_invalid_hostname, permit
+smtpd_relay_restrictions = permit_mynetworks permit_sasl_authenticated defer_unauth_destination
+smtpd_recipient_restrictions = permit_sasl_authenticated, permit_mynetworks, reject_unauth_destination,
+    reject_unauth_pipelining, reject_invalid_hostname, reject_non_fqdn_hostname, reject_unknown_recipient_domain,
```
tomav: `reject_non_fqdn_hostname` is for postfix <2.3 no? It should be `reject_non_fqdn_helo_hostname`.

Same for `reject_invalid_hostname` => `reject_invalid_helo_hostname`

Have a look to http://www.postfix.org/postconf.5.html

tomav:
Hi @stonemaster, thank you for this enhancement on postfix configuration.
I commented your commit because I think that some of `postconf` keys are for `postfix <2.3`
Could you please have a look to http://www.postfix.org/postconf.5.html ?

Thank you.

tomav:
Regarding the tests and CI, don't worry it happen often on this test.

```
not ok 10 checking imap: server is ready with STARTTLS
# (in test file test/tests.bats, line 56)
#   `[ "$status" -eq 0 ]' failed
```

I'll have to find the reason. When you'll submit your modification, tests will be restarted.

stonemaster:
Thanks for the review! That's a nice catch. I have been using those hostname* configuration options since _ever_ and never noticed that have been deprecated for ages :-) It's funny though that nearly every blog post on anti-spam postfix configuration uses the deprecated configuration options. I also checked that the other options are recent and not deprecated.

tomav:
Thank you @stonemaster !

Issue #87, opened 2016-02-26
Sieve filters


echo511:
Is it a desired to-do? Or is it off-topic for this project?

tomav:
Hi @echo511, what do you mean?
I'm not an expert but we already have amavis and spamassassin.
Let me know.

echo511:
Well, that's for eliminating the unwanted. Sieve is for mail ordering based on user specific rules. For example move mail marked as spam by spamassassin to Junk folder etc.

Look here: http://wiki2.dovecot.org/Pigeonhole/Sieve/Examples

echo511:
Ok, it may not be possible to set up Courier-Sieve, no support from Courier. Maby move to Dovecot instead? I'll look into it. Let me know if Sieve-Dovecot sounds good to you. If so I'll send a pull.

tomav:
If we can keep the root concept of using only versionable configuration( no *sql database), why not.
I'm just wondering if dovecot allows to do that.

echo511:
By default it uses system accounts but can be configured to use userfile or sql or whatever. So no problem

00angus:
Dear all, still a work in progress but we have dovecot in v2 :)
We can now add sieve. Someone interested?

stonemaster:
I've always configured my dovecot server to enable Sieve by uncommenting the file following line in `/etc/dovecot/90-sieve.conf`:

`
sieve = file:~/sieve;active=~/.dovecot.sieve
`

That will look for a `~/.dovecot.sieve` in the user's home directory and enable appropriate sieve rules. 

Is there anything more comfortable or elegant needed? For me something like this would be enough given it is possible to configure the **dovecot.sieve** file to the `/var/mail/%d/%u` directory structure.

tomav:
If you want to PR something, feel free to do it :-) Just add necessary tests.

tomav:
Do it on v2 of course.

stonemaster:
Okay I'll have a look :-) And yes v2 of course.

PR #166, opened 2016-04-28
Implement basic sieve support using Dovecot.

stonemaster:
The dovecot-sieve plugin is installed and configured to apply sieve
as soon as a .dovecot.sieve file is encountered in the virtual user's
home directory (that is /var/mail/${domain}/${username}/.dovecot.sieve).

Transport has been changed in the postfix configuration to use
Dovecot LDA (see http://wiki.dovecot.org/LDA/Postfix) to actually
enable sieve filtering.

Tests have been added.

Addresses #87.

### DIFF
--- a/test/tests.bats
+++ b/test/tests.bats
@@ -414,7 +414,7 @@
 # sieve
 #
 
-@test "checking sieve: user1 should have received 1 in folder INBOX.spam" {
+@test "checking sieve: user1 should have received 1 email in folder INBOX.spam" {
   run docker exec mail /bin/sh -c "ls -A /var/mail/localhost.localdomain/user1/.INBOX.spam/new | wc -l"
   [ "$status" -eq 0 ]
   [ "$output" = 1 ]
--- a/test/tests.bats
+++ b/test/tests.bats
@@ -129,9 +129,9 @@
 }
 
 @test "checking smtp: delivers mail to existing account" {
-  run docker exec mail /bin/sh -c "grep 'status=sent (delivered to maildir)' /var/log/mail/mail.log | wc -l"
+  run docker exec mail /bin/sh -c "grep 'status=sent (delivered via dovecot service)' /var/log/mail/mail.log | wc -l"
   [ "$status" -eq 0 ]
-  [ "$output" -eq 2 ]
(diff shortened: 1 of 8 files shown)


stonemaster:
I didn't add documentation specifically. I thought about adding a new entry to the FAQ on how to get started with sieve.

test/tests.bats
```diff
@@ -409,3 +409,13 @@
   [ "$status" -eq 0 ]
   [ "$output" = "my-domain.com" ]
 }
+
+#
+# sieve
+#
+
+@test "checking sieve: user1 should have received 1 in folder INBOX.spam" {
```
tomav: > should have received 1 **email**

tomav:
Thank you @stonemaster, I just let you fix the missing word in the test.
I'll merge as soon it's fixed.

tomav:
Regarding the doc, yes a page `Configuring Sieve filters` in the Wiki will be fine with an example and a link to documentation.

00angus:
cool. It could be useful to be able to provide user-prepared sieve files to be copied in place during setup.
sort of: user@domain.tld.dovecot.sieve if found during startup will be moved into corresponding home into .dovecot.sieve.
What do you think about it? Other ideas?
BTW I was working on the same thing but using lmtp instead of lda... performance wise do you see differences?

stonemaster:
@00angus that's a nice idea. I would suggest have a differen pull request for that feature though.
Regarding LMTP: I have no idea whether this actually makes a difference. LDA is the older (maybe proven?) method. Have you seen any performance tests regarding this?

stonemaster:
@tomav Added a new Wiki page and updated README with a link to it btw.

tomav:
Regarding user specific config, let's open another issue / enhancement.

00angus:
LMTP: needed in case smtp wants to deliver to a different host/container, better descriptive information in case of failures than LDA, LMTP may add additional headers I think ... 
Performance wise, on high email load I think LMTP should perform better as it's a running process sitting in the background waiting for connections: on the contrary LDA will need to spawn a delivery process for each new message. 
Anyway both solutions for me work :) 

If performance becomes a problem for someone please let us know ;)

PR #168, opened 2016-04-28
Sieve filters can now be installed at user setup.

00angus:
Changed Dockerfile to include a missing configuration file for dovecot (mailboxes).
Changed info message when no postfix-main.cf are provided for overriding postfix config (was referencing the older file name).
Moved Postfix setup after DKIM/DMARC/SSL setup, near the override postfix setup.
This one in particular is only aesthetic if you look at the order messages are outputted at startup:

> user 'user2' for domain 'otherdomain.tld' with password '********'
> Postfix configurations
> No DKIM key provided. Check the documentation to find how to get your keys.
> Loaded 'config/postfix-main.cf'

I thought that `Postfix configurations` should appear before the `loading of postfix-main.cf` instead of being interleaved by DKIM, SSL etc.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update -q --fix-missing && \
 RUN sed -i -e 's/include_try \/usr\/share\/dovecot\/protocols\.d/include_try \/etc\/dovecot\/protocols\.d/g' /etc/dovecot/dovecot.conf
 RUN sed -i -e 's/#mail_plugins = \$mail_plugins/mail_plugins = \$mail_plugins sieve/g' /etc/dovecot/conf.d/15-lda.conf
 ADD target/dovecot/auth-passwdfile.inc /etc/dovecot/conf.d/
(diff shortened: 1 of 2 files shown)


stonemaster:
I'd suggest documenting this behaviour on the Wiki page at https://github.com/tomav/docker-mailserver/wiki/Configuring-Sieve-filters

00angus:
ok - I updated that page of the wiki:

- added infos on how to provided filter at setup/startup
- added another example of filter I find useful (forwarding)
- included another link to a web page with handy examples

tomav:
The page should be "Configure" and not "Configuring" if we want do like we've done so far (it's just a detail but we should change it before releasing `v2` officially).

stonemaster:
Just renamed the wiki page accordingly in pull #170.

stonemaster:
@00angus the changes look very good and the documentation too. I'd approve :-)

tomav:
Thanks :-)

tomav:
I'ill fix the link later.

PR #169, opened 2016-04-29
Configure dovecot's managesieve plugin when the environment variable

### LINKED ISSUE #167: Add ManageSieve support to allows server-side filtering configuration for users
Now that Sieve is configured with Dovecot it would be possible to install ManageSieve (http://wiki2.dovecot.org/Pigeonhole/ManageSieve) and let users configure their server-side rules themselves.
There are plugins available for e.g. Thunderbird: https://addons.mozilla.org/de/thunderbird/addon/sieve/

stonemaster:
... ENABLE_MANAGE_SIEVE has been set.

Adapted README and updated tests.

The functionality has successfully been tested using the Sieve
Thunderbird plugin.

Fixes #167

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ MAINTAINER Thomas VIAL
 RUN DEBIAN_FRONTEND=noninteractive apt-get update -q --fix-missing && \
 	apt-get -y upgrade && \
 	apt-get -y install --no-install-recommends \
-	postfix dovecot-core dovecot-imapd dovecot-pop3d dovecot-sieve gamin amavisd-new spamassassin razor pyzor \
+	postfix dovecot-core dovecot-imapd dovecot-pop3d dovecot-sieve dovecot-managesieved gamin amavisd-new spamassassin razor pyzor \
 	clamav clamav-daemon libnet-dns-perl libmail-spf-perl bzip2 file gzip p7zip unzip arj rsyslog \
     opendkim opendkim-tools opendmarc curl fail2ban ed iptables && \
 	curl -sk http://neuro.debian.net/lists/trusty.de-m.libre > /etc/apt/sources.list.d/neurodebian.sources.list && \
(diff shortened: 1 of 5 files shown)


tomav:
Thank you @stonemaster.

The var should be named `ENABLE_SIEVE` no? Why adding `MANAGE`?

Dockerfile
```diff
@@ -74,6 +74,6 @@ RUN curl -s https://letsencrypt.org/certs/lets-encrypt-x1-cross-signed.pem > /et
 ADD target/bin/generate-ssl-certificate target/bin/generate-dkim-config target/start-mailserver.sh /usr/local/bin/
 RUN chmod +x /usr/local/bin/*
 
-EXPOSE 25 587 143 993 110 995
+EXPOSE 25 587 143 993 110 995 4190
```
tomav: Port `4190` really need to be exposed?

stonemaster: Yes it needs to be exposed because the client wants to connect to the manage sieve daemon and edit the sieve files online.

stonemaster:
The daemon is called `managesieved` and the feature is called *ManagedSieve* by the Dovecot guys (http://wiki1.dovecot.org/ManageSieve). But you're right - the current on is misleading. Maybe `ENABLE_MANAGESIEVE` is better? `ENABLE_SIEVE` wouldn't exactly be right because sieve is enabled by default if an appropriate `.dovecot.sieve` file has been installed. The manage sieve extension just allows users to manage their sieve scripts theirselves.

Dockerfile
tomav: Thanks, I didn't know how it worked.

tomav:
`ENABLE_MANAGESIEVE ` will be fine.

stonemaster:
Just pushed the change. Once the pull request is accepted I'll add a section for ManageSieve to the Wiki.

tomav:
Use "Configure" as prefix.
Or perhaps it should be a a sub-section of Sieve page. (same page)

stonemaster:
I will just use the same Sieve page because it fits quite well :-)

tomav:
Merged.

PR #171, opened 2016-04-30
Fix permission issues on wrong dkim path

Zauberstuhl:
Hi, I used your v2 branch and wasn't able to send mails due
```
Apr 30 11:01:11 mail opendkim[521]: mail._domainkey.zauberstuhl.de: key data is not secure: /tmp can be read or written by other users
Apr 30 11:01:11 mail opendkim[521]: 176E2C07: error loading key 'mail._domainkey.zauberstuhl.de'
```

I followed your tutorial and ran `generate-dkim-config` from the console. Maybe I missed a step but in the KeyTable file there is the path to the temporary file configured after creation.

~~You can avoid this kind of issue by simply running `sed` after the copy step. Even if I missed something that fix could avoid problems in future.~~

great work btw.. Cheers!

### DIFF
--- a/target/bin/generate-dkim-config
+++ b/target/bin/generate-dkim-config
@@ -33,10 +33,10 @@ grep -vE '^(\s*$|#)' /tmp/vhost | while read domainname; do
 	fi
 
 	# Write to KeyTable if necessary
-	keytableentry="mail._domainkey.$domainname $domainname:mail:/tmp/docker-mailserver/opendkim/keys/$domainname/mail.private"
+	keytableentry="mail._domainkey.$domainname $domainname:mail:/etc/opendkim/keys/$domainname/mail.private"
 	if [ ! -f "/tmp/docker-mailserver/opendkim/KeyTable" ]; then
 		echo "Creating DKIM KeyTable"
-	 	echo "mail._domainkey.$domainname $domainname:mail:/tmp/docker-mailserver/opendkim/keys/$domainname/mail.private" > /tmp/docker-mailserver/opendkim/KeyTable
+		echo $keytableentry > /tmp/docker-mailserver/opendkim/KeyTable
 	else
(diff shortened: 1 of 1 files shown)


tomav:
The generator mounts a folder to `/tmp`to get it back on your host.
Once you have, this folder is re-mounted to `/tmp` and files are copied to `/etc` and permissions applied.
Have a look to https://github.com/tomav/docker-mailserver/blob/v2/target/start-mailserver.sh#L82-L95

Could you paste your `docker-compose.yml` and a `tree` ouput of your config directory.

You should not experience your issue. Perhaps we've missed something.

Thanks

Zauberstuhl:
> The generator mounts a folder to /tmp to get it back on your host.
Once you have, this folder is re-mounted to /tmp and files are copied to /etc and permissions applied.

yes and that is not the problem. My problem is the content of  `/etc/opendkim/KeyTable`:

`mail._domainkey.example.org example.org:mail:/tmp/docker-mailserver/opendkim/keys/example.org/mail.private`
should be
`mail._domainkey.example.org example.org:mail:/etc/opendkim/keys/example.org/mail.private`

otherwise it will always complain about permission issues since he tries to load the key from `tmp` and not from `etc`

diiigle:
I can confirm this issue. 
My syslog tells me:

```
May  1 13:06:35 vps opendkim[463]: can't load key from /tmp/docker-mailserver/opendkim/keys/example.com/mail.private: Permission denied
May  1 13:06:35 vps opendkim[463]: F186B201FBF: error loading key 'mail._domainkey.example.com'
```
Why don't we edit the `generate-dkim-config` script to write the right path to the KeyTable file?

Zauberstuhl:
> Why don't we edit the generate-dkim-config script to write the right path to the KeyTable file?

ups kinda missed that this is part of the docker-mailserver as well.
that is actually the better approach cheers!

tomav:
Thanks @diiigle, well spotted!

PR #172, opened 2016-05-01
opendkim folder shadowing

diiigle:
./opendkim/ local folder shadows ./config/opendkim
on generation of the keyfiles all files get written to ./config/opendkim leaving ./opendkim empty
thus on startup no config for opendkim is loaded


diiigle:
I also already edited the wiki

tomav:
Thanks.

PR #181, opened 2016-05-08
Merge "volumes" option in docker-compose.yml

Chris54721:
By having two separate `volumes` options, only the last one is used, so they should be merged into one to make sure the `maildata` container is used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,14 +50,13 @@ Adapt this file with your FQDN.
         hostname: mail
         domainname: domain.com
         container_name: mail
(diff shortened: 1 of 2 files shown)


tomav:
This should fix https://github.com/tomav/docker-mailserver/issues/173

Chris54721:
Can confirm, even after running `docker rm -f mail` everything is fine when starting the container again.

tomav:
Thank you @Chris54721

Issue #190, opened 2016-05-22
Don't always fix permissions on start-up


tve:
Fixing permissions on startup https://github.com/tomav/docker-mailserver/blob/master/target/start-mailserver.sh#L220 takes many minutes on my (perhaps a bit wimpy) mail server with many GB of messages stored in maildir format. Please either check whether the dirs already have the right perms or provide an option to skip the check. Thanks for an excellent docker image!

tomav:
Thanks @tve 
Would you submit a PR for that?

tve:
Implemented in #194

PR #194, opened 2016-05-23
Avoid fixing permissions; add regexp alias file, add AWS SES outgoing email support

tve:
This PR is not intended to be merged as-is, please indicate whether you want all/some/none of it and I can cherry-pick accordingly. It has:
- avoid fixing permissions if they're OK #190
- support regexp alias file
- support outgoing email via AWS SES

I added docs to the README, but I would move that to the wiki.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ MAINTAINER Thomas VIAL
 RUN DEBIAN_FRONTEND=noninteractive apt-get update -q --fix-missing && \
 	apt-get -y upgrade && \
 	apt-get -y install --no-install-recommends \
-	postfix dovecot-core dovecot-imapd dovecot-pop3d dovecot-sieve dovecot-managesieved gamin amavisd-new spamassassin razor pyzor \
+	postfix dovecot-core dovecot-imapd dovecot-pop3d dovecot-sieve dovecot-managesieved gamin amavisd-new spamassassin razor pyzor libsasl2-modules \
 	clamav clamav-daemon libnet-dns-perl libmail-spf-perl bzip2 file gzip p7zip unzip arj rsyslog \
     opendkim opendkim-tools opendmarc curl fail2ban ed iptables && \
 	curl -sk http://neuro.debian.net/lists/trusty.de-m.libre > /etc/apt/sources.list.d/neurodebian.sources.list && \
--- a/README.md
+++ b/README.md
@@ -1,32 +1,4 @@
-# docker-mailserver [![Build Status](https://travis-ci.org/tve/docker-mailserver.svg?branch=master)](https://travis-ci.org/tve/docker-mailserver)
-
-This is a fork of https://github.com/tomav/docker-mailserver with some additional features
-described below.
-
-### Sending outbound mail via Amazon SES
-
-Instead of letting postfix deliver mail directly it is possible to forward outgoing email
-through Amazon SES (Simple Email Service). To enable this feature, define the following two
-environment variables in the `docker-compose.yml` with the appropriate values for your AWS SES
-subscription (the values for `AWS_SES_USERPASS` are the "SMTP username" and "SMTP password"
-provided when yuo create SMTP credentials for SES):
-```
-    environment:
-    - AWS_SES_HOST=email-smtp.us-east-1.amazonaws.com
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -1,4 +1,32 @@
-# docker-mailserver [![Build Status](https://travis-ci.org/tomav/docker-mailserver.svg?branch=master)](https://travis-ci.org/tomav/docker-mailserver)
+# docker-mailserver [![Build Status](https://travis-ci.org/tve/docker-mailserver.svg?branch=master)](https://travis-ci.org/tve/docker-mailserver)
+
+This is a fork of https://github.com/tomav/docker-mailserver with some additional features
+described below.
+
+### Sending outbound mail via Amazon SES
```
tomav: Could you move this doc to the Wiki? Simpler to maintain.

README.md
```diff
@@ -1,4 +1,32 @@
-# docker-mailserver [![Build Status](https://travis-ci.org/tomav/docker-mailserver.svg?branch=master)](https://travis-ci.org/tomav/docker-mailserver)
+# docker-mailserver [![Build Status](https://travis-ci.org/tve/docker-mailserver.svg?branch=master)](https://travis-ci.org/tve/docker-mailserver)
```
tomav: This part should not be commit to this repo.

target/start-mailserver.sh
```diff
@@ -78,6 +78,15 @@ if [ -f /tmp/docker-mailserver/postfix-virtual.cf ]; then
 else
   echo "==> Warning: 'config/postfix-virtual.cf' is not provided. No mail alias/forward created."
 fi
+if [ -f /tmp/docker-mailserver/postfix-regexp.cf ]; then
+  # Copying regexp alias file
```
tomav: Could you add test on it (see `test/bats.tests`)

target/start-mailserver.sh
```diff
@@ -216,8 +246,13 @@ else
   echo "==> Warning: 'SASL_PASSWD' is not provided. /etc/postfix/sasl_passwd not created."
 fi
 
-echo "Fixing permissions"
-chown -R 5000:5000 /var/mail
+# Fix permissions, but skip this if 3 levels deep the user id is already set
```
tomav: There's a pending PR from @Josef-Friedrich https://github.com/tomav/docker-mailserver/pull/193
You should discuss the implementation to only keep one.
Thanks.

Dockerfile
```diff
@@ -5,7 +5,7 @@ MAINTAINER Thomas VIAL
 RUN DEBIAN_FRONTEND=noninteractive apt-get update -q --fix-missing && \
 	apt-get -y upgrade && \
 	apt-get -y install --no-install-recommends \
-	postfix dovecot-core dovecot-imapd dovecot-pop3d dovecot-sieve dovecot-managesieved gamin amavisd-new spamassassin razor pyzor \
+	postfix dovecot-core dovecot-imapd dovecot-pop3d dovecot-sieve dovecot-managesieved gamin amavisd-new spamassassin razor pyzor libsasl2-modules \
```
tomav: `libsasl2-modules` is really needed?

README.md
tomav: https://github.com/tomav/docker-mailserver/wiki

tomav:
Thanks for the work @tve 
Let's check these little things and I'll merge.

README.md
tve: That was the plan :-)

README.md
tve: Yup

Dockerfile
tve: Unfortunately, yes. The config you have uses dovecot for SASL, but dovecot doesn't support client auth (http://www.postfix.org/SASL_README.html#client_sasl says "At this time, the Dovecot SASL implementation does not provide client functionality") and what is needed for AWS SES is for postfix to do SASL as a client. The same would be needed for pretty much any scenario where outgoing email gets relayed through an authenticated connection, unless I'm missing something...

target/start-mailserver.sh
tve: Done

target/start-mailserver.sh
tve: Started...

tve:
This PR is ready from my point of view. Please let me know if there's anything else you would like changed.

tomav:
Thank you @tve, really nice work and wiki documentation.

Issue #191, opened 2016-05-22
Persist spool directories


tve:
As far as I can tell, restarting the container looses the postfix spool directories. IMHO these should be mapped to a volume in the sample docker-compose file so they persist.

tomav:
Here again, if you want to modify the wiki / `README.md`, feel free to do it.

I think we should do the same for spamassassin database.

tomav:
The folder for spamassassin is `/var/lib/amavis/`

Josef-Friedrich:
I tried to mount this folders outside:

* `/var/lib/spamassassin`
* `/var/lib/fail2ban`
* `/var/lib/amavis`
* `/var/lib/clamav`

tve:
I see two options here. One is to mount a slew of volumes onto all the directories with state. The other is to mount one or two volumes and move where the various daemons put their data onto the volume using symlinks or config changes.

Mounting a slew of volumes is cleaner is a way in that it exposes what's going on and avoids mysterious symlinks and locations. But it doesn't allow for a nice backup model by mounting "the" volume onto a backup container. Having one volume with everything or perhaps two volumes (one for user mailboxes and one for "spool" stuff) makes the backup story simpler.

Thoughts?

Josef-Friedrich:
I prefer the one (two) volume backup solution.

00angus:
That's a very good idea to allow persistence at least for emails spool.
The approach involving "a slew of volumes" is - IMHO - preferable as it makes explicit what will be shared with the host. After all once you have put all the relevant config in a docker-compose or a script it doesn't matter, I think.
Af for the convenience of the backup, if the slew of volumes come from the same base path ... it's the same :)

PR #195, opened 2016-05-24
Persist mail state directories

tve:
This PR consolidates all directories with mail state into /var/mail-state, which can then be mounted as a volume. This is to address #191.

This PR is based off the changes in PR 194, please look at the one new commit https://github.com/tomav/docker-mailserver/commit/4ca39f9144f42675257af879c7d49558e5d66f1d for the changes.

If you like this approach, I will add docs to the Wiki. Dunno whether you want this to be in the docker-compose.yaml.dist or not. Also, I changed the tests to use the consolidated directories, but I don't really know how you want to proceed, so this PR is a bit of a WIP 'til I hear back...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,11 +19,13 @@ run:
 	docker run -d --name mail \
 		-v "`pwd`/test/config":/tmp/docker-mailserver \
 		-v "`pwd`/test":/tmp/docker-mailserver-test \
+		-v "`pwd`/test/onedir":/var/mail-state \
 		-e SA_TAG=1.0 \
 		-e SA_TAG2=2.0 \
 		-e SA_KILL=3.0 \
 		-e SASL_PASSWD=testing \
 		-e ENABLE_MANAGESIEVE=1 \
+		-e ONE_DIR=1 \
 		-h mail.my-domain.com -t $(NAME)
 	sleep 20
 	docker run -d --name mail_pop3 \
--- a/target/start-mailserver.sh
+++ b/target/start-mailserver.sh
@@ -268,6 +268,29 @@ test -z "$ENABLE_FAIL2BAN" && rm -f /etc/logrotate.d/fail2ban
 # Fix cron.daily for spamassassin
 sed -i -e 's/invoke-rc.d spamassassin reload/\/etc\/init\.d\/spamassassin reload/g' /etc/cron.daily/spamassassin
 
+# Consolidate all state that should be persisted across container restarts into one mounted
+# directory
+statedir=/var/mail-state
+if [ "$ONE_DIR" = 1 -a -d $statedir ]; then
+  echo "Consolidating all state onto $statedir"
+  for d in /var/spool/postfix /var/lib/postfix /var/lib/amavis /var/lib/clamav /var/lib/spamassasin /var/lib/fail2ban; do
+    dest=$statedir/`echo $d | sed -e 's/.var.//; s/\//-/g'`
+    if [ -d $dest ]; then
+      echo "  Destination $dest exists, linking $d to it"
+      rm -rf $d
+      ln -s $dest $d
+    elif [ -d $d ]; then
+      echo "  Moving contents of $d to $dest:" `ls $d`
+      mv $d $dest
+      ln -s $dest $d
+    else
+      echo "  Linking $d to $dest"
+      mkdir -p $dest
+      ln -s $dest $d
+    fi
+  done
+fi
+
 echo "Starting daemons"
 cron
 /etc/init.d/rsyslog start


Josef-Friedrich:
For now I store my mails in a local directory. @tve Is it possible to mount a local directory to this persistent state directory too?

Josef-Friedrich:
I would be very happy if this PR could be merged.

tve:
@Josef-Friedrich if by "local directory" you mean a directory on the docker host, then yes, you can do that. My docker-compose-fu isn't well developed yet, but for sure it works if you use something like the following in your docker-compose.yaml:
```
services:
  mail:
    ...
    volumes:
    - /mnt/mail-state:/var/mail-state
```
Here /mnt/mail-state is a directory on the host. You can then back this dir up. I haven't figured out how to do it with the named volumes syntax.

tve:
Docs for the wiki (waiting to hear from @tomav whether he likes the direction of this PR):

# Configuring Backups

## User mail directories

Assuming that you use docker-compose and a data volume container named maildata, you can backup your user mails like this:

```
docker run --rm \
--volumes-from maildata_1 \
-v "$(pwd)":/backups \
-ti tvial/docker-mailserver \
tar cvzf /backups/docker-mailserver-`date +%y%m%d-%H%M%S`.tgz /var/mail
```

## Mail state directories

In addition to the user mail directories you may want to persist and even back up the state of the various daemons, such as postfix mail queues, clamav virus database, spamassassin learning, etc. Inside the container the files for this are normally found in various directories under `/var/lib` and `/var/spool` but by adding the `ONE_DIR=1` option to the environment these all get moved to `/var/mail-state` at container start-up. This allows you to mount a volume there, which has the first benefit that all this state is persisted if you need to restart the container. As a second benefit, you can also back this volume up to secondary storage, although it is not really required, i.e., no email will get lost unless something was just in transit in the postfix queues.

In docker-compose.yaml you can enable the merging of all daemon state into one directory and the mounting of a volume as follows:
```
services:
  mail:
    ...
    environment:
      ...
    - ONE_DIR=1
    volumes:
    ...
    - mailstate:/var/mail-state

volumes:
  ...
  mailstate:
    driver: local
```

tve:
I'm having trouble getting a build to complete. Travis times out just building the container images :-(

Josef-Friedrich:
The download of the main.cvd is very slow. I'm just building the image on my laptop ...

tve:
Yeah, I did set things up so I could build and test locally, and there the tests pass. It's just not nice to have a PR where "all checks have failed"...

tomav:
Just restarted the build.
https://travis-ci.org/tomav/docker-mailserver/builds/132603315

tomav:
A question: why don't we change where these daemons write their content to the single folder, using configuration files, instead of adding new code in `start-mailserver.sh`.

tve:
That's certainly a very valid alternative. I didn't want to dive into the documentation for the 7 different daemons to find what I had to change and then test that I didn't forget something, etc...

tomav:
I understand, but there's another easier way by adding the folder list to a data container.

```
version: '2'

services:
  mail:
    # build: .
    hostname: mail
    domainname: domain.com
    container_name: mail
    ports:
    - "25:25"
    - "143:143"
    - "587:587"
    - "993:993"
    volumes:
    - maildata:/var/mail
	- maildata:/var/spool/postfix
	- maildata:/var/lib/postfix
	- maildata:/var/lib/amavis
	- maildata:/var/lib/clamav
	- maildata:/var/lib/spamassasin
	- maildata:/var/lib/fail2ban

volumes:
  maildata:
    driver: local
```

tve:
Interesting, I didn't know one could do that. I guess no PR needed in that case? I'll keep using my fork because I want to map these directories to an EC2 EBS volume so I can do snapshot-backups and I can't do that easily with docker volumes, but that may not be of interest to many other folks.

Josef-Friedrich:
@tomav I tried your docker-compose.yml file from above. I got this error:

`mail    | May 29 19:45:31 mail postfix/master[642]: fatal: open lock file /var/lib/postfix/master.lock: cannot create file exclusively: Permission denied`

tomav:
There are some ownerships/permissions to set.
Let's gonna list them here.

    - maildata:/var/spool/postfix
    - maildata:/var/lib/postfix => postfix:postfix
    - maildata:/var/lib/amavis => amavis:amavis with default directories (tmp and db)
    - maildata:/var/lib/clamav => clamav:clamav
    - maildata:/var/lib/spamassasin
    - maildata:/var/lib/fail2ban

tomav:
No time for this right time.

I'm gonna merge your PR and will try to do something more docker compliant later.

tomav:
Thanks for the work anyway!

PR #196, opened 2016-05-24
Add a additional configuration file for dovecot

Josef-Friedrich:
Inspired by the `postfix-main.cf` I created a `dovecot.cf` configuration file which is loaded by dovecot as last config file. So additional configurations for dovecot can be easily appended.

### DIFF
--- a/config/dovecot.cf
+++ b/config/dovecot.cf
@@ -1,0 +1,4 @@
+# File for additional dovecot configurations.
+# For more informations read http://wiki.dovecot.org/BasicConfiguration
+
+#mail_max_userip_connections = 50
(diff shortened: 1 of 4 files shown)


tomav:
Thank you @Josef-Friedrich and sorry for the delay.
Can you add an integration test on it?
On one of the existing test container, load your file sample and check that this container has been updated with it.

We spent a lot of time on tests on this v2 and we'd like to keep it easily maintainable.

I'll merge as soon as the test is added.
Let me know if you need help.

Thanks.

Josef-Friedrich:
I added some tests. By the way: Very nice test framework

tomav:
Good job @Josef-Friedrich

Issue #200, opened 2016-06-01
Increase findtime and bantime for fail2ban?


dominikwinter:
Hi,

I noticed lots of authentication failes from different ip addresses on my server .. for example:
```
root@mail:/# grep "authentication failed" /var/log/syslog | grep 91.200.*.150 | cut -c1-74 | tail -n5
May 31 22:43:40 mail postfix/smtpd[10945]: warning: unknown[91.200.*.150]
May 31 22:58:57 mail postfix/smtpd[10982]: warning: unknown[91.200.*.150]
May 31 23:14:20 mail postfix/smtpd[11002]: warning: unknown[91.200.*.150]
May 31 23:29:46 mail postfix/smtpd[11012]: warning: unknown[91.200.*.150]
May 31 23:45:13 mail postfix/smtpd[11030]: warning: unknown[91.200.*.150]
```
Or:
```
root@mail:/# grep "authentication failed" /var/log/syslog /var/log/syslog.1 | cut -d ' ' -f 7 | grep ^un | cut -c9- | rev | cut -c3- | rev | sort | uniq -c | sort -nr
    103 91.200.*.150
     81 91.200.*.124
     79 96.57.*.106
     79 91.200.*.152
     79 91.200.*.148
     79 91.200.*.142
     75 91.200.*.139
     73 91.200.*.101
     52 91.200.*.147
```
After I set bantime = 21600, findtime = 21600 and maxretry = 2 in /etc/fail2ban/jail.conf and do ..

`service fail2ban reload`

.. everything is working fine again :)
```
root@mail:/# iptables -L
--<8
Chain f2b-dovecot (1 references)
target     prot opt source               destination
REJECT     all  --  ool-6039256a.static.********.net  anywhere             reject-with icmp-port-unreachable
REJECT     all  --  dedic774.******.net  anywhere             reject-with icmp-port-unreachable
REJECT     all  --  dedic774.******.net  anywhere             reject-with icmp-port-unreachable
RETURN     all  --  anywhere             anywhere
8>--
```
Maybe it's a good idea to make these variables configurable or at least increase them by default a bit .. what do you think?

Greets

tomav:
Yep, good idea. Can you do that?

Variable can be named `FAIL2BAN_BANTIME` and should have a default value.
You can check what has been done for Spamassasin.

PR #204, opened 2016-06-04
added ability to overwrite jail.conf

dominikwinter:
#200 

Now it's possible to change configurations for fail2ban in config/ directory, too.


tomav:
Thank you @dominikwinter! Perfect!
Could you also document it in [this page](https://github.com/tomav/docker-mailserver/wiki/Configure-Fail2ban)?

Issue #207, opened 2016-06-10
A way to override master.cf?


lauri865:
Not really an issue with the repo, but probably my lack of knowledge.

Is there a way to override master.cf? I want to add a hook to trigger a php script, but couldn't find a way to override or append to master.cf. Any hints would be much appreciated.

tomav:
There's currently no such feature. But could be added, we do the same thing for some other file.

lauri865:
Pushed a commit related to this. Please have a look if you have time. Thanks!

lauri865:
Took a different stab at it. Should add more flexibility by enabling free execution of postconf

tomav:
Send the PR, I'll be able to see your 2 commit merged and review it.
You should also add integration tests.
Thanks.

tomav:
Hi @lauri865, any update? Thank you.

tomav:
Closed due to inactivity. Reopen if needed.

PR #208, opened 2016-06-10
Add ability to override master.cf (#207)

lauri865:
Should be sufficient I guess?


lauri865:
Will figure out a way to add `postconf -P` support as well

PR #210, opened 2016-06-14
switch to SHA512

Crafter6432:
I removed test/auth/smtp-auth-cram-md5.txt because I don't think that CRAM-MD5-AUTH will work anymore, if we don't have a MD5 hash to compare against.


Crafter6432:
Any idea how to fix this test ?

tomav:
Yes, the test is `diff test/config/example.postfix-accounts.cf test/config/postfix-accounts.cf`
So you have to update `test/config/example.postfix-accounts.cf` as reference with `SHA512` passwords.
Default clear password should be `mypassword`.

Crafter6432:
SHA512-CRYPT is salting the passwords, which means the hash is always different and thus not comparable.

tomav:
Thank you. I update the Wiki.

PR #211, opened 2016-06-14
Usermanagement

Crafter6432:
This would make the process of adding a users a bit more convenient, but requires the config volume to be mounted. 

Adds the following commands:
```
 addmailuser <user@domain> [password]
 delmailuser <user@domain>
```


tomav:
@Crafter6432 Could you add integration tests to be sure that everything works as expected?

Crafter6432:
Seems like the imap STARTTLS is failing from time to time, cloud you rerun the tests ?

Crafter6432:
The said test even fails on master, I guess the dovecot package was updated in ubuntu's repo ?

Tests on master:
https://gist.github.com/Crafter6432/a42c3b5b4cf88a95abd67aed145e60ec

tomav:
Just restarted the build. Sometimes this on fails, probably when container is not 100% started.

tomav:
Better now.

tomav:
Merged.

@Crafter6432 could add some notes in https://github.com/tomav/docker-mailserver/wiki/Configure-Accounts ?

Thank you.

PR #224, opened 2016-07-05
added certs installation and information

kopax:
- added certs installation and information			92d2742
- added key in case of custom certs		6e6ef28

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,30 @@ Otherwise, `iptables` won't be able to ban IPs.
   - custom => Enables custom certificates
   - self-signed => Enables self-signed certificates
 
+__custom__
+
+Mount your certs : 
+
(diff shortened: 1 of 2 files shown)


tomav:
Thank you @kopax 
Can you remove documentation in the `README`?
We'd like to keep it as simple as possible.
Detailed documentation is in the Wiki, we have a [SSL dedicated page](https://github.com/tomav/docker-mailserver/wiki/Configure-SSL).

tomav:
Hi @kopax, I'm back from vacations.
Let me know in this PR, I'll merge once ready.

kopax:
Hi @tomav, sorry I didn't follow this one but I saw you had a dedicated page.

By the way, I have forked your project on my gitlab your project and added LDAP feature.
I tried to kept both (with/without LDAP) to work, but since I wasn't able to make it work (V2) with my signed certs, I did some rewrite of the way you import the certs so it can manage correctly the signed certs (do not use one file if we can use two, being able to import ca cert...)

If you feel like you wan't to test my version, I would be happy to put it on github and create the PR as long as someone take care of the backward compatibility test.

PR #229, opened 2016-07-12
Fix some bugs in the shell scripts.

vifino:
Hello.

This PR contains some fixes: 
* Due to the lack of quotes in `addmailuser`'s usage, it didn't actually display the usage.
* generate-ssl-certificate did not use the fqdn, but only the short hostname.
* Permission errors in scripts.


tomav:
Hi @vifino, thanks for that!

PR #243, opened 2016-07-25
Add paths to CA to postifx

auchri:
Fixes untrusted TLS connections

See: http://giantdorks.org/alain/fix-for-postfix-untrusted-certificate-tls-error/

Before: `Untrusted TLS connection established to gmail...`

After: `Trusted TLS connection established to gmail...`


tomav:
Thank you @auchri!

PR #246, opened 2016-07-28
fixed an issue where sieve complains about ~/.dovecot.sieve is not a directory

methanol:
I am having an issue using ManageSieve where I got the following error message:

```
Jul 28 00:02:27 mail dovecot: imap(xxx@xxx.xxx): Error: stat(/var/mail/xxx.xxx/xxx/.dovecot.sieve/tmp) failed: Not a directory
Jul 28 00:02:35 mail dovecot: message repeated 247 times: [ imap(xxx@xxx.xxx): Error: stat(/var/mail/xxx.xxx/xxx/.dovecot.sieve/tmp) failed: Not a directory]
```

`.dovecot.sieve` seems to be a symbolic link.

I am using RoundCube to configure sieve filter rules.


methanol:
Nevermind... I just checked again and it works as expected... dont know why I got that error

PR #247, opened 2016-07-28
Startup logging

KyleOndy:
Adding two simple echo statements verifying when `ENABLE_FAIL2BAN` or `ENABLE_MANAGESIEVE` are set.


tomav:
Thanks @KyleOndy

Issue #216, opened 2016-06-22
Option to disable AV


Crafter6432:
It would be really nice to have an option to disable spamassassin, amavis and clamav as these services are very resource hungry(especially memory). In my case for instance these 3 services consume 23%(~450mb) which in my opinion isn't worth it as I'm the only user and know how to deal with viruses and spam if there's any.

tomav:
Good idea @Crafter6432!
If you want to PR this enhancement, I'll be happy to merge it into master.
Be sure to read [this page](https://github.com/tomav/docker-mailserver/wiki/How-to-contribute) before.

ep0ch2:
Don't suppose there's an image with this feature that i could test?

KyleOndy:
@tomav Could we just wrap the configuration and starting of each component in a check for an Env var? Similar to the existing `ENABLE_MANAGESIEVE` and such?

I will submit a PR if that sounds reasonable.

tomav:
@KyleOndy of course. But it could have an impact on some config files, like postfix.

tomav:
Done

PR #250, opened 2016-08-04
selective service disable

KyleOndy:
This pull request is in response to #216.

This PR only selectively disabled either clamav, spamassassin, or amavis.

I am not sure if the config files needs to be touched at all.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,6 +54,12 @@ run:
 		-v "`pwd`/test":/tmp/docker-mailserver-test \
 		-e DISABLE_AMAVIS=1 \
 		-h mail.my-domain.com -t $(NAME)
+	sleep 20
+	docker run -d --name mail_disabled_spamassassin \
+		-v "`pwd`/test/config":/tmp/docker-mailserver \
+		-v "`pwd`/test":/tmp/docker-mailserver-test \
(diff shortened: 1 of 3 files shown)


KyleOndy:
Build timed out. It built on my [branch of travis](https://travis-ci.org/KyleOndy/docker-mailserver/builds/149836649).

tomav:
Restarted.

tomav:
Thank you!

ep0ch2:
Many thanks for this, the following settings work:
    - DISABLE_AMAVIS=0
    - DISABLE_SPAMASSASSIN=1
    - DISABLE_CLAMAV=1

however with disable_amavis=1 i can't send/receive emails. so perhaps some extra config changes are needed,

KyleOndy:
Looking into it now.

tomav:
The answer is probably [here](https://github.com/tomav/docker-mailserver/blob/master/target/postfix/main.cf#L71) and perhaps [here](https://github.com/tomav/docker-mailserver/blob/master/target/postfix/master.cf#L84)

tomav:
Perhaps, to simplify, a `DISABLE_FILTERS` should be easier to implement (no matrix with the 3 current vars, but of course, they bring more granularity).

Let me know.

PR #251, opened 2016-08-06
added LDAP to dovecot and postfix (TODO: regression test & improve DOC)

kopax:
My mail-server is running and working, this has been done 1 month ago:

In my memories, `start-mailserver.sh` wasn't ready for signed certs. I do use signed certs, and I prefer to have a separate file for both, the key and the certs.
Signed-certs documentation needs to be adapted for the new changes, other certs conf should not be impacted by the change.

I have added postfix-ldap and dovecot-ldap to the packet

I have updated the dovecot.conf for working with my ldap, however, this cannot be kept for simple authentication, please use two separate dovecot.conf file and choose the proper one on initialization by adding some more code to `start-mailserver.sh`.

I also added some examples on how to add accounts and aliases to openldap.

Most important TODO: do regression test and clean the DOC.

If you need help to start the server with your ldap, I can help here.

Regards and sorry for being in hurry, I was OK to share with the help of @tomav, but vacation period and work decided differently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,8 @@ Don't forget to adapt MAIL_USER and MAIL_PASS to your needs
 
 Now the keys are generated, you can configure your DNS server by just pasting the content of `config/opedkim/keys/domain.tld/mail.txt` in your `domain.tld.hosts` zone.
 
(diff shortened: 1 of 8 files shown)


tomav:
#335 and #352 will replace this one.

PR #264, opened 2016-08-17
Minor corrections to README.md

mathuin:
One example was missing a useful line, which was present in the wiki's example and without which the example does not run.

A typographical error was found in a file.


tomav:
Thanks.

PR #265, opened 2016-08-17
Handle missing files more gracefully.

mathuin:
Several tools expect the presence of certain config files, and fail gracelessly when they are absent.

Both addmailuser and delmailuser assume the existence of the postfix-accounts.cf file, while generate-dkim-config assumes the existence of that file and the postfix-virtual.cf file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@
 docker-compose.yml
 .idea
 test/config/empty/
+test/config/without-accounts/
+test/config/without-virtual/
 test/config/postfix-accounts.cf
 test/config/letsencrypt/mail.my-domain.com/combined.pem
 test/onedir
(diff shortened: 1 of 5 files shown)


tomav:
Build timed out. Restarted.

tomav:
Build is green. But could you add some functional test on this part?

mathuin:
I have never written functional tests before.  They're fun!  If you could make sure that I did them right, I'd appreciate it.

mathuin:
Huh, deleting those files in the test worked here.  Any idea why they cannot be deleted in Travis?

mathuin:
I fixed the functional tests, then rebased against master, which appears to have messed things up.  Would you prefer that I make a clean PR with only my code?

mathuin:
I'm going to force-push my fixed branch here, and see what happens.

tomav:
Thanks @mathuin

PR #273, opened 2016-08-22
Delete orphaned configuration file

Josef-Friedrich:
The config file target/dovecot/dovecot.conf is not used to build the
docker-mailserver image. All tests are working without this file. The
additional config file config/dovecot.cf won’t be loaded if
target/dovecot/dovecot.conf is used.


tomav:
https://github.com/tomav/docker-mailserver/issues/212

PR #278, opened 2016-08-24
add AWS_SES_PORT env

bigpigeon:
ses often connection timeout that used 25 port, therefore I  add a choosesable env

### DIFF
--- a/target/start-mailserver.sh
+++ b/target/start-mailserver.sh
@@ -249,8 +249,11 @@ fi
 
 # Support outgoing email relay via Amazon SES
 if [ ! -z "$AWS_SES_HOST" -a ! -z "$AWS_SES_USERPASS" ]; then
-  echo "Setting up outgoing email via AWS SES host $AWS_SES_HOST"
-  echo "[$AWS_SES_HOST]:25 $AWS_SES_USERPASS" >>/etc/postfix/sasl_passwd
+  if [ -z "$AWS_SES_PORT" ];then
+    AWS_SES_PORT=25
+  fi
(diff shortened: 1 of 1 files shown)


target/start-mailserver.sh
```diff
@@ -249,8 +249,11 @@ fi
...
 if [ ! -z "$AWS_SES_HOST" -a ! -z "$AWS_SES_USERPASS" ]; then
-  echo "Setting up outgoing email via AWS SES host $AWS_SES_HOST"
-  echo "[$AWS_SES_HOST]:25 $AWS_SES_USERPASS" >>/etc/postfix/sasl_passwd
+  if [ -z "$AWS_SES_PORT" ];then
+    AWS_SES_PORT=25
+  fi
+  echo "Setting up outgoing email via AWS SES host $AWS_SES_HOST:$AWS_SES_PORT"
+  echo "[$AWS_SES_HOST]:$AWS_SES_PORT $AWS_SES_USERPASS" >>/etc/postfix/sasl_passwd
   postconf -e \
     "relayhost = [$AWS_SES_HOST]:25" \
```
tomav: We should get AWS_SES_PORT on line 256 too, no?

bigpigeon: you also replace 587 with 25th port
use 25th port is bad idea!

bigpigeon: http://docs.aws.amazon.com/ses/latest/DeveloperGuide/smtp-connect.html

tomav:
Thanks!

PR #286, opened 2016-08-28
Fix build

Josef-Friedrich:
The Refactoring of the scripts 'addmailuser' and 'delmailuser'
destroyed the build process.


tomav:
Thanks!

PR #288, opened 2016-08-29
Minor tweaks to letsencrypt configuration.

mathuin:
The letsencrypt-auto client creates cert.pem, chain.pem, fullchain.pem, and
privkey.pem in its default settings.  The simp_le client creates cert.pem,
fullchain.pem, and key.pem in its default settings.

A check for either privkey.pem or key.pem was added.

The chain.pem file was only used by the letsencrypt code for the creation of
combined.pem, which is not used by either postfix or dovecot.

The code to create a combined.pem file for letsencrypt was removed, as was the
corresponding test.

### DIFF
--- a/target/start-mailserver.sh
+++ b/target/start-mailserver.sh
@@ -121,23 +121,27 @@ case $SSL_TYPE in
   "letsencrypt" )
     # letsencrypt folders and files mounted in /etc/letsencrypt
     if [ -e "/etc/letsencrypt/live/$(hostname)/cert.pem" ] \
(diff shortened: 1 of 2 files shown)


tomav:
Thanks.

mathuin:
I realized this morning that I had accidentally forgotten to commit an important change.  I have since pushed that change to my branch that was merged.  Should I make another pull request?

tomav:
Yep please.

mathuin:
PR submitted, after testing in my own personal production site.  Sorry about that!

PR #289, opened 2016-08-29
Improve fetchmail support

Josef-Friedrich:
Fetchmail is now configured by a script called 'setup-fetchmail'.
The script 'debug-fetchmail' is used inside the 'setup.sh' script.


tomav:
Thanks. Perhaps the Wiki needs a page on that topic?

Josef-Friedrich:
I’ve updated the corresponding [Wiki-Page](https://github.com/tomav/docker-mailserver/wiki/Retrieve-emails-from-a-remote-mail-server-(using-builtin-fetchmail)) and created an [new Wiki-Page about setup.sh](https://github.com/tomav/docker-mailserver/wiki/Setup-docker-mailserver-using-the-script-setup.sh)

PR #291, opened 2016-08-29
Let Travis CI test on all branches

Josef-Friedrich:
Most developers implement new docker-mailserver features on feature
branches before they create a pull request. This commit enables Travis
CI testing on all branches except 'donttestme'. Developers now can
'pre-test' their features branches without permanentaly hacking
'.travis.yml'. More testing is good!


tomav:
Great!

PR #296, opened 2016-08-31
Add the option of manually specifying paths to SSL certificates

keslerm:
I needed to directly specify the paths to my SSL certificates to solve getting this running within a Kubernetes cluster where my SSL certificates are managed by kube-lego.

Kubernetes will let me mount the secret volume (which is where the ssl keys are stored) into the container, but the "self signed" and "custom" options really work with that.

This will let you define two additional env variables

SSL_CERT_PATH
SSL_KEY_PATH

Honestly just being able to manually tell what SSL certificates is very handy since obviously not everyone fits in a specific structure. 

Let me know what you think or if there is anything I can improve!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,6 +75,13 @@ run:
 		-v "`pwd`/test":/tmp/docker-mailserver-test \
 		-e DISABLE_CLAMAV=1 \
 		-h mail.my-domain.com -t $(NAME)
+	docker run -d --name mail_manual_ssl \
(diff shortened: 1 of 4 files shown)


tomav:
Hi @keslerm, could you add a functional test on this part?

Thank you.

keslerm:
Sure can, wanted to get some feedback before I tried to figure out the tests.

tomav:
Thanks.

Issue #298, opened 2016-09-01
Allow Amavis configuration to be overriden


keslerm:
While trying to get this running on a kubernetes cluster I ran into a problem where kubernetes will not allow you to define a hostname for a pod/container. This obviously breaks a few things.

1. The postfix configuration is fairly easy to fix, you can just define the hostname info in the override file.
2. Dovecot does not generally need a hostname but I found it broke without adding in a setting `postmaster_address=email@address.com` to the dovecot.conf - again easy to mend with an override file but seems weird that I would need to.
3. The problem child is Amavis, it tries to use `hostname` to get the hostname and obviously that will fail because it will not be a FQDN. You can easily fix this with a setting but there's no amavis override file.

1 is easy I don't think it needs a change, 2 is weird but i don't think it needs a change but maybe just a note in the wiki

3 will require a change which will allow the Amavis config files to be overridden. This should be pretty easy because Amavis provides a place to store user configuration (`/etc/amavis/conf.d/50-user`).

The only other thing I think would be maybe a wiki entry on running in Kubernetes with the config.

keslerm:
One caveat is that Amavis config files are perl files so it is structured a specific way. I'm partial to requiring a user to provide the entire config file to copy rather than do any injecting into an existing file.

I figure if someone is needing to expand the Amavis config file they are probably aware of what it needs to look like.

tyranron:
Please, look at parameter `subdomain` at [v1.PodSpec](http://kubernetes.io/docs/api-reference/v1/definitions/#_v1_podspec).
For me it resolved the relevant issue.

Example:
```yaml
kind: Deployment
apiVersion: extensions/v1beta1
metadata:
  name: mail-server
  namespace: kube-system
  labels:
    k8s-app: mail-server
    kubernetes.io/cluster-service: 'true'
    version: 'v2'
spec:
  replicas: 1
  revisionHistoryLimit: 2
  selector:
    matchLabels:
      k8s-app: mail-server
      version: 'v2'
  template:
    metadata:
      labels:
        k8s-app: mail-server
        kubernetes.io/cluster-service: 'true'
        version: 'v2'
    spec:
      subdomain: mail-server  # <-- here it is
      containers:
        - name: docker-mailserver
          image: tvial/docker-mailserver:v2
          imagePullPolicy: Always
          ports:
            - name: smtp
              containerPort: 25
            - name: smtp-msa
              containerPort: 587
            - name: imap
              containerPort: 143
[...]

PR #299, opened 2016-09-01
Allow user to provide Amavis configuration

keslerm:
This will allow the user to provide custom Amavis configuration file.

Pretty straightforward.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,6 +75,13 @@ run:
 		-v "`pwd`/test":/tmp/docker-mailserver-test \
 		-e DISABLE_CLAMAV=1 \
 		-h mail.my-domain.com -t $(NAME)
+	docker run -d --name mail_manual_ssl \
+		-v "`pwd`/test/config":/tmp/docker-mailserver \
+		-v "`pwd`/test":/tmp/docker-mailserver-test \
+		-e SSL_TYPE=manual \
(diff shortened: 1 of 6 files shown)


keslerm:
Related to #298

tomav:
Hi @keslerm, could you resolve the conflict? I'll merge as soon its ready.
Thank you.

keslerm:
Updated

tomav:
Thanks, perhaps this needs to be explained in a Wiki page.

kachkaev:
Hi @keslerm, could you please add a few lines to wiki about how to configure amavis? I guess we can now mount some new volume for this to happen, but it's not 100% clear what to do to the newbies like me.

Being more specific, I'm interested in making the container as lite as possible to use the server as a 'reverse-proxy' for emails (i.e. people 'talk' to abcd@example.com and I manage everything with a gmail account). My gut-feeling suggests that I can somehow switch off all spam & virus checking and leave only dkim signatures by means of some custom amavis config (switching dkim off too is probably a bad idea). Feel free to correct me if my assumptions are wrong.

My mail 'reverse-proxy' works already, I just see that some incoming mail is being marked as spam and is therefore not passed upstream to google. What I'd like is to have gmail handle all that. Some users of this great docker image are likely to have the same intention, so your contribution to wiki will be appreciated by many!

Issue #306, opened 2016-09-03
PERMIT_DOCKER & Rancher


CrystalMethod:
I'm using the ``docker-mailserver`` in an Rancher [1] environment. Since some days I'm unable to use SMTP anymore. IMAP still works pretty well. In the logs I can find something like:

```
Adding container ip in my networks
postconf: fatal: -e, -X, or -# accepts no multi-line input
bash: line 1: <official ip>/32: No such file or directory
bash: line 1: <official ip>/32: No such file or directory
172.17.0.3
...
```

Since I'm on vacation I was not able to dig into details.

[1] http://rancher.com/rancher/

Josef-Friedrich:
@Zehir can you please look at this problem?

Zehir:
Hello, I dont Know Rancher.  Can you send me the result of this 2 commands from inside the container?
ip addr show eth0
Ifconfig
Thank

CrystalMethod:
Here comes the result:

```
root@mail:/# ip addr show eth0
67: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP grou
p default
    link/ether 02:54:63:e3:29:b2 brd ff:ff:ff:ff:ff:ff
    inet 172.17.0.3/16 scope global eth0
       valid_lft forever preferred_lft forever
    inet 10.42.15.210/16 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::54:63ff:fee3:29b2/64 scope link
       valid_lft forever preferred_lft forever
```
```
root@mail:/# ifconfig
eth0      Link encap:Ethernet  HWaddr 02:54:63:e3:29:b2
          inet addr:172.17.0.3  Bcast:0.0.0.0  Mask:255.255.0.0
          inet6 addr: fe80::54:63ff:fee3:29b2/64 Scope:Link
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:211590 errors:0 dropped:0 overruns:0 frame:0
          TX packets:278808 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:0
          RX bytes:29632186 (29.6 MB)  TX bytes:399099187 (399.0 MB)

lo        Link encap:Local Loopback
          inet addr:127.0.0.1  Mask:255.0.0.0
          inet6 addr: ::1/128 Scope:Host
[...]

PR #310, opened 2016-09-04
Fix for multiple IPV4

Zehir:
Quick fix for #306 
For now, supplementary ipv4 addresses are not added to mynetwork.

### DIFF
--- a/target/start-mailserver.sh
+++ b/target/start-mailserver.sh
@@ -231,7 +231,7 @@ touch /etc/postfix/vmailbox && postmap /etc/postfix/vmailbox
 touch /etc/postfix/virtual && postmap /etc/postfix/virtual
 
 # PERMIT_DOCKER Option
-container_ip=$(ip addr show eth0 | grep 'inet ' | sed 's/[^0-9\.\/]*//g' | cut -d '/' -f 1)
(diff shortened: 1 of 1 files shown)


Josef-Friedrich:
Many thanks for your quick response

mathuin:
Wow, this fix is horrible.

https://tools.ietf.org/html/rfc1918 is the RFC which refers to private addressing space.

Here's the bit which identifies those ranges:

```
3. Private Address Space

   The Internet Assigned Numbers Authority (IANA) has reserved the
   following three blocks of the IP address space for private internets:

     10.0.0.0        -   10.255.255.255  (10/8 prefix)
     172.16.0.0      -   172.31.255.255  (172.16/12 prefix)
     192.168.0.0     -   192.168.255.255 (192.168/16 prefix)
```

Please note that 172.0.0.0-172.15.255.255 and 172.32.0.0-172.255.255.255 are valid external IP addresses administered by ARIN, the registry responsible for distributing addresses throughout much of North America. With that PR in place, systems at those perfectly legitimate IP addresses are unable to use this container.

**Please** revert this change.  It doesn't actually help, and it can make things worse.

https://github.com/tomav/docker-mailserver/issues/306#issuecomment-246032470 shows that the original reporter's issue was caused by their ISP, and not the presence of two addresses.  If multiple addresses really do cause a problem, whatever fix is written should keep the above-quoted RFC in mind.

Josef-Friedrich:
@mathuin @Zehir Any ideas for a better fix?

Issue #311, opened 2016-09-05
[Question] Is spamd required to be running?


tyranron:
During my investigation of how this image works and reading a lot of tutorials about Postfix + AMaViS + SpamAssassin + ClamAV integrations I was wondered why many tutorials have no words about `spamd` or `service spamassassin start`, but AMaViS requires SpamAssassin to be installed.
That's why I have googled more about how AMaViS uses SpamAssassin and finally found [this archived mail](http://mail-archives.apache.org/mod_mbox/spamassassin-users/201001.mbox/%3C4B43FA52.40401@verizon.net%3E) which makes things clear.

According to that explanation `amavisd` represents some kind of `spamd` by itself and uses SpamAssasin via Perl API directly. So, actually, no `spamd` running is required if we use `amavisd`.
I've tested this stuff making `amavisd` running in separate container which does not run `spamd` process and it has worked OK as expected. The only thing required still to be run periodically is `sa-update` cron job.

So, the question is:
Do I miss some caveats which complicate things? Or, if not, can we improve image by disabling `spamd` as unnecessary process and leave only `sa-update` cron jobs?

tomav:
Hi @tyranron, I think you're right.
`spamd` could be disabled.

You can submit a PR for that, we'll be happy to merge it.

tomav:
Thank you @tyranron

PR #312, opened 2016-09-05
Remove unnecessary spamd process

tyranron:
This PR removes `spamd` process from container because of reasons described in #311.


tomav:
LGTM.

PR #313, opened 2016-09-05
Ignore commented or empty lines from config file

Zehir:
Affected files:

* `config/postfix-virtual.cf`
* `config/postfix-master.cf`

### DIFF
--- a/target/start-mailserver.sh
+++ b/target/start-mailserver.sh
@@ -69,6 +69,7 @@ if [ -f /tmp/docker-mailserver/postfix-virtual.cf ]; then
   cp /tmp/docker-mailserver/postfix-virtual.cf /etc/postfix/virtual
   while read from to
   do
+    ([[ "$from" =~ ^#.*$ ]] || [ -z "$from" ] ) && continue
     # Setting variables for better readability
     uname=$(echo ${from} | cut -d @ -f1)
     domain=$(echo ${from} | cut -d @ -f2)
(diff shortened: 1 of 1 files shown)


tomav:
Could you add a test on it?
Few users had this problem. Let's be sure that it will work, ever ;-)

Zehir:
It's will not work with this type of comments:
`mynetworks=127.0.0.0/8 # comment`
Then I dont know what type of test we can make for this

Josef-Friedrich:
@Zehir If I understand your PR right, all lines containing a comment are ignored? Why not simply clean out all comments? So the `docker-mailserver` admins can write as many comments in`postfix-*` they want. Here a sample code snippet:

```sh
#! /bin/sh

while read line; do
	line=$(echo $line | sed -E 's/#.*//')
	[ -n "$line" ] && echo $line
done <<EOF
postmaster@example.com postmaster # comment
info@example.com       joe # comment
sales@example.com      jane # comment
# Uncomment entry below to implement a catch-all address
# @example.com         jim
EOF
```

Output:

```
postmaster@example.com postmaster
info@example.com joe
sales@example.com jane
```

Zehir:
Yes you can use this too. Add your code in a pull request and reject this one.

tomav:
@Josef-Friedrich What should we do with this PR?

Josef-Friedrich:
I tried a different approach there: https://github.com/Josef-Friedrich/docker-mailserver/commit/c5ecd73e1e5828d483a36433155f796542513e47
I ran into testing issues. Any suggestions? Maybe I can have a look at this on the weekend.

tomav:
Hi @Zehir / @Josef-Friedrich, any update this?

tomav:
Will use https://github.com/Josef-Friedrich/docker-mailserver/commit/c5ecd73e1e5828d483a36433155f796542513e47 when ready. Feel free to ask if you need help.

PR #317, opened 2016-09-09
Avoid many escaped slashes in sed scripts

Josef-Friedrich:
Many escaped slashes can be avoided by changing the deliminter in sed
replace statements. This increases readability a little bit.


tomav:
Thanks

PR #361, opened 2016-10-24
Improve OpenDKIM and OpenDMARC milters integration

tyranron:
This PR provides some improvements for current milters integration inside container:

1. Upgrades `milter_protocol` from `2` to `6`. It seems that there is no reason to keep it `2`, because all apps that uses it (OpenDKIM, OpenDMARC) already supports `6` with their current versions.
2. Disables OpenDMARC milter for outgoing mails as far as it does no job. [DMARC relies on SPF and DKIM](https://dmarc.org/wiki/FAQ#How_does_DMARC_work.2C_briefly.2C_and_in_non-technical_terms.3F) and there is nothing to do with outgoing messages for it (opposite to OpenDKIM which must sign messages).
3. Disables duplicate milters checking when we receive mail back from AMaViS on `10025` Postfix port. Miltering before sending mail to AMaViS is enough.
4. Sets OpenDKIM and OpenDMARC to listen on more "traditional" ports `8891` and `8893`. See blockquote description in [this tutorial](http://www.stevejenkins.com/blog/2015/03/installing-opendmarc-rpm-via-yum-with-postfix-or-sendmail-for-rhel-centos-fedora/) to find out why these numbers are meant to be more "traditional".
5. Configures OpenDMARC parameters `AuthservID` and `TrustedAuthservIDs` with `HOSTNAME` value which, accordingly to [documentation](http://www.trusteddomain.org/opendmarc/opendmarc.conf.5.html), will automatically resolve to container hostname. So no need to inject those values in `start-mailserver.sh` script.
6. Relatively to previous one, moves creating of `/etc/opendmarc/ignore.hosts` file to `Dockerfile` instead of `start-mailserver.sh`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -83,6 +83,7 @@ COPY target/opendkim/default-opendkim /etc/default/opendkim
 # Configure DMARC (opendmarc)
 COPY target/opendmarc/opendmarc.conf /etc/opendmarc.conf
 COPY target/opendmarc/default-opendmarc /etc/default/opendmarc
+COPY target/opendmarc/ignore.hosts /etc/opendmarc/ignore.hosts
 
 # Configure fetchmail
 COPY target/fetchmail/fetchmailrc /etc/fetchmailrc_general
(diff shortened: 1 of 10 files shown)


tomav:
Great work @tyranron and explanation, thanks for that.

Issue #372, opened 2016-11-01
Changing the fetchmail config file has no effect.


jcgruenhage:
Only the first launch of the container with fetchmail enabled changes the config file, changing it later in is impossible without manually deleting files in the container and relaunching it.

The following part should be removed from setup-fetchmail to resolve this:

```bash
if [ -f "$RC" ]; then
	echo "The Fetchmail configuration is file '$RC' is already generated."
	exit 0
fi
```

tomav:
Merged.

PR #373, opened 2016-11-02
Changing the fetchmail config file has no effect (#372)

Josef-Friedrich:


tomav:
Thanks @Josef-Friedrich

PR #381, opened 2016-11-13
Use hostname to get domainname

dennis95stumm:
The domainname comand may return `(none)` use instead `hostname -d` to get the domainname


tomav:
Hi @dennis95stumm, should we also change the check here?
https://github.com/tomav/docker-mailserver/blob/master/target/start-mailserver.sh#L12

dennis95stumm:
Only if we want to check the domain and not the fully qualified domain name.

dennis95stumm:
The hostname

dennis95stumm:
So my opinion is: no.

tomav:
Thank you.

PR #382, opened 2016-11-15
Use "texthash" Postfix lookup format instead of "hash"

tyranron:
As far as Postfix setup inside Docker container does not require to re-apply configuration on-fly, and reconfiguration is done with container restarts (which actually stops and starts Postfix) we can use [`texthash:` Postfix lookup format](http://www.postfix.org/DATABASE_README.html#types) instead of `hash:`, which allows us don't mess with pre-`postmap` commands.

Also, I changed a bit handling of such lookup files to make them survive container restarts a bit better. Initially, if we create a container with accounts configuration, then remove all accounts and restart container, those accounts will be in lookup files still. I made them stripped before accounts configuration is applied.

### DIFF
--- a/target/postfix/main.cf
+++ b/target/postfix/main.cf
@@ -7,8 +7,8 @@ readme_directory = no
 
 # Basic configuration
 # myhostname = 
-alias_maps = hash:/etc/aliases
-alias_database = hash:/etc/aliases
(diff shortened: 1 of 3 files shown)


tomav:
Thank you @tyranron

Issue #357, opened 2016-10-14
Update Password


ckiszka:
Thank you for this docker! The wiki illustrates creating an account and password. Is there a command to update (change) a password? I'm nervous to attempt and lose access to mail without guidance. Any help is greatly appreciated. Thanks in advance. -- Chris

Nesousx:
Hi,

In order to update your password, you can run the following command ::

`docker run --rm -e MAIL_PASS=Y0uR_P4$$w0rd! -ti tvial/docker-mailserver:latest /bin/sh -c 'echo "$(doveadm pw -s SHA512-CRYPT -p $MAIL_PASS)"'`

It will generate a string. Then, just replace the old password string with the new one inside **config/postfix-accounts.cf**.

This config file looks like this:
 

> user@domain.tld | password_string

NB : Please notice the pipe character between email and password.

Brandl:
Maybe a takeaway from this is, that the setup.sh, should maybe support another command:

SUBCOMMANDS:

  email:

    ./setup.sh email add <email> <password>
    ./setup.sh email update <email> <password>
    ./setup.sh email del <email>
    ./setup.sh email list

Search for the EMail and replace the password.

ckiszka:
Nesousx -- Thanks!

tomav:
@ckiszka would you like to contribute to the project adding this feature on `setup.sh`?

ckiszka:
@tomav -- I will give it a shot. This is new to me, it may take me a little longer than you guys ;)

tomav:
No problem, we'll be happy to count you as a contributor!

tomav:
Fixed by #413

PR #383, opened 2016-11-17
Binary to edit a user's password

marceloschmidt:
I'm not sure if this is the best bash script it can be, but it works for me.
This script changes a user password.
We still need to restart the server for changes to be taken into account.

### DIFF
--- a/target/bin/editmailuser
+++ b/target/bin/editmailuser
@@ -1,0 +1,35 @@
+#!/bin/bash
+DATABASE=/tmp/docker-mailserver/postfix-accounts.cf
+
+function usage {
+  echo 'Usage: editmailuser <user@domain.tld> [new password]'
+  exit 1
+}
+
+function sed_string() {
+  SEARCH=$(echo $1 | sed 's/[]\/$*.^|[]/\\&/g')
+  REPLACE=$(echo $2 | sed 's/[\/&]/\\&/g')
+  SED_STRING="^${SEARCH}|.*$/${REPLACE}/"
+}
+
+if [ ! -z "$1" ]; then
+  USER=$1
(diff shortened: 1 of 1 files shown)


Josef-Friedrich:
related to #357

Influencer:
Just a suggestion, sed will do all the work in a single line. Try this out:

Remove lines 9-13(the sed_string function), remove 31 and replace line 32 from
`sed -i "s/${SED_STRING}" $DATABASE`
to
` sed -i.bak "s%^$USER.*%$ENTRY%g" $DATABASE `

That should search the database for $user and replace the line with the new $ENTRY. I added the .bak in case the user made a mistake and needed to rollback the change.

The only issue I could think of is if % was in the encryption. As far as I've been able to tell and test, I have not come across it so I don't believe would be.

Also should update setup.sh to include the usage for:
`$0 email update <email> <password>`

and add the case:
```
update)
    shift
    _docker_image editmailuser $@
    ;;
```

Let me know if I can make a pull request for anything to help.

Josef-Friedrich:
@marceloschmidt Thank you for submitting your script! Can you please write a test. Writing tests with [bats](https://github.com/sstephenson/bats) is so nice.

marceloschmidt:
I'll have to learn some bash programming 😂 but give me a few days and I'll add some tests! Thanks!

tomav:
Yeah, no problem @marceloschmidt

tomav:
Hi @marceloschmidt, did you get a chance to look at `bats`integration test?
We can help you if necessary. Feel free to ask. Tests are important to make this robust to changes.

tomav:
Fixed by #413

PR #390, opened 2016-11-27
Override hostname for kubernetes et al.

iMartyn:
In kubernetes land, it's not possible to set the hostname of a container, it's always set to a variation of the pod name, with random characters in it.

This PR should allow the env var OVERRIDE_HOSTNAME to replace the hostname commands that are used in the script.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,11 @@ Otherwise, `iptables` won't be able to ban IPs.
   - **empty** => admin
   - => Specify the password to bind against ldap
 
+##### OVERRIDE_HOSTNAE
+
+  - **empty** => uses the `hostname` command to get the mail server's canonical hostname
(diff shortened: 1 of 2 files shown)


tyranron:
There is a possibility: [v1.PodSpec hostname & subdomain](http://kubernetes.io/docs/api-reference/v1/definitions/#_v1_podspec). Please, look [here](https://github.com/tomav/docker-mailserver/issues/298#issuecomment-243980790) for example.
I'm successfully using this image in Kubernetes.

iMartyn:
Hi tyanron,

Sure, that's useful, but it doesn't give you the option of returning the dns name.  if you set subdomain, it includes the .svc.cluster stuff.  I still think that giving the option of specifically using env override is better.

tomav:
@iMartyn moved to another project.
Feel free to get this back to life if needed.

tomav:
Moved to #484

Issue #338, opened 2016-09-30
Restructure start-mailserver.sh


alinmear:
From my point of view it would be a big benefit, to structure the script with functions, for better maintenance. 

With this improvement it would be possible to better control the bootstrap of the container:

I would appreciate this step and would spent my time working on this.

Best Regards

tomav:
You're right, this file is a bit messy.
Can you work on that?

All, any thoughts?

Josef-Friedrich:
Functions for start-mailserver.sh would be great!

alinmear:
If you merge my pull request #335 then we would have the logic for container setup with postfix as central part as mta of a full featured mailserver:

Postfix <-> User database and User Validation <-> imap/pop server
Postfix <-> dovecot/postfix local <-> dovecot
Postfix <-> dovecot/postfix ldap (dovecot ldap needs to be implemented)<-> dovecot
Postfix <-> dovecot local/postfix ldap <-> dovecot
Postfix <-> dovecot ldap (needs to be implemented)/postfix local <-> dovecot

Postfix <-> ldap/activedirectory <-> kopano/zarafa

With this in mind, i would suggest the following structure:
do_setup_postfix ()  
do_setup_opendkim()
do_setup_opendmarc()
do_setup_ssl()
do_setup_permit_docker_options()
do_setup_saslauthd()
do_setup_spamassassin()
do_setup_amavis()
do_setup_fail2ban()
do_setup_dovecot () 
do_setup_kopano () 
do_setup_container() [e.g. persistent state)
do_init () 

At the end of the script:
control section --> validating ENV-VARS with possible setup, if not valid, then throw error on container startup e.g. user dovecot and kopano or cyrus sasl and dovecot sasl. After validating and settngs configuration state of the container do_init()

alinmear:
@tomav i will start here, when we can be sure, whether the pull request of @dennis95stumm, and also my new rebased pull request for the kopano feature has been successfully merged or not. Because both will result in some alterations of the start-mailserver script.

tomav:
Roger that!

tomav:
@alinmear tests are missing in your other PR by the way.

tomav:
We should have a pre function that checks configuration to stop the execution if a configuration is provided:
* if FQDN is not provided (already exists at the top of the script)
* display configuration (alias provided or not, regexp-alias provided or not, dkim provided or not...) before it starts the server

Let me know if I can help or if you want to talk about that.

[...]

PR #393, opened 2016-12-01
Restructure start-mailserver.sh #338

alinmear:
To make it easier to extend and maintain the logics of initialization of the mailserver and it's components i restructured the whole script into small junks/functions. 

**Let me explain the logics here:**
We have 5 steps to bring the mailserver up (they are running after one another) :
1. Check: these are functions to check the configuration sanity of the setup
2. Setup: here happens the main configuration initialization. 
3. Fix: within this section we place proper fixes for problems that may occur (like permissions of the maildirs)
4. Misc: within this section we can place functions that needs to be run before starting the daemons (like saving the container state)
5. Start: within this section all daemon processes getting started

I also made a simple registration logic, for **integrating new features** easily:

-  Step 1: Find the section where to place a feature
-  Step 2: Implement the feature within the proper part of the script
-   Step 3: register the function at the top of the script (also within the proper section)

**Integrating Checks and improve this pull request:**
While making integration tests one of the things i recognized was, that we are lacking checks, when a setup step is not performed successfully (for this see my TODOs within the script). Because while executing the 5 steps (check, setup, fix, misc, start-daemons) i implemented to check for a proper return value. right now when there is an error while executing a script the return value $? != 0 --> execution part failed --> script becomes defunc state.

This can be the starting point to make proper checks within each part of the script. If this is not wanted, we should comment out the check whether a function return 0 or not.

**Global and Default Vars:**
On top of the script i made an associative array to place global variables. the _setup_default_vars() function loops over this array and exports the variable. With this we can define Default/Global Variables easily without having to mess around.

Improvements are welcome!

### DIFF
--- a/target/start-mailserver.sh
+++ b/target/start-mailserver.sh
@@ -1,145 +1,433 @@
 #!/bin/bash
 
-die () {
-  echo "## Configuration error:"
-  echo >&2 "> $@"
-  exit 1
+##########################################################################
+# >> DEFAULT VARS
+#
+# add them here. 
+# Example: DEFAULT_VARS["KEY"]="VALUE"
+##########################################################################
+declare -A DEFAULT_VARS
+DEFAULT_VARS["VIRUSMAILS_DELETE_DELAY"]="${VIRUSMAILS_DELETE_DELAY:="7"}"
+##########################################################################
+# << DEFAULT VARS
+##########################################################################
+
+
+##########################################################################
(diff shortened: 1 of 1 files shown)


tomav:
Hi @alinmear, I think it's a pretty good start.
I'm gonna merge it and we'll discuss about improvements.

tomav:
@alinmear could you open an issue with suggested improvements? Thank you.

PR #400, opened 2016-12-17
use strong tls and ciphers

sylvaindumont:
Postfix configuration use strong protocols and ciphers, I don't know if there is a good reason for not using them in dovecot.


tomav:
Bonjour @sylvaindumont, pretty good input, thanks.

icewind1991:
This breaks logging in from my mail clients for me (kmail and gmail on android), the larger cipher suite in #418 fixes it

tomav:
Merged.

tomav:
Having problems with OSX Mail client.

PR #404, opened 2016-12-18
Fail2ban fix for restarting container

arcaine2:
Fail2ban doesn't seems to shutdown cleanly and leaves fail2ban.sock file that prevent it from starting after a container restart. That simple check should do the trick.


tomav:
Just restarted build.

tomav:
LGTM, let's see what Travis says.

PR #408, opened 2016-12-19
Fix Problem with Saslauthd and Postfix master.cf

alinmear:
The provided default postfix master.cf overwrites the configs for
saslauthd within main.cf. To make saslauthd work, we have to comment or
in this case delete the lines from master.cf to make the given configs
in main.cf work.

### DIFF
--- a/target/start-mailserver.sh
+++ b/target/start-mailserver.sh
@@ -442,6 +442,11 @@ log_level: 10
 EOF
 	fi
 
+		 sed -i \
+		 -e "/^[^#].*smtpd_sasl_type.*/s/^/#/g" \
+		 -e "/^[^#].*smtpd_sasl_path.*/s/^/#/g" \
+		 etc/postfix/master.cf
+
 	sed -i \
 		-e "s|^START=.*|START=yes|g" \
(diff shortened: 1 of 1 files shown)


tomav:
Hi @alinmear, what was the issue for the record?

alinmear:
The default values in master.cf makes any changes in main.cf useless. e.g. You are setting up the container, configuring everything properly, ldap, saslauthd and postfix main.cf. But in postfix master.cf there is the default option to use dovecot. So we have to remove this -o option from master.cf that the config options from main.cf are working.

In short the configs of postfix has the following hierarchy:
1. master.cf overrides with -o options main.cf
2. main.cf

In particular:

submission inet n       -       n       -       -       smtpd
  -o syslog_name=postfix/submission
  -o smtpd_tls_security_level=encrypt
  -o smtpd_sasl_auth_enable=yes
  **-o smtpd_sasl_type=dovecot**
  **-o smtpd_sasl_path=private/auth**

alinmear:
@tomav here are some sources for this:
* http://www.postfix.org/master.5.html
* http://serverfault.com/questions/462825/postfix-master-cf-versus-main-cf
* http://www.linuxquestions.org/questions/linux-server-73/postfix-main-cf-vs-master-cf-4175438528/

PR #413, opened 2016-12-21
Script to update users password, made test and updated setup.sh

Influencer:
I'm interested in getting the functionality into master so I'm submitting a separate pull request for this item.

This pull is related to #383 and #357 . 

A few variations as I had mentioned in #383 discussion:
I'm doing all of the searching and replacing in a single Sed statement. In testing I have found reliably no '%' special characters to be in the hashed password so I replaced the default separator with % instead.

I also included the .bak extension so the postfix_accounts.cf files can be rolled back in case of any error or issue.

I have updated setup.sh to include the new update function as well as included a separate test_update_password.bats as I was unsure if the norm was to add to the existing tests script. Also, not sure if I wrote the test correctly, it is my first time working with bats.


Influencer:
Not sure why the CI build failed? I have tested on my system and builds fine for me

tomav:
Thanks @Influencer 
Build hangs because of Clamav update.

```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

I restarted the build.

tomav:
I'm gonna change the config to avoid this issue.

tomav:
@Influencer can you add an integration test using `bats`?
Thank you.

tomav:
your test needs to be included in `test/tests.bats`.
Otherwise it won't be triggered :-)

tomav:
(later we'll have to refactor the test suite)

jcalton88:
I'll move the test into tests.bats. When you say missing integration test do you mean just move my test into the tests.bats or is the integration test a separate test that's needed?

tomav:
Just move it to the `tests.bats`.  (the label is more to know what is the issue status)

jcalton88:
Looks like I made the test fail by not removing the user after the test. Can you take a look and make sure that's the only thing failing

PR #427, opened 2016-12-24
TLS config and Ciphers - should fix #426

### LINKED ISSUE #426: Dovecot/postfix new SSL config breaks Mail OSX client
cc @dbellavista @icewind1991 

I'm gonna revert 2 latest changes.

tomav:


tomav:
cc @dbellavista

Issue #386, opened 2016-11-23
Mail relay does not work when amavis is disabled


mlegenhausen:
Configuration

```yml
  mailserver:
    build: mailserver/
    hostname: mail
    domainname: example.com
    environment:
      DISABLE_AMAVIS: 1
```

Relay server configuration

```
smtp_sasl_auth_enable = yes
smtp_sasl_password_maps = static:xxx:xxx
smtp_sasl_security_options = noanonymous
smtp_tls_security_level = encrypt
header_size_limit = 4096000
relayhost = [smtp.sendgrid.net]:2525
```

The error message

```
Nov 23 11:06:46 mail postfix/smtp[1013]: connect to 127.0.0.1[127.0.0.1]:10024: Connection refused
Nov 23 11:06:46 mail postfix/smtp[1013]: 3B6EB950: to=<to@example.com>, orig_to=<system@server.com>, relay=none, delay=0.14, delays=0.13/0.01/0/0, dsn=4.4.1, status=deferred (connect to 127.0.0.1[127.0.0.1]:10024: Connection refused)
```

The problem is that `content_filter = smtp-amavis:[127.0.0.1]:10024` is statically set even when amavis is disabled.

Current workaround is to add `content_filter =` to the `postfix-main.cf`.

I think this value needs to be setted depending  amavis is enabled or disabled?

tomav:
You're right, this should be fixed. Want to contribute?

mlegenhausen:
I can but I want to make sure that only amavis needs this content_listener whats about clamav?

tomav:
IMHO, clamav should also be disabled because triggered by amavis.

mlegenhausen:
So when `DISABLE_AMAVIS` is set clamAV is disabled automatically?

tomav:
Yes, amavis is the entrypoint for clamav and spamassassin.
If amavis is not running, clamav should not be running neither.
Let me know if I can help.

tomav:
But keeping amavis and disabling clamav should be a working solution too.
There's an old bug reported on this part:
https://github.com/tomav/docker-mailserver/issues/256

tomav:
`DISABLE_AMAVIS` does not exist anymore in `2.1`.
You can enable/disable spam and virus check separately.
I close this issue and add a new one to create `ENABLE_AMAVIS` env.

mlegenhausen:
Thanks for you fix :)

PR #432, opened 2016-12-25
BREAKING CHANGES:

### LINKED ISSUE #256: Bug in selective service
Bug found by @ep0ch2 in https://github.com/tomav/docker-mailserver/pull/250

> Many thanks for this, the following settings work:
> - DISABLE_AMAVIS=0
> - DISABLE_SPAMASSASSIN=1
> - DISABLE_CLAMAV=1
> 
> however with disable_amavis=1 i can't send/receive emails. so perhaps some extra config changes are needed,

@KyleOndy: I'm opening this issue for reminder.
Will you be able to work on it? May I help?

Thank you.

tomav:
* Removed DISABLE_AMAVIS
* Renamed DISABLE_* to ENABLE_* with 0 as default value. (this must be explicit)
* Added missing tests for ENABLE_*
* Improved readme and docker-compose example

Should fix #256 and #386

This will be released as `tvial/docker-mailserver:2.1`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,8 @@ run:
 		-v "`pwd`/test/config":/tmp/docker-mailserver \
 		-v "`pwd`/test":/tmp/docker-mailserver-test \
 		-v "`pwd`/test/onedir":/var/mail-state \
+		-e ENABLE_CLAMAV=1 \
+		-e ENABLE_SPAMASSASSIN=1 \
 		-e SA_TAG=1.0 \
 		-e SA_TAG2=2.0 \
 		-e SA_KILL=3.0 \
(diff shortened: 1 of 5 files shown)


kachkaev:
Great improvement, thanks for your work @tomav!

I'm wondering if this change needs to introduce a new major version an in my current view it should. The reason is that the new variable names are not backwards-compatible, so those who do docker pull for v2 will encounter issues. How about giving the new release v3?

tomav:
Hi @kachkaev, you're right, that's why I released a 2.1 (removed the `v` by the way), but not a 3.0 yet.

tomav:
There's still a `SMTP_ONLY` which should be `ENABLE_IMAP`.

PR #441, opened 2016-12-27
Fix mailuser tools

weo:
I tried to refactor and improve the mailuser tools (add/del/list and update). Certain user names were not exactly matched by the patterns: e.g. "user@domain" matched also "anotheruser@domain", or "a.user@domain" matched "anuser@domain".

I appreciate your project very much, it is really useful to me! Thank you so much!


tomav:
Thanks for spotting this!

PR #446, opened 2017-01-01
Fix /var/mail right at container initialization rather than startup + configure vhosts

kbeaugrand:


tomav:
Don't forget comments on the initial PR. 
Let me know when I can review this new one.

kbeaugrand:
I've commented the initial PR. You can review the second one.
________________________________
De : Thomas VIAL <notifications@github.com>
Envoyé : dimanche 1 janvier 2017 15:13:27
À : tomav/docker-mailserver
Cc : Kevin BEAUGRAND; Author
Objet : Re: [tomav/docker-mailserver] Fix /var/mail right at container initialization rather than startup + configure vhosts (#446)


Don't forget comments on the initial PR.
Let me know when I can review this new one.

-

tomav:
Lets discuss in this PR only, otherwise it won't be easy.

tomav:
What I understand is that you want to add domains but no real email address with these domains.
What is the use case? Only send emails?

target/start-mailserver.sh
```diff
@@ -95,7 +97,7 @@ function register_functions() {
 
 	################### >> fix funcs
 
-	_register_fix_function "_fix_var_mail_permissions"
+#	_register_fix_function "_fix_var_mail_permissions"
```
tomav: * all dovecot configuration is based on `docker:docker`(`5000:5000`). This should be kept.
* if this function becomes useless, it should be remove and not only commented out.

kbeaugrand:
I Want to host multiple domains emails. 
But as I'm in a Doker environnement, my slapd server is hosted on another container.

Working with mydestination will asked to postfix to work with local account and will try t map the email account to a local one (will not work because it's a lapd account). 

So I've to work with virtual domains.

target/start-mailserver.sh
```diff
@@ -16,6 +16,7 @@ DEFAULT_VARS["ENABLE_FETCHMAIL"]="${ENABLE_FETCHMAIL:="0"}"
 DEFAULT_VARS["ENABLE_LDAP"]="${ENABLE_LDAP:="0"}"
 DEFAULT_VARS["ENABLE_SASLAUTHD"]="${ENABLE_SASLAUTHD:="0"}"
 DEFAULT_VARS["SMTP_ONLY"]="${SMTP_ONLY:="0"}"
+DEFAULT_VARS["POSTFIX_MAILBOXDOMAINS"]="${POSTFIX_MAILBOXDOMAINS:="localhost"}"
```
tomav: These domains are additional to those created using accounts and aliases.
The variable may be renamed but I'd like to understand the purpose of the PR.

tomav:
Ok I understand. But I have a question: this is more a workaround than an improvement of the current LDAP implementation? I don't use LDAP so I won't be very helpful but a lot of users here are managing the accounts using LDAP.
Perhaps there's work to do/plan to make things easier and more reliable.

kbeaugrand:
I don't think that is a workaround. 

In fact. on my previous server (that I'm migrating). I also had postfix with ldap. But these were on the same server. I had to change the nsswitch, and so on inorder the make my users known by Linux OS. 
But with docker container I'm not able to do that. That's why I found only this way to make my issue working

I think that other users that use your image have the same problem. Perhaps they have found another way ?? but I didn't know another one.

tomav:
I don't have the answer not being a LDAP user. I think the best thing to do is to open an issue with your detailing your problem and a link to your PR. Perhaps current users may have suggestions.
I'll notify them when the issue will be open.

kbeaugrand:
OK

PR #448, opened 2017-01-02
Fix #443 - RIMAP support

### LINKED ISSUE #443: SASLAUTH rimap: does it work?
I want to use imap auth for smtp submission and configured 
```
-e ENABLE_SASLAUTHD=1 \
-e SASLAUTHD_MECHANISMS=rimap \
```
In start-mailserver.sh, function _setup_saslauthd(), that SASLAUTHD_MECHANISMS setting is overwritten in case SASLAUTHD_LDAP_SEARCH_BASE is empty or not set.

I worked around that and set SASLAUTHD_LDAP_SEARCH_BASE to a dummy value. Then MECHANISMS in /etc/default/saslauthd  is set to "rimap".

But authentication still fails when sending emails.

I used strace on saslauthd and saw no activity. This seems a little strange to me.

I found out that before 782152f authentication with imap worked.

SASL is a mystery to me, can anybody help?

weo:
#443

- Added unit tests to rimap sasl mechanism
- Enabled sasl auth in postfix when rimap mechanism is selected
- Additionally, it seems necessary to start saslauthd with option "-r" ("Combine the realm with the login") in that case

### DIFF
--- a/test/auth/sasl-imap-smtp-auth.txt
+++ b/test/auth/sasl-imap-smtp-auth.txt
@@ -1,0 +1,5 @@
+EHLO mail
+AUTH LOGIN
+dXNlcjFAbG9jYWxob3N0LmxvY2FsZG9tYWlu
+bXlwYXNzd29yZA==
+QUIT
--- a/test/auth/sasl-imap-smtp-auth.txt
+++ b/test/auth/sasl-imap-smtp-auth.txt
@@ -1,5 +1,0 @@
-EHLO mail
-AUTH LOGIN
-dXNlcjFAbG9jYWxob3N0LmxvY2FsZG9tYWlu
-bXlwYXNzd29yZA==
-QUIT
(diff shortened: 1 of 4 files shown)


test/auth/sasl-imap-smtp-auth.txt
```diff
@@ -0,0 +1,5 @@
+EHLO mail
```
tomav: Perhaps you should reuse the smpt auth (renaming the file to something more obvious, specifying the user perhaps). Because if we change a password, it should be done in both file (the probably look the same)

weo: Yes, reusing smtp-auth-login.txt makes sense.

Regarding checks of ENABLE_SASLAUTHD: perhaps we can drop that variable at all and just look at SASLAUTHD_MECHANISMS?

tomav: That's a option. perhaps LDAP users may have an idea on that specific point.
But I'm currently refactoring tests and I can say that's not obvious when we look at the different combinations with these variables.

tomav:
Thanks @weo!

PR #456, opened 2017-01-08
Fix update in setup.sh

theomega:
The updatemailuser needs parameters. Without this, the command can not work and fails with:

```
Usage: updatemailuser <user@domain.tld> [password]
no username specified

```


tomav:
Thank you @theomega !

PR #481, opened 2017-01-18
Update GeoLite db URL as logstash 5.x doesn't support old database fo…

citec:
As defined in [Logstash 5.x no longer supports .dat format Geo Databases](https://github.com/logstash-plugins/logstash-filter-geoip/issues/90)


tomav:
Thanks!

PR #483, opened 2017-01-18
Set transport.host on elasticsearch.yml config file

citec:
Set transport.host on elasticsearch.yml config file to avoid Bootstrap Checks failure on production environments


tomav:
Thanks!

PR #484, opened 2017-01-19
kubernetes fix

kcrawley:
This was derived from #390. I'm not trying to step on toes, I just happened to be setting up a mail server on my kubernetes homelab this evening.

I believe amavis should pass integration tests now with the changes I've made. We'll see.


tomav:
Thank you @kcrawley 
Could you add integration test on this new ENV?
Once done, I'll merge it.
Thank you.

target/start-mailserver.sh
```diff
@@ -18,6 +18,7 @@ DEFAULT_VARS["ENABLE_SASLAUTHD"]="${ENABLE_SASLAUTHD:="0"}"
 DEFAULT_VARS["SMTP_ONLY"]="${SMTP_ONLY:="0"}"
 DEFAULT_VARS["VIRUSMAILS_DELETE_DELAY"]="${VIRUSMAILS_DELETE_DELAY:="7"}"
 DEFAULT_VARS["DMS_DEBUG"]="${DMS_DEBUG:="0"}"
+DEFAULT_VARS["OVERRIDE_HOSTNAME"]="${OVERRIDE_HOSTNAME:="false"}"
```
tomav: I don't think "false" is a good default value in that case.
This should remain empty if not provided, no?

target/start-mailserver.sh
```diff
@@ -315,12 +317,18 @@ function check() {
 }
 
 function _check_hostname() {
-	notify "task" "Check that hostname/domainname is provided (no default docker hostname) [$FUNCNAME]"
+	notify "task" "Check that hostname/domainname is provided or overidden (no default docker hostname/kubernetes) [$FUNCNAME]"
+
+	if [[ ${DEFAULT_VARS["OVERRIDE_HOSTNAME"]} != "false" ]]; then
```
tomav: Cf. previous comment.

tyranron:
Additionally, I think it will be right to remove [this](https://github.com/tomav/docker-mailserver/pull/361/files#diff-2e69d4e5de4179e60cf6533e668bceb7R11) and restore [this](https://github.com/tomav/docker-mailserver/pull/361/files#diff-62fede8c21437c1625639aa79b0a97acL113) stuff for OpenDMARC with this PR.

Currently it uses `HOSTNAME` value in configuration, so [according to documentation](http://www.trusteddomain.org/opendmarc/opendmarc.conf.5.html) automatically resolves to container's `hostname` which is Pod's, not the one we want to specify.
>If the string "HOSTNAME" is provided, the name of the host running the filter (as returned by the gethostname(3) function) will be used.

Makefile
```diff
@@ -18,6 +18,22 @@ generate-accounts:
 
 run:
 	# Run containers
+	docker run -d --name mail_hostoverride \
```
tomav: Could you use an existing container like `mail_smtponly`?
It avoids to launch to many of them consuming CPU and memory.
Otherwise it's ok!
Thanks.

test/tests.bats
```diff
@@ -9,6 +9,11 @@ load 'test_helper/bats-assert/load'
   assert_failure
 }
 
+@test "checking configuration: hostname/domainname override" {
+  run docker exec mail_hostoverride /bin/bash -c "cat /etc/mailname | grep my-domain.com"
```
tomav: Use `mail_smtponly `

kcrawley: :+1:

tomav:
LGTM, waiting for travis (just restarted it)

tomav:
@tyranron this can be merged without your last comment or not?
Perhaps this can be added after?

kcrawley:
https://github.com/tomav/docker-mailserver/blob/master/target/opendmarc/opendmarc.conf#L11-L12

It's not difficult to `sed` and replace those values, would current tests cover that change?

tomav:
I'm gonna merge this.
@kcrawley thanks for the job.
@tyranron feel free to open an issue if some works needs to be done on that topic

Thank you guys.

tomav:
Thanks thanks @iMartyn fir the initial PR.

PR #517, opened 2017-02-10
Hostname override fixes for docker option --net=host in conjunction with OVERRIDE_HOSTNAME

flole:
I have noticed that my emails contain the hostname of my docker host when I use the docker option `--net=host` with the env-var `OVERRIDE_HOSTNAME`. That is because postfix, dovecot and opendmarc user the `hostname` command to retrieve the name of host. Until now the hostname was only overwritten for amavis, but these changes will overwrite it for all mentioned services.


tomav:
Thanks @flole

Issue #305, opened 2016-09-03
[Question] Postfix and Dovecot delivery integration


tyranron:
According to current (release 2.0.0) configuration Postfix delivers mails to account's mailboxes via LDA command:

[/target/postfix/main.cf:64-65](https://github.com/tomav/docker-mailserver/blob/2.0.0/target/postfix/main.cf#L64-L65)
```
# Enable Dovecot Sieve
virtual_transport = dovecot
```

[/target/postfix/master.cf:66-71](https://github.com/tomav/docker-mailserver/blob/2.0.0/target/postfix/master.cf#L66-L71)
```
#
# Dovecot LDA configuration
#

dovecot unix    -       n       n       -       -      pipe
  flags=DRhu user=docker argv=/usr/lib/dovecot/deliver -f ${sender} -d ${user}@${nexthop} -m ${extension}
```

So Postfix uses `/usr/lib/dovecot/deliver` util to deliver message to Dovecot. This applies some additional requirements, i.e. Postfix and Dovecot should have access to this directory under the same user and etc.

My question is:
Why not use LMTP over unix socket as [this page](http://wiki.dovecot.org/LDA) recommends?
LMTP really simplifies configuration and decreases processes coupling.
Do I miss any caveats?

Josef-Friedrich:
Sounds interessting

tomav:
Hi @tyranron, thanks for this input.
@00angus, any comment?

00angus:
lmtp! Very good IMHO.
I was for lmtp even before:
https://github.com/tomav/docker-mailserver/pull/166

Just my 0,02€ :)

-- Marco

On Sep 4, 2016 20:42, "Thomas VIAL" <notifications@github.com> wrote:

Hi @tyranron <https://github.com/tyranron>, thanks for this input.
@00angus <https://github.com/00angus>, any comment?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/tomav/docker-mailserver/issues/305#issuecomment-244621325>,
or mute the thread

Josef-Friedrich:
Why still hesitate? @tyranron can you implement LMTP in `docker-mailserver`?

tyranron:
@Josef-Friedrich yeap, I'll try in next few days.

tomav:
Hi @tyranron, any update on this? Thank you.

tyranron:
[...]

PR #520, opened 2017-02-13
Add support for recipient delimiters with aliases fixes #442

### LINKED ISSUE #442: Use recipient_delimiter with aliases
It is currently not possible to use the `recipient_delimiter` (by default +) with aliases configured in `config/postfix-virtual.cf`. The reason is, that dovecot doesn't know about the aliases.

Let's say you have `me@somedomain.com` configured as an account in `config/postfix-accounts.cf` and an alias `full.name@somedomain.com` in `config/postfix-virtual.cf`. When a mail with recipient `full.name+idonttrustyou@somedomain.com` gets received by the mail server, it is processed just fine by postfix. It then tries to deliver the mail to dovecot (as per `virtual_transport` in `target/postfix/main.cf`). Because dovecot is unaware of the aliases it rejects the mail with 
> User doesn't exist: full.name@somedomain.com

nwolber:
I finally got time to add a test case for this. If there is anything that keeps you from merging, please let me know.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ notifications:
     secure: TTo1z9nbZCWcIdfPwypubNa3y+pwvfgDGlzEVAGEuK7uuIpmEoAcAUNSSPTnbewDGHnDl8t/ml93MtvP+a+IVuAKytMqF39PHyoZO7aUl9J62V+G75OmnyGjXGJm40pQosCS6LzqoRRYXotl9+fwH568Kf4ifXCrMZX1d+ir7Ww=
   webhooks:
     urls:
-      - https://webhooks.gitter.im/e/7c5e56a8257cdec003ab
+      - https://webhooks.gitter.im/e/7ec65291bfb4bf1f8518
     on_success: always
     on_failure: always
     on_start: never
--- a/Makefile
+++ b/Makefile
@@ -149,6 +149,7 @@ fixtures:
 	docker exec mail /bin/sh -c "nc 0.0.0.0 25 < /tmp/docker-mailserver-test/email-templates/amavis-virus.txt"
 	docker exec mail /bin/sh -c "nc 0.0.0.0 25 < /tmp/docker-mailserver-test/email-templates/existing-alias-external.txt"
 	docker exec mail /bin/sh -c "nc 0.0.0.0 25 < /tmp/docker-mailserver-test/email-templates/existing-alias-local.txt"
+	docker exec mail /bin/sh -c "nc 0.0.0.0 25 < /tmp/docker-mailserver-test/email-templates/existing-alias-recipient-delimiter.txt"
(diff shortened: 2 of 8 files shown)


nwolber:
The TravisCI integration seems to be broken. The automatic build got canceled because setting up the containers took more than 20 mins. Is there a way to re-run the tests other than a new commit?

tomav:
Restarted.

flole:
Can it be that this undo the changes made by #305 and #360?

tomav:
@flole, yep it will. - cc @00angus, @tyranron

tyranron:
@tomav @nwolber  
Interesting... I don't think that getting rid of lmtp is best option.  
I'll investigate the issue a lit bit further in few days and will look can be this considered with lmtp or not.

tyranron:
@tomav I've submitted [a PR](https://github.com/tomav/docker-mailserver/pull/528) which adds integration test only provided by @nwolber in this PR. And, as we see, latest `master` passes it.

So, @nwolber, are you sure that this test reproduces your issue?

There is also an [official Dovecot LMTP doc](http://wiki.dovecot.org/LMTP#Address_extension_delivery) which refers us to `lmtp_save_to_detail_mailbox = yes` option. I've tried to enable it, and the only thing is changed that mail is delivered to appropriate folder (specified after delimiter), not to default as happens now.

nwolber:
@tyranron you're totally right. The test only tested the good case. Unfortunately the failure case also produces an e-mail (bounced) with the exact same log message. I added a new step in the test case, that checks for the status=bounced message. I also noticed that the failures seems to occur only if recipient_delimiter is set something other than the standard '+' sign. I also added that to the test case.

I ran the tests both with the lmtp configuration currently present in master and the lda configuration in this PR. The test now fails with lmtp and succeeds with lda.

If there is a solution that gets this working if lmtp I would be more than happy to do that. Unfortunately I don't see how. Any help would be greatly appreciated.

tyranron:
@nwolber thank you for update. I'll dig into and will try to investigate further.

tyranron:
@nwolber I reproduced your test case again in [this PR](https://github.com/tomav/docker-mailserver/pull/528). Just changed `recipient_delimiter` from `-` to `~` to [not collide with other test cases](https://github.com/tomav/docker-mailserver/blob/3d553961d37311ee7b9e60ac808625419817c992/test/email-templates/existing-regexp-alias-external.txt#L6).

The latest master [passes it](https://travis-ci.org/tomav/docker-mailserver/builds/206707386).
The problem was that `recipient_delimiter` was changed in Postfix configuration, but not in Dovecot configuration. This is mentioned in [Dovecot documentation](http://wiki.dovecot.org/LMTP#Address_extension_delivery) I referred earlier.

nwolber:
Yep, you are right. I already had that configuration, but for some other reason it didn't work as expected. I fixed that now and it works. I think this PR can be closed then and #528 should be merged instead. I will also close #442.

Thanks again for your help.

PR #540, opened 2017-03-11
Fixed virus-wiper script when using the default value for VIRUSMAILS_…

danielpanteleit:
…DELETE_DELAY

Moved variable VIRUSMAILS_DELETE_DELAY to Dockerfile to make it visible to cron. Also removed echo in cron job to prevent sending mail for each run.

I would also suggest moving all default values to the Dockerfile so one can easily identify the used values with `docker inspect`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,10 @@
 FROM ubuntu:14.04
 MAINTAINER Thomas VIAL
 
+# Default values
+ENV VIRUSMAILS_DELETE_DELAY=7 \
+    ONE_DIR=0
+
 # Packages
 RUN DEBIAN_FRONTEND=noninteractive apt-get update -q --fix-missing && \
   apt-get -y upgrade && \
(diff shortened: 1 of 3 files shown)


tomav:
Restarted build.

danielpanteleit:
OK, this did not work. Apparently the environment variable is cleared before the cron job runs. Does this actually work when setting `VIRUSMAILS_DELETE_DELAY`?

There are a few options left, but they all have to be done in the start script:

1) Setting `VIRUSMAILS_DELETE_DELAY` in the crontab
2) Changing `virus-wiper` to take a parameter and setting up the cron job in the start script with the value as a parameter.
3) Creating `virus-wiper` with the actual value already in place
4) Writing environment variables to a file and reading it in `virus-wiper`

I would vote for (2) as I think this is the most straight-forward solution.

tomav:
Hmm it seems that current implementation does not work as expected, you're right (i've done that myself).
When I execute `virus-wiper`, manually it works.
It does not when triggered by CRON.

I think that putting the ENV in `/etc/environment` should do the trick if you want to try.

danielpanteleit:
The start script will now write the variable to /etc/environment. I tested it manually this time (by adding a cron job which runs every minute)

tomav:
Thanks @danielpanteleit

PR #550, opened 2017-03-16
Add restart: always

strarsis:
This PR adds restart: always to the containers in docker-compose.yml for 
auto-restarting them in the event of docker service or host restart.


tomav:
Thanks.

PR #553, opened 2017-03-19
Changed max message size

Dschinghis-Kahn:
Added config to increase the maximal message size from 10MB (default) to 40MB.

### DIFF
--- a/target/postfix/main.cf
+++ b/target/postfix/main.cf
@@ -13,6 +13,7 @@ mydestination =
 relayhost =
 mynetworks = 127.0.0.0/8 [::1]/128 [fe80::]/64
 mailbox_size_limit = 0
+message_size_limit = 40960000
 recipient_delimiter = +
 inet_interfaces = all
 inet_protocols = all


tomav:
Hi @Dschinghis-Kahn, why this should go in the default config?

Dschinghis-Kahn:
Hi @tomav, this is a necessary setting if you want to send attachments. For example, with the default you can only send mails with a maximum size of ~7MB (because of base64). That's 2-3 pictures for my camera...
I wonder why there is a size limit on the sending side anyway?! On the receiveing mail server ... okay, but why should one prohibit one self of sending larger mails?
However, this option makes it possible to send larger mails and has no effect when sending small (text only) mails.

tomav:
Hi @Dschinghis-Kahn, thanks for your answer.
Of course I understand the need, but I'm not sure this message size should go in default config and impact all users with default settings.

Do you know that you can set your own settings using [postfix override](https://github.com/tomav/docker-mailserver/wiki/Override-Default-Postfix-Configuration) feature?

thiesschneider:
would be great to have this as env :)

Dschinghis-Kahn:
Overriding with postfix-main.cf worked perfectly. Thanks!

PR #554, opened 2017-03-19
Fix battery drain for mobile clients

huncode:
Regarding too frequent IDLE notifications mobile clients like k9mail have a massive battery drain.
More info:
https://github.com/k9mail/k-9/issues/1290
https://peterkieser.com/2011/03/25/androids-k-9-mail-battery-life-and-dovecots-push-imap/

P.S. Please correct me if I have to add additional changes. I've read contributing guide, but this change doesn't affect tests and build step.


huncode:
As mentioned on http://wiki.dovecot.org/Timeouts after login client idle timeout is set to 30 minutes, so it wouldn't affect desktop clients and others.

tomav:
Thanks @huncode

PR #556, opened 2017-03-19
Fix #526: fatal: no SASL authentication mechanisms

### LINKED ISSUE #526: "fatal: no SASL authentication mechanisms" when running SMTP_ONLY=1
I'm trying to get docker-mailserver up and running, so I'm able to send mails from my containers. 

I created the server with the following compose file:
`version: '2'
services:

  xxx_mail:
    restart: always
    container_name: xxx_mail
    image: tvial/docker-mailserver:2.1
    hostname: mails
    domainname: xxx.xx
    expose:
      - 25
    ports:
      - "xxx.xxx.xxx.xxx:25:25"
      - "127.0.0.1:143:143"
      - "127.0.0.1:587:587"
      - "127.0.0.1:993:993"
    mem_limit: "512m"
    networks:
      - xxx-mail-backend
    volumes:
      - /data/xxx/mailserver/maildata:/var/mail
      - /data/xxx/mailserver/mailstate:/var/mail-state
      - ./config/:/tmp/docker-mailserver
    environment:
      - ENABLE_SPAMASSASSIN=0
      - ENABLE_CLAMAV=0
      - ENABLE_FAIL2BAN=0
      - ENABLE_POSTGREY=0
      - ONE_DIR=1
      - DMS_DEBUG=0
      - SMTP_ONLY=1
      # SSL: empty means disabled. as we don't have any mailclients (yet), we don't need it.
      #- SSL_TYPE=letsencrypt
      - PERMIT_DOCKER=network

networks:
    xxx-mail-backend:
      external:
        name: xxx-mail-backend
`

when trying to send a mail from another container (which is in the same network), it fails. in the logs it just says
Feb 22 00:14:21 mails postfix/smtpd[1000]: error: open database /etc/postfix/vmailbox: No such file or directory
Feb 22 00:14:21 mails postfix/smtpd[1000]: fatal: no SASL authentication mechanisms

from the other container, I'm using ssmtp with mailhub=mail (where mail is the alias mapped into the other container). 
there I just get a "sendmail: Invalid response SMTP server"
returned. 

Any suggestions? I don't want any virus scanning or anything.. I only want to send mails that are DKIM signed. 

thx

alinmear:
When using the container with SMTP_ONLY = 1,  the postfix process within the container fails
on ehlo because there is no valid sasl authentication mechanism available. This happens because sasl has been enabled within postfix/main.cf per default but sasl is not configured.

To fix this _setup_postfix_sasl does not depend anymore on ENABLE_SASLAUTHD and will check in it's logic, whether to enable or disable sasl explicit within postfix/main.cf.

### DIFF
--- a/target/start-mailserver.sh
+++ b/target/start-mailserver.sh
@@ -90,7 +90,6 @@ function register_functions() {
 
 	if [ "$ENABLE_SASLAUTHD" = 1 ];then
 		_register_setup_function "_setup_saslauthd"
-		_register_setup_function "_setup_postfix_sasl"
 	fi
 
 	if [ "$ENABLE_POSTGREY" = 1 ];then
@@ -107,6 +106,7 @@ function register_functions() {
 	_register_setup_function "_setup_postfix_hostname"
 	_register_setup_function "_setup_dovecot_hostname"
 
+	_register_setup_function "_setup_postfix_sasl"
 	_register_setup_function "_setup_postfix_override_configuration"
 	_register_setup_function "_setup_postfix_sasl_password"
 	_register_setup_function "_setup_security_stack"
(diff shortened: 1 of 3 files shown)


tomav:
I think you should add an integration test for that.

alinmear:
Added integration test

alinmear:
Just to be accurate on this issue, i just coverd the testing for this particular problem, checking whether sending an email is working or not. But i think we should cover more test cases for the mailserver state `SMTP_ONLY` use-cases.

Here some examples:

- Check Mail send is working
- Check SASL authentication is working
- Check Relay is working
- Check Relay with sasl authentication is working

Maybe we could create a section for this.

PR #563, opened 2017-03-29
Fixes #496 enabling all domains to get spam headers and spam checks 

### LINKED ISSUE #496: multiple users: spamassassin not working properly
I first configured the container with one user and spamassassin works fine and as expected for months. Then I added two new users with their own directory for receiving mails and for the life of me I can't get spamassassin headers added to incoming emails:

> Return-Path: <eyxdzn@domain.net>
> Delivered-To: support@domain.com
> Received: from localhost (localhost [127.0.0.1])
> 	by mail.domain.com(Postfix) with ESMTP id 8290F1723BE
> 	for <support@domain.com>; Sun, 29 Jan 2017 01:40:10 +0000 (UTC)
> **X-Virus-Scanned: Debian amavisd-new at domain.net**
> Received: from mail.domain.com([127.0.0.1])
> 	by localhost (mail.domain.com[127.0.0.1]) (amavisd-new, port 10024)
> 	with ESMTP id P5EdkQSiJxjI for <support@domain.com>;

So Amavis passes the mail off to Virus Scanning, works fine. But it doesn't pass it to Spamassassin?

My SA Tags are as follows:

> $sa_tag_level_deflt = -999;  # add spam info headers if at, or above that level
> $sa_tag2_level_deflt = 3.4; # add 'spam detected' headers at that level
> $sa_kill_level_deflt = 99; # triggers spam evasive actions
> $sa_dsn_cutoff_level = 10;   # spam level beyond which a DSN is not sent
> 

* I checked permissions on the new users' postfix folders and they look OK. They appear writable.
* Nothing of note is showing up in mail.err in /var/log/mail

1. In theory, amavis is setup to process Spamassassin for all users with a single bayes database/tokens?
2. Is there anything in amavis/spamassassin that would stop Spamassassin from adding X-Spam headers to these additional postfix users' mails?

tomav:
Integration tests included.


tomav:
Here is the PR @makloda & @theomega

theomega:
Not sure if this is a good solution. The underlying idea (of adding the `60-dms_default_config`) is good. But why copy all the files out into this repo. How do we ensure that they are in sync with the packages (what if a bug is fixed in the files by the ubuntu packages?)

tomav:
I can only add the new file instead of the whole directory.
That should be better.
But doing things with `sed` in `Dockerfile` is pretty hard to maintain too.

PR #564, opened 2017-03-29
Include whitelist_clients.local in postgrey setup from config folder

nueaf:
Copies whitelist_clients.local from config dir to /etc/postgrey. 
Test is made by communicating directly with postgrey via nc on port 10023 in order to test for the correct client_name instead of localhost. 

Credit for that solution to this blog post: http://www.brandonchecketts.com/archives/manually-testing-postgrey-through-a-telnet-session

### DIFF
--- a/target/start-mailserver.sh
+++ b/target/start-mailserver.sh
@@ -539,6 +539,7 @@ function _setup_postgrey() {
 	if [ $TEXT_FOUND -eq 0 ]; then
 		printf "POSTGREY_TEXT=\"$POSTGREY_TEXT\"\n\n" >> /etc/default/postgrey
 	fi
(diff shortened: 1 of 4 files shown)


tomav:
Thanks @nueaf

kortenkamp:
This leads to 
`mail    | cp: cannot stat '/tmp/docker-mailserver/whitelist_clients.local': No such file or directory`
on startup if there is no `whitelist_clients.local` file.  After updating to the latest image my server seems to no longer accept mail – while investigating the reason I found this.

nueaf:
Sorry about that. I will add a check to see if the file exists before trying to copy it.

PR #568, opened 2017-04-07
Allow accounts which do not start with a-z (fixes #567)

### LINKED ISSUE #567: [Breaking Change] Mail accounts starting with a non-letter don't work anymore
Due to the change of `target/start-mailserver.sh` in 912cb5676b9538eb1c799b3c229c7e8dc9c94f40, mail accounts whose names do not start with a lowercase letter are ignored when the `postfix-accounts.cf` file is parsed during mailserver startup. As a consequence, logins for those accounts don't work anymore and the accounts are not known to the system so mail is potentially lost. Furthermore, fail2ban blocks the IP of clients which try to access.

In my case, I use a mail account which starts with a number for one of my mail addresses. Also, mailboxes starting with an uppercase letter (or even something else?) will run into problems.

Probably the correct rule would be to ignore lines which are starting with a comment (#) or are empty, but accept everything else (for backwards compatibility).

mplx:
Commit 912cb5676b9538eb1c799b3c229c7e8dc9c94f40 restricts accounts to begin with lower case a-z

The modified regex now ignores empty lines, lines with one or multiple blanks and lines with comments (#) but allows accounts like 123@domain.tld.

Fixes #567

### DIFF
--- a/target/start-mailserver.sh
+++ b/target/start-mailserver.sh
@@ -463,7 +463,8 @@ function _setup_dovecot_local_user() {
 
 		# Creating users
 		# 'pass' is encrypted
-		grep "^[a-z]" /tmp/docker-mailserver/postfix-accounts.cf | while IFS=$'|' read login pass
+		# comments and empty lines are ignored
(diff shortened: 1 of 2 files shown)


tomav:
Thanks

PR #532, opened 2017-02-25
Upgrade to ubuntu:16.04

42wim:
Upgrade to ubuntu:16.04. 
Fixes #500 
All tests are passing.

PR #571, opened 2017-04-13
Upgrade to 16.04

kiliant:
Another try at upgrading to 16.04, as tests work from clean builds just fine locally.

Forked from #532.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ ENV ONE_DIR=0
 RUN apt-get update -q --fix-missing && \
   apt-get -y upgrade && \
   apt-get -y install postfix
-RUN DEBIAN_FRONTEND=noninteractive apt-get update -q --fix-missing && \
+RUN apt-get update -q --fix-missing && \
   apt-get -y upgrade && \
   apt-get -y install --no-install-recommends \
     amavisd-new \
@@ -33,6 +33,7 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update -q --fix-missing && \
     gamin \
     gzip \
     iptables \
+    locales \
     libmail-spf-perl \
     libnet-dns-perl \
     libsasl2-modules \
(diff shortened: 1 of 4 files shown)


kiliant:
Weird thing is, that the local checks should have failed, too. 
```
run docker exec mail_override_hostname /bin/sh -c "ls -A /var/mail/localhost.localdomain/user1/new | wc -l | grep 1"
```
fails, but that is not shown by the test run (bug?).
The Makefile has been fixed and the tests have been adapted.

Dockerfile
```diff
@@ -8,10 +8,14 @@ ENV ONE_DIR=0
 # Packages
 RUN apt-get update -q --fix-missing && \
   apt-get -y upgrade && \
+  apt-get -y install postfix
+RUN DEBIAN_FRONTEND=noninteractive apt-get update -q --fix-missing && \
```
tomav: `DEBIAN_FRONTEND=noninteractive` already defined line 4

test/config/dovecot-lmtp/local.conf
```diff
@@ -1 +1,2 @@
 mail_max_userip_connections = 69
+recipient_delimiter = ~
```
tomav: This is not `+` in postfix?

tomav:
No BC Break detected?

kiliant:
These were changes introduced by @42wim.

42wim:
The dovecot ones are not from me btw. Check my original PR

Dockerfile
42wim: Iirc this was the only way it worked. Putting it in the already defined one doesn't complete postfix install

kiliant:
You're right sorry. Weird, I did not change that either. Reversing.

kiliant:
Those changes were caused by running the tests locally and slipped through the commit - now removed.

tomav:
This will be `2.3`.

tomav:
Thanks @kiliant and @42wim

PR #573, opened 2017-04-15
Mention submodule dependency

arneke:
Mention that you have to manually initialize and update submodules, otherwise you find out after a very long wait.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,6 +23,7 @@ The development workflow is the following:
 
 - Fork project and clone your fork
 - Create a branch using `git checkout -b branch_name` (you can use `issue-xxx` if fixing an existing issue)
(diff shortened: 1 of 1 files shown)


kiliant:
👍 haha had to find that out myself.

ci should be just fine, i had those "hickups" as well. maybe @tomav can restart it

arneke:
Actually, I had intermittent issues with that PGP key as well, the one that killed the build, as well as multiple failures in a row with db.local.clamav.net (round robin DNS afaik). So I have been trying to get clean build for two days now.

Pretty sure it's not an issue with the server I am building this on, it's a bit dated, but on a solid connection. Making the build more resilient, or making a base image to build on top of, may qualify for a separate ticket. But I haven't looked into what you have discussed previously.

tomav:
Restarted

tomav:
Thanks

PR #595, opened 2017-05-06
Support for modifying Postfix' master.cf

MadsRC:
Support for modifying Postfix' master.cf, using the  syntax, in postfix-master.cf

Fixes issue-594


tomav:
@MadsRC any update on this? Thank you.

MadsRC:
Time has been extremely short with an infant at home, so I haven't had the time to dig into the integration testing yet - That, and I'm not familiar with the bats framework.

I'm not sure how much you want the test to cover? Reading the test file (Again, not familiar with bats, so I might be wrong) getting inspiration from the testing of main.cf I suppose a "postfix check" would cover much of it?

tomav:
Yep the same kind we used on main.cf

MadsRC:
@tomav - Added the integration test, and the build succeeds on my local machine when running "make". Seems like the build fails on travis, however, the unit test i made seem to have run fine?

tomav:
Thanks @MadsRC

Issue #589, opened 2017-05-03
Disable Spamassassin Subject Tag


elbracht:
Is there a way to disable the `sa_spam_subject_tag` in `/etc/amavis/conf.d/20-debian_defaults`?

Spamassassin shall not add \*\*\*SPAM*** to the mail subject.

tomav:
Hi @elbracht, true [this file](https://github.com/tomav/docker-mailserver/wiki/FAQ-and-Tips#how-can-i-manage-my-custom-spamassassin-rules) and let me know (not tested)

elbracht:
I found the following on https://wiki.apache.org/spamassassin/SubjectRewrite:

```
Warning: if you are running SA through amavisd or qmail-scanner, those apps do their own message rewriting and SpamAssassin cannot perform these rewrites.
```

Is it possible to change the amavis configuration outside the container?

elbracht:
Implemented in #596

PR #596, opened 2017-05-08
Environment variable for amavis subject tag

elbracht:
Ref: #589

Environment variable to change the amavis tag `sa_spam_subject_tag` outside the container .

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,7 @@ run:
 		-e SA_TAG=-5.0 \
 		-e SA_TAG2=2.0 \
 		-e SA_KILL=3.0 \
+		-e SA_SPAM_SUBJECT="SPAM: " \
 		-e VIRUSMAILS_DELETE_DELAY=7 \
 		-e SASL_PASSWD="external-domain.com username:password" \
 		-e ENABLE_MANAGESIEVE=1 \
--- a/README.md
+++ b/README.md
@@ -146,6 +146,12 @@ Note: this spamassassin setting needs `ENABLE_SPAMASSASSIN=1`
 
 Note: this spamassassin setting needs `ENABLE_SPAMASSASSIN=1`
 
+##### SA_SPAM_SUBJECT
+
+  - **\*\*\*SPAM\*\*\*** => add tag to subject if spam detected
+
+Note: this spamassassin setting needs `ENABLE_SPAMASSASSIN=1`
+
 ##### ONE_DIR
 
   - **0** => state in default directories
@@ -336,4 +342,3 @@ Enabled by ENABLE_POSTFIX_VIRTUAL_TRANSPORT. Specify the final delivery of postf
     - lmtps:inet:<host>:<port> (secure lmtp with starttls, take a look at https://sys4.de/en/blog/2014/11/17/sicheres-lmtp-mit-starttls-in-dovecot/)
     - lmtp:<kopano-host>:2003 (use kopano as mailstore)
     - etc.
-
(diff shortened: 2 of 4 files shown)


tomav:
Hi @elbracht, don't forget integration tests.

elbracht:
@tomav do you have an idea why I can’t use a string as environment variable? If I use a number as environment variable in `Makefile` it works.

Works:

```
-e SA_SPAM_SUBJECT=10.0 \
```

Don’t work:

```
-e SA_SPAM_SUBJECT="SPAM: " \
```

tomav:
Still 2 errors in this build. How can we move forward?

elbracht:
@tomav I can’t use a string as environment variable to update the `$sa_spam_subject_tag` in `start-mailserver.sh`. With a number or a fixed value as environment variable it works.

number (Makefile):

```
docker run -d --name mail \
...
-e SA_SPAM_SUBJECT=10.0 \
```

fixed value (start-mailserver.sh):

```
sed -i -r 's/^\$sa_spam_subject_tag (.*);/\$sa_spam_subject_tag = fixed value';/g' /etc/amavis/conf.d/20-debian_defaults
```

tomav:
The error is [here](https://github.com/elbracht/docker-mailserver/blob/5b9908bd04cecc49366c842e26af37393e4acedb/test/tests.bats#L438)

elbracht:
@tomav I know that the test failed on this line but this is not the problem. 

I think the problem is that the environment variable is not passed to the start script `start-mailserver.sh` ([here](https://github.com/elbracht/docker-mailserver/blob/5b9908bd04cecc49366c842e26af37393e4acedb/Makefile#L30)).

Or the script doesn't write the variable to the config file ([here](https://github.com/elbracht/docker-mailserver/blob/5b9908bd04cecc49366c842e26af37393e4acedb/target/start-mailserver.sh#L936)).

tomav:
I have an error with `sed` when trying to run the line for subject tag configuration.
This seems to work better:

```
sed -i -r 's/^\$sa_spam_subject_tag (.*);/\$sa_spam_subject_tag = '"'$SA_SPAM_SUBJECT'"';/g'
```

Let me know if it works for you.

elbracht:
@tomav Nice! 👍 

Now the local tests have been passed.

tomav:
Here we are!

jholster:
Sorry for reopening this, but it's not clear for me how to disable the subject rewrite entirely. If SA_SPAM_SUBJECT is empty/undefined, it seems to default to ***SPAM***. I would like to keep my subject unchanged, because I'm moving spams to separate folder.

johansmitsnl:
@jholster maybe add a " " space in the variable might fit your needs.

jholster:
@johansmitsnl Tried that, didn't help, maybe the configuration parsers trims the lines? Anyway, sounds  like hack. Can we additionally have SA_ENABLE_SPAM_SUBJECT_REWRITE boolean variable?

I think it's very common case to move spam messages into separate spam folder. Once in a while I skim through my spams and move false positives to inbox/archive folder (to teach SpamAssassin). But the moved messages will have "***SPAM***" in subject line forever, which is kind of ugly.

elbracht:
@jholster The following environment variable should solve your problem:

```
-e SA_SPAM_SUBJECT=undef
```

jholster:
@elbracht Thanks, that finally solved the issue. If anyone else wonders how the .env parser distinguish between undef and "undef" – it does not, "undef" is just a magic string, see [here](https://github.com/tomav/docker-mailserver/blob/master/target/start-mailserver.sh#L1217).

PR #613, opened 2017-05-23
add switchable dovecot feature address extension

gutmensch:
* currently the container does not support to sort
          incoming mails with an address extension into subfolders
          automatically, since the default for dovecot/lmtp is "no"

	* with this feature enabled, mails to user+subfolder@d.com
          will cause dovecot to automatically create subfolder within
          your inbox

	* the feature is IMO useful in a lot of places, e.g. for
          _dmarc records, where you can automatically sort into
          forensic and aggregated report folders


gutmensch:
It's anyway easier to just set the variable in dovecot.cf or make it default, so lets skip this. ;)

PR #629, opened 2017-06-09
Increase message size limit to 25MB

JCBird1012:
This PR increases the limit Postfix enforces on message attachment size.

As of currently with ```docker-mailserver```, if someone were to send you a message, with an attachment greater than 10240000 bytes (roughly 10MB), Postfix would silently drop it, and return an error to the sender.

Especially in this time of large files, attachments are creeping closer and closer to the default Postfix size limit. This will hopefully alleviate some headaches of dropped mail because of attachments that are too large.

However, where some users may potentially run into some issues regarding mailbox size. I believe the current default mailbox limit of Postfix is roughly 50MB... Two messages with full size attachments (at 25MB each) would fill up a user's mailbox. 
That being said, things weren't much better at the default attachment size either (5 messages would fill up a user's mailbox).

Increasing the size limit of users' mailboxes may be beyond of the scope of this PR, but nonetheless something to consider. I can add that fix if you wish.

### DIFF
--- a/target/postfix/main.cf
+++ b/target/postfix/main.cf
@@ -63,6 +63,9 @@ virtual_mailbox_domains = /etc/postfix/vhost
 virtual_mailbox_maps = texthash:/etc/postfix/vmailbox
 virtual_alias_maps = texthash:/etc/postfix/virtual
 
+# Increase the default attachment size limit (10MB) to 25MB
+message_size_limit = 26214400
+
 # Additional option for filtering
 content_filter = smtp-amavis:[127.0.0.1]:10024
 


MadsRC:
Personally I don't see any reason to change the default value - Users can already do this by overriding the default value in the postfix-main.cf file?

JCBird1012:
Correct.. Users can override the value themselves, however they need to know that they should do so. I wouldn't have figured out I needed to make this change unless I was obsessively checking the logs to figure out why some of my mail messages were getting dropped. Many users might not do that. 

Google has allowed Gmail users to [send attachments of up to 25MB](https://support.google.com/mail/answer/6584?co=GENIE.Platform%3DDesktop&hl=en) by default. 
It's only logical that this Docker image follow suit in being able to accept those large attachments.

We tout this image as working "out-of-the-box" with minimal configuration required... There's no reason to stick with the Postfix default as other email providers increase their attachment size limit.

As I said previously, there's really no downside _except_ in the mailbox size limit issue I described.

tomav:
Hi @JCBird1012, and thanks for your first contribution.

As stated above, this was already possible by configuring postfix.
You mentioned that 25MB should be a default and there's no downside.
I was thinking the same the day I changed my config for that.
But this is a mistake because processing a larger file takes also a lot of time and server ressources.
I want people to understand what they do, and deep dive into configuration if default config. does not match their need, because this image offer that option.

Another thing is that email should not be used for large file transfer, there are another and better protocols or services for that.

For that reason I won't merge your PR for now. Sorry for that.

jirislav:
> Personally I don't see any reason to change the default value - Users can already do this by overriding the default value in the postfix-main.cf file?

Manual overriding doesn't take effect after you restart the docker with docker-compose .. I think it would be great to add an environment variable to the [start-mailserver.sh](https://github.com/tomav/docker-mailserver/blob/master/target/start-mailserver.sh) so that I can easily change the default maximum message size value at the `docker-compose.yml` directly if I really wish to.

johansmitsnl:
@jirislav could you create a PR for this?

jirislav:
> @jirislav could you create a PR for this?

Of course, I've just created one, see: #1056

johansmitsnl:
@jirislav just merged it! Thanks for the contribution!

JCBird1012:
Glad I could make something happen — even if it was implemented a little differently.

johansmitsnl:
@JCBird1012 the help is appreciated!

Issue #575, opened 2017-04-16
Cannot deliver emails


changkun:
Hi, thanks for your excellent project, It is really simple to deploy mailing system.

However, I have few troubles with mail delivery. The mailing system somehow cannot deliver emails, but the login is ok.

Here are the docker logs from mail container:

```bash

#
#
# ENV
#
#

HOSTNAME=mail.mydomain.com
DMS_DEBUG=1
VIRUSMAILS_DELETE_DELAY=7
ENABLE_CLAMAV=1
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
ENABLE_FAIL2BAN=1
ENABLE_SPAMASSASSIN=1
PWD=/
SHLVL=1
HOME=/root
ONE_DIR=1
SSL_TYPE=letsencrypt
DEBIAN_FRONTEND=noninteractive
_=/usr/bin/printenv

#
#
# docker-mailserver
#
#

Initializing setup
  Registering check,setup,fix,misc and start-daemons functions
  * _check_environment_variables() registered
  * _check_hostname() registered
  * _setup_default_vars() registered
  * _setup_dovecot() registered
  * _setup_dovecot_local_user() registered
  * _setup_dkim() registered
  * _setup_ssl() registered
  * _setup_docker_permit() registered
  * _setup_mailname() registered
  * _setup_amavis() registered
  * _setup_dmarc_hostname() registered
  * _setup_postfix_hostname() registered
  * _setup_dovecot_hostname() registered
  * _setup_postfix_override_configuration() registered
  * _setup_postfix_sasl_password() registered
  * _setup_security_stack() registered
  * _setup_postfix_aliases() registered
  * _setup_postfix_vhost() registered
  * _setup_environment() registered
  * _fix_var_mail_permissions() registered
  * _fix_var_amavis_permissions() registered
  * _misc_save_states() registered
  * _start_daemons_cron() registered
  * _start_daemons_rsyslog() registered
  * _start_daemons_dovecot() registered
  * _start_daemons_opendkim() registered
  * _start_daemons_opendmarc() registered
  * _start_daemons_postfix() registered
  * _start_daemons_fail2ban() registered
  * _start_daemons_clamav() registered
  * _start_daemons_amavis() registered
Checking configuration
  Check that there are no conflicts with env variables [_check_environment_variables]
  Check that hostname/domainname is provided or overidden (no default docker hostname/kubernetes) [_check_hostname]
[...]

PR #631, opened 2017-06-12
Run services with supervisor

### LINKED ISSUE #434: Move to supervisord
Proposed by @viossat, good idea.

LukeAllen:
With this change, most of the services are run via supervisor, so they'll be automatically restarted if they die. Fixes #434 and #575

Note that to run the processes with supervisor I had to use the 'run in foreground' version of their start commands, rather than the init.d script that was used before.  @tomav you should verify that I haven't lost any special configurations you had for those services. 

Also, I added tests, but I'm having an error actually running the automated test suite on my machine. The problem seems unrelated to my changes. When I run `make`, the container `mail` fails to stay up. It looks like its start-mailserver.sh is exiting with this error:

```
Check that hostname/domainname is provided or overidden (no default docker hostname/kubernetes) [_check_hostname]
  * Setting hostname/domainname is required
```

Is any special environment required to run `make`?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -61,7 +61,9 @@ RUN apt-get update -q --fix-missing && \
 # Enables Clamav
 RUN (echo "0 0,6,12,18 * * * /usr/bin/freshclam --quiet" ; crontab -l) | crontab -
 RUN chmod 644 /etc/clamav/freshclam.conf && freshclam
-RUN sed -i 's/Foreground false/Foreground true/g' /etc/clamav/clamd.conf && mkdir /var/run/clamav && chown -R clamav:root /var/run/clamav
+RUN sed -i 's/Foreground false/Foreground true/g' /etc/clamav/clamd.conf && \
+sed -i 's/AllowSupplementaryGroups false/AllowSupplementaryGroups true/g' /etc/clamav/clamd.conf && \
+mkdir /var/run/clamav && chown -R clamav:root /var/run/clamav
 
 # Configures Dovecot
 RUN sed -i -e 's/include_try \/usr\/share\/dovecot\/protocols\.d/include_try \/etc\/dovecot\/protocols\.d/g' /etc/dovecot/dovecot.conf
--- a/Dockerfile
+++ b/Dockerfile
@@ -63,6 +63,7 @@ RUN apt-get update -q --fix-missing && \
 # Enables Clamav
 RUN (echo "0 0,6,12,18 * * * /usr/bin/freshclam --quiet" ; crontab -l) | crontab -
 RUN chmod 644 /etc/clamav/freshclam.conf && freshclam
+RUN sed -i 's/Foreground false/Foreground true/g' /etc/clamav/clamd.conf && mkdir /var/run/clamav && chown -R clamav:root /var/run/clamav
 
 # Configures Dovecot
 RUN sed -i -e 's/include_try \/usr\/share\/dovecot\/protocols\.d/include_try \/etc\/dovecot\/protocols\.d/g' /etc/dovecot/dovecot.conf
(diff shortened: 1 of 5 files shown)


gegere:
@LukeAllen this is a nice step forward. Supervisor does a good job managing processes, let me know if I can help in any way.

LukeAllen:
@gegere thanks! If you could figure out what went wrong in the automated test, that would be really helpful. It looks like the TravisCI automated test is failing at a different point than on my machine, so I'm kind of confused.

tomav:
Restarted build.

tomav:
I reproduced the issue locally. Will try to have a look.

tomav:
Note that there's no more `mail.log` output in foreground (accessible using `docker logs`), which was a good thing to debug. (this is important otherwise `DMS_DEBUG` has no interest)

tomav:
Problem seems related to Dovecot. I tried multiple times and its not always the same container which stops. But I can always see the same thing in the logs:
```
Fatal: Dovecot is not running (read from /var/run/dovecot/master.pid)
```

LukeAllen:
The output of `start-mailserver.sh` goes to stdout now, so it will be visible. (I had been redirecting the output to /var/log/container-startup.log which is why it wasn't showing. Now it goes both places.)

tomav:
I guess services have to be launched in background mode, and supervisor in foreground.
`start-mailserver.sh` probably can't get `exit 0` from foreground startups and hangs.
Just my 2 cents.

LukeAllen:
Yeah people usually run supervisor in the foreground for docker containers, but it can run in the background too. I ran it in the background to stay compatible with your existing startup script. 

I replaced the old init.d calls with `supervisorctl start appname` which does immediately return `exit 0`. It does return before the app has finished starting though. I'm not sure if that breaks some assumption of the tests or the startup script. I can't figure out where the test is actually failing; is there a way to make the TravisCI tests run with DMS_DEBUG=1?

tomav:
You can add `DMS_DEBUG=1` in the `Makefile`

tomav:
Hi @LukeAllen, any update on this?

LukeAllen:
I fixed the dovecot problem. The issue was that in some configurations, `start-mailserver.sh` starts dovecot, writes some config files, then immediately calls `dovecot reload`. The `dovecot reload` broke because supervisor hadn't finished starting dovecot. I modified it to just write all config files before starting dovecot. So now all the tests run, but there are new failures.

tomav:
HI @LukeAllen, it seems that a lot of tests are not passing
https://travis-ci.org/tomav/docker-mailserver/builds/250677415?utm_source=github_status&utm_medium=notification

alinmear:
@LukeAllen do you need help getting this done? 

I think we should check whether the tasks could run in foreground mode or not. If so (notifications and errors are mapped to stdout and stderr) we should set the following to each supervisord definition:

```
stdout_logfile=/dev/stdout
stdout_logfile_maxbytes=0 
```

Benefit would be that we would have 1 interface for all logging informations.

LukeAllen:
@alinmear good idea; I incorporated your logging suggestion. Thanks for the help offer; yeah I could use some help. I'm having trouble understanding these tests. I've verified that the processes are actually all running, and the container can send mail, so my guess is that some of the tests might make assumptions that aren't true any more. Not sure though.

alinmear:
@LukeAllen will try to check this on the weekend!

evilstiefel:
Another possibility would be to use [baseimage-docker](https://github.com/phusion/baseimage-docker) by phusion which was designed specifically for this purpose (and other enhancements). Maybe they've already solved the problem and we don't need to reinvent the wheel?

I would be willing to invest some time into this. The docker image is perfect for running a mailserver, but I am unsure about dependability and resiliance against service failures...

johansmitsnl:
Hi all,

I have rebased the changes from @LukeAllen  on the new master branch and did some switching of the flow.
Currently I have almost all tests working nicely and have only 8 errors (ldap/log related). I will spend some more time on it to make these go away.

Could someone take a review of my changes on top of the @LukeAllen branch [issue-631-run-services-with-supervisor](https://github.com/johansmitsnl/docker-mailserver/commits/issue-631-run-services-with-supervisor) and give out some ideas if some changes are needed before I wrap it up?

When done I need to make a new pull request or what is the procedure of it?

johansmitsnl:
Solved all tests, see PR #676 because I could not make a PR on this PR.

LukeAllen:
Awesome thanks @johansmitsnl; I was having trouble understanding those tests. Closing this PR now since PR #676 incorporates it.

Issue #491, opened 2017-01-24
Document LDAP usage / tips in Wiki


tomav:
@alinmear or someone else?

gordanyang:
LDAP is work。
I connect openLDAP success.
* 1. dovecot-ldap.conf.ext
···
auth_bind = yes
base = ****
dn = ****
dnpass = **
hosts = **.**.**.**
ldap_version        = 3
scope = subtree
pass_attrs          = uid=cn,userPassword=password
pass_filter         = (&(objectClass=posixAccount)(mail=%u))
#user_attrs          =  uidNumber=uid,mailHomeDirectory=home,mailUidNumber=uid,mailGidNumber=gid,mailStorageDirectory=mail
user_attrs = uidNumber=uid,gidNumber=gid
user_filter         = (&(objectClass=posixAccount)(mail=%u))
···

* 2. ldap-users.cf
```
bind             = yes
bind_dn = ***
bind_pw = *
query_filter     = (&(mail=%s))
result_attribute = mail
search_base = *
server_host = *.*.*.*
version          = 3
```

* 3. ldap-aliases.cf
```
bind             = yes
bind_dn = *
bind_pw = *
query_filter     = (&(mail=%s))
result_attribute = mail
search_base = *
server_host = *.*.*.*
version          = 3
```
* 4. compose environment
```
ENABLE_SPAMASSASSIN=1
ENABLE_CLAMAV=1
ENABLE_FAIL2BAN=1
ONE_DIR=1
DMS_DEBUG=0
SSL_TYPE=manual
SSL_CERT_PATH=/tmp/docker-mailserver/ssl/mail.finogeeks.club.pem
SSL_KEY_PATH=/tmp/docker-mailserver/ssl/mail.finogeeks.club.key
ENABLE_LDAP=1
LDAP_SERVER_HOST=*.*.*.*
LDAP_SEARCH_BASE=*
LDAP_BIND_DN=*
LDAP_BIND_PW=*
POSTMASTER_ADDRESS=*
ENABLE_SASLAUTHD=1
SASLAUTHD_MECHANISMS=ldap
SASLAUTHD_LDAP_SERVER=*.*.*.*
SASLAUTHD_LDAP_SSL=0
SASLAUTHD_LDAP_BIND_DN=*
SASLAUTHD_LDAP_PASSWORD=*
SASLAUTHD_LDAP_SEARCH_BASE=*
SASLAUTHD_LDAP_FILTER=(&(mail=%u))
```

[...]

PR #636, opened 2017-06-23
Also search on the mail so that the email can be dropped

johansmitsnl:
See issue #491


johansmitsnl:
Any idea why this might fail and what the error is?

tomav:
@johansmitsnl this is still relevant after @alinmear improvements? Thanks.

johansmitsnl:
Need to check it. Hope to do it this weekend.

tomav:
Thank you

johansmitsnl:
Not needed anymore, it is fixed with: #644

Issue #591, opened 2017-05-05
LDAP environment variables and ambiguous characters


tailtwo:
Hello,

The user I use to bind to my LDAP server have a '&' character in its password. When this character is parsed, the string that correspond with the password in ldap-users.cf gets cut in two with `bind_pw          = admin`

Let's say my password is ABC&DEF:

docker-compose.yml
```bash
[...]
environment:
  - LDAP_BIND_PW=ABC&DEF
[...]
```

/etc/postfix/ldap-users.cf
```bash
[...]
bind_pw = ABCbind_pw          = adminDEF  
[...]
```

Even with quotes and double quotes, it stays the same.
Is there any way I can fix this other than entering it manually?

alinmear:
Hello @tailtwo, 

to fix this we should have a look at the ldap method, especially the parsing of the env variables. Im pretty sure, that here this is happening.

Do you have time to investigate this?

alinmear:
@tailtwo my pr #644 should fix the issues with the `&` within the ldap configurations. 

hopefully this helps!

johansmitsnl:
It worked for me.

alinmear:
Seems that there are no problems after #644. Closing ...

Issue #641, opened 2017-06-30
LDAP accounts/domain/alias  fully functional


kopax:
I have a old mail-server (v1) with my own ldap implementation.

I have tried to play with the v2 and the following `docker-compose.yml`

```yml
version: '2'

services:
  mail:
    image: tvial/docker-mailserver:2.1
    hostname: mail
    domainname: domain.com
    container_name: mail
    ports:
    - "25:25"
    - "143:143"
    - "587:587"
    - "993:993"
    volumes:
    - maildata:/var/mail
    - mailstate:/var/mail-state
    - ./config/:/tmp/docker-mailserver/
    - /home/me/workspace/private/domain-keys/mail-server:/tmp/ssl:ro
    environment:
    - ENABLE_LDAP=1
    - LDAP_SERVER_HOST=ldap.domain.com
    - LDAP_SEARCH_BASE=dc=mail,dc=domain,dc=com
    - LDAP_BIND_DN=cn=readonly,dc=domain,dc=com
    - LDAP_QUERY_FILTER=(&(objectClass=CourierMailAccount)(mail=%s))
    - LDAP_BIND_PW=abc
    - ENABLE_SPAMASSASSIN=1
    - ENABLE_CLAMAV=1
    - ENABLE_FAIL2BAN=1
    - ENABLE_SASLAUTHD=1
    - SASLAUTHD_MECHANISMS=ldap
    - SASLAUTHD_LDAP_SERVER=ldap.domain.com
    - SASLAUTHD_LDAP_PROTO=ldap://
    - SASLAUTHD_LDAP_SSL=0
    - SASLAUTHD_LDAP_PASSWORD=domain
    - SASLAUTHD_LDAP_BIND_DN=cn=readonly,dc=domain,dc=com
    - SASLAUTHD_LDAP_PASSWORD=abc
    - SASLAUTHD_LDAP_SEARCH_BASE=dc=mail,dc=domain,dc=com
    - SASLAUTHD_LDAP_FILTER=(&(objectClass=CourierMailAccount)(mail=%s))
    - SSL_TYPE=manual
    - SSL_CERT_PATH=/tmp/ssl/mail.domain.com-full.pem
    - SSL_KEY_PATH=/tmp/ssl/mail.domain.com-key.pem
    - ONE_DIR=1
    - DMS_DEBUG=0
    cap_add:
    - NET_ADMIN
    restart: always

volumes:
  maildata:
    driver: local
  mailstate:
    driver: local
```

[...]

Issue #633, opened 2017-06-17
Unable to change LDAP filter for non SASL.


Joshua-Ashton:
Hello!
I am unable to change the LDAP filter to be other than "(&(?objectClass=PostfixBookMailAccount)(uniqueIdentifier=%u))".

No matter what options I put in my DockerCompose.yml I always end up with that filter being executed on my ldap. I do not have a postfixbookmailaccount and I would prefer it if I didn't have to, unless it is necessary.

I have tried 
  - SASLAUTHD_LDAP_FILTER='(&(objectClass=posixAccount)(uid=%U))'
  - LDAP_FILTER='(&(objectClass=posixAccount)(uid=%U))'
  - LDAP_QUERY_FILTER='(&(objectClass=posixAccount)(uid=%U))'
but none of them have any effect.

I am using the docker openldap also, so adding the schema would prove a bit of a mess, unless there's another way to do it?

My apologies if I am being stupid - I am fairly new to Docker.

Thanks and any advice/help is greatly appreciated.
- Josh

alinmear:
Hello @Joshua-Ashton. This is a known Bug. Please have a look at https://github.com/tomav/docker-mailserver/issues/491#issuecomment-306296358 for a workaround/solution.

alinmear:
I will fix this today and do a pull request!

alinmear:
@Joshua-Ashton please try with the latest version of this container. My PR #644 has been merged and should fix this.

Note that your issue should be related to the missing DOVECOT ldap part (fixed with the PR stated above).

I placed a sample config within the README. Maybe you can take a look!

Keep in Mind to use the following variables for configuration:

* `LDAP_QUERY_FILTER_USER`
* `LDAP_QUERY_FILTER_GROUP`
* `LDAP_QUERY_FILTER_ALIAS`
* `SASLAUTHD_LDAP_FILTER`
* `DOVECOT_USER_FILTER`
* `DOVECOT_PASS_FILTER`

alinmear:
@Joshua-Ashton ; can we close this issue. Also special character should work now. Merged #648.

Joshua-Ashton:
Hello! 
Sorry for taking an insane amount of time to get back to you, I'm pretty blind.
Decided to give this a go today but I'm still having issues although it's probably my fault - I'm fairly new to this. I just want to use regular "posixAccount"s as people with mail and no matter what I put as the filter it doesn't like it. Maybe I'm missing something...

Thanks for your help!
[...]

PR #642, opened 2017-06-30
Also search on the domain name for the virtual domain name in the ldap

johansmitsnl:
See #638

PR #644, opened 2017-07-03
Fix ldap related critical Problems

alinmear:
As already written by me (https://github.com/tomav/docker-mailserver/issues/641#issuecomment-312585289) we have some serious issues regarding the setup of ldap with env variables. 

This pr should fix the following issues: #641, #633, #491 (see comments), #636 (can be configured explicit with the new env vars; maybe we should discuss better defaults as stated there), #591 

* `LDAP_QUERY_FILTER_USER`
* `LDAP_QUERY_FILTER_GROUP`
* `LDAP_QUERY_FILTER_ALIAS`
* `DOVECOT_USER_FILTER`
* `DOVECOT_PASS_FILTER`

We may should discuss some defaults and or combinations to make the setup more straight forward (like dovecot user_filter and pass_filter could be the same, when only 1 has been defined or LDAP_QUERY_FILTER_USER could be set for the saslauthd part if not explicitly set).

Hope this helps ...


tomav:
Thanks @alinmear

Joshua-Ashton:
Thanks for this, I will try it out if I get a chance later this week

tailtwo:
Thank you, I really didn't have the time to check the source to fix #591 so this mr will come in handy.

johansmitsnl:
The new variable don't work because they are in comment. Why is this?
Now the dovecot variable don't work, so did not build a new image to test.

https://github.com/tomav/docker-mailserver/blob/21fb3f3c86391f079afb7fb59260b64019fa2e51/target/start-mailserver.sh#L593:L595

alinmear:
@johansmitsnl, these are only default values. If you want to change those values you need to set the corresponding ENV Variables, like:

```
- DOVECOT_PASS_FILTER="(&(objectClass=PostfixBookMailAccount)(uniqueIdentifier=%n))"
- DOVECOT_USER_FILTER="(&(objectClass=PostfixBookMailAccount)(uniqueIdentifier=%n))"
```

I already wrote an example within the readme.

johansmitsnl:
Inside the docker:
```
echo $DOVECOT_PASS_FILTER
(&(objectClass=PostfixBookMailAccount)(|(uniqueIdentifier=%n)(mail=%u)))
```

```
cat /etc/dovecot/dovecot-ldap.conf.ext
pass_attrs          = uniqueIdentifier=user,userPassword=password
pass_filter         = (&(objectClass=PostfixBookMailAccount)(uniqueIdentifier=%n))
user_attrs          = mailHomeDirectory=home,mailUidNumber=uid,mailGidNumber=gid,mailStorageDirectory=mail
user_filter         = (&(objectClass=PostfixBookMailAccount)(uniqueIdentifier=%n))
```

It seems that the variables are not applied but they are set in the ENV.

alinmear:
@johansmitsnl , the problem here is the `|` within your expression; because we take the `|` for the separation of the sed parts; this is causing the problem; you could escape the `|` like:

```bash
(&(objectClass=PostfixBookMailAccount)(\|(uniqueIdentifier=%n)(mail=%u)))
```

This should work. I check how to deal with this kind of characters and sed

johansmitsnl:
@alinmear Escaping worked and solved my ldap issue #636.
It should only escape the special chars as you already noted.

Now we only have the #642 to rework to support virtual domains in the ldap. Will take a look how the changes where made and update the PR.

PR #654, opened 2017-07-11
Change Email-Folder to /var/mail/%d/%u/mails

evilstiefel:
This solves the issue that when running managesieve, the .sieve-folder shows up as subscribable for IMAP clients but will produce an error when selecting. Mainly to remove clutter and to follow dovecot recommendations on not setting the same directory for user home and user mail.

This should not be merged as of yet, because it will break existing mail-stores. Hopefully though, this pull request will let me automatically run the travis-suite to check if the changes are safe for new installations.

Later, if the request is desirable, we could add a config flag like the many other's that already exist,

### DIFF
--- a/target/dovecot/10-mail.conf
+++ b/target/dovecot/10-mail.conf
@@ -27,7 +27,7 @@
 #
 # <doc/wiki/MailLocation.txt>
 #
-mail_location = maildir:/var/mail/%d/%n
+mail_location = maildir:/var/mail/%d/%n/mails
 
 # If you need to set multiple mailbox locations or want to change default
 # namespace settings, you can do it by defining namespace sections.
--- a/target/start-mailserver.sh
+++ b/target/start-mailserver.sh
@@ -534,15 +534,16 @@ function _setup_dovecot_local_user() {
 			# user:password:uid:gid:(gecos):home:(shell):extra_fields
 			# Example :
 			# ${login}:${pass}:5000:5000::/var/mail/${domain}/${user}::userdb_mail=maildir:/var/mail/${domain}/${user}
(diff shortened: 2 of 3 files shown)


evilstiefel:
I noticed that the bats test references the home directory directly, so I need to update these tests (next to preparing a configurable flag).

Some background on why the home directory should be different to the mail directory for virtual users:
https://wiki2.dovecot.org/VirtualUsers/Home

johansmitsnl:
@evilstiefel why use a configurable flag? A migration script in the startup script would also work and it will make a permanent change to the structure.
Not sure what the admins think of this idea though.

Fmstrat:
I think this could be handled in the `user_attrs` option in the dovecot configuration. This is a request in my LDAP issue, too: https://github.com/tomav/docker-mailserver/issues/680

johansmitsnl:
@evilstiefel any progress on this PR?

evilstiefel:
The changes to the tests and files are minimal and complete, as far as I'm concerned. I've tested this extensively and used it on my own setup for quite some time. I cannot debug the LDAP failure in the travis environment, since I'm not knowledgeable in that area. It seems to me that the user folder is also specified in the directory and I am unsure how to change the file.

As to the issue how best to implement this if the PR is desirable, either via a configurable flag or just migrating existing folders in the case of anybody upgrading from an image with the "old" folder-structure, that's up to the maintainers (I see that as a "political" decision).

Since then, I've switched to another setup to be honest, since I didn't feel there was much feedback or activity either here or on the gitter chat.

johansmitsnl:
Clear. I'll patch up your pr.
Will close yours and open a new one from my repository that has LDAP support. I did already work on the LDAP part.
Tomav doesn't have much time at the moment. Gained access to the repo and have some spare time to spend on it. 
Suggestions are always welcome. Thanks for you effort on testing and his pr.

PR #657, opened 2017-07-18
Proper shutdown on container stop, additional whitelisting options, PCRE filtering support, fixing permission issues

felix-schwarz:
New features:
* ADD_PERMIT_NETWORK options allows to specify additional networks to whitelist in mynetworks while retaining the benefits of automatically determining relevant networks through PERMIT_DOCKER. E.g. to add the network "10.0.1.0/24", set ADD_PERMIT_NETWORK to "10.0.1.0/24"
* DISABLE_FILTERS_FOR_MYNETWORKS option to add amavisd-new configuration to disable clamav/spamassassin filters for the IPv4 hosts listed in mynetworks (amavisd-new doesn't seem to support IPv6 here). Set DISABLE_FILTERS_FOR_MYNETWORKS to enable this.
* Added PCRE support to postfix, allowing efficient rejections of mail headers and body based on PCRE regexps. The regexps can be placed in config/postfix-reject_body_checks and config/reject_header_checks accordingly. See postfix main.cf documentation on header_checks and body_checks for more information.
* Added support for the second postgrey whitelist (whitelist_clients.local)

Improvements:
* Postgrey whitelists are also accepted as config/postgrey-whitelist_clients.local and config/postgrey-whitelist_recipients.local
* Stray pid files that could hinder a proper startup of postgrey and dovecot are now removed on container start
* When the container is stopped, all services are now shut down properly
   - previously, the services were just killed, resulting in stray pid files and - possibly - corruption of files modified by the services at the time of termination
   - stopping the container is now much faster (previously took 10 seconds - the standard grace period for containers from being signalled by 'docker stop' and timing out and being killed off)

Bug/Security fixes:
* Fixing permissions problems for /var/mail-state/spool-postfix when ONE_DIR=1:
  - spool-postfix/maildrop and spool-postfix/public folders must be owned by group postdrop
  - spool-postfix/[dev/etc/lib/pid/usr] permissions were not restrictive enough. They should be owned by the root user and root group.


Dockerfile
```diff
@@ -158,13 +159,13 @@ RUN sed -i -r "/^#?compress/c\compress\ncopytruncate" /etc/logrotate.conf && \
...
+
 # Start-mailserver script
 COPY ./target/start-mailserver.sh ./target/docker-configomat/configomat.sh /usr/local/bin/
 RUN chmod +x /usr/local/bin/*
 
 EXPOSE 25 587 143 993 110 995 4190
 
-CMD /usr/local/bin/start-mailserver.sh
-
+CMD ["/bin/bash", "/usr/local/bin/start-mailserver.sh"]
```
johansmitsnl: You test 123 is failing because this change.
You need to set the same command to the setup.sh https://github.com/tomav/docker-mailserver/blob/master/setup.sh#L10

johansmitsnl:
Please take a look at my branch [issue-631-run-services-with-supervisor](https://github.com/johansmitsnl/docker-mailserver/commits/issue-631-run-services-with-supervisor). It has many overlap with a proper shutdown function that you are building because of the use of supervisor.
Can you split the PR for the PCRE and whitelisting?
The proper shutdown will been take care of in PR #631

tomav:
Resolved conflicts. Building in Travis.

johansmitsnl:
Could you rework the PR on the current master?

felix-schwarz:
Sorry for the late reaction. I currently lack the time to rework it, but hope to find some in the coming weeks.

johansmitsnl:
Don't want to rush you but is there any progress?

felix-schwarz:
Seeing as I still lack the time to rework it, I close this pull request for now. If I find the time to rework a current version, I'll open a new one.

PR #664, opened 2017-07-23
using travis_retry for image building

danielpanteleit:
This could fix the problem that clamav is not able to update in docker build


danielpanteleit:
Somehow the tests seem to be flaky. This is not an issue with the change.

tomav:
Restarted build

PR #665, opened 2017-07-23
Dockerfile cleanup

danielpanteleit:
Removed the initial duplicate apt-get update / apt-get upgrade and merged a couple of RUN commands together. This should result in less layers and a smaller image because the apt cache is cleared properly now (the first apt-get update / upgrade stored the caches/lists into a layer).


tomav:
Thanks you @danielpanteleit

PR #667, opened 2017-07-23
Disable logrotate and cron job for updating clamav if clamav is disabled (fixes #660)

### LINKED ISSUE #660: Should the freshclam cron job be disabled when ENABLE_CLAMAV=0?
I'm getting several emails a day about cron job failing to rotate logs and things.  Also, trying to minimize network, memory, and cpu usage by not updating clamav db.

danielpanteleit:
This should fix #660.

Cron-Job is now written to `/etc/cron.d/freshclam`. On start, the cron job and logrotate files regarding clamav are removed if clamav is disabled. This means that this can not be reversed, but changing the environment should also spawn a new container with the files intact (in case someone wants to reenable clamav)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -65,8 +65,7 @@ RUN apt-get update -q --fix-missing && \
   touch /var/log/auth.log && \
   update-locale
 
-# Enables Clamav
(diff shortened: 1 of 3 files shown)


tomav:
Great!

PR #669, opened 2017-07-23
remove /etc/cron.weekly/fstrim since we don't want to fstrim in docker

### LINKED ISSUE #662: Cron runs fstrim
Cron runs `fstrim` on a weekly basis which results in a cron error mail:

```
/etc/cron.weekly/fstrim:
fstrim: /tmp/ssl/private.pem: not a directory
fstrim: /tmp/ssl/public.pem: not a directory
fstrim: /etc/hosts: not a directory
fstrim: /etc/hostname: not a directory
fstrim: /etc/resolv.conf: not a directory
fstrim: /var/mail: FITRIM ioctl failed: Operation not permitted
fstrim: /tmp/docker-mailserver: FITRIM ioctl failed: Operation not permitted
fstrim: /var/mail-state: FITRIM ioctl failed: Operation not permitted
fstrim: /: FITRIM ioctl failed: Operation not permitted
```

Since docker does not have permission to `fstrim` (which should already be done by the host system), `fstrim` should be disabled.

I guess deleting `/etc/cron.weekly/fstrim` should be enough.

danielpanteleit:
This fixes #662

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -63,7 +63,8 @@ RUN apt-get update -q --fix-missing && \
   rm -rf /usr/share/man/* && \
   rm -rf /usr/share/doc/* && \
   touch /var/log/auth.log && \
-  update-locale
+  update-locale && \
+  rm -f /etc/cron.weekly/fstrim
 
 # Enables Clamav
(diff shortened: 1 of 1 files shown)


tomav:
Thanks!

PR #685, opened 2017-08-09
Also test on the submission port if the login is succesful

johansmitsnl:
When making a new PR I noticed that there was no test for smtp authentication on the submission port.
While adding it opendkim is complaining about the permission in the KeyTable location. It is set to /tmp/... but this is wrong.
This PR fixes the path and adds a test for smtp authentication on the submission port.


johansmitsnl:
@tomav my local tests all pass, any idea?

johansmitsnl:
@alinmear @tomav could you trigger a rebuild?

tomav:
Done.

tomav:
Thanks.

PR #687, opened 2017-08-11
Added post setup script functionality.

Fmstrat:
This will enable users to execute postconf or sed commands (or other things like adding mysql for vimbadmin use) without having to fundamentally change anything about the docker container. It simply adds the ability to execute `/config/postsetup.sh` after all other setup is complete. It could help people with issues like:

https://github.com/tomav/docker-mailserver/issues/680
https://github.com/tomav/docker-mailserver/issues/679
https://github.com/tomav/docker-mailserver/issues/670

I'm sure the list goes on and on, but those are just the top 3 from the issues list that a simple postconf command could handle.


Fmstrat:
This is showing a failed check above, but I think someone else is working on LDAP and that's what's breaking, vs this pull request.

Fmstrat:
@tomav Any idea if these checks were failing because of something else on the LDAP side? Unfortunately I can't trigger the tests to run again without resubmitting the pull. Thanks!

tomav:
Restarted build to see if it works better.

Fmstrat:
Looks good, working in my test cases, too. Can we get a merge? Thanks!

tomav:
@Fmstrat I did not lokked at the code before sorry, but could you add an integration test?

Fmstrat:
@tomav Integration test is in and all tests pass. Thanks!

tomav:
Why `postfix-mysql` should be added to the project?

Fmstrat:
Oh, that was supposed to be a separate pull, sorry about that. It's so that we can integrate virtual users in a mariaDB container such as using a tool like https://www.vimbadmin.net/ . (which works like a champ with the addition)

tomav:
@johansmitsnl what do you think about that?

Fmstrat:
I'm not sure what's being asked here. Are you saying you don't want to merge the pre/post script, and that I should roll this PR back, then create a new one for inclusion of `postfix-mysql` with a test case for that?

Issue #686, opened 2017-08-10
After host reboot (and some container restarts) dovecot fails to run correctly


justinvoelker:
I have docker-mailserver setup and running, sending and receiving emails as expected, but sometimes after a server reboot, or a docker-compose restart, dovecot appears to break. The docker log gets flooded with the following lines every second until I stop or restart the container (hopefully working on the second try).

```
INFO spawned: 'dovecot' with pid 1099
INFO success: dovecot entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
INFO exited: dovecot (exit status 89; not expected)
INFO spawned: 'dovecot' with pid 1101
INFO success: dovecot entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
INFO exited: dovecot (exit status 89; not expected)
INFO spawned: 'dovecot' with pid 1103
INFO success: dovecot entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
INFO exited: dovecot (exit status 89; not expected)
```

I also see the following line in the log at sporadic places.

```
mail dovecot: master: Fatal: Dovecot is already running with PID 203 (read from /var/run/dovecot/master.pid)
```

The strange thing is that even though it says dovecot is running, I cannot connect any accounts over IMAP.

Here is my docker-compose.yml

```
version: '2'
services:
  mailserver:
    image: tvial/docker-mailserver:latest
    restart: always
    hostname: mail
    domainname: example.com
    container_name: mail.example.com-mailserver
    ports:
      - "25:25"
      - "587:587"
      - "993:993"
    volumes:
      - ./mailserver/maildata/:/var/mail
      - ./mailserver/mailstate/:/var/mail-state
      - ./mailserver/config/:/tmp/docker-mailserver/
      - /home/justin/docker/nginx-proxy/certificates/:/etc/letsencrypt/live
    environment:
      - ONE_DIR=1
      - SSL_TYPE=letsencrypt
      - DMS_DEBUG=1
```

gutmensch:
I'm seeing the same respawning behavior for amavis
`2017-08-11 19:06:55,127 INFO exited: amavis (exit status 255; not expected)
2017-08-11 19:06:56,130 INFO spawned: 'amavis' with pid 25600
2017-08-11 19:06:56,681 INFO success: amavis entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
2017-08-11 19:06:56,689 INFO exited: amavis (exit status 255; not expected)
[...]

PR #699, opened 2017-08-18
Fix the hook that restarts the amavis server and causes it to detach

johansmitsnl:
See #686
This should fix the amavis issue.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -115,8 +115,9 @@ RUN sed -i -e 's/include_try \/usr\/share\/dovecot\/protocols\.d/include_try \/e
 COPY target/dovecot/dovecot-ldap.conf.ext /etc/dovecot
 COPY target/postfix/ldap-users.cf target/postfix/ldap-groups.cf target/postfix/ldap-aliases.cf /etc/postfix/
 
-# Enables Spamassassin CRON updates
-RUN sed -i -r 's/^(CRON)=0/\1=1/g' /etc/default/spamassassin
+# Enables Spamassassin CRON updates and update hook for supervisor
+RUN sed -i -r 's/^(CRON)=0/\1=1/g' /etc/default/spamassassin && \
+    sed -i -r 's/^\$INIT restart/supervisorctl restart amavis/g' /etc/spamassassin/sa-update-hooks.d/amavisd-new
 
 # Enables Postgrey
 COPY target/postgrey/postgrey /etc/default/postgrey


johansmitsnl:
@tomav a rebuild for this to please

Dockerfile
```diff
@@ -115,8 +115,9 @@ RUN sed -i -e 's/include_try \/usr\/share\/dovecot\/protocols\.d/include_try \/e
 COPY target/dovecot/dovecot-ldap.conf.ext /etc/dovecot
 COPY target/postfix/ldap-users.cf target/postfix/ldap-groups.cf target/postfix/ldap-aliases.cf /etc/postfix/
 
-# Enables Spamassassin CRON updates
-RUN sed -i -r 's/^(CRON)=0/\1=1/g' /etc/default/spamassassin
+# Enables Spamassassin CRON updates and update hook for supervisor
+RUN sed -i -r 's/^(CRON)=0/\1=1/g' /etc/default/spamassassin && \
+    sed -i -r 's/^\$INIT restart/\supervisorctl restart amavis/g' /etc/spamassassin/sa-update-hooks.d/amavisd-new
```
gutmensch: Should there really be a backslash in front of s?! No need to escape anything?

gutmensch:
Thanks for looking into this and fixing it, @johansmitsnl!

Dockerfile
johansmitsnl: Your correct, fixed it

johansmitsnl:
the failures are not related to the chance. They fail every now and then..

tomav:
Merged

init-js:
I've noticed that the process hierarchy of  amavis is different when it is started with init.d (`start-stop-daemon --start --quiet --pidfile /var/run/amavis/amavisd.pid --startas /usr/sbin/amavisd-new --`) versus when it is started with supervisorctl (`/usr/sbin/amavisd-new foreground`). I don't know amavis enough to know if this has subtle implications.

```
root@mx1:/# /etc/init.d/amavis stop; supervisorctl stop amavis; : make it is not running
Stopping amavisd: (not running).
amavis: ERROR (not running)
root@mx1:/# /etc/init.d/amavis start; sleep 5;
Starting amavisd: amavisd-new.
root@mx1:/# ps -eaf | grep amavisd
amavis   27567     0  4 22:28 ?        00:00:00 /usr/sbin/amavisd-new (master)
amavis   27573 27567  0 22:28 ?        00:00:00 /usr/sbin/amavisd-new (virgin child)
root     27587 26926  0 22:28 ?        00:00:00 grep --color=auto amavisd
root@mx1:/# /etc/init.d/amavis stop
Stopping amavisd: amavisd-new.
root@mx1:/# supervisorctl start amavis; sleep 5;
amavis: started
root@mx1:/# ps -eaf | grep amavisd
amavis   27696     5 11 22:30 ?        00:00:01 /usr/sbin/amavisd-new (master)
root     27716 26926  0 22:30 ?        00:00:00 grep --color=auto amavisd

```
In either case, it looks like it's the `master` process which is listening on the unix socket (`127.0.0.1:10024`)

Issue #697, opened 2017-08-16
Sudden amavis crashed


manuelgu:
Startup log: https://paste.md-5.net/koxixuruqo.coffee

After a few hours of uptime, `amavis` ends up in a never-ending restart-crash loop like shown [here](https://paste.md-5.net/wageceroti.vhdl)

danielpanteleit:
I can confirm this:

```
Aug 17 16:56:00 mail amavis[20575]: (!)Net::Server: 2017/08/17-16:56:00 Pid_file already exists for running process (259)... aborting\n\n  at line 144 in file /usr/share/perl5/Net/Server.pm
2017-08-17 16:56:00,104 INFO exited: amavis (exit status 1; not expected)
2017-08-17 16:56:00,125 INFO spawned: 'amavis' with pid 20589
2017-08-17 16:56:00,128 INFO success: amavis entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
```


This also happens to dovecot for me:

```
2017-08-17 16:55:59,701 INFO spawned: 'dovecot' with pid 20583
2017-08-17 16:55:59,731 INFO success: dovecot entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
2017-08-17 16:55:59,732 INFO exited: dovecot (exit status 89; not expected)
Aug 17 16:55:59 mail dovecot: master: Fatal: Dovecot is already running with PID 188 (read from /var/run/dovecot/master.pid)
```

Both seem to use their own pid files. Both mentioned pids are not running.

danielpanteleit:
I think these are two separate problems:

amavis seems to quit on its own at one point and probably restarts itself somehow. Supervisor also starts a new amavis instance which will be killed and restarted all the time:

```
2017-08-14 07:09:28,320 INFO exited: amavis (exit status 0; expected)
2017-08-14 07:09:29,489 INFO spawned: 'amavis' with pid 18990
Aug 14 07:09:29 mail amavis[18981]: starting. /usr/sbin/amavisd-new at ... amavisd-new-2.10.1 (20141025), Unicode aware, LC_ALL="C"
2017-08-14 07:09:29,937 INFO success: amavis entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
Aug 14 07:09:30 mail amavis[18990]: starting. /usr/sbin/amavisd-new at ... amavisd-new-2.10.1 (20141025), Unicode aware
...
[...]

PR #700, opened 2017-08-19
Clean any leftover pid file from a stop

johansmitsnl:
During a stop and start pid files can be left over with wrong information.
Clean them before starting new services.

See #697


johansmitsnl:
@tomav as you can see the 40 seconds is still to low. This build failed 3 times and no change to the code was made. Only when I upped the waiting time it works more stable.
I see other builds as well fail on most of the times the same points, I assume that travis is busy or not as fast as we expect.

tomav:
Ok for me.

Issue #681, opened 2017-08-06
Hide email client details from mail headers


johansmitsnl:
In the email headers there is some sensitive information. Where the connection originated from, which client is used.

Can this be stripped so that the received does not gain information about you and keep it down to only the essential parts?

Example:
```log
Return-Path: <johan@domain.example>
Received-SPF: pass (domain.example: 35.157.0.0 is authorized to use 'johan@domain.example' in 'mfrom' identity (mechanism 'mx' matched)) receiver=mail.example; identity=mailfrom; envelope-from="johan@domain.example"; helo=mail.domain.example; client-ip=35.157.0.0
Received: from mail.domain.example (mail.domain.example [35.157.0.0])
  by mail.domain.example (Postfix) with ESMTPS id 062112E097F
  for <someone@domain.example>; Sun,  6 Aug 2017 12:19:30 +0300 (MSK)
Received: from localhost (localhost [127.0.0.1])
  by mail.domain.example (Postfix) with ESMTP id 7A80518003D;
  Sun,  6 Aug 2017 09:19:28 +0000 (UTC)
X-Virus-Scanned: Debian amavisd-new at domain.example
Received: from mail.domain.example ([127.0.0.1])
  by localhost (mail.domain.example [127.0.0.1]) (amavisd-new, port 10024)
  with ESMTP id ixpQhbfk0t-s; Sun,  6 Aug 2017 09:19:27 +0000 (UTC)
Received: from iMac.local (unknown [88.159.0.0])
  (Authenticated sender: johan.smits@mail.domain.example)
  by mail.domain.example (Postfix) with ESMTPSA id 5CE1D18003A;
  Sun,  6 Aug 2017 09:19:27 +0000 (UTC)
DKIM-Signature: v=1; a=rsa-sha256; c=relaxed/simple; d=domain.example;
  s=mail; t=1502011167;
To: someone@domain.example
From: Johan Smits <johan@domain.example>
Subject: Abuse
Message-ID: <c2f2a6dc-403c-1eb6-5225-7cc4a5121a2b@domain.example>
Disposition-Notification-To: Johan Smits <johan@domain.example>
Date: Sun, 6 Aug 2017 11:19:24 +0200
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.12; rv:52.0)
 Gecko/20100101 Thunderbird/52.2.1
```

Should remove these items:
* User-Agent
* Received: from iMac.local (unknown [88.159.0.0])
[...]

PR #708, opened 2017-09-05
Improve the privacy of the client by removing sensitive details

johansmitsnl:
We should care about our privacy and hide information when we send emails.
See ticket #681 

The PR has tests to ensure the filter is working. Multiple items are added that I think should be filtered. The minimal compatibility level had to be set to 2 in order for pcre to work and not creating errors in the logging.


johansmitsnl:
@tomav could you do a review?

tomav:
Sorry, could you resolve the conflict?

Issue #618, opened 2017-05-29
send emails over SSL port 465


gustavotero7:
Can I send emails using the 465 port? I al ready have ssl certificate (comodo) also I have specified the ssl certificates manually in the docker compose file, now i can send emails using the port 587/25 with tls encryption but the ssl /465 port dont works.

Also for email clients I only able to use tls, how i can use ssl porperly?

what can i do?

arneke:
I believe you cannot do it until #595 lands, or you have to build your own image (which is not that hard) to overwrite master.cf.

You need to add something like
```
smtps     inet  n       -       n       -       -       smtpd
  -o smtpd_tls_wrappermode=yes
  -o smtpd_sasl_auth_enable=yes
```

MadsRC:
The only way to do it right now is to monkey-patch the master.cf file after container creation.

As @arneke mentioned, #595 adds support for this, however it won't get merged until I'm done making unit tests for it (However, I've run my private mailserver with the #595 patch for a month now, without problems, so it should work fine)

0xQSL:
I've tried using @MadsRC's #595 to enable tls on port 465.
Using the configuration posted above I get this warning:
```
postconf: warning: unmatched request: "smtps     inet  n       -       n       -       -       smtpd"
```
Is this because you're parsing the script line by line?

johansmitsnl:
@gustavotero7 @0xQSL I just created a PR #709 for this feature.
Waiting for @tomav to review and comment on it.

tomav:
Merged, image building.

PR #709, opened 2017-09-06
Listen also on the ssl port 465

johansmitsnl:
Also use the SSL port for clients to deliver emails.

Added tests to verify tls strength and authentication.

See #618


johansmitsnl:
@tomav could you review it?

tomav:
LGTM, thanks

gustavotero7:
@johansmitsnl That's great thanks.

PR #724, opened 2017-09-26
SSL Bundle

Kabe0:
Added in the functionality for bundle files for SSL.
To use this feature, add the following to the wideside variables SSL_BUNDLE_PATH
This will work identically to SSL_KEY_PATH and SSL_CERT_PATH. 
The field is not required and can be ignored if not required.


johansmitsnl:
Issue got solved by using the correct crt with the intermediate cert in it.

PR #731, opened 2017-10-02
Enable variable polling rates for fetchmail

dmcgrandle:
* Setup two base fetchmailrc files, one with default polling of 300 and
  one with none (target/fetchmail/fetchmailrc_general and _nopoll).
* Modify Dockerfile to copy both these files into container.
* Modify setup-fetchmail to parse user-provided fetchmail.cf and if
  user gave a poll interval then use _nopoll base file, else _general.
* Modify supervisor-app.conf to remove hard-code of 300 second poll
  interval.


target/supervisor/conf.d/supervisor-app.conf
```diff
@@ -104,7 +104,7 @@ autorestart=true
 stdout_logfile=/var/log/supervisor/%(program_name)s.log
 stderr_logfile=/var/log/supervisor/%(program_name)s.log
 user=fetchmail
-command=/usr/bin/fetchmail -f /etc/fetchmailrc -v --nodetach --daemon 300 -i /var/lib/fetchmail/.fetchmail-UIDL-cache --pidfile /var/run/fetchmail/fetchmail.pid
```
johansmitsnl: What does it do when you remove all your changes, just change the deamon time to 60 (or so) and use the nopoll configuration?
If it polls every 60 seconds the fix is more simple then I thought. I don't have a setup like this so could you verify this?

dmcgrandle: I can test this later tomorrow, but I’m quite sure it will poll at whatever value it is set to.  However, the point of this change wasn’t to change the default from 300 seconds to 60 - that is easy.  The point of this change is to allow the user to set it to whatever value they want in their config/fetchmail.cf file.  To make it user-definable. The change you are suggesting simply makes a new default of 60 without allowing the user to change it, unless I am misunderstanding something?

johansmitsnl: If his works it can be easily changed with just a single env variable and there is no need for a bigger change.

dmcgrandle: Oh, I think I see what you are getting at now.  Rather than mess with the existing files or code, we could simply set an environmental variable such as FETCHMAIL_POLL to the desired poll rate and in the supervisor conf file we could then use "--daemon %(ENV_FETCHMAIL_POLL)" instead of "--daemon 300", and set the default to 300 in case the user doesn't define it.  Interesting.  I will try this out and let you know - yes, it would be a much simpler change.  :)

dmcgrandle:
Johan - this is passing all tests and completing successfully on my local machine.  Any idea why this failed the Travis CI build?

dmcgrandle:
I pushed an empty commit to trigger another Travis CI build and this time it passed.  Strange.

target/supervisor/conf.d/supervisor-app.conf
johansmitsnl: You got it ;)

Issue #552, opened 2017-03-18
[documentation] Reload/restart after accounts/virtual change?


strarsis:
Because no postmap files are used - is it required to reload/restart the server daemons after 
changing the 'postfix-account.cf' and/or 'postfix-virtual.cf' file?
Do both daemons (postfix + dovecot) need a restart or can a softer reload be used after config change? 

Config files are changed without setup.sh (synced/uploaded), hence no setup.sh is involved 
(that may perform additional action to ensure the new config takes effect).

voldedore:
I didn't take much time to view the setup.sh

But I don't think that the server daemons will know to update user list when you just change the postfix-account.cf.

As for me, I used setup.sh script to add new user and had to restart container to take effect.

strarsis:
Correct, according to documentation, process will not detect changes to texthash type: 
http://www.postfix.org/DATABASE_README.html#types

Postfix has to be reloaded for using the new config: 
http://www.postfix.org/DATABASE_README.html#detect

postfix reload takes away performance each time - it is recommended in documentation to use a config type that doesn't require reloading - like postmap.

tomav:
https://github.com/tomav/docker-mailserver/pull/382

This can be changed (again). We can talk about that here.

tyranron:
@strarsis why not just restart whole container on config change? Docker containers are rarely designed in a way to reload stuff inside it. I suppose there can appear more issues related to it.

strarsis:
@tyranron: During container restart mail clients fail to connect and retrieve mail.
Also I had issues in the past where a restart fails because of incorrect ownership/permissions while the container still runs - resulting in downtime until it is discovered and manually fixed.

tomav:
Reloading postfix != Restart container
Reloading is enough and much faster.

strarsis:
So when the accounts/aliases configuration files have been changed, 
postfix has to be reloaded because the currently used configuration file format requires a reload after each change.

tomav:
`texthash` on aliases does not need reload/restart.
`hash` needs to, with a `postmap` before.

tyranron:
I understand your concerns.

But again, in most cases HA problem in Docker world is solved outside container by orchestration stuff (rolling restarts, multiple instances, etc). There are few exceptions I've seen related to situations when applications represents some kind of controller to work with dynamic services discovery.

The #382 was made from that point of view:
[...]

PR #738, opened 2017-10-07
Fix for #552

kamuri:


kamuri:
Ok tested everything. All works now as expected.

target/check_for_changes.sh
```diff
@@ -0,0 +1,124 @@
...
+	fi
+    
+    # Set right new if needed
+	if [ `find /var/mail -maxdepth 3 -a \( \! -user 5000 -o \! -group 5000 \) | grep -c .` != 0 ]; then
+		chown -R 5000:5000 /var/mail
+	fi
+    
+    
+    #supervisorctl restart postfix # Buggy causes error while testing.
+    postfix reload
```
johansmitsnl: This is my only concern because the process might lose the pid. Is the reload nesesary?
When we use the postmap a reload is not needed (http://www.postfix.org/postmap.1.html)

PR #754, opened 2017-10-25
Implement AWS SES mail fetching from S3 bucket

Zauberstuhl:
A view days ago I decided to use AWS as main MX and the docker instance as backup MX.
I wanted to share my modification cause that could be interesting for more then only me.

I haven't tested this PR, yet! If you decide on implementing this I will test and "prettify"  it.

---

HowTo; We can achieve that by creating a bucket called mail-aws-ses
within the bucket the mail should be stored in a directory
with the domain name as folder name e.g. s3://mail-aws-ses/zauberstuhl.de/...
This can be achieved by creating the right filter rules in aws console

After that we need an aws ID and secret token with full-s3-permissions
and add it to the docker environment variables AWS_SES_KEY_ID and AWS_SES_SECRET

Optional you can set a default region with AWS_SES_REGION variable

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,7 @@ RUN apt-get update -q --fix-missing && \
   apt-get -y install --no-install-recommends \
     amavisd-new \
     arj \
+    awscli \
     binutils \
     bzip2 \
     ca-certificates \
@@ -195,6 +196,9 @@ COPY ./target/bin /usr/local/bin
 # Start-mailserver script
 COPY ./target/check-for-changes.sh ./target/start-mailserver.sh ./target/fail2ban-wrapper.sh ./target/postfix-wrapper.sh ./target/docker-configomat/configomat.sh /usr/local/bin/
 RUN chmod +x /usr/local/bin/*
+
+# Setup AWS SES Fetch Mail
+RUN (echo "*/1 * * * * /usr/local/bin/fetch-aws-ses-mail"; crontab -l) | crontab -
 
 # Configure supervisor
 COPY target/supervisor/supervisord.conf /etc/supervisor/supervisord.conf
--- a/target/bin/fetch-aws-ses-mail
+++ b/target/bin/fetch-aws-ses-mail
@@ -1,5 +1,9 @@
 #!/bin/bash
 
+if [ -z "$AWS_SES_KEY_ID" ] || [ -z "$AWS_SES_SECRET" ]; then
+  exit 0; # AWS mail fetching is not configured
+fi
+
 find /var/mail -maxdepth 1 -name '*.*' |while read -r domainPath; do
   domain=${domainPath##*/};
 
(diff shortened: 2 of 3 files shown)


target/bin/fetch-aws-ses-mail
```diff
@@ -0,0 +1,24 @@
...
+  aws s3 mv s3://mail-aws-ses/${domain} \
+           /var/mail-aws-ses/${domain} --recursive || exit 1;
+
+  find /var/mail-aws-ses -type f |while read -r file; do
+    while read -r name; do
+      mail_dir="/var/mail/${domain}/${name}";
+      if [ ! -f ${mail_dir}/new/${file##*/} ]; then
+        mv -v $file ${mail_dir}/new;
+        sieve-filter -e -W -C -u ${name}@${domain} \
+          ${mail_dir}/sieve/rainloop.user.sieve INBOX
```
Zauberstuhl: Check if file exists

Dockerfile
```diff
@@ -196,6 +197,9 @@ COPY ./target/bin /usr/local/bin
 COPY ./target/check-for-changes.sh ./target/start-mailserver.sh ./target/fail2ban-wrapper.sh ./target/postfix-wrapper.sh ./target/docker-configomat/configomat.sh /usr/local/bin/
 RUN chmod +x /usr/local/bin/*
 
+# Setup AWS SES Fetch Mail
+RUN (echo "*/1 * * * * /usr/local/bin/fetch-aws-ses-mail"; crontab -l) | crontab -
```
Zauberstuhl: Should I do this while booting in the mailserver-start-script?

johansmitsnl:
* I understand the flow and code, but what is the main benefit of receiving email this way?
* It adds an extra delay because you need to "fetch" it from the S3 bucket. Having a normal MX and backup MX would not solve the issue to?
* Also how did you plan to test this addition with the tests in this project?

target/start-mailserver.sh
```diff
@@ -929,6 +933,19 @@ function _setup_postfix_relay_amazon_ses() {
 		"smtp_tls_CAfile = /etc/ssl/certs/ca-certificates.crt"
 }
```
Zauberstuhl: Rename function. It has nothing to do with postfix

Dockerfile
johansmitsnl: Yes you should.

Zauberstuhl:
@johansmitsnl aws ses has no real way of importing the received mail into a existing mailserver. You can store the mail in S3 or use a aws webmail solution. Thats why I did that.

In the worst case the delay would be 60 seconds thats true. Maybe i can tweak that but personally I don't care if my mail is a minute late.

I haven't checked the test framework yet. But it shouldn't be too hard using some appropriate s3 wrapper scripts.

Zauberstuhl:
If this is not the right place let me know then I do a blog post instead of a PR ;) Simply wanted to share my modification

johansmitsnl:
I think it is a creative solution that might be helpful to you.
Don't think it is something that fits in this project. Closing it because of this reason.

Feel free to add a WIKI section and create a modification based on this docker image as a base so that people can use it if there interested.

PR #767, opened 2017-11-25
Implement undef option for SA_SPAM_SUBJECT

elbracht:
Use `undef` to disable the sa_spam_subject_tag in amavis config. Currently the string "undef" is written before the email subject.


johansmitsnl:
You also could add a empty string if you don't want the tag?

elbracht:
I think the script use the default value for empty strings.

```
SA_SPAM_SUBJECT=${SA_SPAM_SUBJECT:="***SPAM*** "} && ...
```

johansmitsnl:
Correct, could you add a test to this PR to verify that it works?

elbracht:
Is there any other way than create a new docker container in `Makefile` to test this case? I think it's overkill for this test.

johansmitsnl:
This variable is already tested in:
https://github.com/tomav/docker-mailserver/blob/0748734b198ddbe9a67e7589bf73c4b011e4983a/Makefile#L21-L37

I guess adding a new container is the best way to test this.

elbracht:
I think the failed tests are not my fault

PR #769, opened 2017-11-30
 Add ability to override fail2ban.conf with fail2ban.local values.

mwlczk:
The test fail2ban overrides fail2ban.cf is succesfull but a little bit week, I assume the standard logging level is INFO.


johansmitsnl:
@mwlczk Thanks for the contribution

Issue #324, opened 2016-09-12
Test suite deficiencies


KyleOndy:
As I started looking back into #256 and #250 some maintainability issues have become apparent within the test suite. Given the current set of tests, coverage is low since we are only [testing the ability to send mail on the `mail` docker image.](https://github.com/tomav/docker-mailserver/blob/f42a0c193528b74777faef18683a8794d7cbb848/Makefile#L82-L101)

For instance, in #250, the test passes because [we only test that the service is not running](https://github.com/tomav/docker-mailserver/blob/f42a0c193528b74777faef18683a8794d7cbb848/test/tests.bats#L50-L53), however the ability to send an email is never tested, allowing the bug to be introduced.

As I see it, an ideal solution that would avoid the bugs above is to run through the basic functions of the mail server for each option that is configured.

---
# Proposal

_Possible implementation would need to be explored_
### A set of core tests that are common across all configurations

This is a mail server, so always test SMTP, IMAP, log configs, accounts and any other core services are always correctly working no matter what combination of environmental variables are set.
### Each environmental variable would have a set of core tests associated with each possible value.

**example:** `SSL_TYPE` would have 5 sets of tests associated with it.
1. `SSL_TYPE=` => verify no ssl is configured
2. `SSL_TYPE=letsencrypt` => verify Let's Encrypt certificates
3. `SSL_TYPE=custom` => verify custom certificates
4. `SSL_TYPE=manual` => verify custom locations of your SSL certificates for non-standard cases
5. `SSL_TYPE=self-signed` => verify self-signed certificate work
### Some kind of preprocessorfor BATS that generates all tests cases

Once this information is defined, it would be possible to generate complete test coverage for every configuration (all combinations of environmental variables) by using the associated set of tests for each option being tested.
## Possible Issues
1. Compexity in creating / maintaing a BATS preprocessorfor.
2. The number of combinations grows exponentially, making testing all combinations infeasible.

---

The more I think about this particular solution, the messier it seems. Ultimately, I still think there are coverage issues that should be solved.

Thoughts?

tomav:
Hi @KyleOndy, great summary.
Same conclusion here, but few points to move forward:
- `bats` allows to [skip tests](https://github.com/sstephenson/bats#skip-easily-skip-tests)
[...]

PR #784, opened 2017-12-28
debian stretch slim

mwlczk:
switch to debian:stretch-slim:
 - first step correct the testdata, as newer packages are more strict
about the mail-structure.

 - add missing build-step in `make`
 - clean the userdb aswell
 - use timeout of netcat (nc -w), as postgrey would not close the connection
 - there is 2 extra mail-log-files (mail.info and mail.warn) -> count only mail.log
 - cosmetic: use "" instead of ''

new image benefits:
 - smaller size (52MB less)
 - 0 CVEs compared to 11 CVEs in ubuntu 16.04 Image

better backport situation:
 - postfix 3.1.6 vs 3.1.0
 - fail2ban 0.9.6 vs 0.9.3
 - openssl 1.1.0f vs 1.0.2g
 - dovecot 2.2.27 vs 2.2.22
 ....

changes in Dockerfile and configfiles needed because of stretch-slim:
- add missing gnupg and iproute2 package
- remove non-free rar, unrar-free should do
- rsyslog does not add syslog user and has different conf-structure
- pyzor command discover was deprecated and is missing in the new
stretch package

- dovecot does not know SSLv2 anymore. removed because of warnings in
log

- iptables does not know imap3, IMAP working group chose imap2 in favor
of imap3

- s. following commits aswell


mwlczk:
The tests run fine on my machine. Need to look into it, why Travis breaks. They changed the architecture mid-December though. Any hints on that would be great

mwlczk:
retrigger travis ci to get a closer look into the build.

mwlczk:
this is odd: travis might have some caching (or no-cache) issues.
First build run had 9 failures.
Second build run had only 1 failure - but no source code has been changed in between.

Gonna hunt that last failure down now.

mwlczk:
One more Travis try

mwlczk:
Last two Travis runs exit at different tests.

mwlczk:
Travis makes me wonder. If I seemingly fix one bug/test, it fails another test, which is executed earlier in the queue and passed all the times before.

johansmitsnl:
@mwlczk First of great PR!

A very good contribution, myself also tried to change this image to Debian because of better backport support that lacks in Ubuntu sometimes, couldn't agree more.

Sometimes travis is busy and due to the amount of containers that are started some tests are not yet ready to be tested and this gives sometimes some mixed test results. This is a improvement that needs to be applied to this project #324.

Looking forward to merging this PR but some questions:

* Could you remove commented old code?
* Besides the tests are you using the PR live?
* Any notes that should be considered when merging, todo's or ...?

Dockerfile
```diff
@@ -60,7 +62,7 @@ RUN apt-get update -q --fix-missing && \
     postfix-pcre \
     postfix-policyd-spf-python \
     pyzor \
-    rar \
+    # rar \
```
johansmitsnl: Remove the line if not used

Dockerfile
```diff
@@ -177,8 +180,9 @@ RUN sed -i -r "/^#?compress/c\compress\ncopytruncate" /etc/logrotate.conf && \
   chown -R clamav:root /var/log/mail/clamav.log && \
   touch /var/log/mail/freshclam.log && \
   chown -R clamav:root /var/log/mail/freshclam.log && \
-  sed -i -r 's|/var/log/mail|/var/log/mail/mail|g' /etc/rsyslog.d/50-default.conf && \
-  sed -i -r 's|;auth,authpriv.none|;mail.none;mail.error;auth,authpriv.none|g' /etc/rsyslog.d/50-default.conf && \
+  # no separate 50-default.conf
```
johansmitsnl: Remove the line if not used

test/tests.bats
```diff
@@ -166,8 +166,8 @@ load 'test_helper/bats-assert/load'
 }
 
 @test "checking postgrey: there should be a log entry about the whitelisted and passed e-mail user@whitelist.tld in /var/log/mail/mail.log" {
-  run docker exec mail_with_postgrey /bin/sh -c "nc 0.0.0.0 10023 < /tmp/docker-mailserver-test/nc_templates/postgrey_whitelist.txt"
-  sleep 8
+  run docker exec mail_with_postgrey /bin/sh -c "nc -w 8 0.0.0.0 10023 < /tmp/docker-mailserver-test/nc_templates/postgrey_whitelist.txt"
+  # sleep 8
```
johansmitsnl: Remove the line if not used

mwlczk:
@johansmitsnl
- removed commented old code
- am using this PR live
- Notes: We should have a look at the cipher-suites for all 
  TLS-Endpoints (smtp, smtpd, imaps, pop3s, ...), because newer
  package (e.g. openssl) might cause problems with older stuff. 
  Anyway we should keep an eye to update them in a next PR.

PR #746, opened 2017-10-15
Find the right interface for the container #607

dlion:

Issue #607, opened 2017-05-16
Assumes existance of eth0 device


nalt:
If there is not eth0 device, setup fails with "Device eth0 not found". Some docker network configs do not have eth0.

tomav:
True, `eth0` seems hardcoded here: https://github.com/tomav/docker-mailserver/blob/master/target/start-mailserver.sh#L845

johansmitsnl:
Could you create a PR?

cofilippo:
I notice this issue is still open. Does this mean that I won't be able to use my bridged config properly? (br0 bridges eth0 on my machine)

erik-wramner:
Closed in favor of #1227 (duplicate).

PR #785, opened 2018-01-02
Update command to set container_network

gkobeaga:


johansmitsnl:
Check #746 and #607 that we want to dump the hardcoded eth0 name.

gkobeaga:
What about setting it as environmental variable? I am using Docker Swarm to deploy it, which uses multiple interfaces. The automatic finding might cause problems.

johansmitsnl:
I see that test `not ok 119 checking PERMIT_DOCKER: my network value` is failing. Could you verify this?

erik-wramner:
Is this still relevant? No changes since January 2018, should we close it or make it work?

PR #800, opened 2018-01-24
Added support for Dovecot and Postfix LDAP TLS

madmath03:
Added 2 parameters:
- `LDAP_START_TLS` to enable Postfix LDAP TLS
- `DOVECOT_TLS` to enable Dovecot LDAP TLS

Tests and README updated.

Wiki will need to be updated to once PR has been accepted.


target/postfix/ldap-domains.cf
```diff
@@ -5,4 +5,5 @@ query_filter     = (&(|(mail=*@%s)(mailalias=*@%s))(mailEnabled=TRUE))
 result_attribute = mail
 search_base      = ou=people,dc=domain,dc=com
 server_host      = mail.domain.com
+start_tls        = no
 version          = 3
```
johansmitsnl: Could you add the new line

PR #804, opened 2018-01-26
Add dovecot mysql

julianhille:
This PR allows to configure dovecot etc to have a mysql server as user backend. Actual mysql / mariadb has to be another docker machine.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,6 +32,7 @@ RUN apt-get update -q --fix-missing && \
     dovecot-imapd \
     dovecot-ldap \
     dovecot-lmtpd \
+    dovecot-mysql \
     dovecot-managesieved \
     dovecot-pop3d \
     dovecot-sieve \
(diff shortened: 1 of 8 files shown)


julianhille:
is someone able to tell me why the tests are failing? This looks like a general error instead of one introduced by me.

julianhille:
Will have a look if i can dive into the test system. Will take me some days to finish this up. But yes indeed tests are needed

unomind:
any new for this one?

erik-wramner:
Personally I would vote for closing this as it goes against the basic idea with this project: "Only configuration files, no SQL database". If we want to support SQL we need a whole range of additional tests. In my view it would be better to create a fork with a server that only uses SQL. I think that would be easier to maintain.

erik-wramner:
Waiting for contributor action since March 2018, no response on comments and in my view outside of the project's scope. Closing.

keducoop:
In case that somebody lands here we've just created a [docker image](https://hub.docker.com/repository/docker/kedu/docker-mailserver-mysql) merging [dovecot-mysql](https://github.com/julianhille/docker-mailserver/tree/dovecot-mysql) branch with current [tomav/docker-mailserver](https://github.com/tomav/docker-mailserver), [here](https://github.com/Kedu-SCCL/docker-mailserver-mysql) the code.

Cheers.

Issue #795, opened 2018-01-16
How to move Spam messages into Spam folder?


modInfo:
```
    environment:
    - ENABLE_SPAMASSASSIN=1
    - ENABLE_CLAMAV=0
    - ENABLE_FAIL2BAN=0
    - ENABLE_POSTGREY=0
    - SA_KILL=1000
    - ONE_DIR=1
    - DMS_DEBUG=0
    - SSL_TYPE=manual
    - SSL_CERT_PATH=/etc/letsencrypt/live/mail./fullchain.pem
    - SSL_KEY_PATH=/etc/letsencrypt/live/mail./privkey.pem
```

## My solution:

/var/lib/dovecot/sieve/default.sieve
```
require "fileinto";
if header :contains "X-Spam-Flag" "YES" {
        fileinto "Spam";
}
```

/etc/dovecot/conf.d/90-sieve.conf
```
  sieve_default = /var/lib/dovecot/sieve/default.sieve
```

Compile this file by sieve compiler
```
cd /var/lib/dovecot/sieve/
sievec default.sieve
```

do not forget to commit changes
`docker commit <container-id> tvial/docker-mailserver:latest`

17Halbe:
It‘s the default and configurable implementation: 
See also #768

modInfo:
How do it automatically move to Junk?
What settings have I forgotten to add to docker-compose?

17Halbe:
From the link, I provided: 

> It it gets over the kill level it deletes the email. If it is below but also spam it gets in je Junk folder.
This is default.

So if you set the kill level to something high(or low(?)) even the worst spam won‘t be deleted but also be moved to your Junk folder

modInfo:
Junk folder doesn't exist for new users how to create it?

17Halbe:
Are you sure? It should have gotten created by dovecot at creation time. But one is not "subscribed" to it. To see if had been created:
`./setup.sh debug login
ls -al /var/mail/<DOMAIN>/<USERNAME>`
Where you replace <DOMAIN> with your domain (like domain.net) and <USERNAME> with the address portion of the email-address in question ( user in case of user@domain.net)
You should then see a folder `.Junk` (besides the rest of the mailbox)

You could subscribe to it using your Mail-client. I guess..

modInfo:
```bash
root@mail:/var/mail/my.eu/test# ls -al
total 64
drwx------ 10 docker docker 4096 Jan 16 22:42 .
[...]

Issue #768, opened 2017-11-29
How to create a global sieve filter ? (Spam moving)


H4R0:
I need help with setting up a sieve filter that runs for every user, how to do that ?

Why does the container delete spam by default ? I missed some important mails because of that (false positive). The log shows they hit a score of 8 and got rejected.

I now configured Spam handling the following way:
- Setting env SA_KILL to 1000 which prevents blocking / deletion of spam mails.
- Running docker exec -it mail sed -i 's/mailbox Junk {/mailbox Spam {\n    auto = subscribe/g' /etc/dovecot/conf.d/15-mailboxes.conf 
which will create the Spam folder for every User automatically and marks it as Junk
- setting up a sieve filter for every users to move Spam into the Spam folder
cat << 'EOF' > /srv/mail/config/user@domain.com.dovecot.sieve
require "fileinto";
if header :contains "X-Spam-Flag" "YES" {
  fileinto "Spam";
}
EOF

However i would prefer to just create 1 Sieve filter which gets applied to all Users. That should be possible but i don't know how. Could somebody please help me with it, thanks.

johansmitsnl:
It it gets over the kill level it deletes the email. If it is below but also spam it gets in je Junk folder. 
This is default.

dar72:
I think it would be a good idea to not delete anything by default, simply mark as spam without deleting.

I potentially lost important mail due to false positives on initial setup

H4R0:
Solution at #795

PR #806, opened 2018-01-27
Changed Junk folder to be created and subscribed by default

H4R0:
The Junk folder is currently not created by default. Also there is no documentation on how to setup the Junk folder. Imho the Junk folder should be created and subscribed by default as the other folders. #795 #768


johansmitsnl:
Why does the ldap test fail?

Issue #811, opened 2018-01-31
Postfix reject_sender_login_mismatch


TechnicLab:
Consider using reject_sender_login_mismatch in postfix to prevent users from sending emails from any address.

johansmitsnl:
Could you suggest a PR?

TechnicLab:
I’m happy to but it looks like there are some new configuration needed for current users: http://www.postfix.org/postconf.5.html#smtpd_sender_login_maps, also I’m not sure how to configure this with LDAP.

johansmitsnl:
I think no extra configuration is needed, but if you have a branch I can test it (i use a ldap setup).

TechnicLab:
Ok, I will create it this evening.

PR #819, opened 2018-02-06
Added  reject_sender_login_mismatch

TechnicLab:
Added reject_sender_login_mismatch parameter to prevent users from sending email using arbitrary email address. #811


17Halbe:
Hmm.. I might got this wrong, but don't you need a mapping file to make that work? Is this really denying a logged in user to spoof it's FROM address? I always had to implement a mapping file.

Which would probably look like:
main.cf:
```
...
smtpd_sender_login_maps = pcre:mapping.file
...
```
and the 
mapping.file:
```
/^(.*)$/   ${1}
```

All from memory, but it took me quite a while back than to figure that out.. So it sticks! ;) 
And yes, this would map every user to only his mailaccount.

TechnicLab:
Great idea! It is not working in my implemention, I actually created another fork request but there was a problem with requipment of mapping users. Could you fix this please?

17Halbe:
I will fix it, if you implement a test for it! ;)

TechnicLab:
I am not familiar with tests that are used in this image, but I could try...

TechnicLab:
https://github.com/tomav/docker-mailserver/pull/825 - here is another pull request. Unfortunately forking only with ldap.

17Halbe:
Oh, great find! :) That would probably be replaceable with my idea.. I'll have a look

PR #821, opened 2018-02-07
Nist tls ciphers

mwlczk:
Did some testing with https://www.htbridge.com/ssl/ 
according to them (NIST; HIPAA) the two POLY ciphers are not compliant, so removed.
I use Atom for editing, and this editor trims all the lines in every file while saving it.


johansmitsnl:
Do tests pass on your machine?
I ran travis twice and both fail with a timeout.

mwlczk:
was a quick hack this morning during breakfast - not a good idea ;)
gonna redo it better some time later.

mwlczk:
issues resolved in #831 .

PR #825, opened 2018-02-07
Added smtpd_sender_login_maps

TechnicLab:
Added smtpd_sender_login_maps for reject_sender_login_mismatch postfix option.


johansmitsnl:
This file does not needs to be only set when using ldap?

TechnicLab:
Totally forgot about it - yes, this modification is for ldap. Not sure where to find file with those contents without ldap:
"The controlled_envelope_senders table specifies the binding between a sender envelope address and the SASL login names that own that address:

/etc/postfix/controlled_envelope_senders
    # envelope sender           owners (SASL login names)
    john@example.com            john@example.com
    helpdesk@example.com        john@example.com, mary@example.com
    postmaster                  admin@example.com
    @example.net                barney, fred, john@example.com, mary@example.com
"

johansmitsnl:
So this can be merged or are changes still needed?

TechnicLab:
Now it works in my ldap setup. Waiting for tests.

17Halbe:
Can you please check if this would work with ldap as well. Otherwise we would have to differentiate between ldap and non ldap builds.

Please see the comments.

TechnicLab:
A am a beet confused. You would like me to check your solution from comment in another pull request? I will test it when I will be near PC.

TechnicLab:
A am a beet confused. You would like me to check your solution from comment in another pull request? I will test it when I will be near PC.

17Halbe:
Sorry for the confusion.. 
Yes, exactly! That would be great.
Your PR is only working for ldap users. So in case we would implement smtpd_sender_login_maps for everyone else as well, we would have to distinguish upon server start, which mapping file we need to use.
(Depending if that's a ldap or a "standard" setup) 
So, if the regex suggestion works for you/ldap as well, than we just need this one mapping file for everyone.

I hope this was less confusing!

johansmitsnl:
@TechnicLab could you fix the conflict?

johansmitsnl:
Almost all tests fail, looks like a typo somewhere?

TechnicLab:
Actually there was no confusion anymore, it was too late in my time zone. Unfortunately stilll no near PC(

TechnicLab:
Unfortunately it is not working with my ldap setup. I guess we need to separate ldap and simple configurations.

johansmitsnl:
I personally use ldap but I don't encounter any issues at the moment with master. Does this mean the restriction is not working?

ps. I have extracted the ldap service in this project in it own docker: https://hub.docker.com/r/jsmitsnl/docker-openldap-postfix-book

TechnicLab:
I manually added changes from 17Halbe and they lead to inability to send any mail. In current master you can still send messages from any email.

johansmitsnl:
Ok, why do the tests for this PR fail do you know why?

TechnicLab:
Actually I have no idea.

17Halbe:
Ok thanks for trying.. Shouldn't this be the default(not letting every signed in user spoof its address)?

TechnicLab:
No, this is not the default behavior.

17Halbe:
I know, but IMHO it should be! 
\So it has to be implemented for everyone else as well!

TechnicLab:
I absolutely agree with you, but maybe there is a reason for such default configuration?

17Halbe:
I'll provide a PR on this in a few days. Should be easy enough.

TechnicLab:
warning: pcre map /etc/postfix/sender.cf, line 1: no closing regexp delimiter "+": ignoring this rule. It looks like there is an error with your regexp.

17Halbe:
You copied the + from the comment as well.. Which is actually great, since this leaves a silverlining, that it will still probably work.
Just remove the leading + from the expression! Does it work?

TechnicLab:
Great! It working now.

TechnicLab:
Well, this pull request can be merged when we will determine why it is causing tests failitures.

17Halbe:
Can you add a testcase for a user spoofing it's address?

TechnicLab:
I am making changes mainly online on GitHub without cloning repo, so I can’t check whatever test work locally or not. I will look into projects’s tests closer and will try to create spoofing test.

17Halbe:
Did you check, if you could still receive mail from a foreign address? I think this is actually be blocking anyone sending mails to the server.
If that's the case(which I am pretty sure of), try replacing reject_sender_login_mismatch  with `reject_authenticated_sender_login_mismatch`

TechnicLab:
Looks like you are right, my test container rejects mail from foreign addresses.

17Halbe:
and reject_authenticated_sender_login_mismatch? Does that work as expected?

TechnicLab:
Yes

17Halbe:
/etc/postfix/senders.cf: No such file or directory
Modify the Dockerfile to copy that file over to the container

TechnicLab:
Thats's strange: ldap tests failed but it works locally.

johansmitsnl:
@TechnicLab Could you fix the conflict and update the branch for a retest?

TechnicLab:
Well, it should work now. I will add tests today.

johansmitsnl:
@TechnicLab love to merge the PR, could you inform me about the state of adding tests?

TechnicLab:
I’m sorry, I am currently ill and was not in great working condition.

johansmitsnl:
@TechnicLab hope you get well soon

17Halbe:
@TechnicLab  I hope you don't consider this being rude, I just needed some distraction from work and for a change wanted to contribute to something useful! :wink: 
So I continued your work and provided a PR here: #872

17Halbe:
Since #872 has been merged, this can be closed, right?

erik-wramner:
Closing due to no response, probably fixed by #872.

PR #418, opened 2016-12-21
Secure TLS protocols

dbellavista:
Use suggested protocols and ciphers (modern systems) (https://wiki.mozilla.org/Security/Server_Side_TLS#CipherScan)

See issue #410

PR #831, opened 2018-02-09
Nist tls

mwlczk:
A new PR because some "spaces", line breaks etc. destroyed the main.cf in #821 
This time added dovecot tls-cipher corrections as well.


mwlczk:
But if I read this (https://blog.cloudflare.com/do-the-chacha-better-mobile-performance-with-cryptography/), I wonder and would love to have little bit of a discussion before we merge this PR.

It seems like these two ciphers (ChaCha20-Poly1305) are not NIST-compliant, but who is NIST to tell us what is secure ;)

PLEASE do not merge now!

17Halbe:
Ok, I tried to follow the article and have to admit, that I always took the cipher lists from [BetterCrypto.org](https://bettercrypto.org/static/applied-crypto-hardening.pdf). Which are at this time probably outdated. 
(Which resulted in those chiphers: `EDH+CAMELLIA:EDH+aRSA:EECDH+aRSA+AESGCM:EECDH+aRSA+SHA256:EECDH:+CAM
ELLIA128:+AES128:+SSLv3:!aNULL:!eNULL:!LOW:!3DES:!MD5:!EXP:!PSK:!DSS:
!RC4:!SEED:!IDEA:!ECDSA:kEDH:CAMELLIA128-SHA:AES128-SHA`)
So I really appreciate your effort! :+1: 

Sounds like a good idea supporting ChaCha20-Poly1305, even more so since they are negotiable anyway. So there shouldn't be a downside, right? And it seems like a good idea for mobile devices too..
Anyhow, I agree: Who is NIST to tell us! ;) (And who knows if they are still trustable anyway. Remember the NSA elliptic curves discussion back than? )

So for me it would be fine.

By the way: Why did you remove those [ ] around the arguments in the usage part of the setup shell script? They are optional and should be surrounded by them, [right?](https://en.wikipedia.org/wiki/Usage_message)

johansmitsnl:
@mwlczk could you solve the conflict so that I can merge it?

mwlczk:
@17Halbe @johansmitsnl 
We need a proper discussion about this topic. I read #418 - there was a discussion about this before, of course. 
I digged as well into:
OWASP: https://www.owasp.org/index.php/TLS_Cipher_String_Cheat_Sheet
Mozilla: https://wiki.mozilla.org/Security/Server_Side_TLS
BetterCrypto: https://bettercrypto.org/static/applied-crypto-hardening.pdf

All Mac OS X before High Sierra (latest) are stuck at OpenSSL 0.9.8 -> only TLSv1 is supported. We cannot use "modern" (mozilla) / "A+/A" (owasp)  cipher suites for this type of clients.
Many Mail-servers might not implement newer technology aswell because the OP-teams might be insufficient. As BetterCrypto states in this case in it better to have some encyrption than not to have any and be forced to use plain. 

We should therefore use as the default  "intermediate" (mozilla) / "B" (owasp) ciphersuites to have wide support of encryption with clients and servers. And of course we should keep the newer ChaCha20-Poly1305 as they are a lot faster in software implementations - most smartphones don't have a dedicated AES-NI acceleration.
Additionaly we can provide an easy way to "optin" the stronger / modern cipher-suites via "postfix-main.cf" and "dovecot.cf" overrides.

I would commit the changes to this PR, if we all agree to that strategy.

P.S.
I would stick to OpenSSL (1,1.0 should be a good step forward) and not to expermient with LibreSSL for now. We have some other refactorings in the last weeks for this mail project.

P.P.S
@17Halbe I would remove my changes of the brackets to setup.sh, don't why they disapperad. Sorry for that.

mwlczk:
> We should therefore use as the default "intermediate" (mozilla) / "B" (owasp) ciphersuites to have wide support of encryption with clients and servers. And of course we should keep the newer ChaCha20-Poly1305 as they are a lot faster in software implementations

> Additionaly we can provide an easy way to "optin" the stronger / modern cipher-suites via "postfix-main.cf" and "dovecot.cf" overrides.

@johansmitsnl @17Halbe are we fine with this (my recommendation)? Or should we do the opposite: default is strong, and opt-in for moderate/intermediate ciphers.

johansmitsnl:
My personal preference is to have strong settings and weaken (easily) if needed. Then you make the decision about weaken them and making it less secure.
But again it should be "easy" to do so.

17Halbe:
I second @johansmitsnl suggestion. Which was your proposal as well, right?
How do we implement the "weakening"? Is a wiki entry sufficient? 
You basically just would have to supply a `postfix-main.cf` and  a `dovecot.cf` with altered cipher entries.
But a wiki entry would be necessary since this is quite "in depth" stuff.

@mwlczk No worries, You didn't hurt my feelings, I was just wondering! ;)

johansmitsnl:
@17Halbe it was indeed.
I guess a WIKI should be sufficient indeed, it could also explain how to change them in general when needed with a disclaimer that they should only change when they know what they are doing.

mwlczk:
oh, well i'm gonna run the tests tomorrow morning again. had run two ci-build accidently at the same time.

mwlczk:
restart travis, because tests don't fail on my machine.

mwlczk:
as so often, Travis fails on different tests with the same code base.
The refactoring of the test suite gets more and more important.

johansmitsnl:
@mwlczk restarted

PR #834, opened 2018-02-11
Diffie-Hellman 2048 Bit Parameters should be changed regularly.

17Halbe:
Since it is assumed that the NSA uses rainbowtables to break default-DHE-Parameters, one is encouraged to change those parameters periodically.

See: 
Paper: [Imperfect Forward Secrecy: How Diffie-Hellman Fails in Practice](https://weakdh.org/imperfect-forward-secrecy-ccs15.pdf)
Consideration: [Wie die NSA den Diffie-Hellman Schlüsseltausch knackt](https://www.heinlein-support.de/blog/news/wie-die-nsa-den-diffie-hellman-schluesseltausch-knackt/)(German)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -178,7 +178,9 @@ RUN mkdir /var/run/fetchmail && chown fetchmail /var/run/fetchmail
 COPY target/postfix/main.cf target/postfix/master.cf /etc/postfix/
 COPY target/postfix/sender_header_filter.pcre /etc/postfix/maps/sender_header_filter.pcre
 RUN echo "" > /etc/aliases && \
-  openssl dhparam -out /etc/postfix/dhparams.pem 2048
+  openssl dhparam -out /etc/postfix/dhparams.pem 2048 && \
+  echo "@daily FILE=`mktemp` ; openssl dhparam -out $FILE 2048 > /dev/null 2>&1 && mv -f $FILE /etc/postfix/dhparams.pem" > /etc/cron.d/dh2048
+
 
 # Configuring Logs
 RUN sed -i -r "/^#?compress/c\compress\ncopytruncate" /etc/logrotate.conf && \


17Halbe:
As usual: tests ... work for me! ;)

Starbix:
If we are so concerned about security, why aren't 4096 bit DH parameters used? 
Also, isn't daily a bit much ? I mean it doesn't hurt, but I feel like weekly should be enough.

ulm0:
+1 to weekly 4096 bit DH

17Halbe:
Even though this has been merged, I can change that daily to weekly if you'd like.. 

Considering 4096 bit: 1024bit DH is only just getting feasable, 2048 is far in the "cannot break it" zone and bear in mind, that *every additional *bit** **doubles** the attack space.. 
This PR was more or less a precautionary thing and common best practice.

Regardless: The last time I tried switching to 4096 bit, around 3 years ago, couple of problems arouse. Clients weren't able to connect because of lacking support and mobile (android) devices took (I guessed) too long to compute the 4096bit(at least I had timeout issues with them)

But I agree, 2048bit would be sufficient to be regenerated every week..

Starbix:
4096 bit DH parameters work great for my nginx setup and all clients I've tried were able to connect to it, but I can imagine that mail clients don't have have up-to-date crypto libraries like browsers do. So you're probably right about that. 
I'd recommend 3072 bit, but support for that is worse than for 4096 bit if I remember correctly.
So generating a new 2048 bit param weekly is the way to go imo.

johansmitsnl:
@17Halbe could you add a PR for a weekly update?

Starbix:
I added one #836

hardware:
Instead of generating your own DH groups with "openssl dhparam", you should use the pre-defined DH groups ffdhe2048, ffdhe3072 or ffdhe4096 recommended by the IETF in [RFC 7919 https://tools.ietf.org/html/rfc7919]. These groups are audited and may be more resistant to attacks than ones randomly generated.

And the Postfix SMTP server EDH parameters file is not secret, these parameters are sent in clear to all remote SMTP clients.

17Halbe:
How fast those things change! ;) 
So would you like to do a PR? I'm out of time till maybe next week..

PR #836, opened 2018-02-12
Generate new DH param weekly instead of daily

Starbix:
#834


17Halbe:
@Starbix Thanks for jumping in!

Issue #844, opened 2018-02-16
Error in default configuration


Tygore:
Hello,
Standard settings. Changed only in .env:
hostname: mail
domainname: mycooldomain.com

I can not receive and send mail. In the logs of the server writes:
`

- Feb 16 11:03:54 mail postfix/postscreen[12994]: CONNECT from [172.21.0.1]:54410 to [172.21.0.2]:25
- Feb 16 11:03:54 mail postfix/postscreen[12994]: PASS OLD [172.21.0.1]:54410
- Feb 16 11:03:54 mail postfix/smtpd[13007]: error: open database /tmp/docker-mailserver/postfix-send-access.cf: No such file or directory
- Feb 16 11:03:54 mail postfix/smtpd[13007]: error: open database /tmp/docker-mailserver/postfix-receive-access.cf: No such file or directory
- Feb 16 11:03:54 mail postfix/smtpd[13007]: connect from unknown[172.21.0.1]
- Feb 16 11:03:54 mail postfix/smtpd[13007]: Anonymous TLS connection established from unknown[172.21.0.1]: TLSv1.2 with cipher ECDHE-RSA-AES256-GCM-SHA384 (256/256 bits)
- Feb 16 11:03:54 mail postfix/smtpd[13007]: warning: texthash:/tmp/docker-mailserver/postfix-send-access.cf is unavailable. open database /tmp/docker-mailserver/postfix-send-access.cf: No such f
- ile or directory
- Feb 16 11:03:54 mail postfix/smtpd[13007]: warning: texthash:/tmp/docker-mailserver/postfix-send-access.cf: table lookup problem
- Feb 16 11:03:54 mail postfix/smtpd[13007]: NOQUEUE: reject: RCPT from unknown[172.21.0.1]: 451 4.3.5 <xxxx@gmail.com>: Sender address rejected: Server configuration error; from=<xxxx@gmail.com> to=<tygore@mycooldomain.com> proto=ESMTP helo=<mail-lf0-x232.google.com>
- Feb 16 11:03:54 mail postfix/smtpd[13007]: disconnect from unknown[172.21.0.1] ehlo=2 starttls=1 mail=1 rcpt=0/1 data=0/1 quit=1 commands=5/7
- Feb 16 11:06:20 mail postfix/postscreen[13505]: CONNECT from [172.21.0.1]:54416 to [172.21.0.2]:25

`

My DNS: 
mail		300	MX	10	mail.mycooldomain.com
[...]

PR #846, opened 2018-02-16
postscreen implementation altered

17Halbe:
made the config files `postfix-sender-access.cf` as well as `postfix-receive-access.cf` optional.

main.cf modification happens when either the user uses the setup.sh script or when providing a ...-access.cf file.

Please have a look if the postfix restart is implemented correctly. (I wasn't sure if `service postfix reload`, `supervisorctl restart postfix` or `/etc/init.d/postfix restart` is desired) (`service postfix reload` gets trapped anyway.)
This will most probably solve #844


17Halbe:
Ah forgot to mention that the tests succeed on my machine.

17Halbe:
Tests still passing on my side..

johansmitsnl:
Could you fix the conflict so that I can merge it?

17Halbe:
See https://github.com/tomav/docker-mailserver/pull/846#issuecomment-366501435
;)

17Halbe:
Should have closed #847 and #844

PR #851, opened 2018-02-19
Latest commit 5e09074d587114ef4d771fe65c911b67735d4fcb by @17Halbe br…

srslynow:
Access control features are nice, but they break the default setup.. These files are required for startup. Were removed in previous commit 5e09074d587114ef4d771fe65c911b67735d4fcb


17Halbe:
Those files shouldn‘t be neccessary anymore. And be fully optional. 
It‘s being tested if they exist during mailserver startup and if so, the required entries in main.cf are made. 
Do you still experience startup problems?

17Halbe:
Should actually work as expected, otherwise the tests wouldn‘t have been past successfully.

srslynow:
Try sending an email, you'll get a sender adress rejected error. Startup works fine because those files are only checked at receive or send events.

17Halbe:
Thanks for the report. I might have overlooked something. It’s working fine here...
So merge please. It won‘t hurt anyway.


...von unnerwegs...

17Halbe:
Hmm. Not here.. I‘ll check again. This isn‘t as elegant as I intended it to be, ;) but it‘s a definitly working workaround. 
And I‘m still quite a bit puzzled why this isn‘t working.

17Halbe:
Ah, got the same problem... BUT the docker image has not been updated yet. It doesn't reflect the changes I implemented.
Can you have a look if you got a result of the following command:
`docker exec mail cat /usr/local/bin/start-mailserver.sh |grep access_control`
replace mail of course with your container name..

17Halbe:
@johansmitsnl Are you also in charge of the docker builds? Can you get them running?

config/postfix-receive-access.cf
```diff
@@ -0,0 +1 @@
+-
```
17Halbe: missing new line

johansmitsnl:
@17Halbe yes I can restart them.

srslynow:
@17Halbe Seems you were right about the update, ```docker exec mail cat /usr/local/bin/start-mailserver.sh |grep access_control``` doesn't return anything. The errors were definitely pointing towards non-existant ```/tmp/docker-mailserver/postfix-receive-access.cf``` and ```/tmp/docker-mailserver/postfix-send-access.cf``` files though.

17Halbe:
Yes, and you are still right, since the problem's not solved yet! Your PR wouldn't have helped though either! ;)  
Really appreciated your PR, otherwise we wouldn't have noticed docker isn't building the image anymore! :+1:

srslynow:
Strange, creating these files did fix the problem for me. I have the extra newlines in those files on my linux box running the mailserver. The MR was created from a Windows laptop, and I erroneously left out the required newlines there. Anyways, you're aware of the issue, thanks for the fast responses!

boaorm:
I guess this happened to me too when rebuilding everything this morning. So now I'm a bit worried of using v2tec/watchtower for auto-updating this project. My intention is to use docker-mailserver in 'production' (personal use).

srslynow:
@boaorm this is the exact setup I'm running as well, I'm going to exclude this image from watchtower today unfortunately. This is not the first time things break. I love this image and I'm really grateful to the contributors but its releases are just not stable enough for blind auto-updating.

PR #866, opened 2018-02-27
update postmaster_address in dovecot config according to POSTMASTER_ADDRESS env var

andihit:
Update the postmaster_address in the `start-mailserver.sh` script.

At the moment this config is set in the `Dockerfile` at: https://github.com/tomav/docker-mailserver/blob/c36e878d76827fc7062558f1edada05d7ba38880/Dockerfile#L111
But this code is executed only when the docker image is built and not at runtime.


andihit:
The config variable is documented already (but didn't work correctly before): https://github.com/tomav/docker-mailserver/tree/c36e878d76827fc7062558f1edada05d7ba38880#postmaster_address

Issue #855, opened 2018-02-20
openDKIM key length


pn2:
This is a wish for a feature:

Some DNS servers restrict the length of TXT fields. 
The opendkim default key length seems to have been increased from 1024 to 2048 bits. In my case, the DKIM entry now does not fit the TXT field anymore.

I now would like to have an option to shorten the key length back to 1024 bits to make it fit. I understand that it of course would be better to keep the longer keys and fix the problem on the DNS side, but my provider seems unable to do so.
 
What about a corresponding environment variable, like DKIM_KEY_LENGTH ?

leokun:
I have the same problem, I cannot past all the key (about 16 digit missing)

johansmitsnl:
Can you provide a PR?

mwlczk:
This is a duplicate of #786 .
In fact debian uses key size length of 2048 as default since January 2017(!) Those bloody debians living on the edge ;)

> opendkim (2.11.0~alpha-9) unstable; urgency=medium
  * Set umask to 0007 in opendkim.service so opendkim socket is group readable
    (Closes: #851141)
  * Set opendkim-genkey to default to 2048 bit sha-256 only keys based on
    current best practices
  * Add debian/patches/nsupdate_output.patch to fix opendkim-genzone output to
    be usable for nsupdate (Closes: #849540)
 -- Scott Kitterman <scott@kitterman.com>  Wed, 25 Jan 2017 09:58:07 -0500

can anyone (@cruisemaniac, @youaresofunny, @leokun) provide a PR accompanied by a small test for the size generated once as 1024 and once as 2048 Bit key?

johansmitsnl:
#868 is merged. You can now generate with any key size

Issue #786, opened 2018-01-04
Namecheap does not allow more than 1024 bits in the DKIM


cruisemaniac:
With reference to this: https://www.namecheap.com/support/knowledgebase/article.aspx/317/2237/how-do-i-add-txtspfdkimdmarc-records-for-my-domain

This can be fixed by adding a `-b 1024` parameter to the `generate-dkim-config` and `generate-dkim-domain` scripts.

Is this acceptable? Or is there another way? I'm on namecheap for my domains and I cannot move from there in the near future.

johansmitsnl:
Expanding the 2 scripts to accept the option and add this documentation feels better. Changing defaults may break other users expectations in automatisation.

cruisemaniac:
@johansmitsnl Totally agree with this. Let me dig around and send in a PR. Does that sound fair?

johansmitsnl:
Sounds good

johansmitsnl:
Please add a PR. Will close the issue for now.

mwlczk:
@johansmitsnl @cruisemaniac 
If it comes to namecheap, we don't need any changes in source-code for now. This is an extract from http://www.opendkim.org/opendkim-genkey.8.html:
> (−−bits=n) Specifies the size of the key, in bits, to be generated. The **default is 1024** which is the value recommended by the DKIM specification.

We should be fine with that dkim-key and namecheap (most likely most dns-services).

youaresofunny:
@mwlczk it is not only namecheap problem, got same at yandex and reg.ru

PR #868, opened 2018-02-28
Dkim key size

mwlczk:
the keysize needs to be provided as the first parameter of the command generate-dkim-config
e.g.: `generate-dkim-config 1024` for a 1024 key size
default (or left empty) is set to 2048.
Should help as a quick workaround for #855 and #786 
cheers


mwlczk:
If this is merged, need to update the wiki aswell

mwlczk:
@johansmitsnl great job of closing all the issues, cudos.
I was implementing the TODOs i stated above, but got stuck at tests, because the bats-process is not allowed to read the dkim-keys (mail.txt). I am gonna use a less nice evaluation method of the size of the file instead of evaluating the key via openssl.
When I am done (soon hopefully) I push here again ;)

Issue #839, opened 2018-02-13
Delivery monitoring


zzvara:
Hi,

Please suggest a tool that would monitor delivery from/to this mail server. Unfortunately many services fail within the container, that builds up a Postfix queue. Currently these problems are identified from the logs. Are there any open-source monitoring tools, that the author or others would recommend? It would be a great addition to this stack.

Cheers,
Zoltán

akmet:
Hey,
you mean something like https://linux.die.net/man/1/pflogsumm ?
I am running this as a cron every night and it sends me an email what happend on my mail server

zzvara:
Seems like a useful tool, I will check it out. Are you using this with `tomav:mailserver`?

johansmitsnl:
@akmet could you suggest a PR for this?

akmet:
I am using it currently on my host but I think it would be much better when it is inside the docker. Over the next week I should have enough time for that! :)

What do you guys would prefer:
An env variable like DAILY_SUMM and then every night a summery is send to postmaster or a SUMM_USER mail account? Using cron with @daily and output yesterdays Summery?

akmet:
@johansmitsnl I could not found any mail client. Wasn't it just not needed in the past or is there an agreement to use other methods? I have not worked this much with mail servers in the past and maybe just dont really understand enough whats going on 😆

johansmitsnl:
You could add to the `apt-install` the package: `bsd-mailx`
Somewhere here:
https://github.com/tomav/docker-mailserver/blob/253c8faef61c3544ddb006c291105b02dd8febc3/Dockerfile#L20-L22

akmet:
@johansmitsnl could you close this?

PR #881, opened 2018-03-07
Adding daily mail review from Issue 839

akmet:
I added the daily summery email as requested in #839.
I have not added any tests, because I never worked with this kind of tests and does not really know how to make them and what exact test for. Hope someone can help here?

Currently the script analyzes the /var/log/mail/mail.log and I think logrotate will cause a loss in information when it gets executed. One way to solve this problem could be running logrotate every day at midnight (what I am using in production system) and analyze the mail.log.1 which represents the day before.
Another way could be to analyze the combination of mail.log.1 and mail.log but this just doesn't feels right. Any thoughts?


target/bin/dailysummery
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+test -x /usr/sbin/pflogsumm || exit 0
+test -x /usr/bin/bsd-mailx || exit 0
+
+/usr/sbin/pflogsumm /var/log/mail/mail.log -d yesterday --problems-first | /usr/bin/bsd-mailx -s "Daily Summery for HOSTNAME"  DAILY_SUMM_EMAIL
```
johansmitsnl: Should the variable be: `$DAILY_SUMM_EMAIL` ?

target/start-mailserver.sh
```diff
@@ -1064,6 +1065,20 @@ function _setup_elk_forwarder() {
 		> /etc/filebeat/filebeat.yml
 }
 
+function _setup_daily_summery() {
+	notify 'task' 'Setting up daily summery'
+
+	DAILY_SUMM_EMAIL=${DAILY_SUMM_EMAIL:="0"}
```
johansmitsnl: The default should be set at the start of the script.

target/start-mailserver.sh
```diff
@@ -1064,6 +1065,20 @@ function _setup_elk_forwarder() {
...
+	if [[ ! "$DAILY_SUMM_EMAIL" == 0 ]]; then
+		notify 'inf' "Enable daily mail summery with recipient $DAILY_SUMM_EMAIL"
+
+		sed -i -r 's/HOSTNAME/'$HOSTNAME'/g' /usr/local/bin/dailysummery
+		sed -i -r 's/DAILY_SUMM_EMAIL/'$DAILY_SUMM_EMAIL'/g' /usr/local/bin/dailysummery
+
+		echo "/usr/local/bin/dailysummery" > /etc/cron.daily/dailysummery
```
johansmitsnl: In the daily directory should contain executable scripts?
Just to confirm.

target/bin/dailysummery
akmet: I got problems with sed when used $. Removing it fixed the error..

Starbix:
It'd be great if you could change `summery` to `summary`

johansmitsnl:
@17Halbe could you assist with the tests?

target/bin/dailysummery
17Halbe: Even if you encapsulate the variable with `"`?

17Halbe:
Hey, what a great idea!
I'll be busy the next few days, but I'll give it a try.
What about sendmail instead of bsd-mailx wasn't it working properly?
I don't know pflogsumm, but if we already introduce a new env variable, what about letting the user decide between daily/weekly/monthly(?) mails?
The tests would need to confirm, that a mail is being sent by cron, right? That shouldn't be a problem.

Dockerfile
```diff
@@ -21,6 +21,7 @@ RUN apt-get update -q --fix-missing && \
     amavisd-new \
     arj \
     binutils \
+    bsd-mailx \
```
17Halbe: Isn't sendmail working for this?

akmet: I wasn't able to do so. But maybe it was just my fault.
Could you provide an example on how to use sendmail? Is there something already existing here I could adapt?

17Halbe: https://github.com/tomav/docker-mailserver/blob/a73692cc9fb98849bb0eac72311f4966ecaaca41/test/tests.bats#L1368
I haven't tried using it, but the tests seem to work if one is using it like this.. 
I don't like the way sendmail works either. It's different on every system, sometimes a symlink to some other mail service and then there are about a couple different implementations, all with different syntaxes... :roll_eyes:

akmet:
@17Halbe Just had a look at manpage
It can just analyze today, yesterday or the complete log file.
But we could do this with logrotation and after each rotation we just analyze the old log and send an email.
What do you think?

17Halbe:
I think that would be overkill. ;) In that case, I think it‘s not worth the effort. 
Unless you like to do it for fun and practice! :) (Isn‘t that the reason we are all doing this anyway?)

akmet:
When using in production we have to edit logrotate anyway because if not we are skipping one day as I mentioned above..

akmet:
Travis failed cause because the postfix-accounts.cf file keeps removing while the container runs.
I can reproduce this on my desktop (win10 with docker running on Windows and Ubuntu Bash) and I don't understand why. Can someone reproduce this?

johansmitsnl:
Is the file gone or is it empty?

akmet:
Like both?
Windows Explorer shows the file, but I cannot open it ("File does not exist. Create File?") and I cannot write to it. The file only disappears when restarting the Docker Daemon.
I have no idea why this happens and I think that I then had to use my notebook for changes and git.

johansmitsnl:
I don't know why that is, maybe change de filesystem of docker?

Could you also add a test by calling the function and verify the email has been send and it contains the correct subject and to address?

17Halbe:
I‘ll take care of the test. Probably today.

akmet:
Me neither. But I installed Ubuntu next to Windows and started with a fresh master and now it seems to run.
I would also test if the interval is correctly set and if the logrotate config is set with defaults when no variable is set.

target/bin/dailysummery
akmet: Used the following line and it does not replace the variable:
sed -i -r 's/$HOSTNAME/'$HOSTNAME'/g' /usr/local/bin/postfix-summary
and
sed -i -r 's/\$HOSTNAME/'$HOSTNAME'/g' /usr/local/bin/postfix-summary

target/start-mailserver.sh
akmet: due to using logrotate this is obsolet

johansmitsnl:
Good work!

Issue #843, opened 2018-02-15
Check of cert.pem in Let's Encrypt


dolanor:
Configuring my letsencrypt for my container, I found this.
https://github.com/tomav/docker-mailserver/blob/b08c9b42ed2a211784499a3adbc46d4a80ad4b71/target/start-mailserver.sh#L731

As I was using letsencrypt certificates generated by my [caddy](http://caddyserver.com/), it stored it differently. It only had the fullchain and the key. No simple `cert.pem`.
Then I had to investigate how to make the container letsencrypt configuration work. And this line made me wonder why we check for `cert.pem` existence, since we don't use it afterwards: both postfix and dovecot use `fullchain.pem`.

For info, I made it work for my use case with caddy by volume mounting.
```
    volumes:
      - ${HOME}/.caddy/acme/acme-v01.api.letsencrypt.org/sites/mail.mydomain.org/mail.mydomain.org.crt:/etc/letsencrypt/live/mail.mydomain.org/fullchain.pem
      - ${HOME}/.caddy/acme/acme-v01.api.letsencrypt.org/sites/mail.mydomain.org/mail.mydomain.org.crt:/etc/letsencrypt/live/mail.mydomain.org/cert.pem
      - ${HOME}/.caddy/acme/acme-v01.api.letsencrypt.org/sites/mail.mydomain.org/mail.mydomain.org.key:/etc/letsencrypt/live/mail.mydomain.org/privkey.pem

```

johansmitsnl:
Check this section of the wiki: https://github.com/tomav/docker-mailserver/wiki/Configure-SSL#custom-certificate-files I think it will answer your question.

dolanor:
You're right, even if I use let's encrypt, it's not what you recommended (certbot), therefore, it is as foreign to let's encrypt as a custom certificate.
Still, my question about the verification of `cert.pem` in the script still holds, since it is not used/referenced by the rest of the script.

johansmitsnl:
The cert.pem is indeed not used with Letsencrypt, I guess this is some copy/paste when adding Letsencrypt support in the past.

johansmitsnl:
Can you provide a PR to fix it?

PR #882, opened 2018-03-07
Removed unneeded check for Let's encrypt cert.pem

akmet:
As in #843 I have removed the check for Let's encrypts cert.pem


dolanor:
Thank you for taking the time for it.

Issue #887, opened 2018-03-08
FEATURE REQUEST: Adding /var/lib/dovecot to mailstate


baxerus:
The folder /var/lib/dovecot should be added to the mailstate, so that ssl-parameters.dat doesn't have to be recreated everytime the service is recreated by docker-compose.

The folder has only to be added target/start-mailserver.sh

Best regards

Benedict

johansmitsnl:
Could you add a PR?

baxerus:
Have done so 😉 
A one liner

PR #890, opened 2018-03-09
Add /var/lib/dovecot to mailstate persistence

baxerus:
Added `/var/lib/dovecot/` to the list of folders that get persisted in `mailstate`. So the creation of `ssl-parameters.dat` has not to be done every restart again. This may only take some dozen seconds, but can be very long on systems with high load.

### DIFF
--- a/target/start-mailserver.sh
+++ b/target/start-mailserver.sh
@@ -1175,7 +1175,7 @@ function _misc_save_states() {
 	statedir=/var/mail-state
 	if [ "$ONE_DIR" = 1 -a -d $statedir ]; then
 		notify 'inf' "Consolidating all state onto $statedir"
-		for d in /var/spool/postfix /var/lib/postfix /var/lib/amavis /var/lib/clamav /var/lib/spamassassin /var/lib/fail2ban /var/lib/postgrey; do
+		for d in /var/spool/postfix /var/lib/postfix /var/lib/amavis /var/lib/clamav /var/lib/spamassassin /var/lib/fail2ban /var/lib/postgrey /var/lib/dovecot; do
 			dest=$statedir/`echo $d | sed -e 's/.var.//; s/\//-/g'`
 			if [ -d $dest ]; then
 				notify 'inf' "  Destination $dest exists, linking $d to it"


baxerus:
Closes #887

mwlczk:
objection, maybe it is wise to generate the SSL/TLS - params from scratch once in a while. 
We should look into it, and find out if we need the same cron-job like renewal as with DH provided by @17Halbe

johansmitsnl:
@mwlczk dovecot does regeneration by default:

https://dovecot.org/list/dovecot/2015-May/100928.html
> When Dovecot starts up for the first time, it generates new 512bit and  
1024bit Diffie Hellman parameters and saves them into  
<prefix>/var/lib/dovecot/ssl-parameters.dat. After the initial  
creation they're by default regenerated every week.

baxerus:
No, automatic generation is disabled in the configuration already by setting ssl_parameters_regenerate = 0 (try: `docker exec -it mail doveconf | grep ssl_parameters_regenerate`).

Also even the dovecot developer think nowadays, that the regeneration wasn't a very good idea: "Dovecot v2.1.x and older regenerated them every week by default, but because the extra security gained by the regeneration is quite small, Dovecot v2.2 disabled the regeneration feature completely." (see: https://wiki.dovecot.org/SSL/DovecotConfiguration under SSL security settings).

I'm pro security, but only if it is useful. And, only because the `ssl-parameters.dat` is persisted across restarts doesn't mean, that it is still possible to configure your system in a way, that the file is regenerated every whatever hours. But I think that is better, than slowing down every restart (especially when setting up a system and trying settings and setups and so).

akmet:
But after what amount of time would you suggest to regenerate the ssl parameters?

baxerus:
Actually I do not know. And it seems like the dovecot developers do not recommend to do so anymore (otherwise they wouldn't "disabled the regeneration feature completely" as they write in the wiki).

As always it will depend on your needed level of security, if you feel like all secret services of the world are trying to hack your mail server, you should probably regenerate them quite often (and use more than the standard 2048 length). But than there are much more important things to make secure before dovecots SSL encryption.
On normal security level installations I would guess that it would be a good approach to regenerate the DH parameters every time you change the SSL certificates. You touch one part of the SSL anyway, so resetup everything. Easy to remember, easy to do and most likely often enough.

BUT again as a disclaimer, I am no security expert and I do not run a high value site. The things above are only my thought! You have to decide how big your need for security is by yourself (but than also question yourself if downloading a ready solution from github without a complete code review is a clever thing to do 😉).

akmet:
In my opinion we should not automatically regenerate the ssl parameters by default. I believe the dovecot team does not just drop a security feature when it has some enhancement. So we could just provide a wikipage for the more security-concerned people. Or when someone wants to create an enviromental variable to enable automatical regeneration if needed.

Issue #878, opened 2018-03-06
Correct way of removing unused accounts


vedtam:
Hi,

Thanks for this wonderful docker image.

I would like to remove some of the unused accounts and wondering if all I have to do is to remove the account from the `postfix-accounts.cf` and log into the container to delete its folder from `/var/mail/myaccount.com` as well?

Or is there maybe another way?
Thanks.

17Halbe:
If you got aliases set up you could also delete those in the respective file. Except that, your actions would be enough, yes.
You could also use the [setup script](https://github.com/tomav/docker-mailserver/wiki/Setup-docker-mailserver-using-the-script-setup.sh) but as I just realized this wouldn't delete the maildir itself nor the aliases but only remove the user.
Probable Enhancement?! Would you like to contribute?

vedtam:
Hi @17Halbe,

Thanks for the suggestion. My bash knowledge is rather scarce, but I would like to help and contribute if I can.

17Halbe:
No problem, thanks for the idea! :+1:

PR #897, opened 2018-03-13
Delmailuser

17Halbe:
solves #878

- `delmailuser` is now capable of deleting multiple accounts at once, prompting for confirmation to delete the maildir as well. An additional `-y` flag was introduced to assume yes on this question.
I had to modify most of the shell scripts which are modifying the postfix-accounts and postfix-virtual databases due to the modified delmailuser script.
- `updatemailuser` was just deleting and readding the user as a new one. This wasn't possible with the new deletion script anymore.
- `addalias, delalias` was changed to reflect the correct implementation of multiple recipients for one alias (`,` as saparator only, actually that was making it much easier for me, getting the sed commands do what I wanted them to do! ;) )

### DIFF
--- a/target/bin/addalias
+++ b/target/bin/addalias
@@ -6,7 +6,7 @@ EMAIL="$1"
 RECIPIENT="$2"
 
 usage() {
-	echo "Usage: addalias <user@domain> <recipient@other>"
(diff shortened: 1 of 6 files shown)


17Halbe:
Tests did pass on my machine, could you give it another try?

PR #898, opened 2018-03-13
Log errors of letsencrypt on startup

akmet:
Added error messages to letsencrypt on startup so when some files are not found we get an output directing us in the right direction.

I think this should also be added to other setup functions soon.


akmet:
The tests doesn't failed on my computer and there is nothing that could fail so yeah :)

johansmitsnl:
Can you update the submodule to the latest release? It reverts the master release.

akmet:
Oh, sorry. Thought I had done it but looks like I forgot to update them..

PR #907, opened 2018-03-18
test/config backup & restore

17Halbe:
Adapted the config folder backup strategy to the test folder.
Should be safer regarding future implementations.


johansmitsnl:
Test failed due to permission/missing files, could you verify this?

17Halbe:
tests succeeded on my setup. I do run the tests as root though. 
A conflict may occur if someone sets the repository up as a unprivileged user and then runs the tests as root.
In that case the config and test/config folders would change their ownership to root.
That being said. Why not preserving the ownership and permissions when copying?! ..Hold my beer! ;)

17Halbe:
So I ended up introducing a backup rule. If the copy process is run during fixtures, some files appear to be locked and are unable of being read. fixtures wasn't the right place to do the backup anyway, since files already got modified till the backup(fixtures rule) happened. 
Should have checked more thoroughly, and not rely on the copying being done on the right place and time. Sorry.

PR #914, opened 2018-03-20
Password creation fix

### LINKED ISSUE #908: Password with spaces is stripped to first word
With ./setup.sh you cannot save password with spaces.
Bug is with creation of new email and update. Both will save only first word as password.

## Expected Behavior

Saved password will be same as I write to command.

## Actual Behavior

If password contains space, only first word is saved as password.

## Steps to Reproduce

1. `./setup.sh email update somemail@example.com "super secrete password"`
2. Try login with "super secrete password", it doesn't work
3. Login will works only with password "super"

## Your Environment

* Mailserver version used: Latest, with lastest setup.sh from master
* Docker version used: 17.05.0-ce, build 89658be

17Halbe:
closes #908


17Halbe:
The same problem with the tests as on my side. They actually succeed rarely. I don't know why. There is no problem if I run the commands manually.

test/tests.bats
```diff
@@ -1243,27 +1260,12 @@ load 'test_helper/bats-assert/load'
...
-  run grep "test1@example.org" ./config/postfix-virtual.cf
-  assert_failure
-
-  run grep "test2@example.org" ./config/postfix-virtual.cf
+  run cat ./config/postfix-virtual.cf | wc -l | grep 0
   assert_success
-
-  ./setup.sh -c mail alias del test2@example.org test1@forward.com
-  run grep "test2@example.org" ./config/postfix-virtual.cf
-  assert_failure
```
johansmitsnl: Why are these lines removed?

17Halbe: Thanks for reviewing. I have no idea, I was working on too branches at the same time.
:+1:

johansmitsnl:
Just to confirm, why is this file being removed? `test/config/dovecot-lmtp/userdb`

akmet:
Think this file gets deleted by make clean. There was another PR by @17halbe last week adding this file after it gets deleted accidentially by the mail summary PR..

17Halbe:
Yepp I did, and I was wrong. TL;DR: It can be deleted.

The `userdb` got introduced here #461, became obsolete with #784 and @akmet wisely removed it in #901. I wasn't so wisely and thought it got accidentally deleted. My sincere apologies!

17Halbe:
However I'm really not happy with those tests. They fail to frequently and I don't know why. Anyone got an idea? 
Or shall I just leave them out? Though the creation of the password hash is never being tested and I would like to see that being tested.

johansmitsnl:
@17Halbe I restarted the job several times, do you know why it fails?

17Halbe:
\o/ 

But it‘s a dirty fix. It sleeps for 30(!) seconds for either of those two tests. :/ 
I‘ll leave it like this. I could probably move at least the addmailuser command to the Makefile.
I plan on refactoring the tests anyway in the near future. What do you think?

johansmitsnl:
The sleep is needed indeed to allow the watcher to pickup the changed file.
Sounds good to me.

PR #919, opened 2018-03-28
add headers to postfix summary mail

andihit:
this PR adds the `To`, `Content-Transfer-Encoding` and `Content-Type` headers to the postfix summary mails


andihit:
not sure why the travis tests fail, they work locally
but the one test affected by this change also works on travis (`ok 160 checking pflogsum delivery`)

PR #926, opened 2018-03-31
Implementation of multi-domain relay hosts, fixes #922

### LINKED ISSUE #922: Multi-domain Mailgun configuration
The AWS SES settings work well for a single domain using [Mailgun](https://www.mailgun.com/) to relay email. However for a multi-domain setup it isn't quite enough, as we want to authenticate as different users depending on the sender's domain.

The configuration is described well in [this forum post][1].

From looking at the code, it looks like this will need code adding in `start-mailserver.sh#_setup_dovecot_local_user()` and in `check-for-changes,sh` and a new function along the lines of `start-mailserver.sh#_setup_postfix_relay_amazon_ses()`

I'm happy to have a go at this so here's my plan:
* maintain another file postfix-sasl-password.cf to provide the multiple username passwords
* add extra env variables:
  * `MULTI_DOMAIN_RELAY_HOST`
  * `MULTI_DOMAIN_RELAY_PORT`
  * `MULTI_DOMAIN_RELAY_DEFAULT_USER`
  * `MULTI_DOMAIN_RELAY_DEFAULT_PASSWORD`
* update `relayhost_map` file at startup and on changes
* (bonus) add a function to setup.sh which creates/updates `postfix-sasl-password.cf`

A couple of thoughts/questions:
1. Relay passwords will be stored in plain text in `config/postfix-sasl-password.cf` (I guess it's not really worse than the current solution which needs the password in the docker-compose.yml or in .env)
2. `MULTI_DOMAIN_RELAY_HOST` and `AWS_SES_HOST` can't both be true - can this merely be documented in README?

[1]: https://community.rackspace.com/products/f/email-products-forum/3897/how-to-setup-postfix-with-a-mailgun-smtp-relay-when-using-multiple-domains

crazystick:
The implementation is pretty much as per the description in #922 

There is a bit of additional flexibility vs. what I planned - you can add relay maps to postfix-relaymap.cf which allows for per-domain relay hosts as well as per-domain authentication. I have tested that successfully having one domain send via Mailgun and another send via Sendgrid.

This implementation is pretty convenient for my use-case - however it did occur to me that we could require users to build the whole configuration in those two config files, and skip building `/etc/postfix/relayhost_map` on-the-fly from mail accounts. That would reduce the amount of code somewhat.

I didn't spot any tests for AWS_SES env variables, and so I'm not sure quite how this should be tested. Any advice would be appreciated.


target/bin/addrelayhost
```diff
@@ -0,0 +1,33 @@
...
+
+if [ -z "$PORT" ]; then
+	PORT=25
+fi
+
+if grep -qi "^@$DOMAIN" $DATABASE 2>/dev/null; then
+	sed -i "s ^@"$DOMAIN".* "@$DOMAIN"\t\t["$HOST"]:"$PORT" " $DATABASE
+else
+	echo -e "@$DOMAIN\t\t[$HOST]:$PORT" >> $DATABASE
+fi
```
johansmitsnl: Could you add a new line

target/bin/addsaslpassword
```diff
@@ -0,0 +1,35 @@
...
+	read -s -p "Enter Password: " PASSWD
+	echo
+	[ -z "$PASSWD" ] && errex "Password must not be empty"
+fi
+
+if grep -qi "^@$DOMAIN" $DATABASE 2>/dev/null; then
+	sed -i "s ^@"$DOMAIN".* "@$DOMAIN"\t\t"$USER":"$PASSWD" " $DATABASE
+else
+	echo -e "@$DOMAIN\t\t$USER|$PASSWD" >> $DATABASE
+fi
```
johansmitsnl: Same here

README.md
```diff
@@ -538,3 +538,25 @@ Note: This postgrey setting needs `ENABLE_POSTGREY=1`
...
+
+#### RELAY_USER
+
+  - **empty** => no default
+  - default relay username (if no specific entry exists in postfix-sasl-password.cf)
+
+#### RELAY_PASSWORD
+
+  - **empty** => no default
+  - password for default relay user
```
johansmitsnl: Can you use the env and the setup script independently or dos the env needs to be set?

crazystick: as it stands, at least the RELAY_HOST env variable is required, and the config files add functionality on top (the env variables alone basically work like the AWS_SES_* vars used to work). It would be possible to drop the env and do it all by config files, but then we have to keep the AWS_SES code for backwards compatibiilty. I believe that unauthenticated relays are possible (although dangerous) so I think it's valid to have relay user/password be optional.

PR #835, opened 2018-02-12
 Introduced auto. spam learning & global sieve filter

17Halbe:
#### Introduced auto-spam-learning via sieve. 
Spam is automatically trained to spamassassin, when moved via **imap-only** to the spam folder.
And so is "Ham" when it's moved out of the Spam folder.

Additionally two new optional config files have been introduced:
- global-before.dovecot.sieve
- global-after.dovecot.sieve

The former being a global sieve file which is going to be executed before the user specific rules are applied and the later, you might have guessed it, is applied, after the user specific rules have been executed.

**Attention:**

This has not been tested with actual Spam, Ham movement. We need a testcase or at least someone to test it before merging, alone I don't know how to programmatically move mails between two imap-folders. Anyone else has a hint? Or can try this out?

Tests are passing on my machine.
closes  #384

PR #934, opened 2018-04-04
 Introducing global filters.

17Halbe:
On my way to getting spam handling done. This was part of PR #835 but should be handled as a separate PR I think. 
This introduces the optional global sieve files `after.dovecot.sieve`/`before.dovecot.sieve` which apply to **every** incoming email. They are executed before and after any user specific `{user}.dovecot.sieve` filter would be applied.

Wiki entry will follow once accepted.

- added optional after.dovecot.sieve/before.dovecot.sieve files
- added global filter test

### DIFF
--- a/target/dovecot/90-sieve.conf
+++ b/target/dovecot/90-sieve.conf
@@ -71,7 +71,7 @@ plugin {
   # (http://pigeonhole.dovecot.org) for available plugins.
   # The sieve_extprograms plugin is included in this release.
   #sieve_plugins =
-  sieve_plugins = sieve_extprograms sieve_imapsieve
+  sieve_plugins = sieve_extprograms
 
   # The separator that is expected between the :user and :detail
   # address parts introduced by the subaddress extension. This may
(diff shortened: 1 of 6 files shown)


17Halbe:
tests are succeeding on my end. (Except the `checking setup.sh: setup.sh email add` test, which shouldn't be affected)

target/dovecot/90-sieve.conf
```diff
@@ -31,14 +31,14 @@ plugin {
   # executed. The order of execution within a directory is determined by the
   # file names, using a normal 8bit per-character comparison. Multiple script
   # file or directory paths can be specified by appending an increasing number.
-  #sieve_before =
+  sieve_before = /usr/lib/dovecot/sieve-global/before.dovecot.sieve
```
johansmitsnl: Does this fail when no file it copied in the start-mailserver.sh?

johansmitsnl:
About test 130 something broke this test, you have any idea why?

17Halbe:
This is the test from the #914 PR, where I already had my problems with it. I ended up inserting a 30 sec sleep before commencing the test. It was enough to get it running once. 
I'll move the `addmailuser` command to the `Makefile`.

target/dovecot/90-sieve.conf
17Halbe: Yes, I was afraid of that as well. I couldn't find anything in the Docs though(It just states, that it's safe to not have any files in a directory when configuring a directory as a target) and just ran the tests with it. Dovecot wasn't complaining(the `after.dovecot.sieve` file doesn't exist during the tests) so I basically forgot about it. 
Do you think we should take the safer route and comment those lines out, when not used?

17Halbe:
Yeah, I'm just too slow in answering it! ;)

target/dovecot/90-sieve.conf
johansmitsnl: I guess it is safer indeed, if you could change it I can merge the request.

17Halbe: Let me check again..

johansmitsnl:
Thanks!

Issue #950, opened 2018-04-15
CI is not reliable (?)


callmemagnus:
Still while investigating my broken builds, I see that the builds running on Travis are unreliable.

I believe that this is related to the amount of processing needed to start the containers completely. I see that this has probably already been an issue in the past when I see all the "sleep" commands present in the Makefile and in some tests.

The following screen show illustrates it at best:
![image](https://user-images.githubusercontent.com/232478/38782930-897199d6-40fb-11e8-8861-fc43aaaf8544.png)
Source: https://travis-ci.org/callmemagnus/docker-mailserver/branches

Both branches are the same. "Patch-1" is the source of my PR and its build did not trigger on your repo (?).

I am pretty convinced that the result of the builds are dependent on the availability and power of Travis build slave.

johansmitsnl:
What you see is true. They are sometimes unreliable depending on the time the build is created.
However when you investigate the errors it is mostly test 6 that fails.
This can be improved by detecting if the images are running and available and then send the test data for processing and then again wait if this data is processed.

Any suggestions?

callmemagnus:
I didn't look into all the details of the build.
As far as I can see it is the `dovecot/ssl-param` that requires a lot of CPU time, with current make I have 3-4 at the same time on my laptop.

It seems that the container is able to receive the test data but not to process it. Which breaks some tests (?)

Open questions:

# Are all the containers needed for the tests?
# Are all the containers to be started upfront? Would it be possible to do start image run specific tests for that image, then go to next tests.

I agree that the detection of "readyness" is probably a good idea. Maybe embed a script in the image that would tell based on several criteria whether the "boot" process is finished for a given container.

I see these "checks":

* presence of ssl-param in `ps` is a no-go
* `netstat` should list all listening ports
* maybe do a local "telnet" to see the response of a given service (?)

johansmitsnl:
Now there a clear steps that it goes trough for each combination.

1. build
2. start the different configurations
3. Put in test data for each configuration
4. Test if it worked correctly

This can be separated in several tests #945 can have a advantage in development time and also in test stability.

With `nc` you can also probe ports. But the SSL generation for Dovecot starts to listen on the port and accepts a connection but keeps it in hold until it is ready with the SSL generation.
Not sure how to test if it is fully running.

callmemagnus:
I'm trying something like this in the Makefile just after the list of docker run...

```
#!/bin/bash

function is_container_ready () {
  local container=$1
[...]

Issue #945, opened 2018-04-10
Separate test cases


17Halbe:
## Context
With the increasing amount of test cases testing is becoming more and more the bottleneck in development. Even a fast cached (`make all-fast`) build on a current PC currently takes around 17 minutes to complete.

My proposal would be to split the tests into separate test files according to the containers they are executed from. 
The user could then decide to execute all containers and therefore all tests or just the particular container and its associated tests.
The `Makefile` would be getting a new optional argument `CONTAINER='container1 container2'`

Furthermore I would like to introduce a optional argument `PERF=(int)` setting, to respect different performance capabilities of different setups. 
The setting would basically be a factor applied to any `sleep` command in the `Makefile`. 

## Pros
- shorter debugging times(Performance factor and selective testing)
- smaller and therefore more readable test files
- more granular debugging
- less ressource heavy. Containers would be removed after their tests exited successfully.
- downward compatible

## Cons
- not one single searchable tests.bats file

johansmitsnl:
Sounds good and it is indeed an issue.

We could also take advantage of Parallelizing the builds/tests https://docs.travis-ci.com/user/speeding-up-the-build/ ?
The issue is that the docker container needs to be build and building on each node does not sounds efficient. Any ideas on this?

mwlczk:
What about splitting the big container/services into a "composition" of postfix, dovecot and antivir as proposed in  #539. We could have many "service"-specific tests, and some "global" tests at the end.

17Halbe:
@johansmitsnl That's a great idea, I'm not really familiar with Travis though. But as we would already separate the tests, we could easily perform the tests directly after container creation in parallel. I'll have a look into parallelizing Travis.
If I'm not totally wrong about this, we should be able to deny multithreading to the build target and then create the containers all based on the built test image.
@mwlczk This is an interesting idea, but doesn't really have to do with testing, right? Sounds like a huge undertaking though! ;)

johansmitsnl:
@mwlczk there is already a ticket about this: #539 take a look at the conversation there.

crazystick:
It would be great if tests would not break each other. For example, the mail container is used by a bunch of tests, and adding a new test which sends an additional email breaks earlier tests that count emails, and debugging is hard because yet another test rotates the log file and so there's no data to find out why something failed earlier on. Maybe each test should have a separate file which also does setup/teardown?

17Halbe:
[...]

PR #955, opened 2018-04-19
Tests separation

### LINKED ISSUE #949: "email del" does not remove data directory, even if it says so
## Context
While investigating my failing test on my PR (https://github.com/tomav/docker-mailserver/pull/884), I discovered that one feature is not working as expected and its test fails to provide guard.

When running `./setup -c mail email del a.user@server.com`, there is an option to delete the data folder.

Sadly when that option is selected (-y or y at run time), it does not delete the directory.

## Expected Behavior

The user should be removed from the account file and if I wish the maildir folder should be deleted too.

## Actual Behavior

Mail folder is not deleted.

## Possible Fix

The script delmailuser is ran from setup.sh in a docker "run" command. This command only mount the configuration directory and not the persistence of the mail folder so obviously the "normal" server will not have its maildir (inside the container or on the host filesystem) deleted as it only tries to delete a folder in a docker run once container.

## Steps to Reproduce

1. launch the setup for tests "mail, mail-xx"
2. add a user through the script
3. delete a user through the script
4. check the existence of the folder on the main "mail" container

## The impact

The test that is supposedly validating this feature does not its job and fails to "see" the issues.

I will comment that test in my PR as it does not work as expected (when I finish)

17Halbe:
**This is more a test for Travis than the final product.** 

- introduced multithreaded container startup
- separated test cases according to the container they are run on
- Introduced optional `PERF` option for `make`. (The higher the PERF number, the lower the sleep time in the Makefile)
- Introduced optional `CONTAINER` option for `make` to decrease development and testing time. Specific Container can be chosen to be tested. It is provided as a space separated list of container names. List of all Container:`CONTAINER=from_image mail mail_privacy mail_pop3 mail_smtponly mail_smtponly_without_config mail_override_hostname mail_fail2ban mail_fetchmail mail_disabled_clamav_spamassassin mail_manual_ssl mail_with_ldap mail_postscreen mail_lmtp_ip mail_with_postgrey mail_undef_spam_subject mail_with_relays mail_with_imap`
- added a `clean` prerequisite prior building to the main/default targets. One would want to get everything reset before commencing a new test.
- cleaned up obsolete setup.sh arguments (-c CONTAINER for functions running via the image)
- made some tests less flexible by replacing the dynamic image name lookup with a fixed value, to make them independent from the running mail container)
- Fixes userdel bug (closes #949)

The number of tests didn't change:
screenshot from 2018-04-19 09-34-53

Right now the tests aren't multithreaded. There are quite a lot of problems regarding processor load, timing and displaying the results. 
`tests.bats` is still included for reference. It might look a bit unordered, but it's the final version I separated the tests from.
See #945
Concerns #950


17Halbe:
Ok this seems to work.
Todo: 
- Parallizing tests
- Separating fixtures into container specific targets
- Creating Test/Container specific config directories
- What else?

Makefile
```diff
@@ -218,47 +240,76 @@ run:
...
-	docker exec mail /bin/sh -c "nc 0.0.0.0 25 < /tmp/docker-mailserver-test/email-templates/existing-user-and-cc-local-alias.txt"
-	docker exec mail /bin/sh -c "nc 0.0.0.0 25 < /tmp/docker-mailserver-test/email-templates/existing-regexp-alias-external.txt"
-	docker exec mail /bin/sh -c "nc 0.0.0.0 25 < /tmp/docker-mailserver-test/email-templates/existing-regexp-alias-local.txt"
-	docker exec mail /bin/sh -c "nc 0.0.0.0 25 < /tmp/docker-mailserver-test/email-templates/existing-catchall-local.txt"
-	docker exec mail /bin/sh -c "nc 0.0.0.0 25 < /tmp/docker-mailserver-test/email-templates/sieve-spam-folder.txt"
-	docker exec mail /bin/sh -c "nc 0.0.0.0 25 < /tmp/docker-mailserver-test/email-templates/sieve-pipe.txt"
-	docker exec mail /bin/sh -c "nc 0.0.0.0 25 < /tmp/docker-mailserver-test/email-templates/non-existing-user.txt"
-	docker exec mail_disabled_clamav_spamassassin /bin/sh -c "nc 0.0.0.0 25 < /tmp/docker-mailserver-test/email-templates/existing-user1.txt"
+	$(eval NEEDSSLEEP=false)
+  ifneq (,$(filter mail,$(CONTAINER)))
```
crazystick: Just wondering, why not use the same fixtures_<container> pattern like you use for build_<container>?

Makefile
```diff
@@ -218,47 +240,76 @@ run:
...
+  ifneq (,$(filter mail_privacy,$(CONTAINER)))
+		docker exec mail_privacy /bin/sh -c "openssl s_client -quiet -starttls smtp -connect 0.0.0.0:587 < /tmp/docker-mailserver-test/email-templates/send-privacy-email.txt"
+		$(eval NEEDSSLEEP=true)
+  endif
+  ifneq (,$(filter mail_override_hostname,$(CONTAINER)))
+		docker exec mail_override_hostname /bin/sh -c "nc 0.0.0.0 25 < /tmp/docker-mailserver-test/email-templates/existing-user1.txt"
+		$(eval NEEDSSLEEP=true)
+  endif
+  # Wait for mails to be analyzed
+	@if [ $(NEEDSSLEEP) = true ]; then \
```
crazystick: Sleeps should just be part of the fixture with duration defined by the test author - if I make one fixture to deliver one email, I don't need to sleep for 120s if I'm just running my own fixture.

Makefile
crazystick: Ah - is that your TODO item 2?

Makefile
17Halbe: You are right and that‘s on the todo list.

17Halbe:
Ok, did some more testing, but really don't find the time right now. Unfortunately the tests are as or even less reliable than before. 
Those sleep commands are a real problem. Their usefulness really depends on the available resources of the underlying hardware. 
So how do we figure out, if a container is up and running?
- finding its pid and checking for its resource usage,
- greping the log for a "up'n'running" message from the container(after startup and after test-message processing.(Probably via [postqueue](http://www.postfix.org/postqueue.1.html)?))

Does anyone has a bit more experience and ideas regarding that issue?

johansmitsnl:
@17Halbe the issue is indeed how to detect if a container is running.
Even when you add to the log "running" dovecot can still be generating ssl keys or whatever the services do when they are started from the main script until they are ready.
Maybe you can open a tcp connection and check for certain output in a while loop of Postfix and Dovecot.

If this works then send the test data into the container, wait for processing (post queue sounds good) and then start the actual tests.

17Halbe:
Sorry for the "long time no see". I was sick and away for it. I won't be able to contribute for some time. Sorry. Feel free to close this one.

erik-wramner:
While this is outdated the changes as such would be extremely valuable. I think we should split it into smaller parts and get them into the master branch one change at a time. I'll leave this open for now.

erik-wramner:
Closing as similar work is under way in other PRs.

PR #963, opened 2018-04-25
Possibility to disable authentication for relay hosts

mariusburfey:
I prepared a pull request for option 2 in this issue: https://github.com/tomav/docker-mailserver/issues/962

With the changes you can disable authentication for relay hosts by setting `smtp_sasl_auth_enable` in postfix.


mariusburfey:
I found a much simpler solution:
Do not set `RELAY_HOST` but override `relayhost = relay.example.com` in `postfix-main.cf`...

The result is the same.

PR #965, opened 2018-04-30
Report sender

ixeft:
I added the possibility to specify which address is in the From field of the summary generated by pflogsum


ixeft:
Hmm, the test were passing by my side.

ixeft:
Tests does not fail on my computer, and tests unrelated to my pull request are failing on travis, is that possible that there is a problem with it ?and to launch again the checks ?

akmet:
I did not checked the travis result but sometimes when there are to much builds resulting in some failures

PR #970, opened 2018-05-03
fix SASL domain fixes #892

### LINKED ISSUE #892: Spoof failure with valid account and sender address
@17Halbe introduced in #872 spoof protection but this fails with ldap accounts.

```
 Mar 11 09:30:44 mail postfix/submission/smtpd[1874]: NOQUEUE: reject: RCPT from unknown[172.16.38.1]: 553 5.7.1 <johan@domain.net>: Sender address rejected: not owned by user username@mail.domain.net; from=<johan@domain.net> to=<johan@other.net> proto=ESMTP helo=<macbook>
```

The email server has `mail.domain.net` as the hostname and the email address is johan@domain.net. It  verifies with my username and mail host domain to verify the sender address.

This is incorrect because it should check if the email address matches what is in the ldap as the `mail` field or `mailalias` fields.

n00dl3:
setting value as `$myhostname` will make sasl look for users `user@mail.domain.tld` instead of `user@domain.tld`

### DIFF
--- a/target/postfix/main.cf
+++ b/target/postfix/main.cf
@@ -72,7 +72,7 @@ smtpd_sasl_path = /var/spool/postfix/private/auth
 smtpd_sasl_type = dovecot
 
 smtpd_sasl_security_options = noanonymous
-smtpd_sasl_local_domain = $myhostname
+smtpd_sasl_local_domain = $mydomain
 broken_sasl_auth_clients = yes
 
 # Mail directory


johansmitsnl:
Reading the Postfix manual:

```
Default authentication domain
Postfix can append a domain name (or any other string) to a SASL login name that does not have a domain part, e.g. "john" instead of "john@example.com":

/etc/postfix/main.cf:
    smtpd_sasl_local_domain = example.com
This is useful as a default setting and safety net for misconfigured clients, or during a migration to an authentication method/backend that requires an authentication REALM or domain name, before all SMTP clients are configured to send such information.
```

It should only have effect when no domain is given and this is indeed in my case the issue.

Restarted Travis.

Issue #979, opened 2018-05-24
TLS options for LDAP


tailtwo:
## Context
In order to use this container with my LDAP server, I have to use TLS with SASLauth and Dovecot. Unfortunately, I don't see any envvar that would allow me to set the desired options.

## Expected Behavior
`- SASLAUTHD_LDAP_START_TLS=yes` set `ldap_start_tls: yes` in `/etc/saslauthd.conf`
`- SASLAUTHD_LDAP_TLS_CHECK_PEER=no` set `ldap_tls_check_peer: no` in `/etc/saslauthd.conf`
`- DOVECOT_TLS_REQUIRE_CERT=never` set `tls_require_cert = never` in `/etc/dovecot/dovecot-ldap.conf.ext`

Generally, I would like to be able to use more ldap envvars without mounting a file in the container (I'm using Kubernetes and it's not very practical), like those :
https://blog.sys4.de/cyrus-sasl-saslauthdconf-man-page-en.html
http://ftp.ics.uci.edu/pub/centos0/ics-custom-build/BUILD/dovecot-1.1.7/doc/dovecot-ldap-example.conf

## Possible Fix
Add more envvars to configure LDAP for dovecot and salsauth.

johansmitsnl:
It is merged

PR #980, opened 2018-05-25
Add saslauthd option for ldap_start_tls & ldap_tls_check_peer

fkefer:
I needed start_tls myself and also found that it solves issue #979 

Not entirely sure about the tls_check_peer - [documentation](https://github.com/winlibs/cyrus-sasl/blob/master/saslauthd/LDAP_SASLAUTHD) says you "you must specify ldap_tls_cacert_file or ldap_tls_cacert_dir" - unless I guess if CA is already known to the system


fkefer:
Just to be clear about the "Not entirely sure" part of the comment: tls_check_peer works perfectly fine with default "no", and it works also fine with "yes" when the Cert is e.g. from letsencrypt (as in my case) and sure other accepted CAs. the only thing I'm not sure about is whether the other options should be also provided - or if users with self-signed certs should rather fall back to "no"...

PR #989, opened 2018-06-13
DOMAINNAME can fail to be set in postsrsd-wrapper.sh

jamebus:
if the container doesn’t have a proper hostname, postsrsd will fail to start
because SRS_DOMAIN is empty. Make a best effort to figure out the domain name
and provide a way to set one if needed.


johansmitsnl:
Could you add a test?

jamebus:
Yes. That's a great idea. I'll give it a go.

PR #1011, opened 2018-07-25
Sqlite as storage

jrd:
I find your docker which is exactly what I was looking for.
Unfortunately, until now, postfixadmin cannot directly read postfix config files, but relies on a database. As I like it, I added a way to store some postfix config in sqlite database (so no server running).

The database is mainly used for reading, rarely for writing so that shouldn't be a problem.

I've read the guidelines, and found that 2 spaces were required in files, so I did a first commit to clean some files that uses tabs or 4 spaces or both ;-)

The real addition is in the second commit. It includes some tests with `bats`.

P.S. I also bumped the compose file to version 3 as I think version 2 is quite old now. Feel free to remove this is you think version 2 should be kept as example.


jrd:
The travis build failed on ONE test: 

> checking setup.sh: email add login validation

I don't understand why!? Could it be a sync/time problem?

johansmitsnl:
Could you undo this addition: `# vim: et ai si ts=8 sts=8 sw=8` and other files that have only whitespace changes?

I restarted the test to see if it was a false positive.

jrd:
Yes I can remove the commit which changes spaces/tabs.

I will propose it in another, independent, pull request to be clean.
What is the willing whitespace standard you want to use for docker-mailserver? tabs? spaces? If spaces, how many? Currently some files are a mix of spaces/tabes depending on who did some modifications.

I will also rebase my commit to master tip (and hence resolve changelog conflict)

ekozan:
any new on this one :)

johansmitsnl:
Could you not change the changelog (that is only done on release).

Also could you resolve the conflict?

erik-wramner:
@jrd nothing has happened for a very long time and by now there are massive merge conflicts. What do you want to do about this one?

If you want to get it in I would suggest starting fresh from the current master and just adding the required changes in a new PR. I'm ambivalent about it. I think it is great in some ways, but at the same time it dramatically increases the number of test cases (need to test with and without sqlite as there are so many if statements) and script complexity.

Either way I'd like to close this old PR as it seems to be stuck. What do you think?

erik-wramner:
Closing due to inaction.

PR #1032, opened 2018-09-10
Fix allow sending emails from regexp aliases when spoof protection is enabled

Birkenstab:
## How to reproduce
* Set `SPOOF_PROTECTION=1`
* Create a alias in `postfix-regexp.cf`: `/^test[0-9]@domain.tld/ mail@domain.tld`
* Try to send an email as user mail@domain.tld with sender address `test1@domain.tld`
=> Mail gets rejected with error message:
`postfix/smtps/smtpd[1312]: NOQUEUE: reject: RCPT from xxxxx[xxx.xxx.xxx.xxx]: 553 5.7.1 <test1@domain.tld>: Sender address rejected: not owned by user mail@domain.tld; from=<test1@domain.tld> to=<receiver@example.com> proto=ESMTP helo=<xxxxxx>`

After adding `/etc/postfix/regexp` to `smtpd_sender_login_maps` it worked as expected.

For LDAP this probably has to be fixed in a similar way.

Before I discovered this bug I set a alias for a whole domain in `postfix-virtual.cf`: `@domain.tld mail@domain.tld`. This worked for receiving mails as expected but not for sending mails. The sending was rejected with same error as above. Although the file `/etc/postfix/virtual` is included in the `smtpd_sender_login_maps` so this might be a bug in Postfix?

## Sidenote
When I was trying to build this project, I first got a error message from docker that it couldn't find a file. Until I realized that there is git submodule I hadn't cloned yet. I suggest adding it to the Readme (maybe create a build section with additional infos?) in case someone stumbles upon the same problem.

### DIFF
--- a/target/start-mailserver.sh
+++ b/target/start-mailserver.sh
@@ -658,7 +658,7 @@ function _setup_spoof_protection () {
 	sed -i 's|smtpd_sender_restrictions =|smtpd_sender_restrictions = reject_authenticated_sender_login_mismatch,|' /etc/postfix/main.cf
 	[ "$ENABLE_LDAP" = 1 ] \
 		&& postconf -e "smtpd_sender_login_maps=ldap:/etc/postfix/ldap-users.cf ldap:/etc/postfix/ldap-aliases.cf ldap:/etc/postfix/ldap-groups.cf" \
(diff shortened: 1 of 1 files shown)


johansmitsnl:
Thanks for the contribution!

Please add a PR for the readme and I'll accept it.

markusstephanides:
Good work, thanks!

PR #1036, opened 2018-09-20
Update start-mailserver.sh

djoq:
Possible resolutions to bugs like relay access denied and host loops back to myself.

Helps in solving issues mentioned here: https://github.com/tomav/docker-mailserver/issues/805#issuecomment-361462181 and here: https://github.com/tomav/docker-mailserver/issues/794#issuecomment-378059130


johansmitsnl:
Some critical tests are failing, could you verify them?

tomav:
`HOSTNAME` (fqdn) and `DOMAINNAME` have different purpose.
The reason there are 2 different variables si because they can have 2 different values.

PR #1043, opened 2018-10-01
Restart Postfix and Dovecot when SSL files change

mathuin:
This is a slightly different approach to the problem for which #976 is intended to solve.

The major differences:
* A new script `target/watch-ssl-files.sh` was created to handle the situation, instead of extending `target/check-for-updates.sh`.
* Tests for the script's functions are included in the script, and are called by a test in the `test./tests.bats` file.

The primary benefit of a second script is that `target/check-for-updates.sh` currently runs when LDAP is not in use, so LDAP users who use SSL cannot take advantage of functionality added to this script.  The largest risk of a second script is "battling restarts", where one script restarts Postfix while another is attempting to do the same.

The tests are for the script's functions -- to generate a set of checksums for files and test those files for changes -- and not on the script's actual behavior (to restart the services when those files change), for simplicity's sake.  

If desired, I am 100% willing to merge the new script into the existing script.  It will require some general cleanup of the original script, but the end result would be a single task -- with tests -- checking on whether certain files have changed and taking the appropriate action.  Just ask, and I'll take care of it!

There are some additional inadvertent changes in this pull request as a result of Atom's efforts to remove trailing white space and white space from blank lines.  As this brings the code further into compliance with the style guide in the documentation, I decided to leave them in.  I can revert those changes if desired.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -209,7 +209,7 @@ RUN curl -s https://letsencrypt.org/certs/lets-encrypt-x3-cross-signed.pem > /et
 
 COPY ./target/bin /usr/local/bin
 # Start-mailserver script
-COPY ./target/check-for-changes.sh ./target/start-mailserver.sh ./target/fail2ban-wrapper.sh ./target/postfix-wrapper.sh ./target/postsrsd-wrapper.sh ./target/docker-configomat/configomat.sh /usr/local/bin/
+COPY ./target/watch-ssl-files.sh ./target/check-for-changes.sh ./target/start-mailserver.sh ./target/fail2ban-wrapper.sh ./target/postfix-wrapper.sh ./target/postsrsd-wrapper.sh ./target/docker-configomat/configomat.sh /usr/local/bin/
 RUN chmod +x /usr/local/bin/*
 
 # Configure supervisor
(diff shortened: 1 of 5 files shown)


mathuin:
There's something non-deterministic about the tests here.  I noticed it on the other PR but blamed my home network's NFS mount -- but I doubt Travis has that problem.  Here's my Travis: https://travis-ci.org/mathuin/docker-mailserver/builds It doesn't show that the first run failed, only that the second one succeeded.  The failed tests are unrelated to the code changes, which is really frustrating.  On the off chance that the change is related to the "battling reset" issue somehow, I've merged the new script into the old script and cleaned up the old script.  All tests pass here at home, so I'm pushing it to this PR branch.  If the cleanup is excessive, let me know!

mathuin:
I am not sure why the tests are failing, as all the tests pass at home.  I'm running this branch in production at home, and will update this ticket if there are any failures.

Is there reason to believe that the tests can be trusted on Travis?  Looking over the history of builds, there are many many more failures than successes.

johansmitsnl:
Restarted the build, do test all pass on your setup?
Normally test 6 fails but all others pass, but in this case email delivery fails so that is critical.

mathuin:
193 tests, 0 failures -- I can pass along the typescript if you'd like.

vaceslav:
any updates for this PR?

Issue #964, opened 2018-04-28
unable to create lock file /var/mail/mail.lock: Permission denied


jirislav:
There is a bug in the currently latest version.

## Context
When anyone sends me an email, I got the error as I pasted it into the title, which suggests there is permission collision when updating received mail.

Good thing is that the sender is notified correctly about the inability of receiving messages.

## Expected Behavior
The incoming email should be saved.

## Actual Behavior
The incoming email is not saved

## Possible Fix
Permissions (/var/mail/mail owns `5000:5000` with mod of `0600`, but user `mail` tries to write there, which implies this error)

## Steps to Reproduce
1. Start latest docker-mailserver
2. Choose working mailserver & send a test message from it to the latest docker-mailserver
3. See logs of the docker-mailserver, you should see there the message

sleif:
in my case "target/start-mailserver.sh: function _fix_var_mail_permissions()" breaks postfix and dovecot.
from LDAP the real uidNumber and gidNumber will be used by postfix und dovecot. So both processes run into trouble if all ownership in /var/mail belongs to uid 5000.

jirislav:
I've found out the solution:
- postfix tries to write into `/var/mail/mail` only when it doesn't use virtual transport (it is Off by default) - perhaps it should be On by default?
  - you can enable this by setting `ENABLE_POSTFIX_VIRTUAL_TRANSPORT=1` to env while starting the Docker
- you also have to configure proper dovecot socket path so that the virtual transport functions properly:
  - `POSTFIX_DAGENT=lmtp:unix:/var/run/dovecot/lmtp`

Only then the rules generated within the `/etc/postfix/vmailbox` will be respected resulting in no need to write to `/var/mail/mail`, but into the appropriate mailbox instead.

**I'm not closing this issue as the default docker behavior is technically broken unless you specify all of those ENVs mentioned above**
 - someone should make the virtual transport the default

aendeavor:
This issue was closed due to one or more of the following reasons:

1. Age
2. Contributor inactivity
3. The issue seems to be resolved

If you think this happened by accident, or feel like this issue was not actually resolved, please feel free to re-open it. If there is an issue you could resolve in the meantime, please open a PR based on the current master branch so we can review it.

PR #1051, opened 2018-10-09
fix: /var/mail permissions are hardcoded to 5000 #964

jirislav:
On April 29th I've issued [a bug](https://github.com/tomav/docker-mailserver/issues/964) with file permissions in `/var/mail`, but no one paid attention to this. That's why I've created this PR.


jirislav:
In the end, I've found that this wasn't the problem.

PR #1056, opened 2018-10-15
feat: added postfix message & mailbox size limits to ENV settings #629

jirislav:
See discussion at #629 .


techgourmet:
Finally!!!!!

PR #1070, opened 2018-11-02
add LDAP_QUERY_FILTER_DOMAIN env to documentation

mpanneck:
Currently the LDAP_QUERY_FILTER_DOMAIN env var is used in [Makefile](https://github.com/tomav/docker-mailserver/blob/master/Makefile#L175) and [start-mailserver.sh](https://github.com/tomav/docker-mailserver/blob/master/target/start-mailserver.sh#L594).
This is documented nowhere.
This PR adds the env var to documentation.


johansmitsnl:
Thanks for the PR! :+1:

PR #1072, opened 2018-11-03
Allow regex aliases

shimunn:
Since it like to use a unique email address for each service I use and I've got tiered on creating a new account each time I thought it would be a good idea to support regex aliases via postfix.


johansmitsnl:
Many tests fail now specific to email delivery. Could you verify/fix this?

shimunn:
I don't think that's caused by my modifications, tests for master and stable branches are failing as well.

johansmitsnl:
Restarted the build

erik-wramner:
Nothing has happened since November 2018. If this is still relevant, please rebase from the current master where the tests are green to see if that helps. If there is no activity for two weeks I think we should close this PR.

erik-wramner:
Closing for inactivity.

PR #1074, opened 2018-11-04
Calling supervisord directly instead of via shell

### LINKED ISSUE #1047: Graceful stop of container
`docker stop` does not stop the container gracefully. Instead after 10s (default) it gets SIGKILL and simply removed from memory by kernel.
This is because `/bin/sh` is PID 1 in the container which does not have an explicit handler for SIGTERM nor does it forward the signal to its child processes.

Is it for any reason I do not know? Is it for any reason better to SIGKILL the container?

## Expected Behavior
Graceful stop of all processes in container.

## Actual Behavior
Sudden kill of all processes no matter in which state they are.

## Possible Fix
I would suggest something like [dumb-init](https://github.com/Yelp/dumb-init) as ENTRYPOINT.
I can create a pull request with a proper PID 1 process (dumb-init, tini, pid1) optimized for container use.
As a side effect it may be possible to route SIGHUP to postfix master to reload changed config without restart.

## Steps to Reproduce
`docker stop` your container and see exit code 137. (128 + 9[SIGKILL])
or
`docker exec -ti YOUR_CONTAINER_NAME ps -ef` and you see PID 1 is `/bin/sh ...`

danielpanteleit:
This fixes #1047 by calling supervisor directly. The container won't fail now anymore, because there is no way to tell supervisor to exit with an error, so I had to change the test that expects the container to fail.

I recommend increasing the stopping time, though. Since the container can take longer than 10 seconds to stop gracefully. This can be done using "stop_grace_period" in docker-compose.yml or "--stop-timeout" using "docker run" (also docker stop -t)

### DIFF
--- a/test/tests.bats
+++ b/test/tests.bats
@@ -6,7 +6,7 @@ load 'test_helper/bats-assert/load'
 
 @test "checking configuration: hostname/domainname" {
   run docker run `docker inspect --format '{{ .Config.Image }}' mail`
-  assert_failure
+  assert_success
 }
 
 @test "checking configuration: hostname/domainname override" {
@@ -1783,5 +1783,14 @@ load 'test_helper/bats-assert/load'
 
 @test "checking that mail for root was delivered" {
   run docker exec mail grep "Subject: Root Test Message" /var/mail/localhost.localdomain/user1/new/ -R
+  assert_success
+}
+
+#
+# clean exit
+#
+
+@test "checking that the container stops cleanly" {
+  run docker stop -t 60 mail
   assert_success
 }
--- a/Dockerfile
+++ b/Dockerfile
@@ -219,6 +219,6 @@ COPY target/supervisor/conf.d/* /etc/supervisor/conf.d/
 
 EXPOSE 25 587 143 465 993 110 995 4190
 
-CMD supervisord -c /etc/supervisor/supervisord.conf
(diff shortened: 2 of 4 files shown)


test/tests.bats
```diff
@@ -6,7 +6,7 @@ load 'test_helper/bats-assert/load'
 
 @test "checking configuration: hostname/domainname" {
   run docker run `docker inspect --format '{{ .Config.Image }}' mail`
-  assert_failure
+  assert_success
```
johansmitsnl: Why did you change this test? It does not seems to be related to your change?

danielpanteleit: Supervisor is now the main process (PID 1) and we can't stop it in the start-mailserver.sh script in a way that would result in a failure. Supervisor is not designed to stop with an error if one of its processes is failing.

So there are basically two options:
1. stopping supervisor gracefully or
2. using a different init process.

I think 1) should be ok, since the container only failed if the setup was failing (due to configuration errors). For correctly configured containers this would not matter.

For option 2) we could either (a) use docker's built-in init process (docker run --init; or docker-compose option "init: true") or (b) add an external init program (like tini) and kill supervisor non-gracefully.

Option 2a) could be problematic for already deployed instances, since the docker-compose files (or docker run commands) need to be changed. I did not check out options for 2b)

PR #1104, opened 2019-01-06
Support for setting relayhost in main.cf

jjtt:
I'm running docker-mailserver for a single domain with ~30 email accounts and all of them forwarding all messages to other email providers. The current RELAY_HOST setting provides much more functionality than I needed and it seems to require that RELAY_USER and RELAY_PASSWORD are also set. However my ISP hasn't provided me with a user and password for the SMTP server and accepts (like most of them do?) SMTP connections from local addresses without authentication.

So the global default relay host setting in main.cf seemed like the simplest solution for my use case.


crazystick:
I agree that overall the final configuration is simpler, but I assume that you did just try setting the existing RELAY_HOST and RELAY_PORT settings? For your single domain you should just end up with an `/etc/postfix/relayhost_map` containing one line `@your.domain      [mail.yourisp.com]:25` which I would expect to work. You'll get a warning "No relay auth file found and no default set" at startup which you could safely ignore. Have to confess I didn't test this though.

johansmitsnl:
Is this PR ready?

jjtt:
I could fix the typo and tidy up the documentation since, as @crazystick noticed, this probably doesn't interfere with RELAY_HOST after all. Other than that, I have nothing more in mind. This setup is in use on my tiny domain.

I did try the RELAY_HOST without user and password and didn't manage to get it to work. But now that I think about it, the reason could have been something completely different. I'll try to find the time to test it again.

PR #1124, opened 2019-02-05
Fix port descriptions

501st-alpha1:
Port 465 is for implicit SSL/TLS connections, and port 587 is for StartTLS, so I fixed those descriptions.

Ports 110, 143, and 587 are actually plain text by default, with the option to upgrade to StartTLS, so I noted that as well.

(I haven't tested *all* of these myself to verify, but am taking this from what I know about email ports.  See e.g. [SSL vs TLS vs STARTTLS](https://www.fastmail.com/help/technical/ssltlsstarttls.html).)

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,10 +103,10 @@ Your config folder will be mounted in `/tmp/docker-mailserver/`. To understand h
 #### Exposed ports
 * 25 receiving email from other mailservers
 * 465 TLS/SSL Client email submission
-* 587 StartTLS Client email submission
-* 143 StartTLS IMAP client
+* 587 Plaintext / StartTLS Client email submission
+* 143 Plaintext / StartTLS IMAP client
 * 993 TLS/SSL IMAP client
-* 110 POP3 client
+* 110 Plaintext / StartTLS POP3 client
 * 995 TLS/SSL POP3 client
 
 `Note: Port 25 is only for receiving email from other mailservers and not for submitting email. You need to use port 465 or 587 for this.`
--- a/README.md
+++ b/README.md
@@ -102,8 +102,8 @@ Your config folder will be mounted in `/tmp/docker-mailserver/`. To understand h
 
 #### Exposed ports
 * 25 receiving email from other mailservers
-* 465 SSL Client email submission
-* 587 TLS Client email submission
+* 465 TLS/SSL Client email submission
+* 587 StartTLS Client email submission
 * 143 StartTLS IMAP client
 * 993 TLS/SSL IMAP client
 * 110 POP3 client


erik-wramner:
I'm not sure that the descriptions are correct. For port 587 I think the server insists on getting a STARTTLS, so it is not a plain-text port at all. I don't think it is possible to submit e-mail messages on port 587 in plain text. On the other hand I'm pretty sure that both 110 and 143 are plain-text even though they might support STARTTLS (not sure). Personally I would not expose them as there is a clear risk that a client will send user name and password in the clear.

501st-alpha1:
> For port 587 I think the server insists on getting a STARTTLS, so it is not a plain-text port at all. I don't think it is possible to submit e-mail messages on port 587 in plain text. On the other hand I'm pretty sure that both 110 and 143 are plain-text even though they might support STARTTLS (not sure).

Okay, that makes sense; I didn't really test plaintext or STARTTLS since I only need implicit TLS.  I can test these if you'd like.

> Personally I would not expose them as there is a clear risk that a client will send user name and password in the clear.

I agree; I disable plaintext ports on my installs.

erik-wramner:
Please do; if we document the ports the documentation should be correct.

erik-wramner:
This has been hanging for a long time. Shall we close it, or would you like to verify the above and get it in?

polarathene:
Ello, looked into this(but not yet tested).

[Dovecot default configuration](https://doc.dovecot.org/configuration_manual/quick_configuration/#authentication) is to prevent supporting plaintext auth/usage, STARTTLS is enforced to establish a secure connection. [`docker-mailserver` Dovecot config](https://github.com/tomav/docker-mailserver/blob/master/target/dovecot/10-auth.conf) doesn't appeear to deviate from that.

Regarding SSL/TLS, [the docs state](https://doc.dovecot.org/configuration_manual/dovecot_ssl_configuration/#dovecot-ssl-configuration):

> `ssl=yes` and `disable_plaintext_auth=yes`: SSL/TLS is offered to the client, but **the client isn’t required to use it**. The client isn’t allowed to use plaintext authentication, unless SSL/TLS is enabled first. However, if non-plaintext authentication mechanisms are enabled **they are still allowed even without SSL/TLS**.

> Depending on how secure they are, the authentication is either fully secure or **it could have some ways for it to be attacked**.

> `ssl=required`: **SSL/TLS is always required**, even if non-plaintext authentication mechanisms are used. Any attempt to authenticate before SSL/TLS is enabled will cause an authentication failure.

[This projects related `ssl` dovecot config](https://github.com/tomav/docker-mailserver/blob/master/target/dovecot/10-ssl.conf) is following upstream default of `yes`, which is presumably fine, since non-plaintext authentication isn't enabled by default:

> If you have only plaintext mechanisms enabled (e.g. `auth { mechanisms = plain login }` ), `ssl=yes` and `ssl=required` are completely equivalent because in either case the authentication will fail unless SSL/TLS is enabled first.

That is followed up by a word of caution that regardless of `yes` or `required, lack of conformance of the protocol can result in leaking credentials:

> With both `ssl=yes` and `ssl=required` it’s still possible that the client attempts to do a plaintext authentication before enabling SSL/TLS, which exposes the plaintext password to the internet.

There's further clarification that `ssl=required` encrypts the entire connection, while `disable_plaintext_auth=yes` only encrypts the auth exchange?(unclear if `ssl=yes` is equivalent for plaintext auth here, presumably it is).

And then advice of best practice:

> Nowadays you most likely should be using SSL/TLS anyway for the entire connection, since the cost of SSL/TLS is cheap enough. Using both SSL/TLS and non-plaintext authentication would be the ideal situation since it protects the plaintext password even against man-in-the-middle attacks.

Finally there is mention of exceptions where those settings don't apply, and the connection is *assumed* secure because it's on localhost, or the remote IP is the same as the host, there's also trusted IPs(which I assume can be anything on the internal docker network).

Meanwhile, the supported [auth mechanisms are `PLAIN` and `LOGIN`](https://github.com/tomav/docker-mailserver/blob/master/target/dovecot/10-auth.conf#L100), [Dovecot docs point out that `LOGIN` is obsolete](https://doc.dovecot.org/configuration_manual/authentication/):

> The LOGIN mechanism is obsolete, but still used by old Outlooks and some Microsoft phones.

Further down there is a section about usage with a proxy, where you might be enforcing TLS connections to the server at a different layer and terminating them before reaching dovecot, in which case changing the settings like below can treat it as one would with HTTPS(443) being routed to HTTP(80) in reverse proxies?:

> `ssl = no`
> `disable_plaintext_auth = no`

There is some more coverage about the `auth_mechanisms` option covered in [dovecot docs](https://doc.dovecot.org/configuration_manual/authentication/authentication_mechanisms/). General advice as best practice these days is to enforce TLS which makes plaintext auth safe([but should be cautioned with STARTTLS](https://doc.dovecot.org/admin_manual/ssl/)). Note @tomav expressed [*wanting to tackle this since back in 2016*](https://github.com/tomav/docker-mailserver/issues/332#issuecomment-255347167),

- `login` should be dropped(if still valid for a minority, could be referenced in docs for enabling).
- `plain` should probably be kept, STARTTLS will be enforced, but has potential to be insecure outside of `docker-mailserver` with non-conforming clients. Docs should have examples with the port omitted/commented out and advise that implicit TLS ports should be preferred.

---

For testing, you can check ports 25, 110, 143, 587 with STARTTLS [as described here](https://serverfault.com/a/926495):

```
openssl s_client -starttls smtp -connect mail.example.com:25
openssl s_client -starttls smtp -connect mail.example.com:587
openssl s_client -starttls imap -connect mail.example.com:143
```

Ports 110, 143, 587 *should enforce STARTTLS*, **port 25 is required to make STARTTLS support optional** when exposed on the host to allow inbound mail from the public internet. Not that everyone follows that, especially with MTA-STS/DANE?

---

## TL;DR

> For port 587 I think the server insists on getting a STARTTLS, so it is not a plain-text port at all. 

[`docker-mailserver` enforces TLS](https://github.com/tomav/docker-mailserver/blob/master/target/postfix/master.cf#L17), but this is not required when supporting 587 on an SMTP server. 

Do note the port is still plain-text to begin with, even with enforcement of TLS for transit, prior to establishing the TLS connection all data exchanged is in plain-text. Some clients are known to insecurely send credentials prior to TLS being established.

> I don't think it is possible to submit e-mail messages on port 587 in plain text. 

It is. 587 is a submissions port, STARTTLS enforcing TLS is optional, but supplying credentials is required.

> I'm pretty sure that both 110 and 143 are plain-text even though they might support STARTTLS (not sure).

These ports were originally plain-text only prior to STARTTLS which provides opt-in TLS support. Dovecot by default(and this project) [is configured to enforce TLS](https://github.com/tomav/docker-mailserver/blob/master/target/dovecot/10-auth.conf#L10) for 110 and 143(failure to negotiate TLS rejects connection).

This can still be insecure prior to establishing TLS connection, thus implicit TLS ports 993 and 995 are encouraged instead. Same applies for 587.

polarathene:
How about the following, or is the associated notes too long for the README?

#### Exposed ports

| Protocol | Opt-in Encryption<sup>1</sup> | Forced Encryption | Purpose    |
|----------|---------------------|-------------------|------------|
| SMTP     | 25                  | N/A               | Relay<sup>2</sup>    |
| ESMTP    | 587                 | 465<sup>3</sup>             | Submission |
| POP3     | 110                 | 995               | Retrieval  |
| IMAP4    | 143                 | 993               | Retrieval  |

1. Originally plain-text only, STARTTLS when supported on both ends of a connection will upgrade to a secured TLS connection. This can be enforced instead of optional, rejecting the connection if it cannot be established securely. Enforcement in `docker-mailserver` is default for ports 110, 143, 587. Port 25 when exposed to the internet is expected to be opt-in, as not all legitimate email received supports securing the connection.
2. Receives email from external servers on the internet. Many hosting providers and ISPs block port 25(outbound) as it requires no authentication, risking misconfigured servers to become an [Open Relay](https://en.wikipedia.org/wiki/Open_mail_relay) for spam. It is required to be exposed to the public internet if you intend to receive email to an inbox. Sending outbound email in `docker-mailserver` by default is restricted to your submission ports(465, 587), see [**PERMIT_DOCKER**](https://github.com/tomav/docker-mailserver/blob/master/README.md#permit_docker) ENV for relaxing that.
3. Since 2018 with RFC 8314, port 465 is a submission port(requires authentication), not a secure variant of port 25.

NOTE: While STARTTLS can establish secure connections, doing so requires conformance between both parties which is not always guaranteed. A third-party may [intercept and perform a downgrade attack](https://it.slashdot.org/story/14/11/11/2349244/isps-removing-their-customers-email-encryption) preventing a secure connection that would otherwise be supported, when STARTTLS is enforced it would prevent the email from being sent.

Where possible, it is encouraged to prefer ports 465, 993, 995 for implicit TLS to secure connections. If you do not need ports 25, 110, 143, do not map them to the host ports. They will still be available within the internal docker network for your own containerized services to use.

erik-wramner:
I think it is a bit long even though the content is good. Perhaps the table, a short note on best practice and a link to a wiki page?

polarathene:
How about this revision? Would you like me to raise a new PR for it?

The other content can be placed in the wiki(and the wiki link can be updated to an appropriate section for it), but I think these notes are important enough to convey to the user up front when they're viewing this overview on ports.

I know for me, it was fairly confusing how the ports were to be used or what I needed. Especially with conflicting information about port 465 on the internet, or how secure is STARTTLS(and why it exists when there are secure ports etc). Further compounded with docs about reverse proxies being added to the mix, and how STARTTLS doesn't mix well into that and LetsEncrypt.

---

| Protocol | Opt-in Encryption<sup>1</sup> | Forced Encryption | Purpose    |
|----------|---------------------|-------------------|------------|
| SMTP     | 25                  | N/A               | Relay<sup>2</sup>    |
| ESMTP    | 587                 | 465<sup>3</sup>             | Submission |
| POP3     | 110                 | 995               | Retrieval  |
| IMAP4    | 143                 | 993               | Retrieval  |

1. A connection *may* be secured over TLS when both ends support `STARTTLS`. On ports 110, 143 and 587, `docker-mailserver` will reject a connection that cannot be secured. Port 25 is [required](https://serverfault.com/questions/623692/is-it-still-wrong-to-require-starttls-on-incoming-smtp-messages) to support insecure connections.
2. Receives email and filters for spam and viruses. You should prefer the submission ports(which require authentication) to submit outgoing email. Unless a relay host is configured, outgoing email will leave the server via port 25(thus outbound must not be blocked by your provider).
3. Submission port since 2018, [RFC 8314](https://tools.ietf.org/html/rfc8314). Originally a secure variant of port 25.

See the [wiki](https://github.com/tomav/docker-mailserver/wiki) for further details and best practice advice, especially regarding security concerns.

erik-wramner:
Looks good, please raise a new PR.

501st-alpha1:
Apologies for the delay on my part, this got buried on my todo list.

Thanks @polarathene for all the research and your revision to the port info!  I agree that the table looks much better, so I'll go ahead and close this PR.

PR #1159, opened 2019-04-24
Add Kubernetes helm chart

funkypenguin:
Hey guys,

This looks like a scary-big PR, I know, but you'll note that aside from some additional .travis.yml tests, all the changes are to a new directory called /helm-chart/docker-mailserver.

What is this?
--
This is a helm chart for docker-mailserver - there's a comprehensive README.md included, at https://github.com/funkypenguin/docker-mailserver/blob/add-helm-chart/helm-chart/docker-mailserver/README.md

The chart has been carefully tested (both manually and with [CI](https://www.funkypenguin.co.nz/note/continuous-integration-for-helm-charts/)), and I believe it's ready for general use.

Does CI pass?
--
Well, no. But the CI tests for the _chart_ pass, the existing CI for docker-mailserver itself is currently failing, which seems to be the case prior to this PR anyway (see logs at https://travis-ci.org/funkypenguin/docker-mailserver/jobs/523771460)

I'm happy to take questions / update the wiki as necessary. Using this chart, I'm a happy Kubernetes docker-mailserver user :)

D


erik-wramner:
I think this looks great, but at the same time I believe it would work best as a downstream project using this as upstream rather than as an integrated part. It adds quite a lot of code/complexity and as far as I can see it it can use the upstream docker images more or less as they are? Perhaps making the helm chart a top-level project and linking to it from the docker-mailserver readme would be better? Just my view though. Good work regardless!

funkypenguin:
Thanks @erik-wramner - I have now moved the chart to a standalone repository (https://github.com/funkypenguin/helm-docker-mailserver), and published it at https://hub.helm.sh/charts/funkypenguin/docker-mailserver.

How about a new PR for a simple README update to point to the separate repo, instead?

(And yes, the chart simply refers to the upstream docker images as-is)

erik-wramner:
I think that would be great, but unfortunately I don't have commit rights so @tomav needs to approve.

tomav:
I do agree with @erik-wramner.
We can link this at the end of the README.md in a `related projects` section.
ELK should also be moved out.
I let you open a new MR.
Thanks for this, we'll have more and more K8s users in the future.

PR #1176, opened 2019-06-13
1175: specify user for cron.d freshclam file

erik-wramner:
See issue 1175, cron complains that the freshclam entry in cron.d is invalid as it lacks a user. Fixed by adding root as user.

Note that the unit tests were failing before this, they are still failing after this and I don't really know how to test this with a unit test. A good manual test is to let the cron job run once with and without the fix. Without it there are errors in the logs and the cron job does not run.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,6 +27,7 @@ The development workflow is the following:
 - Code :-)
 - Add integration tests in `test/tests.bats`
 - Use `make` to build image locally and run tests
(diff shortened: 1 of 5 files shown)


erik-wramner:
Merged in the fixed unit tests from my other pull request (dovecot fixes and root alias change script fix) so that the tests are green. It would be nice to get this in?

tomav:
Thanks @erik-wramner

PR #1177, opened 2019-06-13
Fix Dovecot SSL parameters and generate dhparams as for Postfix

erik-wramner:
This pull request combines https://github.com/tomav/docker-mailserver/pull/1149 and https://github.com/tomav/docker-mailserver/pull/710. It fixes the outdated Dovecot configuration options and it generates the dh.pem file the first time the server starts, keeping it in the mail state for future use.

Tests failed before so they are most likely going to fail after as well, unfortunately.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,6 +27,7 @@ The development workflow is the following:
 - Code :-)
 - Add integration tests in `test/tests.bats`
 - Use `make` to build image locally and run tests
+  Note that tests work on Linux only; they hang on Mac and Windows.
 - Document your improvements in `README.md` or Wiki depending on content
 - [Commit](https://help.github.com/articles/closing-issues-via-commit-messages/), push and make a pull-request
 - Pull-request is automatically tested on Travis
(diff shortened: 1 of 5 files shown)


erik-wramner:
Fixed the broken unit tests! The root cause was the script that checks for changes, it broke the /etc/aliases file. Please accept this or add the changes in some other way, it would be great to get a green build again.

erik-wramner:
@tomav (or by all means someone else with commit rights), wouldn't it be nice to get the tests green after 8 months in the red?

martin-schulze-vireso:
@johansmitsnl I second @erik-wramner 's wish to have green tests again. I have little faith in the testing process now and it probably slows down PR acceptance.

Is there anything I can do to facilitate a solution? Should the fix be unbundled from the other commit in this PR so they can land separately?

erik-wramner:
@martin-schulze-vireso the fix for the tests should ideally not be un-bundled because the broken dovecot configuration was one of the errors. It is possible to accept the other fix separately and keep the broken config, but then there will still be broken tests. Just fewer.

martin-schulze-vireso:
@erik-wramner I see, just took a glance. Thanks for clearing that up.

martin-schulze-vireso:
@erik-wramner Please have a look at #1186

PR #1180, opened 2019-06-25
Replace ENV with ARG for DEBIAN_FRONTEND

j-marz:
Best practice suggests not using ENV for this setting as it persists after build. ARG is only set during build.


tomav:
Thanks.

PR #1183, opened 2019-06-29
Fixed self-signed cert generation

j-marz:
Added optional FQDN arguement to setup.sh script which avoids using temporary container hostname for cert names. Also fixed issue with certs being saved outside config volume.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -70,7 +70,7 @@ SUBCOMMANDS:
   config:
 
     $0 config dkim <keysize> (default: 2048)
-    $0 config ssl
+    $0 config ssl <fqdn>
 
   relay:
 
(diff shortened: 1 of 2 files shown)


tomav:
I'm wondering if you should not remove that part. I created this myself when I started the project few years ago. It's not a good usage and now letsencrypt is a standard. Anyone?

j-marz:
> I'm wondering if you should not remove that part.

@tomav which part are you referring to?

> It's not a good usage and now letsencrypt is a standard.

Are you suggesting self-signed is not good practice or this scripts implementation to generate the self-signed cert is not good practice? 

Tweaking the key size and expiry could make it stronger.

tomav:
Yep, I'm thinking about simplify docker-mailserver globally and have natively only best practices.
For now I merge this MR.
Thank you.

PR #1190, opened 2019-07-06
Allow postfix master.cf overrides to start with numbers

tweibert:
Change startup script so that it accepts overrides for `master.cf` that start with a number (not only a character). This is required for changes that modify the `127.0.0.1:10025/inet` service, as e.g.:

    127.0.0.1:10025/inet/smtpd_proxy_filter=

Very simple edit to `start-mailserver.sh`, chaging the regular expression used to filter `postfix-master.cf`.


tomav:
Thanks

PR #1200, opened 2019-07-28
FIX: escape symbols in cron command

dimalo:
Hi @tomav, thanks for the good work!
I've found a bug after inspecting the logs in compose:

```
mail                     | Jul 28 09:51:55 mail postfix/smtpd[2658]: warning: cannot load 1024-bit DH parameters from file /etc/postfix/dhparams.pem -- using compiled-in defaults
mail                     | Jul 28 09:51:55 mail postfix/smtpd[2658]: warning: TLS library problem: error:0906D06C:PEM routines:PEM_read_bio:no start line:../crypto/pem/pem_lib.c:686:Expecting: DH PARAMETERS:
```
`cat /etc/postfix/dhparams.pem` was showing an empty file...

Logs from docker exec -it {container_name: mail} /bin/bash:
```
root@mail:~# cat /etc/cron.d/dh2048
@weekly FILE=/tmp/tmp.aFRqiZNPZT ; openssl dhparam -out  2048 > /dev/null 2>&1 && mv -f  /etc/postfix/dhparams.pem
root@mail:~# echo "@weekly FILE=\`mktemp\` ; openssl dhparam -out \$FILE 2048 > /dev/null 2>&1 && mv -f \$FILE /etc/postfix/dhparams.pem" > /etc/cron.d/dh2048
root@mail:~# cat /etc/cron.d/dh2048
@weekly FILE=`mktemp` ; openssl dhparam -out $FILE 2048 > /dev/null 2>&1 && mv -f $FILE /etc/postfix/dhparams.pem
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -181,7 +181,7 @@ COPY target/postfix/main.cf target/postfix/master.cf /etc/postfix/
 COPY target/postfix/header_checks.pcre target/postfix/sender_header_filter.pcre target/postfix/sender_login_maps.pcre /etc/postfix/maps/
 RUN echo "" > /etc/aliases && \
   openssl dhparam -out /etc/postfix/dhparams.pem 2048 && \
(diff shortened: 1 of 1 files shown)


tomav:
Thank you.

PR #1205, opened 2019-08-03
set Nameservers in opendkim.conf at start-up

j-marz:
Fix for https://github.com/tomav/docker-mailserver/issues/1204


erik-wramner:
Looks good. Have you tried it with a real server?

It would also be nice with a test, see test/tests.bats. You could assert that the config file contains the expected line(s).

j-marz:
> Looks good. Have you tried it with a real server?

Yes, running on my server.

> It would also be nice with a test, see test/tests.bats. You could assert that the config file contains the expected line(s).

I'll try do this tomorrow.

j-marz:
@erik-wramner test added and confirms 8a1584c3cb40aa6ba7c40c4a1c4d8968722ba4a5 is working.

PR #1194, opened 2019-07-23
WIP: fix linting errors reported by hadolint

fbartels:
Hadolint is a nice too to detect common pitfalls in Dockerfiles. If desired I can add it to the travis file as well.

PR #1210, opened 2019-08-06
Make check for connected-networks more dynamic

fbartels:
On my system the creation of the bridge failed since there was an overlap in the address space

continuation of https://github.com/tomav/docker-mailserver/pull/1194

Signed-off-by: Felix Bartels <felix@host-consultants.de>


martin-schulze-vireso:
Okay, that was already anticipated by @erik-wramner in #1194 but as I said in https://github.com/tomav/docker-mailserver/issues/1079#issuecomment-518553879 this will probably not be sufficient. Since the test in its current form did not expose the problem that you experienced in your tests I would propose to enhance the test to try to really send a mail from the second connected network. I already had code for that but then saw the other PERMIT_DOCKER tests only checking the config.

fbartels:
Obsoleted by https://github.com/tomav/docker-mailserver/pull/1215

PR #1211, opened 2019-08-06
Fix linting errors reported by hadolint

fbartels:
Continuation of https://github.com/tomav/docker-mailserver/pull/1194

Signed-off-by: Felix Bartels <felix@host-consultants.de>


fbartels:
@erik-wramner @martin-schulze-vireso I tried to understand the test failure around opendkim, but it does not really make sense to me.

The way I see it the test starts a new container based on the testing tab mounting a folder out of the local dataset and then complains that the mounted data is empty. When running the command locally it succeeds for me. Can one of you shine some light on this?

erik-wramner:
Well, several commands use `ls -1 etc/opendkim` which works in the root folder, but not elsewhere. Perhaps it would help to use `/etc/opendkim` instead? Is this error consistent, or does it come and go?

I'm getting mad at the "can't login as newly added user" error, it always works locally but fails often in Travis... Probably because Postfix and Dovecot have been restarted and are not ready, but I don't know if there is more to it.

fbartels:
No, that error was consistent and while thinking about it I now also know why it happened (but fixed the paths for ls nonetheless). The Dockerfile was using a cd in between and I changed this to switching the workdir (since this is a best practice), but that also mean that after starting the container we were not in /, but in /usr/share/dovecot. 

I have now added another workdir switch locally.

Will now add a linting task to the makefile.

fbartels:
Yay, its finally green!

Lets see if it stays green with the linting part of the testing.

(this pr should be squashed before merging)

Issue #1206, opened 2019-08-04
Improve development by modularizing the tests


martin-schulze-vireso:
Currently the tests are a large monolith which makes their development cumbersome. They should be separated into several lightweight suites that can be called independently.

## Context
Taken from https://github.com/tomav/docker-mailserver/pull/1185#issuecomment-517947172

- [x] separate tests into smaller, independent chunks (e.g. one suite per container)
- [x] ensure cleanup really does clean up
- [x] replace `sleep`s with health checks and timeouts
- [x] add a simple test that ensures the startup of the service was successful before running the testsuite
- [x] ensure that test containers are stopped automatically after their test has run
- [x] fix non deterministic tests (where restarting does change the result)
- [x] reduce interference between containers by separating their config folders

martin-schulze-vireso:
I think the separation into a test suite per container would allow for the setup/teardown functionality of BATS to take that over from the Makefile. However, we still should do the common setup at a central place.

erik-wramner:
See PR #955 (unfortunately a bit outdated and never fully completed).

fbartels:
> would allow for the setup/teardown functionality of BATS to take that over from the Makefile

That sounds like an interesting proposition. Although the Makefile could the still be used for proper orchestration of tasks.

martin-schulze-vireso:
@erik-wramner Thanks, I will have a look at that.

@fbartels 
> the Makefile could the still be used for proper orchestration of tasks.

Yeah, I'd envision something like a pattern rule for .bats files in the test folder, which would allow to run specific tests via `make test/<suite-name>.bats` and the general test target depends on all test/*.bats

The biggest problem that I see right now is that setup and teardown are called per test case, that might be prohibitively slow if we start and stop/remove the container each time. It may be workable if we get the sleep times down as much as possible but I am afraid we will only find out by trying.

martin-schulze-vireso:
@erik-wramner After skimming through the diff and rebasing that onto current master I am a bit at a loss: There are close to 200 tests in the project, the PR seems to be mostly WIP. Since there never was a green state I am not sure if that is the right starting point. I might get some inspiration there but it would probably be much more helpful if @17Halbe could give me some advice on the pitfalls that were encountered so I don't make the same mistakes again. However, since it has been so long since this was done most of that knowledge might already be lost.

The main takeaway from the PR for me is: Relying on the (sleep) timing is very brittle, but getting rid of that is already one of the main goals of this ticket.

[...]

PR #1215, opened 2019-08-07
Test PERMIT_DOCKER=connected-networks by sending an email from second network

martin-schulze-vireso:
also showcase timeouts and makefile integration as needed for #1206 
This should also make #1210 obsolete

At first I just wanted to fix the tests but then I ran against the monolith again so I tried this as a playground for the modularization. The extracted tests for this PR now run in under 20s.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -212,7 +212,7 @@ RUN curl -s https://letsencrypt.org/certs/lets-encrypt-x3-cross-signed.pem > /et
 
 COPY ./target/bin /usr/local/bin
 # Start-mailserver script
-COPY ./target/check-for-changes.sh ./target/start-mailserver.sh ./target/fail2ban-wrapper.sh ./target/postfix-wrapper.sh ./target/postsrsd-wrapper.sh ./target/docker-configomat/configomat.sh /usr/local/bin/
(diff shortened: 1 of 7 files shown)


martin-schulze-vireso:
I also noticed that the code for host just assumes a /16 network. My development setup generates /20 networks, so this is bound to lead to problems. I am not sure if we should leave it as is because changing it might break existing use cases or if we should fix it.

fbartels:
Nice work. I've run the tests once with the old docker container and it failed (as expected) and then did a rebuild and with the next run the test succeeded.

The only comment I would have is that I would still keep the test around (but then now in the new dedicated file) to check mynetworks in the container. I see scenarios where patching of mynetworks will fail, but sending could succeed afterwards.

martin-schulze-vireso:
I don't see a reason why the test should have failed, I only changed code in the new .bats file. Can we restart the test run?

fbartels:
@martin-schulze-vireso just restarted the build

martin-schulze-vireso:
@fbartels thx

erik-wramner:
I restarted the build again, but it fails on the fixtures not on a test step (last time anyway). If it still doesn't work I can try to build it locally.

martin-schulze-vireso:
Now it ran through. I'll add non deterministic tests to the list for #1206

PR #1228, opened 2019-08-14
Add ShellCheck tests and recommendations

j-marz:
This is a WIP. The build will fail until all fixes are applied to shell scripts.

- [x] fail travis build if shellcheck finds issues with shell scripts
- [ ] fix issues found in existing shell scripts

https://github.com/koalaman/shellcheck

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,9 @@ install:
 - travis_retry travis_wait make build-no-cache
 script:
 - make generate-accounts run generate-accounts-after-run fixtures tests
+- shellcheck *.sh
+- shellcheck target/*.sh
+- shellcheck target/bin/*
 after_script:
 - make clean
 notifications:


fbartels:
In another project I am using the following snippet in a Makefile (could be added the the recently added lint target):

```
grep -rIl '^#![[:blank:]]*/bin/\(bash\|sh\|zsh\)' \
        --exclude-dir=.git --exclude=*.sw? \
        | xargs shellcheck -x
```

j-marz:
I like the idea @fbartels - no need to update `.travis.yml` every time a new shell script is added.
I guess we could use that approach, but limit it to the directories that will be expected to host shell scripts related to this project? 
My concern is breaking builds for shell syntax issues found in shell scripts that come with the base image or packages.

fbartels:
> My concern is breaking builds for shell syntax issues found in shell scripts that come with the base image or packages.

as shellcheck is performed in the checked out directory it will check scripts that are then only part of the build docker image. The concern is valid though for files pulled in by git submodules (bats for example).

erik-wramner:
This was a good idea, but it seems to have stalled. Were there too many errors to fix?

j-marz:
> This was a good idea, but it seems to have stalled. Were there too many errors to fix?

Unfortunately, I haven't been able to spend any more time on this PR. Happy for someone else to pick it up and take over. Otherwise i'll aim to continue the work later this year.

aendeavor:
@j-marz I refactored `setup.sh` already, currently waiting for the Travis tests to finish. Would be nice to see you continuing :+1:

aendeavor:
@j-marz I've got time now, will tend to it.

aendeavor:
Can be closed due to #1601

Issue #1221, opened 2019-08-11
Format scripts according to standard


erik-wramner:
The scripts are not formatted according to the standard and as a consequence there is a mix of indentation sizes, sometimes tabs are used but mostly spaces. It makes it harder to work with the code, in particular when blocks of code are copied/moved between files that use different conventions. In particular the start-mailserver.sh script uses tabs in some places and spaces in others and indents with 8 spaces, whereas the standard is to only use two spaces and no tabs.

## Possible Fix
I can fix the scripts (or at least start-mailserver.sh to begin with), but as this will cause major changes I would like to do it at an agreed point in time when we try to avoid having outstanding work. Merging real changes when there are whitespace changes all over is not my idea of fun!

fbartels:
One way to tackle this would be to add an .editorconfig config file into the repository. Not all editors support this standard out of the box, but this would be a way to define the expected behavior in this project.

https://editorconfig.org

Current output of ´eclint´ (xargs stops after the first command has failed, therefore only errors of ´setup.sh´ are shown):

```
$ git ls-files --exclude='*.sh' --ignored | xargs --max-lines=1 eclint check
setup.sh
    152:01 ❌ invalid indent size: 3, expected: 4       (EditorConfig indent_size  https://goo.gl/QtGB8R)
    153:01 ❌ invalid indent size: 5, expected: 6       (EditorConfig indent_size  https://goo.gl/QtGB8R)
    154:01 ❌ invalid indent size: 5, expected: 6       (EditorConfig indent_size  https://goo.gl/QtGB8R)
    276:01 ❌ invalid indent style: tab, expected: space
                                                         (EditorConfig indent_style https://goo.gl/8Qkrbr)
```

erik-wramner:
Yes, but the problem with the big change and hence merge issues with any outstanding PRs remains. It would be much easier to do this when there are no or at least as few PRs waiting as possible. We also need to consider the new Debian branch and any changes that have been done there.

aendeavor:
I'll have a look at this. Should be about the right time.

aendeavor:
Done with #1619

PR #1229, opened 2019-08-15
enforce formatting according to .editorconfig

fbartels:
relates to #1221


fbartels:
@erik-wramner as you already thought this chance is quite invasive (and current prs will probably need to be rebased and merge conflicts resolved).

@j-marz i've seen you started work on a shellcheck pr. what are your thoughts towards this?

erik-wramner:
Yes, it affects a lot of code. Looking at the active PRs we have:

- ShellCheck - needs to rebase/merge or commit first
- Modularize tests - needs to rebase/merge or commit first
- Port descriptions - just docs, not affected
- SA short circuit - waiting for action, stale? Can probably close.
- Regex aliases - stale, will close Tomorrow unless I get an answer
- Sqlite - I think we should reject that as outside the project scope
- Check for updates of certs - I think we should reject that in favor of frequent updates
- Test separation - stale, kept for reference, no need to rebase
- Change pw - should ideally use other container? Relatively free-standing
- Dovecot mysql - I think we should reject that as outside the project scope
- New Debian version (branch) - not sure of impact but would need to rebase/merge

Not that bad except for the new Debian version. I don't know how many changes Tom has made to the code there. It would be nice to get some input from @tomav on this.

.editorconfig
```diff
@@ -27,6 +26,10 @@ indent_size = 2
 indent_style = space
 indent_size = 2
 
-[target/start-mailserver.sh]
```
erik-wramner: What happens with start-mailserver.sh? Do we use 2 spaces now, or is it not checked at all?

config/postfix-accounts.cf
```diff
@@ -0,0 +1 @@
+# TODO is this file even used?
```
erik-wramner: I think this and the other similar files are placeholders that make it easy for new user to find where the can make changes. The .bak file has probably been checked in by mistake though.

.editorconfig
fbartels: With this change `start-mailserver.sh` is simply treated like all the other bash scripts in this repository.  But even with the exception for start-mailserver there were quiet a lot of lines that would have needed fixing (did not count).

config/postfix-accounts.cf
fbartels: It feel a bit like this is a relic of a change of the past. I did not immediately find where this file is put into the container and therefore decided to add this todo.

.editorconfig
erik-wramner: Yes, I've noticed.

PR #1233, opened 2019-08-20
Improved documentation for SA_KILL

jamesfryer:


erik-wramner:
Thanks!

Unfortunately I don't think the path is correct. It depends on several things. The mail state is only used with ONE_DIR=1, otherwise the files are in other locations. Furthermore the exact path that docker uses for volumes is likely to differ. Docker can provision volumes with many drivers.

I think you should document the paths that are used inside the container. There will be one path with ONE_DIR=1 and another with ONE_DIR=0.

jamesfryer:
I believe the values inside the container will be:

ONE_DIR=1: /var/mail-state/lib-amavis/virusmails/
ONE_DIR=0: /var/lib/amavis/virusmails/

Does that look right to you?

erik-wramner:
Yes, that should be correct. The start script sets the state directory for amavis to /var/lib/amavis or /var/mail-state/lib-amavis depending on ONE_DIR.

jamesfryer:
I have changed it as you suggest. I now worry it is turning into an essay.

erik-wramner:
Well, it is a bit long, but at least it is informative. Thanks!

PR #1243, opened 2019-08-29
Use dovecot repo to get CVE-2019-11500 patched dovecot 2.3.7.2

hof:


erik-wramner:
Unfortunately the tests fail with this PR and it is not a random failure. I've restarted the tests and they consistently fail in the same way. We need to solve that before we merge this.

fbartels:
Has been implemented with https://github.com/tomav/docker-mailserver/pull/1245

PR #1245, opened 2019-08-30
Use Dovecot repo to fix cve-2019-11500 take two

erik-wramner:
Switch repo and fix broken configuration.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,7 @@ ENV SASLAUTHD_MECH_OPTIONS=""
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
 # Packages
-# hadolint ignore=DL3015
+# hadolint ignore=DL3015,SC2016
 RUN echo "deb http://ftp.debian.org/debian stretch-backports main" | tee -a /etc/apt/sources.list.d/stretch-bp.list && \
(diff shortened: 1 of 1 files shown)


hof:
Nice work! 

Maybe remove the reference to `stretch-backport` on line 135.

erik-wramner:
@hof, I'll do that in a separate commit to speed things up. Thanks.

@fbartels, what do you think? Shall we merge this now or wait for backports? It is running on one of my servers, so far it seems to work.

erik-wramner:
@fbartels to save time, if you like it just merge it. If you think we should wait for backports then we'll do that instead. I'm fine with either. @tomav if you read this you naturally have a vote as well!

Dockerfile
```diff
@@ -17,13 +17,14 @@ ENV SASLAUTHD_MECH_OPTIONS=""
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
 # Packages
-# hadolint ignore=DL3015
+# hadolint ignore=DL3015,SC2016
```
fbartels: I want to later look why the next exception is needed, but this should not block

fbartels:
Lets get this merged then.

@tomav @erik-wramner I think we should then also really tag a new release. What do you think, should this be a v6.2.0?

Issue #1252, opened 2019-09-05
Make some use case examples that reproducibly work


yjiangnan:
I have been struggling for the last several days to make docker-mailserver to work on my Mac, but only with limited success and even the simplest setups throw problems at me. I know mail-server is hard, due to the large numbers of components. But isn't it the purpose of docker to make the configurations contained as reproducible tool sets to be easily used in the future or by others? Email is an decades-old tool without much change anyway! I will brief describe the problems I faced and then make some suggestions about making some standardized docker images.

I tried to get a simple setup following the documentation:
```
mkdir mailtest && cd mailtest/
docker pull tvial/docker-mailserver:latest
curl -o setup.sh https://raw.githubusercontent.com/tomav/docker-mailserver/master/setup.sh; chmod a+x ./setup.sh
curl -o docker-compose.yml https://raw.githubusercontent.com/tomav/docker-mailserver/master/docker-compose.yml.dist
curl -o .env https://raw.githubusercontent.com/tomav/docker-mailserver/master/.env.dist
```
Then I edited `hostname`, `domainname` and `container_name` and attached my SSL certificate in volumes in `docker-compose.yml`. Then I run `sudo docker-compose up mail` and get the following error:
```
mail    | grep: /tmp/docker-mailserver/postfix-accounts.cf: No such file or directory
mail    | Error: Unless using LDAP, you need at least 1 email account to start the server.
mail    | Error: Please fix your configuration. Exiting...
mail    | 2019-09-05 02:48:35,716 INFO exited: mailserver (exit status 1; not expected)
```
So, I do not have email accouts. But why does the documentation asks me to run the `docker-compose up` at this point?? And in the detach mode! Even better, it asks me to edit the `.env` file which I have no idea what to do with and went through the file for a whole day! So, I then run
```
./setup.sh email add notify@mydomain.org pswd1112
```
I get another error:
```
"docker inspect" requires at least 1 argument.
See 'docker inspect --help'.

Usage:  docker inspect [OPTIONS] NAME|ID [NAME|ID...]

Return low-level information on Docker objects
```
It seems that this problems is caused by line 7 of setup.sh that return emty results. So, I had to change `INFO=$(docker ps \` to `INFO=$(docker ps -a\` and it at least gives me a first taste of success. WOW!!!
Then I run `./setup.sh config dkim` and finally after running `sudo docker-compose up mail` I can successfully send emails using the following Python code:

```
import smtplib
[...]

PR #1255, opened 2019-09-08
Make the documentation more readible and error-free

yjiangnan:
This has been tested on both Mac and Ubuntu, although Ubuntu didn't work for me due to the Cloud provider blocking outgoing traffic on port 25.
Refer to discussion:
 Make some use case examples that reproducibly work #1252


erik-wramner:
Well, this was a fairly large PR. @fbartels I would like your views on this as I'm not sure how to best handle it.

First we have the code change where `setup.sh` is changed to use `docker ps -a` for Mac support. On the surface that seems fine, but I have two possible objections. One is that it doesn't do the same thing, with the `-a` flag stopped containers are included and trying to connect to a stopped container using the container name will fail. That could break things. The other and much less serious objection is that I'm a bit hesitant to make the scripts work better with Mac unless we are willing to really support Mac hosts and I think that could be a lot of work. Better to run into problems quickly than to go into production and find the issues there.

Second we have the docs. We don't want to remove the environment options from the main readme. It is nice to have them in their own file, but I think the main readme is copied to Docker hub when the image is built (not sure) or worst case manually with copy & paste. It is worth a lot to have the configurable options where people find the image so they don't have to go to Github to find them.

Plus by and large I think the main readme is good, I don't want to refactor it too much. Adding extra detail is good, but then it needs to be correct as well - for example I don't think we need at least one account, as I recall it is quite possible to run an `SMTP_ONLY `server without any accounts.

Finally the examples. I don't think we should make examples for advanced configurations (the ones that are waiting for contributor action) at all. It is better to document the options provided and trust the users to read up on what they need. Running a large scale mail solution is a complex job and should be treated as such or there will be unexpected issues down the line.

The quick getting started guide could be a good addition; however, it feels as if it overlaps what we already have in the readme and on the wiki (SSL guide for LetsEncrypt, FAQ, ...) and I don't want to suggest using Python code that may not work as intended. I'd rather use the tools that are already in the image that we use for the tests. Perhaps we could just elaborate a bit on the step-by-step guide that is in the readme to clear up any misconceptions?

I feel that I'm overly negative here, so I'd appreciate other views.

fbartels:
@erik-wramner will have a deeper read in the evening. My first reaction this morning was that by moving a lot of text around I cannot directly see where this is correcting errors (because of the `error-free` statement).

yjiangnan:
@erik-wramner 
I tested on Ubuntu linux. The old `setup.sh` also pops out error messages on Linux when there are no containers. I strongly suggest you to test it on a completely fresh machine, with no other containers and with no email counts previously setup. I am not sure what `docker ps ` was trying to achieve. But the real question is *not* whether to support Mac, but whether to support new users. I am not sure what is the best way to solve the problem when no contains exist, but `docker ps -a` at least worked for me on both Mac and Linux.

The huge list of environmental variables on the main doc greatly increase the complexity and reduces readability. It also creates a lot of confusion for new users. The main doc should be the entry point of the whole project, with clear directions for where to go for different users, instead of an all-including huge piece of documentation on all the technical details. The new organization may create a little bit inconvenience by one more mouse clicking, but it is going to save huge amount of efforts for new users to just get started with. So, the question is again whether you want to encourage new users, or just to make it a little bit more convenient for old users. But the way, for old experienced users, why would you go to GitHub to look for reference? Isn't the documentation also present in your local machine in the `.env` file? In fact, I thought about completely remove the documentation of environment variables on Github because it is almost identical duplication of the `.env` file, which would make the maintenance more complicated. The `.env` would be a superior source for reference because it is directly next to the real code that is going to run on your machine. Therefore, if the default values change in the future and you just go to the GitHub README file for reference, you would be assuming the wrong thing.

` I don't think we need at least one account, as I recall it is quite possible to run an SMTP_ONLY server without any accounts.` Please test your thought on a new machine. I tested on Linux and Mac, they both fail with the instructions from the documentation. If you think `SMTP_ONLY` server would work, then we have to make it the default value, because the README instruction says nothing about setting up a SMTP_ONLY server.

I do feel that good examples are very helpful. At least, when it is well organized, it doesn't affect users who do not need the examples. As you said, mail servers are complex, so why not make it simple for users when it is possible to do so? `Running a large scale mail solution is a complex job and should be treated as such or there will be unexpected issues down the line.` That is why we need some standardized configurations to avoid users repeatedly running into the unexpected issues.

`it feels as if it overlaps what we already have` Sometimes a little bit overlap is helpful because it brings things into context. So, users do not have read the whole repo to get the simplest things done. The Python code is simple and worked well for me. I am not aware of the test tools because they are not in the doc. If they are simpler and more transparent, I have no problem using them. The bottom line is, users need some simple way to test whether the mail server is working and the documentation does not have such information currently.

erik-wramner:
I'm not trying to bash you. The setup command is used for many different things and in many contexts (new container, existing container that is running, existing container that is not running), so while your specific scenario works there is no guarantee that we don't break one of the other use cases. The real question for that has nothing to do with new users as the script works as is; the question is if the change has nasty side effects or if it is benign. It could well be benign, but it is not obvious. Furthermore it is not clear why the Mac would need to check the stopped containers. It sounds as if a container failed and stopped for some reason, so that the real problem is not in the setup script but somewhere else.

As for the docs I beg to disagree. Having the configuration options on docker hub along with the image is highly beneficial and many new users for docker-mailserver are likely to have prior experience with other servers. They know the terms and want to see if this image provides X and Y and Z. Of course there are complete beginners as well, but having the options listed below the quick installation instructions won't hurt them. On the contrary it may make them think about what they want enabled in their server so that they can search for the terms and make an educated decision.

The `SMTP_ONLY` option allows you to run a server with only Postfix, i.e. without Dovecot. That means you can send mail, but you can't read mail. That is not a good default option, but it is a useful option in some cases and it is supported.

Anyway, let's see what @fbartels thinks.

yjiangnan:
@fbartels 
The main errors come from two sources:
- First, when there are no email accounts, the container will exit with message:
```
mail    | grep: /tmp/docker-mailserver/postfix-accounts.cf: No such file or directory
mail    | Error: Unless using LDAP, you need at least 1 email account to start the server.
mail    | Error: Please fix your configuration. Exiting...
mail    | 2019-09-09 15:01:56,726 INFO exited: mailserver (exit status 1; not expected)
```
The solution is to move the steps of creating accounts and DKIM setup before starting the mail server.
- Second, `./setup.sh email add no-reply@domain.com passwd` prints out the following 
```
"docker inspect" requires at least 1 argument.
See 'docker inspect --help'.

Usage:  docker inspect [OPTIONS] NAME|ID [NAME|ID...]

Return low-level information on Docker objects
```
This seems like an error for users. Maybe it is not. I am not sure, because the account has actually been created after further examination. This at least needs some fix, either in the code, or in the documentation.
I again reproduced the above problems on Ubuntu linux for the sake of this reply.

yjiangnan:
@erik-wramner 
It hurts, because it is a never ending list of configurations and takes a long time to read through. You may know that it is just a list of configurations, but how do new readers know *before* reading everything?? If they do not read through everything, how can they be sure that they did not miss something important? The solution? Put them in a separate file. If you are worried about the readers missing the point that they could do a lot of configurations, just briefly mention them in the main page and then point to the right location.

erik-wramner:
For the record I have just removed all docker images from a Linux host and performed a clean installation following the original readme:

```
docker pull tvial/docker-mailserver:latest
curl -o setup.sh https://raw.githubusercontent.com/tomav/docker-mailserver/master/setup.sh; chmod a+x ./setup.sh
curl -o docker-compose.yml https://raw.githubusercontent.com/tomav/docker-mailserver/master/docker-compose.yml.dist
curl -o .env https://raw.githubusercontent.com/tomav/docker-mailserver/master/.env.dist
docker-compose up -d mail
./setup.sh email add test@domain.com test123
./setup.sh config dkim
```
It worked (on a Linux host - probably not on Mac) and this is straight out of the readme from Usage and about one page down. From there the rest will depend on what the specific installation is all about.

yjiangnan:
@erik-wramner 
To see the error message, you need to run 
`docker-compose up mail`
instead of 
`docker-compose up -d mail`
before creating accounts. If you just run in the detach mode, the container appears to be still running, but I am not sure whether it would function correctly given the error message.

martin-schulze-vireso:
> To see the error message, you need to run
> `docker-compose up mail`
> instead of
> `docker-compose up -d mail`

Are you talking about log messages from the container? If so, you can also use `docker-compose logs mail` to see the logs even when running in detached mode. Furthermore, just to be sure: are you aborting the non detached run (via CTRL+C) to enter new commands? If so, this would explain why the container is stopped.

I think the best solution to get everybody on the same page would be a shell script that can be used by all parties to get the same behavior.

ENV.MD
```diff
@@ -0,0 +1,427 @@
+# Environment variables
+
+Please check [how the container starts](https://github.com/tomav/docker-mailserver/blob/master/target/start-mailserver.sh) to understand what's expected. Also if an option doesn't work as documented here, check if you are running the latest image!
```
fbartels: Instead of pointing towards the start-mailserver.sh script to "understand whats expected" the script should use default values for all configuration options.

ENV.MD
```diff
@@ -0,0 +1,427 @@
...
+  - 1 => Enables POP3 service
+
+##### ENABLE_FAIL2BAN
+
+  - **0** => fail2ban service disabled
+  - 1 => Enables fail2ban service
+
+If you enable Fail2Ban, don't forget to add the following lines to your `docker-compose.yml`:
+
+    cap_add:
```
fbartels: this should be a code block

ENV.MD
```diff
@@ -0,0 +1,427 @@
...
+  - 1 => Clamav is enabled
+
+##### ONE_DIR
+
+  - **0** => state in default directories
+  - 1 => consolidate all states into a single directory (`/var/mail-state`) to allow persistence using docker volumes
+
+##### ENABLE_POP3
+
+  - **empty** => POP3 service disabled
```
fbartels: the default value here is then probably 0 instead of empty

ENV.MD
```diff
@@ -0,0 +1,427 @@
...
+If you enable Fail2Ban, don't forget to add the following lines to your `docker-compose.yml`:
+
+    cap_add:
+      - NET_ADMIN
+
+Otherwise, `iptables` won't be able to ban IPs.
+
+##### SMTP_ONLY
+
+  - **empty** => all daemons start
```
fbartels: here 0 as the default makes also more sense

ENV.MD
```diff
@@ -0,0 +1,427 @@
...
+  - 1 => only launch postfix smtp
+
+##### SSL_TYPE
+
+  - **empty** => SSL disabled
+  - letsencrypt => Enables Let's Encrypt certificates
+  - custom => Enables custom certificates
+  - manual => Let you manually specify locations of your SSL certificates for non-standard cases
+  - self-signed => Enables self-signed certificates
+  - _any other value_ => SSL required, settings by default
```
fbartels: this option does not make sense to me (without checking the code)

ENV.MD
```diff
@@ -0,0 +1,427 @@
...
+
+Please read [the SSL page in the wiki](https://github.com/tomav/docker-mailserver/wiki/Configure-SSL) for more information.
+
+##### TLS_LEVEL
+
+  - **empty** => modern
+  - modern => Enables TLSv1.2 and modern ciphers only. (default)
```
fbartels: instead of empty, modern should be marked as the default

ENV.MD
```diff
@@ -0,0 +1,427 @@
...
+  - intermediate => Enables TLSv1, TLSv1.1 and TLSv1.2 and broad compatibility ciphers.
+  - old => NOT implemented. If you really need it, then customize the TLS ciphers overriding postfix and dovecot settings [ wiki](https://github.com/tomav/docker-mailserver/wiki/
+
+##### SPOOF_PROTECTION
+Configures the handling of creating mails with forged sender addresses.
+  - **empty** => Mail address spoofing allowed. Any logged in user may create email messages with a forged sender address. See also [Wikipedia](https://en.wikipedia.org/wiki/Email_spoofing)(not recommended, but default for backwards compatibility reasons)
```
fbartels: 0 instead of empty

ENV.MD
```diff
@@ -0,0 +1,427 @@
...
+Set different options for mynetworks option (can be overwrite in postfix-main.cf)
+  - **empty** => localhost only
+  - host => Add docker host (ipv4 only)
+  - network => Add the docker default bridge network (172.16.0.0/12); **WARNING**: `docker-compose` might use others (e.g. 192.168.0.0/16) use `PERMIT_DOCKER=connected-networks` in this case
+  - connected-networks => Add all connected docker networks (ipv4 only)
+
+##### VIRUSMAILS_DELETE_DELAY
+
+Set how many days a virusmail will stay on the server before being deleted
+  - **empty** => 7 days
```
fbartels: instead of empty, the default value seems then to be 7

ENV.MD
```diff
@@ -0,0 +1,427 @@
...
+
+
+##### ENABLE_POSTFIX_VIRTUAL_TRANSPORT
+
+This Option is activating the Usage of POSTFIX_DAGENT to specify a ltmp client different from default dovecot socket.
+
+- **empty** => disabled
```
fbartels: 0 instead of empty

ENV.MD
```diff
@@ -0,0 +1,427 @@
...
+- 1 => enabled
+
+##### POSTFIX_DAGENT
+
+Enabled by ENABLE_POSTFIX_VIRTUAL_TRANSPORT. Specify the final delivery of postfix
+
+- **empty**: fail
```
fbartels: instead of failing the container on an undefined value we should stick to a sensible default. ´lmtp:unix:private...´ seems a good choice.

ENV.MD
```diff
@@ -0,0 +1,427 @@
...
+- `lmtp:unix:private/dovecot-lmtp` (use socket)
+- `lmtps:inet:<host>:<port>` (secure lmtp with starttls, take a look at https://sys4.de/en/blog/2014/11/17/sicheres-lmtp-mit-starttls-in-dovecot/)
+- `lmtp:<kopano-host>:2003` (use kopano as mailstore)
+- etc.
+
+##### POSTFIX\_MAILBOX\_SIZE\_LIMIT
```
fbartels: does this really need the backflashes (same as with the next option)

ENV.MD
```diff
@@ -0,0 +1,427 @@
...
+
+Set the mailbox size limit for all users. If set to zero, the size will be unlimited (default).
+
+- **empty** => 0 (no limit)
```
fbartels: better

**0** => no limit

ENV.MD
```diff
@@ -0,0 +1,427 @@
...
+
+##### POSTFIX\_MESSAGE\_SIZE\_LIMIT
+
+Set the message size limit for all users. If set to zero, the size will be unlimited (not recommended!)
+
+- **empty** => 10240000 (~10 MB)
```
fbartels: it does not seem to be empty by default, but 10240000

ENV.MD
```diff
@@ -0,0 +1,427 @@
...
+
+##### ENABLE_MANAGESIEVE
+
+  - **empty** => Managesieve service disabled
```
fbartels: 0 instead of empty

ENV.MD
```diff
@@ -0,0 +1,427 @@
...
+  - 1 => Enables Managesieve on port 4190
+
+##### OVERRIDE_HOSTNAME
+
+  - **empty** => uses the `hostname` command to get the mail server's canonical hostname
+  - => Specify a fully-qualified domainname to serve mail for.  This is used for many of the config features so if you can't set your hostname (e.g. you're in a container platform that doesn't let you) specify it in this environment variable.
+
+##### POSTMASTER_ADDRESS
+
+  - **empty** => postmaster@domain.com
```
fbartels: is that then really postmaster@domain.com or is domain.com depending on the hostname?

ENV.MD
```diff
@@ -0,0 +1,427 @@
...
+  - 1 => `fetchmail` enabled
+
+##### FETCHMAIL_POLL
+  - **300** => `fetchmail` The number of seconds for the interval
+
+## LDAP
+
+##### ENABLE_LDAP
+
+  - **empty** => LDAP authentification is disabled
```
fbartels: 0 instead of empty

ENV.MD
```diff
@@ -0,0 +1,427 @@
...
+  - 1 => LDAP authentification is enabled
+  - NOTE:
+    - A second container for the ldap service is necessary (e.g. [docker-openldap](https://github.com/osixia/docker-openldap))
+    - For preparing the ldap server to use in combination with this container [this](http://acidx.net/wordpress/2014/06/installing-a-mailserver-with-postfix-dovecot-sasl-ldap-roundcube/) article may be helpful
+
+##### LDAP_START_TLS
+
+  - **empty** => no
```
fbartels: no instead of empty

ENV.MD
```diff
@@ -0,0 +1,427 @@
...
+  - yes => LDAP over TLS enabled for Postfix
+
+##### LDAP_SERVER_HOST
+
+  - **empty** => mail.domain.com
+  - => Specify the dns-name/ip-address where the ldap-server
```
fbartels: this sentence ends abruptly

ENV.MD
```diff
@@ -0,0 +1,427 @@
...
+  - NOTE: If you going to use the mailserver in combination with docker-compose you can set the service name here
```
fbartels: grammar

ENV.MD
```diff
@@ -0,0 +1,427 @@
...
+
+##### LDAP_SEARCH_BASE
+
+  - **empty** => ou=people,dc=domain,dc=com
```
fbartels: instead of empty, the value seems to be ou=people...

ENV.MD
```diff
@@ -0,0 +1,427 @@
...
+  - => e.g. LDAP_SEARCH_BASE=dc=mydomain,dc=local
+
+##### LDAP_BIND_DN
+
+  - **empty** => cn=admin,dc=domain,dc=com
```
fbartels: empty is not really empty

ENV.MD
```diff
@@ -0,0 +1,427 @@
...
+  - => take a look at examples of SASL_LDAP_BIND_DN
+
+##### LDAP_BIND_PW
+
+  - **empty** => admin
```
fbartels: empty is not really empty

ENV.MD
```diff
@@ -0,0 +1,427 @@
...
+  - => Specify the password to bind against ldap
+
+##### LDAP_QUERY_FILTER_USER
+
+  - e.g. `(&(mail=%s)(mailEnabled=TRUE))`
```
fbartels: these settings have no default?

ENV.MD
```diff
@@ -0,0 +1,427 @@
...
+  - => Specify how ldap should be asked for aliases
+
+##### LDAP_QUERY_FILTER_DOMAIN
+
+- e.g. `(&(|(mail=*@%s)(mailalias=*@%s)(mailGroupMember=*@%s))(mailEnabled=TRUE))`
+- => Specify how ldap should be asked for domains
+
+##### DOVECOT_TLS
+
+  - **empty** => no
```
fbartels: default seems no

yjiangnan:
@martin-schulze-vireso 
Yes. I ran the `setup.sh` script without any containers running. That explains why it didn't work as expected. However, if I run `docker-compose up -d mail` before creating any account, then `docker-compose logs mail` still gives the same error message.
Maybe it is true that `setup.sh` does not have to be changed, but in that case, `docker-compose up -d mail` must run without any error messages. I feels that this might be more difficult to achieve. It may still be best to create account first, but somehow suppressing the `"docker inspect" requires at least 1 argument.` message.

ENV.MD
```diff
@@ -0,0 +1,427 @@
...
+  - yes => LDAP over TLS enabled for Dovecot
+
+## Dovecot
+
+The following variables overwrite the default values for ```/etc/dovecot/dovecot-ldap.conf.ext```.
+
+##### DOVECOT_USER_FILTER
+
+  - e.g. `(&(objectClass=PostfixBookMailAccount)(uniqueIdentifier=%n))`
```
fbartels: this could also mention the default values. maybe its sufficient to say that unless set dovecot-ldap.conf.ext will not be touched

ENV.MD
```diff
@@ -0,0 +1,427 @@
...
+## SASL Auth
+
+##### ENABLE_SASLAUTHD
+
+  - **0** => `saslauthd` is disabled
+  - 1 => `saslauthd` is enabled
+
+##### SASLAUTHD_MECHANISMS
+
+  - empty => pam
```
fbartels: instead of empty list pam, make it as default and explain pam (its the only unexplained at the moment)

ENV.MD
```diff
@@ -0,0 +1,427 @@
...
+  - `ldap` => authenticate against ldap server
+  - `shadow` => authenticate against local user db
+  - `mysql` => authenticate against mysql db
+  - `rimap` => authenticate against imap server
+  - NOTE: can be a list of mechanisms like pam ldap shadow
+
+##### SASLAUTHD_MECH_OPTIONS
+
+  - empty => None
```
fbartels: None instead of empty

ENV.MD
```diff
@@ -0,0 +1,427 @@
...
+  - e.g. with SASLAUTHD_MECHANISMS rimap you need to specify the ip-address/servername of the imap server  ==> xxx.xxx.xxx.xxx
+
+##### SASLAUTHD_LDAP_SERVER
+
+  - empty => localhost
```
fbartels: localhost instead of empty

ENV.MD
```diff
@@ -0,0 +1,427 @@
...
+
+##### SASLAUTHD_LDAP_SSL
+
+  - empty or 0 => `ldap://` will be used
```
fbartels: drop the "empty or"

ENV.MD
```diff
@@ -0,0 +1,427 @@
...
+  - 1 => `ldaps://` will be used
+
+##### SASLAUTHD_LDAP_BIND_DN
+
+  - empty => anonymous bind
```
fbartels: anonymous bind instead of empty

ENV.MD
```diff
@@ -0,0 +1,427 @@
...
+  - specify an object with privileges to search the directory tree
+  - e.g. active directory: SASLAUTHD_LDAP_BIND_DN=cn=Administrator,cn=Users,dc=mydomain,dc=net
+  - e.g. openldap: SASLAUTHD_LDAP_BIND_DN=cn=admin,dc=mydomain,dc=net
+
+##### SASLAUTHD_LDAP_PASSWORD
+
+  - empty => anonymous bind
```
fbartels: anonymous bind instead of empty

ENV.MD
```diff
@@ -0,0 +1,427 @@
...
+
+##### SASLAUTHD_LDAP_SEARCH_BASE
+
+  - empty => Reverting to SASLAUTHD_MECHANISMS pam
```
fbartels: not quite sure what to suggest here

Examples/README.md
```diff
@@ -0,0 +1,13 @@
+This folder contains several examples to show different configurations of the mailserver for different purposes. You can choose from what you need. This list is still limited. Contributions are encouraged.
+
+# Basic functionality
+[basic.md](basic.md) allows sending notifications/alerts, with TLS setups and spoof protection. This can also be a beginner's guide to get a feeling about how the setup works.
+
+# With LDAP functionality
+[ldap.md](ldap.md) shows the setup with LDAP. However, it maybe unclear to a new user what LDAP is and why should one use it. What is the context for it to be useful? Contributions are needed here.
```
fbartels: > Contributions are needed here.

Ah, now I get @erik-wramner comment. instead of asking for contributions in the document I would put this into an issue. Same for the other examples.

Do document what you know, and directly list missing pieces in the readme. Putting questions in a document hardly ever pushes people to contribute. to these questions.

Examples/basic.md
```diff
@@ -0,0 +1,59 @@
+This setup allows sending notifications/alerts, with TLS setups and spoof protection. This can also be a beginner's guide to get a feeling about how the setup works.
+
+First, get your TLS certificate from [Letsencrypt](https://letsencrypt.org) or follow a docker example [here](https://www.humankode.com/ssl/how-to-set-up-free-ssl-certificates-from-lets-encrypt-using-docker-and-nginx). Do not forget to also add your subdomain `mail.mydomain.org` to the list of domains during creating the certificate.
+
+Then, use the following `docker-compose.yml` file (edit with your own domain name) instead of the default one to re-create your mailserver following the instructions on the main page (or simply re-run `sudo docker-compose up -d mail`).
+
+```
+version: '2'
```
fbartels: putting the docker-compose.yml into the readme makes it much harder to maintain it. as you need to copy it out and your changes back in. Rather put the yml alongside of the readme.

Examples/basic.md
```diff
@@ -0,0 +1,59 @@
...
+    container_name: mail
+    ports:
+    - "25:25"
+    - "143:143"
+    - "587:587"
+    - "993:993"
+    volumes:
+    - maildata:/var/mail
+    - mailstate:/var/mail-state
+    - ./config/:/tmp/docker-mailserver/
```
fbartels: I personally dislike this combination of docker volumes and bind mounts

Examples/basic.md
```diff
@@ -0,0 +1,59 @@
...
+    - SYS_PTRACE
+    restart: always
+volumes:
+  maildata:
+    driver: local
+  mailstate:
+    driver: local
+
+```
+Then you can use the following Python code or other methods to send email with your own domains!
```
fbartels: Instead of pasting a script here, maybe rather document how you would setup your client (or even better point to documentation for clients)

Examples/ldap.md
```diff
@@ -0,0 +1,56 @@
+The following example `docker-compose.yml` file shows the setup with LDAP. However, it maybe unclear to a new user what LDAP is and why should one use it. What is the context for it to be useful? What other services should be setup together? Contributions are needed here.
```
fbartels: Like written before I would not put questions that need answers into the readme.

Examples/basic.md
```diff
@@ -0,0 +1,59 @@
...
+
+First, get your TLS certificate from [Letsencrypt](https://letsencrypt.org) or follow a docker example [here](https://www.humankode.com/ssl/how-to-set-up-free-ssl-certificates-from-lets-encrypt-using-docker-and-nginx). Do not forget to also add your subdomain `mail.mydomain.org` to the list of domains during creating the certificate.
+
+Then, use the following `docker-compose.yml` file (edit with your own domain name) instead of the default one to re-create your mailserver following the instructions on the main page (or simply re-run `sudo docker-compose up -d mail`).
+
+```
+version: '2'
+services:
+  mail:
+    image: tvial/docker-mailserver:latest
```
fbartels: I personally dislike using :latest

README.md
```diff
@@ -63,26 +80,32 @@ Download the docker-compose.yml, the .env and the setup.sh files:
...
-  - Don't quote your values.
-  - Variable substitution is *not* supported (e.g. `OVERRIDE_HOSTNAME=$HOSTNAME.$DOMAINNAME`).
+- The `.env` file contains the default environmental variables and their explanations. You can also find them in the more readable file [ENV.MD](ENV.MD). These variables will be passed to `docker-compose.yml` during running. You can change these variables in `docker-compose.yml`.
 - Install [docker-compose](https://docs.docker.com/compose/) in the version `1.7` or higher.
 
-#### Start Container
-    docker-compose up -d mail
-
-#### Create your mail accounts
+#### Create your mail accounts (at least one account)
```
fbartels: agree with leaving out the "at least one account"

README.md
```diff
@@ -63,26 +80,32 @@ Download the docker-compose.yml, the .env and the setup.sh files:
...
+Also create an `A` record to point `mail.youdomain.tld` to your machine public IP address.
+
+#### Start Container
+    docker-compose up -d mail
 
-#### Restart and update the container
+#### Restart and update the container if necessary
 
     docker-compose down
     docker pull tvial/docker-mailserver:latest
```
fbartels: the `docker pull` part is not necessary. instead suggest to do `docker-compose pull`

ENV.MD
yjiangnan: This part confuses me. I am not sure what it means. If I am going to set it, I do need to change "domain" to something like "mydomain" and "com" to something like "com" or "org"? Or should they remain the way it is? Then why "dc=" appeared twice?

Examples/basic.md
yjiangnan: I still haven't set up any clients yet. For now, I am only using the mail server to send alert (and actually use the above Python code). For setting up a client, I would make it another use case example.

ENV.MD
yjiangnan: I am actually not sure what to expect from so many lines of code.

README.md
yjiangnan: But this is what the mailserver tells me to do
`mail    | Error: Unless using LDAP, you need at least 1 email account to start the server.`
It keeps complaining unless I do this.

fbartels:
Wow this setup.sh script is a piece of work. I have started a new pr to look into how this can be cleaned up and simplified at https://github.com/tomav/docker-mailserver/pull/1258.

>  One is that it doesn't do the same thing, with the -a flag stopped containers are included and trying to connect to a stopped container using the container name will fail. That could break things. 

Pretty much agree with this statement. Instead of relying on `-a` the script should stop if it cannot find a running container. On top of that this first query is a bit bogus to me, since it tries to identify the docker-mailserver containers by taking the first container that has supervisord running as the main process.

> The other and much less serious objection is that I'm a bit hesitant to make the scripts work better with Mac unless we are willing to really support Mac hosts

I would just label this as "community supported". As long as Mac compat does not break the rest, why not?

> main readme is copied to Docker hub when the image is built 

yes, the Docker Hub sets this during an automated build (but could also be done both manually or with a script)

> Please test your thought on a new machine. I tested on Linux and Mac, they both fail with the instructions from the documentation. 

Had the same failure on my test. container did not start because the config directory does not exist. could either be added to the bootstrapping commands or (better) we should move to data volumes that get auto created.

ENV.MD
fbartels: Indeed. its good to point to the script for "deeper understanding", but the container should start without any of the explained values explicitly set. (or them being set through compose).

ENV.MD
fbartels: That is just how a dn (distinguished name) is structured in LDAP,

README.md
fbartels: Message comes from https://github.com/tomav/docker-mailserver/blob/9e7348c8f93de24d3d5f796f0b8a4ef09b9968d9/target/start-mailserver.sh#L626-L631

So it is only necessary to create an account if you are not using LDAP.

erik-wramner:
@yjiangnan @martin-schulze-vireso I followed the instructions from the readme word for word and it works (on Linux). I don't really care if there are errors in the logs as they are obviously not important. It is possible to setup the system like this following the readme and later on (when the accounts have been added) the errors will be resolved. I see no burning need to change the working setup instructions?

As for Mac most likely the container stopped due to some error. That is why `-a` was needed. We don't support Mac, nor do we support Windows which would also be possible. We could look into Mac support and by all means Windows support, but I fear that would be a much bigger undertaking. Some features such as fail2ban will almost certainly not work.

Haven't looked at @fbartels review yet, will check later.

yjiangnan:
@erik-wramner 
Personally, I am very uncomfortable to just ignore the error messages. Two reasons:
1. Most users are unaware of whether error messages matter. By default, they matter because the original creator put them there most likely for a reason. So, the users would try really hard to fix this. If their mail server is not working due to the complexity of setup (say SSL problems, IP being blocked, firewall problems, etc.), they would instead think that the errors are what the error messages say. I have a feeling that you tend to ignore the struggle of new users. If this is just your personal project, that is fine. But if you intend this to be open source, that would be a disaster. For me personally, I spent several days to combat with the error messages (including some errors that I cannot reproduce so I did not mention here) and even thought about switching to mailcow. 
2. Error messages indicate errors and the program may not function properly. If you do not know how the error messages are generated inside out, you are then just burying your head in the sand, opening up to potential malfunctioning or even security holes.

erik-wramner:
@yjiangnan as I recall you said that the error was about not having accounts and the very next thing we do is to add an account, fixing the problem? If you had followed the getting started guide on Linux you probably wouldn't even have noticed. I'm not against adding some extra text to make that clearer, but as you say elsewhere that means the getting started guide will be longer. As it is it almost fits on my laptop screen.

yjiangnan:
@erik-wramner 
Then you need to first add a disclaimer that the setup only works on Linux (which is not true because I have got it to work on Mac). Then, after adding the account, the user should restart the container (and not for update purposes). 

Then, why not simply add the account first? 

I think in an open source context it is not the best idea to assume that the users would follow exactly your steps in your evironment setups. One thing is that a mailserver requires SSL setup and it is not covered in the main doc and can be tricky. The fewer the unnecessary error messages, the less burden there would be on the users.

erik-wramner:
@yjiangnan I think a disclaimer is in order, yes. "Have got it to work" is not the same as "everything works reliably". The tests only work on Linux. As long as the tests consistently fail it doesn't work in my book. Don't get me wrong, as @fbartels suggested the community could work towards making it stable on Mac and Windows and certainly you and others can use it on Mac and Windows, but be prepared for some rough edges.

I didn't write the instructions, but I think they are fairly short and to the point and they do their job. With docker I would assume users to follow the quick start guide unless they are comfortable debugging the errors they find. Then they can throw away the scratch container and do it for real, knowing how it works. Perhaps migrating existing accounts rather than creating a new one.

Finally for SSL there are several supported alternatives and some are quite complex, which is probably why they are not covered in the quick start guide. The self-signed certificate is there from the start and is good enough in some cases. For others setups see the wiki for details.

polarathene:
Not a proper review, as it's been some time, content in the commits would have become outdated, the extraction of ENV to it's own document is in an early commit and unclear what's different there, maybe the individual commits provide some insight? Otherwise a bit of a hassle to diff to original, and then to current master..

Many comments in the review for this PR are about the `empty` default that maps to the actual default value. These usually request to replace `empty` with the default value, or imply the `empty` value is `0`. Is that an assumption? 

I've not looked into how the values are being read or if it's equivalent to an unset ENV var. May be better to just avoid the list/map entry, and start with `Default: foobar`(where default is an unset ENV var), then list options that can be set to change that.

I do agree that on DockerHub, while it's nice to get some overview of configuration, the current bulk of options at the bottom half don't really belong there. Those are just noise to most coming across it and may be considered advanced configuration/tuning. Common config/features should be covered, the rest that isn't likely to matter to most users could be provided in another document or the Wiki which is linked to several times for further information.

aendeavor:
@yjiangnan Please have a look at the current PRs. This PR has become outdated due to #1995, #1996 and #1601.


If you have the time, please close this PR and re-open another one, based off of current master, implementing the changed proposed by @fbartels . If you do not have the time, please close this PR. I will wait for a reply - otherwise close this PR myself.

aendeavor (changes requested):
I agree with @fbartels  regarding the points to be changed

aendeavor:
This issue was closed due to one or more of the following reasons:

1. Age
2. Contributor inactivity
3. The issue seems to be resolved

If you think this happened by accident, or feel like this issue was not actually resolved, please feel free to re-open it. If there is an issue you could resolve in the meantime, please open a PR based on the current master branch so we can review it.

PR #1258, opened 2019-09-09
WIP: setup.sh review

fbartels:
setup.sh is a mess. Lets see if it can be done better.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -2,24 +2,20 @@
 
 # make script fail on errors
 # TODO add exit trap to report the failing line? at least say there was an error
-set -euo
+set -eu
 IFS=$'\n\t'
 ##
 # Wrapper for various setup scripts included in the docker-mailserver
 #
 
-WISHED_CONFIG_PATH=""
-
-# TODO grepping for supervisord is not a good way to find the right container
-INFO=$(docker ps \
-  --no-trunc \
-  --format="{{.Image}}\t{{.Names}}\t{{.Command}}" | \
-  grep "supervisord -c /etc/supervisor/supervisord.conf")
-
-IMAGE_NAME=$(echo $INFO | awk '{print $1}')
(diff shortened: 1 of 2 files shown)


setup.sh
```diff
@@ -1,19 +1,22 @@
...
 #
 
-INFO=$(docker ps \
-  --no-trunc \
-  --format="{{.Image}}\t{{.Names}}\t{{.Command}}" | \
-  grep "supervisord -c /etc/supervisor/supervisord.conf")
-
-IMAGE_NAME=$(echo $INFO | awk '{print $1}')
-CONTAINER_NAME=$(echo $INFO | awk '{print $2}')
+IMAGE_NAME=tvial/docker-mailserver:latest
```
erik-wramner: Hard-coding the image may not be a good idea as there can be changes between latest, stable and versioned releases. We should ideally use the same version that the user intends to run (or is running). As a default image name this works fine.

fbartels: Yes, I totally agree with that. In the end the script should check if there is any running container that is based on tvial/docker-mailserver.

Additionally the script has (at the moment) the ability to specify the name of the running container.

erik-wramner: That feature (specifying container name) is used extensively from the tests and is important. It is probably less important in real life, in theory someone could run multiple docker images on the same server but that is bound to be exceedingly rare.

erik-wramner:
Can we wrap up this? Perhaps keep some of the not-very-nice code for backwards compatibility (and possible later action) to be able to merge the good parts? In particular we could keep the old image name code unless you have a better idea for finding it; just hard-coding latest will break things.

erik-wramner:
FYI this needs to be rebased to get the podman changes.

fbartels:
+1 yes, I need to get back to this at some time.

fbartels:
Closing in favour of #1590

PR #1259, opened 2019-09-10
Fix missing error output from timeout

martin-schulze-vireso:
The exits would prevent the printing code from running. Thus we would get a failed test but no error message.


erik-wramner:
Looks good. There is no WIP in the title, but I'll double-check anyway: ready for merge?

martin-schulze-vireso:
Yes it is ready. I wanted to isolate this little fix to get it in ASAP.

PR #1262, opened 2019-09-15
Fix #1093, pflogsumm and logwatch

### LINKED ISSUE #1093: Is there a way to have logwatch and pflogsum added and act/deactivated by an Environment var?
Is there a way to have logwatch and pflogsum added and act/deactivated by an Environment var?

## Context

Currently I'm using logwatch and pflogsum to monitor my mailserver. The way I'm installing it is using some docker-exec and docker cp after creating and running the container. 

It would be nice to have them both added in the default-image.

## Steps to Reproduce

1.
2.
3.
4.

## Your Environment

* Amount of RAM available:
* Mailserver version used:
* Docker version used:
* Environment settings relevant to the config:
* Any relevant stack traces ("Full trace" preferred):

erik-wramner:
Added support for running pflogsumm with cron for Yesterday while keeping backwards compatibility. Also added support for logwatch reports.

### DIFF
--- a/README.md
+++ b/README.md
@@ -357,23 +357,23 @@ Set the message size limit for all users. If set to zero, the size will be unlim
 
 ##### PFLOGSUMM_TRIGGER
 
-  Enables a report being sent (created by pflogsumm) on a regular basis.
-  - **none** => No report
-  - daily_cron => Daily report for Yesterday will be sent
-  - logrotate => Full report for period will be sent when logs are rotated
+  Enables regular pflogsumm mail reports.
+  - **not set** => No report
+  - daily_cron => Daily report for the previous day
+  - logrotate => Full report based on the mail log when it is rotated
 
 This is a new option. The old REPORT options are still supported for backwards compatibility.
 If this is not set and reports are enabled with the old options, logrotate will be used.
 
 ##### PFLOGSUMM_RECIPIENT
 
-  Recipient address for pflogsumm reports if they are enabled.
+  Recipient address for pflogsumm reports.
   - **not set** => Use REPORT_RECIPIENT or POSTMASTER_ADDRESS
   - => Specify the recipient address(es)
 
 ##### PFLOGSUMM_SENDER
 
-  From address for pflogsumm reports if they are enabled.
+  From address for pflogsumm reports.
   - **not set** => Use REPORT_SENDER or POSTMASTER_ADDRESS
   - => Specify the sender address
 
--- a/README.md
+++ b/README.md
@@ -132,6 +132,7 @@ services:
     volumes:
       - maildata:/var/mail
       - mailstate:/var/mail-state
+      - maillogs:/var/log/mail
       - ./config/:/tmp/docker-mailserver/
     environment:
       - ENABLE_SPAMASSASSIN=1
@@ -148,6 +149,8 @@ volumes:
   maildata:
     driver: local
   mailstate:
+    driver: local
+  maillogs:
     driver: local
 ```
 
@@ -170,6 +173,7 @@ services:
     volumes:
       - maildata:/var/mail
       - mailstate:/var/mail-state
+      - maillogs:/var/log/mail
       - ./config/:/tmp/docker-mailserver/
     environment:
       - ENABLE_SPAMASSASSIN=1
@@ -205,6 +209,8 @@ volumes:
   maildata:
     driver: local
   mailstate:
+    driver: local
+  maillogs:
     driver: local
 ```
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -347,28 +353,81 @@ Set the message size limit for all users. If set to zero, the size will be unlim
   - ignore => Ignore the failure of this test. Allow other tests to complete. Repeat this test the next time the client connects. This option is useful for testing and collecting statistics without blocking mail.
 
 
-##### REPORT_RECIPIENT
+## Reports
+
+##### PFLOGSUMM_TRIGGER
+
+  Enables a report being sent (created by pflogsumm) on a regular basis.
```
fbartels: Reads a bit odd, how about

> Enables regular mail reported (created by pflogsumm)

?

README.md
```diff
@@ -347,28 +353,81 @@ Set the message size limit for all users. If set to zero, the size will be unlim
...
+  - **none** => No report
+  - daily_cron => Daily report for Yesterday will be sent
```
fbartels: Daily report for the previous day

README.md
```diff
@@ -347,28 +353,81 @@ Set the message size limit for all users. If set to zero, the size will be unlim
...
+  - logrotate => Full report for period will be sent when logs are rotated
```
fbartels: is this "full" different to the other options? Maybe specify that reporting will be weekly (have not checked the interval)

README.md
```diff
@@ -347,28 +353,81 @@ Set the message size limit for all users. If set to zero, the size will be unlim
...
+
+This is a new option. The old REPORT options are still supported for backwards compatibility.
+If this is not set and reports are enabled with the old options, logrotate will be used.
+
+##### PFLOGSUMM_RECIPIENT
+
+  Recipient address for pflogsumm reports if they are enabled.
```
fbartels: would drop the "if they are enabled" (same further down)

README.md
erik-wramner: It is different. Full means whatever was in the logs when they got rotated. If the log rotate period is monthly that will be a month, but there will also typically by a few entries for the new period. So if the period is daily, the report will often be for two days where one is very short. That is how it works now and what the cron alternative aims to fix.

README.md
erik-wramner: Adjusted the text a bit.

README.md
erik-wramner: OK

README.md
erik-wramner: OK

mindrunner:
I migrated to the new functionality, however I am getting the following errors from logrotate:
```
error: skipping "/var/log/mail/clamav.log" because parent directory has insecure permissions (It's world writable or writable by group which is not "root") Set "su" directive in config file to tell logrotate which user/group should be used for rotation.
error: skipping "/var/log/mail/freshclam.log" because parent directory has insecure permissions (It's world writable or writable by group which is not "root") Set "su" directive in config file to tell logrotate which user/group should be used for rotation.
error: skipping "/var/log/mail/mail.log" because parent directory has insecure permissions (It's world writable or writable by group which is not "root") Set "su" directive in config file to tell logrotate which user/group should be used for rotation.
error: skipping "/var/log/mail/mail.info" because parent directory has insecure permissions (It's world writable or writable by group which is not "root") Set "su" directive in config file to tell logrotate which user/group should be used for rotation.
error: skipping "/var/log/mail/mail.warn" because parent directory has insecure permissions (It's world writable or writable by group which is not "root") Set "su" directive in config file to tell logrotate which user/group should be used for rotation.
error: skipping "/var/log/mail/mail.err" because parent directory has insecure permissions (It's world writable or writable by group which is not "root") Set "su" directive in config file to tell logrotate which user/group should be used for rotation.
```

Did I miss out any special configuration? I had to chmod the mounted maillog folder to 777, because of the antivirus daemon not starting otherwise.

erik-wramner:
Hm, in my case `/var/log/mail` is owned by syslog and both clamav and freshclam logs to it:

`drwxr-xr-x 2 syslog root         7 Sep 20 09:08 mail`

I didn't do anything special, just changed the compose file to use a volume. Please start by changing the ownership and permissions to see if it works with my settings; then we need to find out why we got different results.

mindrunner:
Hi @erik-wramner 
I assume your folder is owned by the syslog user IN the container, right?

Cheers
Lukas

mindrunner:
After `chown -R syslog:root /var/log/mail/` it looks better. Cheers :)

erik-wramner:
@mindrunner yes it is in the container. Did you mount a volume or a host folder by the way? And are you using ONE_DIR or not? I wonder why it didn't work for you as it works for me.

mindrunner:
`ONE_DIR` yes
mounted a folder, instead of a volume

I think, I just had a permission problem and everything seems fine now.

erik-wramner:
Good, if it was a mounted folder we can drop it as a one-off.

helmutundarnold:
Hello Lukas,
you're right

drwxr-xr-x 1 syslog root      4096 Sep 22 06:31 mail

Regards
         Helmut

Am 21.09.19 um 16:53 schrieb lukas:

mindrunner:
still getting this error: 
```
/etc/cron.daily/logrotate:
error: skipping "/var/log/mail/clamav.log" because parent directory has insecure permissions (It's world writable or writable by group which is not "root") Set "su" directive in config file to tell logrotate which user/group should be used for rotation.
error: skipping "/var/log/mail/freshclam.log" because parent directory has insecure permissions (It's world writable or writable by group which is not "root") Set "su" directive in config file to tell logrotate which user/group should be used for rotation.
error: skipping "/var/log/mail/mail.log" because parent directory has insecure permissions (It's world writable or writable by group which is not "root") Set "su" directive in config file to tell logrotate which user/group should be used for rotation.
error: skipping "/var/log/mail/mail.info" because parent directory has insecure permissions (It's world writable or writable by group which is not "root") Set "su" directive in config file to tell logrotate which user/group should be used for rotation.
error: skipping "/var/log/mail/mail.warn" because parent directory has insecure permissions (It's world writable or writable by group which is not "root") Set "su" directive in config file to tell logrotate which user/group should be used for rotation.
error: skipping "/var/log/mail/mail.err" because parent directory has insecure permissions (It's world writable or writable by group which is not "root") Set "su" directive in config file to tell logrotate which user/group should be used for rotation.
```

Probably because my maillog folder itself had too insecure permissions. I changed it to the following now:

```
[le@nx]: ~/src/mail>$ ls -ahl maillog/
total 97M
dr-xr-xr-x 2 113 root  4.0K Sep 19 14:00 .
drwxr-xr-x 6 le  users 4.0K Sep 19 08:25 ..
-rw-r----- 1 113 root   21K Sep 22 19:03 clamav.log
-rw-r----- 1 113 root  7.8K Sep 21 14:01 freshclam.log
-rw-r----- 1 113 root   744 Sep 22 19:37 mail.err
-rw-r----- 1 113 root   755 Sep 15 08:51 mail.err.1
-rw-r----- 1 113 root   934 Sep  8 08:51 mail.err.2.gz
-rw-r----- 1 113 root   417 Aug 25 08:54 mail.err.3.gz
-rw-r----- 1 113 root   28M Sep 22 20:25 mail.info
-rw-r----- 1 113 root   25M Sep 15 08:51 mail.info.1
-rw-r----- 1 113 root  5.5M Sep  8 08:51 mail.info.2.gz
-rw-r----- 1 113 root  784K Aug 25 08:54 mail.info.3.gz
-rw-r----- 1 113 root   28M Sep 22 20:25 mail.log
-rw-r----- 1 113 root  4.6M Sep 15 08:51 mail.log.1
-rw-r----- 1 113 root  3.3M Sep 22 20:25 mail.warn
-rw-r----- 1 113 root  3.1M Sep 15 08:51 mail.warn.1
-rw-r----- 1 113 root  441K Sep  8 08:51 mail.warn.2.gz
-rw-r----- 1 113 root   57K Aug 25 08:54 mail.warn.3.gz

```


Waiting for the next rotate...

fbartels:
@mindrunner for testing purposes you can also force a log rotation https://www.shellhacks.com/logrotate-force-log-rotation/

erik-wramner:
You could also use a regular volume and copy/sync the files from there. Any particular reason you want to mount a folder?

mindrunner:
Yes, I do not like docker volumes for persisting data. (e.g. how do you backup a docker volume?) They are nothing else, than a folder, but one which i am not supposed to access from the outside. For me /var/lib/docker is never a persistent storage. Everything I put in there is data I am willing to lose. This includes volumes. Thus, my data, I want to keep and backup is mounted into directories.

erik-wramner:
Right. Well, I use volumes but I can understand your position. Good luck with the permissions.

mindrunner:
Since a volume is nothing else than a folder, permissions should not make any difference. I figured out, however, that the amavis/clamav logfiles go to `/var/log/mail` as well, but need another ownership (amavis) applied.

mindrunner:
Sorry for digging this out once again. I am getting this in my daily log summary now:
```
--------------------- clam-update Begin ------------------------

 No updates detected in the log for the freshclam daemon (the
 ClamAV update process).  If the freshclam daemon is not running,
 you may need to restart it.  Other options:

 A. If you no longer wish to run freshclam, deleting the log file
    (configured is /var/log/clamav/freshclam.log ) will suppress this error message.

 B. If you use a different log file, update the appropriate
    configuration file.  For example:
       echo "LogFile = log_file" >> /etc/logwatch/conf/logfiles/clam-update.conf
    where log_file is the filename of the freshclam log file.

 C. If you are logging using syslog, you need to indicate that your
    log file uses the syslog format.  For example:
       echo "*OnlyService = freshclam" >> /etc/logwatch/conf/logfiles/clam-update.conf
       echo "*RemoveHeaders" >> /etc/logwatch/conf/logfiles/clam-update.conf

 ---------------------- clam-update End -------------------------
```
I suppose, I missed out on somehting, right?

PR #1270, opened 2019-09-20
Separate environments and simplify docker-compose.yml

mindrunner:
There are two environment files now. The implicit .env file for docker-compose and the mail-specific configurations. The container needs the variables of the compose script as well, thus .env is imported into the container environment, too.


erik-wramner:
Would you mind changing the name of `.env-mailserver.dist` to `env-mailserver.dist` or something? I really hate it when important files are hidden.

It would also be nice if the readme explained what should go where, i.e. modify the domain name and container name in the `.env` file and keep all other settings in the other.

If possible I would prefer to delete the `.env` file and move the two remaining settings into the compose file instead, but perhaps that won't work? Not a big deal, but getting rid of the hidden file would be a plus!

mindrunner:
> Would you mind changing the name of .env-mailserver.dist to env-mailserver.dist or something? I really hate it when important files are hidden.
Sure thing.

> It would also be nice if the readme explained what should go where, i.e. modify the domain name and container name in the .env file and keep all other settings in the other.
Isn't that obvious by having example configs?

> If possible I would prefer to delete the .env file and move the two remaining settings into the compose file instead, but perhaps that won't work? Not a big deal, but getting rid of the hidden file would be a plus!
Surely, this would work, but i not my personal preference. I like to separate configs and sensitive data. But, its your baby, so I am happy to do it this way.

mindrunner:
Also, I think, `HOSTNAME` and `DOMAINNAME` are shared in both compose file and container. Thus, eliminating `.env` in favor of editing the compose file directly, will lead to the necessity of maintaining this variables in two places.

erik-wramner:
Looks good, I'll test the changes on one of my servers; then I'll merge.

erik-wramner:
@mindrunner, I tested it and it works well. I'll merge it, but I have one question. I used to have this in my compose file:
```
-    hostname: ${HOSTNAME}
-    domainname: ${DOMAINNAME}
-    container_name: ${CONTAINER_NAME}
```
It didn't work, but setting the options directly in the compose file works, i.e. `hostname: mail`. My question is, if we need to set the options manually in the compose file (and we do in the examples), what good is the `.env` file?

mindrunner:
> @mindrunner, I tested it and it works well. I'll merge it, but I have one question. I used to have this in my compose file:
> 
> ```
> -    hostname: ${HOSTNAME}
> -    domainname: ${DOMAINNAME}
> -    container_name: ${CONTAINER_NAME}
> ```
> 
> It didn't work, but setting the options directly in the compose file works, i.e. `hostname: mail`. My question is, if we need to set the options manually in the compose file (and we do in the examples), what good is the `.env` file?

Are you sure, that you have a file `.env` next to the `docker-compose.yml`? And it still does not work? That's odd. I will try to reproduce this here.

erik-wramner:
Hm, right. Perhaps it is because I'm using docker-compose from a docker container (no local installation on the host)? All the `.env` were propagated into the container, so it does find the file.

mindrunner:
can you provide a minimal example of our setup to reproduce this?

erik-wramner:
Yes, that is easy. I can just follow the setup instructions in the readme. Copying the templates gives me:
```
services:
  mail:
    hostname: ${HOSTNAME}
    domainname: ${DOMAINNAME}
    container_name: ${CONTAINER_NAME}
```
And:
```
HOSTNAME=mail
DOMAINNAME=domain.com
CONTAINER_NAME=mail
```
Starting the container and checking the host name:
```
docker exec mail cat /etc/hosts
# other lines
172.22.0.2	ef0581c1f85a.domain.com ef0581c1f85a
```
Editing the compose file:
```
  mail:
    hostname: mail
    domainname: domain.com
```
Gives:
```
172.22.0.2	mail.domain.com mail
```

erik-wramner:
```
Docker image docker/compose 1.24.0  4624ec04635e
docker-compose version 1.24.0, build 0aa59064
docker-py version: 3.7.2
CPython version: 3.6.8
OpenSSL version: OpenSSL 1.1.0j  20 Nov 2018
```
Running on Linux Mint (Ubuntu) 18.3.

mindrunner:
I can confirm your problem when starting the container from within another container by passing in the socket.
```
docker run  -ti --entrypoint "/bin/sh" -v /home/le/temp/compose-test/:/code/  -v /var/run/docker.sock:/var/run/docker.sock docker/compose:1.24.0
cd /code
docker-compose up
```
This is reproducible in both docker/compose and my own image runmymind/docker-with-compose. Both based on Alpine.

However, running compose directly from my local box behaves differently (and as I would expect).

Is this issue new since I refactored the env-import?

erik-wramner:
No and it is not a big deal. I mentioned it because the only things that are kept in the `.env` file now are settings that don't have any effect for me, so if it was the same for others we could have removed the file. Most people probably run compose directly on the host though.

PR #1284, opened 2019-10-07
Added optional file user-patches.sh for own patches without recompiling

vortex852456:
As @erik-wramner suggested in https://github.com/tomav/docker-mailserver/issues/1281 I created this pull request.

Sadly I wasn't able to test with the current tip of master. Error while starting:
`2019-10-07 20:45:23,793 INFO spawned: 'mailserver' with pid 8`
`supervisor: couldn't exec /usr/local/bin/start-mailserver.sh: ENOENT`
`supervisor: child process was not spawned`

This is my first pull request. Feel free to inform me whenever I was doing domething wrong.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ run:
 		-v "`pwd`/test/config":/tmp/docker-mailserver \
 		-v "`pwd`/test/test-files":/tmp/docker-mailserver-test:ro \
 		-v "`pwd`/test/onedir":/var/mail-state \
-		-v "`pwd`/test/config/user-patches/user-patches.sh":/tmp/docker-mailserver/user-patches.sh:ro \
+		-v "`pwd`/test/config/user-patches/user-patches.sh":/tmp/docker-mailserver/user-patches.sh \
 		-e ENABLE_CLAMAV=1 \
 		-e SPOOF_PROTECTION=1 \
 		-e ENABLE_SPAMASSASSIN=1 \
--- a/Makefile
+++ b/Makefile
@@ -31,6 +31,7 @@ run:
 		-v "`pwd`/test/config":/tmp/docker-mailserver \
 		-v "`pwd`/test/test-files":/tmp/docker-mailserver-test:ro \
 		-v "`pwd`/test/onedir":/var/mail-state \
+		-v "`pwd`/test/config/user-patches/user-patches.sh:/tmp/docker-mailserver/user-patches.sh:ro \
 		-e ENABLE_CLAMAV=1 \
 		-e SPOOF_PROTECTION=1 \
 		-e ENABLE_SPAMASSASSIN=1 \
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ run:
 		-v "`pwd`/test/config":/tmp/docker-mailserver \
 		-v "`pwd`/test/test-files":/tmp/docker-mailserver-test:ro \
 		-v "`pwd`/test/onedir":/var/mail-state \
-		-v "`pwd`/test/config/user-patches/user-patches.sh:/tmp/docker-mailserver/user-patches.sh:ro \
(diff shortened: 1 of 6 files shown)


Undercover1989:
The error does not seem to have anything to do with the changes I made. See https://github.com/tomav/docker-mailserver/issues/1285

erik-wramner:
I'm afraid that the error is caused by your change. I triggered a rebuild and got the same failure, but master builds. You should be able to run the tests locally if you have a Linux box (the tests don't work with Windows or Mac), see https://github.com/tomav/docker-mailserver/blob/master/CONTRIBUTING.md. In particular you need to init the submodules.

Undercover1989:
I've problems fixing this. When I start my self build mailserver, I get following unusal output from log:
`/tmp/docker-mailserver /`
`/`
`Starting Misc`
`user-patches hello world!` (this is an echo from my testing user-patches.sh)
`Error: Please fix your configuration. Exiting...`

When I comment out the line `/tmp/docker-mailserver/user-patches.sh` the server boots as expected. I dont't get it.

Undercover1989:
Now I'm totally confused. I've added `DMS_DEBUG=1` to my environment. With this the mailserver boots up...

`  Setting up configuration checksum file,`
`/tmp/docker-mailserver /,`
`  * Creating /tmp/docker-mailserver-config-chksum,`
`/,`
`  Checking /var/mail permissions,`
`  * Permissions in /var/mail look OK,`
`  Checking $amavis_state_dir permissions,`
`  * Permissions in /var/lib/amavis look OK,`
`  Cleaning up disabled Clamav,`
`  Cleaning up disabled spamassassin,`
`Starting Misc,`
`  * Executing user-patches.sh,`
`user-patches hello world!,`
`  * user-patches.sh executed,`
`Starting mail server`

erik-wramner:
I'll look into it when I get some free time. The CI tests still fail on the fixtures though, so something must be seriously wrong.

Undercover1989:
I've tested `make build-no-cache generate-accounts run generate-accounts-after-run fixtures tests clean` with commit https://github.com/tomav/docker-mailserver/pull/1284/commits/275a83667a76fef1046fa3306a09f72aa4686352 and it was successful on my machine (debian-buster). But the Travis CI build is still failing.

vortex852456:
Great! I'm taking care of it this or next weekend. Thanks for your effort!

erik-wramner:
Looks good so far, I'll wait for the test before merging. Have a nice weekend!

vortex852456:
Hope you had two nice weekends. Sorry for the delay.

Can I add the new test to https://github.com/tomav/docker-mailserver/blob/master/test/tests.bats ? Or where would you place it?

I must admit that implementing the test is a not so easy step for me. Have to learn and understand how it works before I can start. If you could give me some coarse instructions it may help me a lot on getting this implemented faster.

erik-wramner:
Well, either in test.bats or in one of the new mail_something.bats. We are trying to break up the tests to avoid having to start all the  docker images up front.

In your case you could use test.bats and add a new function in the "configuration checks" section. Make sure that an example script is mounted into one of the containers (for example mail) in the `Makefile` but not into all containers. Then add a new test, something like (pseudo code):
```

@test "checking configuration: user-patches.sh executed" {
  run docker logs mail | grep expected_output_from_user_patches
  assert_output "expected output..."
}

```

vortex852456:
Argh, the new test doesn't pass. I guess I have to set up an local IDE to examine this further. You will hear from me :-) Thanks so far!

vortex852456:
image

sebka:
@erik-wramner pls have a look @ the new test.
i would really like to know why bats could not match those two strings.

erik-wramner:
Hm, strange. @fbartels do you have any idea?

At least I merged the PR. Thanks for taking the time!

fbartels:
Hmm, the only thing I could think of are trailing spaces.

PR #1297, opened 2019-10-17
Removing unnecessary maildir folders creation

gmasse:
According to Dovecot configuration `/etc/dovecot/conf.d/15-mailboxes.conf`
`Drafts`, `Junk`, `Trash` and `Sent` folders will be created on the fly (reference documentation: https://wiki.dovecot.org/MailboxSettings)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,7 +8,6 @@ insert_final_newline = true
 
 [*.md]
 indent_style = space
-indent_size = 4
 trim_trailing_whitespace = false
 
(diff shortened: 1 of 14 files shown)


erik-wramner:
Looks good except for one thing. We are not removing the folders - we just want Dovecot to create them for us, right? Ideally I'd like to keep the old checks in the test and create the folders by making a Dovecot connection to the server from the test. Then the test verifies that the auto-creation works as intended.

PR #1311, opened 2019-11-01
Removing unnecessary maildir folders creation

gmasse:
Test added. See PR #1297

_According to Dovecot configuration /etc/dovecot/conf.d/15-mailboxes.conf
Drafts, Junk, Trash and Sent folders will be created on the fly (reference documentation: https://wiki.dovecot.org/MailboxSettings)_


erik-wramner:
Looks great, thanks!

Issue #1313, opened 2019-11-03
Add support for Dovecot sdbox and mdbox format


gmasse:
Current release only support maildir format.
Add a new option `DOVECOT_MAILBOX_FORMAT`to set Dovecot `mail_location` accordingly.
PR #1314

gmasse:
Merged

PR #1314, opened 2019-11-03
New option DOVECOT_MAILBOX_FORMAT

gmasse:
Add a new option `DOVECOT_MAILBOX_FORMAT`to set Dovecot `mail_location` accordingly.
Issue #1313

### DIFF
--- a/README.md
+++ b/README.md
@@ -357,6 +357,13 @@ Set the message size limit for all users. If set to zero, the size will be unlim
   - drop => Drop the connection immediately with a 521 SMTP reply. Repeat this test the next time the client connects.
   - ignore => Ignore the failure of this test. Allow other tests to complete. Repeat this test the next time the client connects. This option is useful for testing and collecting statistics without blocking mail.
 
+##### DOVECOT_MAILBOX_FORMAT
+
+  - **maildir** => uses very common Maildir format, one file contains one message
+  - sdbox => (experimental) uses Dovecot high-performance mailbox format, one file contains one message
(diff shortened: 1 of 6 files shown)


erik-wramner:
Looks good, two very minor things. One is the indentation for the case/esac block in the start script. I guess it mixes tabs and spaces because the code doesn't line up. Not a big deal but easy to fix.

The other is the readme. Perhaps you could add a note that the two new formats were added to docker-mailserver in November 2019 and should be considered experimental until they have been tested in the wild for a while? I don't expect problems, but one never knows and then at least prospective new users have been warned.

gmasse:
I agree @erik-wramner, I will fix and add the comment.
Why not creating a `Testing` branch? `Master` would continue to be fed with bugfixes and minor changes when `Testing` would integrate big changes and enhancements. My 2 cents ;)

erik-wramner:
We could do that, but most of the work consists of fixes that everyone should get and then `latest` followed by `stable` works well. The next big thing is the upgrade to buster and there is a separate branch for that.

gmasse:
Grr I am fighting to correctly push my updates...

erik-wramner:
Are you done, or still fighting?

gmasse:
PR is good now.
Test has failed; I suspect a race condition or something like that. I cannot easily identify the root cause.
@erik-wramner can you relaunch the build?

erik-wramner:
Restarted. That error is not one of the frequent ones, but I have seen it before. Let's see what happens!

erik-wramner:
It worked, I merged the changes. Thanks!

PR #1318, opened 2019-11-08
Update start-mailserver.sh

iNDicat0r:
Argument type should be -e and not -P

### DIFF
--- a/target/start-mailserver.sh
+++ b/target/start-mailserver.sh
@@ -1161,7 +1161,7 @@ function _setup_postfix_override_configuration() {
 	if [ -f /tmp/docker-mailserver/postfix-master.cf ]; then
 		while read line; do
 		if [[ "$line" =~ ^[0-9a-z] ]]; then
-			postconf -P "$line"
+			postconf -e "$line"
 		fi
 		done < /tmp/docker-mailserver/postfix-master.cf
 		notify 'inf' "Loaded 'config/postfix-master.cf'"


erik-wramner:
I think you are wrong. Check http://www.postfix.org/postconf.1.html. For updating `master.cf` parameters it should be `-P`. For `main.cf` it should be `-e`. The block you have changed is for `master.cf`.

iNDicat0r:
> I think you are wrong. Check http://www.postfix.org/postconf.1.html. For updating `master.cf` parameters it should be `-P`. For `main.cf` it should be `-e`. The block you have changed is for `master.cf`.

The -P option doesn't accept $line of type "key=value" and therefore when you try to load a master.cf from the `config` folder it throws an error.

I tried to change some values of postfix and had them under `postfix-master.cf` and then tried to inspect the logs and saw that postfix -P $line was throwing errors.

erik-wramner:
Right. Perhaps we need both `-P` and `-e` or something, I need to check. That code is quite old though, really surprising if it has been wrong all that time. Of course, not many are overriding those options.

iNDicat0r:
> Right. Perhaps we need both `-P` and `-e` or something, I need to check. That code is quite old though, really surprising if it has been wrong all that time. Of course, not many are overriding those options.

The error I was getting was: 

```
postconf -Pe or -PX requires service_name/type/parameter
```

and i had the following value in my postfix-master.cf:
```
smtpd_client_connection_count_limit = 500
```

erik-wramner:
This is poorly documented, but I think you are simply using the wrong format. The postfix-master.cf does not use `name=value`, it uses `service_name/type/name=value`. For example `127.0.0.1:10025/inet/smtpd_soft_error_limit=1002` results in `postconf -P 127.0.0.1:10025/inet/smtpd_soft_error_limit=1002` which works.

If you enter the container (`docker exec -it mail /bin/bash`) and run `postconf -P` it will show the active options with the correct format except that there are spaces that must be removed.

erik-wramner:
I've updated the documentation in the FAQ. Are you happy with that?

iNDicat0r:
Alright thank you

PR #1323, opened 2019-11-12
fix: dont use TTY in docker if not present, rem TTY necessity in tests

ap-wtioit:
When running tests in Gitlab CI they fail because currently docker commands are hardcoded with `-ti` argument. This gives the following errors:
```
not ok 182 checking amavis: VIRUSMAILS_DELETE_DELAY override works as expected
# (from function `assert_success' in file test/test_helper/bats-assert/src/assert.bash, line 114,
#  in test file test/tests.bats, line 705)
#   `assert_success' failed
# 
# -- command failed --
# status : 1
# output : the input device is not a TTY
# --
# 
not ok 183 checking amavis: old virusmail is wipped by cron
# (from function `assert_success' in file test/test_helper/bats-assert/src/assert.bash, line 114,
#  in test file test/tests.bats, line 711)
#   `assert_success' failed
# 
# -- command failed --
# status : 1
# output : the input device is not a TTY
# --
# 
not ok 184 checking amavis: recent virusmail is not wipped by cron
# (from function `assert_success' in file test/test_helper/bats-assert/src/assert.bash, line 114,
#  in test file test/tests.bats, line 719)
#   `assert_success' failed
# 
# -- command failed --
# status : 1
# output : the input device is not a TTY
# --
# 
```

This fix only passes `-ti` if a TTY was detected in setup.sh and skips the `-ti` in the failing tests where it is not necessary.

Info @wt-io-it

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -98,10 +98,14 @@ _docker_image_exists() {
   fi
 }
 
+if [ -t 1 ] ; then
+  USE_TTY="-ti"
+fi
+
 _docker_image() {
   if [ "$USE_CONTAINER" = true ]; then
     # Reuse existing container specified on command line
-    docker exec -ti "$CONTAINER_NAME" "$@"
+    docker exec ${USE_TTY} "$CONTAINER_NAME" "$@"
   else
     # Start temporary container with specified image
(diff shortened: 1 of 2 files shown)


erik-wramner:
Makes sense.

PR #1324, opened 2019-11-12
fix: fix test setup.sh debug fetchmail on wildcard domains

ap-wtioit:
on hosts that belong to wildcard domains pop3.example.com might
 actually resolve to pop3.example.com.[mydomain.com] and give a valid ip
 the return code of fetchmail then no longer is 11 (dns failure) but
 something else (2 for socket error in our case)

to make sure we always get return code 11 we use the domain name
pop3.example.com. that is not allowed to be resolved to a subdomain.

Info @wt-io-it


erik-wramner:
Looks good, thanks.

Issue #1322, opened 2019-11-12
Will docker-mailserver support podman?


OrwillT:
RHEL/CentOS 8 provide podman and buildah as the default container tools. So will docker-mailserver support podman?

fbartels:
Hi @OrwillT,

the whole podman eco-system is meant to be compatible with the Docker world, so I would suspect for it to work. But the only way to really find out is for you to try and then lets us know.

It should either fail early or not at all.

OrwillT:
And the answer is no.
I can't deploy the docker-compse file with [podman-compose](https://github.com/containers/podman-compose), and also can't use setup.sh to visit container.

erik-wramner:
This is a community effort and the docker image itself might run even if the compose scripts (which are only provided as a template anyway) fail. If you can make it work we'll be happy to consider a PR and/or to update the documentation! Personally I don't intend to use podman any time soon, though.

OrwillT:
I believe I'll give it a try when I could as soon as possible.

OrwillT:
```
Nov 21 19:11:53 los00 amavis[261]: Net::Server: Group Not Defined.  Defaulting to EGID '111 111'
Nov 21 19:11:53 los00 amavis[261]: Net::Server: User Not Defined.  Defaulting to EUID '109'
```

Is this correct?
Not only amavis have this issue, opendmarc, opendkim, dovecot have this user/group not defined issues too.

erik-wramner:
I get that error as well with docker. I don't like it, but as you say everything seems to work regardless.

OrwillT:
@erik-wramner  @fbartels  #1327 Here it is.

PR #1327, opened 2019-11-22
Add podman support to setup.sh

OrwillT:
As what I want in #1322, New setup.sh could automatically decide which CRI should it use and work normally with podman. 

I will write a new wiki about how to manually install docker-mailserver with podman CLI and systemd tomorrow to make this function have minimum support.

Then I will slowly write a standalone install&update management script for podman and a kubernetes YAML for autodeploy.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -1,12 +1,32 @@
-#! /bin/sh
+#!/usr/bin/env bash
 
 ##
 # Wrapper for various setup scripts included in the docker-mailserver
 #
 
-INFO=$(docker ps \
+CRI=
+
+_check_root() {
+  if [[ $EUID -ne 0 ]]; then
+    echo "Curently docker-mailserver doesn't support podman's rootless mode, please run this script as root user." 
+    exit 1
+  fi  
+}
+if [ -z "${CRI}" ]; then
+  if [ ! -z "$(command -v docker)" ]; then
+    CRI=docker
+  else if [ ! -z "$(command -v podman)" ]; then
+    CRI=podman
+    _check_root
+  else do
+    echo "No Supported Container Runtime Interface Detected."
+    exit 1
+  fi
+fi
+
+INFO=$(${CRI} ps \
   --no-trunc \
-  --format="{{.Image}}\t{{.Names}}\t{{.Command}}" | \
+  --format="{{.Image}} {{.Names}} {{.Command}}" | \
   grep "supervisord -c /etc/supervisor/supervisord.conf")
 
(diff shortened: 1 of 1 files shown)


erik-wramner:
I haven't worked much with podman, but isn't it common to create an alias for docker? And wouldn't `command -v docker` succeed in that case even if it uses podman?

OrwillT:
> I haven't worked much with podman, but isn't it common to create an alias for docker? And wouldn't `command -v docker` succeed in that case even if it uses podman?
> 
> EDIT: and why do we need podman to run as root? The way I see it the only real advantage with podman is that it can run as a non-privileged user. Would it be possible to run as a non-root user by not using some features (such as fail2ban) or are we stuck with root?

Podman is compatible with most of docker's features, but it's not just an alias for docker. ` command -v docker ` won't succeed if you only install podman. Even if you alias podman to docker or install a stupid package named ` podman-docker ` which basiclly work the same as alias, this script will still work normally. The only situation this script won't work is someone install both docker and podman, and try to install mailserver to podman. In this case he need to manually edit setup.sh and set variable ` CRI ` to podman, which I will mentioned in the tutorial.

Rootless is great, but the mapping under port 1024 must have root privileges, and the default podman version Centos 8 provide is 1.0.5 release around february which is too old, and it only have minimum support now. I'll create another pull requset after I test all the advanced feature like manage container with pod like k8s, rootless, and autodeploy.

erik-wramner:
OK fine. One thing though. I noticed you changed `#/bin/sh` into  `#!/usr/bin/env bash`. I don't like that. It is likely that bash is available, but not 100% certain (this is on the host, not in the container) and even if it is there may not be a `/usr/bin/env` present. Can you change that, please?

OrwillT:
Done.

erik-wramner:
Merged, thanks!

oblitum:
> Then I will slowly write a standalone install&update management script for podman and a kubernetes YAML for autodeploy.

@OrwillT, did you manage to port it to kubernetes YAML?

alexanderadam:
Thank you for your work @OrvilleQ! :raised_hands: 

> Rootless is great, but the port binding under 1024 must have root privileges

Is this for within the container or on the host mapping side? The port within the container image shouldn't matter (since it can be mapped at the host), right?
And on the outside only the `CAP_NET_BIND_SERVICE` capability should be needed, right?

I'm still new to podman so it could be that I totally misunderstood the whole topic. :wink: 

> I'll create another pull requset after I test all the advanced feature like manage container with pod like k8s, rootless, and autodeploy.

Did you have any time to look into the rootless functionality?

PR #1334, opened 2019-12-09
Add documentation example for Goolgle Mail users

hack2root:
Added command which helps people to work with GMail servers properly (If you compose email using Web client, it uses and alias for your email address, <user>, not FQDN in form <user@domain>)

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,10 @@ Download the docker-compose.yml, the .env and the setup.sh files:
 #### Create your mail accounts
 
     ./setup.sh email add <user@domain> [<password>]
+    
+#### GMail aliases for mail accounts (in case you encounter error 550 5.1.1)
+
+    ./setup.sh alias add <user@domain> <user>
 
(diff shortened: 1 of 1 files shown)


erik-wramner:
I wonder about two things. The first is "GMail aliases for mail accounts", the command creates a normal alias which really has nothing to do with GMail? I'd prefer to remove GMail from the header and instead provide a short description of what the alias can solve.

That brings me to the second point which is the problem itself. Can you elaborate a bit? If I create an e-mail account `some.user@example.com` with an identical alias `some.user@example.com` there should be a conflict. How do you define your user and your alias? If for example the user is created as `some.user@mail.example.com` then an easier solution would be to simply change that.

In short I can merge this but I would like to understand it first and possibly add some text to make sure others understand it as well.

erik-wramner:
@hack2root?

erik-wramner:
Closing this due to no response.

PR #1343, opened 2019-12-20
Remove lower-case check for postfix-main.cf

julakali:
The check excludes any line with uppercase letters, such as cipher lines. Remove it because it is pretty rudimentary anyways.

I tried to add the following config line by using postfix-main.cf:
smtpd_tls_exclude_ciphers = aNULL,MD5,DES,3DES
It does not get applied because it fails the regex check.

### DIFF
--- a/target/start-mailserver.sh
+++ b/target/start-mailserver.sh
@@ -1152,11 +1152,7 @@ function _setup_postfix_override_configuration() {
 
 	if [ -f /tmp/docker-mailserver/postfix-main.cf ]; then
 		while read line; do
-		# all valid postfix options start with a lower case letter
-		# http://www.postfix.org/postconf.5.html
-		if [[ "$line" =~ ^[a-z] ]]; then
-			postconf -e "$line"
-		fi
+		postconf -e "$line"
 		done < /tmp/docker-mailserver/postfix-main.cf
 		notify 'inf' "Loaded 'config/postfix-main.cf'"
 	else


erik-wramner:
I don't think we should remove the check. There may be empty lines and comments and whatnot in existing configuration files that are currently ignored. Removing the check will apply them and may break existing installations.

Furthermore the check is (almost) correct as described in the comment. It requires the option to START with a lower-case letter. In your case `smtpd_tls_exclude_ciphers` starts with `s`, so it should match. There is actually an option that starts with a digit (`2bounce_notice_recipient`), but your option should work.

erik-wramner:
I wrote a very simple test program to play around with the regex:
```
#!/bin/bash
if [[ "$1" =~ ^[a-z] ]]; then
  echo yes
else
  echo no
fi
```
As expected it responds with yes when I pass in your config line; it matches. If your change really solved your problem there must be something else that is wrong. Perhaps you have garbage in the config file (Windows CRLF line endings or a UTF-8 byte order mark for example)?

julakali:
Thanks for testing the regex again.
As I was trying to find a solution, I removed the check and it worked for me - but i also changed the postfix-main.cf.
The real problem was that my postfix-main.cf only contains 1 line - and **no** EOL sequence (written with vscode).
Thus, `read line` returned false (exit status of read is non-zero if EOF).

I'm sorry for this mistake. You could use `while read line || [ -n "$line" ]; do` instead to support lines with no EOL sequence. But i guess it's a rare case, so I'm fine.

erik-wramner:
Right, I'll close this then. Merry Christmas!

PR #1363, opened 2020-01-11
Integrated OpenARC

mindrunner:


erik-wramner:
Thanks, looks good!

mindrunner:
Does this have a docker tag automatically now as well? Also, can I get notified about changes in master, so that I can rebase the ARC branch?

erik-wramner:
Well, this was disappointing. I had hoped that there would be an arc image automatically, but the docker auto-build does not seem to be configured that way and I don't have the credentials to fix it. Only @tomav  does, but I've been trying to get help with a similar change (scheduled builds) for a long time, so I don't know when he has time to help. I'll write an issue about it and hope for the best.

I don't think we should rebase the arc branch as that will affect others who have used it, I'll try to remember to cherry-pick or merge changes from master into arc instead.

mindrunner:
> I don't think we should rebase the arc branch as that will affect others who have used it, I'll try to remember to cherry-pick or merge changes from master into arc instead.

Well, if you prefer to merge instead of rebase, then for sure we can merge. I like rebases more, because of the clean history. But that's more a religious question. My request was not about the how, but rather about the when and how do I know! ;)

erik-wramner:
I also prefer rebase for my own private branches or when I share with a few known developers, but for public branches it can be troublesome. I'll try to remember to mention you when I do a merge, so that you'll get a notification. Alternatively I can cherry-pick changes to arc from time to time if you like. We don't have that many changes after all.

mindrunner:
Makes sense! :) Cheers. Already cherry-picked the fixes for https://github.com/tomav/docker-mailserver/issues/1366 into the `arc` branch :)

PR #1119, opened 2019-01-29
activate SA shortcircuit features via env, fixes #1118

phish108:

PR #1372, opened 2020-01-14
activate SA shortcircuit features via env, fixes #1118 (again)

### LINKED ISSUE #1118: Allow to use Spamassassin's shortcircuit bayes_99 and bayes_00 feature
Spamassassin allows administrators to give their Bayes databases greater emphasis by short circuiting the analytical process. 

This is normally configured in /etc/spamassassin/local.cf at the very end. However, it is rather annoying, to edit this file with every update of the container.

Having two environment variables to configure this would be very helpful.

Proposal: 

- ```SA_SHORTCIRCUIT_BAYES_SPAM``` - if set to 1, then it will activate ```shortcircuit bayes_99 spam```
- ```SA_SHORTCIRCUIT_BAYES_HAM```- if set to 1, then it will activate ```shortcircuit bayes_00 ham```

phish108:
This is the requested minimised PR based on the current master. This replaces #1119

### DIFF
--- a/README.md
+++ b/README.md
@@ -474,6 +474,22 @@ Note: this spamassassin setting needs `ENABLE_SPAMASSASSIN=1`. By default, the m
 
 Note: this spamassassin setting needs `ENABLE_SPAMASSASSIN=1`
 
+##### SA_SHORTCIRCUIT_BAYES_SPAM
+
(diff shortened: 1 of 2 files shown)


erik-wramner:
Thanks.

PR #1373, opened 2020-01-15
remove not needed log files after build

mindrunner:
remove not needed log files after build


erik-wramner:
The test failed in an unanticipated way, I'm rerunning it now.

Concerning the change set this looks a bit dirty. The merge approach is not so clean after all. Perhaps it is better if I merge master into arc on changes so that you can just rebase/pull arc and create PRs for your own changes? The disadvantage is that I need to fix any merge conflicts, but having other people's changes look like yours in this way is not very good.

mindrunner:
> The test failed in an unanticipated way, I'm rerunning it now.\

Running locally as well. Looked good to me before.

> Concerning the change set this looks a bit dirty. The merge approach is not so clean after all. Perhaps it is better if I merge master into arc on changes so that you can just rebase/pull arc and create PRs for your own changes? 

Yes, I think so too. I just cherry picked everything out of master including my own change based on the discussion we had before. Let's try your approach. :)

> but having other people's changes look like yours in this way

Was never my intention!

erik-wramner:
The tests passed this time, so I'll merge this.

Issue #1376, opened 2020-01-21
Incorrect permissions for sieve-pipe scripts


tweibert:
I'm using a sieve-pipe script to pipe incoming mail into an osTicket instance (osticket_automail.pl). As per the documentation, the script is placed into `config/sieve-pipe`, from where it gets copied over to `/usr/lib/dovecot/sieve-pipe` in the running container.

In the `_setup_dovecot()` function in `start-mailserver.sh`, there are these two lines that "fix" the permissions on all sieve configuration:

	chown docker:docker -R /usr/lib/dovecot/sieve*
	chmod 550 -R /usr/lib/dovecot/sieve*

Unfortunately this removes the executable flag on the sieve-pipe script, and dovecot will no longer call it, giving an error like this:

> Jan 21 16:56:17 mail dovecot: lmtp(xxx@xxxxx.com)<26992><cEiLHSctJ15waQAA/p0Tyw>: Error: program exec:/usr/lib/dovecot/sieve-pipe/osticket_automail.pl (28697): Forcibly terminated with signal 15

My workaround is to fix the permissions in `user-patches.sh`:

    chmod +x /usr/lib/dovecot/sieve-pipe/*

I think the issue can be easily solved by adding this line into `start-mailserver.sh` as well.

tweibert:
Added PR with my suggested solution.

DuncanvR:
@tweibert, since your PR has been merged, can this issue be closed?

tweibert:
Oh yes, sure. Thank you.

PR #1377, opened 2020-01-21
Add executable flag for scripts in /usr/lib/dovecot/sieve-pipe

tweibert:
Fixes missing executable permissions for sieve-pipe scripts (see #1376).


erik-wramner:
The tests have failed twice. I don't think it is due to your change, but I want them to succeed. I'll start them again.

Meanwhile, what happens if there are no files at all in the directory? Does the command fail, or doesn't it matter?

tweibert:
Good point about the empty directory. Added the -f flag to suppress any error message in that case.

Let me know if there's any issue with the checks.

PR #1391, opened 2020-02-10
Enabling Multi-Domain-Support in saslauthd

theoneandonly-vector:
Enabling Multi-Domain-Support.
! Needs to be mentioned in "README.md" as an important change! -> users won't be able to just use "username" anymore.

-> Combine the realm with the login (with an '@' sign in between). e.g. login: "foo" realm: "bar" will get passed as login: "foo@bar".


erik-wramner:
@theoneandonly-vector the tests fail. I have restarted them, so it is not just a transient error, something is broken. Please check.

Apart from that I'm not willing to add a breaking change just like that. You'll need to find a way to make it configurable. If it is possible to patch the supervisor files from the start script (or from user-patches.sh) that is a better option; it can then be turned on and off. Not sure if supervisor re-reads the configuration file though.

PR #1407, opened 2020-02-21
Allow calling setup.sh from other scripts

casperklein:
Calling setup.sh from other scripts fails, for example when adding a new mailbox:

    the input device is not a TTY

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -127,7 +127,7 @@ _docker_image_exists() {
   fi
 }
 
-if [ -t 1 ] ; then
+if tty -s ; then
   USE_TTY="-ti"
 fi
 


erik-wramner:
I'll merge this as the tests pass (sorry for being slow), but can you point to any resource that shows the difference between the built-in -t option and the tty command? I'd like to understand why tty is better than -t.

casperklein:
I noticed, that I was a little bit wrong. However the change is still valid. Let me explain:

The problem was not that my script was calling setup.sh, but my script was not attached to a TTY when used like:

`migrate.sh < users.txt`.

Without using `< users.txt` there would be no problem.

[`[ -t FD ]` `True if file descriptor FD is open and refers to a terminal.`](http://tldp.org/LDP/Bash-Beginners-Guide/html/sect_07_01.html)

`-t` only checked if there is a terminal, but not if there is a TTY attached.

`docker -ti` however expects a TTY.

Thats why `-t 1` failed in my setup. `USE_TTY` was wrongly set, even if there was no TTY. Thats why the docker command failed with: `the input device is not a TTY`.

`tty -s` checks if there is a TTY and is therefor the correct test for the `USE_TTY` variable.

PR #1409, opened 2020-02-22
Missing options from README.md added.

casperklein:


erik-wramner:
Thanks and sorry for the very slow feedback, I haven't had time. I will gladly merge this except that I don't think the old options are deprecated. They are still supported and can be used going forward for people who prefer to have a single option that controls all their reports. Can you just remove the "(deprecated)" parts, please?

casperklein:
I just copy pasted from the current `README.md`, where these options are stated as deprecated.

This was introduced with https://github.com/tomav/docker-mailserver/commit/008b8e6bce0a8e146b0eb2c67d1dcfe5b71f0cf8

> Can you just remove the "(deprecated)" parts, please?

No problem, but then we should also adjust the `README.md`. What do you think?

erik-wramner:
OK, we'll run with this then.

PR #1427, opened 2020-03-13
Allow to set comfortably inet_protocols

Yajo:
Setting `inet_protocols = ipv4` is almost a requirement when running behind Docker. Provide a way to make it easy.

@Tecnativa TT22925

### DIFF
--- a/README.md
+++ b/README.md
@@ -304,6 +304,8 @@ Set different options for mynetworks option (can be overwrite in postfix-main.cf
   - network => Add the docker default bridge network (172.16.0.0/12); **WARNING**: `docker-compose` might use others (e.g. 192.168.0.0/16) use `PERMIT_DOCKER=connected-networks` in this case
   - connected-networks => Add all connected docker networks (ipv4 only)
 
+Note: you probably want to [set `POSTFIX_INET_PROTOCOLS=ipv4`](#postfix_inet_protocols) to make it work fine with Docker.
+
 ##### VIRUSMAILS_DELETE_DELAY
 
 Set how many days a virusmail will stay on the server before being deleted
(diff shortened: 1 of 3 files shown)


Yajo:
Please could you help with travis? I don't understand those logs... :confused:

erik-wramner:
I restarted the test, but that kind of error indicates a major problem of some sort. It didn't even start to run the tests but failed very early.

Yajo:
Then what should I do?

erik-wramner:
The best would be if you could reproduce locally. However, I'm curious as to why you need this? Most if not all seem to be running fine without it. My docker image does not have ipv6 addresses even though the host sports them. Are you using host networking for some reason?

Yajo:
No, I'm using docker networks inside a docker container. However, I found some logs of a blocked server that was attempting to contact `smtp.gmail.com` with ipv6, although the container itself didn't have any ipv6 networks. However, if you run `postconf -p mynetworks`, you'll see some ipv6 networks that seem to make postfix believe it can use ipv6 normally.

Adding this setting saved my day, so I just wanted to make it easier for everyone in my situation. I use this image just as a local relay in this scenario.

Actually I believe ipv4-only should be the default, but I decided to not touch the default to avoid breaking others'.

I found a bug, let's see if CI works now.

Yajo:
:heavy_check_mark: now.

erik-wramner:
Great. One small request though. The current code sets the value "all" by default. That is not strictly backwards compatible. The old code did not touch the setting. Small thing but it can make a difference when it interacts with people's custom configuration.

Can you add an empty (or "default") option that is the default and skip the postconf command entirely unless the new option is set, please? Otherwise it looks fine.

Yajo:
Well I just imitated the way it was done for other settings. If someone overwrites the setting in `postfix-main.cf` it should work just like with any other, isn't it? :thinking:

casperklein:
Something like that in `target/start-mailserver.sh`:

```
if [ "$POSTFIX_INET_PROTOCOLS" != "all" ]; then
	_register_setup_function "_setup_inet_protocols"
fi
```

This ensures, that nothing is changed in a already running setup.

Yajo:
OK, done.

PR #1434, opened 2020-03-20
Filebeat in its own container

gmasse:
### Implementation of Filebeat in its own container

Filebeat configuration is done through _Hints based autodiscover_ (following [Elastic](https://www.elastic.co/guide/en/beats/filebeat/current/configuration-autodiscover-hints.html) practice).
Nevertheless, I have decided to change the default behavior (logs from all containers) in order to only log from mailserver (I guess some people can run other containers and do not want to push all the logs to filebeat).

**Filebeat with local ELK container:**
- `docker-compose.elk.yml.dist`: updated with the new filebeat container

**Filebeat forwarding to remote ELK:**
- `docker-compose.filebeat.yml.dist`: new file for those who only need filebeat (remote elk)
- `config/filebeat.docker.yml` to update with ELK server IP

Any references to filebeat previous implementation are removed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -119,17 +119,6 @@ RUN \
   rm -f /etc/postsrsd.secret && \
   rm -f /etc/cron.daily/00logwatch
 
-# install filebeat for logging
-# SKIP and run in an external container instead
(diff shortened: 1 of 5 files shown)


erik-wramner:
Thanks, looks good.

PR #1448, opened 2020-04-06
Allow change of mail_max_userip_connections through variable

michaeljensen:
I found that I was exceeding the default max logins per IP (which is 10) with 4 devices logging in at once from a home.  This allows for the config setting to be changed through an environment variable.


erik-wramner:
This is already supported using the dovecot configuration file in `config/dovecot.cf`. In fact if you check that file your exact use case is used as an example. Simply uncomment the line:
`
#mail_max_userip_connections = 50
`
That should work. If not we need to investigate why. We can't expose all the Postfix and Dovecot options with environment variables, so where it is possible it is better to use the configuration file overrides.

If this works for you please close the PR. If not let's discuss it! Regardless thanks for taking the time to get involved.

target/start-mailserver.sh
```diff
@@ -586,6 +587,12 @@ function _setup_dovecot() {
 	sed -i -e 's/#ssl = yes/ssl = required/g' /etc/dovecot/conf.d/10-ssl.conf
 	sed -i 's/^postmaster_address = .*$/postmaster_address = '$POSTMASTER_ADDRESS'/g' /etc/dovecot/conf.d/15-lda.conf
 
+	# Set mail_max_userip_connections
+	if [[ ! -z ${DEFAULT_VARS["DOVECOT_MAX_USERIP_CONNECTIONS"]} ]]; then
```
casperklein, line 591: Default value is 10. So this test is always true.
Better would be: `if [ ${DEFAULT_VARS["DOVECOT_MAX_USERIP_CONNECTIONS"]} != "10" ]; then`

michaeljensen:
Thank you for the kind reply.
I was trying to make it so that the container wouldn't have to be rebuilt in order to implement this config change.

It makes sense to not expose all of the postfix & dovecot config options through environment variables.

target/start-mailserver.sh
michaeljensen: Good catch!  Thanks!

michaeljensen: Do you think it would be better to make your change or remove the line from the DEFAULT_VARS section altogether?

erik-wramner:
@michaeljensen you don't have to rebuild the container, you simply need to restart it. See https://github.com/tomav/docker-mailserver/wiki/Override-Default-Dovecot-Configuration. Don't get confused by the "Override configuration" section, everything you need to do for this simple change is to uncomment the existing line in `config/dovecot.cf` and restart (or do it before starting). That is all.

In short your change should be supported without rebuilding the container by simply editing the bundled example file which is intended for just this kind of changes.

target/start-mailserver.sh
casperklein: @erik-wramner is right, I don't think we need that configurable via a variable. I just wanted you to know that there is a logical error in your code.
Regarding your last question, I would stay with the default variable defined. That way, it's easier to keep track, which variables are used in the script.

erik-wramner:
As this is still open, to be even more clear: in the installation right after the curl commands that download files, create a directory named config in the folder where you have the other files. Check the templates in https://github.com/tomav/docker-mailserver/tree/master/config and download the files you think you need and put them in config.

In your case you only need https://github.com/tomav/docker-mailserver/blob/master/config/dovecot.cf. Edit the file. In your case all you need to do is to remove the # for the line with mail_max_userip_connections. Save the file. Restart the container:

`
docker-compose down && docker-compose up -d mail
`

You are done. If that works we can close this? If not please ask. It should work.

michaeljensen:
I finally had some time to test out the dovecot.cf file and it works as expected.  I like that way of managing this config much better than what I was proposing in this pull request.
In the process, I discovered the user-patches.sh which seems like a good "catch all" for any other custom modifications.

Thank you for the info and the help on this!

PR #1450, opened 2020-04-10
Upgrade packages

casperklein:
Some packages from the base image are upgradable. For example, that's the case for `libgnutls30` at the moment.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,7 +28,7 @@ ENV SASLAUTHD_MECH_OPTIONS=""
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
 # Packages
-# hadolint ignore=DL3015
+# hadolint ignore=DL3015,DL3005
 RUN \
   apt-get update -q --fix-missing && \
   apt-get -y upgrade && \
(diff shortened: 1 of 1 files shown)


erik-wramner:
The tests fail with:
`Dockerfile:32 DL3005 Do not use apt-get upgrade or dist-upgrade
`

casperklein:
From https://github.com/hadolint/hadolint/wiki/DL3005:
> Updates will be baked into the based images you don't need to apt-get upgrade your containers. 

I didn't know that, however that seems not the case for all images out there. Debian seems to update their images not that often. The last image update for buster-slim was 10 days ago.

erik-wramner:
And now it is 11 days old. Perhaps we should do the update anyway. @fbartels and @tomav, what do you think?

tomav:
Can we assume that if it's available as Debian package, it's production ready?

erik-wramner:
@tomav if it is in the main repository so that apt-get can get it I would call it production ready. We have not enabled testing or experimental. Still, it could break things, some people only enable security. That is an option for us as well I guess; do the updates for security and wait for the docker image to catch up for the rest.

casperklein:
fyi: the `libgnutls30` update is part of security and fixes [CVE-2020-11501](https://nvd.nist.gov/vuln/detail/CVE-2020-11501).

tomav:
Yeah, I think upgrade is fine.

PR #1458, opened 2020-04-18
Cleanup obsolete file removal

casperklein:
`/etc/cron.weekly/fstrim` does not exist, so no need to remove it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -111,8 +111,8 @@ RUN \
   rm -rf /usr/share/doc/* && \
   touch /var/log/auth.log && \
   update-locale && \
-  rm -f /etc/postsrsd.secret && \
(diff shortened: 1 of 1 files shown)


casperklein:
There is no need for the `-f` switch on these instructions:

```
rm -f /etc/postsrsd.secret && \
rm -f /etc/cron.daily/00logwatch
```

If we remove `-f`, the build will fail, if these files vanish in future builds. Should I create a PR for that too? What are your opinions?

erik-wramner:
You can do that. Removing unused code is good. The test failed with a nonsense error and for some reason I can no longer resubmit failed tests, so you need to push something to run it again. Can you add the suggested changes to this PR please? Then hopefully the tests will pass and I can merge.

PR #1467, opened 2020-04-22
rsyslog logrotate warning fixed

casperklein:
Fix for https://github.com/tomav/docker-mailserver/issues/1465


erik-wramner:
Excellent, hope your fix is more complete than mine!

PR #1473, opened 2020-04-27
chore: Clarify exposed ports section on README

polarathene:
Documentation improvement based on prior stale PR and discussions at https://github.com/tomav/docker-mailserver/pull/1124

- Avoids historical ambiguity of SSL/TLS, if users aren't famliar with these ports, they primarily want to know their purpose and which ones are secure.
- Establishes that STARTTLS *may* be insecure, and how it's configured for `docker-mailserver` on a best effort to be secure and compliant.
- Clarifies that port 25 is not only for receiving external/inbound email, while raising awareness of a common issue for sending outgoing email, with encouragement for submission ports for MUA/MSA to submit email for delivery my the MTA.
- Clearly communicates port 465 as a submission port for implicit TLS, as there is much outdated/conflicting information about it's history.

Presently the wiki link is to the main wiki page. Contributions to improve the wiki in future can later update that link to a more appropriate section for further guidance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,15 +125,19 @@ Your config folder will be mounted in `/tmp/docker-mailserver/`. To understand h
 `restart: always` ensures that the mail server container (and Filebeat/ELK containers when using the mail server together with ELK stack) is automatically restarted by Docker in cases like a Docker service or host restart or container exit.
 
 #### Exposed ports
-* 25 receiving email from other mailservers
-* 465 SSL Client email submission
-* 587 TLS Client email submission
-* 143 StartTLS IMAP client
-* 993 TLS/SSL IMAP client
-* 110 POP3 client
-* 995 TLS/SSL POP3 client
 
-`Note: Port 25 is only for receiving email from other mailservers and not for submitting email. You need to use port 465 or 587 for this.`
+| Protocol | Opt-in Encryption<sup>1</sup> | Enforced Encryption | Purpose              |
+|----------|-------------------------------|---------------------|----------------------|
+| SMTP     | 25                            | N/A                 | Transfer<sup>2</sup> |
+| ESMTP    | 587                           | 465<sup>3</sup>     | Submission           |
+| POP3     | 110                           | 995                 | Retrieval            |
+| IMAP4    | 143                           | 993                 | Retrieval            |
+
+1. A connection *may* be secured over TLS when both ends support `STARTTLS`. On ports 110, 143 and 587, `docker-mailserver` will reject a connection that cannot be secured. Port 25 is [required](https://serverfault.com/questions/623692/is-it-still-wrong-to-require-starttls-on-incoming-smtp-messages) to support insecure connections.
+2. Receives email and filters for spam and viruses. For submitting outgoing mail you should prefer the submission ports(465, 587), which require authentication. Unless a relay host is configured, outgoing email will leave the server via port 25(thus outbound traffic must not be blocked by your provider or firewall).
+3. A submission port since 2018, [RFC 8314](https://tools.ietf.org/html/rfc8314). Originally a secure variant of port 25.
+
+See the [wiki](https://github.com/tomav/docker-mailserver/wiki) for further details and best practice advice, especially regarding security concerns.
 
 ##### Examples with just the relevant environmental variables:
 


polarathene:
Preview of the section updated:

---

#### Exposed ports

| Protocol | Opt-in Encryption<sup>1</sup> | Enforced Encryption | Purpose              |
* 587 TLS Client email submission	|----------|-------------------------------|---------------------|----------------------|
* 143 StartTLS IMAP client	| SMTP     | 25                            | N/A                 | Transfer<sup>2</sup> |
* 993 TLS/SSL IMAP client	| ESMTP    | 587                           | 465<sup>3</sup>     | Submission           |
* 110 POP3 client	| POP3     | 110                           | 995                 | Retrieval            |
* 995 TLS/SSL POP3 client	| IMAP4    | 143                           | 993                 | Retrieval            |


`Note: Port 25 is only for receiving email from other mailservers and not for submitting email. You need to use port 465 or 587 for this.`	1. A connection *may* be secured over TLS when both ends support `STARTTLS`. On ports 110, 143 and 587, `docker-mailserver` will reject a connection that cannot be secured. Port 25 is [required](https://serverfault.com/questions/623692/is-it-still-wrong-to-require-starttls-on-incoming-smtp-messages) to support insecure connections.
2. Receives email and filters for spam and viruses. For submitting outgoing mail you should prefer the submission ports(465, 587), which require authentication. Unless a relay host is configured, outgoing email will leave the server via port 25(thus outbound traffic must not be blocked by your provider or firewall).
3. A submission port since 2018, [RFC 8314](https://tools.ietf.org/html/rfc8314). Originally a secure variant of port 25.

See the [wiki](https://github.com/tomav/docker-mailserver/wiki) for further details and best practice advice, especially regarding security concerns.

youtous:
Nice!
Perhaps we could use a [spoiler] for details.

```
<details>
  <summary>Click to expand</summary>
  whatever
</details>
```
----
### Result
| Protocol | Opt-in Encryption<sup>1</sup> | Enforced Encryption | Purpose              |
|----------|-------------------------------|---------------------|----------------------|
| SMTP     | 25                            | N/A                 | Transfer<sup>2</sup> |
| ESMTP    | 587                           | 465<sup>3</sup>     | Submission           |
| POP3     | 110                           | 995                 | Retrieval            |
| IMAP4    | 143                           | 993                 | Retrieval            |

<details>
  <summary>Click for details</summary>

1. A connection *may* be secured over TLS when both ends support `STARTTLS`. On ports 110, 143 and 587, `docker-mailserver` will reject a connection that cannot be secured. Port 25 is [required](https://serverfault.com/questions/623692/is-it-still-wrong-to-require-starttls-on-incoming-smtp-messages) to support insecure connections.
2. Receives email and filters for spam and viruses. For submitting outgoing mail you should prefer the submission ports(465, 587), which require authentication. Unless a relay host is configured, outgoing email will leave the server via port 25(thus outbound traffic must not be blocked by your provider or firewall).
3. A submission port since 2018, [RFC 8314](https://tools.ietf.org/html/rfc8314). Originally a secure variant of port 25.

</details>

See the [wiki](https://github.com/tomav/docker-mailserver/wiki) for further details and best practice advice, especially regarding security concerns.

README.md
```diff
@@ -125,15 +125,19 @@ Your config folder will be mounted in `/tmp/docker-mailserver/`. To understand h
...
-* 995 TLS/SSL POP3 client
-
-`Note: Port 25 is only for receiving email from other mailservers and not for submitting email. You need to use port 465 or 587 for this.`
+
+| Protocol | Opt-in Encryption<sup>1</sup> | Enforced Encryption | Purpose              |
+|----------|-------------------------------|---------------------|----------------------|
+| SMTP     | 25                            | N/A                 | Transfer<sup>2</sup> |
+| ESMTP    | 587                           | 465<sup>3</sup>     | Submission           |
+| POP3     | 110                           | 995                 | Retrieval            |
+| IMAP4    | 143                           | 993                 | Retrieval            |
```
Yajo, line 134: I suggest to clarify this too:

```suggestion
| POP3     | 110                           | 995                 | Retrieval (unsynced)           |
| IMAP4    | 143                           | 993                 | Retrieval (synced)           |
```

README.md
```diff
@@ -125,15 +125,19 @@ Your config folder will be mounted in `/tmp/docker-mailserver/`. To understand h
...
+
+1. A connection *may* be secured over TLS when both ends support `STARTTLS`. On ports 110, 143 and 587, `docker-mailserver` will reject a connection that cannot be secured. Port 25 is [required](https://serverfault.com/questions/623692/is-it-still-wrong-to-require-starttls-on-incoming-smtp-messages) to support insecure connections.
+2. Receives email and filters for spam and viruses. For submitting outgoing mail you should prefer the submission ports(465, 587), which require authentication. Unless a relay host is configured, outgoing email will leave the server via port 25(thus outbound traffic must not be blocked by your provider or firewall).
```
Yajo, line 137:
```suggestion
2. Receives email (usually from other SMTP servers) and filters for spam and viruses. For submitting outgoing mail you should prefer the submission ports (465, 587), which require authentication. Unless a relay host is configured, outgoing email will leave the server via port 25 (thus outbound traffic must not be blocked by your provider or firewall).
```

polarathene:
> Perhaps we could use a [spoiler] for the details.

Is that supported on DockerHub? Personally, those notes should be visible as they're a common source of confusion, placing them next to the ports info in the docs increases their visibility and relevance.

I've often seen misleading information about `STARTTLS` implying it's secure, and outdated information about port 465. Port 25 stuff is pretty common too. Best to clear that up here, and communicate what to expect with the image. More details will be covered in the wiki.

README.md
polarathene: > (usually from other SMTP servers)

Is that not covered with the following sentence about submitting outgoing mail? 

Port 25 serves the purpose of an MTA (transfer), but can also handle submissions without authentication, such as from your own containers or as an open relay. The user is encouraged to use the submission ports (MSA), so they should only need to think of port 25 for receiving email from a third-party to an inbox, not to be sent out of the server(submission).

youtous:
> Is that supported on DockerHub? Personally, those notes should be visible as they're a common source of confusion, placing them next to the ports info in the docs increases their visibility and relevance.

Dockerhub supports spoiler syntax ([tested here](https://hub.docker.com/repository/docker/youtous/fork-wip-docker-mailserver)).

> I've often seen misleading information about STARTTLS implying it's secure, and outdated information about port 465. Port 25 stuff is pretty common too. Best to clear that up here, and communicate what to expect with the image. More details will be covered in the wiki.

In order to keep README the more concise as much as possible we could use the [home page](https://github.com/tomav/docker-mailserver/wiki) as _Getting started_ and list important points there.

README.md
polarathene: I'm not sure if that's suitable here? If the user doesn't know the difference between POP and IMAP protocols, they're either going to support both or need to read up about them. "Retrieval" keeps it simple for the purpose served.

Wiki section should touch on setup that if using the image to retrieve from an inbox is desirable, that IMAP is most common these days and preferred. 

I don't know if we need to document the behaviour of these too much, other than the ports and what is considered secure. I trust that if the user is deploying this, they're capable of reading up about the basics of POP/IMAP, or already aware of it. 

Some users, will not want to support retrieval and just use the image for sending or relaying transactional email (albeit this image is a bit overkill for such, it has a high popularity / contributions and generally trusted more to be properly configured / secure and documented). These users just need to be aware of what ports/protocols matter for them, they're possibly not familiar with MTA/MSA but would know that retrieval isn't important.

polarathene:
> Dockerhub supports spoiler syntax (tested here).

Requires login to access that link. 

It's good that it's supported, but I still think that information is important not to require the user to click, otherwise I'd just place it in the wiki. There is plenty of other information on the README that could be relocated as it's value is far lower to the majority of users.

> In order to keep README the more concise as much as possible we could use the home page as Getting started and list important points there.

I'll be contributing more details on the wiki, as this PR indicates. This PR is just to improve the section on exposed ports, it's an important part to cover for this image. I've kept the 3 points terse (see earlier revision in related PR that was linked).

If it's not clear, those numbered points match the numbers in the table that they reference.

README.md
Yajo: OK, just was a suggestion.

README.md
Yajo: Yes, although I cannot find that sentence anywhere... where's that from?

The note you deleted seemed more straightforward: *Note: Port 25 is only for receiving email from other mailservers and not for submitting email. You need to use port 465 or 587 for this.* It clarifies the difference among transfer and submission.

polarathene: > It clarifies the difference among transfer and submission.

Port 25 is for an MTA, an MTA can handle both transfer and submission, however port 25 is unauthenticated, thus exposing it to the public internet for submission(not possible by default this image is configured to restrict submission to local IP iirc?) would allow spammers to use the server as an open relay. You could use it to submit email from your containers without them having to provide authentication to do so.

To reduce risk, it's simpler to direct users ignore port 25 for submission, and just deal with the MSA on port 587 / 465, these require authentication upfront making them safer. Before the MSA was a thing though, it's what the MTA handled as well, as such it is capable of that.

Insisting that port 25 is only for receiving and incapable of submission can be a bit confusing if you've learned about that it can prior be used for such. It would be fine if it was more clear on the fact it was best practice and for security reasons to only use port 25 for receiving email and the submission ports for submitting. 

Not only that, I recall reading here that if you do submit to port 25, it would still go through the spam filters and other configuration that differentiates it from submission ports in this image. I haven't looked into that or verified, but that would make the port less attractive to submit to if your own email being sent out gets blocked from doing so..

> Yes, although I cannot find that sentence anywhere... where's that from?

What sentence? I clearly state after the receiving email sentence, that preference should be given to submission ports when you want to send email yourself(either via some email client  or your own containers).

The other half of `2` isn't covered by the previous note that I've replaced. My note only indicates port 25 as capable of receiving email.

erik-wramner:
Good work, thanks!

PR #1474, opened 2020-04-27
chore: Remove obsolete postfix parameter `smtpd_use_tls`

polarathene:
`smtpd_tls_security_level` has replaced this config parameter. `smtpd_use_tls` has no relevance in the config anymore.

See: http://www.postfix.org/postconf.5.html#smtpd_tls_security_level

> this overrides the obsolete parameters `smtpd_use_tls` and `smtpd_enforce_tls`.

### DIFF
--- a/target/postfix/main.cf
+++ b/target/postfix/main.cf
@@ -26,7 +26,10 @@ smtpd_tls_security_level = may
 smtpd_tls_loglevel = 1
 smtp_tls_security_level = may
 smtp_tls_loglevel = 1
+
+# Reduces CPU overhead with `NO_COMPRESSION`, SMTP not at risk of CRIME attack (see git blame for details)
 tls_ssl_options = NO_COMPRESSION
+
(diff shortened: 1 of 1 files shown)


erik-wramner:
Looks good but the tests fail and I can't resubmit them. Can you make a nonsense commit (some whitespace will do) to trigger a new build, please? I'll merge as soon as the tests pass.

polarathene:
Done, tests pass.

Added a note to the config about disabled compression. I was aware of CRIME, but wasn't sure how it applied to SMTP. 

May be useful for anyone else in future, mentioned to use `git blame` to view the lines commit description for reviewers in future to reference.

target/postfix/main.cf
```diff
@@ -26,7 +26,10 @@ smtpd_tls_security_level = may
 smtpd_tls_loglevel = 1
 smtp_tls_security_level = may
 smtp_tls_loglevel = 1
+
+# Reduces CPU overhead with `NO_COMPRESSION`, SMTP not at risk of CRIME attack (see git blame for details)
 tls_ssl_options = NO_COMPRESSION
```
youtous, line 31: We could add `NO_RENEGOTIATION` in order to mitigate CPU exhaustion attack. http://www.postfix.org/postconf.5.html#tls_ssl_options

>    Postfix ≥ 3.4. This can reduce opportunities for a potential CPU exhaustion attack.

erik-wramner: Yes that is a good idea. I'll merge this though, so let's make a separate PR for that.

PR #1479, opened 2020-04-30
fix 1478

youtous:
When using LDAP, dovecot quota service was still enabled in postfix.
https://github.com/tomav/docker-mailserver/issues/1478


target/start-mailserver.sh
```diff
@@ -637,6 +637,9 @@ function _setup_dovecot_quota() {
     if [ "$ENABLE_LDAP" = 1 ]; then
       notify 'inf' "Dovecot quota is not implemented with LDAP."
 
+      # LDAP => postfix service for dovecot quota not enabled
```
erik-wramner, line 640: It looks like you closed the if statement wrong? Before the SMTP_ONLY check was outside the ENABLE_LDAP check, now it is inside?

youtous: The last commit remove the nested condition, it's cleaner.
The function should only disable or enable quota.

Quota is disabled when LDAP is enabled or when SMTP_ONLY is enabled, otherwise dovecot quota is enabled.

Do you agree this logic?

target/start-mailserver.sh
```diff
@@ -663,11 +664,8 @@ function _setup_dovecot_quota() {
 		    echo -n >/tmp/docker-mailserver/dovecot-quotas.cf
```
erik-wramner, line 664: This can overwrite configured quotas, right? Is it perhaps better to use touch?

erik-wramner:
As this fixes a bug I'll merge now, but please check comment above.

target/start-mailserver.sh
casperklein: btw: `echo -n` is not needed. `>/tmp/docker-mailserver/dovecot-quotas.cf` is enough.
Should we replace all `echo -n >`?

PR #1482, opened 2020-05-01
Feature quota optional

youtous:
Allow user to disable dovecot quota.


erik-wramner:
Thanks for the quick fix.

PR #1484, opened 2020-05-02
Updating SPAMASSASSIN_SPAM_TO_INBOX to support default amavis override

VanVan:
Updating SPAMASSASSIN_SPAM_TO_INBOX to support the default overload in the amavis/conf.d/49-docker-mailserver file.
If the user decides to override this value, then an error message will be displayed during initialization to tell the user to make a choice.

Issue #1396, opened 2020-02-12
incoming spam is sometimes bounced, please add parameter


hanscees:
## Context

Working on an email-host with local mailboxes for 4 domains facing the internet.

When testing if spammassasin works I sended some test emails coming form the internet. 
Since I had defined a spam-header (####spam####) I was expecting that spam would be passed to endusers and tagged.

However, some spam was bounced to the internet.

logging:

```

Feb 12 20:55:30 mail amavis[1043]: (01043-01) Blocked SPAM {BouncedInternal,Quarantined}, LOCAL [172.19.3.101]:57278 <kees@vlees.nl> -> <hanscees@hansbees.com>, quarantine: L/spam-LMHsfmxldtMS.gz, Queue-ID: 35EE580A20, mail_id: LMHsfmxldtMS, Hits: 1005.482, size: 544, 999 ms
Feb 12 20:55:30 mail postfix/smtp[13246]: 35EE580A20: to=<hanscees@hansbees.com>, relay=127.0.0.1[127.0.0.1]:10024, delay=28, delays=27/0/0.01/1, dsn=2.5.0, status=sent (250 2.5.0 Ok, id=01043-01, BOUNCE)
Feb 12 20:55:30 mail postfix/qmgr[1975]: 35EE580A20: removed

Feb 12 20:55:30 mail postfix/smtp[13254]: Trusted TLS connection established to vlees-nl.mail.protection.outlook.com[104.47.2.36]:25: TLSv1.2 with cipher ECDHE-RSA-AES256-GCM-SHA384 (256/256 bits)
Feb 12 20:55:31 mail postfix/smtp[13254]: 88DC380B30: to=<kees@vlees.nl>, relay=vlees-nl.mail.protection.outlook.com[104.47.2.36]:25, delay=0.84, delays=0/0.01/0.18/0.65, dsn=2.6.0, status=sent (250 2.6.0 <SSLMHsfmxldtMS@mail.hansbees.net> [InternalId=11055245821936, Hostname=AM6PR06MB5845.eurprd06.prod.outlook.com] 10723 bytes in 0.137, 75.892 KB/sec Queued mail for delivery)


```

Users in my setup want to get all email (except from rbl-tagged hosts etc), since non of it can be missed. So rejecting mail on the basis of mail content is not desirable.

Secondly sending bounces to probably forged recipient email addresses is not clever.

## Expected Behavior

spam should be delivered to users and tagged.

## Actual Behavior

[...]

PR #1485, opened 2020-05-02
Feature: Spam to Junk folder

youtous:
This PR is related to https://github.com/tomav/docker-mailserver/issues/1396

Features:
- Fix `SPAMASSASSIN_SPAM_TO_INBOX` behavior
- `MOVE_SPAM_TO_JUNK=1` permits to move spams to Junk folder using a global sieve rule
- Warning raised when `SPAMASSASSIN_SPAM_TO_INBOX=0`
- Tests

`MOVE_SPAM_TO_JUNK=1` will allow [SpamAssassin Lear](https://github.com/tomav/docker-mailserver/wiki/FAQ-and-Tips#how-can-i-make-spamassassin-learn-spam) to work. 

~TODO:~
- ~README update: strong warning about the default behavior of spam bouncing~

fix https://github.com/tomav/docker-mailserver/issues/1396

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -125,6 +125,7 @@ RUN echo "0 */6 * * * clamav /usr/bin/freshclam --quiet" > /etc/cron.d/clamav-fr
 # Configures Dovecot
 COPY target/dovecot/auth-passwdfile.inc target/dovecot/??-*.conf /etc/dovecot/conf.d/
 COPY target/dovecot/scripts/quota-warning.sh /usr/local/bin/quota-warning.sh
+COPY target/dovecot/sieve/* /usr/lib/dovecot/sieve-global/
 WORKDIR /usr/share/dovecot
 # hadolint ignore=SC2016,SC2086
 RUN sed -i -e 's/include_try \/usr\/share\/dovecot\/protocols\.d/include_try \/etc\/dovecot\/protocols\.d/g' /etc/dovecot/dovecot.conf && \
--- a/README.md
+++ b/README.md
@@ -161,6 +161,7 @@ services:
       - ./config/:/tmp/docker-mailserver/
     environment:
       - ENABLE_SPAMASSASSIN=1
+      - SPAMASSASSIN_SPAM_TO_INBOX=1
       - ENABLE_CLAMAV=1
       - ENABLE_FAIL2BAN=1
       - ENABLE_POSTGREY=1
@@ -202,6 +203,7 @@ services:
       - ./config/:/tmp/docker-mailserver/
     environment:
       - ENABLE_SPAMASSASSIN=1
+      - SPAMASSASSIN_SPAM_TO_INBOX=1
       - ENABLE_CLAMAV=1
       - ENABLE_FAIL2BAN=1
       - ENABLE_POSTGREY=1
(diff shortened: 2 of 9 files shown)


erik-wramner:
See suggestion in #1396. Can we amend it that way? Also I don't want a warning when `SPAMASSASSIN_SPAM_TO_INBOX=0`, it is a perfectly legitimate setting. You can show the warning when `SPAMASSASSIN_SPAM_TO_INBOX` has not been set (which should default to 0 for backwards compatibility) though. Then people who have not made up their minds will see it and make an informed selection.

youtous:
> You can show the warning when SPAMASSASSIN_SPAM_TO_INBOX has not been set (which should default to 0 for backwards compatibility) though. 

I agree with you.

Current default value is 0.
```
DEFAULT_VARS["SPAMASSASSIN_SPAM_TO_INBOX"]="${SPAMASSASSIN_SPAM_TO_INBOX:="0"}"
```

Any idea for getting the original env value without (proper way) default assignment? ` :=0`

VanVan:
If SPAMASSASSIN_SPAM_TO_INBOX=1 is the default configuration, this is right, there is no need for a warning, because the user made the change himself after reading it. If not the default value, you can keep warning message.
However, I think you should have kept the original configuration and PR #1484 in the start-mailserver.sh file, as it allows you to keep the modification of the 20-debian_defaults file in case the 49-docker-mailserver file is not present (Update from previous version by keeping its configuration files, which is my case for example).
This allows to keep the banned_destiny configuration, and to display an error if a configuration in 50-user makes SPAMASSASSIN_SPAM_TO_INBOX unusable, in this case, it is better to disable this value if other functions use this value to modify the operation.

VanVan:
@youtous No modification in this file, only in env-mailserver.dist. Previous users will keep their configurations. #1488

youtous:
> Your forget to also create Junk directory if not exist on file target/dovecot/sieve/before.spam.sieve

As the tests show, no need to create a directory.

> modification of the 20-debian_defaults

`20-debian_defaults` should be left untouched.

> in case the 49-docker-mailserver file is not present
 
User updates must be done using `50-user` (https://github.com/tomav/docker-mailserver/blob/master/target/amavis/conf.d/49-docker-mailserver#L3), there is no reason for `49-docker-mailserver` to be missing in a normal usage (correct me if I'm wrong).

> display an error if a configuration in 50-user makes SPAMASSASSIN_SPAM_TO_INBOX unusable

Isn't `50-user` made for overwriting changes made in previous configurations?

> If SPAMASSASSIN_SPAM_TO_INBOX=1 is the default configuration, this is right, there is no need for 

`SPAMASSASSIN_SPAM_TO_INBOX` is set to `0` by default. A warning will be raised if `SPAMASSASSIN_SPAM_TO_INBOX` is not explicitly defined.

youtous:
> See suggestion in #1396. Can we amend it that way?

Implemented behavior regarding `SPAMASSASSIN_SPAM_TO_INBOX`:
- `SPAMASSASSIN_SPAM_TO_INBOX=0` (default), spam messages are bounced (current behavior)
- `SPAMASSASSIN_SPAM_TO_INBOX=1`, spam messages are delivered and eventually marked as spam (according SA config)
- `SPAMASSASSIN_SPAM_TO_INBOX not set`: raise a warning asking user to explicitly confirm the desired behavior.

Are you ok with this @erik-wramner?

erik-wramner:
This looks fine. Unfortunately I merged the other PR first and I'm not allowed to resolve the merge conflict. Can you rebase and just keep your changes, please? Then I'll merge.

youtous:
Rebase completed, tests are running.
Closes https://github.com/tomav/docker-mailserver/pull/1484.

PR #1491, opened 2020-05-03
Fix sieve scripts not loaded

youtous:
Currently, sieve scripts are loaded using  an individual `sieve_before/after` key associated with a file in `90-sieve.conf`.

For instance:
```
sieve_before = /usr/lib/dovecot/sieve-global/before.dovecot.sieve
sieve_before2 = /usr/lib/dovecot/sieve-global/before.spam.sieve
```

This way of loading sieve scripts requires first script to be present for loading the second script.
In the following example, ` /usr/lib/dovecot/sieve-global/before.spam.sieve` will not be loaded because `sieve_before` was commented:
```
#sieve_before = /usr/lib/dovecot/sieve-global/before.dovecot.sieve <--- this script is disabled
sieve_before2 = /usr/lib/dovecot/sieve-global/before.spam.sieve
```
I've noticed this behavior when using `MOVE_SPAM_TO_JUNK` without using a `before.spam.sieve` script.

#### Solution

Loading sieve scripts using a directory scheme for `sieve_before/sieve_after`, any script present in the defined directory will be loaded (_directory used is unchanged_).

This fix does not break existing installations as sieve filters are copied and compiled each time `start-mailserver.sh` is executed.


erik-wramner:
Looks good. One question though. I'm not that familiar with sieve, does the order matter? Changing to a directory means there is less control of the load order for the scripts, right? Probably they are loaded in alphabetic order (or random order)?

If the order is alphabetic we should adopt a convention where the files we use start with two digits, i.e. `20-before-dovecot.sieve`. If the order doesn't matter then forget it. What do you think?

youtous:
This make sense. According`90-sieve.conf`:
> The order of execution within a directory is determined by the file names, using a normal 8bit per-character comparison

I've added two digits for those scripts:
- `before/00-before.dovecot.sieve`
- `after/00-after.dovecot.sieve`
- `before/15-spam.sieve`

What do you think?

erik-wramner:
One thing I learned when I wrote BASIC programs with line numbers in the early 80s was to always leave free numbers before and after every line, as manually renumbering a big program could be a royal pain. Later they added the `renum` command, what an improvement...

If we use 00-something then our before script will run first and it will be hard to get in front. Likewise the after script will also run first in that group and that may not be the best thing.

I would suggest using 50 for both. My example used 20 as that happens to be the debian default file for dovecot, but there is nothing magic about it. As long as there are free numbers before and after anything works.

How about changing to 50-before.dovecot.sieve, 60-spam.sieve and 50-after.dovecot.sieve?

youtous:
It's a good practice indeed.
Updated with suggested numbers ;)

PR #1503, opened 2020-05-09
update ports recommendations according RFC8314

youtous:
Fix https://github.com/tomav/docker-mailserver/issues/1501.

This update best practices concerning ports usage ~and highlights possible MITM when port 587 is available~. 

~This update permits not to change `smtpd_tls_security_level` to `encrypt`, users using port `587` will be informed related security consequences on TLS negocation.~

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,16 +125,17 @@ Your config folder will be mounted in `/tmp/docker-mailserver/`. To understand h
 
 #### Exposed ports
 
-| Protocol | Opt-in Encryption<sup>1</sup> | Enforced Encryption | Purpose              |
-|----------|-------------------------------|---------------------|----------------------|
-| SMTP     | 25                            | N/A                 | Transfer<sup>2</sup> |
-| ESMTP    | 587                           | 465<sup>3</sup>     | Submission           |
-| POP3     | 110                           | 995                 | Retrieval            |
-| IMAP4    | 143                           | 993                 | Retrieval            |
+| Protocol | Opt-in Encryption<sup>1</sup>   | Enforced Encryption | Purpose              |
+|----------|---------------------------------|---------------------|----------------------|
+| SMTP     | 25                              | N/A                 | Transfer<sup>2</sup> |
+| ESMTP    | 587 _(deprecated<sup>4</sup>)_  | 465<sup>3</sup>     | Submission           |
+| POP3     | 110 _(deprecated<sup>4</sup>)_  | 995                 | Retrieval            |
+| IMAP4    | 143 _(deprecated<sup>4</sup>)_  | 993                 | Retrieval            |
 
-1. A connection *may* be secured over TLS when both ends support `STARTTLS`. On ports 110, 143 and 587, `docker-mailserver` will reject a connection that cannot be secured. Port 25 is [required](https://serverfault.com/questions/623692/is-it-still-wrong-to-require-starttls-on-incoming-smtp-messages) to support insecure connections.
+1. A connection *may* be secured over TLS when both ends support `STARTTLS`. On ports 110, 143 and 587, `docker-mailserver` will reject a connection that cannot be secured _(preventing [MITM attacks](https://stackoverflow.com/questions/15796530/what-is-the-difference-between-ports-465-and-587/32460763#32460763) trough a downgrading)_. Port 25 is [required](https://serverfault.com/questions/623692/is-it-still-wrong-to-require-starttls-on-incoming-smtp-messages) to support insecure connections.
 2. Receives email and filters for spam and viruses. For submitting outgoing mail you should prefer the submission ports(465, 587), which require authentication. Unless a relay host is configured, outgoing email will leave the server via port 25(thus outbound traffic must not be blocked by your provider or firewall).
 3. A submission port since 2018, [RFC 8314](https://tools.ietf.org/html/rfc8314). Originally a secure variant of port 25.
+4. [RFC 8314](https://tools.ietf.org/html/rfc8314) is recommending that clear text exchanges to be abandoned and that all three common IETF mail protocols to be used only in implicit mode.
 
 See the [wiki](https://github.com/tomav/docker-mailserver/wiki) for further details and best practice advice, especially regarding security concerns.
 


polarathene:
Isn't this already covered in the first point for that column? Would it not be better to instead further document in the wiki instead?

> 1. A connection may be secured over TLS when both ends support STARTTLS. On ports 110, 143 and 587, `docker-mailserver` will reject a connection that cannot be secured. Port 25 is required to support insecure connections.

AFAIK, credentials won't be handled until a secure connection can be established, but a client/MUA may not respect that and send them forward anyway.

---

> This update permits not to change `smtpd_tls_security_level` to `encrypt`, users using port `587` will be informed related security consequences on TLS negocation.

What is this about? I don't see any such mention in the commit for this PR?

youtous:
> Isn't this already covered in the first point for that column? Would it not be better to instead further document in the wiki instead?

Yes, it is. Paragraph 1 explains what is the behavior of this image.

Paragraph 4 justifies best practices concerning email port usage. I've edited the last sentence of §4 and merged the comment about MITM _- not relevant, could be removed -_ in §1.

I don't think §4 should be moved to the wiki, it's an important recommendation, not attentive users could miss it; `docker-compose` and README examples are using `143` and  `587`, marking these ports as _obsolete_ is a clear signal for not using them.

polarathene:
The URL is referring to the entire discussion thread or a specific answer like [this one](https://stackoverflow.com/questions/15796530/what-is-the-difference-between-ports-465-and-587/32460763#32460763)?

Linking directly that one sort of conveys the 4th point you've added, and I still feel that it's better suited for the wiki. Port 587 is not really obsolete, it's still valid to use, especially with this image, all 3 ports in postfix and dovecot are expected to successfully establish secure connections before any submission/retrieval occurs.

If the user is aware of the benefit of encryption/TLS, they'd likely be aware that it's meant to prevent such attacks. What is not initially clear to a user is that there is two types with email, explicit/opportunistic and implicit/enforced TLS. Further complicated by port 25 allowing for insecure connections while still seeming to use the same protocol/functionality as 587(to a new user before they grok it at least).

The current table addresses that fairly well. I did originally have something similar but it was agreed to move that into the wiki. Encouraging implicit TLS ports vs explicit may be valid to add to the README, but otherwise I'm not sure if it's worth changing too much.

Perhaps @erik-wramner or others can weigh in with their opinions. I think the link is mostly good, just not sure if it belongs there.

polarathene:
> I don't think §4 should be moved to the wiki, it's an important recommendation, not attentive users could miss it; docker-compose and README examples are using 143 and 587, marking these ports as obsolete is a clear signal for not using them.

The more content presented to the reader, the less non-attentive users will bother to read. Thus it's important to keep terse and expand via wiki.

Having a clear column of opt-in encryption vs enforced makes the distinction between ports fairly clear security wise in which to prefer. I believe the implicit TLS ports for receiving mail are commonly used, but 587 for submission is still common place (at least for relays, where some also offer 465), from a MUA perspective, 465 is what should be preferred. This image isn't just for MUA users though afaik.

youtous:
> The URL is referring to the entire discussion thread or a specific answer like this one?

I was referring to the linked answer which is the only up to date in the thread.

> Port 587 is not really obsolete
According the RFC8314, 465 should be prefered over 587, when the transition will end, 587 will be obsolete. A more appropriate tag could be _deprecated_.

Something like: `143`, `110` and `587` are deprecated by RFC8314

> Encouraging implicit TLS ports vs explicit may be valid to add to the README, but otherwise I'm not sure if it's worth changing too much.

I agree with this, it will makes sense if we could have a **Best security practices** section in the wiki, as you [recommended](https://github.com/tomav/docker-mailserver/issues/1501#issuecomment-626229126).

> The more content presented to the reader, the less non-attentive users will bother to read. Thus it's important to keep terse and expand via wiki.

:sweat_smile: I couldn't have said it better.

> I believe the implicit TLS ports for receiving mail are commonly used, but 587 for submission is still common place (at least for relays, where some also offer 465)

I don't know much about the state of the art of relays, but I believe relays will be RFC8314 compliant soon or later.

erik-wramner:
Good discussion.

I agree in principle, but I don't think the STARTTLS ports are deprecated from our point of view. On the contrary the RFC you link to says mail servers SHOULD implement both STARTTLS on port 587 and Implicit TLS on port 465, so the mail server should provide both, but encourage the use of the latter. Deprecated in this context tells me that `docker-mailserver` is about to stop supporting the feature and that is not true here.

In addition too many footnotes make it hard to grasp for newcomers.

How about simply changing the header "Enforced Encryption" to "Enforced Encryption (preferred)" where preferred is a link to https://tools.ietf.org/html/rfc8314 and remove the rest of the text or move it to the existing wiki page? Short and to the point.

polarathene:
Could just add to note 1, to prefer implicit/enforced TLS when possible, or add ^2 to the column and assign it's own note, bumping the existing 2/3 notes to 3/4. Seems better to just append to 1 though.

chikamichi:
I've created https://github.com/tomav/docker-mailserver/wiki/A-mail-server's-101 as an attempt to provide a beginner-friendly summary on (but not limited to) the topic of security and ports.

Hopefully it achieves that, in which case it could be linked from the README and replace all the footnotes.

polarathene:
@chikamichi nice contribution to the wiki, thanks for that!

I noticed you're referring to port 465 as SMTPS, but this is not what it is and the RFC 8314 does not claim it to be that, just that was it's historical use(lowercase "smtps" as a port service, rather than protocol) - I assume your misunderstanding was from section 7.3?

It is [ESMTP](https://en.wikipedia.org/wiki/Extended_SMTP) as the RFC points out in section 4.3, it is known as a Message Submissions port, just like 587 - however similar to the older "smtps" designation, where port 587 is designated as the "submission" port, port 465 is the "submissions" port(with s appended, as explained in section 3.3). These aren't differences the user should really need to care about however, just referring to them both as submissions ports with the difference of explicit and implicit TLS is sufficient.

SMTPS would be an implicit TLS variant of port 25, no authentication required. ESMTPA is exchanged in ESMTP protocol to indicate the authentication extension (there is also ESMTPSA, where the extra S indicates successful STARTTLS negotiation). It's better just to refer to it as ESMTP or SMTP with explicit or implicit TLS(RFC 8314, section 3 cites it will be referred to as such, at least within the document), where ports 587 and 465 should according to RFCs mandate authentication. 465 is only for submission now(MUA-to-MSA), not transfer/relay(MTA-to-MTA).

> an alternate mechanism where TLS is negotiated immediately at connection start on a separate port (referred to in this document as "Implicit TLS") has been deployed more successfully.  To encourage more widespread use of TLS and to also encourage greater consistency regarding how TLS is used, this specification now **recommends the use of Implicit TLS** for POP, IMAP, **SMTP Submission**, and all other protocols used between an MUA and an MSP. - [RFC 8314, Section 3](https://tools.ietf.org/html/rfc8314#section-3)

As mentioned in [section 3.3](https://tools.ietf.org/html/rfc8314#section-3.3), SMTP-AUTH is a requirement, credentials must be submitted, it is not like HTTPS where only a secure connection is established, as that would defeat the purpose at preventing spam submissions. Therefore SMTPS is a bit misleading, you would achieve the same with port 587 once a secure connection is established, the message transfers would be encrypted and technically "SMTPS", provided 587 was implemented correctly as RFC 8314 warns, and the main motivation for preferring 465 going forward. 

Note that some providers/clients may still treat 465 as SMTPS(legacy, no auth) or in some cases an alternative 587 iirc(still using STARTTLS), you'll still see articles in 2020 misrepresenting the usage of the port, so confusion is understandable.

Mentions of deprecation could probably link to [RFC 8314 section 4.1](https://tools.ietf.org/html/rfc8314#section-4.1) where it's described as being phased out over several years, and how it should ideally be approached. There may still be clients or services where port 465 is still handled as a legacy port or only support 587.

---

> I incorporated a slightly revised copy of the README's table & associated footnotes.

The updated port table doesn't need to add implicit protocol variants imo. Why was the `Purpose` column moved? It's secondary information, not primary. 

Note 1 modifications seem to add more noise for no apparent reason? The only potentially worthwhile addition is the mention of MITM attack, which has a spelling error (through not trough), and mentions *downgrading* while introducing *upgrading* earlier, it attempts STARTTLS but downgrades to insecure if that could not be established(or in this projects case is rejected if it's not port 25). This concern could be better explained in the wiki elsewhere instead of modifying the table note.

Note 2 modifications have issues as well. 

- Being more explicit about receiving *incoming* mail is fine(as opposed to client retrieval, although that should be apparent that it is different from receiving and *incoming* may not clarify that difference by much?).
- *outgoing* definition here with ports 587 and 465 is invalid. The intention here was for a user sending outgoing mail via their MUA(eg ThunderBird, and possibly any local service on the server wanting to send outgoing mail) to the `docker-mailserver`(MSA) to handle and send out itself. 
- The MSA will either be configured to forward to a relay server such as SendGrid typically with authentication using ports 587 or 465(SendGrid also offers ports 25 and 2525 to receive on, but as [their docs mention](https://sendgrid.com/docs/for-developers/sending-email/getting-started-smtp/), each port expects auth, they're non-standard to workaround constraints), or your MSA is not going through an explicit relay and sending the e-mails directly to the MTA of an e-mail address, this would be over port 25. 
- The wiki could better explain this better without trying to cram it into a table note.

Note 3 modifications are invalid, you cite it being made a submissions port *again* which it was not prior, but it was confused as being one, RFC 8314 mentions this(see section 7.3). You then repeat the same mistake, saying it was a secure variant of port 25, but is now dedicated to SMTPS.. what actual difference does that indicate? It's now a submissions port(requries authentication as an earlier note clarifies), but unlikely port 587, uses implicit TLS, just like the legacy usage did, which lacked requiring authentication. The modification here adds no value, if anything it spreads misinformation.

The new 4th note addition, you could drop the *(deprecated)* addition to the table, and mention it as the first word in the 4th note, while linking to the relevant RFC 8314 section as mentioned earlier in this comment. Or, alternatively indicate the deprecation within the 1st note. Personally, if the user has gone so far as to view this information, communicating preference to the implicit TLS ports along with the earlier notes already raising awareness of security concerns on these ports, this is also adding to noise and may be better expanded on out of tables notes in a relevant wiki section.

---

> I'd propose removing the footnotes and simplify the table, granted the full-fledged version is now in the wiki.

No. The purpose of the table was to communicate important information in a terse manner, and not expect a user to further seek more details unless they legitimately could be bothered. The table serves to communicate the port info and purpose clearly, with the terse footnotes to clear up common misunderstandings / confusion, deferring to the wiki for further clarification and details.

If the footnotes were omitted, the user may visit the wiki and be presented with a much larger document similar to what they may come across with various blog articles on the subject. Like with a educational video, it's not clear where the more relevant information is to seek, and thus the user may opt-out of going through it.

The foot notes point out(for the most part):

1) STARTTLS is better than nothing, but you should really prefer implicit TLS ports.
2) Submission ports should be preferred when you or a service you run wants to send mail through `docker-mailserver`, using port 25 to avoid authentication, even if not exposed to the public internet as an open-relay, may cause unexpected problems such as your service e-mail not being sent out due to the spam/virus filters raising a false positive(at least with this project).
3) Sending mail to another server via port 25 *may* be secure but it's not something you can realistically enforce through transit(nor 587/465 for that matter when the relay sends to another MTA).
4) `docker-mailserver` tries to be secure(to the best of it's ability), by rejecting insecure STARTTLS connections on ports 110, 143, 587. Something worth knowing if you're searching for mail server project to use, but haven't decided on one so you're going over READMEs briefly while researching what is out there and their differences.

The wiki for additional details are less important, or can better expand on these details.

The README should retain it's footnotes with the table, the majority of alterations to the notes however are not relevant there or misinformation in themselves, while also working against keeping it terse.

polarathene:
> with an opt-in TLS upgrade of the client-to-server connection using using STARTTLS.

"using" repeated twice.

> A final setup exists and is akin SMTP+STARTTLS, but over port 25. That port has historically been reserved specifically for plain text mail exchange. One may upgrade the connection on port 25 to a TLS-encrypted one, but that should be considered a non-normative usage. It's better reserving port 25 for plain text trafic in order to support older clients, and inter-MTA exchange (although obviously non-secure).

chikamichi:
Now that's what I call _feedback_! Thank you :+1:

I can't help but thinking: why didn't he edit the wiki directly? :laughing:

Read it once but I'll go through it all over again later on, and will come up with a better version of the content on the wiki.

chikamichi:
I've reworked https://github.com/tomav/docker-mailserver/wiki/A-mail-server's-101.

- It's written with the beginner's concerns in mind. That's why I decided to structure the document as a "story": it starts with simple (thus simplifying) definitions & examples, then highlights definitions for the most important technical terms, shares extensive details about technologies pertaining to emails exchange, and eventually bridges the gap between that and docker-mailserver specifically.
- It first features Submission, Retrieval and Transfer/Relay as "instinctive" guidelines for the beginner to relate to (sending / fetching / relaying emails are simple, visual concepts), instead of focusing on protocols and ports and whatnots. Those are covered eventually though.
- I removed the table borrowed from the README. That's one less problem with this document.

Hopefully I've fixed all misinformation. @polarathene I feel like I'm in the same process right now than you were a few months ago, taming the Mail stack beast :) So please forgive me for any remaining mistakes.

chikamichi:
Erik wrote:

> I agree in principle, but I don't think the STARTTLS ports are deprecated from our point of view. On the contrary the RFC you link to says mail servers SHOULD implement both STARTTLS on port 587 and Implicit TLS on port 465, **so the mail server should provide both, but encourage the use of the latter.**

What about adding to this PR/Draft, by enabling `"465:465"` in the port mapping of docker-compose.yml.dist? There is already `"993:993"` for IMAP4S.

polarathene:
@chikamichi I tried reviewing and making some changes to the wiki page twice now, but while doing so, you continued to submit many changes(which is fine). That is a bit difficult for me to review again and know if any changes I made are still relevant to apply, along with matching the content if it changed a fair bit like with my first change. I will hold off until you feel you're done editing (I assumed the message here was indication of that, but the edits keep coming).

It has changed quite a bit from the original version, so I don't have any feedback until your revisions settle down for me to go through again. When making changes, especially so frequently, it can be nice to be a bit more descriptive with the title vs re-using the same generic "update" title that they all have. Not a big issue, since as long as I remember the commit hash I last reviewed, I can view a diff against the latest, but for anyone that wants to inspect the changes, it's unfortunate noise that requires individual inspection depending on why they're looking over the revision history.

I cannot comment on the docker-compose suggestion to add port 465 right now, I'll respond to that after wiki changes are sorted out.

chikamichi:
Oh, I thought you would be asleep at the other edge of the Earth :p
Let me finish my current edit and you'll be fine.

chikamichi:
Alright @polarathene I'm done with editing for now.

polarathene:
I have done a brief review and made some minor adjustments. Mostly some small spelling/grammar fixes, referring to the project name with `inline code block`, and changing `plain text` to `cleartext`, which is the correct term(I only learned that today since this is a public wiki, using the correct terminology is more important). 

I also added a [new page](https://github.com/tomav/docker-mailserver/wiki/Understanding-the-ports) with the first change I had been working on with the table you removed, and clarifying Explicit vs Implicit TLS, while keeping the mention of the deprecation warning that had been previously added as a footnote. 

Fixed up the sidebar too, as some links were on the same line but better suited to another level of bullet points.

---

This is a confusing statement, it talks about Implicit TLS, then switches to explicit midway but only indicates that at the end of the paragraph. 

> Implicit TLS means the server _enforces_ the client into using an encrypted TCP connection, using [TLS](https://en.wikipedia.org/wiki/Transport_Layer_Security). With this kind of connection, the MUA _has_ to establish a TLS-encrypted connection from the get go. The mail server would deny any client attempting to submit email in cleartext (unencrypted, not secure) or requesting a cleartext connection to be upgraded to a TLS-encrypted one (secure if `STARTTLS` is successful). It is also known as Enforced TLS.

It should probably be rephrased so that it's clear you're now talking about Explicit TLS.

---

I did have a rendering issue when viewing the ASCII flows you made in browser text field or when viewing the rendered markdown page. However I'm on Linux and inspecting CSS it's using a linux mono font, that may render the content differently, so if it looks worse on your end, please revert the change, I made sure to do so via a separate revision. 

I am probably a minority with font issues like that, only way to fix it for everyone is if Github used a webfont, instead of relying on system mono fonts.

---

Other than all that, the wiki page looks like it could be improved still, but I wouldn't worry about it(not worth the time/effort), it's a fantastic contribution and much appreciated. I'm sure it'll be helpful to new users regardless :)

I am a bit surprised that anyone can make changes to the wiki without approval though.

chikamichi:
Thank you very much @polarathene. I especially appreciate your edits on grammar, for English is not my native language.

I tried my best to further improve the content, with https://github.com/tomav/docker-mailserver/wiki/A-mail-server's-101/_compare/fac8d4ae50a7a93d39b9f75da8df3bd6bce4aa4d...986e60965a13ef9331b08be1006123e156496d09.

chikamichi:
Now that https://github.com/tomav/docker-mailserver/wiki/Understanding-the-ports exists and integrates youtous' proposed changes, I guess this Draft/PR could be closed?

polarathene:
Reviewed. Made some small changes/corrections, and updated my separate wiki page on the ports, which seemed to touch on a similar topic you had written about with your new changes(inbound vs outbound).

It didn't occur to me earlier, but the example email hostname you're using `dms.io`, why are you using this in the wiki? Is it meant to be an abbreviation of `docker-mailserver`? Is that used anywhere else? I only see env var `DMS_DEBUG` with a quick search.

---

Do you have any reference of your Inward/outward submission explanation? This did differ from inbound/outbound traffic that my wiki page covers. 

An MTA can handle submission of email to send(or an MSA can specifically handle submissions, but just thinking of the MTA is fine for wiki). Receiving email from another MTA is not email submission. I think your terminology of what you're trying to describe is using the wrong words here.

Port 25 can be used to deliver email to your MTA, or to relay through your MTA(but afaik `docker-mailserver` does not allow by default for this unless it's an internal/local IP address). Submission ports 465 and 587 are only for submission to your MTA, which can then be sent out to another MTA on port 25 or 465/587 if your MTA is not responsible for the destination email address. The inbound and outbound ports can differ as this is two separate tasks/queues. 

My wiki page includes a flow chart graphic that should better visualize that explanation.

Your wiki section probably shouldn't refer it to as "Inward Submission" for incoming deliveries to your MTA. Likewise, "Outward Submission" describes submitting to your own MTA for potential outgoing delivery/relay if the address hostname is not represented by your MTA, but the name suggests you're submitting email externally(while the description states submission is handled by your MTA, followed by relaying to an external third-party MTA. And the ASCII graphic that it is only between MUA and your MTA, not covering leaving the MTA).

---

In your wiki page, there seems to be some incorrect statements(or missing information) about defaults? Such as port 25 on the ports table and statements like this one:

> It does not enable Explicit TLS (STARTTLS) on port 25 by default.

https://github.com/tomav/docker-mailserver/blob/6abcc8ce5fa9352e1b5aedc2340f31dc72e6f657/target/postfix/main.cf#L25

Both `smtp`(Postfix as client, to send email to another MTA) and `smtpd`(Postfix receiving connection as an MTA) are configured by `docker-mailserver` to have their [`tls_security_level`](http://www.postfix.org/postconf.5.html#smtp_tls_security_level) set to `may` as shown above. This will attempt to use `STARTTLS` if possible between the two MTA.

This setting is specific to port 25, 587 is handled in the other `master.cf` config where auth is required and connections unable to establish with `STARTTLS` successfully are rejected. `encrypt` enforces that:

https://github.com/tomav/docker-mailserver/blob/6abcc8ce5fa9352e1b5aedc2340f31dc72e6f657/target/postfix/master.cf#L19

To re-iterate, port 25 supports encryption through `STARTTLS`, just like port 587, but it is not mandatory due to some RFC requirement, and that rejection would prevent receiving email that you might actually want. That's a non-issue for submitting email to your MTA though(where you are in control of both sides of the connection), so your submission ports can enforce it in addition to requiring authentication. Port 25 allows for untrusted/anonymous sources, while the submission ports allow for trusted sources to interact with the MTA.

polarathene:
> I guess this Draft/PR could be closed?

Yes, I think so.

@youtous could you close your PR? Thanks.

polarathene:
Oops, clearing through my tabs and noticed I haven't sent this response to you @chikamichi 

> There is already `"993:993"` for IMAP4S.

It is IMAPS btw, I don't think I've ever seen IMAP4S?

> What about adding to this PR/Draft, by enabling `"465:465"` in the port mapping of docker-compose.yml.dist? 

Yes, that seems like a good improvement.

The support for port 465 in the Postfix `master.cf` config added it in 2017 using the `smtps` service port name, this was before RFC 8314, but they also added requiring authentication in the config, so afaik that is configured as intended(implicit TLS as a submissions port). I guess it was something the contributor missed with their 2017 PR(including tomav who approved it).

Regarding the config, the correct service name according to RFC 8314 would be `submissions` (587 is `submission`) instead of the legacy `smtps`, but I'm not sure if that's supported by Postfix, or that it matters much.

IIRC, we do however lack support in Postfix config (or documented example) of using Implicit TLS on port 465 for relay (outbound email), there is a lengthy issue about it (where I started to get involved in the project and my learning journey).

youtous:
> > I guess this Draft/PR could be closed?
> 
> Yes, I think so.
> 
> @youtous could you close your PR? Thanks.

Sorry for the delay, and thank you for all this work <3
Awesome!

Issue #1274, opened 2019-09-25
letsencrypt certificate search path


neuralp:
**EDITED**

The root of this problem has a lot less to do with the _mydestinations_ in the main.cf file and more to do with finding the letsencrypt certificate in the proper path.  So I've changed the title and amended this on to the original report.

## Context

In the process of setting up my mail server, I am setting my environment variables to 

```
HOSTNAME=mail
DOMAINNAME=abc.com
CONTAINER_NAME=mail
OVERRIDE_HOSTNAME=abc.com
```
The override hostname is intentionally set because the letsencrypt certificate will only work for the base of the domain (abc.com) and not (mail.abc.com) and are stored in the abc.com folder when letsencrypt is mounted in the container.  In order for the container to automatically find the correct certificate, the hostname can only be of the domain.com format.

This is all fine with the exception of the virtual tables created when I add an e-mail to the database.  If I add an e-mail for 'joe@abc.com' to the container (using the setup.sh script) it will add a virtual domain and address for this.  The postfix main.cf file has the line:

```
mydestination = $myhostname, localhost.$mydomain, localhost
```
Because $myhostname has been overridden (instead of using the output of 'hostname') to get lets encrypt to work, the mydestination contains abc.com and the virtual domain tables contain abc.com.  This obviously creates a conflict and as such mail will not be delivered (in my setup anyhow).  I don't remember if this was an ERROR or a WARNING, but either way the end result is that mail is not delivered.

## Expected Behavior

I would expect there to be no conflict between 'mydestination' and the 'vritual' tables used for other domains.

## Actual Behavior
Mail is not delivered because there is a conflict between the same hostname in 'mydestination' and the 'virtual' mail tables.

## Possible Fix

Every time I download a new image of the container, I stand it up as is.  After that I log into the container and edit the main.cf in the postfix directory and remove the $myhostname from the 'mydestination=' line in the file.  I then just commit my own image and stand up containers from that.

I propose that the $myhostname be removed from the 'mydestination' configuration item.  Correct me if I am wrong, but should all email accounts exist exclusively in the virtual tables?

## Steps to Reproduce

1. set OVERRIDE_HOSTNAME=abc.com in the environment
2. add an e-mail address for the same domain as the OVERRIDE_HOSTNAME is set to
3. conflict will then exist between mydestinations and virtual tables

## Your Environment

* Amount of RAM available: 1G total for virtual server I'm running this on
* Mailserver version used: Image ID 8ae795dc146c
[...]

PR #1509, opened 2020-05-12
Modified letsencrypt support to add domain name checking in addition to

neuralp:
hostname checking.  Added necessary tests and renamed original manual
ssl test to a name that supports adding the other SSL tests.

Reference to issue #1274


erik-wramner:
There are changes not only for letsencrypt but also for the other alternatives. Are there real changes there, or is it only indentation? If there are real changes I need to review more carefully.

There are some places where the if and fi are not aligned, probably you have mixed tabs and spaces. Can you fix that, please? I know that the script is full of indentation problems as is, but at least we can make sure the new parts are aligned.

test/mail_ssl_letsencrypt.bats
```diff
@@ -0,0 +1,66 @@
...
+    run docker exec mail_lets_hostname /bin/sh -c 'grep -ir "/etc/letsencrypt/live/mail.my-domain.com/" /etc/postfix/main.cf | wc -l'
+    assert_success
+    assert_output 2
+    run docker exec mail_lets_hostname /bin/sh -c 'grep -ir "/etc/letsencrypt/live/mail.my-domain.com/" /etc/dovecot/conf.d/10-ssl.conf | wc -l'
+    assert_success
+    assert_output 2
+}
+
+@test "checking ssl: letsencrypt cert works correctly" {
+    run docker exec mail_lets_domain /bin/sh -c "timeout 1 openssl s_client -connect 0.0.0.0:587 -starttls smtp -CApath /etc/ssl/certs/ | grep 'Verify return code: 10 (certificate has expired)'"
```
youtous, line 57: Just a thought; could you cover `465` too?

test/mail_ssl_letsencrypt.bats
```diff
@@ -0,0 +1,66 @@
...
+    docker rm -f mail_lets_hostname
+}
+
+# this test must come first to reliably identify when to run setup_file
+@test "first" {
+    skip 'Starting testing of letsencrypt SSL'
+}
+
+@test "checking ssl: letsencrypt configuration is correct" {
+    run docker exec mail_lets_domain /bin/sh -c 'grep -ir "/etc/letsencrypt/live/my-domain.com/" /etc/postfix/main.cf | wc -l'
```
youtous, line 42: `postconf` tests are more explicit than counting word

test/mail_ssl_letsencrypt.bats
```diff
@@ -0,0 +1,66 @@
...
+    assert_success
+    assert_output 2
+    run docker exec mail_lets_domain /bin/sh -c 'grep -ir "/etc/letsencrypt/live/my-domain.com/" /etc/dovecot/conf.d/10-ssl.conf | wc -l'
```
youtous, line 45: Same for dovecot, `doveconf` is more accurate (https://wiki.dovecot.org/Tools/Doveconf)

target/start-mailserver.sh
```diff
@@ -1042,119 +1042,131 @@ function _setup_ssl() {
...
 
-					notify 'inf' "SSL configured with 'letsencrypt' certificates"
-				else
-					notify 'err' "Key filename not set!"
-				fi
-			else
-				notify 'err' "Cannot access '/etc/letsencrypt/live/"$HOSTNAME"/fullchain.pem'"
-			fi
 		;;
-	"custom" )
```
youtous, line 1075: As @erik-wramner suggested, could you unstage these changes if this is not part of this PR?

target/start-mailserver.sh
```diff
@@ -1042,119 +1042,131 @@ function _setup_ssl() {
...
+            local LETSENCRYPT_DOMAIN=""
+            local LETSENCRYPT_KEY=""
+			
+            # first determine the letsencrypt domain by checking both the full hostname or just the domainname if a SAN is used in the cert
+            if [ -e "/etc/letsencrypt/live/$HOSTNAME/fullchain.pem" ]; then
+                LETSENCRYPT_DOMAIN=$HOSTNAME
+            elif [ -e "/etc/letsencrypt/live/$DOMAINNAME/fullchain.pem" ]; then
+                LETSENCRYPT_DOMAIN=$DOMAINNAME
+			else
+				notify 'err' "Cannot access '/etc/letsencrypt/live/"$HOSTNAME"/fullchain.pem' or '/etc/letsencrypt/live/"$DOMAINNAME"/fullchain.pem'"
```
youtous, line 1057: Shouldn't the script terminate with an error here? `return 1` or something like that?

target/start-mailserver.sh
```diff
@@ -1042,119 +1042,131 @@ function _setup_ssl() {
...
+
+            # then determine the keyfile to use
+			if [ -n "$LETSENCRYPT_DOMAIN" ]; then
+				if [ -e "/etc/letsencrypt/live/$LETSENCRYPT_DOMAIN/privkey.pem" ]; then
+					LETSENCRYPT_KEY="privkey"
+				elif [ -e "/etc/letsencrypt/live/$LETSENCRYPT_DOMAIN/key.pem" ]; then
+					LETSENCRYPT_KEY="key"
 				else
-					notify 'err' "Cannot access '/etc/letsencrypt/live/"$HOSTNAME"/privkey.pem' nor 'key.pem'"
+					notify 'err' "Cannot access '/etc/letsencrypt/live/"$LETSENCRYPT_DOMAIN"/privkey.pem' nor 'key.pem'"
```
youtous, line 1067: Same here

neuralp:
> There are changes not only for letsencrypt but also for the other alternatives. Are there real changes there, or is it only indentation? If there are real changes I need to review more carefully.
> 
> There are so(me places where the if and fi are not aligned, probably you have mixed tabs and spaces. Can you fix that, please? I know that the script is full of indentation problems as is, but at least we can make sure the new parts are aligned.

Since I was working on the SSL function, I figured that I would attempt to fix the indentation since it was all over the place.  I'll revisit this for the function and make sure that all of the tabs get converted to spaces.  Apparently my editor didn't read the .editorconfig.

test/mail_ssl_letsencrypt.bats
neuralp: To tell you the truth, the tests were already this way (the word count way) so I just duplicated them for the second container. I'll take a look at a more accurate test for this then, not a problem.

target/start-mailserver.sh
neuralp: Just trying to fix indentation, I will remove and follow the .editorconfig more closely.

target/start-mailserver.sh
neuralp: Currently if SSL encounters an error, it just leaves the default configuration in place.  It does make sense that if you configured SSL, it would be assumed to be running correctly if no errors were actually returned from the function.  A return 1 sounds look a good thing to implement.

test/mail_ssl_letsencrypt.bats
neuralp: Could you please clarify what you are referring to? I can't find an issue with that ID or a PR.

neuralp: Never mind, I've got it.  Test port 465 as well.

neuralp: 465 is now covered in the tests with the updated pull request

test/mail_ssl_letsencrypt.bats
neuralp: tests now use _postconf_ and search for the full expected certificate name

test/mail_ssl_letsencrypt.bats
neuralp: same for _doveconf_ which is now used and checked

target/start-mailserver.sh
neuralp: indentation updates have been corrected and dropped

target/start-mailserver.sh
neuralp: In addition to the error, return 1 is used after the error.  Upon successfull completion of the letsencrypt portion, return 0 is given.

target/start-mailserver.sh
neuralp: Resolved with return 1 upon error.

neuralp:
Your suggestions have been taken into account and the PR has been modified to include these suggestions.  Please review at your earliest convenience.

erik-wramner:
Looks good, thanks!

PR #1514, opened 2020-05-18
Fix #1507 - Do not try to reload freshclam daemon after log rotation

### LINKED ISSUE #1507: Logrotate Error
Got this today:

```
/etc/cron.daily/logrotate:
logrotate_script: 6: logrotate_script: Syntax error: "fi" unexpected
error: error running non-shared postrotate script for /var/log/mail/freshclam.log of '/var/log/mail/freshclam.log '
```

gmasse:
Freshclam is not run as daemon. Fix #1507


erik-wramner:
Much better than my "fix". Thanks.

PR #1535, opened 2020-06-11
Improve container/image name detection

casperklein:
Problem: `setup.sh` fails, if more than one container uses `CMD ["supervisord", "-c", "/etc/supervisor/supervisord.conf"]`

Current container/image name detection: List all containers, grep for "supervisor".

New approach: List container, with label `org.label-schema.name="docker-mailserver"`

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -25,13 +25,10 @@ if [ -z "$CRI" ]; then
   fi
 fi
 
-INFO=$($CRI ps \
-  --no-trunc \
-  --format="{{.Image}} {{.Names}} {{.Command}}" | \
-  grep "supervisord -c /etc/supervisor/supervisord.conf")
+INFO=$($CRI ps --no-trunc --format "{{.Image}};{{.Names}}" --filter label=org.label-schema.name="docker-mailserver")
 
-IMAGE_NAME=$(echo $INFO | awk '{print $1}')
-CONTAINER_NAME=$(echo $INFO | awk '{print $2}')
+IMAGE_NAME=${INFO%;*}
+CONTAINER_NAME=${INFO#*;}
 DEFAULT_CONFIG_PATH="$(pwd)/config"
 USE_CONTAINER=false
 
--- a/setup.sh
+++ b/setup.sh
@@ -25,7 +25,7 @@ if [ -z "$CRI" ]; then
   fi
 fi
 
-INFO=$($CRI ps --no-trunc --format "{{.Image}};{{.Names}}" --filter label=org.label-schema.name="docker-mailserver")
+INFO=$($CRI ps --no-trunc --format "{{.Image}};{{.Names}}" --filter label=org.label-schema.name="docker-mailserver" | tail -1)
 
 IMAGE_NAME=${INFO%;*}
 CONTAINER_NAME=${INFO#*;}
--- a/setup.sh
+++ b/setup.sh
@@ -25,7 +25,11 @@ if [ -z "$CRI" ]; then
   fi
 fi
 
-INFO=$($CRI ps --no-trunc --format "{{.Image}};{{.Names}}" --filter label=org.label-schema.name="docker-mailserver" | tail -1)
+INFO=$($CRI ps \
+  --no-trunc \
+  --format "{{.Image}};{{.Names}}" \
+  --filter label=org.label-schema.name="docker-mailserver" | \
+  tail -1)
 
 IMAGE_NAME=${INFO%;*}
 CONTAINER_NAME=${INFO#*;}


polarathene:
Any reason you changed the command for getting INFO to a single line instead of leaving it formatted to multi-line? Requires scrolling horizontally now to view in full, and is a bit less readable as a result.

The next two changes, the previous statements were easier to grok roughly what was happening, even to someone not too familiar with the shell and awk, but your replacement looks bewildering(and perhaps less obvious for someone to google what is happening if they don't bother checking `git blame`). Perhaps add a comment? (I personally am not familiar with the syntax you've used myself, but echo/awk was clear enough)

casperklein:
> Any reason you changed the command for getting INFO to a single line instead of leaving it formatted to multi-line?

No. I changed it to multi-line.

> The next two changes, the previous statements were easier to grok ..

As a best-practice, I used the shell build-in functions instead of calling external programs. Because _Parameter expansions_ are already heavily used in this project, I think explicit comments are not necessary.

polarathene:
I disagree personally. That specific syntax is not used in this shell file for either variant, and isn't immediately obvious what is being done to those not familiar with it, nor obvious what keywords to look for when trying to look up what that syntax is imo.

Whomever is to approve the PR can make that call I guess, at the very least the commit can probably be traced back to this PR thread via github or the commit diff itself should imply they're potential equivalents to the prior statements.

[This article](https://linuxconfig.org/introduction-to-bash-shell-parameter-expansions#h5-1-1-remove-matching-pattern-from-the-beginning-of-the-string) explains it well as substring extraction. You're taking the `$INFO` variable and everything before the first `;` as the container name, and everything after the last `;`as the image name. 

Honestly, not something I personally would remember with all the other things I do, rarely needing to use such syntax when dealing with the shell. I don't know about you, but for stuff I rarely use and then have to come back to for maintenance/refactor 6 months or years later, I'd wish I had left a comment, same when I contribute to a third-party project and run into something made terse for the sake of it. 

If you're regularly using that syntax, I can understand why it seems unnecessary to throw in a comment. I have looked at other shell files in the project, most use of `#` is in grep/sed regular expressions which is more recognizable and acceptable, `%` for this syntax is far more rare though(within this project):

https://github.com/tomav/docker-mailserver/blob/6abcc8ce5fa9352e1b5aedc2340f31dc72e6f657/target/helper_functions.sh#L21-L22

https://github.com/tomav/docker-mailserver/blob/6abcc8ce5fa9352e1b5aedc2340f31dc72e6f657/target/postsrsd-wrapper.sh#L7

Heavy usage in [`start-mailserver.sh`](https://github.com/tomav/docker-mailserver/blob/master/target/start-mailserver.sh), but sed/grep specific.

Not sure if this is this is the same, but perhaps illustrates my issue with it:

https://github.com/tomav/docker-mailserver/blob/6abcc8ce5fa9352e1b5aedc2340f31dc72e6f657/target/check-for-changes.sh#L128

---

> I think explicit comments are not necessary.

I respectfully disagree, and see no harm in adding something like `# Get values from `$INFO` via substring removal patterns`. I could then look up something like `bash substring removal` to understand the syntax.

That aside, seems like a great contribution, thanks! :)

erik-wramner:
Looks good, thanks.

casperklein:
@polarathene Sorry for the late reply.

> and isn't immediately obvious what is being done to those not familiar with it

That is true for all kind of things, isn't it? If you are e.g. not familar with `sed/awk`, it can also be hard to understand whats going on exactly. In that cases, I think it's a good opportunity to learn something new 😉

> I have looked at other shell files in the project, most use of # is in grep/sed regular expressions

I don't know what you are referring to exactly, but that (grep/sed) has nothing to do with shell expansion. I guess you mix things up here.

Some files I had in mind were:

    test/bats/install.sh:13:    cd "${path%/*}"
    test/bats/install.sh:14:    local name="${path##*/}"
    test/bats/libexec/bats:38:    cd "${path%/*}"
    test/bats/libexec/bats:39:    local name="${path##*/}"
    test/bats/libexec/bats-exec-test:170:  local lineno="${frame%% *}"
    test/bats/libexec/bats-exec-test:176:  local rest="${frame#* }"
    test/bats/libexec/bats-exec-test:177:  local fn="${rest%% *}"
    test/bats/libexec/bats-exec-test:183:  local rest="${frame#* }"
    test/bats/libexec/bats-exec-test:184:  local filename="${rest#* }"
    test/bats/libexec/bats-exec-test:307:  BATS_TMPDIR="${TMPDIR%/}"
    test/bats/libexec/bats-format-tap-stream:142:    name="${line#* $index }"
    test/bats/bin/bats:38:    cd "${path%/*}"
    test/bats/bin/bats:39:    local name="${path##*/}"
    target/postsrsd-wrapper.sh:7:  domain_name="${OVERRIDE_HOSTNAME#*.}"
    target/helper_functions.sh:21:  IP=${1%%/*}
    target/helper_functions.sh:22:  PREFIX_LENGTH=${1#*/}

> Not sure if this is this is the same, but perhaps illustrates my issue with it:

No it isn't. The `#` is before the variable name and has a complete different meaning. ${#FOO} returns the string length of the variable and ${#FOO[@]} returns the element count of the array.

Here are some nice overviews with examples I can recommend:

- https://devhints.io/bash#parameter-expansions
- https://github.com/dylanaraps/pure-bash-bible#parameter-expansion

Independently of that we got different opinions at some points, I appreciated your feedback. 👍

polarathene:
> That is true for all kind of things, isn't it? If you are e.g. not familar with `sed`/`awk`, it can also be hard to understand whats going on exactly. In that cases, I think it's a good opportunity to learn something new

Well with sed/awk those are a bit more obvious what to google, and the option flags to look up. When the syntax is a variable with a single character like `%` or `#` as the feature, it's a little less obvious and I wouldn't expect google to be super helpful with `bash %` or `bash #`, the results don't indicate much either, you know it's bash, but have to go over the syntax in general and scan for where it might be relevant.

Further more the full part was `${INFO#*;}`, not being all that experienced in bash, or knowing this particular syntax at the time, I don't know what the role of `*;` plays into that as well, and that wouldn't help with the google keywords. 

Yes, it can be figured out, but it detracts time for someone to go and grok vs a small comment that explains what's going on from the decision to be terse. You could show the before and after to a variety of devs, and most will likely have a fair idea of what's going on with the awk variant(without knowing about awk), and those who don't know bash are perhaps going to have a hard time looking at those two lines to understand what they're achieving.

This is just my own perspective and based on my past experiences since I've had to learn and work with a lot of things that I've regretted not documenting something I rarely have experience with but understood well at the time originally writing it. 

Regardless, it got merged as-is, so doesn't seem to be a concern for maintaining this project.

---

> I don't know what you are referring to exactly, but that (grep/sed) has nothing to do with shell expansion. I guess you mix things up here.

Quite possibly! Furthers my point though about how obscure it looks to those who want to grok something let alone contribute. Time to grok by seeking external explanation vs a comment can be a valid concern and add up. 

Here's a line for grep and sed that had `#` in use, I assumed since both are using them as  a pattern syntax, that it had the same usage here.

https://github.com/tomav/docker-mailserver/blob/e1503073393958b8fd8839dc7bf3ae108a25d2dc/target/start-mailserver.sh#L587

https://github.com/tomav/docker-mailserver/blob/e1503073393958b8fd8839dc7bf3ae108a25d2dc/target/start-mailserver.sh#L1366

> Some files I had in mind were

Ah, I missed all those bats ones :) fair enough!

> No it isn't. The `#` is before the variable name and has a complete different meaning. ${#FOO} returns the string length of the variable and ${#FOO[@]} returns the element count of the array.

Illustrates the hassle to grok it via external information when all that is evident to a user that doesn't understand is `bash` + `#`, not the relevance of before or after the variable having different meaning.

If this were an internal project, it'd be less of an issue, especially with all the test files where the syntax is in use. You made a good case for that. But for the files a user may want to review/audit, not just contribute to, it works against their benefit and is cryptic to look at the first time. 

I suppose if they're auditing trustworthiness of what the script is doing, they should probably know bash syntax to begin with though.

> Here are some nice overviews with examples I can recommend:

Cheers, I linked to an article that covered it for me, I think I had viewed a few results on google prior with no explanation.

> Independently of that we got different opinions at some points, I appreciated your feedback.

All good, you've made good points to justify it, I don't disagree with you entirely. Cheers for the response.

PR #1547, opened 2020-06-22
Fix dovecot variable with whitespace

MrFreezeex:
Signed-off-by: Arthur Outhenin-Chalandre <arthur@cri.epita.fr>

This fixes shell expansion for dovecot variables by adding quotes. I saw this bug while trying to use a ldap bind dn with a space in it.


erik-wramner:
The tests fail. I don't think it is caused by your change, but unfortunately I can't trigger them manually, can you make a nonsense commit please?

MrFreezeex:
Done, the CI is green now :tada:

PR #1550, opened 2020-06-28
Fix inconsistent indentation

GiovanH:


erik-wramner:
The tests fail. Please push again to trigger a re-run.

aendeavor:
During my work on #1595, I realised I had seen the same error. Could this be related to an error in `setup.sh` @erik-wramner ?

erik-wramner:
@aendeavor it could but this is one of a few errors that crop up from time to time. It is irritating. The tests are much more stable now than they were a year ago, but they are still not 100%.

aendeavor:
> @aendeavor it could but this is one of a few errors that crop up from time to time. It is irritating. The tests are much more stable now than they were a year ago, but they are still not 100%.

Looking at #1595 too, what can I do about that? After we decide with @fbartels on whether to use `bash`, I would implement the error tracing and push again. In the meantime, this error needs to be resolved. But I don't know how.

aendeavor:
Will be closed as soon as #1601 is merged and I open the new PR for this.

aendeavor:
Obsolete due to #1605.

PR #1553, opened 2020-06-30
Letsencrypt traefik v2 acme json

### LINKED ISSUE #1525: Update LetsEncrypt certificates
Hello,

Just want to confirm that this is the correct procedure to update the LetsEncrypt certificates every ~3 months:

I have in my docker-compose file:

```
    volumes:
    - /opt/mailserver/letsencrypt:/etc/letsencrypt
```

I generate/renew the certificates on the host machine, then:

```
cp -R -L /etc/letsencrypt/live/mail.mydomain.com/* /opt/mailserver/letsencrypt/live/mail.mydomain.com/

docker exec mailserver openssl s_client -connect 0.0.0.0:25 -starttls smtp -CApath /etc/ssl/certs/
Verify return code: 0 (ok)

docker exec mailserver openssl s_client -connect 0.0.0.0:143 -starttls smtp -CApath /etc/ssl/certs/
Verify return code: 0 (ok)
```

Is this correct and enough? I don't need to restart any services or even the container?

Thank you very much :)

MichaelSp:
Will extract certificates from acme.json as written by traefik v2 for usage in dovecot and postfix.
Also watches acme.json for changes. 

For this to work the file has to be mounted/present at `/etc/letsencrypt/acme.json`

This suppose to replace https://github.com/tomav/docker-mailserver/pull/976 and fix #1525


erik-wramner:
Thanks for the PR. Many of the changes seem to be indentation fixes, which is not ideal, but never mind.

First question is if this is compatible with non-letsencrypt setups? I think it should be (the file is ignored if missing), am I right?

Second question is traefik. I'm not sure how common it is to use traefik for letsencrypt in this community? Most issues I've seen mention caddy. It would have been nice to have a solution not tied to traefik (though of course supporting one product is better than supporting none...). Is the traefik format supported by other tools, for example caddy?

MichaelSp:
indentation fixes: yes, you are right. Not ideal, but also not easy to keep the files clean. I had to laugh when reading this comment from almost 1 year ago: `        # Not fixing indentation yet to reduce diff (fix later in separate commit)` :) I tried to find a sweet-spot.

1. If there is no `acme.json` nothing happens. 
2. I cant say much about the rest of the community. It certainly scratches my itch. I was also not able to find any documentation on the caddy format. The traefik acme.json is also poorly documented. Here is an example: https://github.com/tomav/docker-mailserver/blob/32c732e276b2637cdda1fd1cc8b5ba318a765687/test/config/letsencrypt/acme-changed.json The `le` is something like "Encryption Provider Identifier" and can be chosen freely.

youtous:
It would be nice to use a different variable than `$DOMAINNAME` / `$HOSTNAME` to allow certificates from  wildcard name.

> The traefik acme.json is also poorly documented.

And is not compatible with traefik v1. 
Concerning this point, how do you handle errors when loading an `acme.json` from [traefik v1](https://github.com/youtous/docker-mailserver-traefik/tree/master/fixtures)?

MichaelSp:
Yes, this is traefik v2 (updated the PR-name).

I'll update the PR to support something like `CERTIFICATE_DOMAIN` with fall-back to `DOMAINNAME` / `HOSTNAME`

MichaelSp:
Since I don't use `DNS-01` (which is required to generated wildcard certs), I couldn't find out how a `acme.json` would look like in this case.
Can you give me an example? Is it this?

```json
 {
  "le": {
    "Account": {},
    "Certificates": [
      {
        "domain": {
          "main": "*.example.com",
          "sans": []
        },
        "certificate": "X",
        "key": "Y",
        "Store": "default"
      }
    ]
  }
}

```

youtous:
I'm not using `DNS-01` but I already faced an user asking for it.
According to https://stackoverflow.com/a/58171898, it should be something like (not tested):
```json
 {
  "le": {
    "Account": {},
    "Certificates": [
      {
        "domain": {
          "main": "example.com",
          "sans": ["example.com", "*.example.com"]
        },
        "certificate": "X",
        "key": "Y",
        "Store": "default"
      }
    ]
  }
}
```

MichaelSp:
again an unrelated test failure as in #1560

erik-wramner:
Indeed, it is irritating. Just do a nonsense change (or force push with no change?) to trigger a rebuild. I used to be able to trigger that manually, but I can't.

However, there are several changes in this PR that I don't understand:

- disable_plaintext_auth=no (used to be yes in comments)
- ssl=yes (used to be required)
- hostname and postmaster address (used to be undefined)
- quotas
- recipient_delimiter

Why have you changed that?

MichaelSp:
Oh I'm sorry thanks for your careful observation.
I guess these files are changed by the tests. I shouldn't have committed the changes...

erik-wramner:
That's better. Unfortunately that stupid test that keeps failing is at it again. You have unusually bad luck, it normally happens less than one build in ten. Can you push a gain to trigger a rebuild, please? We really have to fix that, no time though. The problem with working for free :-).

MichaelSp:
now it's green @erik-wramner

wernerfred:
Nice work. @MichaelSp will you update the Wiki with instructions on how to use traefik with docker-mailserver? Or do i misinterprete the current information in the wiki about traefik?

MichaelSp:
@wernerfred thanks for the heads up. I've just updated the wiki: https://github.com/tomav/docker-mailserver/wiki/Configure-SSL#traefik

wernerfred:
Functionality somewhat broken currently anyways: See #2239 
Maybe someone from this thread has an idea 💡

MohammedNoureldin:
@polarathene, So the workflow should look like this:

1. The user wants to use letsencrypt with ***Nginx*** and acma-companion, then only set `SSL_TYPE=letsencrypt`.
2. The user wants to use letsencrypt with ***Traefik*** (with custom certificate path), then he needs firstly to set `SSL_TYPE=manual and `, `CERT_PATH=SOME_PATH` and `KEY_PATH=SOME_PATH`, then he should somehow be able to tell the **internal** "*certificate auto extractor*" script (I don't know how it is called in the mailserver container) to extract the certs to the same place as defined in the env vars. And it should know that there are some certificates to be extracted because the script will check the existence of *acme.json*?

Is that correct? Could your refacoting consider that please when you have time?

NorseGaud:
> And extracting the certificates directly inside the same folder where the *acme.json* lives is not good, this is exactly my point, because you may have the *acme.json* stored in a Docker named volume, and named volumes can be mounted only as whole (namely as folder), not using single files. 

We can allow a variable to be passed in to override the location of the acme.json. Would that work for you? I have a PR open for some changes of this code and could include it.

MohammedNoureldin:
Hi @NorseGaud,

it is actually about overriding the location of the extracted (generated) *.crt* and *.key*. Would that be possible?

polarathene:
> if I understand correctly, you think that Docker mailserver uses the acme.json directly?

@MohammedNoureldin 

My understanding is the certs content is available in the `acme.json` and we extract that out into `.pem` files. Those files get placed in the container `/etc/letsencrypt/live` AFAIK, as part of the `SSL_TYPE=letsencrypt` support.

> Hopefully I could explain it. Please let me know if I misunderstood or missed anything.

You can use a bind volume mount instead of a data volume. Or mount to a different location in the container and add a symlink to the `acme.json` pointing to that location from `/etc/letsencrypt` should also work.

Likewise you can mount a data volume in the reverse direction, although it's a hassle and I don't recommend it, it would allow you to bindmount read-only the `acme.json` to the `docker-mailserver` container.

I'd personally advise the symlink approach if you insist on using a data volume instead of a bind mount.

---

> I have a PR open for some changes of this code and could include

@NorseGaud 

Delay that to a follow up please.

MohammedNoureldin:
@polarathene please see my comment above, I added it just a 1 second before you, so it may be missed :D

Issue #1526, opened 2020-05-31
Allow for auth_bind LDAP authentication (here: Dovecot)


guardiande:
## Is your feature request related to a problem?
Currently it is not possible to allow for `auth_bind` authentication for Dovecot by simple container configuration.

## Describe the solution you'd like
For security reasons (https://wiki.dovecot.org/AuthDatabase/LDAP/AuthBinds) it is recommended to use BIND authentication against the LDAP server.

I think that configomat should allow for easy implementation of that feature.

## Alternatives you've considered
Going with the password lookup. It works for most people (but then I think Active Directory does not allow password lookups).

## Who will that feature be useful to?
All users looking into more secure setup and maybe AD users.

## What have you done already?
I just started into implementing a container based mail setup and I'm in progress of evaluating the available options. So I did not yet implement this solution.

## What are you going to contribute?
I'm neither an export on Dovecot nor LDAP. But of course I'll do the testing.

aendeavor:
What's the current status here @guardiande ?

guardiande:
@aendeavor Should be fixed by #1557.

PR #1557, opened 2020-07-04
Update dovecot-ldap.conf.ext

ubenmackin:
add auth_bind = no so that it can be overridden via the env-mailserver file used by docker compose.

This is related to #1526

### DIFF
--- a/env-mailserver.dist
+++ b/env-mailserver.dist
@@ -270,6 +270,11 @@ DOVECOT_PASS_FILTER=
 # default is maildir, supported values are: sdbox, mdbox, maildir
 DOVECOT_MAILBOX_FORMAT=maildir
 
+# empty => no
+# yes => Allow bind authentication for LDAP
+# https://wiki.dovecot.org/AuthDatabase/LDAP/AuthBinds
+DOVECOT_AUTH_BIND=
+
 # -----------------------------------------------------------------------------------------------------------------------------
 # ---------------- Postgrey section -------------------------------------------------------------------------------------------
 # -----------------------------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


ubenmackin:
As @guardiande mentions in #1526 auth_bind is needed for some setups. Setting a default of no, so it doesn't break the existing setup. But by having the value in the dovecot-ldap.conf.ext it should be overridable with a env variable like DOVECOT_AUTH_BIND.

erik-wramner:
Is this PR work in progress? You mention an env variable, but it is not there?

guardiande:
> Is this PR work in progress? You mention an env variable, but it is not there?

From `start-mailserver.sh`

```
configomat.sh "DOVECOT_" "/etc/dovecot/dovecot-ldap.conf.ext"
```

So creating the container with (that's the variable you're missing, it's not part of the PR 😊)

```
environment:
      - DOVECOT_AUTH_BIND=yes
```

would allow for closing #1526

ubenmackin:
Ah, ok. I misunderstood. I didn't think we had to specify all environment variables up front, only if you needed it used in the container. So I didn't add this to the env file, so it would retain the current default behavior.

But I can update the env file to include it.

guardiande:
> Ah, ok. I misunderstood. I didn't think we had to specify all environment variables up front, only if you needed it used in the container. So I didn't add this to the env file, so it would retain the current default behavior.
> 
> But I can update the env file to include it.

No, you don't have to include the variable in the env file. You did everything right. Only if someone - like me - wants to have auth_bind set to yes she would have to override the default setting (which is no) to yes by including the `DOVECOT_AUTH_BIND` variable set to yes in the environment.

ubenmackin:
> > Ah, ok. I misunderstood. I didn't think we had to specify all environment variables up front, only if you needed it used in the container. So I didn't add this to the env file, so it would retain the current default behavior.
> > But I can update the env file to include it.
> 
> No, you don't have to include the variable in the env file. You did everything right. Only if someone - like me - wants to have auth_bind set to yes she would have to override the default setting (which is no) to yes by including the `DOVECOT_AUTH_BIND` variable set to yes in the environment.

Ya, I noticed there were a few other DOVECOT_ variables that could be used, but were not specified (DOVECOT_USER_ATTRS, DOVECOT_PASS_ATTRS, and DOVECOT_DEFAULT_PASS_SCHEME). But since the AUTH_BIND wasn't defined in the dovecot ldap.conf, it couldn't be overridden with environment variable.

I just committed another change to add DOVECOT_AUTH_BIND to the env file, so if @erik-wramner wants to review now, I think this can be pulled.

PR #1560, opened 2020-07-07
consistent format according to .editorconfig

MichaelSp:
feel free to ignore/close if you don't like it


MichaelSp:
is this test still known to be flaky?

```
 # Dovecot has been restarted, but this test often fails so presumably it may not be ready
  # Add a short sleep to see if that helps to make the test more stable
  # Alternatively we could login with a known good user to make sure that the service is up
```

MichaelSp:
seems there is no need for this

PR #1570, opened 2020-07-15
Fix sasl_password generation to allow passwords containing hashes

guardiande:
This is a pull request to allow postfix SASL account to contain hashes (#).


erik-wramner:
One of the quota tests failed. Not sure if it was caused by your change, but it could be. Do the tests succeed if you run them locally?

Feel free to push again (force push or make a nonsense change) to trigger a rebuild. If we get the same error again it must have been caused by the change, which otherwise looks good.

guardiande:
Made a dummy change and pushed again. Works now. Will revert the dummy change and push again.

guardiande:
Hmm... Failed again. Seems to be something indeterministic in the tests as the failing do not seem to have anything in common with the small change.

I'd suggest merging.

PR #1572, opened 2020-07-19
Improve acme.json parsing when it contains multiple certs

julianbreiteneicher:
Hi,

first of all, thanks for your work on this project! I just discovered it recently and it's really awesome!  :smiley:

This PR improves the certificate parsing of traefik's acme.json file.
The parsing fails if the acme.json file contains multiple certificates because the helper method always tries to extract the first cert even if it's the wrong one.


martinwepner:
Hi @julianbreiteneicher, I just found out, that we may ran into similar problems. I think the current python code is not very readable, I thought it'd be better to refactor it. 
Could you check, if your issue is also fixed by my PR in #1574?

julianbreiteneicher:
Hi @martinwepner, thanks for the hint! I just tested your PR and it also fixes my issue. I agree that your refactoring makes the code more readable.
So since your PR solves both issues, I think mine should be closed...

erik-wramner:
Right, closing. Thanks anyway!

PR #1573, opened 2020-07-19
addalias: check if two arguments are given

casperklein:
Fix for https://github.com/tomav/docker-mailserver/issues/1563

### DIFF
--- a/target/bin/addalias
+++ b/target/bin/addalias
@@ -18,7 +18,8 @@ escape() {
 	echo "${1//./\\.}"
 }
 
-[ -z "$EMAIL" ] && { usage; errex "no email specified"; }
+[ -z "$EMAIL" ] && { usage; errex "Error: No alias specified"; }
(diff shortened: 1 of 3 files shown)


erik-wramner:
Does `delalias` really need a recipient? Isn't it enough for that command to specify the alias itself?

casperklein:
Currently `$RECIPIENT` is needed: https://github.com/tomav/docker-mailserver/blob/398b1dd554ce41c03767079919744affead6b4a7/target/bin/delalias#L27-L29

I think we can simplify that to this:

    sed -i '/^$EMAIL\s/d' $DATABASE

However, I am not sure, which use cases are covered by line 28/29 above.

casperklein:
I wasn't aware, that you can define one alias, with more than one recipient:

https://github.com/tomav/docker-mailserver/issues/909
https://github.com/tomav/docker-mailserver/pull/910

Therefor, the recipient argument is mandatory.

Issue #1590, opened 2020-08-13
shellcheck & setup.sh


aendeavor:
## Subject

**Other - `setup.sh` may be prone to errors.** Related to #1228 and #1258.

## Description

Shellcheck (`v0.7.0`) is complaining alot about `setup.sh`. You may wan't to have a look:

![shellcheck_complains](https://user-images.githubusercontent.com/44545919/90148268-731dab80-dd83-11ea-9778-40f62bb731e3.png)

erik-wramner:
There is already a PR for this in #1258 but it is stuck. If you have the skills and time please feel free to continue that work or by all means start fresh. If you do start over, please check the PR first to avoid double work and mistakes.

aendeavor:
Done with the merge of #1595.

Issue #1137, opened 2019-03-05
Gmail only 550-5.7.1 this message does not have authentication information or fails to pass authentication checks


voordev:
## Context
Mailserver with SPF records added, with SPF config added.
Works fine on all domains except gmail.com => mails get returned with Gmail only 550-5.7.1 this message does not have authentication information or fails to pass authentication checks

## Expected Behavior
Mail should get delivered to gmail account.

## Actual Behavior
Gmail only 550-5.7.1 this message does not have authentication information or fails to pass authentication checks

## Steps to Reproduce
```
version: '2'
services:
  mail:
    image: tvial/docker-mailserver:latest
    hostname: mail
    domainname: domain.ltd
    container_name: mail
    ports:
    - "25:25"
    - "143:143"
    - "587:587"
    - "993:993"
    volumes:
      - ./mail/data/mail:/var/mail
      - ./mail/data/mail-state:/var/mail-state
      - ./mail/config/:/tmp/docker-mailserver/
      - /server/letsencrypt/etc:/etc/letsencrypt/live
      - /server/mailserver/var/mail:/var/mail
      - ./config/postfix-policyd-spf.conf:/etc/postfix-policyd-spf-python/policyd-spf.conf
    environment:
    - ENABLE_SPAMASSASSIN=1
    - ENABLE_CLAMAV=0
    - ENABLE_FAIL2BAN=1
    - ENABLE_POSTGREY=1
    - ONE_DIR=1
    - DMS_DEBUG=0
    #- DMS_DEBUG=1
    #- ENABLE_CLAMAV=1
    #- ONE_DIR=${ONE_DIR}
    #- ENABLE_POP3=${ENABLE_POP3}
    #- ENABLE_FAIL2BAN=${ENABLE_FAIL2BAN}
    #- ENABLE_MANAGESIEVE=${ENABLE_MANAGESIEVE}
    #- OVERRIDE_HOSTNAME=${OVERRIDE_HOSTNAME}
    #- POSTMASTER_ADDRESS=${POSTMASTER_ADDRESS}
    #- POSTSCREEN_ACTION=${POSTSCREEN_ACTION}
    #- REPORT_RECIPIENT=${REPORT_RECIPIENT}
    #- REPORT_SENDER=${REPORT_SENDER}
    #- REPORT_INTERVAL=${REPORT_INTERVAL}
    #- SMTP_ONLY=${SMTP_ONLY}
    - SSL_TYPE=letsencrypt
[...]

PR #1595, opened 2020-08-24
major refactoring for setup.sh (#1590)

aendeavor:
# Improvements for `setup.sh`

Following #1590, I re-wrote and refactored `setup.sh` to

- make shellcheck happy
- improve the overall readability of the script
- make the script consistent

**This PR obsoletes** #1258.

## Rationale

Looking at #1258, I adopted some changes, namely `set -eu` and the use of default values. This PR has the podman changes already in it and does therefore not need to be rebased. Looking at the use of `set -eu`, I am really not sure whether this is the best solution to be honest, and I would like to hear an opinion on this.

Furthermore, I restructured the script, setting description and default values in the beginning, followed by all functions. I believe this to be more consistent.

I also included the changes to `README.md` to reflect what @fbartels already did in #1258.

## Disclaimer

Not sure whether I made (one or more) (minor / major) mistakes while refactoring. The test will show. Bash is still Bash in the end.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,6 @@
 
 [![Build Status](https://travis-ci.org/tomav/docker-mailserver.svg?branch=master)](https://travis-ci.org/tomav/docker-mailserver) [![Docker Pulls](https://img.shields.io/docker/pulls/tvial/docker-mailserver.svg)](https://hub.docker.com/r/tvial/docker-mailserver/) [![Docker layers](https://images.microbadger.com/badges/image/tvial/docker-mailserver.svg)](https://microbadger.com/images/tvial/docker-mailserver) [![Github Stars](https://img.shields.io/github/stars/tomav/docker-mailserver.svg?label=github%20%E2%98%85)](https://github.com/tomav/docker-mailserver/) [![Github Stars](https://img.shields.io/github/contributors/tomav/docker-mailserver.svg)](https://github.com/tomav/docker-mailserver/) [![Github Forks](https://img.shields.io/github/forks/tomav/docker-mailserver.svg?label=github%20forks)](https://github.com/tomav/docker-mailserver/) [![Gitter](https://img.shields.io/gitter/room/tomav/docker-mailserver.svg)](https://gitter.im/tomav/docker-mailserver)
 
-
 A fullstack but simple mail server (smtp, imap, antispam, antivirus...).
 Only configuration files, no SQL database. Keep it simple and versioned.
 Easy to deploy and upgrade.
@@ -13,12 +12,12 @@ At this point we have merged the next branch based on Debian Buster into master.
 That means the docker image latest uses Buster. The change may break things!
 
 The following possibly breaking changes are known:
+
 - Filebeat is removed and should be handled by another container, see [Wiki](https://github.com/tomav/docker-mailserver/wiki/).
 - Dovecot will be downgraded a little bit (same major version) so that we can use the official Debian version.
 
 If you want to stick to the old version a while longer, either switch to stable or to a specific version.
 If you run into problems, please raise issues and ask for help. Don't forget to provide details.
-
 
 Includes:
 
@@ -50,11 +49,13 @@ Before you open an issue, please have a look this `README`, the [Wiki](https://g
 ## Requirements
 
 Recommended:
+
 - 1 CPU
 - 1-2GB RAM
 - Swap enabled for the container
 
 Minimum:
+
 - 1 CPU
 - 512MB RAM
 
@@ -62,7 +63,7 @@ Minimum:
 
 ## Usage
 
-#### Get the tools
+### Get the tools
 
 Download the docker-compose.yml, the .env and the setup.sh files:
 
@@ -76,7 +77,7 @@ curl -o .env https://raw.githubusercontent.com/tomav/docker-mailserver/master/.e
 curl -o env-mailserver https://raw.githubusercontent.com/tomav/docker-mailserver/master/env-mailserver.dist
 ```
 
-#### Create a docker-compose environment
+### Create a docker-compose environment
 
 - Edit the files `.env` and `env-mailserver` to your liking:
   - `.env` contains the configuration for docker-compose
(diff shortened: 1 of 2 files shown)


fbartels:
Thanks for reviving and completing my work from #1258 I did close this pr just now.

> Looking at the use of set -eu, I am really not sure whether this is the best solution to be honest, and I would like to hear an opinion on this.

The main intention behind this is to force some "hygiene" when working with bash scripts. `set -u` makes scripts terminate when an assigned variable does not have a value and `set -e` terminates the script when one of the commands it uses gives a non-zero return code (i.e. produced an error).

aendeavor:
Alright. I was aware of what `set -eu` does, I was just curious because I read that the use of `set -e` can have unintentional side effects. Sadly, the Travis build failed. Not sure why thought. @fbartels, can you help?

fbartels:
> the use of set -e can have unintentional side effects

I would say it the other way around. If a step in between fails, but the script continues that could very much have unintentional side effects. But yes, forcing `set -e` will mean that scripts need to come with a certain level of robustness, make them clean up after themselves, etc. I am ok with removing it, but keeping it forces a higher level of care.

The concrete errors I see are:

```
not ok 257 checking setup.sh: setup.sh email add and login
#  in test file test/tests.bats, line 1251)
#   `assert_output "passdb: setup_email_add@example.com auth succeeded"' failed
# 
# -- output differs --
# expected : passdb: setup_email_add@example.com auth succeeded
# actual   : passdb: setup_email_add@example.com auth failed
# --
# 
```
Test https://github.com/tomav/docker-mailserver/blob/f225e14a219824e4ebb34e63b144e6d7c1314bcb/test/tests.bats#L1250-L1251

This indeed looks like its unrelated to your change.

```
not ok 267 checking setup.sh: setup.sh config dkim
# (from function `assert_success' in file test/test_helper/bats-assert/src/assert.bash, line 114,
#  in test file test/tests.bats, line 1415)
#   `assert_success' failed
# 
# -- command failed --
# status : 2
# output : ./setup.sh: 279: ./setup.sh: 2: parameter not set
# --
# 
```
Not enough parameters left after `fail2ban       ) shift ; _docker_container fail2ban "$@" ;;`?

```
ot ok 268 checking setup.sh: setup.sh debug fetchmail
# (in test file test/tests.bats, line 1426)
#   `[ "$status" -eq 11 ]' failed
```

this is the line of the tests https://github.com/tomav/docker-mailserver/blob/f225e14a219824e4ebb34e63b144e6d7c1314bcb/test/tests.bats#L1424-L1426. what happens when you manually run setup.sh with these parameters?

aendeavor:
Seems like you mixed up line 279 with 297. The script fails because with `set -u`, the default value (2048) is not being used - instead the script is stopped with a non-zero exit code. I corrected this with

``` SH
279 dkim     ) _docker_image generate-dkim-config "${2:-2048}" ;;
```

instead of just `$2`.

aendeavor:
Investigated error tracking. There was the idea of showing where an error was, and that there was an error. I tried

``` SH
set -euE
trap '_report_err() $LINENO' ERR

function _report_err()
{
  echo "Error happened on line $1" | tee -a $LOG
}
```

until I realized:

1. `set -E` is not standard POSIX sh and therefore incompatible
2. `trap '...' ERR` is not standard POSIX sh and therefore incompatible
3. the `function` keyword is not POSIX sh too!?

The first two are important (obviously). Unless we switch to a different interpreter (looking at you, `bash`), this feature cannot easily be implemented in `sh`.

erik-wramner:
Well, personally I wouldn't mind using bash. The setup script does run on the host though, so bash may not be available. In practice I haven't seen a Linux host without bash for ages, though. @fbartels what do you think? Upgrade to bash?

aendeavor:
> Well, personally I wouldn't mind using bash. The setup script does run on the host though, so bash may not be available. In practice I haven't seen a Linux host without bash for ages, though. @fbartels what do you think? Upgrade to bash?

In addition to the already mentioned points, `setup.sh` uses `EUID` as an environment variable. This variable is not defined in POSIX `sh` too, making the podman part basically unreliable. I had to default to a value to not trigger `set -u`.

fbartels:
Yes I agree. Almost every host (especially if its also running Docker) will have bash or something that provides the ability to interpret bash. So upgrading this helper script should be no problem.

aendeavor:
Implemented the feature, but still: 268 is failing. Something seems wrong in `setup.sh`.

aendeavor:
Yes, nevermind. My refactoring was incomplete. This should in practice work now. Can you approve @fbartels @erik-wramner ?

fbartels (approved):

aendeavor:
@fbartels I am currently working on a bigger PR obsoleting #1228 and #1137 

Therefore, I would like to know when this PR (#1595) is to be merged. I'd like to become a contributor to manage these PR's. But first, this and #1596 must be merged, as the new PR I'm currently working on is based on them.

Do you have any information? What about @erik-wramner ?

fbartels:
Usually I leave the merging to @erik-wramner, but you changes look good to me so I'll merge them in now.

casperklein:
Image/Container name detection was removed. Was that on purpose?

```
INFO=$($CRI ps \
  --no-trunc \
  --format "{{.Image}};{{.Names}}" \
  --filter label=org.label-schema.name="docker-mailserver" | \
  tail -1)

IMAGE_NAME=${INFO%;*}
CONTAINER_NAME=${INFO#*;}
```

aendeavor:
> Image/Container name detection was removed. Was that on purpose?
> 
> ```
> INFO=$($CRI ps \
>   --no-trunc \
>   --format "{{.Image}};{{.Names}}" \
>   --filter label=org.label-schema.name="docker-mailserver" | \
>   tail -1)
> 
> IMAGE_NAME=${INFO%;*}
> CONTAINER_NAME=${INFO#*;}
> ```

Not sure what you mean @casperklein, but looking at lines 186 ff. in `function main()`, both in this PR as well as on `master`, name detection is there.

casperklein:
You are absolutely right. I don't know, why I missed that. Sorry for the noise.

PR #1596, opened 2020-08-24
Populate relayhost_map from virtual accounts (and other fixes)

mwnx:
I started this branch to enable relaying through virtual account domains (before, only domains found in postfix-accounts.cf and postfix-relaymap.cf were taken into account). I bumped into some bugs along the way so I ended up having to do a bit of fixing and refactoring (see first two commits).

### DIFF
--- a/target/check-for-changes.sh
+++ b/target/check-for-changes.sh
@@ -16,7 +16,6 @@ if [ ! -f postfix-accounts.cf ]; then
 fi
 
 # Verify checksum file exists; must be prepared by start-mailserver.sh
-CHKSUM_FILE=/tmp/docker-mailserver-config-chksum
 if [ ! -f $CHKSUM_FILE ]; then
    echo "${log_date} ${CHKSUM_FILE} is missing! Start script failed? Exit!"
    exit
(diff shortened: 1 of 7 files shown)


erik-wramner:
I think I will need some time to review this (unless you beat me to it @fbartels). One question though. You start by writing to the checksum file without sleep, but that should already have been done in _setup_chksum_file in the startup script which runs before this script is allowed to start. Moving the write to before the sleep just overwrites the file produced by the startup script.

That is unnecessary work, but it can also hide changes. Assume that there were changes after the startup script ran and before the change script created the file. We would not act on them as the checksum file says that nothing has changed.

In short I'd like to remove that first step and use the file created by the startup script to eliminate the risk for race conditions.

mwnx:
On Wed, Aug 26, 2020 at 05:10:18AM -0700, Erik Wramner wrote:
> I think I will need some time to review this (unless you beat me to it @fbartels). One question though. You start by writing to the checksum file without sleep, but that should already have been done in _setup_chksum_file in the startup script which runs before this script is allowed to start. Moving the write to before the sleep just overwrites the file produced by the startup script.

Oh, I had not noticed the '_setup_chksum_file' step in start-mailserver.sh.

> That is unnecessary work, but it can also hide changes. Assume that there were changes after the startup script ran and before the change script created the file. We would not act on them as the checksum file says that nothing has changed.

> In short I'd like to remove that first step and use the file created by the startup script to eliminate the risk for race conditions.

I agree. Do you mind if I factor out the checksum generation code into
helper_functions.sh while I'm at it?

--
mwnx
GPG: AEC9 554B 07BD F60D 75A3  AF6A 44E8 E4D4 0312 C726
     ipfs.io/ipfs/QmdLc7sFYA3SeAocaFC5AaVJKfruj5ecePoj1iW2rWzgPS

fbartels (approved):
> helper_functions.sh while I'm at it?

yes, sounds like a good idea.

If first glance I could not see anything else, than what was already said about the checksum function. lgtm

mwnx:
I updated the first commit with the changes we talked about, so it should be ready for merging.

aendeavor (approved):
Looks fine.

PR #1601, opened 2020-09-05
Changes from #1599 without `start-mailserver.sh`

aendeavor:
# The First Half

This PR includes the first half of #1599, i.e. everything that is not `start-mailserver.sh`.

## Changes

- outsourced badges-links in `README.md` to clean this section up
- refactored `.travis.yml` with Prettier and added **shellcheck capabilities**
- refactored Makefile to make it verbose where it should be and added a **shellcheck target**
- refactored all scripts in `target/` (except `start-mailserver.sh`)
- added **coding style guidelines** to `CONTRIBUTING.md` so everyone can and must adhere to consistent guidelines
- `start-mailserver.sh` had to be touched due to renaming of functions
- added a table of contents to Markdown documents

## What's new

Bash scripts are now streamlined and follow a certain paradigm where strictness is key.

TravisCI will now check scripts with shellcheck, and throws an error if warnings were found. Guidelines have been written to make coding more streamlined.

## To-Do

- [x] closed #1228

## Current Errors

None. Ready for review and merge.

## Post Scriptum

Thank you very much @tomav for invinting me!

PR #1605, opened 2020-09-10
Complete refactoring for `start-mailserver.sh`

aendeavor:
# The Second Half

This one is rather a blue whale. I completely refactored `start-mailserver.sh`. Therefore, it is now included into `shellcheck`. This was indeed a slow decend into madness.

## Changes

- completely refactored `start-mailserver.sh`
- adjusted Makefile to check this script too

## To Do

- [x] close #1550
- [x] get rid of as many `#shellcheck disable...` comments as possible
- [x] correct the `Makefile` to use `/usr/bin/shellcheck`

## Post Scriptum

@erik-wramner I accidentally merge-commited all commits from #1601; I will squash them next time. I'm sorry for the inconvenience.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,10 +51,10 @@ The development workflow is the following:
 
 ### Bash and Shell
 
-When refactoring, writing or altering Script, that is Shell and Bash scripts, in any way, adhere to these rules:
+When refactoring, writing or altering scripts, that is Shell and Bash scripts, in any way, adhere to these rules:
 
 1. **Adjust your style of coding to the style that is already present**! Even if you do not like it, this is due to consistency. Look up the GNU coding style guide. There was a lot of work involved in making these scripts consistent.
-2. **Use `shellcheck` to check your scripts**! Your contributions are checked by TravisCI with shellcheck.
+2. **Use `shellcheck` to check your scripts**! Your contributions are checked by TravisCI with shellcheck. You can check your scripts like Travis with `make shellcheck`.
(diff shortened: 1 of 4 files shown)


erik-wramner:
I have this on my TODO list, I'll try to find time in the weekend.

aendeavor:
any news on your side @erik-wramner ? @fbartels have you had time?

fbartels (approved):
Sorry, quite a large changeset and therefore not that easy to review. But in the end it looks like almost exclusively formatting changes and no functional changes (at least I cannot remember seeing one of those).

aendeavor:
That's correct @fbartels. In the end, it's only a refactoring:) The only "major" changes are those where I substituted code, for example

``` BASH
array=$(cat $THIS | tr ":" "\n"
```

which is ugly and not robust became 

``` BASH
declare -a ARRAY
IFS=':'
read -r -a ARRAY < <(cat "${THIS}")
```

which is more verbose, but `read` was made for exactly this case, therefore being robust.

There were similar changes here and there. The majority of line changes came from brace substitutions and `""` around variables.

aendeavor:
@erik-wramner I see you have very little time at hand, plus the PR is pretty big. I really understand this and I don't want to force you into reviewing these changes. If it's okay with you, just tell me: @fbartels already approved and the test look good, so there should be no risk involved in merging this if you are okay with not having review this one. What's your take on this?

erik-wramner (approved):

erik-wramner:
I've approved the changes. I'll try to make a release this weekend provided that master is stable (i.e. no serious bugs reported).

PR #1613, opened 2020-09-20
Extract even more tests

martin-schulze-vireso:
Continuation of the work in #1206.

- [x] each test got its own copy of config directory -> no more interference between the tests
- [x] no more container starts from the makefile, containers are moved into the test files
- [x] Investigate why there are no failing tests for the "generate accounts after run" step that has not yet been migrated?
- [x] update to newer bats version to use more modern features (like parallelization, native setup/teardown_file)


aendeavor:
This is great to see @martin-schulze-vireso. I marked this as high priority. I'll have a look at Travis later. Keep on the good work!

martin-schulze-vireso:
Just a quick update: I am still working on this but the going is rough. The quota tests reliable hang when the whole test.bats is executed but if I skip some tests, then they run fine. @aendeavor @erik-wramner any ideas what could cause the quota script to hang indefinitely? I thought about the read but that should never be reached in these tests.

aendeavor:
> Just a quick update: I am still working on this but the going is rough. The quota tests reliable hang when the whole test.bats is executed but if I skip some tests, then they run fine. @aendeavor @erik-wramner any ideas what could cause the quota script to hang indefinitely? I thought about the read but that should never be reached in these tests.

I'll investigate and report back.

martin-schulze-vireso:
@erik-wramner I am currently working detecting during the tests when there are still undetected configuration changes to wait for them to be applied. To ease this, I move the checksum file update after the configuration (see https://github.com/tomav/docker-mailserver/pull/1613/commits/191fe4c920e2d0f079d0c8786e482ba9d649e65d). Do you see any negative consequences?

martin-schulze-vireso:
Some more debugging info: It seems one of the `docker exec` commands gets stuck while returning. It seems to be a problem with the container as I noticed that I could not exit a manual `docker exec -it` session when the container is in this state. Even weirder, the tests stop the container but it is still marked as running in `docker ps`. However, when trying to `docker exec` it again I get 

> OCI runtime exec failed: exec failed: cannot exec a container that has stopped: unknown

I also saw another crash pattern where I'd need to restart docker to be able to start the tests again, because the `docker run` command would fail otherwise.

aendeavor:
@martin-schulze-vireso During my current work I noticed the script hang when issuing the `restrict-access` file. Maybe this is related in some way to this problem of tests hanging, who knows...

aendeavor:
@martin-schulze-vireso I noticed you wrote

> - [x] update to newer bats version to use more modern features (like parallelization, native setup/teardown_file)

but I cannot find any diff which shows the bats-submodule being updated to the current master. How shall we go about that? I can update this right now in this repo if this helps you. In the end, this needs to be updated somehow.

martin-schulze-vireso:
@aendeavor 

> but I cannot find any diff which shows the bats-submodule being updated to the current master.

I am sorry for the confusion. I used these checkboxes to track my progress within the ticket, so there is no change on master yet.

However, we can discuss if we want to do this. I used local changes to improve the test development and will PR them in the bats-core project, once they are polished.

aendeavor:
> However, we can discuss if we want to do this. I used local changes to improve the test development and will PR them in the bats-core project, once they are polished.

Alright. I saw that you had already contributed on the bats project (latest commit). When I realized the checked-out commit in the submodule is more than three years old, I figured updating would be a good idea. But this was just the first idea when I saw this. Would you like to wait until you pushed everything into the bats project, and then I would merge it?

martin-schulze-vireso:
I'd say we wait for the next release, the current change is not in yet and local development can benefit from it: When the tests hang and you abort via CTRL-C, the last command is printed as aborted/failed.

aendeavor:
> I'd say we wait for the next release, the current change is not in yet and local development can benefit from it: When the tests hang and you abort via CTRL-C, the last command is printed as aborted/failed.

Alright, then we'll wait. Call me back when this work is to be picked up again.

martin-schulze-vireso:
@aendeavor I finally had a breakthrough with the problems. The test helper functions had a problem which lead to the call to `mail ...` instead of `docker exec mail ...` which did not progress without input for obvious reasons. In hindsight, I think we need some tests for the test helpers as well, they are growing quite complex.

I am sure there are still problems remaining but I banged my head too long against this wall to see the cracks, so I hope you can help me out with a review.

aendeavor:
Of course, I will review what has been done so far. This seems to be much work, and is greatly appreciated! Because I want to take my time, and don't have any today, I can review this PR tomorrow at the earliest. However I guess this is not a big problem.

I noticed there was an interesting _final line ending_ error in `test/test_helper/common.bash`, stating `Wrong line endings or new final newline`, which I have not seen before, but this should be an easy fix:)

The other one is the `261 checking quota: warn message received when quota exceeded` test, which is known to fail. This is more interesting, and we should investigate in this.

You will hear back from me when I review your changes.

test/default_relay_host.bats
```diff
@@ -2,7 +2,7 @@ load 'test_helper/common'
 
 function setup() {
     docker run -d --name mail_with_default_relay \
-		-v "`pwd`/test/config/relay-hosts":/tmp/docker-mailserver \
+		-v "$(duplicate_config_for_container relay-hosts/)":/tmp/docker-mailserver \
```
aendeavor, line 5: Not sure whether the `/` in the end makes a difference here. Could you double check?

aendeavor:
I already marked a few stylistic issues, which is a good thing, since no other issues occured to me. I would love to see curly braces around the variables. As described in CONTRIBUTING.md, you can use this regex `\$([^{("\\'\/])([a-zA-Z0-9_]*)([^}\/ \t'"\n.\]:(=\\-]*)` on regex101.com to automatically change most of the occurrences.

martin-schulze-vireso:
> Because I want to take my time, and don't have any today, I can review this PR tomorrow at the earliest.

Take your time, I did mit rush this either.

> That might actually be a good idea

Okay, I will see what I can do.

test/default_relay_host.bats
martin-schulze-vireso: I removed the / but since this was a directory, I don't see much reason why this should be problematic

aendeavor (approved):
The only complaints I could find were of stylistic nature. Therefore, I can approve these changes, they look good to me. I have to admit, there was much to review, so I might overlooked something, but you never know.

I will provide a PR for you so you don't have to fix all my complaints :)

aendeavor:
You may have noticed, I opened a PR (martin-schulze-vireso/docker-mailserver#1) so you don't have to work on these changes.

martin-schulze-vireso:
I converted this to draft again until we figured out what is going on with test 244 https://github.com/martin-schulze-vireso/docker-mailserver/pull/1#issuecomment-712166217

aendeavor:
That's a good idea.

martin-schulze-vireso:
Ah, test 108 also failed again now. Adding it to the list.

martin-schulze-vireso:
I tried to reproduce the errors by simulating a slow single CPU system (aka Travis' free tier) via `docker run`'s `--cpus 0.5`.

This triggered many errors due to the emails still being in the queue for scanning. I added a helper that will wait until the mail queue is empty as a replacement for the `sleep 80` that I commented out. I am not very versed with the postfix commands so I simply consumed `mailq`'s output. Let me know if there is a more reliable way to detect that programmatically.

The failure on 244 seems to be similar: /var/run/clamav/clamd.ctl was not yet there when the test emails were being sent. I added another repeat_until_success loop that waits for the file to exist. However, I am not sure if this is masking other issues that should by caught by the 'log is error free' test. 

The error 108 did not occur in my local tests, so it might have been hit with these fixes.

Please review these parts extra carefully @aendeavor .

aendeavor:
I will add a review in a few hours.

martin-schulze-vireso:
I think I broke Travis. The failed build fails exactly the things I fixed in the last commit. It is telling, that the failure command looks like before the fix. The triggering commit is also listed wrongly as https://github.com/tomav/docker-mailserver/commit/1a7f7618b8b9cde06c760cc93ffba918ff3ef4b2

aendeavor:
Travis seems to be getting out of hand for useful testing here. I just looked this up: <https://stackoverflow.com/questions/21053657/how-to-run-travis-ci-locally>

I'll will try this out and see what _real_ CPU cores can do here.

martin-schulze-vireso:
Well, no need to run travis for local execution. The whole point of my series of work is geared towards smaller testable chunks for local execution when developing new features and more deterministic test outcomes.

However, there is one problem: your computer is probably too fast. Travis clocks in at ~1000 seconds for all tests. I did not test on my machine but it is definitely below 10 minutes. Many of the problems we see are timing dependent race conditions which are not reproducible on faster machines.

aendeavor:
> Ah, it looks like travis is not running the commit itself but a tentative merge of the previous(?) commit with master.

I can confirm this. This is what the Docker container (`travisci/ci-ubuntu-2004:packer-minimal-1602658599-606a7440`) does as well:

travis-local

aendeavor:
`87` and `106` broken again

martin-schulze-vireso:
I am more concerned of 277. Where does this additional mail come from?

aendeavor:
I have looked over every change. Are there any particular lines you want to check? Since this is a big diff, it's probably better if you tell me where you think I should be careful.

martin-schulze-vireso:
I was referring specifically to Tests 108 and 244, which should now be 106 and 242

aendeavor (approved):
I looked over it again and it actually seems fine to me. I will request a review from @erik-wramner here too, safety first.

@martin-schulze-vireso if you're ready, tell me how are we going to upgrade the Bats submodule?

martin-schulze-vireso:
@aendeavor I'm afraid that bats-core upstream is not moving fast enough for it to get in here. I would propose to wait for the next release and get that in but I cannot tell how long until then. The project is currently understaffed and looking for maintainers.

erik-wramner (approved):

erik-wramner:
Looks good to me!

aendeavor:
@martin-schulze-vireso tell me when you're ready for merging. (Or merge it yourself, if you can - I don't know whether you're a collaborator :D)

martin-schulze-vireso:
I am ready for merging and just a lowly contributor ;) so go ahead please.

PR #1618, opened 2020-09-23
Add possibility to use a domain as a send only domain

wzrdtales:
This changes enable to exclude domains for local delivery options. This is important to use this module as a send only instance.

I do not have too much time at hand, so I am just leaving that here, there is probably some work to do to pull it in, however I will very probably respond slow, so feel free to copy it over in your own terms, I am using my own image already though.

### DIFF
--- a/target/check-for-changes.sh
+++ b/target/check-for-changes.sh
@@ -222,7 +222,7 @@ s/$/ regexp:\/etc\/postfix\/regexp/
         if [[ -n ${SENDONLY_DOMAINS} ]]
         then
           declare -a SEND_DOMAIN
-          IFS=',' read -r -a SEND_DOMAIN <<< "${SENDONLY_DOMAINS}"
+          IFS=',' ; read -r -a SEND_DOMAIN <<< "${SENDONLY_DOMAINS}"
           unset IFS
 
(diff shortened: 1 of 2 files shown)


aendeavor:
Thank you @wzrdtales for following the contributing guideline with opening an issue and a corresponding PR; we value contributions to this project. I will post this answer here as well as on the PR, so don't get confused.

I think this idea is worth merging, but there are some flaws which will need to be fixed up, by you if you want this in `master`. First of all, for this feature, we will need to see tests to verify that it works properly. Moreover, the documentation needs extension too. But you said:

> I do not have too much time at hand, so I am just leaving that here, there is probably some work to do to pull it in [...]

Hearing you don't have time, and neither have we for this feature request, there is little chance this is going to be merged soon. If you're willing to invest more time in this, we will happily review and merge these changes. Therefore, I will close the PR for now and mark it as frozen. When the time has come for you or anyone else to pick up and finish this work, please rebase from (then) current master and create a new PR. I will close this issue too. You will not need to re-open it in case you create a new PR, because I now know about it. **Feel free to re-open this issue if you changed your mind and want to finish this work sooner than later**.

PS: Have a look at `make shellcheck`. The build failed just because of the shellcheck-tests.

PR #1614, opened 2020-09-20
Getting rid of ELK & Update Compose

aendeavor:
This PR starts the process of removing ELK from the current builds.

## What has already been done?

- [x] The ELK directory has been removed
- [x] Removed ` docker-compose.elk.yml.dist`
- [x] Removed ` docker-compose.filebeat.yml.dist`
- [x] Updated Docker Compose files

## What needs to be done?

- [x] The elk build on Docker Hub will be removed by @tomav  

## Rationale

For ELK, it's clear. Filebeat was removed with release 7, so removed this compose file too.

Docker Compose is currently on stable version `1.26.2` which translates to compose file version `3.8`. I updated to `3.7` and formatted the YAML files properly.

PR #1620, opened 2020-09-24
New Stable Version

aendeavor:
This is a draft I prepared @erik-wramner to merge `master` into `stable` when you're ready.

Before we merge this into stable, I would like to find out why the `/elk` build is still there, seeing that it should have been deleted with #1614.

Furthermore, @erik-wramner I saw that `stable` is 57 behind, but also **2 commits *ahead*** of master. Is this just the default merge commit when merging from \<insert branch name here\>?


erik-wramner:
@aendeavor the reason for stable being 2 commits in front of master is the build environment. As I can't trigger a build from Docker hub the only way to publish the branch if the build fails is to push again to trigger a new build. Irritating as the build fails from time to time due to the unstable tests, but there you are.

PR #1621, opened 2020-09-24
added option to use non-default network-interface, resolves #1227

### LINKED ISSUE #1227: Invalid mynetworks when host interface is not named 'eth0'
All SMTP mail gets rejected due to an invalid mynetworks configuration when the host interface is not named 'eth0'. 

In this function, the primary interface of the host is assumed to be eth0, but it's not always named as such, and so an invalid mynetworks line gets written as ```mynetworks = 127.0.0.0/8 [::1]/128 [fe80::]/64 /32``` (the /32 is invalid)

https://github.com/tomav/docker-mailserver/blob/9ac164ab648df166b5fd6d6a944af9d137ace290/target/start-mailserver.sh#L206

aendeavor:
This PR adds the possibility to use a network interface of choice. When using Kubernetes with Host Networking, this was an issue as `eth0` was the default and not changeable.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,77 @@
+
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our community a harassment-free experience for everyone, regardless of age, body size, visible or invisible disability, ethnicity, sex characteristics, gender identity and expression, level of experience, education, socio-economic status, nationality, personal appearance, race, religion, or sexual identity and orientation.
+
+We pledge to act and interact in ways that contribute to an open, welcoming, diverse, inclusive, and healthy community.
+
+## Our Standards
+
+Examples of behavior that contributes to a positive environment for our community include:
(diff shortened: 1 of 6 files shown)


fbartels (approved):

erik-wramner:
The PR includes changes not related to the issue, but it's fine.

erik-wramner (approved):

aendeavor:
> The PR includes changes not related to the issue, but it's fine.

Although minor, that's true, and bad style - pardon me:) Should not happen again, but since the test were failing two times in a row, I thought it what I could improve on.

sundbry:
I don't see how sliding in the contributor covenant was an appropriate element to "accidenrally" include here.

aendeavor:
This was by no means on accident. Although I agree with you @sundbry that this is not the best style, I don't feel like this is harmful in any way to this project. Meeting [GitHub's Community Standards here](https://github.com/tomav/docker-mailserver/community) is a good thing.

Nevertheless, I will resolve and revert these changes, if you present good reasons for it.

PR #1626, opened 2020-09-26
Dovecot community repo info updated

casperklein:


aendeavor:
I was curious: Could we just get rid of the comments you moved to the top? I had the same idea, you've done it @casperklein. Maybe we can even drop the comments.

erik-wramner:
I think we should get rid of that part entirely. It was kept for convenience before we decided which way to go.

aendeavor:
> I think we should get rid of that part entirely. It was kept for convenience before we decided which way to go.

Completely with you there.

casperklein:
For reference: It was commented out, while preparing the Buster release, because there are no ARM packages available: https://github.com/tomav/docker-mailserver/issues/1195#issuecomment-599049180

I liked it commented out, so everyone building an amd64 only image and wanting to use the more recent dovecot version, just had to uncomment it.
However, the old comment was broken, because it was still referring to Debian Stretch. I fixed that with this PR and moved it to the top.

Feel free to keep the comment, for easy activation, for those who want it or completely drop it. I am fine with both ways.

aendeavor:
I understand the reason behind leaving this as is. Nevertheless, I'd "vote" for removing it, because I feel like the cleaner look of the file is worth more than the option for some to get the latest Dovecot. But this is just my POV.

PR #1629, opened 2020-09-30
WIP / FEAT: Dual certificate support (eg ECDSA with RSA fallback)

polarathene:
Since Postfix 3.4 (Feb 2019), `smtpd_tls_cert_file` and `smtpd_tls_key_file` have been deprecated in favor of [`smtpd_tls_chain_files`](http://www.postfix.org/postconf.5.html#smtp_tls_chain_files) which supports a list of values where a single or sequence of file paths provide a private key followed by it's certificate chain. 

## 1st commit - DRY logic, subtle changes/improvements, small bugfix

The first commit refactors the relevant code to support this settings change in Postfix, while also moving common logic for each TLS setup variation into a function. 

`sed` lines were also revised as they often contained unnecessary options(`-e`, `-r`) and regex tokens/syntax(`g`), or specific paths that were better extracted out as variables or unnecessary for a match(use `.*` instead).

The `self-signed` case had an additional `-combined.pem` file used instead of the `-cert.pem` for Dovecot's `ssl_cert` value. This didn't make much sense (unless there's a known reason for doing this), the `-combined.pem` is a bundle file containing the private key followed by the certificate, I see no reason to support that so I've replaced it with the same `-cert.pem` that Postfix uses for it's certificate. This makes it more consistent with other cases.

`self-signed` Dovecot paths to apply `sed` to were also **invalid** due to whitespace within the file paths AFAIK.

The `custom` case also differed, in that it provides a similar key+cert bundle as the self-signed `-combined.pem`, this one is `-full.pem` which I've not inspected but assume contains a full certificate chain (leaf + issuer chain), which is useful for including CA certs in a chain that don't belong to a systems trust store.

An additional bonus of this refactor is that previously a `docker-compose up` when exited, persisted the container's internal filesystem state which would be used by a follow up `docker-compose up`, which caused unexpected behaviour / breakage in the event that you changed from `self-signed` to `manual` for example or if `HOSTNAME` were to change. `docker-compose down` was required to properly clear this state to a "fresh" idempotent state. **That issue may be present elsewhere in the codebase btw**.

Postfix `main.cf` also had a slight consistency fix to maintain spaces between `=` assignment like the rest of the config file and related configs such as Dovecot use.

I've not tested each case personally, only the `manual` case. I'm not deeply familiar with the codebase or the `bats` tests, this PR has been validated to work for `manual`, it might benefit from an extra `bats` test covering the dual cert support?

## 2nd commit - Dual cert feature

`smtpd_tls_chain_files` allows for multiple key+cert bundles so that you can provide different key types, such as ECDSA and RSA.

To maintain compatibility with the current `SSL_CERT_PATH` and `SSL_KEY_PATH` ENV variables only a 2nd certificate is supported.

Since Dovecot 2.2.31 (June 2017) [a related feature is also available](https://wiki.dovecot.org/SSL/DovecotConfiguration#Multiple_SSL_certificates), but it is limited to only providing one alternative certificate via separate cert and key settings. That limitation is a non-issue as this feature is only supporting a 2nd alternative certificate via `SSL_ALT_CERT_PATH` and `SSL_ALT_KEY_PATH` ENV vars.

---

## TL;DR

This feature enables support for multiple certificates, eg for serving modern ECDSA cert with an RSA cert as fallback.

---

I needed this for testing my active PR. (which I'm already using locally, but figured was worth polishing up for a separate PR) 

I presently do not have time to learn about `bats` and how to implement any tests required for this PR, any assistance would be appreciated regarding that, even if it's just some guidance on what needs to be tested and roughly how to go about achieving that.

### DIFF
--- a/target/dovecot/10-ssl.conf
+++ b/target/dovecot/10-ssl.conf
@@ -11,6 +11,8 @@
 # certificate, just make sure to update the domains in dovecot-openssl.cnf
 ssl_cert = </etc/dovecot/ssl/dovecot.pem
 ssl_key = </etc/dovecot/ssl/dovecot.key
+#ssl_alt_cert = </path/to/alternative/cert.pem
+#ssl_alt_key = </path/to/alternative/key.pem
 
 # If key file is password protected, give the password here. Alternatively
 # give it when starting dovecot with -p parameter. Since this file is often
--- a/target/postfix/main.cf
+++ b/target/postfix/main.cf
@@ -18,10 +18,9 @@ inet_interfaces = all
 inet_protocols = all
 
 # TLS parameters
-smtpd_tls_cert_file=/etc/ssl/certs/ssl-cert-snakeoil.pem
-smtpd_tls_key_file=/etc/ssl/private/ssl-cert-snakeoil.key
-#smtpd_tls_CAfile=
-#smtp_tls_CAfile=
+smtpd_tls_chain_files = /etc/ssl/certs/ssl-fullkeychain-snakeoil.pem
+#smtpd_tls_CAfile =
+#smtp_tls_CAfile =
 smtpd_tls_security_level = may
 smtpd_tls_loglevel = 1
 smtp_tls_security_level = may
--- a/target/start-mailserver.sh
+++ b/target/start-mailserver.sh
@@ -1061,6 +1061,24 @@ function _setup_ssl()
 {
   _notify 'task' 'Setting up SSL'
 
+  # Primary certificate to serve for TLS
+  # sed substituion delimiter uses `~` instead of `/` due to file paths as values
+  function _set_certificate()
+  {
+    local postfix_fullkeychain=$1
+    local dovecot_cert=$2
+    local dovecot_key=$3
+
+    # Postfix configuration
+    # NOTE: `smtpd_tls_chain_files` expects private key defined before public cert chain
+    # May be a single PEM file or a sequence of files, so long as the order is key->leaf->chain
+    sed -i 's~^smtpd_tls_chain_files =.*~smtpd_tls_chain_files = '"${postfix_fullkeychain}~" /etc/postfix/main.cf
+
+    # Dovecot configuration
+    sed -i 's~^ssl_cert = <.*~ssl_cert = <'"${dovecot_cert}~" /etc/dovecot/conf.d/10-ssl.conf
+    sed -i 's~^ssl_key = <.*~ssl_key = <'"${dovecot_key}~" /etc/dovecot/conf.d/10-ssl.conf
+  }
+
   # TLS strength/level configuration
   case "${TLS_LEVEL}" in
     "modern" )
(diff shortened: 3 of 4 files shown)


aendeavor:
Looks very nice. Keep on this good work.

A word about the current situation with the test: They are being refactored and rewritten in #1613 as this is absolutely necessary. Therefore, I cannot recommend writing any more tests before this PR has been merged.

polarathene:
Will that delay the merging of this feature? If so no worries :)

I'm not particularly great at naming things, I've used `fullkeychain` a few times to describe a file with private key + certificate chain(leaf + issuing certificates from CA) - I've seen this called `-full.pem`, `-combined.pem`, `fullchain.pem` (this one seemed to only be the certificate chain I think without a private key), no clue if there is a proper jargon word for referring to it?

`cert_chain` can just be the leaf certificate, or may contain multiple additional certificates such as intermediate and root certs from a CA.

aendeavor:
> Will that delay the merging of this feature? If so no worries :)

Not if you ask me :)

> I'm not particularly great at naming things, I've used `fullkeychain` a few times to describe a file with private key + certificate chain(leaf + issuing certificates from CA) - I've seen this called `-full.pem`, `-combined.pem`, `fullchain.pem` (this one seemed to only be the certificate chain I think without a private key), no clue if there is a proper jargon word for referring to it?

I guess it's fine. But have a look at polarathene#1 :D

martin-schulze-vireso:
> They are being refactored and rewritten in #1613 as this is absolutely necessary.

The current situation is centered mostly around getting all tests with their setup into .bats files instead of the makefile. Changing existing .bats files is absolutely fine with this PR. Changing tests.bats might light to some conflicts but I think if it is only addition of new test cases, they should be easy to resolve. In summary: No need to compromise the quality by leaving out tests. :)

aendeavor:
Thanks for clarifying @martin-schulze-vireso. If this is the case, a test or two won't be of harm - they should be added:)

polarathene:
I welcome any help with sorting out tests for this then, but I'll try take a shot at it if no one can spare the time to.

For now my spare time is allocated to the security update PR, slowly piecing together my review.

aendeavor:
I'm currently working on another k8s deployment, I'm sorry. Take your time.

polarathene:
Just occurred to me that there should probably be some logic to "reset"/disable the alternative config if the ENV var is removed. It'd only affect Dovecot AFAIK, but should be reproducible with:

- `docker-compose up`
- `CTRL+C`
- Remove ENV var
- `docker-compose up`
- `docker-compose exec <service name> bash`
- The dovecot config for alternative cert/key would not be commented/disabled, and the referenced file that was copied from docker volume or bind mount would probably still exist until `docker-compose down`.

aendeavor:
@polarathene This is just FYI: I was going through your commits when I recognized that they are not signed. If you'd like to improve on this, you can have a look here <https://docs.github.com/en/free-pro-team@latest/github/authenticating-to-github/signing-commits> (improving on authenticity). This is just a recommendation from my side:)

aendeavor:
Because lately we've been seeing a lot of PRs, I will mark this as a draft too. Just mark it ready for review when it's done.

aendeavor:

Issue #1446, opened 2020-04-05
Amavis config file permissions not secure, keeps looping


mattarnster:
When spawning a new docker-mailserver, amavis complains that its config file is writable and then exit-loops.

## Context
I spawned a fresh copy of this docker container using the instructions in the readme.

Here's the log from the container:
```mail    | Apr  5 01:42:49 <host> amavis[167]: starting. /usr/sbin/amavisd-new at <domain> amavisd-new-2.10.1 (20141025), Unicode aware
mail    | Apr  5 01:42:49 <host> amavis[167]: Net::Server: Group Not Defined.  Defaulting to EGID '111 111'
mail    | Apr  5 01:42:49 <host> amavis[167]: Net::Server: User Not Defined.  Defaulting to EUID '109'
mail    | Apr  5 01:42:49 <host> amavis[167]: (!!)FATAL: Config file "/etc/amavis/conf.d/61-dms_auto_generated" is writable, UID 109, EUID 109, EGID 111 111
mail    | Apr  5 01:42:49 <host> amavis[167]: (!!)TROUBLE in pre_loop_hook: SECURITY PROBLEM, ABORTING at /usr/sbin/amavisd-new line 12567.
mail    | 2020-04-05 01:42:49,623 INFO exited: amavis (exit status 255; not expected)
```

Not sure this is having an effect on me either way since spamassassin is disabled. 

## Expected Behavior
This shouldn't happen from a fresh `docker-compose up` command.

## Actual Behavior
See above for the log.

## Steps to Reproduce
Not sure how to reproduce, as this happened right from the start for me.

## Your Environment

* Amount of RAM available: 12GB
* Mailserver version used: tvial/docker-mailserver:latest
* Docker version used: Docker version 1.13.1, build cccb291/1.13.1
* Environment settings relevant to the config:
```
HOSTNAME=<host>
DOMAINNAME=<domain>
CONTAINER_NAME=mail
```

* Any relevant stack traces ("Full trace" preferred):
As log file above, copied below:
```mail    | Apr  5 01:42:49 <host> amavis[167]: starting. /usr/sbin/amavisd-new at <domain> amavisd-new-2.10.1 (20141025), Unicode aware
mail    | Apr  5 01:42:49 <host> amavis[167]: Net::Server: Group Not Defined.  Defaulting to EGID '111 111'
mail    | Apr  5 01:42:49 <host> amavis[167]: Net::Server: User Not Defined.  Defaulting to EUID '109'
[...]

PR #1632, opened 2020-10-01
Remove write access to group and others to Amavis configuration files

artonge:
Remove write access to group and others to Amavis configuration files

Fix: #1446

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -168,7 +168,8 @@ RUN sed -i -r 's/#(@|   \\%)bypass/\1bypass/g' /etc/amavis/conf.d/15-content_fil
   # no syslog user in debian compared to ubuntu
   adduser --system syslog && \
   useradd -u 5000 -d /home/docker -s /bin/bash -p "$(echo docker | openssl passwd -1 -stdin)" docker && \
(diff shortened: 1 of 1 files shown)


aendeavor (approved):

aendeavor:
I will merge this, but I want to hear @erik-wramner opinion in #1446 first.

erik-wramner:
The chmod affects not only the copied files but any other files that happen to be in the folder as well. I would suggest another approach, see https://github.com/moby/moby/pull/34263. It is now possible to set the permissions and owner in the COPY command in the Dockerfile. If we do that we save one layer in the final image and we don't risk changing existing files in a negative way.

@artonge, can you try that instead, please?

aendeavor (changes requested):
> It is now possible to set the permissions and owner in the COPY command in the Dockerfile. If we do that we save one layer in the final image and we don't risk changing existing files in a negative way.

Exactly why I wanted to hear @erik-wramner on this. Thanks :D

erik-wramner:
Hm, chmod will be supported (https://github.com/moby/moby/issues/34819) but is not there yet. I'll merge this then.

PR #1636, opened 2020-10-02
Outsourced linting tests into its own file

aendeavor:
Since @martin-schulze-vireso started refactoring the tests in #1613, effectively cleaning up the Makefile on the way, I figured it would make sense to be consistent here and outsource the linting tests into their own file.


erik-wramner:
Looks good except that there are test failures. I don't think these are the normal ones due to timing either.

aendeavor:
This was an issue on my side. Corrected one thing, broke another. This is an easy fix though, since it's only logging output.

PR #1638, opened 2020-10-04
Comments updated

casperklein:
Comment added about modifying the supervisord configuration. See https://github.com/tomav/docker-mailserver/issues/1608


aendeavor:
weird quota test fails. Can you re-push to re-trigger the build?

PR #1641, opened 2020-10-06
Miscellaneous cleanup / housekeeping

aendeavor:
This PR is just housekeeping and cleanup here and there, nothing in particular, but it bugged me to see these small areas where I could improve on.


casperklein:
One question regarding: https://github.com/tomav/docker-mailserver/pull/1641/commits/060e2c5427ae3e640bf03a7a18c99a15bb0518c5

>: >/etc/postfix/vmailbox

I normaly use `> /foo/bar` to create an empty file. What is the colon for in this case?

aendeavor:
> One question regarding: [060e2c5](https://github.com/tomav/docker-mailserver/commit/060e2c5427ae3e640bf03a7a18c99a15bb0518c5)
> 
> > : >/etc/postfix/vmailbox
> 
> I normaly use `> /foo/bar` to create an empty file. What is the colon for in this case?

Using `> /foo/bar` is actually fine because it's clear that you're redirecting an empty `stdout` in that file essentially creating an empty one. But it is good style in idiomatic Bash seeing a command before the redirection. The colon (and in this case `true` as well) would serve this no-op function:

noop

However this will not actually affect the outcome.

casperklein:
Ok, thank you for clarification.

aendeavor:
No problem:)

PR #1653, opened 2020-10-16
Documentation and Script Updates trying to fix #1647

aendeavor:
I accidentally removed a branch and closed #1651. This is the same PR (without the unwanted commits for another change in another PR).

PR #1654, opened 2020-10-16
Complete Refactor for `target/bin`

aendeavor:
During small work on something else, I discovered that scripts under `target/bin/` are not checked since (obviously) they do not end with `.sh`. When I looked at them, and please forgive m choice of words here, they were as horrendous as the other scripts I tended to before refactoring.

I figured it would be best to edit them as well.

Rationale:

1. Currently, they are **a mess**
    1. Sometimes using `/bin/bash`, sometimes `/usr/bin/env sh`
    2. The constructs are all over the place, very few times actually being **idomatic Bash**
    3. They use **unreliable** and **unstable** splitting methods (e.g. ` | sed -i 's/,/ /g'`)
    4. They are non-coherent, and do **not adhere to our** styling guide and **standards**
    5. They use deprecated methods of evaluation (e.g. `expr index ...`)
2. Refactoring them makes working with them in the future easier
    1. They have become clear in structure
    2. Although working right now, this is however not a guarantee since (as I mentioned), some of the used methods are unreliable and unstable. Therefore, making a **stronger** guarantee with idiomatic scripts is obviously better.

**Note**: This PR accidentally includes #1653. Therefore, this will be merged when #1653 was merged.


target/bin/addsaslpassword
```diff
@@ -2,34 +2,29 @@
...
-
-if [ -z "$PASSWD" ]; then
-	read -s -p "Enter Password: " PASSWD
-	echo
-	[ -z "$PASSWD" ] && errex "Password must not be empty"
+DOMAIN="${1}"
+USER="${2}"
+PASSWD="${3}"
+
+function usage  { echo "Usage: addsaslpassword <domain> <username> <password>" ; }
```
fbartels, line 9: This part repeats itself in the other scripts, would be a candidate to move it into a shared file and source it from there.

target/bin/delmailuser
```diff
@@ -1,69 +1,90 @@
 #! /bin/bash
+# shellcheck disable=SC2094
```
fbartels, line 2: would move the ignore line to the part that should actually be ignored

target/bin/setup-fetchmail
```diff
@@ -1,13 +1,14 @@
-#! /bin/sh
+#! /bin/bash
```
fbartels, line 1: Strictly speaking it is still not the identical shebang in all scripts. some have a space after the pound.

fbartels (approved):

target/bin/delmailuser
aendeavor: This is done to disable this linting target for the whole file. Otherwise I would need to write it twice (and at one place where it somehow doesn't really work, Bash is complaining somehow, idk why). What do you think?

target/bin/setup-fetchmail
aendeavor: Fixed in the latest commit.

target/bin/addsaslpassword
aendeavor: I have outsourced what I thought appropriate into a new file and sourced this file from every other file which needs the mentioned functions. This is done in the latest commit.

target/bin/delmailuser
erik-wramner: Regardless of where it is I would like to have a short comment that says in plain text why it is needed.

erik-wramner (approved):

target/bin/delmailuser
aendeavor: Good idea. I have added it with the latest commit.

PR #1660, opened 2020-10-18
Exchanging `.dist` files

aendeavor:
A while ago, I figured the `*.dist` files existed for the ELK stack and other use cases, all of which were removed. Therefore, I deemed a quality of life improvement worthwhile.

I assingned you here @erik-wramner, because although I searched the whole repository for remains of `*.dist`-file-usage, I couldn't find any, but that doesn't mean there isn't a reason these files exists. If there is anything a overlooked, please tell me.

I cannot assign you @martin-schulze-vireso and @casperklein, but would like to hear your opinion too:)


.gitignore
```diff
@@ -3,10 +3,7 @@
 #################################################
 
 .DS_Store
-.env
```
fbartels, line 6: I would keep the file ignored. This file is often used to store secrets and no longer ignoring it could mean that these could accidentally be committed and pushed to a public repository.

If you still want to include an initial copy of it with the checkout you could force commit it once.

fbartels (approved):

.gitignore
aendeavor: Fixed in the latest commit.

casperklein (approved):

erik-wramner (approved):
This is fine, but do we need the compose upgrade? Some people run old versions of compose as their distros are conservative (that is the nice word for it), so having if we don't use the newer features the old version may work? It is just a template file though, so I'm fine with it either way.

aendeavor:
I feel like for our template, we can afford using the latest version. People who run old versions have this file already written with the correct version. This shouldn't be an issue.

PR #1662, opened 2020-10-21
Env Update 2

aendeavor:
I just figured, why do we actually name the environment files so poorly, just so we can later explain the difference? I think, this change makes it much clearer where the environment variables are expanded, in addition to the note that is being kept, of course.

Furthermore, this provides **proper** syntax highlighting in IDEs.


README.md
```diff
@@ -88,28 +88,28 @@ Minimum:
...
+- Edit the files `.env` and `mailserver.env` to your liking:
   - `.env` contains the configuration for docker-compose
-  - `env-mailserver` contains the configuration for the mailserver container
+  - `mailserver.env` contains the configuration for the mailserver container
   - These files supports only simple `VAR=VAL` lines (see [Documentation](https://docs.docker.com/compose/env-file/)).
   - Don't quote your values.
   - Variable substitution is *not* supported (e.g. `OVERRIDE_HOSTNAME=$HOSTNAME.$DOMAINNAME`).
 
-**Note:**: Variables in `.env` are expanded in the `docker-compose.yml` file **only** and **not** in the container. The file `env-mailserver` serves this case where environment variables are used in the container.
+**Note:**: Variables in `.env` are expanded in the `docker-compose.yml` file **only** and **not** in the container. The file `mailserver.env` serves this case where environment variables are used in the container.
```
fbartels, line 112: with .env provided via env_file the statement regarding it is not technically true.

I think .env should be removed from docker-compose env_files instead if rephrasing this sentence.

fbartels (approved):

README.md
aendeavor: I should have realised this earlier. It's fixed now.

aendeavor:
The test error should in theory be harmless, since

1. These changes did not change one thing which could alter test results
2. It seems like Travis messed this up

casperklein:
The variables defined in `.env` are automatically read by docker-compose and used. See [here](https://docs.docker.com/compose/environment-variables/#the-env-file).

If that file is renamed (to compose.env), these variables are not be available for use in docker-compose:

    WARNING: The HOSTNAME variable is not set. Defaulting to a blank string.
    WARNING: The DOMAINNAME variable is not set. Defaulting to a blank string.
    WARNING: The CONTAINER_NAME variable is not set. Defaulting to a blank string.

The `env_file` section in docker-compose is for variables, which should be available within the container and are not used by docker-compose.

aendeavor:
@casperklein I tried my best to understand what you're up to, but my guess right now is that you misunderstood the changes this and the previous PR introduced. Let me explain:

---

> The variables defined in .env are automatically read by docker-compose and used.

This was already perfectly clear. A few weeks ago I introduced a change to the README, which states:

``` MARKDOWN
**Note**: Variables in `.env` are expanded in the `docker-compose.yml` file only and not in the container. The file `env-mailserver` serves this case where environment variables are used in the container.
```

Because I had already dealt with some issues regarding this I figured I might as well write it in the README.

> If that file is renamed (to compose.env), these variables will not be available for use in docker-compose

This is technically speaking true, but you could have made this argument earlier, since originally, the `compose.env` file was called `.env.dist`. What you missed is the update of the README, which uses the following commands:

``` BASH
...
curl -o .env https://raw.githubusercontent.com/tomav/docker-mailserver/master/compose.env
...
```

So to actually use it, as it has always been the case, you `curl` the file and name it `.env`.

> The env_file section in docker-compose is for variables, which should be available within the container and are not used by docker-compose.

I think this is now pretty clear, and that's why I named it `mailserver.env`:)

---

To summarize: The changes introduced here do perfectly well what you describe, but with meaningful filenames.

casperklein:
> **Note**: Variables in `.env` are expanded in the `docker-compose.yml` file only and not in the container.

As of now, that is not true. The `.env` file is referenced in `docker-compose.yml` in the `env_file section`. Therefor all variables set in `.env` are also available in the container and not only used by docker-compose itself. Put something like foo=bar in `.env` and start the container. You'll see, that `echo $foo` will print `bar` in the container.

>Because I had already dealt with some issues

I think I know which issues you mean. If I remember correctly, imho the "real" problem was that variables definied in .env were later overwritten with default values from `env-mailserver` (which is referenced second in the docker-compose env_file section).
That was the root problem I think. And not the assumption, that `.env` variables are not passed to the container at all.

>What you missed is the update of the README, which uses the following commands:

Full ack. I missed that. You can ignore my previous comment 😄

aendeavor:
> > **Note**: Variables in `.env` are expanded in the `docker-compose.yml` file only and not in the container.
> 
> As of now, that is not true. The `.env` file is referenced in `docker-compose.yml` in the `env_file section`.

This is one of the reasons this PR exists:)
 
> I think I know which issues you mean. If I remember correctly, imho the "real" problem was that variables definied in .env were later overwritten with default values from `env-mailserver` (which is referenced second in the docker-compose env_file section).
> That was the root problem I think. And not the assumption, that `.env` variables are not passed to the container at all.

I think that was it.

PR #1667, opened 2020-10-25
Change default logrotate settings for /var/mail/maillog

casperklein:
See also: https://github.com/tomav/docker-mailserver/issues/1666

When Logrotate interval:

- daily => rotate 7 times (1 week)
- weekly => rotate 5 times (1 month)
- monthly => rotate 12 times (1 year)


aendeavor:
Have you checked why pflogsum delivery is failing?

casperklein:
It fails because of: https://github.com/tomav/docker-mailserver/blob/451bbfdf40a47d20f9cde6a4b2ce309bf3c36cbe/test/mail_with_ldap.bats#L223

If I set this to _weekly_ the test succeed. But I start wondering, if for some reason, daily rotation could be needed? At the moment, I do not understand the motivation for this test.

aendeavor:
I think this test is just there to ensure a correct Setup. We may als @erik-wramner here.

aendeavor (approved):

erik-wramner:
It is there to verify that the expected default works. However, I don't really like this PR (sorry).

I can't see the benefit in changing the default interval. If you want another interval, then set the option. Changing the default "breaks" existing installations that use the default. That is fine if there is a clear benefit, but here it is more a matter of taste. I don't think we should do that.

Second, why do you change the number of retained log files and (more specifically) why do you want to keep 7 days, 5 weeks and 12 months respectively? I don't quite see the logic and it might affect the reports. I'm not sure (would have to check) but I think they check all the archived log files. Going through a whole year for the summary report will take time...

casperklein:
@aendeavor ready for merge.

PR #1672, opened 2020-10-30
Add SELinux support

### LINKED ISSUE #1637: SELinux Fix?
I have an issue with this line when SELinux is on.

https://github.com/tomav/docker-mailserver/blob/108bdbd46361b23671b11940cf5827a98f6975b0/setup.sh#L169

Editing the script and adding `:Z` fixes the issue.
`-v "${CONFIG_PATH}":/tmp/docker-mailserver:Z \ `

I would do a PR but unsure of the best way to detect SELinux running.

omarc1492:
Added SELinux support to docker-compose and the script `setup.sh`. Closes #1637


aendeavor:
This is waiting for review. @casperklein what's your take on this? @erik-wramner any opinions on this?

casperklein (approved):
Looks good to me. All additions are opt-in. So there are no drawbacks for existing setups.

erik-wramner:
I agree, looks good and should not affect existing users. I'll merge it.

PR #1690, opened 2020-11-16
Add additional pem file for _monitored_files_checksums

dayld:
Hi,

[start-mailserver.sh](https://github.com/tomav/docker-mailserver/blob/master/target/start-mailserver.sh#L1113
) checks if private key "privkey.pem" or "key.pem"
but [helper-functions.sh](https://github.com/tomav/docker-mailserver/blob/master/target/helper-functions.sh#L204) monitors only key.pem 
My certbot generates only "privkey.pem" thats why, there could be issues while updating private key.


casperklein (approved):

aendeavor:
This will be merged into master after `v7.2.0-stable` is released. Approximately 5 days from now.

aendeavor (approved):

PR #1703, opened 2020-12-06
Small fixes for extracting certs from the acme.json file

BrandonSchmitt:
As I wanted to setup TLS with certificates provided by traefik I ran into two small bugs:

The first was a python exception. I have an acme.json containing sections which look like:
```
  "httpChallenge": {
    "Account": null,
    "Certificates": null
  }
```
The script didn't test if `Certificates` was `null`.

The second problem was that the environment variable `SSL_DOMAIN` was only used while watching for changes in the acme.json, not during start-up.

Both bugs are fixed with this PR.

### DIFF
--- a/target/helper-functions.sh
+++ b/target/helper-functions.sh
@@ -56,11 +56,12 @@ import sys,json
 acme = json.load(sys.stdin)
 for key, value in acme.items():
     certs = value['Certificates']
-    for cert in certs:
-        if 'domain' in cert and 'key' in cert:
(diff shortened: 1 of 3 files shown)


aendeavor:
Thank you for providing these fixes @BrandonSchmitt. I will happily merge this as soon as possible, but before I approve of these changes, I would like to take a look at the `[[ ... ]] && ... || ...` construct that was introduced. I left a comment in my review:)

aendeavor:
EDIT: Sorry for the inconvenience, forgot to press the submit button for the review (here too).

aendeavor (approved):

Issue #1697, opened 2020-11-29
Periodical maintenance operation for Dovecot mailbox format


gmasse:
## Is your feature request related to a problem?
Dovecot [dbox format](https://github.com/tomav/docker-mailserver#dovecot_mailbox_format) requires periodical purge to free unused space. Details in the official documentation:
> Expunging a message only decreases the message's refcount. The space is later freed in "purge" step. This is typically done in a nightly cronjob when there's less disk I/O activity. The purging first finds all files that have refcount=0 mails. Then it goes through each file and copies the refcount>0 mails to other mdbox files (to the same files as where newly saved messages would also go), updates the map index and finally deletes the original file. So there is never any overwriting or file truncation.
> 
> The purging can be invoked explicitly running doveadm purge.

## Describe the solution you'd like
Daily crontab within the container:
```
$ cat /etc/cron.d/dovecot-purge
#
#  Dovecot purge maintenance
#
# m h dom mon dow user	command
38 2	* * *	root    doveadm purge -A
```
**Question:** should this file be added automatically when enabling dbox format?
(DOVECOT_MAILBOX_FORMAT=sdbox or mdbox)

## Alternatives you've considered
Run from the host's crontab (not in the container).

## What are you going to contribute?
I can implement the solution.

aendeavor:
Thanks for this post. As the dbox format is marked experimental, I think it is fine to add this Cron entry unconditionally. Therefore, from my point of view, the answer to

>  should this file be added automatically when enabling dbox format?

is yes.

You're very welcome with providing and implementing a solution. I will review it when the PR is ready.

PS: I dislike the idea of running Cron from the host and not inside the container. We should not consider this alternative. Nevertheless worth the mention so we can rule this out.

erik-wramner:
@gmasse please make sure that you delete the cron job if dbox format is not used as well (if it is present of course).

aendeavor:
This somehow wasn't linked. Closed with #1705.

PR #1705, opened 2020-12-07
Add purge cron job for Dovecot dbox format

gmasse:
Related to feature request #1697

### DIFF
--- a/target/start-mailserver.sh
+++ b/target/start-mailserver.sh
@@ -543,9 +543,13 @@ function _setup_dovecot
 
   # set mail_location according to mailbox format
   case "${DOVECOT_MAILBOX_FORMAT}" in
-    sdbox|mdbox|maildir )
+    sdbox|mdbox )
       _notify 'inf' "Dovecot ${DOVECOT_MAILBOX_FORMAT} format configured"
       sed -i -e 's/^mail_location = .*$/mail_location = '"${DOVECOT_MAILBOX_FORMAT}"':\/var\/mail\/%d\/%n/g' /etc/dovecot/conf.d/10-mail.conf
+
+      _notify 'inf' "Enabling cron job for dbox purge"
+      mv /etc/cron.d/dovecot-purge.disabled /etc/cron.d/dovecot-purge
+      chmod 644 /etc/cron.d/dovecot-purge
       ;;
     * )
       _notify 'inf' "Dovecot maildir format configured (default)"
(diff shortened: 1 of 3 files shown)


target/start-mailserver.sh
```diff
@@ -543,13 +543,25 @@ function _setup_dovecot
 
   # set mail_location according to mailbox format
   case "${DOVECOT_MAILBOX_FORMAT}" in
-    sdbox|mdbox|maildir )
+    sdbox|mdbox )
```
erik-wramner, line 546: Why has maildir been removed here? The cron job is not needed for that, but how about the other changes?

target/start-mailserver.sh
```diff
@@ -543,13 +543,25 @@ function _setup_dovecot
...
       _notify 'inf' "Dovecot ${DOVECOT_MAILBOX_FORMAT} format configured"
       sed -i -e 's/^mail_location = .*$/mail_location = '"${DOVECOT_MAILBOX_FORMAT}"':\/var\/mail\/%d\/%n/g' /etc/dovecot/conf.d/10-mail.conf
+
+      _notify 'inf' "Creating cron job for dbox purge"
+      cat <<EOF >/etc/cron.d/dovecot-purge
+#
```
erik-wramner, line 552: This works, but personally I would prefer to keep the file separate and just make a copy here. That keeps the start script a bit simpler and makes it easier to change the cronjob.

erik-wramner:
The unit tests failed on quota. Not sure why, but it is not one of the common failures. Please push again to re-trigger the tests so that we can see if it was coincidence or an error. Perhaps related to the maildir change?

target/start-mailserver.sh
gmasse: maildir is the default behavior

target/start-mailserver.sh
gmasse: Where is the best place to put a crontab file? It could be in `target/dovecot/`.

aendeavor:
@erik-wramner seems like to quota test still is somewhat random, yes - I had hoped to not see this again. But at the current failure rate, this randomness is somewhat acceptable and way better than a few weeks ago.

@gmasse I agree with what @erik-wramner has proposed. Nice job:)

target/start-mailserver.sh
erik-wramner: Sounds good to me.

aendeavor:
@gmasse could you do a dummy-commit to re-trigger the tests?

gmasse:
@aendeavor, I will push a new commit integrating @erik-wramner advice.

erik-wramner (approved):

aendeavor (approved):

Issue #1707, opened 2020-12-08
Is this project mailgun/sendgrid DIY replacement?


stasbar:
Hello,
I'm looking for mailgun/sendgrid free and opensource replacement. I just need email service to send transactional emails like reset email and email confirmation, also it would be nice if I could listen on mails that arrive to my server so my app can respond to them. 

I would like to be as independent of external services as possible. 

That's why I found this project, but I'm wondering if this is too much (or maybe not enough?) regarding my needs. I don't need any webui that why Mailcow, Mailu, [mailinabox](https://github.com/mail-in-a-box/mailinabox) are out. And projects like [postal](https://github.com/postalhq/postal) and [cuttlefish](https://github.com/mlandauer/cuttlefish) don't work in Docker, which is also a must-have for me.

I know this is kinda offtopic, but I couldn't find a better place to ask this.

MakerMatrix:
Yes, although for what you described you could probably get by with just postfix by itself (or this configured as SMTP-only).  You can configure this to do a great many things, including using it in conjuction with services like Mailgun/Sendgrid.  In my case, my ISP blocks outbound traffic on port 25, so I have my docker-mailserver container configured to relay outbound mail through Sendgrid on port 587 (which my ISP does not block).  I will probably never hit more than 100 emails per day so it should remain free for me.

I also use the built-in dovecot service for IMAP mail.  I think this project is uperior to Mailu (have not looked at the others) because of the full suite of integrated antispam/antimalware features.  All very valuable if you plan to publish email addresses on your server in public places, like I do.

aendeavor:
I agree with @MakerMatrix. You could use `SMTP_ONLY` if you like, and the other solutions that were already proposed.

stasbar:
Thanks. Fortunately, my ISP does not block any port so I managed to run it on my home server. Sending and receiving emails works ✓. Now it's the second part "listening to mails that arrive at my server so my app can respond to them." Is it possible to do with postfix only? Do I need to implement my own MDA where postfix will deliver the mails to? Or is there any place where I can attach a hook that executes a script on every email delivery?

stasbar:
Ok

MakerMatrix:
You can use pipes in /etc/aliases to redirect incoming email text to arbitrary programs.  So if you have an email address like "myapp@mydomain.tld" you’d put something like this as in /etc/aliases:

myapp: | myprog <args>

check the postfix docs for syntax, I probably got something wrong but you get the idea.

[...]

Issue #1706, opened 2020-12-07
Default mail aliases - what mechanism makes the most sense?


MakerMatrix:
* [x] I would like to contribute to the project (code, documentation, advocacy, integration, ...)
I want to make a PR that adds the following default aliases (or similar) at the time of container image creation or deployment
postmaster -> root
amavis -> root
dovecot -> root

## Description
After looking under the hood of this container for a bit, it seems there are multiple, sometimes competing ways to handle aliases:

1: An (empty) /etc/aliases file gets created inside the Dockerfile
2: The start-mailserver.sh script (that gets run out of supervisord?) overwrites /etc/aliases with one that points root -> postmaster
3: The check-for-changes.sh (that gets run out of supervisord?) also overwrites /etc/aliases.  There is a comment here that identfies that as a bug, presumably since it competes with (2).
4: /tmp/docker-mailserver/postfix-aliases.cf (external to the container, making it not a great option to set container defaults).
5: Similar to mechanism (4), /tmp/docker-mailserver/postfix-virtual.cf

I find these multiple options both confusing and somewhat dangerous for unsuspecting users of the package.  For example if you do "postfix standard things" to set up aliases then the scripts run out of supervisord will simply blow them away (silently, at that).

Need to brainstorm with somebody on what is a good way to streamline this and hopefully define *one* "internal" way to create default aliases, while keeping the external post-setup way.  I think it makes sense to put the defaults in /etc/aliases and leave the /tmp/docker-mailserver supervisord stuff alone, so that external configs are not hampered/complicated by the process.  I guess that implies changing the start-mailserver.sh and then fixing the seemingly documented bug in the check-for-changes.sh script.

Thoughts?

erik-wramner:
Well, personally I add the accounts to the `config/postfix-virtual.cf` file using `setup.sh`. In addition I don't use root (who ever logs in with the root account in the docker image?), rather I forward the accounts to one of my own mail accounts.

If we fix this with documentation that is the route I would propose. We could also add a warning from the start script if these aliases have not been defined.

If we are going to add the aliases regardless then /etc/aliases (with some discipline down the line) is probably the way to go.

MakerMatrix:
[...]

PR #1709, opened 2020-12-08
Add default /etc/aliases for amavis, clamav.  Add bsd-mailx.

MakerMatrix:
Also fixed dovecot keygen so it doesn't generate an error msg about RANDFILE.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,6 +38,7 @@ RUN \
   apt-transport-https \
   arj \
   binutils \
+  bsd-mailx \
   bzip2 \
   ca-certificates \
   cabextract \
@@ -134,6 +135,8 @@ RUN sed -i -e 's/include_try \/usr\/share\/dovecot\/protocols\.d/include_try \/e
   # create directory for certificates created by mkcert
   mkdir /etc/dovecot/ssl && \
   chmod 755 /etc/dovecot/ssl  && \
+  touch /etc/dovecot/ssl/.rand && \
+  sed -i 's/dev\/urandom/etc\/dovecot\/ssl\/\.rand/' /usr/share/dovecot/dovecot-openssl.cnf && \
   ./mkcert.sh  && \
   mkdir -p /usr/lib/dovecot/sieve-pipe /usr/lib/dovecot/sieve-filter /usr/lib/dovecot/sieve-global && \
   chmod 755 -R /usr/lib/dovecot/sieve-pipe /usr/lib/dovecot/sieve-filter /usr/lib/dovecot/sieve-global
@@ -197,7 +200,7 @@ RUN mkdir /var/run/fetchmail && chown fetchmail /var/run/fetchmail
 # Configures Postfix
 COPY target/postfix/main.cf target/postfix/master.cf /etc/postfix/
 COPY target/postfix/header_checks.pcre target/postfix/sender_header_filter.pcre target/postfix/sender_login_maps.pcre /etc/postfix/maps/
-RUN echo "" > /etc/aliases
+RUN touch /etc/aliases
 
 # Configuring Logs
 RUN sed -i -r "/^#?compress/c\compress\ncopytruncate" /etc/logrotate.conf && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -200,7 +200,7 @@ RUN mkdir /var/run/fetchmail && chown fetchmail /var/run/fetchmail
 # Configures Postfix
 COPY target/postfix/main.cf target/postfix/master.cf /etc/postfix/
 COPY target/postfix/header_checks.pcre target/postfix/sender_header_filter.pcre target/postfix/sender_login_maps.pcre /etc/postfix/maps/
-RUN touch /etc/aliases
+RUN : >/etc/aliases
 
 # Configuring Logs
 RUN sed -i -r "/^#?compress/c\compress\ncopytruncate" /etc/logrotate.conf && \
@@ -210,6 +210,9 @@ RUN sed -i -r "/^#?compress/c\compress\ncopytruncate" /etc/logrotate.conf && \
   chown -R clamav:root /var/log/mail/clamav.log && \
   touch /var/log/mail/freshclam.log && \
   chown -R clamav:root /var/log/mail/freshclam.log && \
+  mkdir -p /var/log/supervisor && \
+  touch /var/log/supervisor/supervisord.log && \
+  chown -R syslog:root /var/log/supervisor && \
   sed -i -r 's|/var/log/mail|/var/log/mail/mail|g' /etc/rsyslog.conf && \
   sed -i -r 's|;auth,authpriv.none|;mail.none;mail.error;auth,authpriv.none|g' /etc/rsyslog.conf && \
   sed -i -r 's|LogFile /var/log/clamav/|LogFile /var/log/mail/|g' /etc/clamav/clamd.conf && \
--- a/target/check-for-changes.sh
+++ b/target/check-for-changes.sh
@@ -82,7 +82,7 @@ do
       # Generate default /etc/aliases
       echo "root: ${PM_ADDRESS}" >/etc/aliases
       echo "amavis: ${PM_ADDRESS}" >>/etc/aliases
-      if [[ ${ENABLE_CLAMAV} -1 1 ]]
+      if [[ ${ENABLE_CLAMAV} -eq 1 ]]
       then
         echo "clamav: ${PM_ADDRESS}" >>/etc/aliases
       fi
--- a/target/check-for-changes.sh
+++ b/target/check-for-changes.sh
@@ -79,8 +79,12 @@ do
         esac
       done
 
-      # regenerate postix aliases
+      # Generate default /etc/aliases
       echo "root: ${PM_ADDRESS}" >/etc/aliases
+      echo "amavis: ${PM_ADDRESS}" >>/etc/aliases
+      echo "clamav: ${PM_ADDRESS}" >>/etc/aliases
+
+      # Append external postifx aliases
       if [[ -f /tmp/docker-mailserver/postfix-aliases.cf ]]
       then
         cat /tmp/docker-mailserver/postfix-aliases.cf >>/etc/aliases
--- a/target/check-for-changes.sh
+++ b/target/check-for-changes.sh
@@ -82,7 +82,10 @@ do
       # Generate default /etc/aliases
       echo "root: ${PM_ADDRESS}" >/etc/aliases
       echo "amavis: ${PM_ADDRESS}" >>/etc/aliases
-      echo "clamav: ${PM_ADDRESS}" >>/etc/aliases
+      if [[ ${ENABLE_CLAMAV} -1 1 ]]
+      then
+        echo "clamav: ${PM_ADDRESS}" >>/etc/aliases
+      fi
 
       # Append external postifx aliases
       if [[ -f /tmp/docker-mailserver/postfix-aliases.cf ]]
--- a/target/start-mailserver.sh
+++ b/target/start-mailserver.sh
@@ -122,6 +122,7 @@ function register_functions
 
   _register_setup_function "_setup_dkim"
   _register_setup_function "_setup_ssl"
+  _register_setup_function "_setup_postfix_aliases"
 
   [[ ${POSTFIX_INET_PROTOCOLS} != "all" ]] && _register_setup_function "_setup_inet_protocols"
 
@@ -137,7 +138,6 @@ function register_functions
   _register_setup_function "_setup_postfix_sasl"
   _register_setup_function "_setup_postfix_sasl_password"
   _register_setup_function "_setup_security_stack"
-  _register_setup_function "_setup_postfix_aliases"
   _register_setup_function "_setup_postfix_vhost"
   _register_setup_function "_setup_postfix_dhparam"
   _register_setup_function "_setup_postfix_postscreen"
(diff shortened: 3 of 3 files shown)


aendeavor:
I like this changes - thanks for creating this PR:)

I've added some comments I'd like to resolve first, and I've requested a review from @erik-wramner as weill, which I would like to see before merging.

MakerMatrix:
On reviewing your comments I think it would be much better if we wrapped the the clamav alias inside an if stmt, to check whether the user asked for that feature or not.  I put it in because I noticed that cron job(s) get run as the clamav user.  But I guess only if user asked for that feature.

aendeavor:
> On reviewing your comments I think it would be much better if we wrapped the the clamav alias inside an if stmt, to check whether the user asked for that feature or not. I put it in because I noticed that cron job(s) get run as the clamav user. But I guess only if user asked for that feature.

I accidentally did not press the submit button. You can now see my comments, sorry:) These are the ones I was referring to.

aendeavor:
In #1706 you mentioned that `/etc/aliases` is completely emptied on "check-for-changes.sh`-runs. I've already mentioned it in a comment - this causes some headache. I cannot confidently say whether this may break things.

MakerMatrix:
> In #1706 you mentioned that `/etc/aliases` is completely emptied on "check-for-changes.sh`-runs. I've already mentioned it in a comment - this causes some headache. I cannot confidently say whether this may break things.

Fair question, but if you look closely you see that line was already there - I only made the comment, and then added the two lines below to append amavis and clamav.  It was always the case that /etc/aliases got recreated from scratch when these functions are called.  There are some comments about this and a TODO about making these things atomic instead of rebuilding them from scratch, [here](https://github.com/tomav/docker-mailserver/blob/2629b575900759920a31ad6a5c0603cbc7b25af7/target/check-for-changes.sh#L56).

aendeavor:
Oh my... I totally missed that. I didn't look close enough. Resolved:)

target/start-mailserver.sh
```diff
@@ -1031,8 +1031,12 @@ s/$/ pcre:\/etc\/postfix\/regexp/
 
   _notify 'inf' "Configuring root alias"
 
+  # Generate default /etc/aliases
   echo "root: ${POSTMASTER_ADDRESS}" > /etc/aliases
```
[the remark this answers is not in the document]
aendeavor: Turns out: I didn't look close enough :D

Dockerfile
```diff
@@ -197,7 +200,7 @@ RUN mkdir /var/run/fetchmail && chown fetchmail /var/run/fetchmail
 # Configures Postfix
 COPY target/postfix/main.cf target/postfix/master.cf /etc/postfix/
 COPY target/postfix/header_checks.pcre target/postfix/sender_header_filter.pcre target/postfix/sender_login_maps.pcre /etc/postfix/maps/
-RUN echo "" > /etc/aliases
+RUN touch /etc/aliases
```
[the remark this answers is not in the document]
MakerMatrix: It was not clear to me why that line is needed at all.  I made the assumption that the goal was to create the file anew, since it's the Dockerfile here.  I changed it because I thought the old way was clunky.

I would have liked to have the defaults put in here but I don' think the environment for $HOSTNAME and $DOMAIN are exposed inside the Dockerfile?  I had tested versions that did things like:

amavis: postmaster

...etc...but that syntax in /etc/aliases forced the mail to get delivered locally (/var/mail) or fail if there is no user (i.e. postmaster).  Only if you specify something after @, will it deliver through SMTP.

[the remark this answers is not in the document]
aendeavor: I understand. I'd propose the following to get the original functionality without the clunkyness:

``` DOCKERFILE
RUN : >/etc/aliases
```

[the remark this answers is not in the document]
MakerMatrix: I'm not a bash user, really.  If you say it does that, then I like its cleanliness.

How do I update it now that it's already a PR?  Just push this change to the same branch?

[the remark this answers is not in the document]
aendeavor: > I'm not a bash user, really.  If you say it does that, then I like its cleanliness.

I just wrote a little test Dockerfile - my suggested solution should work.

> How do I update it now that it's already a PR?  Just push this change to the same branch?

Exactly:)

target/start-mailserver.sh
```diff
@@ -122,6 +122,7 @@ function register_functions
 
   _register_setup_function "_setup_dkim"
   _register_setup_function "_setup_ssl"
+  _register_setup_function "_setup_postfix_aliases"
```
aendeavor, line 125: What's this about?:)

target/check-for-changes.sh
```diff
@@ -79,8 +79,15 @@ do
         esac
       done
 
-      # regenerate postix aliases
+      # Generate default /etc/aliases
       echo "root: ${PM_ADDRESS}" >/etc/aliases
+      echo "amavis: ${PM_ADDRESS}" >>/etc/aliases
+      if [[ ${ENABLE_CLAMAV} -1 1 ]]
```
aendeavor, line 85: this is invalid, you accidentally used `-1` instead of `-eq` :)

MakerMatrix: OMG

target/start-mailserver.sh
MakerMatrix: Sloppiness.  I was moving it around to test why aliases do not get created when using default mailserver.env.  It's still a mystery.

aendeavor:
Looks somewhat fine to me now. As mentioned in #1706 I am not sure whether the change now adds the aliases for root, amavis and clam, but I really don't see why this would be case. So is this working?:)

aendeavor:
What's the status here? @erik-wramner  what do you think?

Dockerfile
```diff
@@ -197,7 +200,7 @@ RUN mkdir /var/run/fetchmail && chown fetchmail /var/run/fetchmail
 # Configures Postfix
 COPY target/postfix/main.cf target/postfix/master.cf /etc/postfix/
 COPY target/postfix/header_checks.pcre target/postfix/sender_header_filter.pcre target/postfix/sender_login_maps.pcre /etc/postfix/maps/
-RUN echo "" > /etc/aliases
+RUN : >/etc/aliases
```
casperklein, line 203: Why not just `RUN >/etc/aliasses` which is even shorter ;-)

aendeavor: True indeed, and that would work, but I think we talked about this a while ago :D This nooo `:` makes the intention clear and is idiomatic Bash :)

casperklein: Now that you mention it, I can remember 😇  But _personally_ I don't like using functionless commands like in that case and don't see how that makes the intention any clearer. Using something like `while :; do foo; done` is a better noop use case.
But in the end, it's not worth discussing about +/- 2 bytes 😅 👍

aendeavor: +1

MakerMatrix:
> Looks somewhat fine to me now. As mentioned in #1706 I am not sure whether the change now adds the aliases for root, amavis and clam, but I really don't see why this would be case. So is this working?:)

Yeah it works, in conjuction with #1707. In 1706 I had made some faulty assumptions about the default testing container config, just learning how to build and test. :).  It was a simple matter of getting a a couple things configured away from defaults which in any real-world deployment would be the first things somebody would set.

aendeavor:
@erik-wramner when you've reviewed this PR, feel free to either merge it or close it. SInce we are working on the alias code anyways, I will leave it up to you what to do with this PR.

Dockerfile
```diff
@@ -134,6 +135,8 @@ RUN sed -i -e 's/include_try \/usr\/share\/dovecot\/protocols\.d/include_try \/e
   # create directory for certificates created by mkcert
   mkdir /etc/dovecot/ssl && \
   chmod 755 /etc/dovecot/ssl  && \
+  touch /etc/dovecot/ssl/.rand && \
+  sed -i 's/dev\/urandom/etc\/dovecot\/ssl\/\.rand/' /usr/share/dovecot/dovecot-openssl.cnf && \
```
erik-wramner, line 139: Why are you changing the dovecot random function in this way and how is it related to /etc/aliases? If this changes should be done (and please explain why) it should probably go in a separate commit.

Dockerfile
```diff
@@ -207,6 +210,9 @@ RUN sed -i -r "/^#?compress/c\compress\ncopytruncate" /etc/logrotate.conf && \
   chown -R clamav:root /var/log/mail/clamav.log && \
   touch /var/log/mail/freshclam.log && \
   chown -R clamav:root /var/log/mail/freshclam.log && \
+  mkdir -p /var/log/supervisor && \
+  touch /var/log/supervisor/supervisord.log && \
+  chown -R syslog:root /var/log/supervisor && \
```
erik-wramner, line 215: Please explain the need for this as well and how it is related to the aliases. It looks like you are trying to solve something, but what?

target/check-for-changes.sh
```diff
@@ -79,8 +79,15 @@ do
...
+      # Generate default /etc/aliases
       echo "root: ${PM_ADDRESS}" >/etc/aliases
+      echo "amavis: ${PM_ADDRESS}" >>/etc/aliases
+      if [[ ${ENABLE_CLAMAV} -eq 1 ]]
+      then
+        echo "clamav: ${PM_ADDRESS}" >>/etc/aliases
+      fi
+
+      # Append external postifx aliases
       if [[ -f /tmp/docker-mailserver/postfix-aliases.cf ]]
```
erik-wramner, line 91: This is probably fine, but we could check if postfix-aliases.cf has a mapping for amavis. Many existing installations may already have one (mine does). Not a big deal though, perhaps it is better to keep the code simple and write something about it in the README.

target/start-mailserver.sh
```diff
@@ -1031,8 +1031,15 @@ s/$/ pcre:\/etc\/postfix\/regexp/
 
   _notify 'inf' "Configuring root alias"
 
+  # Generate default /etc/aliases
   echo "root: ${POSTMASTER_ADDRESS}" > /etc/aliases
+  echo "amavis: ${POSTMASTER_ADDRESS}" >> /etc/aliases
+  if [[ ${ENABLE_CLAMAV} -eq 1 ]]
+  then
+    echo "clamav: ${POSTMASTER_ADDRESS}" >> /etc/aliases
+  fi
```
erik-wramner, line 1041: See previous comment.

erik-wramner:
@aendeavor @MakerMatrix I don't understand what several of the changes are for. Perhaps the important fixes are done elsewhere? If so let's close this. If not feel free to motivate and perhaps we can merge instead. It feels as if the changes are trying to fix several things that are not related to the alias issue at all?

aendeavor:
Very true. There are some hotfixes in here that could probably go in another PR, you're right. @MakerMatrix can explain them and then we can decide if we need another PR.

Dockerfile
MakerMatrix: This is an unrelated build fix that I tossed in quickly.  As I recall, during the build phase, openssl throws an error when generating the dovecot key.  IIRC it's because it tries to write entropy to  /dev/urandom which of course is not allowed.  So now I have it creating an empty, writable file for it to put its entropy in, and that filename gets put in dovecot-openssl.cnf.

Dockerfile
MakerMatrix: It's not related to aliases, sorry.  I believe I found errors in the logs where supervisord was throwing an error when trying to write to /var/log/supervisor/supervisord.log and it did not exist.  I do apologize for not documenting this one better, I see I didn't even put a comment along with the PR and frankly even I cannot remember exactly where I found this one.

target/check-for-changes.sh
MakerMatrix: Hmmm, mine didn't.   My /tmp/dockermailserver/postfix-aliases.cf is empty.  If that's a better place to do these, then by all means.

target/start-mailserver.sh
MakerMatrix: Ditto.

Dockerfile
MakerMatrix: I'll amend the first sentence of the previous to say that it *was* tangentially related to aliases in the sense that I found it when trying to discover exactly what was going on with aliases when supervisord brings up the mail service the first time after container creation.  There are some conditions where they don't get added, which appear to be based on how some of the environment variables are set.  I never narrowed it down further than that but thought it important for supervisord to be able to write its logs "out of the box"

Dockerfile
casperklein: That is the build message:

    Generating a RSA private key
    ....+++++
    ........+++++
    writing new private key to '/etc/dovecot/ssl/dovecot.key'
    -----
    Cannot write random bytes:
    140123681975557:error:2407007A:random number generator:RAND_write_file:Not a regular file:../crypto/rand/randfile.c:183:Filename=/dev/urandom

If I understand https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=922732#20 correctly, it should be safe to remove RANDFILE completely from the configuration:

    | OpenSSL 1.1.1 introduced a new random generator (CSPRNG) with an improved
    | seeding mechanism. The new seeding mechanism makes it unnecessary to
    | define a RANDFILE for saving and restoring randomness. This option is
    | retained mainly for compatibility reasons.

In any case, this should be discussed/done in a separate PR.

aendeavor: @casperklein I'm completely with you there.

@MakerMatrix when we have finished the migration to https://github.com/docker-mailserver/docker-mailserver, could open a few several small PRs that go through each of your changes one by one?

aendeavor: @casperklein @MakerMatrix I'm currently working on a bigger improvement of the `Dockerfile`. @casperklein How would I go about removing the RANDFILE completely?

aendeavor: @casperklein Nevermind - I did it by removing `RANDFILE` from `mkcert.sh`. Done.

MakerMatrix: > @MakerMatrix when we have finished the migration to https://github.com/docker-mailserver/docker-mailserver, could open a few several small PRs that go through each of your changes one by one?

Happy to.

MakerMatrix:
At this point I think this can be closed out without merging.  I will try to recreate the supervisord.log issue and open that as a separate PR after the migration to the docker-mailserver github account (will I know when that happens?).  I think everything else here including the additional default aliases is being addressed in different fixes.

aendeavor:
> At this point I think this can be closed out without merging.  I will try to recreate the supervisord.log issue and open that as a separate PR after the migration to the docker-mailserver github account (will I know when that happens?).  I think everything else here including the additional default aliases is being addressed in different fixes.

A good decision. You will see when the repository migration is ready in the README and when the other PRs are all closed - just look out for that.

PR #1711, opened 2020-12-09
Get rid of syslogd warnings about kernel logging.

MakerMatrix:
- Modify Dockerfile to comment out noisy/errant kernel logging line in /etc/rsyslogd.conf
- Add lines in README.md that prompt users to create an alias that captures messages to postmaster.
- Add commented line in /etc/postfix/main.cf to easily set 'myorigin=$mydomain' if desired.  Then e.g. postmaster emails will be addressed from "postmaster@domain.tld" instead of "postmaster@host.domain.tld".  This may require a regex or an additional or modified virtual alias for postfix to accept mail from such an address.  I have not documented that in these patches.  If people watch mail.log it will be clear.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -222,7 +222,7 @@ RUN sed -i -r "/^#?compress/c\compress\ncopytruncate" /etc/logrotate.conf && \
   sed -i -e 's/^\(POLICYHELPER=\).*/\1/' /usr/sbin/invoke-rc.d && \
   # Prevent syslog warning about imklog permissions.
   # We won't want to look at that inside a container anyway.
-  sed -i -e 's/^module\(load="imklog"\)/#module\(load="imklog"\)/' /etc/rsyslog.conf && \
+  sed -i -e 's/^module(load=\"imklog\")/#module(load=\"imklog\")/' /etc/rsyslog.conf && \
   # prevent email when /sbin/init or init system is not existing \
   sed -i -e 's|invoke-rc.d rsyslog rotate > /dev/null|/usr/bin/supervisorctl signal hup rsyslog >/dev/null|g' /usr/lib/rsyslog/rsyslog-rotate
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -220,6 +220,9 @@ RUN sed -i -r "/^#?compress/c\compress\ncopytruncate" /etc/logrotate.conf && \
   # prevent syslog logrotate warnings \
   sed -i -e 's/\(printerror "could not determine current runlevel"\)/#\1/' /usr/sbin/invoke-rc.d && \
   sed -i -e 's/^\(POLICYHELPER=\).*/\1/' /usr/sbin/invoke-rc.d && \
+  # Prevent syslog warning about imklog permissions.
+  # We won't want to look at that inside a container anyway.
+  sed -i -e 's/^module\(load="imklog"\)/#module\(load="imklog"\)/' /etc/rsyslog.conf && \
   # prevent email when /sbin/init or init system is not existing \
   sed -i -e 's|invoke-rc.d rsyslog rotate > /dev/null|/usr/bin/supervisorctl signal hup rsyslog >/dev/null|g' /usr/lib/rsyslog/rsyslog-rotate
 
--- a/README.md
+++ b/README.md
@@ -118,6 +118,7 @@ chmod a+x ./setup.sh
 docker-compose up -d mail
 
 ./setup.sh email add <user@domain> [<password>]
+./setup.sh alias add postmaster@<domain> <user@domain>
 ./setup.sh config dkim
 ```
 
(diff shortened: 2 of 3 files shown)


aendeavor:
Requested a review from @erik-wramner and @fbartels because I am not confidently review these changes - more on the script side myself:)

aendeavor:
Damn it, the quota tests again. @martin-schulze-vireso this is the second time close after one another where this test fails. Seems like this hasn't been resolved after all...

martin-schulze-vireso:
It looks like a true failure. The second large mail should not have been delivered but we got a quota message on the third. I am not sure if there is some kind of race condition due to sending the mails too fast. However, that would likely be outside of our control to fix. 

Maybe we should wait for the first mail to be delivered before sending the next one.

target/postfix/main.cf
```diff
@@ -10,6 +10,8 @@ readme_directory = no
 alias_maps = hash:/etc/aliases
 alias_database = hash:/etc/aliases
 mydestination = $myhostname, localhost.$mydomain, localhost
+# Uncomment for container accounts to masquerade as the domain instead of the host
+# myorigin = $mydomain
```
fbartels, line 14: I am not quite sure about this one as it changes the default behaviour for everyone. Rather it should be made as a configuration whish as afaik already possible in this project.

Dockerfile
```diff
@@ -220,6 +220,9 @@ RUN sed -i -r "/^#?compress/c\compress\ncopytruncate" /etc/logrotate.conf && \
   # prevent syslog logrotate warnings \
   sed -i -e 's/\(printerror "could not determine current runlevel"\)/#\1/' /usr/sbin/invoke-rc.d && \
   sed -i -e 's/^\(POLICYHELPER=\).*/\1/' /usr/sbin/invoke-rc.d && \
+  # Prevent syslog warning about imklog permissions.
+  # We won't want to look at that inside a container anyway.
+  sed -i -e 's/^module(load=\"imklog\")/#module(load=\"imklog\")/' /etc/rsyslog.conf && \
```
fbartels, line 225: 👍

fbartels (changes requested):

target/postfix/main.cf
aendeavor: When I looked over this, I was not sure too. I think there already is an option to do this: Looking at the README, there's a line stating

"Note: If you want to use a bare domain (host name equals domain name) see FAQ."

I guess this change is not necessary and may possibly break things. I would rather like to remove it as well.

aendeavor:
@martin-schulze-vireso you're right. But before doing that, I will see what the next test result bring, whether the quota test is "properly" failing or whether this is random as well.

target/postfix/main.cf
MakerMatrix: It's commented, I put it there to make it easier for people (like me, which took some time) to find.

And it does break things.  And I was too lazy to document it.  Either way.

MakerMatrix: I checked the sesction of the FAQ that you referenced.  That part of the FAQ is not about mail originating from the container host.  It's about situations where the MX record == the domainname.

Here is the relevent section of the postfix docs if you want to understand what I'm trying to do with this:

http://www.postfix.org/postconf.5.html#myorigin

aendeavor (approved):

Dockerfile
casperklein, line 225: I think, the quotes do not need to be escaped:
`sed -i -e 's/^module(load="imklog")/#module(load="imklog")/' /etc/rsyslog.conf && \`

erik-wramner:
I think this looks good. One thing though. Where we add the postmaster alias we could also add an alias for amavis.

There is another PR about adding both in the start script. @aendeavor I think we should either remove the postmaster alias from the readme in this PR and implement it in the other or we should add the amavis alias here. We could of course also do both and then remove the aliases from the readme later when the other PR has been released.

README.md
```diff
@@ -118,6 +118,7 @@ chmod a+x ./setup.sh
 docker-compose up -d mail
 
 ./setup.sh email add <user@domain> [<password>]
+./setup.sh alias add postmaster@<domain> <user@domain>
```
erik-wramner, line 121: Can we add an alias for amavis here as well, please? The system sometimes sends mail to amavis.

erik-wramner (changes requested):

README.md
aendeavor: I think you're right. I'd prefer doing "both", as you said, removing the entries from the README when the other PR is released. Thanks for noticing:)

Dockerfile
aendeavor: Yeah, @casperklein is correct. I overlooked that:)

README.md
MakerMatrix: This was really meant to be added to #1709 but I forgot.  The idea is, the alias amavis->postmaster gets auto-added to /etc/aliases (also clamav->postmaster and then in the docs we prompt users to alias postmaster to a monitored inbox.

aendeavor: What's your take here @erik-wramner ?

fbartels (approved):

erik-wramner (approved):

PR #1736, opened 2021-01-02
Allow manual domains for dkim generator

### LINKED ISSUE #1735: The dkim key generator does not work for ldap setups
The dkim key generator does not work for ldap setups

An option should be provided to the `generate-dkim-config` script for ldap setups to force domain names
Relying on files makes no sense for ldap setups

williamdes:
Fix #1735 - The dkim key generator does not work for ldap setups

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -329,13 +329,13 @@ function _main
       shift ; case ${1:-} in
         set      ) shift ; _docker_image setquota "${@}" ;;
         del      ) shift ; _docker_image delquota "${@}" ;;
-        *        )   _usage ;;
+        *        ) _usage ;;
       esac
       ;;
 
     config)
       shift ; case ${1:-} in
-        dkim     ) _docker_image generate-dkim-config "${2:-2048}" ;;
+        dkim     ) _docker_image generate-dkim-config "${2:-2048}" "${3:-}" ;;
         ssl      ) _docker_image generate-ssl-certificate "${2}" ;;
         *        ) _usage ;;
       esac
--- a/README.md
+++ b/README.md
@@ -139,6 +139,20 @@ docker-compose up -d mail
 
 ### Miscellaneous
 
+### DKIM manual domain mode
+
+LDAP setups would need to use this mode because the user accounts are not in `postfix-accounts.cf` or `postfix-virtual.cf`.
+
+``` BASH
+# without SELinux
+docker-compose up -d mail
+./setup.sh config dkim 2048 'domain1.com,domain2.fr'
+
+# with SELinux
+docker-compose up -d mail
+./setup.sh -Z config dkim 2048 'domain1.com,domain2.fr'
+```
+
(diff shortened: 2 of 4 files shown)


setup.sh
```diff
@@ -335,7 +335,7 @@ function _main
 
     config)
       shift ; case ${1:-} in
-        dkim     ) _docker_image generate-dkim-config "${2:-2048}" ;;
+        dkim     ) _docker_image generate-dkim-config "${2:-2048}" "${3}"
```
aendeavor, line 338: I would like to see `${3:-}` here since the `set -u` from above will end the script immediately of `$3` is not provided, and this is only the case for LDAP setups. The tests should show this. This is rather intricate, and you could not possibly know this since this script has some intricate `set ...` options

aendeavor (approved):
Looks very fine to me. Please only adjust the `setup.sh` script as I suggested.

williamdes:
It should be okay, I did script that without testing it and if the bats tests passes, then it is okay.

I will test it but for now I am working on ldap domains

aendeavor:
I have now corrected the scripts myself as there was a syntax error and a misuse of `-z` still in it. We'll check tests then.

williamdes:
Looks like I should add back the volume mounts like the test that I copied it from

aendeavor:
Not sure whether that fixes the `PermissionDenied` errors, but it's worth a shot.

williamdes:
No more errors but tests fail
https://travis-ci.org/github/tomav/docker-mailserver/builds/752566900#L2397

Looks like the domains option does not work or something else

aendeavor:
That's unfortunate - but we'll need to see all tests passing before merging this PR. I guess some more debugging needs to be done here.

williamdes:
This should work now, `-z` was the good one and not `-n`
It works in my container

williamdes:
F***
`toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit`
Well, the tests will run another day..

techgourmet:
You could try getting a free account at docker
That should remove the amount of pulls.

Kind regardst,

Roy van der Linden

Certified Data Protection Officer DPCDPO1026251-2018-01

TechGourmet
Communication Consultancy




Dit bericht is alleen bestemd voor de geaddresseerde en kan informatie bevatten die privé, vertrouwelijk of anderszins niet
openbaar mag worden gemaakt. In de lezer niet de beoogde ontvanger van dit bericht is, wordt deze hierbij geinformeerd dat
de informatie in dit bericht niet gekopieerd of anderszins verspreidt mag worden. Tevens wordt u hierbij verzocht om ons hiervan
direct op de hoogte te stellen en het bericht aan ons te retourneren. Bij voorbaat dank voor uw medewerking.

aendeavor:

PR #1737, opened 2021-01-02
Fix shebangs

williamdes:
- Fixes shebangs

### DIFF
--- a/config/user-patches.sh.dist
+++ b/config/user-patches.sh.dist
@@ -1,4 +1,4 @@
-#!/bin/bash
+#! /bin/bash
 ##
 # This user patches script runs right before starting the daemons. That means, all the other configuration is in place, so the script can make final adjustments.
 # If you modify any supervisord configuration, make sure to run "supervisorctl update" afterwards.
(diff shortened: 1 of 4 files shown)


aendeavor:
I do not see how this fixes anything - don't get me wrong, I do not want to impolite, but the choices of shebangs and readme fences were on purpose. I think they are fine they way they are.

williamdes:
Okay, but for me the readme fences are incorrect

Please let me know why some script use a space and some do not so I can revert that :)

aendeavor:
This project uses all-uppercase readme fences with a space between the three hyphens. There are still some inconsistencies - feel free to resolve them. The scripts should all use `#! /SOME/PATH`, i.e. with a space after `#!`. If there are inconsistencies too, I'd be happy to see them resolved too.

williamdes:
> This project uses all-uppercase readme fences with a space between the three hyphens.

But why and where does it come from, it just feels strange to me

Okay I will revert and fix the non spaced files

aendeavor:
> > This project uses all-uppercase readme fences with a space between the three hyphens.
> 
> But why and where does it come from, it just feels strange to me
> 
> Okay I will revert and fix the non spaced files

I guess personal preference. To me, it feels strange to do it in another way. I just feel like this is clean and I would not really like to change whats stable there:) I'd like the readme fences to stay as they are.

aendeavor (approved):

Issue #1694, opened 2020-11-23
virtual_alias_domains


dashohoxha:
## Subject

* [ ] I would like to configure a not documented mail server use case
* [ ] I would like some feedback concerning a use case

## Description

Maybe it is explained somewhere on the docs, but I couldn't find it out.

I'd like to use the mailserver for multiple mail domains.
What is the recommended way for adding `virtual_alias_domains`?

aendeavor:
Hey there @dashohoxha.

Although I do not know the answer to your question, I will mark @erik-wramner / @casperklein. They know better than I do :)

If this use case really is not documented, we will add this to the documentation.

dashohoxha:
By testing I found out that **mailserver** supports multiple email domains out of the box. So, no need to add `virtual_alias_domains` or anything else on the configuration. All that is need is to use `./setup.sh` to add email accounts or aliases (forwarders). For example:

```
./setup.sh email add user1@test1.example.org pass123
./setup.sh email add user1@test2.example.org pass123
./setup.sh alias add user2@test1.example.org xyz@gmail.com
./setup.sh alias add user2@test2.example.org xyz@gmail.com
```

Everything works as expected. I am sorry for opening this issue without proper testing.

However, maybe it is worth to mention somewhere on the docs that **mailserver** supports multiple email domains out of the box.

aendeavor:
Oh, then I misread your question. I thought there were some other difficulties involved. Sorry, I could've told you that earlier:) But your suggestion is good, I will make sure this note ends up somewhere in the `README`.

aendeavor:
Done with ec290acd1f92e1e44773085df8d8872f6f13cb25.

williamdes:
I think `virtual_alias_domains` is needed for ldap setups.

williamdes:
Currently investigating with another test, will let you know soon

williamdes:
Test added as https://github.com/desportes/infrastructure/commit/1ae727a49f808115ad489a2ad10a1e44d387c17a
Reproduced ([log](https://github.com/desportes/infrastructure/runs/1636654139?check_suite_focus=true#step:9:20)) !
```
run-tests_1  | PHPUnit 9.5.0 by Sebastian Bergmann and contributors.
run-tests_1  | 
run-tests_1  | ........sendmail: RCPT TO:<contact@another-domain.com> (554 5.7.1 <contact@another-domain.com>: Relay access denied)
[...]

PR #1739, opened 2021-01-03
Add POSTFIX_VIRTUAL_ALIAS_DOMAINS and POSTFIX_VIRTUAL_ALIAS_MAPS

williamdes:
- Add POSTFIX_VIRTUAL_ALIAS_DOMAINS
- Add POSTFIX_VIRTUAL_ALIAS_MAPS

Fixes: https://github.com/tomav/docker-mailserver/issues/1694

### DIFF
--- a/test/mail_postfix_virtual.bats
+++ b/test/mail_postfix_virtual.bats
@@ -1,0 +1,71 @@
+load 'test_helper/common'
+
+@test "checking postfix virtual alias domains no ENV set" {
+  local PRIVATE_CONFIG
+  PRIVATE_CONFIG="$(duplicate_config_for_container . )"
+  docker run -d --name mail_postfix_virtual_alias_domains \
+            -v "${PRIVATE_CONFIG}":/tmp/docker-mailserver \
+            -v "$(pwd)/test/test-files":/tmp/docker-mailserver-test:ro \
+            -h mail.my-domain.com -t "${NAME}"
+
+  teardown() { docker rm -f mail_postfix_virtual_alias_domains; }
+
+  wait_for_finished_setup_in_container mail_postfix_virtual_alias_domains
+
+  run docker exec mail_postfix_virtual_alias_domains postconf virtual_alias_domains
+  assert_output "virtual_alias_domains = ???"
+  assert_success
+}
+
+@test "checking postfix virtual alias domains with ENV" {
+  local PRIVATE_CONFIG
+  PRIVATE_CONFIG="$(duplicate_config_for_container . )"
+  docker run -d --name mail_postfix_virtual_alias_domains \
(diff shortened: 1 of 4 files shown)


test/mail_postfix_virtual.bats
```diff
@@ -0,0 +1,71 @@
...
+            -v "${PRIVATE_CONFIG}":/tmp/docker-mailserver \
+            -v "$(pwd)/test/test-files":/tmp/docker-mailserver-test:ro \
+            -h mail.my-domain.com -t "${NAME}"
+
+  teardown() { docker rm -f mail_postfix_virtual_alias_domains; }
+
+  wait_for_finished_setup_in_container mail_postfix_virtual_alias_domains
+
+  run docker exec mail_postfix_virtual_alias_domains postconf virtual_alias_domains
+  assert_output "virtual_alias_domains = ???"
```
williamdes, line 16: All this file is not tested because the tests do not run on my machine..
I will have to debug with travis ..
If anyone can help :pray:

aendeavor:
It seems like Travis is slowly giving up. I will tend to the migration of this project to get GitHub Actions in the mix. When this is done, could you open this PR again in the new repo? I don't think we can use Travis any longer.

aendeavor:
Closed due to merging of docker-mailserver#6 :)

aendeavor:
Check that you do not interfere with:

<https://github.com/docker-mailserver/docker-mailserver/blob/master/target/scripts/start-mailserver.sh#L1030-L1039>

These lines will be altered by #1754, which will likely be merged first.

aendeavor:
On second thought, this might very well be integrated into the `_setup_postfix_aliases` function. If you think that's fine, I'd like to see that:)

aendeavor:
Hmm, I already suspected that there are some tests that check exactly that. I'm not sure whether altering the tests is necessary, but I think so. But you will need to be _very_ cautious when altering tests.

wernerfred:
Just wondering why you will add  there ENV vars as your last update on the issue was that they are not needed due to your misconfigured LDAP filters. :)

> Just to update everyone, after some more debug in production I found out that my LDAP filters where not correct.
So the alias/domains where never returned to postfix ldap maps.
Fixed by: https://github.com/desportes/infrastructure/commit/5678ceb6275c5932d428a52cdd8e6ca563cdbced

> No more [virtual_alias_domains](http://www.postfix.org/postconf.5.html#virtual_alias_domains) needed, because they default to virtual_alias_maps that are filled with the ldap seach config (that is itself wired to an ENV passed filter)

_Originally posted by @williamdes in https://github.com/docker-mailserver/docker-mailserver/issues/1694#issuecomment-754280125_

williamdes:
Yeah, good point :smile: 
I abandon this PR finally, that will make users search for good reasons rather than hacks ^^

PR #1752, opened 2021-01-16
Support null USE_TTY for non-interactive use, e.g. ssh

buck:
setup.sh does not work without a tty, as a null USE_TTY leaves stray single quotes ('') on the command line.  The extra double quotes are unnecessary.


aendeavor:
I'm sorry for the inconvenience, but we just migrated to docker-mailserver/docker-mailserver. Please re-open your PR over there. This is the new development repository.

PR #1753, opened 2021-01-16
Allow manual domains for dkim generator

### LINKED ISSUE #1735: The dkim key generator does not work for ldap setups

williamdes:
Fix #1735 - The dkim key generator does not work for ldap setups

Replaces: docker-mailserver/docker-mailserver#1736


wernerfred:
Like in the previous PR there are some tests failing:
<details>
  <summary>Failing tests</summary>

```
not ok 242 checking opendkim: generator creates default keys size
#  in test file test/tests.bats, line 479)
#   `assert_output 6' failed
# 
# -- output differs --
# expected : 6
# actual   : 4
# --
# 
not ok 243 checking opendkim: generator creates key size 2048
#  in test file test/tests.bats, line 505)
#   `assert_output 6' failed
# 
# -- output differs --
# expected : 6
# actual   : 4
# --
# 
not ok 244 checking opendkim: generator creates key size 1024
#  in test file test/tests.bats, line 531)
#   `assert_output 6' failed
# 
# -- output differs --
# expected : 6
# actual   : 4
# --
# 
not ok 245 checking opendkim: generator creates keys, tables and TrustedHosts
#  in test file test/tests.bats, line 553)
#   `assert_output 6' failed
# 
# -- output differs --
# expected : 6
# actual   : 4
# --
# 
not ok 246 checking opendkim: generator creates keys, tables and TrustedHosts without postfix-accounts.cf
#  in test file test/tests.bats, line 584)
#   `assert_output 5' failed
# 
# -- output differs --
# expected : 5
# actual   : 4
# --
# 
not ok 247 checking opendkim: generator creates keys, tables and TrustedHosts without postfix-virtual.cf
#  in test file test/tests.bats, line 615)
#   `assert_output 5' failed
# 
# -- output differs --
# expected : 5
# actual   : 4
# --
# 
not ok 248 checking opendkim: generator creates keys, tables and TrustedHosts using domain name
#  in test file test/tests.bats, line 646)
#   `assert_output 6' failed
# 
# -- output differs --
# expected : 6
# actual   : 4
# --
# 
```

</details>

Do you need help solving the issues?

williamdes:
> Do you need help solving the issues?

Yes, please I am low on time now to fix this
Any help would be awesome, I rebased my work

wernerfred:
I fixed the problem by removing the quotes from the argument as they are already set in function call ([in this line](https://github.com/docker-mailserver/docker-mailserver/blob/07b1d3b16d9b196b1984f02a2f291f435c3ef845/setup.sh#L326))

```GIT
diff --git a/target/bin/generate-dkim-config b/target/bin/generate-dkim-config
index 11b50dd..73ca42f 100755
--- a/target/bin/generate-dkim-config
+++ b/target/bin/generate-dkim-config
@@ -5,7 +5,7 @@ touch /tmp/vhost.tmp
 # if no keysize is provided, 2048 is default.
 KEYSIZE=${1:-2048}
 # optional domain names
-DOMAINS="${2:-''}"
+DOMAINS=${2:-''}
 
 if [[ -z "${DOMAINS}" ]]
 then
```

Next i wanted to add tests for this kind of functionality but then i discovered [a test that is checking dkim only by using domain name](https://github.com/docker-mailserver/docker-mailserver/blob/11bbda4adfe33a801241380093ffd33a89790fb5/test/tests.bats#L636) which was introduced 4 years ago by https://github.com/docker-mailserver/docker-mailserver/commit/984537526a7838710ac169456b0482471b3a2dd7

To be fair, this functionality isn't documented anywhere (got removed from README.md and is not in Wiki currently).

I propose that you take a closer look if `generate-dkim-domain` would fit your needs too and we then can decide if we want to abandon the "old" script (which is an additional file/command _and_ does not allow the setting of keysize _and_ is not documented _but_ is already tested) in favor of this PR (which would be in the same file/command _and_ does allow the setting of keysize _and_ is documented _but_ is not tested currently)

Either way would be a bit of work to invest. What is your opinion on that?

wernerfred:
I would like to decide this before 8.0.0 as removing a old way of generating the dkim keys might break existing setups @aendeavor

aendeavor:
Actually LGTM to me. @wernerfred is right, we should merge this before `8.0.0`. As I have merged #1770, this is the last step before the next major release.

This was labelled with `pr/missing integration tests`. @williamdes how fast can you write up two or three tests?

aendeavor:
I have added another commit that changes the behavior of the default value. I made this "mistake" earlier - using `${VAR:=}` is better than `${VAR:-''}`, since the new variable `VAR` have a completely empty value, instead of the string `''` :)

wernerfred:
> This was labelled with `pr/missing integration tests`. @williamdes how fast can you write up two or three tests?

Maybe I find some time today for writing the test (at least in the same "dirty" style the other dkim tests are)

But that means that you prefer to delete the `generate-dkim-domain` script inclusive the tests and instead use the functionality of this PR?

target/bin/generate-dkim-config
```diff
@@ -4,29 +4,36 @@ touch /tmp/vhost.tmp
 
 # if no keysize is provided, 2048 is default.
 KEYSIZE=${1:-2048}
+# optional domain names
+DOMAINS=${2:=}
```
wernerfred, line 8: Actually the tests are failing with:

`line 8: $2: cannot assign in this way`

wernerfred (changes requested):

williamdes:
Hi
Thanks for the comments and work
Actually I would let that up to you about writing tests because I am not working on my infrastructure this month

Let me know if you absolutely need something :)

I will be glad to review anyway

And dropping the old script looks like a good idea to me

target/bin/generate-dkim-config
wernerfred: Why did you change it to default variable assignment (`:=`) from the original default variable usage (`:-`)?
The result shouldn't be a different as we are _assigning_ the value anyways to `DOMAINS`

wernerfred: Take a look at [Bash Reference Manual - 3.5.3 Shell Parameter Expansion](https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Shell-Parameter-Expansion)
```
${parameter:-word}
If parameter is unset or null, the expansion of word is substituted. Otherwise, the value of parameter is substituted.

${parameter:=word}
If parameter is unset or null, the expansion of word is assigned to parameter. The value of parameter is then substituted. Positional parameters and special parameters may not be assigned to in this way.
```

Important is this one: **Positional parameters and special parameters may not be assigned to in this way.**

wernerfred: I will revert the change back to `:-` during test implementation

aendeavor: I wrote so many of them in `start-mailserver.sh` that I misinterpreted it here. In `start-mailserver.sh` assignments like this happen:

``` BASH
SOME_VAR="${SOME_VAR:=}"
```

which works obviously since it's the same variable. In this case, you're right, `:=` won't work since I was trying to give `$2` a new value. I was thinking about whether `:=` or `:-` but took the wrong one. Of course, `:-` is right. Thanks. As you said you'd revert this, I think this is resolved:)

aendeavor:
@wernerfred please by all accounts merge this yourself if you think this is ready to be merged. Reach back to me in @docker-mailserver/maintainers in the discussion for releasing `8.0.0`. We can then finalize the release.

wernerfred:
@williamdes pls test this asap! All further commits from my side will now be documentation and typo fixing etc.

wernerfred (approved):

aendeavor:
@williamdes just pinging you here:)

williamdes (commented):
The diff looks good to me
I already tested the change and it worked
Not testing the tests because the tests test :smile:
Thanks a lot!

One note: I would have used 4096 in tests rather than an old 2048 value

aendeavor:
@wernerfred if you'd like to bump to 4096 that's fine with me - just remember to adjust the test which checks the default value too.

wernerfred:
Good point as this is a breaking change, too.
Please merge #1771 first. I will implement the changes during the day.

aendeavor:
> Good point as this is a breaking change, too.
> 
> Please merge #1771 first. I will implement the changes during the day.

Tell me when you're ready. I will merge #1771 then if it has not been done until then.

aendeavor:
LGTM

ready for merging?

aendeavor (approved):

aendeavor:
30secs, nice :D

PR #1756, opened 2021-01-19
FIX: build inconsistencies for razor

aendeavor:


wernerfred (approved):
What an ugly workaround for such a simple line of code. Don't get me wrong I'm not complaining about your code - more than the fact that this is necessary :unamused:

LGTM tho as the action will hopefully pass now constantly (remains unclear if the fix will kick in or it will be just random again)

casperklein (approved):

casperklein:
@aendeavor 

Out of interest, did you try the variant with two RUN statements?

aendeavor:
> @aendeavor 
> 
> 
> 
> Out of interest, did you try the variant with two RUN statements?

I actually didn't. I was considering it, but now I'm not sure about this anymore, since the error we're seeing is most likely yet not related to the two command's sequence, but to the second command having a network error of some sort. This will ultimately not be fixed with different positions in the Dockerfile, but by retrying.

Does that sound logical? What's your take?

casperklein:
I didn't fully get what the root cause of the problem actually is. All my local builds were successful, even with the sleep command completely removed. I had never problems with the register command, so wondering why the GH CI fails.

Do we know, why `razor-admin -register` exactly fails? Is there some error output?
Typically a sleep is needed, when for example a daemon is started, that needs some time to startup and open a port. But that shouldn't be the case with `razor-admin -register`? At the moment, we are retrying 10 times in the hope it succeed. If we know the root cause why it fails, maybe we are able to fix it a better way, instead of just "dump" retrying.

The retry approach is ugly, but obviously needed for the Github CI, if we don't want to deal with failed builds from time to time. So for now, I am fine with this fix. 👍

aendeavor:
> Do we know, why `razor-admin -register` exactly fails? Is there some error output?

When you execute it locally _before_ `razor-admin -create` was called, you'll see an error reported with exit code 2 - I guess this is what we're seeing. However, retrying is in this case the best approach since it's a CI inconsistency:)

PR #1758, opened 2021-01-21
Improve logging significantly – color is back!

aendeavor:
This one is important to me. What happened:

1. I added an environment variable which enables users to set the log-level
2. I re-enabled **color** :D See yourself:

2021-01-21-094536_701x136_scrot

3. Enhanced some output that did not use `_notify` but just `echo`
4. Improved supervisor configuration
5. Got rid of (confusing) `CRIT` messages during startup

**Here is why you should like and approve it**:

---

First of all, the log right now is **all over the place**. See:

``` INI
2021-01-21 00:01:16,377 CRIT Supervisor is running as root.  Privileges were not dropped because no user is specified in the config file.  If you intend to run as root, you can set user=root in the config file to avoid this message.
2021-01-21 00:01:16,377 INFO Included extra file "/etc/supervisor/conf.d/saslauth.conf" during parsing
2021-01-21 00:01:16,377 INFO Included extra file "/etc/supervisor/conf.d/supervisor-app.conf" during parsing
2021-01-21 00:01:16,394 INFO RPC interface 'supervisor' initialized
2021-01-21 00:01:16,394 CRIT Server 'unix_http_server' running without any HTTP authentication checking
2021-01-21 00:01:16,394 INFO supervisord started with pid 8
2021-01-21 00:01:17,398 INFO spawned: 'mailserver' with pid 11
 
2021-01-21 00:01:17,444 INFO success: mailserver entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
[ TASKLOG ]  Welcome to docker-mailserver!
 
[ TASKLOG ]  Initializing setup
[ TASKLOG ]  Checking configuration
[ TASKLOG ]  Configuring mail server
Cert found in /etc/letsencrypt/acme.json for mail.Domain.com

[ TASKLOG ]  Startin misc
[ TASKLOG ]  Starting mail server
2021-01-21 00:01:21,967 INFO spawned: 'cron' with pid 264
2021-01-21 00:01:21,968 INFO success: cron entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
cron: started

```

Let's break this down. At the moment, there is no color in our logs. With this PR, this is not the case anymore. More importantly, there are two `CRIT` messages, which **absolutely should not be there**:

``` INI
2021-01-21 00:01:16,377 CRIT Supervisor is running as root.  Privileges were not dropped because no user is specified in the config file.  If you intend to run as root, you can set user=root in the config file to avoid this message.

2021-01-21 00:01:16,394 CRIT Server 'unix_http_server' running without any HTTP authentication checking
```

This was due to a sloppy supervisor configuration. This is now fixed. The configuration file has been cleaned up and improved.

Someone introduced a plain `echo` here:

``` INI
Cert found in /etc/letsencrypt/acme.json for mail.itbsd.com
```

This is now a `_notify`. (Note here that I had to adjust the unit tests to not check against our own logs. We never do that and we can't because of a limitation of ANSI-escape parsing. That is absolutely fine though.)

You can see that I changed the supervisor log-level to `warn`. Before you light this PR up, let me explain:) Currently, you will see this _next_ to our log:

``` INI
2021-01-21 00:01:16,377 INFO Included extra file "/etc/supervisor/conf.d/saslauth.conf" during parsing
2021-01-21 00:01:16,377 INFO Included extra file "/etc/supervisor/conf.d/supervisor-app.conf" during parsing
2021-01-21 00:01:16,394 INFO RPC interface 'supervisor' initialized
2021-01-21 00:01:16,394 INFO supervisord started with pid 8
2021-01-21 00:01:17,398 INFO spawned: 'mailserver' with pid 11
 
2021-01-21 00:01:17,444 INFO success: mailserver entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
```

This will now not be printed anymore. Is this a problem? I argue it's not: 

This will always work, because it is baked into the Dockerfile:

``` INI
2021-01-21 00:01:16,377 INFO Included extra file "/etc/supervisor/conf.d/saslauth.conf" during parsing
2021-01-21 00:01:16,377 INFO Included extra file "/etc/supervisor/conf.d/supervisor-app.conf" during parsing
```

If it's not, a warning is issued and you will see it. Then for

``` INI
2021-01-21 00:01:16,394 INFO RPC interface 'supervisor' initialized
```

it's the same line of reasoning. If you#re interested in the PIDs:

``` INI
2021-01-21 00:01:16,394 INFO supervisord started with pid 8
2021-01-21 00:01:17,398 INFO spawned: 'mailserver' with pid 11
```

run `docker exec mail ps`. The last message:

``` INI
2021-01-21 00:01:17,444 INFO success: mailserver entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
```

has the same line of reasoning as the first two arguments. Now, when `start-mailserver.sh` is running, the default log-level will be warn **if you want** (and by default). But now you have the choice. You could set it back to `info`.

---

I'd like to see this merged as fast as possible :D

### DIFF
--- a/target/supervisor/supervisord.conf
+++ b/target/supervisor/supervisord.conf
@@ -1,28 +1,31 @@
-; supervisor config file
-
 [unix_http_server]
-file=/dev/shm/supervisor.sock    ; (the path to the socket file)
-chmod=0700                       ; sockef file mode (default 0700)
+file = /dev/shm/supervisor.sock
+chmod = 0700
+chown = nobody:nogroup
+username = docker-mailserver
+password = docker-mailserver-password
 
 [supervisord]
-logfile=/var/log/supervisor/supervisord.log ; (main log file;default $CWD/supervisord.log)
-pidfile=/var/run/supervisord.pid ; (supervisord pidfile;default supervisord.pid)
-childlogdir=/var/log/supervisor            ; ('AUTO' child log dir, default $TEMP)
+user = root
+loglevel = warn
+nodaemon = true
+strip_ansi = false
+logfile = /var/log/supervisor/supervisord.log      ; default $CWD/supervisord.log
+pidfile = /var/run/supervisord.pid                 ; default supervisord.pid
+childlogdir = /var/log/supervisor                  ; default $TEMP ('AUTO' child log dir)
 
-; the below section must remain in the config file for RPC
-; (supervisorctl/web interface) to work, additional interfaces may be
-; added by defining them in separate rpcinterface: sections
+[supervisorctl]
+serverurl = unix:///dev/shm/supervisor.sock        ; use a 'unix://' path for a unix socket
+username = docker-mailserver
+password = docker-mailserver-password
+
+; must remain in config file for RPC (supervisorctl/web interface) to work, additional
+; interfaces may be added by defining them in separate rpcinterface: sections
 [rpcinterface:supervisor]
 supervisor.rpcinterface_factory = supervisor.rpcinterface:make_main_rpcinterface
 
-[supervisorctl]
-serverurl=unix:///dev/shm/supervisor.sock ; use a unix:// URL  for a unix socket
-
-; The [include] section can just contain the "files" setting.  This
-; setting can list multiple files (separated by whitespace or
-; newlines).  It can also contain wildcards.  The filenames are
-; interpreted as relative to this file.  Included files *cannot*
-; include files themselves.
-
+; The [include] section can just contain the "files" setting.  This setting can list multiple
+; files (separated by whitespace or newlines).  It can also contain wildcards.  The filenames
+; are interpreted as relative to this file.  Included files *cannot* include files themselves.
 [include]
 files = /etc/supervisor/conf.d/*.conf
--- a/ENVIRONMENT.md
+++ b/ENVIRONMENT.md
@@ -15,6 +15,18 @@
 - **0** => Debug disabled
 - 1 => Enables debug on startup
 
+##### SUPERVISOR_LOGLEVEL
+
+Here you can adjust the [log-level for Supervisor](http://supervisord.org/logging.html#activity-log-levels). Possible values are
+
+- critical => Only show critical messages
+- error => Only show erroneous output
+- warn => Show warnings
+- **info** => Normal informational output
+- debug => Also show debug messages
+
+The log-level will show everything in its class and above.
+
 ##### ENABLE_CLAMAV
 
 - **0** => Clamav is disabled
--- a/ENVIRONMENT.md
+++ b/ENVIRONMENT.md
@@ -21,8 +21,8 @@ Here you can adjust the [log-level for Supervisor](http://supervisord.org/loggin
 
 - critical => Only show critical messages
 - error => Only show erroneous output
-- warn => Show warnings
-- **info** => Normal informational output
+- **warn** => Show warnings
+- info => Normal informational output
 - debug => Also show debug messages
 
 The log-level will show everything in its class and above.
--- a/target/scripts/helper-functions.sh
+++ b/target/scripts/helper-functions.sh
@@ -97,7 +97,7 @@ for key, value in acme.items():
 
     echo "${KEY}" | base64 -d >/etc/letsencrypt/live/"${HOSTNAME}"/key.pem || exit 1
     echo "${CERT}" | base64 -d >/etc/letsencrypt/live/"${HOSTNAME}"/fullchain.pem || exit 1
-    echo "Cert found in /etc/letsencrypt/acme.json for ${1}"
+    _notify 'inf' "Cert found in /etc/letsencrypt/acme.json for ${1}"
 
     return 0
   else
--- a/target/scripts/helper-functions.sh
+++ b/target/scripts/helper-functions.sh
@@ -110,22 +110,17 @@ export -f _extract_certs_from_acme
 
 function _notify
 {
-  local FINAL_MSG=''
-  local MSG="${2:-}"
-  local TYPE="${1:-}"
+  { [[ -z ${1:-} ]] || [[ -z ${2:-} ]] ; } && return
 
-  case "${TYPE}" in
-    'none'    ) FINAL_MSG=' ' ;;
-    'tasklog' ) FINAL_MSG="[ \e[0;92mTASKLOG\e[0m ]  ${MSG}" ;; # light green
-    'warn'    ) FINAL_MSG="[ \e[0;93mWARNING\e[0m ]  ${MSG}" ;; # light yellow
-    'err'     ) FINAL_MSG="[  \e[0;31mERROR\e[0m  ]  ${MSG}" ;; # light red
-    'fatal'   ) FINAL_MSG="[  \e[0;91mFATAL\e[0m  ]  ${MSG}" ;; # red
-    'inf'     ) [[ ${DMS_DEBUG} -eq 1 ]] && FINAL_MSG="[[ \e[0;34mINFO\e[0m ]]   ${MSG}" ;; # light blue
-    'task'    ) [[ ${DMS_DEBUG} -eq 1 ]] && FINAL_MSG="[[ \e[0;37mTASK\e[0m ]]   ${MSG}" ;; # light grey
-    *         ) ;;
+  case ${1} in
+    tasklog  ) echo "-e${3:-}" "[ \e[0;92mTASKLOG\e[0m ]  ${2}" ;; # light green
+    warn     ) echo "-e${3:-}" "[ \e[0;93mWARNING\e[0m ]  ${2}" ;; # light yellow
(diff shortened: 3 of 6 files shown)


aendeavor:
Just FYI: This is what is will look like with log-level kept to 'warn':

new_logs

Please note that I used `_notify` in `user-patches.sh` too and I changed Amavis' log level in the `user-patches.sh` - that's why you're not seeing it here.

target/supervisor/supervisord.conf
```diff
@@ -1,28 +1,31 @@
-; supervisor config file
-
 [unix_http_server]
-file=/dev/shm/supervisor.sock    ; (the path to the socket file)
-chmod=0700                       ; sockef file mode (default 0700)
+file = /dev/shm/supervisor.sock
+chmod = 0700
+chown = nobody:nogroup
+username = docker-mailserver
+password = docker-mailserver-password
```
wernerfred, line 6: Where do these values originate from?

target/supervisor/supervisord.conf
```diff
@@ -1,28 +1,31 @@
...
+strip_ansi = false
+logfile = /var/log/supervisor/supervisord.log      ; default $CWD/supervisord.log
+pidfile = /var/run/supervisord.pid                 ; default supervisord.pid
+childlogdir = /var/log/supervisor                  ; default $TEMP ('AUTO' child log dir)
 
 [supervisorctl]
-serverurl=unix:///dev/shm/supervisor.sock ; use a unix:// URL  for a unix socket
+serverurl = unix:///dev/shm/supervisor.sock        ; use a 'unix://' path for a unix socket
+username = docker-mailserver
+password = docker-mailserver-password
```
wernerfred, line 20: Same as above

wernerfred (approved):
LGTM 👍🏻 

I personally like the changes you introduced to color up the output and allow setting the supervisor loglevel.

Regarding "to not break backwards compatability" my opinion (apart from this PR/change but in general) is that major releases are justified by breaking changes so if you/we think keeping the default by "warn" from now on then there is no better timing as now as we are close to 8.0.0.

target/supervisor/supervisord.conf
aendeavor: I just came up with them. They could be anything, it actually doesn't matter, because they only exist to make the `CRIT`disappear.

target/supervisor/supervisord.conf
aendeavor: They have to equal the values from above - like a login.

aendeavor:
> 
> 
> LGTM 👍🏻
> 
> I personally like the changes you introduced to color up the output and allow setting the supervisor loglevel.
> 
> Regarding "to not break backwards compatability" my opinion (apart from this PR/change but in general) is that major releases are justified by breaking changes so if you/we think keeping the default by "warn" from now on then there is no better timing as now as we are close to 8.0.0.

I like this idea. I will adjust this PR to use `warn` by default.

casperklein:
> Now, when start-mailserver.sh is running, the default log-level will be warn if you want (and by default). But now you have the choice. You could set it back to info

Doesn't work for me. I've set it to `info`, but don't get any `info` logs. For example on container stop, there should be stuff like:

```
2021-02-01 00:40:02,464 WARN received SIGTERM indicating exit request
2021-02-01 00:40:02,464 INFO waiting for rsyslog, postfix, cron, amavis, changedetector, fail2ban, opendkim, opendmarc, dovecot, mailserver to die
2021-02-01 00:40:02,464 INFO stopped: mailserver (terminated by SIGTERM)
```

Can someone confirm this?

aendeavor:
> > Now, when start-mailserver.sh is running, the default log-level will be warn if you want (and by default). But now you have the choice. You could set it back to info
> 
> 
> 
> Doesn't work for me. I've set it to `info`, but don't get any `info` logs. For example on container stop, there should be stuff like:
> 
> 
> 
> ```
> 
> 2021-02-01 00:40:02,464 WARN received SIGTERM indicating exit request
> 
> 2021-02-01 00:40:02,464 INFO waiting for rsyslog, postfix, cron, amavis, changedetector, fail2ban, opendkim, opendmarc, dovecot, mailserver to die
> 
> 2021-02-01 00:40:02,464 INFO stopped: mailserver (terminated by SIGTERM)
> 
> ```
> 
> 
> 
> Can someone confirm this?

Do you see at least the `WARN` message?

casperklein:
Yes. The `warn` message is always there. And the modification to supervisord.conf seems to be successful:


```
root@mail:/# grep loglevel /etc/supervisor/supervisord.conf
loglevel = info
```

casperklein:
>Can you then set SUPERVISOR_LOGLEVEL to something ridiculous

That leads to this startup warning as expected:

`[ WARNING ]  SUPERVISOR_LOGLEVEL value 'aendeavor-told-me-so' unknown. Defaulting to 'warn'`

aendeavor:
> >Can you then set SUPERVISOR_LOGLEVEL to something ridiculous
> 
> 
> 
> That leads to this startup warning as expected:
> 
> 
> 
> `[ WARNING ]  SUPERVISOR_LOGLEVEL value 'aendeavor-told-me-so' unknown. Defaulting to 'warn'`
> 
> 

Good. Then I suspect supervisord needs more than an update? Maybe it needs to be restarted too?

casperklein:
The Problem is, that supervisord does not seem to recognize the change.

Additional to `supervisorctl update`, I've also tried this without success: `supervisorctl reread`

aendeavor:
> The Problem is, that supervisord does not seem to recognize the change.
> 
> 
> 
> Additional to `supervisorctl update`, I've also tried this without success: `supervisorctl reread`

Is there something like a self-restart command? I investigate too, but I'll do this in a few hours:)

casperklein:
> Maybe it needs to be restarted too?
I am not sure, but I don't think that is possible, as supervisor is our PID1. Killing it will stop the container.

aendeavor:
> > Maybe it needs to be restarted too?
> 
> 
> 
> I am not sure, but I don't think that is possible, as supervisor is our PID1. Killing it will stop the container.

No, dumb-init is - you created the PR yourself :D

aendeavor:
Oh my - `supervisorctl reload` seems to have worked...

I will provide a PR.

casperklein:
>No, dumb-init is - you created the PR yourself :D

Yes, I know. Take a look at this 😉 

```
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root         1  0.0  0.0   2144   740 ?        Ss   11:39   0:00 /usr/bin/dumb-init -- supervisord -c /etc/supervisor/supervisord.conf
```

casperklein:
>> the default log-level will be warn if you want (and by default). But now you have the choice. You could set it back to info.

BTW Thats not 100% true. It's not fully possible to get the old behavior back. INFO messages that would appear before `start-mailserver.sh` modifies supervisord.conf are lost.

aendeavor:
> > the default log-level will be warn if you want (and by default). But now you have the choice. You could set it back to info.
> 
> BTW Thats not 100% true. It's not fully possible to get the old behavior back. [...]

I made this point clear from the start:

> You can see that I changed the supervisor log-level to warn. Before you light this PR up, let me explain:) Currently, you will see this next to our log:
>
> 2021-01-21 00:01:16,377 INFO Included extra file "/etc/supervisor/conf.d/saslauth.conf" during parsing
> 2021-01-21 00:01:16,377 INFO Included extra file "/etc/supervisor/conf.d/supervisor-app.conf" during parsing
> 2021-01-21 00:01:16,394 INFO RPC interface 'supervisor' initialized
> 2021-01-21 00:01:16,394 INFO supervisord started with pid 8
> 2021-01-21 00:01:17,398 INFO spawned: 'mailserver' with pid 11
>
> 2021-01-21 00:01:17,444 INFO success: mailserver entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
>
> _This will now not be printed anymore_. Is this a problem? I argue it's not...

Only these 5 INFO messages are "lost", and I believe this to be fine.

#1778 should take care of the rest.

casperklein:
Sorry for not being more precise. I had read it and it's fine. But I thought that should be mentioned in the README?

When someone sets a log level, then the assumption probably is, that no messages are suppressed in that log level. For example you set it to `debug` for some reason, you could be wondering why there a no debug message from the startup until start-mailserver is invoked.

aendeavor:
> Sorry for not being more precise. I had read it and it's fine. But I thought that should be mentioned in the README?
> 
> 
> 
> When someone sets a log level, then the assumption probably is, that no messages are suppressed in that log level. For example you set it to `debug` for some reason, you could be wondering why there a no debug message from the startup until start-mailserver is invoked.

We could add this to `mailserver.env`, if you want. I think it's not that important, since this project is not debugging supervisord, and the messages are only for PIDs and loaded conf files, which cannot be altered before anyway:)

Issue #1759, opened 2021-01-21
[FR] Push docker images to Gihub Container Registry


joao-p-marques:
# Feature Request

Would it be possible to push the Docker image also to the [Gihub Container Registry](https://docs.github.com/en/packages/guides/about-github-container-registry) (ghcr.io)?

## Context

Images are only on DockerHub,.

### Is your Feature Request related to a Problem?

It would help avoid the new pull limits and some possible network restrictions from DockerHub.

### Describe the Solution you'd like

It can be easily done with the Github Action you already have to push to DockerHub. 

### Are you going to implement it?

Yes, because I know the probability of someone else doing it is low and I can learn from it.

### What are you going to contribute??

An update to the Github action to implement this. Some secrets set up would also be needed from the maintainers.

### Who will that Feature be useful to?

People hitting the DockerHub limits or behind a network that restricts access to it, and not to GHCR.

### What have you done already?

I have applied this approach in some other projects.

cc @Yajo 
@Tecnativa TT27819

aendeavor:
What do you think @radicand?

I've got to admit, not sure about this one. I see little use in pushing to two different registries to be honest. Firstly, why wouldn't you be able to pull from docker.io when you're able to pull from ghcr.io. Secondly, the pull limit can be avoided with local caching registries.

However, if it's easy to implement, and does not increase complexity in the workflow, I guess there's not reason not to do it either.

---

PS: Aren't you guys the maintainers of tecnativa/docker-socket-proxy? What about tecnativa/docker-socket-proxy#33? I'm really tempted to couple this to the release of versioned tags 😆 😆 😆  Of course, I'm not going to do that, but seeing some tags would be really appreciated :D

Yajo:
Oops we can't hide anymore! 🙈 😆 



> why wouldn't you be able to pull from docker.io when you're able to pull from ghcr.io

Docker Hub bans some countries, as you see in https://github.com/docker/hub-feedback/issues/369#issuecomment-214540296. Github doesn't: https://docs.github.com/en/github/site-policy/github-and-trade-controls#what-is-available-and-not-available.

wernerfred:
[...]

PR #1762, opened 2021-01-22
Apply image template

joao-p-marques:
Closes https://github.com/docker-mailserver/docker-mailserver/issues/1759

This applies https://github.com/Tecnativa/image-template to this repo. It sets up the CI workflow to build, test and push the images to the Github Container Registry.

**A few changes and comments about how this affects your current flow:**
1. I've tried to merge the whole functionality implemented by the `default_on_push`, `scheduled_builds` and `test_merge_requests` files into the new `ci` file that will apply to **PRs**, push to **master and stable branches**, **scheduled** and **manual** actions. The workflows were basically similar, and I tried to merge the different cases on the same workflow through some conditionals. However, the new workflow won't affect your current ones, as **only the funcionality of pushing to GHCR** is enabled. If, in the future, you would like to remove the other actions and merge everything into this new flow, a simple update from the template is enough (I explain more bellow)
2. The tagging mechanism is slightly different, as mentioned in https://github.com/docker-mailserver/docker-mailserver/issues/1759#issuecomment-765307941. This allows us to generate the docker tags with the https://github.com/crazy-max/ghaction-docker-meta/ action and transverse some logic from the git versioning to docker. It basically tags `:latest` as the latest *released* version in git if a git tag is being pushed, `:edge` as the version that is in *the repo's default* branch, and then each individual git released version with `:{{version}}`, `:{{major}}` and `:{{major}}.{{minor}}` according to the context of the current build. This basically means that:
2.1. In a PR, it will generate a tag of the type `pr-X`
2.2. Once the PR arrives at the default branch, it will push a new `:edge` image and, if a git tag is being applied, it will also push new `:latest` and `:vX-X-X` images, again, according to the templates mentioned above.
3. The images are always built in all the available architectures (linux/amd64, linux/arm/v7 and linux/arm64) even for testing. IMHO, this covers a broader spectrum when you are testing (for example, in a PR), helping to identify potential issues. If we are on a master branch or on a scheduled action (where the goal is to push), the cache is used and the build is very quick, so it won't take much longer.
4. As the images are always built with the same tagging system, it no longer makes sense to tag a `:ci` image and, instead, use the specific image tag generated by the docker meta action (See https://github.com/crazy-max/ghaction-docker-meta/#basic)

As this just adds the new push to GHCR feature in a separate workflow, you can try this configurations there without affecting your DockerHub images and push flow. At the same time, you can try the `image-template` mentioned above which makes it easier to maintain these configurations.

**Some notes about the template:**
https://github.com/Tecnativa/image-template is a [copier](https://github.com/copier-org/copier) template. For updating, check the README on the template.
If you want to change some variables in the template itself, you can do that by simply running a normal update (`copier update`) and changing your answers to some questions.
So, for example, if you end up concentrating all the GH Actions on the `ci.yml` file generated by this template, you'd simply change your answer to the `dockerhub_image` question during the update.
If you have more questions about it, feel free to ask. :smiley: 

@Tecnativa
TT27819

ping @Yajo @aendeavor @wernerfred

### DIFF
--- a/.copier-answers.image-template.yml
+++ b/.copier-answers.image-template.yml
@@ -1,0 +1,17 @@
+# Changes here will be overwritten by Copier; do NOT edit manually
+_commit: v0.1.0
+_src_path: https://github.com/Tecnativa/image-template.git
+dockerhub_image: ''
+image_platforms:
+- linux/amd64
+- linux/arm/v7
+- linux/arm64
+main_branches:
+- stable
+- master
+project_name: docker-mailserver
+project_owner: docker-mailserver
+push_to_ghcr: true
+pytest: false
+python_versions:
+- '3.9'
--- a/README.md
+++ b/README.md
@@ -1,3 +1,6 @@
+[![Last image-template](https://img.shields.io/badge/last%20template%20update-v0.1.0-informational)](https://github.com/Tecnativa/image-template/tree/v0.1.0)
+[![GitHub Container Registry](https://img.shields.io/badge/GitHub%20Container%20Registry-latest-%2324292e)](https://github.com/orgs/docker-mailserver/packages/container/package/docker-mailserver)
+
 # Docker Mailserver
 
 [![ci::status]][ci::github] [![docker::pulls]][docker::hub]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,124 @@
+
+name: test
+
+"on":
+  pull_request:
+  push:
+    branches:
+      - stable
+      - master
+  schedule:
+    - cron: "0 0 * * 5"
+  workflow_dispatch:
+
+env:
+  LANG: "en_US.utf-8"
+  LC_ALL: "en_US.utf-8"
+
+jobs:
+    build-push:
+      runs-on: ubuntu-20.04
+      services:
+        registry:
+          image: registry:2
+          ports:
+            - 5000:5000
+      env:
+        DOCKER_IMAGE_NAME: ${{ github.repository }}
+        PUSH: ${{ toJSON(github.event_name != 'pull_request') }}
+      steps:
+        # Set up Docker Environment
+        - uses: actions/checkout@v2
+          if: github.event_name != 'schedule'
+          with:
+            submodules: recursive
+        # HACK: scheduled actions run on the default branch and GH Actions doesn't allow a conditional "ref" assignment
+        - uses: actions/checkout@v2
+          if: github.event_name == 'schedule'
+          with:
+            submodules: recursive
+            ref: stable
+        - uses: actions/cache@v2
+          with:
+            path: |
+              /tmp/.buildx-cache
+            key:
+              buildx|${{ secrets.CACHE_DATE }}|${{ runner.os }}|${{
+              hashFiles('Dockerfile') }}
+        - name: Set up QEMU
+          uses: docker/setup-qemu-action@v1
+        - name: Set up Docker Buildx
+          id: buildx
+          uses: docker/setup-buildx-action@v1
+          with:
+            driver-opts: network=host
+            install: true
+        # Build and push
+        - name: Docker meta for local images
+          id: docker_meta_local
+          uses: crazy-max/ghaction-docker-meta@v1
+          with:
+            images: localhost:5000/${{ env.DOCKER_IMAGE_NAME }}
+            tag-edge: true
+            tag-semver: |
+              {{version}}
+              {{major}}
+              {{major}}.{{minor}}
+        - name: Build and push to local (test) registry
+          uses: docker/build-push-action@v2
+          with:
+            context: .
+            file: ./Dockerfile
+            platforms: |
+              linux/amd64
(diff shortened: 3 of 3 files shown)


README.md
```diff
@@ -1,3 +1,6 @@
+[![Last image-template](https://img.shields.io/badge/last%20template%20update-v0.1.0-informational)](https://github.com/Tecnativa/image-template/tree/v0.1.0)
```
aendeavor, line 1: I'm sorry to interrupt this so early, but the README modifiation here is from my PoV unacceptable. If you'd like to link to the package, please do it somewhere further down below, and especially _below_ the heading. Please do not use bades for this as well, as the badges at the top shall remain the only ones. The miscellaneous section is in my eyes appropriate,

.copier-answers.image-template.yml
```diff
@@ -0,0 +1,17 @@
+# Changes here will be overwritten by Copier; do NOT edit manually
```
aendeavor, line 1: I have put a lot of effort in this project to clean the directories from unnecessary files, especially the root directory. Keeping this project clean (i.e. having only needed files) I think is a top priority. I'm being blunt here, sorry, but I cannot accept this as I think this addition is unnecessary and I don't want to see the root directory cluttered with files I deem inappropriate or unnecessary.

aendeavor (changes requested):
I have read what you're trying to achieve. I really respect the work you did, but I've got to admit that I do not really like the idea of it. From the linked issues, my understanding was a _simple_ workflow addition _that does not incorporate complexity_.

This PR even introduces a template. Something I deem absolutely not necessary for what we wanted to achieve. The CI integration that is currently present shall not be changed soon - it has just been integrated and works well. The name of the file you created (`ci`) suggests this is a _replacement_, which you also stated. I can only talk for myself here, but I wanted to see an addition, especially an easy one, and not a complete, full-featured replacement with a template I don't understand. And I actually don't want to dig into it either.

I cannot approve this as is. Again, I don't want to be disrespectful, but merging this as is, from my PoV, is out of question.

radicand:
Having briefly reviewed the proposed changes, I would also agree - based on what the goal is, the simplest path of adjustment would be to amend the existing prep/TAGS stage to add the ghcr.io target(s), and ensure there is also a step for login the GHCR in addition to Dockerhub so both can be pushed. There shouldn't need to be any special copy/rewrite of the CI going on to accomplish this.

aendeavor:
> Having briefly reviewed the proposed changes, I would also agree - based on what the goal is, the simplest path of adjustment would be to amend the existing prep/TAGS stage to add the ghcr.io target(s), and ensure there is also a step for login the GHCR in addition to Dockerhub so both can be pushed. There shouldn't need to be any special copy/rewrite of the CI going on to accomplish this.

Completely agree.

wernerfred:
> I have read what you're trying to achieve. I really respect the work you did, but I've got to admit that I do not really like the idea of it. From the linked issues, my understanding was a simple workflow addition that does not incorporate complexity.

Agree 100%

> This PR even introduces a template. Something I deem absolutely not necessary for what we wanted to achieve.

This is the main issue i think. I like the idea of this template and I see a huge benefit for other projects and the community as well. But looking at the mailserver project, how it started and how it envolved this is a change that doesn't fit in the design and simplicity. It is _simply_ not necessary to bring in further dependencies especially if the result can be achieved so much easier.

Nevertheless thumbs up for developing this template in such a short time and thinking of how other could benefit and easy integrate your project even if they don't completely understand how it works. Additionally thank your for providing such a detailed description on the PR. I really appreciate this.

> I cannot approve this as is. Again, I don't want to be disrespectful, but merging this as is, from my PoV, is out of question.

I would come up with a similar conclusion.

aendeavor:
> I like the idea of this template and I see a huge benefit for other projects and the community as well. But looking at the mailserver project, how it started and how it envolved this is a change that doesn't fit in the design and simplicity. It is _simply_ not necessary to bring in further dependencies especially if the result can be achieved so much easier.

That sums up it quite nicely:)

One of the core concepts of this project is staying as simplistic as possible while providing everything that is needed. We as maintainers try to keep this project true to this maxim:)

aendeavor:
Closed per replacement in #1763

joao-p-marques:
Hi! Thanks for the comments and reviews. 

In essence, our end goal was to, indeed, have this image available in GHCR as well as DockerHub, and that can be done with a simple change in the workflow files, as you have done in https://github.com/docker-mailserver/docker-mailserver/pull/1763

The introduction of the template comes from the fact that we have a lot of docker projects to manage and needed a way to centralize and automate that. Even though that template came from our own needs, we tried to make it useful for everyone, and I thought it could be useful here.
It's true that the changes applied by this PR seem a bit big in comparison to your current spec, but my idea was to simply introduce a potentially easier way for you to manage the workflow. Obviously, that is not necessary and, if you think that is against your view for the project, that's fine by me :smile: 

Thanks also for the follow-up with the other PR!

wernerfred:
The `:edge` version is already available in ghcr.
Rest will follow subsequently with new releases

PR #1763, opened 2021-01-23
Change tag convention and allow multiple registries

### LINKED ISSUE #1759: [FR] Push docker images to Gihub Container Registry

wernerfred:
Closes #1759, replaces #1762 and introduces the new tagging convention of the docker images.

As discussed by the project maintainer team this PR brings in the new tagging convention:
| Event           | Ref                                       | Commit SHA | Docker Tags                         |
|--------------|-------------------------------|----------------|-------------------------------|
| `push`          | `refs/heads/master`             | `cf20257`       | `edge`                    |
| `push`          | `refs/heads/stable`              | `cf20257`        | `stable`                                |
| `push tag`    | `refs/tags/1.2.3`                 | `ad132f5`       | `1.2.3`, `1.2`, `1`, `latest`        |
| `push tag`    | `refs/tags/v1.2.3`                 | `ad132f5`       | `1.2.3`, `1.2`, `1`, `latest`        |

Furthermore it is now possible to push to multiple registries **only by adding them to the `image` section** (only isn't completely true as one needs to add a login step to the additional registry but **no** changes to the logic is needed) of the preparation job:
```
images: |
  docker.io/mailserver/docker-mailserver
  ghcr.io/docker-mailserver/docker-mailserver
```
These values can of course again be transformed into github secrets.

An additional secret/PAT (`${{ secrets.GHCR_PASSWORD }}`) for the login to ghcr is needed with the permissions:
- `read:packages`
- `write:packages`
- `delete:packages`

Make sure to deselect the `repo` scope when creating the PAT (according to [the security guideline for github actions](https://docs.github.com/en/packages/guides/pushing-and-pulling-docker-images). @aendeavor that might be your task).

Furthermore (thanks @radicand) we need `${{ secrets.GHCR_USERNAME }}` set appropriate (see [comment](https://github.com/docker-mailserver/docker-mailserver/pull/1763#discussion_r563211955)).

To be able to push to the github container registry @aendeavor (someone else hast access too?) will need to enable the [improved container support](https://docs.github.com/en/packages/guides/enabling-improved-container-support) on the organization level.

Additionally during my test i discovered that the current triggers didn't run the worklflow (at least when using the new action) when tags were pushed so i added the appropriate trigger matching only semver tags:
```
tags:
    - '*.*.*'
``` 

~~One question to @radicand:~~
~~Are the following lines needed:~~
<details>
<summary>Code</summary>

```
build-args: |
  VCS_REF=${{ github.sha }}
  VCS_VER=${{ github.ref }}
```
</details>

~~On the first sight I did not get the point of these vars but I might have overlooked something~~

__________

Checklist/Requirements
- [x] Add `${{ secrets.GHCR_USERNAME }}`
- [x] Add `${{ secrets.GHCR_PASSWORD }}`
- [x] Enable improved container support for organization

Optional:
- [x] Add  `${{ secrets.GHCR_REPOSITORY }}`
- [x] Add  `${{ secrets.DOCKER_REPOSITORY }}` // edit by @aendeavor,changed `DOCKERHUB_...` to `DOCKER_...`

### DIFF
--- a/.github/workflows/default_on_push.yml
+++ b/.github/workflows/default_on_push.yml
@@ -59,8 +59,8 @@ jobs:
         uses: crazy-max/ghaction-docker-meta@v1
         with:
           images: |
-            docker.io/mailserver/docker-mailserver
-            ghcr.io/docker-mailserver/docker-mailserver
+            ${{ secrets.DOCKER_REPOSITORY }}
+            ${{ secrets.GHCR_REPOSITORY }}
           tag-edge: true
           tag-semver: |
             {{major}}
--- a/.github/workflows/default_on_push.yml
+++ b/.github/workflows/default_on_push.yml
@@ -2,7 +2,11 @@ name: "Build, Test & Deploy"
 
 on:
   push:
-    branches: [ "master", "stable" ]
+    branches:
+      - master
+      - stable
+    tags:
+      - '*.*.*'
 
 jobs:
   build-and-test-image:
@@ -52,14 +56,16 @@ jobs:
           submodules: recursive
       - name: Prepare tags
         id: prep
-        run: |
-          DOCKER_IMAGE=${{ secrets.DOCKER_REPOSITORY }}
-          VERSION=latest
-          [[ $GITHUB_REF == refs/tags/* ]] && VERSION=${GITHUB_REF#refs/tags/v}
-          [[ $GITHUB_REF == 'refs/heads/stable' ]] && VERSION=stable
-          TAGS="${DOCKER_IMAGE}:${VERSION}"
-          [[ $VERSION =~ ^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}$ ]] && TAGS="$TAGS,${DOCKER_IMAGE}:latest"
-          echo ::set-output name=tags::${TAGS}
+        uses: crazy-max/ghaction-docker-meta@v1
+        with:
+          images: |
+            docker.io/mailserver/docker-mailserver
+            ghcr.io/docker-mailserver/docker-mailserver
+          tag-edge: true
+          tag-semver: |
+            {{major}}
+            {{major}}.{{minor}}
+            {{major}}.{{minor}}.{{patch}}
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v1
       - name: Set up Docker Buildx
@@ -77,6 +83,12 @@ jobs:
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GHCR_PASSWORD }}
       - name: Build image locally
         uses: docker/build-push-action@v2
         with:
--- a/.github/workflows/default_on_push.yml
+++ b/.github/workflows/default_on_push.yml
@@ -87,7 +87,7 @@ jobs:
         uses: docker/login-action@v1
         with:
           registry: ghcr.io
-          username: ${{ github.repository_owner }}
+          username: ${{ secrets.GHCR_USERNAME }}
           password: ${{ secrets.GHCR_PASSWORD }}
       - name: Build image locally
         uses: docker/build-push-action@v2
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,7 +34,7 @@ The development workflow is the following:
 7. [Commit][commit] and [sign your commit][gpg], push and create a pull-request to merge into `master`
    1. Pull requests are automatically tested against the CI and will be reviewed when tests pass
    2. When your changes are validated, your branch is merged
-   3. CI builds the new `:latest` image
+   3. CI builds the new `:edge` image immediately and your changes will be includes in the next version release.
 
 ## Coding Style
 
--- a/README.md
+++ b/README.md
@@ -61,6 +61,21 @@ A fullstack but simple mail server (SMTP, IMAP, LDAP, Antispam, Antivirus, etc.)
 
 ## Usage
 
+### Available image sources / tags
+
+The [CI/CD workflows](https://github.com/docker-mailserver/docker-mailserver/actions) automatically build, test and push new images to container registries. Currently, the following registries are supported:
+- [DockerHub](https://hub.docker.com/repository/docker/mailserver/docker-mailserver)
+- [GitHub Container Registry](https://github.com/orgs/docker-mailserver/packages?repo_name=docker-mailserver)
+
+All workflows are using the **tagging convention** listed below. It is subsequently applied to all images pushed to supported container registries:
+
+| Event        | Ref                   | Commit SHA | Image Tags                    |
+|--------------|-----------------------|------------|-------------------------------|
+| `push`       | `refs/heads/master`   | `cf20257`  | `edge`                        |
+| `push`       | `refs/heads/stable`   | `cf20257`  | `stable`                      |
+| `push tag`   | `refs/tags/1.2.3`     | `ad132f5`  | `1.2.3`, `1.2`, `1`, `latest` |
+| `push tag`   | `refs/tags/v1.2.3`    | `ad132f5`  | `1.2.3`, `1.2`, `1`, `latest` |
+
 ### Get the tools
 
 Download the `docker-compose.yml`, `compose.env`, `mailserver.env` and the `setup.sh` files:


aendeavor:
> ```
> 
> images: |
> 
>   docker.io/mailserver/docker-mailserver
> 
>   ghcr.io/docker-mailserver/docker-mailserver
> 
> ```
> 
> These values can of course again be transformed into github secrets.
> 
> 

I will create these secrets.

> An additional secret/PAT (`${{ secrets.GHCR_PASSWORD }}`) for the login to ghcr is needed with the permissions:
> 
> - `read:packages`
> 
> - `write:packages`
> 
> - `delete:packages`
> 
> 
> 
> Make sure to deselect the `repo` scope when creating the PAT (according to [the security guideline for github actions](https://docs.github.com/en/packages/guides/pushing-and-pulling-docker-images). @aendeavor that might be your task.
> 

And I will see whether I can take care of this.

> One question to @radicand:
> 
> Are the following lines needed:
> 
> ```
> 
> build-args: |
> 
>   VCS_REF=${{ github.sha }}
> 
>   VCS_VER=${{ github.ref }}
> 
> ```
> 
> On the first sight I did not get the point of these vars but I might have overlooked something
> 
> 
 
I might as well answer:) This is needed. Have a look at the `Dockerfile`, these are build arguments passed at build time:) They are, indeed, needed.

.github/workflows/default_on_push.yml
```diff
@@ -77,6 +83,12 @@ jobs:
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
```
radicand, line 90: I've been doing some similar tests as well (unsuccessfully) - in your testing, did this work end-to-end and result in the image going into GHCR? When I attempt this, it authenticates successfully at login, but on push to registry it gets a 403. I've given the PAT in my tests the read/write:packages scope per docs. Curious if you've had this work in practice yet.

Additionally, the username line will need to change to a secret for the user whose PAT we use for this. This variable as-is works fine for user-owned repos, but for org-repos, you'll be authenticating with someone's specific PAT, and you need that username in the login.

radicand (changes requested):

radicand:
> > Make sure to deselect the `repo` scope when creating the PAT (according to [the security guideline for github actions](https://docs.github.com/en/packages/guides/pushing-and-pulling-docker-images). @aendeavor that might be your task.

Note there's a bug in Github that automatically selects the repo scope without opportunity to de-select it when you add the package scope. The workaround (from Github) is to edit the PAT after you create it, and you can remove repo scope then.

.github/workflows/default_on_push.yml
wernerfred: Yes, running into the same issues first but was able to solve it by enabling the improved container support for my personal account. This would be needed for the organization as well ([docs](https://docs.github.com/en/packages/guides/enabling-improved-container-support)). I will add it to the initial description as it provides the whole overview.

Here is the output of my private test:
```
#6 exporting to image
#6 exporting layers
#6 exporting layers 0.2s done
#6 exporting manifest sha256:42daa0610f0df6ad71fa4288d0335083e7958116fdc31a8f51102272cb3113f9 done
#6 exporting config sha256:ae1f6955dbdff627e8eebebccc8704f08d90bb3364d24f083f8e4613c54d5166 done
#6 pushing layers
#6 pushing manifest for docker.io/wernerfred/testing:edge
#6 pushing manifest for docker.io/wernerfred/testing:edge 0.1s done
#6 pushing layers 1.5s done
#6 pushing manifest for ghcr.io/wernerfred/testing:edge
#6 pushing manifest for ghcr.io/wernerfred/testing:edge 0.5s done
#6 DONE 3.0s
```

> Additionally, the username line will need to change to a secret for the user whose PAT we use for this. This variable as-is works fine for user-owned repos, but for org-repos, you'll be authenticating with someone's specific PAT, and you need that username in the login.

Thanks for that hint I will add `${{ secrets.GHCR_USERNAME }}`

wernerfred: Solved with https://github.com/docker-mailserver/docker-mailserver/pull/1763/commits/cb8034262a981d3a197fa467dcdc7fe5c4327262

wernerfred:
Updated [top description](https://github.com/docker-mailserver/docker-mailserver/pull/1763#issue-560489939)

.github/workflows/default_on_push.yml
```diff
@@ -52,14 +56,16 @@ jobs:
...
-          VERSION=latest
-          [[ $GITHUB_REF == refs/tags/* ]] && VERSION=${GITHUB_REF#refs/tags/v}
-          [[ $GITHUB_REF == 'refs/heads/stable' ]] && VERSION=stable
-          TAGS="${DOCKER_IMAGE}:${VERSION}"
-          [[ $VERSION =~ ^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}$ ]] && TAGS="$TAGS,${DOCKER_IMAGE}:latest"
-          echo ::set-output name=tags::${TAGS}
+        uses: crazy-max/ghaction-docker-meta@v1
+        with:
+          images: |
+            docker.io/mailserver/docker-mailserver
```
radicand, line 62: One last thing to stickler on, if we can make these variables again - e.g.;

```yaml
images: |
  ${{ secrets.DOCKER_REPOSITORY }}
  ghcr.io/${{ github.repository_owner }}/docker-mailserver
```

Again, use-case being forks not having to change these values in the YAML. I don't know if the variable reference on that is correct (e.g., single or double `{` needed, but do test.

radicand (changes requested):

.github/workflows/default_on_push.yml
aendeavor: > > ```
> > 
> > images: |
> > 
> >   docker.io/mailserver/docker-mailserver
> > 
> >   ghcr.io/docker-mailserver/docker-mailserver
> > ```
> > 
> > 
> > These values can of course again be transformed into github secrets.
> 
> I will create these secrets.

These secrets are now created:

```  INI
DOCKER_REPOSITORY = docker.io/mailserver/docker-mailserver
GHCR_REPOSITORY = ghcr.io/docker-mailserver/docker-mailserver
```

The old value from `DOCKER_REPOSITORY` just stays the same:) @wernerfred you can now adjust the lines to use both secrets.

aendeavor:
> Yes, running into the same issues first but was able to solve it by enabling the improved container support for my personal account. This would be needed for the organization as well (docs).

I had already enabled this feature from the beginning when we migrated, so no troubles here. This should work fine.

aendeavor:
`GHCR_USERNAME` is now set to my username. `GHCR_PASSWORD` is filled with a PAT that follows the recommended security guidelines. I have updated the token to not use repo-scope afterwards as @radicand said, thank you for the hint to the bug:) It was exactly as you said.

@wernerfred All secrets should be set now.

.github/workflows/default_on_push.yml
wernerfred: Resolved with https://github.com/docker-mailserver/docker-mailserver/pull/1763/commits/d6c2eed2f003f043f8f81e5d6374b28d25b6febf

wernerfred:
So now everything from the checklist is done and should work as expected.

I have a few more places where I like to edit/correct the information:
- on [CONTRIBUTING.md](https://github.com/docker-mailserver/docker-mailserver/blob/master/CONTRIBUTING.md) edit the line "CI builds the new :latest image" to "CI builds the new :edge image"
- on [README.md](https://github.com/docker-mailserver/docker-mailserver/blob/master/README.md) Add a badge for ghcr pulls next to the dockerhub pulls
- on [README.md](https://github.com/docker-mailserver/docker-mailserver/blob/master/README.md) Add a section ("Available sources/tags") under [Miscellaneous](https://github.com/docker-mailserver/docker-mailserver#miscellaneous) and include the two current supported registries as well as the table with the tagging convention we agreed on.

Please vote 👍🏻 or 👎🏻 on this comment and I will edit these files based on the result.

aendeavor:
>     * on [CONTRIBUTING.md](https://github.com/docker-mailserver/docker-mailserver/blob/master/CONTRIBUTING.md) edit the line "CI builds the new `:latest` image" to "CI builds the new `:edge` image"

You will need to do this in following places too:

1. `Makefile`
2. `common.bash` under `test/test-helper`
3. In all other workflow files

>     * on [README.md](https://github.com/docker-mailserver/docker-mailserver/blob/master/README.md) Add a badge for ghcr pulls next to the dockerhub pulls

Good idea. Please use the "for-the-badge` style like the other badges already do:)

>     * on [README.md](https://github.com/docker-mailserver/docker-mailserver/blob/master/README.md) Add a section ("Available sources/tags") under [Miscellaneous](https://github.com/docker-mailserver/docker-mailserver#miscellaneous) and include the two current supported registries as well as the table with the tagging convention we agreed on.

Very good! Especially the information about the tagging convention.

wernerfred:
> 3\. In all other workflow files

What do you mean by that? Every other workflow uses either `:ci` for local builds or `:stable` (for the scheduled builds)

> on README.md Add a section ("Available sources/tags") under Miscellaneous and include the two current supported registries as well as the table with the tagging convention we agreed on. (Maybe it is better to add a new heading "Available images/tags" at the end of the README instead of putting them under "Miscellaneous" as it lives under "Usage")

Would you put it under Miscellaneous or in a separate section?

aendeavor:
> What do you mean by that? Every other workflow uses either :ci for local builds or :stable (for the scheduled builds)
And in common.bash and the Makefile it is a overwritable ENV

My bad, I mixed something up. Nevermind.

> Would you put it under Miscellaneous or in a separate section?

I would bring the "available sources" thing in the `Usage` section, maybe even as the first point, explaining that users now have a registry choice. The "tagging" should be explained there as well, since we differ from Docker's convention. Every user is more likely to read the usage section than the misc section. That's why I would prefer usage.

wernerfred:
Sadly there is (currently) no support for ghcr download stats due to problems with PAT permissions, see https://github.com/badges/shields/pull/4101

I will include it when it's working properly.

aendeavor:
> Sadly there is (currently) no support for ghcr download stats due to problems with PAT permissions, see https://github.com/badges/shields/pull/4101
> 
> 
> 
> I will include it when it's working properly.

That's fine for now. We do not rely on this or really need it:)

aendeavor (approved):

radicand (approved):
LGTM - side note, does the scheduled build workflow need to get updated as well?

wernerfred:
Nope as it clones the stable branch and is coded to build the stable image there is no adaption needed

wernerfred:
Ah my fault. There is no change needed for the logic **but** we need to add the ghcr login step too. My bad -.-
I will add this directly on the master if it's okay for both of you? @radicand @aendeavor

radicand:
If the push action running now succeeds, I don't have a strong reaction to a hot fix for the scheduled file.

wernerfred:
So the pipeline went through. It created the correct `:edge` tag for both registries, logged in successful in both registries and pushed the `:edge` image to DockerHub and ghcr.
Where the image can be found [on Dockerhub](https://hub.docker.com/layers/mailserver/docker-mailserver/edge/images/sha256-915bacc48c07b7d4e99bc1ea8c863d4bebbd692d0a1acae127e29ebbaa479999?context=explore) ghcr does not list any package now.
But on the projects site i see a small `1` next to packages:
image
But when accessing it it shows "nothing available":
image

Maybe this is a permission problem? @aendeavor can you double check as we are using your PAT at least you should see something.
Other possibility: ghcr is incredible slow in indexing and updating its content.

aendeavor:
@wernerfred you can of course push on master if you deem it necessary. In general, this is discouraged, but for safe additions, it's absolutely fine - needn't ask;)

I will have a look at the provided link and see what I can do.

PR #1776, opened 2021-01-27
Replace $_ in error trap

casperklein:
`$_` returns the last argument to the previous command.

Example: `false foo`

` $_` returns "foo"
`$BASH_COMMAND` returns the failing command and argument(s) "false foo"


aendeavor:
LGTM

Can you replace this in `setup.sh` and `test/linting/lint.sh` too?

aendeavor (approved):

Issue #1781, opened 2021-01-31
[BUG] bl.spamcop.net domain expired, resulting in rejected e-mails


mrPjer:
See also https://www.reddit.com/r/sysadmin/comments/l9asw7/spamcop_domain_expiredparked/.

I noticed I'm not receiving e-mails recently. Checked the logs and found this:

```
mail_1                          | Jan 31 12:39:29 mail postfix/smtpd[21240]: NOQUEUE: reject: RCPT from $SERVER: 554 5.7.1 Service unavailable; Client host [$IP] blocked using bl.spamcop.net; from=<$SENDER> to=<$RECIPIENT> proto=ESMTP helo=<$SERVER>
```

Turns out the spamcop domain expired and now all lookups against the blocklist are treated as failures.

I assume spamcop will fix the domain issue. However, can the blocklist check be improved so that it doesn't reject all e-mail if the service is not available?

mrPjer:
Here's a quick fix until this whole situation is resolved. To disable the spamcop check:

```
CONTAINER="your-container-name"
docker cp ${CONTAINER}:/etc/postfix/main.cf main.cf
# edit main.cf to comment out spamcop (in my case on line 47 and line 58)
docker exec $CONTAINER postfix reload
docker main.cf cp ${CONTAINER}:/etc/postfix/main.cf
```

I also tried overriding the hostname with
```
        extra_hosts:
            - "bl.spamcop.net:104.91.4.234"
```

in the compose file, but that didn't work.

perry-amsterdam:
I have the same problem:
NOQUEUE: reject: RCPT from out-18.smtp.github.com[192.30.252.201]: 554 5.7.1 Service unavailable; Client host [192.30.252.201] blocked using bl.spamcop.net; from=<noreply@github.com> to=<perry@couprie.net> proto=ESMTP helo=<smtp.github.com>

adrianrudnik:
Had to kill bl.spamcop.net. Thats quite the mishap sadly. Had to copy out `/etc/postfix/main.cnf` sections that contain `bl.spamcop.net`.

Then put it into the overriding `config/postfix-main.cf` file and remove any breaklines, leading to

```
[...]

PR #1782, opened 2021-01-31
removing bl.spamcop.net for 8.0.1

aendeavor:
Fixes: #1781


wernerfred (approved):
LGTM 👍🏻 

Just publish a release with tag v8.0.1 afterwards and the actions will kick in. The hotfix patch is then available in `edge`, `8`, `8.0`, `8.0.1` and `latest`.

PR #1783, opened 2021-01-31
doc: fix service name now mailserver

m-schmoock:
This was just an outdated piece of example in the README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,7 +190,7 @@ We are currently providing support for Linux. Windows is _not_ supported and is 
 ``` BASH
 docker-compose down
 docker pull docker.io/mailserver/docker-mailserver:<VERSION TAG>
-docker-compose up -d mail
+docker-compose up -d mailserver
 ```
 
 You're done! And don't forget to have a look at the remaining functions of the `setup.sh` script with `./setup.sh -h`.


aendeavor (approved):

aendeavor:
Thanks! LGTM

m-schmoock:
@aendeavor 

Quick question since your around and so I don't open up a PR unnecessary: Do we need to include the `compose.env` in the `docker-compose.yml` or is this done by docker-compose internally? I don't see any useful references to `compose.env` when I do a `git grep compose.env`
```YAML
    env_file:
       - mailserver.env
       - compose.env
```

aendeavor:
> @aendeavor
> 
> Quick question since your around and so I don't open up a PR unnecessary: Do we need to include the `compose.env` in the `docker-compose.yml` or is this done by docker-compose internally? I don't see any useful references to `compose.env` when I do a `git grep compose.env`
> 
> ```yaml
>     env_file:
>        - mailserver.env
>        - compose.env
> ```

Your question is valid, but have a look at the README: "_Variables in .env are expanded in the docker-compose.yml file only and not in the container. The file mailserver.env serves this case where environment variables are used in the container._" Above that, you see the command

``` BASH
wget -O .env https://raw.githubusercontent.com/docker-mailserver/docker-mailserver/master/compose.env
```

Therefore, `compose.env` becomes your `.env` file if you need one, and Docker will automatically acpand variables in your Compose-file when it detects a `.env` file. Therefore, the name `compose.env` was chosen.

The `docker-compose.yml` tries to make this clear by only using `env_files: mailserver.env` instead of

``` YAML
env_file:
  - mailserver.env
````

i.e. using a list.

PR #1785, opened 2021-01-31
Escaped the underscore of SCORE

froks:
When you read the markdown on github, the underscore by default just causes the word to be italic - when you blindly copy that configuration, it won't work, because you don't see the underscore prefix/suffix.


aendeavor:
@casperklein I don't anything that could be off, if you don't either, we'll can merge it;)

casperklein:
Good find. Thank you @froks.

PR #1787, opened 2021-02-01
correct application of the new SUPERVISOR_LOGLEVEL variable

aendeavor:
Follow up on #1758

### DIFF
--- a/target/scripts/start-mailserver.sh
+++ b/target/scripts/start-mailserver.sh
@@ -356,7 +356,7 @@ function setup
 
 function _setup_supervisor
 {
-  if ! grep "loglevel = ${SUPERVISOR_LOGLEVEL}" /etc/supervisor/supervisord.conf &>/dev/null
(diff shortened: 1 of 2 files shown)


target/scripts/start-mailserver.sh
```diff
@@ -356,22 +356,22 @@ function setup
...
-      _notify 'warn' \
-        "SUPERVISOR_LOGLEVEL value '${SUPERVISOR_LOGLEVEL}' unknown. Defaulting to 'warn'"
-      sed -i -E \
-        "s+loglevel.*+loglevel = warn+g" \
-        /etc/supervisor/supervisord.conf
-      ;;
-  esac
-
-  supervisorctl update
+  if ! grep "loglevel = ${SUPERVISOR_LOGLEVEL}" /etc/supervisor/supervisord.conf &>/dev/null
```
casperklein, line 359: `grep -q` to avoid `&>/dev/null`

aendeavor: Done.

casperklein (approved):

PR #1559, opened 2020-07-05
Support extra user_attributes in accounts configuration

abh:
This allows you to add for example

    |userdb_mail=mbox:~/mail:INBOX=~/inbox

to the end of an account to have a different mailbox configuration.

PR #1792, opened 2021-02-06
Support extra user_attributes in accounts configuration

abh:
This allows you to add for example

    |userdb_mail=mbox:~/mail:INBOX=~/inbox

 to the end of an account to have a different mailbox configuration.

### DIFF
--- a/test/tests.bats
+++ b/test/tests.bats
@@ -60,6 +60,7 @@ setup_file() {
 	docker exec mail /bin/sh -c "nc 0.0.0.0 25 < /tmp/docker-mailserver-test/email-templates/existing-alias-recipient-delimiter.txt"
 	docker exec mail /bin/sh -c "nc 0.0.0.0 25 < /tmp/docker-mailserver-test/email-templates/existing-user1.txt"
 	docker exec mail /bin/sh -c "nc 0.0.0.0 25 < /tmp/docker-mailserver-test/email-templates/existing-user2.txt"
+	docker exec mail /bin/sh -c "nc 0.0.0.0 25 < /tmp/docker-mailserver-test/email-templates/existing-user3.txt"
 	docker exec mail /bin/sh -c "nc 0.0.0.0 25 < /tmp/docker-mailserver-test/email-templates/existing-added.txt"
 	docker exec mail /bin/sh -c "nc 0.0.0.0 25 < /tmp/docker-mailserver-test/email-templates/existing-user-and-cc-local-alias.txt"
 	docker exec mail /bin/sh -c "nc 0.0.0.0 25 < /tmp/docker-mailserver-test/email-templates/existing-regexp-alias-external.txt"
@@ -248,6 +249,7 @@ teardown_file() {
  1 <user1@localhost.localdomain>, orig_to=<root>
  1 <user1~test@localhost.localdomain>
  2 <user2@otherdomain.tld>
+ 1 <user3@localhost.localdomain>
 EOF
 }
 
@@ -338,7 +340,8 @@ EOF
   assert_success
   assert_line --index 0 "user1@localhost.localdomain"
   assert_line --index 1 "user2@otherdomain.tld"
-  assert_line --index 2 "added@localhost.localdomain"
+  assert_line --index 2 "user3@localhost.localdomain"
+  assert_line --index 3 "added@localhost.localdomain"
 }
 
 @test "checking accounts: user mail folder for user1" {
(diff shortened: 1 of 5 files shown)


abh:
This is a rebased patch from #1559 

I tried adding a test, too, but "make tests" just hangs for me after doing the first test, so hopefully the CI will tell me if it worked?

aendeavor:
> I tried adding a test, too, but "make tests" just hangs for me after doing the first test, so hopefully the CI will tell me if it worked?

Submodules are initialized? What OS are you on. The test only support Linux, no other OS. If you're on Linux and submodules are innitialized, hanging tests indicate something is wrong.

test/tests.bats
```diff
@@ -1,3 +1,5 @@
+set -x
```
aendeavor, line 1: Very dangerous for a file that is this large. Probably the reasons tests are unstable. This would need extensive testing. It also does not work in functions and will result in unexpected behavior.

aendeavor:
Had to cancel the workflow as it run for 42 minutes when around 10 minutes is expected. Please make it work local first. This cannot be tested in our CI as it's really broken.

aendeavor (changes requested):
CI completely broken with this PR. Unusable in this state. Please make it work locally first. Init submodules and run `make install_linters`.

test/tests.bats
abh: whoops -- thanks. Fixed!

abh:
> > I tried adding a test, too, but "make tests" just hangs for me after doing the first test, so hopefully the CI will tell me if it worked?
> 
> Submodules are initialized?

Yes.

> What OS are you on.

Latest CentOS 7.

> The test only support Linux, no other OS. If you're on Linux and submodules are innitialized, hanging tests indicate something is wrong.

Indeed!  I didn't figure out what fixed it (a yum upgrade? I used a server that I don't usually use for Docker) but it's running now, albeit really really slowly.

I'll get the tests fixed and submit again.

aendeavor:
You will need to rebase on current master to resolve merge conflicts. A file was deleted, which is now shown as a conflict.

abh:
> You will need to rebase on current master to resolve merge conflicts. A file was deleted, which is now shown as a conflict.

yeah, I submitted the PR for removing the file after finding it wasn't used when trying to use it in my tests here. 🤣

aendeavor:
LGTM

But I will request another 4 eyes to look at it, just for good measure

aendeavor (approved):

wernerfred:
Can you please elaborate on the following points:
- Why are those changes necessary?
- Are there unwanted side effects possible
- How and where do you set that configuration
I am missing a few background information / motivation about this PR.

abh:
> Can you please elaborate on the following points:
> 
> * Why are those changes necessary?

Dovecot supports overriding many settings per user; adding them in the "extra attributes" field is the way to do it for the authentication we use in docker-mailserver (when not using LDAP anyway).

My use case is having some users use a different mailbox format (in one case because the usage pattern is weird and in some others to ease migration of some large mailboxes from an old dovecot installation).

Another use case I was planning to use is to have a particular robot/system account be restricted to only connect from the network it's expected to be on.

https://doc.dovecot.org/configuration_manual/authentication/user_database_extra_fields/  (the very last example is the one that applies to how docker-mailserver works).

https://doc.dovecot.org/configuration_manual/authentication/passwd_file/

> * Are there unwanted side effects possible

No, we already use the field for the quotas.

> * How and where do you set that configuration
 
Just by adding the third column in postfix-accounts.cf as I did in the test file (generated by the Makefile). I can write an example for the wiki after it's merged.

I looked at if there were overrides that made sense to promote to the level of support that quotas have, but I wasn't sure where the natural cut-off point is. The feature as-is is a good "safety valve" to allow the per-user dovecot features to be used.

wernerfred:
Thanks for taking the time and explaining your motivation and changes 💡
Imho nothing speaks against a merge.

> Just by adding the third column in postfix-accounts.cf as I did in the test file (generated by the Makefile). I can write an example for the wiki after it's merged.

Yes, please write down an example of this functionality in the wiki otherwise nobody knows it exists. :writing_hand:

> I looked at if there were overrides that made sense to promote to the level of support that quotas have, but I wasn't sure where the natural cut-off point is.

You are right, it's hard to find this point. Imho your change is something only a few will use (never saw it requested by an issue) but maybe you will save someones day with this feature :+1: 

Thanks for your contribution :handshake:

wernerfred (approved):

aendeavor:
Merging when the last review comes in:)

casperklein (approved):

casperklein:
@abh Thank you for this PR.

As @wernerfred wrote, please add some documentary/example to the README or Wiki.

PR #1798, opened 2021-02-07
Enhancement for function _setup_postfix_sasl fixing #1796 & More

### LINKED ISSUE #1796: [BUG] SASL is enabled with ENABLE_SASLAUTHD=0 but only when SMTP_ONLY=1
# Bug Report

I believe this is the same as #1687. I do not want SASL.

Using `ENABLE_SASLAUTHD=0` and `SMTP_ONLY=1` results in the following log error (and `could not connect to this SMTP server` error in Apple Mail).

This is the relevant log error:

```
Feb  7 14:15:31 mailer-ams postfix/submission/smtpd[892]: warning: SASL: Connect to private/auth failed: Connection refused
Feb  7 14:15:31 mailer-ams postfix/submission/smtpd[892]: fatal: no SASL authentication mechanisms
```

## Context

Fresh installation.

Using SMTP.

### When does this occur?

Every time.

I follow the [default installation instructions](https://github.com/docker-mailserver/docker-mailserver#get-the-tools).

Besides adding domain and hostname to `.env`, this is what I edit in `mailserver.env`:

```
SMTP_ONLY=1
ONE_DIR=1
PERMIT_DOCKER=host
SSL_TYPE=manual
SSL_CERT_PATH=/path/to/fullchain.cer
SSL_KEY_PATH=/path/to/cert.key
```
I keep `ENABLE_SASLAUTHD=0` with its default value.

`docker-compose.yml` only mounts `/path/to/`. https://www.checktls.com is happy with the settings, gives 100% and valid certificate.

# Workaround

Setting `SMTP_ONLY=` (to empty) fixes the issue.

## Your Environment

- version: `docker.io/mailserver/docker-mailserver:latest`
- available RAM: `8GB`
- Docker version: `Docker version 20.10.3, build 48d30b5`

aendeavor:
# Description

SASL is now disabled correctly. The `Dockerfile` has been adjusted to not show the `delaying debconf because apt-utils is not installed` error anymore. `function _misc_save_states` was refactored too as it was in an unacceptable state. Especially proud of the inline variable substitutions that got rid of `echo ... | sed ...`.

Fixes #1796

## Type of change

## Checklist:

- [x] I have made corresponding changes to the documentation (README.md or ENVIRONMENT.md or the Wiki)

---

PS: I like the new template :D

### DIFF
--- a/target/scripts/start-mailserver.sh
+++ b/target/scripts/start-mailserver.sh
@@ -925,7 +925,7 @@ function _setup_spoof_protection
 
 function _setup_postfix_access_control
 {
-  _notify 'inf' "Configuring user access"
+  _notify 'inf' 'Configuring user access'
 
   if [[ -f /tmp/docker-mailserver/postfix-send-access.cf ]]
   then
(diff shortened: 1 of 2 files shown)


wernerfred:
> PS: I like the new template :D

🥳 

My hope is that people get motivated providing some more information than they would without it

wernerfred (approved):
LGTM

target/scripts/start-mailserver.sh
```diff
@@ -1970,42 +1969,45 @@ function misc
...
+      lib/amavis
+      lib/clamav
+      lib/spamassassin
+      lib/fail2ban
+      lib/postgrey
+      lib/dovecot
     )
 
-    for d in "${FILES[@]}"
+    for FILE in "${FILES[@]}"
```
casperklein, line 1992: FILE should be "local"

target/scripts/start-mailserver.sh
```diff
@@ -1970,42 +1969,45 @@ function misc
 
 function _misc_save_states
 {
-  # consolidate all states into a single directory (`/var/mail-state`) to allow persistence using docker volumes
-  statedir=/var/mail-state
+  # consolidate all states into a single directory (`/var/mail-state`)
+  # to allow persistence using docker volumes
 
-  if [[ ${ONE_DIR} -eq 1 ]] && [[ -d ${statedir} ]]
+  local STATEDIR=/var/mail-state
```
casperklein, line 1975: Quotes missing. Not necesary, but to keep one standard across the whole file.

STATEDIR="/var/mail-state"

target/scripts/start-mailserver.sh
aendeavor: didn't know `for local FILE in ...` was even possible, oder did you mean

``` BASH
local FILE
for FILE in ...
```

which is to the best of my knowledge not needed since `FILE` only exists in the loop. And the FILE I explicitly define in the loop is marked `local`. What did I miss?:)

target/scripts/start-mailserver.sh
aendeavor: True. I think this is okay for now, and I will push this with a future commit.

target/scripts/start-mailserver.sh
casperklein: I meant the second example from you.

`FILE` is not declared local before it's first usage in line 1992. So at this point, it's global and keeps the value `spool/postfix`.

If you run e.g. `echo $FILE` after the function call, you would get `spool/postfix`.
```
_misc_save_states
echo $FILE
```

aendeavor: I always thought this variable is only valid inside the loop, but I'm not actually surprised that it's not - I guess you can always learn something new.  Bash is still full of surprises.

I will do a follow-up here.

PR #1804, opened 2021-02-10
Exlude files from triggering workflows

wernerfred:
# Description

In this PR I added the functionality to our CI/CD pipelines to exclude some files to trigger the workflow.
Basically i added the following lines:
```
paths-ignore: 
  - '**.md'
  - 'LICENSE'
  - '.github/**'
```
Imho this change can save up (waiting)time of unnecessary workflow runs as a change or hotfix on the excluded files will **never** impact the image. Furthermore I think it is possible to exclude even more files as they won't affect the image as well but I want your opinion on the topic itself and then the files that can be excluded too to not miss anything important.

The files that could be also excluded:
- `docker-compose.yml`
- `compose.env`
- `mailserver.env`
- `setup.sh`

## Type of change

## Checklist:

- [x] I have made corresponding changes to the documentation (README.md or ENVIRONMENT.md or the Wiki)

### DIFF
--- a/.github/workflows/default_on_push.yml
+++ b/.github/workflows/default_on_push.yml
@@ -5,6 +5,10 @@ on:
     branches:
       - master
       - stable
+    paths-ignore:
+      - '**.md'
+      - 'LICENSE'
+      - '.github/**'
     tags:
       - '*.*.*'
 
--- a/.github/workflows/default_on_push.yml
+++ b/.github/workflows/default_on_push.yml
@@ -5,10 +5,12 @@ on:
     branches:
       - master
       - stable
-    paths-ignore:
-      - '**.md'
-      - 'LICENSE'
-      - '.github/**'
+    paths: 
+      - 'target/**'
+      - '.dockerignore'
+      - '.gitmodules'
+      - 'Dockerfile'
+      - 'setup.sh'
     tags:
       - '*.*.*'
 
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -3,8 +3,16 @@ name: "Lint"
 on:
   pull_request:
     branches: [ "*" ]
+    paths-ignore: 
+      - '**.md'
+      - 'LICENSE'
+      - '.github/**'
   push:
     branches: [ "master", "stable" ]
+    paths-ignore: 
+      - '**.md'
+      - 'LICENSE'
+      - '.github/**'
 
 jobs:
   lint:
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -3,16 +3,8 @@ name: "Lint"
 on:
   pull_request:
     branches: [ "*" ]
-    paths-ignore: 
-      - '**.md'
-      - 'LICENSE'
-      - '.github/**'
   push:
     branches: [ "master", "stable" ]
-    paths-ignore: 
-      - '**.md'
-      - 'LICENSE'
-      - '.github/**'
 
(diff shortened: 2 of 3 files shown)


aendeavor:
A very good idea.

> The files that could be also excluded:

docker-compose.yml       <- yes, not tested
compose.env                  <- not tested either, yes
mailserver.env                <- again, not tested, fine
setup.sh                         <- **not okay**, since CI tests this script: https://github.com/docker-mailserver/docker-mailserver/blob/62e93f0dcc898095b45619f4f481fc5e43619b0a/test/tests.bats#L1242-L1256 ...

A very welcome addition.

aendeavor (approved):

wernerfred:
> setup.sh <- **not okay**, since CI tests this script:

Perfectly that's why I wanted more eyes to look on this. Testing those files is the one thing but beeing needed for the build (or let's say changing the behaviour of the build) is the other thing (an example would be the `.dockerignore` file).

wernerfred:
Now after seeing @casperklein proposing the other way around in #1805 I am wondering if it would make sense here, too?

Here is the list of all files and directories at top level:

- [ ] CODE_OF_CONDUCT.md  
- [ ] CONTRIBUTING.md    
- [x] .dockerignore   
- [x] .gitmodules     
- [ ] Makefile   
- [x] target
- [ ] compose.env         
- [ ] docker-compose.yml  
- [ ] .editorconfig   
- [ ] .github     
- [ ] LICENSE         
- [ ] README.md  
- [ ] test
- [ ] CHANGELOG.md  
- [ ] config              
- [x] Dockerfile          
- [ ] ENVIRONMENT.md  
- [ ] .gitignore  
- [ ] mailserver.env  
- [x] setup.sh

And I think it is only necessary to trigger ci/cd if we have changes on the ticked ones (correct me if I missed one). I excluded `tests/` on purpose as changes to tests won't justify a new image beeing build. And new tests can and should be run local during development anyways. So imho it is absolutely fine to run the CI/CD only if files are affected that actually change the image.

What do you think?

aendeavor:
> I excluded tests/ on purpose as changes to tests won't justify a new image beeing build. And new tests can and should be run local during development anyways. So imho it is absolutely fine to run the CI/CD only if files are affected that actually change the image.

We would need to include `test/` for testing PRs, since we cannot rely that changes to tests are not breaking anything. But that's not a problem as we can just do this for the workflow file that tests PRs.

Other than that, I don't see a problem. I actually quite like this and #1805 since it's like the paradigm of modern programming languages like Rust: making everything private and marking things that should be public - analogous this approach explicitley describes what changes should trigger a new CI run.

wernerfred:
Alright, I will wait if one more throws his opinion in and then implement the changes like shown in the ticklist - with the exception to include `tests/` on PRs.

aendeavor:
> Alright, I will wait if one more throws his opinion in and then implement the changes like shown in the ticklist - with the exception to include `tests/` on PRs.

Then I'd suggest to request one review or more so @docker-mailserver/maintainers notice it :D

casperklein:
I like the approach. Go for it 👍

fbartels (approved):

wernerfred:
I removed the `paths` option on the linting check as every file will be at least checked with `eclint`. The others now use explicit `paths` instead of the previous proposed `paths-ignore`.

I will now check if an update to `CONTRIBUTING.md` or other documentation is needed and add appropriate parts. Afterwards I will merge the PR.

PR #1831, opened 2021-02-23
linting: use local scripts to remove sudo need

ap-wtioit:
# Description

this change installs the linting tools in a local tools directory. this removes the need for `sudo curl` and replacing binaries in /usr/local/bin and /usr/bin.

Issues with the previous code:
* /usr/local/bin and /usr/bin might be shared on CI systems by multiple builds influencing each other
* the old code might leave some of the tools shared with other build unexecutable when downloading them to /usr/local/bin and /usr/bin for short periods of time (breaking other builds)

## Type of change

## Checklist:

Info @wt-io-it


aendeavor:
I like this change.

LGTM

aendeavor (approved):

wernerfred (commented):
A few thoughts from my side:

Imho there is a problem when "installing" to a local path in the repository as one might run these linting checks locally to test befor opening a PR. Therefore the local repository would include the tools as a local change. There are multiple possibilities to solve this. For example make it optional to remove them afterwards via `make clean` or to add `tools/**` to `.gitignore`.

For a long sight: I am preparing a PR (which I came up with already before the migration started) to switch the linting tools etc to their docker versions. This will solve the issue with "installing" completely and should not be a problem for anyone as everyone is using docker already when using docker-mailserver. 
But for the meantime i like your change and will approve if the issue mentioned above is resolved/discussed.

ap-wtioit:
@wernerfred please look at the Files changed tab, tools is already added to .gitignore

aendeavor:
There are multiple possibilities to solve this. For example make it optional to remove them afterwards via `make clean` or to add `tools/**` to `.gitignore`.

Did I miss something?

``` IGNORE
###  Linting Tools  #############################

tools/
```

?

wernerfred:
omg, my bad. Looked at the changed files of course but (howsoever) must have overlooked changes to `.gitignore`

wernerfred (approved):

PR #1851, opened 2021-03-13
Fix update docs to use `setup.sh help`

analogue:
# Description

`setup.sh -h` was replaced with `setup.sh help` somewhere along the line.  Just reflecting change in README


casperklein:
Thank you!

PR #295, opened 2016-08-30
Improve and extend setup.sh

Josef-Friedrich:
Add subcommands 'debug show-mail-logs', 'inspect', 'login'. Add option
'-c' to specify the name of the running container. Add option '-i' to
specify the image name.

PR #1874, opened 2021-03-29
Fix for #1808

### LINKED ISSUE #1808: [BUG] Mailbox contents not deleted
# Bug Report

## Context
Mailbox contents remain in `maildata` subdirectory after running `sudo ./setup.sh email del test@example.com`.

Deleting mailboxes.
### When does this occur?
When deleting a mailbox with the `setup.sh` script.

```
sudo ./setup.sh email del test@example.com
Password: 
sed: can't read /tmp/docker-mailserver/postfix-virtual.cf: No such file or directory
Do you want to delete the mailbox as well(all mails will be removed)?(y/n) y

Mailbox deleted.
```
Check contents of `maildata` subdirectory.

### Actual Behavior
The mailbox is no longer listed when running `sudo ./setup.sh email list`. But the `test` mailbox and its contents are still in the `maildata` subdirectory. 
### Expected Behavior
The mailbox contents should have been deleted.
## Your Environment
Synology DS918+ with latest DSM and Docker version 18.09.8, build bfed4f5.

### Environment Variables

```yml
---
version: "2"

# Define services
services:
  
  roundcubemail:
    image: roundcube/roundcubemail:latest
    container_name: roundcubemail
    restart: unless-stopped
    volumes:
      - /v/d/roundcube/www:/var/www/html
      - /v/d/roundcube/sqlite:/var/roundcube/db
    ports:
      - 127.0.0.1:8000:80
    environment:
      - ROUNDCUBEMAIL_DB_TYPE=sqlite
      - ROUNDCUBEMAIL_SKIN=elastic
      - ROUNDCUBEMAIL_DEFAULT_HOST=ssl://mail.example.com
      - ROUNDCUBEMAIL_SMTP_SERVER=ssl://mail.example.com
      - ROUNDCUBEMAIL_DEFAULT_PORT=993
      - ROUNDCUBEMAIL_SMTP_PORT=465
  
  solr:
    container_name: mailserver_solr
    image: lmmdock/dovecot-solr:latest
    volumes:
      - /v/d/solr-dovecot:/opt/solr/server/solr/dovecot
    restart: always
  
  mailserver:
    hostname: mail
    domainname: example.com
    container_name: mailserver
    ports:
      - "465:465"
      - "993:993"
    volumes:
      - /v/d/mailserver/maildata:/var/mail
      - /v/d/mailserver/mailstate:/var/mail-state
      - /v/d/mailserver/maillogs:/var/log/mail
      - /v/d/mailserver/config:/tmp/docker-mailserver/
      - /v/d/ssl/output:/tmp/ssl:ro
      - /v/d/mailserver/solr-dovecot/10-plugin.conf:/etc/dovecot/conf.d/10-plugin.conf:ro
    environment:
      - SSL_TYPE=manual
      - POSTFIX_MESSAGE_SIZE_LIMIT=102400000
      - ENABLE_SPAMASSASSIN=0
      - ENABLE_CLAMAV=0
      - ENABLE_FAIL2BAN=0
      - ENABLE_POSTGREY=0
      - ENABLE_SASLAUTHD=0
      - ONE_DIR=1
      - DMS_DEBUG=0
    cap_add:
      - NET_ADMIN
      - SYS_PTRACE
    restart: always
```
## Workaround?
Contents of mailbox are deleted when running `sudo docker exec mailserver rm -R /var/mail/example.com/test/` afterwards.

casperklein:
# Description

It was not possible to delete the mail directory when using `setup.sh`

```
./setup.sh email del foo@example.com
[..]
Mailbox directory '/var/mail/example.com/foo' did not exist.
```

That was caused by not using the running mail container, instead a new container was started without `maildata` being mounted.

Fixes #1808

## Type of change

## Checklist:

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -340,7 +340,7 @@ function _main
       case ${2:-} in
         add      ) shift 2 ; _docker_image addmailuser "${@}" ;;
         update   ) shift 2 ; _docker_image updatemailuser "${@}" ;;
-        del      ) shift 2 ; _docker_image delmailuser "${@}" ;;
+        del      ) shift 2 ; _docker_container delmailuser "${@}" ;;
         restrict ) shift 2 ; _docker_container restrict-access "${@}" ;;
         list     ) _docker_image listmailuser ;;
         *        ) _usage ;;


wernerfred (commented):
If that is the reason wouldn't it then make sense to modify the other commands to use `_docker_container` instead of `_docker_image` too?

aendeavor:
> If that is the reason wouldn't it then make sense to modify the other commands to use `_docker_container` instead of `_docker_image` too?

First of all: Why was the `_docker_image` thing chosen in the first place. I'm under the impression that there's a valid reason and I feel like someone may know more: @erik-wramner do you have an idea?

casperklein:
`_docker_image` was fine, until #897 got merged. That commit introduced the "delete mailbox feature".
This makes it necessary to access the mailbox files (to delete them). `delmailuser` works fine within the container. But when called via `setup.sh`, a new container was started, that didn't have access to the mailbox files.

`_docker_image` was chosen before that feature, because no access to the mailbox was necessary. Only access to `/tmp/docker-mailserver/postfix-accounts.cf` was needed, which `_docker_image` provided when starting a new container:

https://github.com/docker-mailserver/docker-mailserver/blob/0cd723208cb12b95589625250d13cf9ba0fc0cce/setup.sh#L235-L237

polarathene:
[History just shows the original `_docker` method for image use being replaced](https://github.com/docker-mailserver/docker-mailserver/pull/295/files) (2016) with separate command for image and a new one for running container, there wasn't much discussion about the inclusion.

casperklein:
> If that is the reason wouldn't it then make sense to modify the other commands to use `_docker_container` instead of `_docker_image` too?

The advantage of `_docker_image` is, that no running container is necessary. So you can for example create new mailboxes, before having a fully configured/running docker-mailserver yet. The downside is, that when you already have a running container, starting a new separte one will cost some resources.

Regarding the deletion of the mailbox:
Because we don't know, where the maildata exactly resides on the host (no we should not think about parsing docker-compose.yaml :pray:), we cannot start a new container and mounting the maildata easily. That's why for this kind of task, a running container is probably the best way to have access to the needed mounts.

To sum things up:
`_docker_image` for tasks that only need access to `/tmp/docker-mailserver`
`_docker_container` for tasks that need also access to `maildata`, `maillogs` and `mailstate` mounts. Also for tasks that obviously need a running container like `setup.sh debug login`. 

While writing this, I noticed, that `_docker_image` also has the ability, to use a running container. But then, `-c` (`CONTAINER_NAME`) must be  given to `setup.sh`.

https://github.com/docker-mailserver/docker-mailserver/blob/0cd723208cb12b95589625250d13cf9ba0fc0cce/setup.sh#L221-L226

We could discard this PR, but then `-c` would be mandatory for the `setup.sh email del` command.

polarathene:
> We could discard this PR, but then `-c` would be mandatory for the `setup.sh email del` command.

If there isn't a situation where it'd make sense to use `_docker_image`, then the PR should be merged?

---

Regarding the rest of your comment, that seems like useful information to add to documention? Seems some of it would be relevant to maintainers/contributors and another part maybe for the existing `setup.sh` docs page?

fbartels (approved):

aendeavor:
After reading through all of this, I could provide a PR that reverses my last PR. With my last PR I disabled the possibility to delete more than one user at a time with one run of `setup.sh`.

Ping me if I should re-enable this.

---

This look good however:)

aendeavor (approved):

wernerfred (approved):
Are you going to document this within this PR? Just asking bc i will have a Close Look then at the workflows etc For building Docs If anything will work as planned

casperklein:
>Regarding the rest of your comment, that seems like useful information to add to documention? Seems some of it would be relevant to maintainers/contributors and another part maybe for the existing setup.sh docs page?

I had a look at the documentation, but to be honest, I've no idea where to put it at best. What exactly would you suggest to put at [setup.sh](https://docker-mailserver.github.io/docker-mailserver/edge/config/setup.sh/)?

Creating a new category for developers, which e.g. describes the internal works of `setup.sh`, would be a bigger task. If someone is going to start this, then it should be done in a separate PR.

> Ping me if I should re-enable this.

Depends on your spare time :wink: Go ahead if it's no big deal for you 👍  But if we leave it as it is, it's also fine I think.

> Are you going to document this within this PR?

I don't think so at the moment. But I'll wait for @polarathene feedback or from anybody else of course.

polarathene:
> What exactly would you suggest to put at [setup.sh](https://docker-mailserver.github.io/docker-mailserver/edge/config/setup.sh/)?

It could be added as a `tip` admonition. The syntax in markdown for that can be noticed by `!!!` or `???`, followed by a type such as `tip`,`note`,etc and a quoted string for title, a blank line afterwards followed by content that is indented by 4 spaces.

I can add it if you like. I'll add it to my backlog?

I'm presently chipping away at the contributor section for documentation (both a contributor and maintainer page). That has priority as it will help better inform contributors how to write documentation.

> Creating a new category for developers, which e.g. describes the internal works of `setup.sh`, would be a bigger task. If someone is going to start this, then it should be done in a separate PR.

This can just be a maintainer notes page with a heading/section for notes on `setup.sh`. That information can be more informal and sparse, it's still useful to collect in a common place rather than scattered issue threads :)

We have a [roughed out a contributing section already](https://github.com/docker-mailserver/docker-mailserver/tree/master/docs/content/contributing) with stubs for some pages like writing tests for example. Notes for maintainers could be added to a new `maintainers.md` file or similar there.

---

From a new user perspective, viewing the current [`setup.sh` docs page](https://docker-mailserver.github.io/docker-mailserver/edge/config/setup.sh/), there is a bunch of example commands shown without any usage of `-i` or `-c`. There's no way to differentiate when it's being handled differently or is required to be specifically `-c` via this command help/examples.

> The advantage of `_docker_image` is, that no running container is necessary. So you can for example create new mailboxes, before having a fully configured/running docker-mailserver yet. The downside is, that when you already have a running container, starting a new separate one will cost some resources.

This could be a note after the `Usage` section on the docs page for users. But with `_docker_image` changed to be tailored as a note about when to choose `-i` vs `-c`, or if they don't need to bother with either (default `-i` value may be sufficient, or `-c` may be required as noted here?).

I'm not great at parsing the `setup.sh` myself, it looks like there is some fallback to try get the container name without using `-c`? 

> - `_docker_image` for tasks that only need access to `/tmp/docker-mailserver`
> - `_docker_container` for tasks that need also access to `maildata`, `maillogs` and `mailstate` mounts. But also for tasks that obviously need a running container like `setup.sh debug login`.

Probably useful for anyone maintaining/auditing the script. Could be added to maintainer docs, or inline documentation within the script file itself? (The docs Usage section with the CLI output looks like it's possibly out of date/sync?)

> While writing this, I noticed, that `_docker_image` also has the ability, to use a running container. But then, `-c` (`CONTAINER_NAME`) must be given to `setup.sh`.

Would this have alleviated the issue prior to any changes? Is it an indication that maintainers and users aren't as aware of this approach and it could be better highlighted/emphasized in our new docs?

```markdown
!!! tip "Target a running container instance"

    By default `setup.sh` commands will run in a new container without any volumes attached. This is useful when setting up a mailserver, however for maintenance sometimes you need to target a running instance that has volumes attached.

    Use `-c <container name>` to target an existing `docker-mailserver` instance for `setup.sh` commands to run within.
```

That may be a bit inaccurate, I haven't actually used this project properly in some time and forget what it's like using this file :)

It seems to suggest it's run externally of the container, to run some commands within the container? A little puzzling since we have `docker exec` / `docker-compose exec` that could run the correct `setup.sh` embedded within the image release that's running the container.

May be better to move this off to a separate discussion/issue thread. I have no opinion on merging or closing the PR personally, just an interest in the UX and maintainance of `setup.sh` itself.

casperklein:
> I can add it if you like. I'll add it to my backlog?

Yes please :heart:

> there is a bunch of example commands shown without any usage of -i or -c.

I guess, that's because `-i/-c` aren't required in a standard setup. `IMAGE_NAME` is hardcoded to `docker-mailserver:latest` and `CONTAINER_NAME` is automatically detected. So normally you don't have to use one of those switches. I only mentioned it, as a workaround for the currently not working `email del` command. My point of view is, that `setup.sh` should work out of the box and not requiring to manually give the image or container name. I cannot imagine a real use case for those switches.

> Would this have alleviated the issue prior to any changes? 

I am not sure. The whole error was a bit hard to track down IMHO. Also it was a clear bug. Not some kind of misusage of `setup.sh`.

> Is it an indication that maintainers and users aren't as aware of this approach and it could be better highlighted/emphasized in our new docs? it could be better highlighted/emphasized in our new docs?

More/better documentation is always good 👍 

> May be better to move this off to a separate discussion/issue thread.

Yes. I am going to merge this now. As you may noticed, documentation does not belong to my strengths :confused:. However I am happy to support/review/help when it goes to technical contents.

NorseGaud:
@casperklein 

> My point of view is, that setup.sh should work out of the box and not requiring to manually give the image or container name.

I just spent hours of my life trying to figure out what the heck is wrong with my delete and create commands, all to find out that -c is necessary... This definitely needs to work out of the box :)

casperklein:
As this has been fixed, you should update to the _latest_ [setup.sh](https://raw.githubusercontent.com/docker-mailserver/docker-mailserver/master/setup.sh) file.

NorseGaud:
@casperklein , that doesn't seem to work:



```
[ec2-user@ip-172-31-3-247 ~]$ wget https://raw.githubusercontent.com/docker-mailserver/docker-mailserver/master/setup.sh
--2021-05-19 00:31:13--  https://raw.githubusercontent.com/docker-mailserver/docker-mailserver/master/setup.sh
Resolving raw.githubusercontent.com (raw.githubusercontent.com)... 185.199.108.133, 185.199.109.133, 185.199.110.133, ...
Connecting to raw.githubusercontent.com (raw.githubusercontent.com)|185.199.108.133|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 11940 (12K) [text/plain]
Saving to: ‘setup.sh’

100%[=================================================================================================>] 11,940      --.-K/s   in 0s      

2021-05-19 00:31:13 (77.2 MB/s) - ‘setup.sh’ saved [11940/11940]

[ec2-user@ip-172-31-3-247 ~]$ chmod +x setup.sh 

[ec2-user@ip-172-31-3-247 ~]$ ./setup.sh email list
doveconf: Fatal: Error in configuration file /etc/dovecot/conf.d/10-ssl.conf line 50: ssl_dh: Can't open file /etc/dovecot/dh.pem: No such file or directory
/usr/local/bin/listmailuser: line 12: 1024 *  : syntax error: operand expected (error token is "*  ")
/usr/local/bin/listmailuser: line 12: 1024 *  : syntax error: operand expected (error token is "*  ")
* nathan@XXXXXXXXXXXXX (  /  ) [%]

doveconf: Fatal: Error in configuration file /etc/dovecot/conf.d/10-ssl.conf line 50: ssl_dh: Can't open file /etc/dovecot/dh.pem: No such file or directory
/usr/local/bin/listmailuser: line 12: 1024 *  : syntax error: operand expected (error token is "*  ")
/usr/local/bin/listmailuser: line 12: 1024 *  : syntax error: operand expected (error token is "*  ")
* mike@XXXXXXXXXXXXXXX(  /  ) [%]
    [ aliases -> mike.manfredi@XXXXXXXXXXXXXXXXXXX]

doveconf: Fatal: Error in configuration file /etc/dovecot/conf.d/10-ssl.conf line 50: ssl_dh: Can't open file /etc/dovecot/dh.pem: No such file or directory
/usr/local/bin/listmailuser: line 12: 1024 *  : syntax error: operand expected (error token is "*  ")
/usr/local/bin/listmailuser: line 12: 1024 *  : syntax error: operand expected (error token is "*  ")
* mike@XXXXXXXXXXXXXXX (  /  ) [%]
```

PR #1879, opened 2021-03-31
setup.sh color variables added

casperklein:
# Description

To make the source more readable, color variables have been added.

## Type of change

## Checklist:

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -15,10 +15,7 @@ CYAN="\e[96m"
 BLUE="\e[34m"
 LBLUE="\e[94m"
 BOLD="\e[1m"
-RESET="\e[39m"
-RESET2="\e[0m"
-
-RESET="$RESET2" # for testing
+RESET="\e[0m"
 
 set -euEo pipefail
 trap '__log_err "${FUNCNAME[0]:-?}" "${BASH_COMMAND:-?}" "${LINENO:-?}" "${?:-?}"' ERR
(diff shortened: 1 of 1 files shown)


casperklein:
@aendeavor There are two escape sequences in use. See `$RESET` and `$RESET2`. I think it's safe to only use `$RESET2`. Can you confirm that? In my test, I did not notice any difference.

wernerfred:
Just Reading this on my mobile but didnt we Had Something similar from @aendeavor which was reverted then as colorized Output Breaks with current bats?
Ist this resolved now? Didnt have the PR # in my Kind nur there was a discussion about that already. 
I will See If i can find it again later

aendeavor:
> @aendeavor There are two escape sequences in use. See `$RESET` and `$RESET2`. I think it's safe to only use `$RESET2`. Can you confirm that? In my test, I did not notice any difference.

Yes, it's safe to use `\e[0m` only. This sequence removes _all_ applied attributes, which is desired.

`\e[39m` sets the Color to the default value again - this is why there is no difference here, as we're just working with colours.

aendeavor:
> Just Reading this on my mobile but didnt we Had Something similar from @aendeavor which was reverted then as colorized Output Breaks with current bats?

Not quite the same. In our case, the problem was supervisord not handling the escape sequences properly. As a result, the bats were not able to handle the output.

> Ist this resolved now? Didnt have the PR # in my Kind nur there was a discussion about that already. 

Supervisor can still not handle color... which makes me angry.

aendeavor (approved):

aendeavor:
Very nice. From my PoV, this can be merged:)

Issue #1884, opened 2021-04-05
[BUG] readlink does not support param -f on macos systems


mommel:
### What is affected by this bug?
Only the  setup.sh on mac

### When does this occur?
when useing ./setup.sh 

### How do we replicate the issue?

1. run  ./setup.sh with any param on macos 

## Behavior
Will fail because of the wrong readlink version

### Actual Behavior
Fails with : illegal option -- f

### Expected Behavior
Does not fail with : option -- f

## Your Environment

Macos

### Environment Variables

not relevant

### Relevant Stack Traces
-

aendeavor:
I'd like to see what happens on your system when you run `setup.sh`. Please show the full output when running it.

I'm curious because the `if ... else ...` should catch your issue just fine. It's not clear to me why this does not work (anymore), since I fixed this issue myself for macOS users.

Can you show the output of your shells version and which shell you're using?

reneploetz:
Note that readlink is an external utility which exists in different implementations for GNU and BSD-based systems. BSD-based systems (such as FreeBSD, MacOSX, ...) provide readlink with a different "-f" option. 

See manpages:
GNU: https://linux.die.net/man/1/readlink
MacOS: https://www.unix.com/man-page/osx/1/readlink/

If one decides not to support MacOS X, we do not need to change any code in the project itself as one can either use a bash alias for readlink binary or modify the path like so: https://apple.stackexchange.com/a/69332

wernerfred:
So after #1886 getting merged and the discussion in #1885 together with the answer of @reneploetz can we close #1884 and #1885 @mommel?

PR #1885, opened 2021-04-05
adds switch for macos systems to use greadlink instead

mommel:
Issue on macos is readlink does not support -f greadlink instead does

# Description

readlink on macos does not support -f 

Fixes # (issue)
https://github.com/docker-mailserver/docker-mailserver/issues/1884

## Type of change

## Checklist:

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -36,11 +36,11 @@ ly and use ./setup.sh help and read the VERSION section.\n" >&2
 
 function _use_readlink
 {
-  if uname -a | grep Darwin == "" &>/dev/null
+  if uname -a | grep -q Darwin
   then
-    READLINK="$(readlink -f "${0}")"
-  else
     READLINK="$(greadlink -f "${0}")"
+  else
+    READLINK="$(readlink -f "${0}")"
   fi
 }
 
--- a/setup.sh
+++ b/setup.sh
@@ -34,13 +34,24 @@ the version / tag of docker-mailserver. Please read the
 ly and use ./setup.sh help and read the VERSION section.\n" >&2
 }
 
+<<<<<<< HEAD
 function _use_readlink
+=======
+function _unset_vars
+{
+  unset CDIR CRI INFO IMAGE_NAME CONTAINER_NAME DEFAULT_CONFIG_PATH
+  unset USE_CONTAINER WISHED_CONFIG_PATH CONFIG_PATH VOLUME USE_TTY
+  unset SCRIPT USING_SELINUX
+}
+
+function _get_script_path
+>>>>>>> 513202f (more meaningful names)
 {
   if uname -a | grep -q Darwin
   then
-    READLINK="$(greadlink -f "${0}")"
+    SCRIPT_PATH="$(greadlink -f "${0}")"
   else
-    READLINK="$(readlink -f "${0}")"
+    SCRIPT_PATH="$(readlink -f "${0}")"
   fi
 }
 
--- a/setup.sh
+++ b/setup.sh
@@ -34,11 +34,21 @@ the version / tag of docker-mailserver. Please read the
 ly and use ./setup.sh help and read the VERSION section.\n" >&2
 }
 
+function _use_readlink
+{
+  if uname -a | grep Darwin == "" &>/dev/null
+  then
+    READLINK="$(readlink -f "${0}")"
+  else
+    READLINK="$(greadlink -f "${0}")"
+  fi
+}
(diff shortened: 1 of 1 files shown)


casperklein:
Thank you for the contribution 👍  However there were two bugs:

1. `_use_readlink` was never called
2. The if condition did not work

I fixed both issues. Please review and test my changes on MacOS.

aendeavor:
This PR fixes a problem which should in theory not even exist. Looking at the code, the failure of calling `readlink -f` should have been catched by the `if... else...` construct. @casperklein what's your take?

setup.sh
```diff
@@ -42,11 +42,21 @@ function _unset_vars
   unset SCRIPT USING_SELINUX
 }
 
-function _get_current_directory
+function _get_script_path
```
aendeavor, line 45: This should, from my PoV be a single function call together with `_get_script_directory`. I find the introduction of a new variable, which is not even local (because it cannot be) or `unset` in then end rather misleading.

setup.sh
```diff
@@ -42,11 +42,21 @@ function _unset_vars
...
 {
-  if dirname "$(readlink -f "${0}")" &>/dev/null
+  if uname -a | grep -q Darwin
```
aendeavor, line 47:
``` DIFF
- if uname -a | grep -q Darwin
+ grep -q "Darwin" < <(uname -a)
```

setup.sh
```diff
@@ -42,11 +42,21 @@ function _unset_vars
...
   then
-    CDIR="$(dirname "$(readlink -f "${0}")")"
+    SCRIPT_PATH="$(greadlink -f "${0}")"
```
aendeavor, line 49: We should not use `greadlink`. We do not officially support macOS. The script _should_ work in a way which

1. Checks if `dirname "$(readlink -f "${0}")"` work, if not, go on
2. Checks if `dirname "$(readlink -f "${0}")"` works, if not, use the value provided on initialization, which is
3. The value of `$(pwd)`

aendeavor (changes requested):

setup.sh
casperklein: I thought about this too. But decided to leave it as it is, for the mantra "one function does one thing". That way, $SCRIPT_PATH could also be reused if needed (in the future).

setup.sh
casperklein: Personal taste. I find my variant more readable in that case.

setup.sh
casperklein: I agree. That was one of the reasons, I haven't approved this PR yet. I don't have a Mac to test with. That would imply, that neither `pwd` nor `realpath` is available in @mommel's environment, which sounds strange.

setup.sh
aendeavor: Alright

setup.sh
aendeavor: The good, old piping into `grep` :D Alright, fine with me :)

setup.sh
fbartels: `pwd` does exist on Big Sur, `realpath` does not.

aendeavor: @fbartels Thanks for pointing this out! But as long as one of the three commands work, this should not be a problem in the end, i.e. I still have no clue as to why this should not work :)

casperklein: What about getting rid of the whole `_get_current_directory` function? 😄 

`CDIR=${0%/*}` should all we need. Pure Bash and no dependencies to other binarys.

aendeavor: > What about getting rid of the whole `_get_current_directory` function? smile
> 
> `CDIR=${0%/*}` should all we need. Pure Bash and no dependencies to other binarys.

Quite nice, but only under the condition that working with relative paths is okay. I do not see an problems at the moment, but I'm not 100% confident either.

casperklein: Yes, I've considered this. In the current code, there is no need for an absolute path.

aendeavor: > Yes, I've considered this. In the current code, there is no need for an absolute path.

Although we're "losing" the capability to resolve softlinks, this should not be an issue altogether. I will close this PR. Can you provide a new one which implements this behavior?

casperklein: Unfortunately, an absolute path is required for `docker -v` command :confused:

aendeavor:
This would need rebasing. But I guess opening a new PR (if still needed) is easier.

Issue #1821, opened 2021-02-21
[Review/Improve] fail2ban service


casperklein:
# Review/Improve Request

## Context

At the moment, the fail2ban service seems not to operate optimal.

### Is your Feature Request related to a Problem?

At least:

https://github.com/docker-mailserver/docker-mailserver/pull/1661
https://github.com/docker-mailserver/docker-mailserver/issues/1810

### Describe the Solution you'd like

The whole fail2ban implementation should be reviewed and optimized.

### Are you going to implement it?

No (at least, not in short-term). This is more for tracking/discussing the topic.

### What are you going to contribute?

Lifetime 😄

# Things to discuss / clarify / To-do

- ~~Is `https://github.com/docker-mailserver/docker-mailserver/blob/master/target/fail2ban/jail.conf` still needed, what was the original intent not to use the default shipped configuration?~~ #1895 
- ~~What services are covered by fail2ban?~~ dovecot, postfix, postfix-sasl
- ~~Should an attacker be blocked completely? Currently, the IP address is only blocked for a specific service. IMHO if someone tries to brute force POP3, the IP address should also be blocked for IMAP etc, or even better completely.~~
- ~~Default ban behavior of fail2ban is to REJECT packets. Maybe DROP is better?~~
- ~~Check if `update-alternatives --set iptables /usr/sbin/iptables-legacy # switch iptables and ip6tables to legacy for Fail2Ban` from Dockerfile is (still) needed~~

There are probably other things to consider. Just let me know and I'll keep this list up-to-date.

### Block all ports instead of service port only
`sed -i '/^banaction = iptables-multiport$/ s|iptables-multiport|iptables-allports|' /etc/fail2ban/jail.conf`

### DROP connections instead of REJECT
`sed -i 's/^blocktype = REJECT.*/blocktype = DROP/' /etc/fail2ban/action.d/iptables-common.conf`

### Proposed configuration options

| Variable | Values | Default | Description |
| --- | --- | ---  | --- |
| FAIL2BAN_BLOCK_TYPE | REJECT, DROP | DROP | iptables block behaviour; `REJECT` sends *ICMP unreachable*; `DROP ` sends no reply |
| FAIL2BAN_BLOCK_ALL | true, false | true | If *true*, block an attacker on all ports, not only the attacked one. |
| FAIL2BAN_BLOCK_PERMANENT | true, false | false | Block an attacker forever |

aendeavor:
[...]

PR #1895, opened 2021-04-10
fail2ban cleanup

casperklein:
# Description

This is the first part from #1821. This is a cleanup / reorganization, so no breaking change is expected.

The file `target/fail2ban/jail.conf` has been deleted. Now the more recent debian `jail.conf` is used.
The default customization for this project takes now place in `target/fail2ban/jail.local`.
User customization from `/tmp/docker-mailserver/fail2ban-jail.cf` is now reflected in `/etc/fail2ban/jail.d/user-jail.local`.

The file `target/fail2ban/filter.d/postfix-sasl.conf` has been deleted, because its no longer needed. I guess, the file was used in the past, because fail2ban had no default config for postfix-sasl? In my manual tests, bans worked perfect for postfix-sasl. Also all bat tests passed successfully.

The file `config/fail2ban-fail2ban.cf` has been updated with the current debian configuration as example.
The file `config/fail2ban-jail.cf` has been updated with the current debian defaults.

On debian, the sshd jail is activated automatically. As we don't use sshd at all, I've disabled this jail and adjusted the corresponding bat test.

## Type of change

## Checklist:


wernerfred:
Are Updates to Docs needed? Does Overriding Files and fail2ban section mention one of the deleted Files?

wernerfred (commented):
Are Updates to Docs needed? Does Overriding Files and fail2ban section mention one of the deleted Files?

casperklein:
These are only internal changes. From a users perspective, there is no change. The two deleted files are all in `target/`, so only used for image building.

Dockerfile
```diff
@@ -172,9 +172,11 @@ RUN for _ in {1..10}; do su - amavis -c "razor-admin -create" ; sleep 3 ; \
 # ––– Fail2Ban, DKIM & DMARC ––––––––––––––––––––
 # –––––––––––––––––––––––––––––––––––––––––––––––
 
-COPY target/fail2ban/jail.conf /etc/fail2ban/jail.conf
-COPY target/fail2ban/filter.d/postfix-sasl.conf /etc/fail2ban/filter.d/postfix-sasl.conf
-RUN mkdir /var/run/fail2ban
+COPY target/fail2ban/jail.local /etc/fail2ban/jail.local
+RUN ln -s /var/log/mail/mail.log /var/log/mail.log && \
```
aendeavor, line 176: Just styling: We use

``` DOCKERFILE
RUN \
  do_something_command \
  do_some_more_command
```

i.e. use the `\` first when multi-line `RUN` statements appear:)

aendeavor:
From what I can see: Nice! Just this one style-related change and I'm good :D

Dockerfile
casperklein: Oh, I must have overlooked that, thx.

casperklein: My mistake was, that I looked at this example directly above my change, which seems also not correct then? https://github.com/docker-mailserver/docker-mailserver/blob/34c6e46f68f5de60ac08f815252f4e1632eb78b0/Dockerfile#L167-L169

casperklein: I've fixed both styling issues.

wernerfred (approved):

aendeavor (approved):

PR #1897, opened 2021-04-11
Mailserver needs recreation after adding new volume

shyim:
# Description

With just an restart the volume are missing and the follow up command will fail

## Type of change

## Checklist:

### DIFF
--- a/docs/content/config/advanced/full-text-search.md
+++ b/docs/content/config/advanced/full-text-search.md
@@ -46,7 +46,7 @@ The [dovecot-solr Plugin](https://wiki2.dovecot.org/Plugins/FTS/Solr) is used in
 
 3. Start the solr container: `docker-compose up -d --remove-orphans solr`
 
-4. Restart the mailserver container: `docker-compose restart mailserver`
+4. Recreating the mailserver container: `docker-compose up -d`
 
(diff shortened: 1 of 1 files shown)


aendeavor:
Am I missing something or does the `docker-compose down` not appear? Is this needed?

shyim:
What do you mean? When the config is changed and you call up it will recreate the container 🤔

aendeavor:
> What do you mean? When the config is changed and you call up it will recreate the container thinking

I'm running Swarm for so long that I don't really remember compose anymore. This update seems fine.

aendeavor (approved):

casperklein:
@aendeavor is right. If the container is already running, `docker-compose up -d` will do nothing.

`docker-compose down mailserver && docker-compose up -d mailserver` is the correct way IMO.

aendeavor:
> @aendeavor is right. If the container is already running, `docker-compose up -d` will do nothing.
> 
> 
> 
> ~~`docker-compose down mailserver && docker-compose up -d mailserver` is the correct way IMO.~~
> 
> 
> 
> Edit:
> 
> 
> 
> `docker-compose down` is not possible for just a single service. So the best way would just be a complete `docker-compose down; docker-compose up -d`

@casperklein Thank you for clarifying. Could you provide a small follow-up PR?

Issue #805, opened 2018-01-27
What is DOVECOT_PASS_FILTER and DOVECOT_USER_FILTER?


jeffwji:
I have been very confused by this two variables:
``` 
     - DOVECOT_PASS_FILTER="(&(objectClass=PostfixBookMailAccount)(uniqueIdentifier=%n))"
     - DOVECOT_USER_FILTER="(&(objectClass=PostfixBookMailAccount)(uniqueIdentifier=%n))"
```

I'm trying to setup the mailbox against with AD, I think I should have something like
```
      - LDAP_QUERY_FILTER_USER="(&(objectCategory=Person)(sAMAccountName=%s)(memberOf=cn=emailusers,dc=company,dc=com))"
```

But what is DOVECOT_PASS_FILTER and DOVECOT_USER_FILTER? `start-mailserver.sh` has this contents from line 558:
```
	# Not sure whether this can be the same or not
	# _dovecot_ldap_mapping["DOVECOT_PASS_FILTER"]="${DOVECOT_PASS_FILTER:="${LDAP_QUERY_FILTER_USER}"}"
	# _dovecot_ldap_mapping["DOVECOT_USER_FILTER"]="${DOVECOT_USER_FILTER:="${LDAP_QUERY_FILTER_USER}"}
```

Which has been commented out, and even funny thing is looks like the script author himself is not sure how to use this two variables?

dolanor:
Well, I've been fighting it also, and I'm taming the beast as we speak.
So, yes, that [comment](https://github.com/tomav/docker-mailserver/blob/master/target/start-mailserver.sh#L558) is valid, depending on the point of view.

Can DOVECOT_PASS_FILTER and DOVECOT_USER_FILTER be the same ? I'd say yes. This is what I do (with an OpenLDAP, but should be the same for AD)
Can DOVECOT_PASS_FILTER and DOVECOT_USER_FILTER use the same value as LDAP_QUERY_FILTER_USER ? Definitely NO.

LDAP_QUERY_FILTER_USER use %s as a value passed to it. %s is simply the user string passed to postfix (and handled via saslauthd). So it can be the username, or the user@domain.com email
DOVECOT_PASS_FILTER and DOVECOT_USER_FILTER can use [dovecot variables](https://wiki2.dovecot.org/Variables). So you can use %u, %d, %n if you want the username, the full email, or the domain.

jeffwji:
[...]

PR #1901, opened 2021-04-12
Use dovecot's LDAP uris option instead of hosts (fixes #1510)

### LINKED ISSUE #1510: Dovecot not working with ldaps:// protocol
When setting e.g. `LDAP_SERVER_HOST=ldaps://farafin.de:636`, Dovecot can't authenticate any users (`dovecot: auth: Error: LDAP: Can't connect to server: ldaps://example.org:636`).<

## Context

I have an LDAP server that only allows TLS connections - the mailserver works great with that, except for IMAP (which fails with the above error message).

### *What* is affected by this bug?
IMAP authentication

### *When* does this occur?
When using LDAP over TLS (not STARTTLS!).

## *How* do we replicate the issue?

1. Set up an LDAP server with TLS
2. Set LDAP_SERVER_HOST to `ldaps://your-ldap-server-host:636`
3. Try to authenticate against the mailserver with correct credentials from the LDAP server

## Actual Behavior

The authentication fails.

## Expected behavior (i.e. solution)

The authentication succeeds.

## Possible Solution
If the LDAP_SERVER_HOST contains `://`, the Dovecot config (`/etc/dovecot/dovecot-ldap.conf.ext`) should use the `uris` instead of the `hosts` parameter (see https://doc.dovecot.org/configuration_manual/authentication/ldap/).

I'm using v7.0.0.

moqmar:
# Description

According to https://doc.dovecot.org/configuration_manual/authentication/ldap/, the "uris" option has to be used when using LDAPS, while the "hosts" option seems to only work with STARTTLS. So, this PR fixes the problem that docker-mailserver can't be used with LDAPS.

The change is made in a way that allows backwards-compatibility with configurations that assume the "hosts" option to be set by LDAP_SERVER_HOST, in which case it's using "ldap://" as a default protocol if none is specified. The only change for the user is that LDAP_SERVER_HOST now supports specifying a protocol like "ldaps://...".

Fixes #1510

## Type of change

## Checklist:

### DIFF
--- a/ENVIRONMENT.md
+++ b/ENVIRONMENT.md
@@ -379,7 +379,7 @@ Note: The defaults of your fetchmailrc file need to be at the top of the file. O
 ##### LDAP_SERVER_HOST
 
 - **empty** => mail.domain.com
-- => Specify the dns-name/ip-address where the ldap-server
+- => Specify the dns-name/ip-address where the ldap-server is listening, or an URI like `ldaps://mail.domain.com`
 - NOTE: If you going to use the mailserver in combination with docker-compose you can set the service name here
 
 ##### LDAP_SEARCH_BASE
--- a/ENVIRONMENT.md
+++ b/ENVIRONMENT.md
@@ -446,10 +446,12 @@ The following variables overwrite the default values for ```/etc/dovecot/dovecot
 - **empty** => same as `LDAP_BIND_PW`
 - => Password for LDAP dn sepecifified in `DOVECOT_DN`.
 
-##### DOVECOT_HOSTS
+##### DOVECOT_URIS
 
 - **empty** => same as `LDAP_SERVER_HOST`
-- => Specify a space separated list of LDAP hosts.
+- => Specify a space separated list of LDAP uris.
+- Note: If the protocol is missing, `ldap://` will be used.
+- Note: This deprecates `DOVECOT_HOSTS` (as it didn't allow to use LDAPS), which is currently still supported for backwards compatibility.
 
 ##### DOVECOT_LDAP_VERSION
 
@@ -477,6 +479,7 @@ The following variables overwrite the default values for ```/etc/dovecot/dovecot
 ##### DOVECOT_PASS_FILTER
 
 - e.g. `(&(objectClass=PostfixBookMailAccount)(uniqueIdentifier=%n))`
+- **empty** => same as `DOVECOT_USER_FILTER`
 
 ##### DOVECOT_PASS_ATTRS
 
--- a/target/dovecot/dovecot-ldap.conf.ext
+++ b/target/dovecot/dovecot-ldap.conf.ext
@@ -2,7 +2,7 @@ base                = ou=people,dc=domain,dc=com
 default_pass_scheme = SSHA
 dn                  = cn=admin,dc=domain,dc=com
 dnpass              = admin
-hosts               = mail.domain.com
+uris                = ldap://mail.domain.com
 tls                 = no
 ldap_version        = 3
 pass_attrs          = uniqueIdentifier=user,userPassword=password
--- a/target/scripts/startup/setup-stack.sh
+++ b/target/scripts/startup/setup-stack.sh
@@ -437,10 +437,17 @@ function _setup_ldap
 
   declare -A _dovecot_ldap_mapping
 
+  # Add protocol to LDAP_SERVER_HOST so that we can use dovecot's "uris" option:
+  # https://doc.dovecot.org/configuration_manual/authentication/ldap/
+  dovecot_ldap_server_uri="${LDAP_SERVER_HOST}"
+  if ! echo "${dovecot_ldap_server_uri}" | grep -F '://' >/dev/null; then
+    dovecot_ldap_server_uri="ldap://${dovecot_ldap_server_uri}"
+  fi
+
   _dovecot_ldap_mapping["DOVECOT_BASE"]="${DOVECOT_BASE:="${LDAP_SEARCH_BASE}"}"
   _dovecot_ldap_mapping["DOVECOT_DN"]="${DOVECOT_DN:="${LDAP_BIND_DN}"}"
   _dovecot_ldap_mapping["DOVECOT_DNPASS"]="${DOVECOT_DNPASS:="${LDAP_BIND_PW}"}"
-  _dovecot_ldap_mapping["DOVECOT_HOSTS"]="${DOVECOT_HOSTS:="${LDAP_SERVER_HOST}"}"
+  _dovecot_ldap_mapping["DOVECOT_URIS"]="${DOVECOT_URIS:="${dovecot_ldap_server_uri}"}"
 
   # Not sure whether this can be the same or not
   # _dovecot_ldap_mapping["DOVECOT_PASS_FILTER"]="${DOVECOT_PASS_FILTER:="${LDAP_QUERY_FILTER_USER}"}"
--- a/target/scripts/startup/setup-stack.sh
+++ b/target/scripts/startup/setup-stack.sh
@@ -435,27 +435,26 @@ function _setup_ldap
 
   _notify 'inf' "Configuring dovecot LDAP"
 
-  declare -A _dovecot_ldap_mapping
+  declare -A DOVECOT_LDAP_MAPPING
 
-  # Add protocol to LDAP_SERVER_HOST so that we can use dovecot's "uris" option:
+  DOVECOT_LDAP_MAPPING["DOVECOT_BASE"]="${DOVECOT_BASE:="${LDAP_SEARCH_BASE}"}"
+  DOVECOT_LDAP_MAPPING["DOVECOT_DN"]="${DOVECOT_DN:="${LDAP_BIND_DN}"}"
+  DOVECOT_LDAP_MAPPING["DOVECOT_DNPASS"]="${DOVECOT_DNPASS:="${LDAP_BIND_PW}"}"
+  DOVECOT_LDAP_MAPPING["DOVECOT_URIS"]="${DOVECOT_URIS:="${DOVECOT_HOSTS:="${LDAP_SERVER_HOST}"}"}"
+
+  # Add protocol to DOVECOT_URIS so that we can use dovecot's "uris" option:
   # https://doc.dovecot.org/configuration_manual/authentication/ldap/
-  dovecot_ldap_server_uri="${LDAP_SERVER_HOST}"
-  if ! echo "${dovecot_ldap_server_uri}" | grep -F '://' >/dev/null; then
-    dovecot_ldap_server_uri="ldap://${dovecot_ldap_server_uri}"
+  if [[ "${DOVECOT_LDAP_MAPPING["DOVECOT_URIS"]}" != *'://'* ]]
+  then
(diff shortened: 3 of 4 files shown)


target/scripts/startup/setup-stack.sh
```diff
@@ -437,10 +437,17 @@ function _setup_ldap
...
+  dovecot_ldap_server_uri="${LDAP_SERVER_HOST}"
+  if ! echo "${dovecot_ldap_server_uri}" | grep -F '://' >/dev/null; then
+    dovecot_ldap_server_uri="ldap://${dovecot_ldap_server_uri}"
+  fi
+
   _dovecot_ldap_mapping["DOVECOT_BASE"]="${DOVECOT_BASE:="${LDAP_SEARCH_BASE}"}"
   _dovecot_ldap_mapping["DOVECOT_DN"]="${DOVECOT_DN:="${LDAP_BIND_DN}"}"
   _dovecot_ldap_mapping["DOVECOT_DNPASS"]="${DOVECOT_DNPASS:="${LDAP_BIND_PW}"}"
-  _dovecot_ldap_mapping["DOVECOT_HOSTS"]="${DOVECOT_HOSTS:="${LDAP_SERVER_HOST}"}"
+  _dovecot_ldap_mapping["DOVECOT_URIS"]="${DOVECOT_URIS:="${dovecot_ldap_server_uri}"}"
```
aendeavor, line 450: NOTE: This note is for maintainers, and should not affect this PR.

This whole section was never worked on, therefore the old style is present. This should not be a problem as of now, but we should re-write this in the future to have consistent styling.

target/dovecot/dovecot-ldap.conf.ext
```diff
@@ -2,7 +2,7 @@ base                = ou=people,dc=domain,dc=com
 default_pass_scheme = SSHA
 dn                  = cn=admin,dc=domain,dc=com
 dnpass              = admin
-hosts               = mail.domain.com
+uris                = mail.domain.com
```
wernerfred, line 5: Doesn't this need a protocol then?

fbartels: Yes, ihmo this is missing the protocol
https://doc.dovecot.org/configuration_manual/authentication/ldap/

moqmar: Yes, it should be `ldaps://mail.domain.com` - it's replaced though by the config script, so it shouldn't influence functionality.

moqmar: I have fixed that now, see https://github.com/docker-mailserver/docker-mailserver/compare/6ac6c241039efd1a22c5809a5434978d37585e94..0a3ab8767f7bddc0a4491494d95cec1173a8d9f2

target/scripts/startup/setup-stack.sh
```diff
@@ -437,10 +437,17 @@ function _setup_ldap
 
   declare -A _dovecot_ldap_mapping
 
+  # Add protocol to LDAP_SERVER_HOST so that we can use dovecot's "uris" option:
+  # https://doc.dovecot.org/configuration_manual/authentication/ldap/
+  dovecot_ldap_server_uri="${LDAP_SERVER_HOST}"
+  if ! echo "${dovecot_ldap_server_uri}" | grep -F '://' >/dev/null; then
```
aendeavor, line 443: Should be `grep -q` instead of using `>/dev/null`

target/scripts/startup/setup-stack.sh
```diff
@@ -437,10 +437,17 @@ function _setup_ldap
 
   declare -A _dovecot_ldap_mapping
```
aendeavor, line 438: We use uppercase variables, see CONTRIBUTING.md for styling

aendeavor (changes requested):

target/scripts/startup/setup-stack.sh
moqmar: I'm using `grep -F '://'` as your version would replace `ldaps://example.org` with `ldap://ldaps://example.org` - we just need any protocol. the `-F` is for fixed strings (no regex), so I could probably also use a bash comparison here.

moqmar: Would `[[ ${DOVECOT_LDAP_SERVER_URI} != *://* ]]` be fine?

aendeavor: Probably a good idea to use a Bash comparison.

target/scripts/startup/setup-stack.sh
moqmar: Are underscore prefixes for internal variables still preferred, or should this just be DOVECOT_LDAP_MAPPING?

aendeavor: It should just be `DOVECOT_LDAP_MAPPING`. Underscores are for function :)

target/scripts/startup/setup-stack.sh
```diff
@@ -437,10 +437,17 @@ function _setup_ldap
...
+    dovecot_ldap_server_uri="ldap://${dovecot_ldap_server_uri}"
+  fi
+
   _dovecot_ldap_mapping["DOVECOT_BASE"]="${DOVECOT_BASE:="${LDAP_SEARCH_BASE}"}"
   _dovecot_ldap_mapping["DOVECOT_DN"]="${DOVECOT_DN:="${LDAP_BIND_DN}"}"
   _dovecot_ldap_mapping["DOVECOT_DNPASS"]="${DOVECOT_DNPASS:="${LDAP_BIND_PW}"}"
-  _dovecot_ldap_mapping["DOVECOT_HOSTS"]="${DOVECOT_HOSTS:="${LDAP_SERVER_HOST}"}"
+  _dovecot_ldap_mapping["DOVECOT_URIS"]="${DOVECOT_URIS:="${dovecot_ldap_server_uri}"}"
 
   # Not sure whether this can be the same or not
```
moqmar, line 452: That was discussed in #805, so I guess I can remove the three lines here as well if I'm at it? I could also make `DOVECOT_PASS_FILTER` optional (as it seems to often be the same), but would still leave it configurable separately:

```bash
DOVECOT_LDAP_MAPPING["DOVECOT_PASS_FILTER"]="${DOVECOT_PASS_FILTER:="${DOVECOT_USER_FILTER}"}"
```

target/scripts/startup/setup-stack.sh
moqmar: I just noticed that this deprecates the DOVECOT_HOSTS variable, I'll change it so that it's still supported?

target/scripts/startup/setup-stack.sh
aendeavor: Yeah, I guess :D #805 is rather old, and we can definitely change this by now.

target/scripts/startup/setup-stack.sh
aendeavor: > I just noticed that this deprecates the DOVECOT_HOSTS variable, I'll change it so that it's still supported?
> Edit: I markeed DOVECOT_HOSTS as deprecated and use it as a fallback value for DOVECOT_URIS now.

Exactly. We cannot get rid of this in an instant as we need to ensure backwards compatibility in a major version. When 10.0.0 is release, we can change this. The deprecation notice is the absolutely right way.

aendeavor (dismissed):
I can approve this PR style-wise. I cannot say anything about the functionality. I re-requested a review from @wernerfred

moqmar:
Whoops, seems like I have broken something with the style changes. Will evaluate. 🙃

aendeavor:
Please also note that your commit signatures are invalid - you may want to fix this as well :D

wernerfred:
@reneploetz you commented on #1902. Can you please review here as well?

reneploetz:
> @reneploetz you commented on #1902. Can you please review here as well?

As I'm not using dovecot + LDAP I'm unable to test this on an actual test system (my system is using Kopano + LDAP). But the change *appears* to fine to be given that there is at least on test that is checking the correct replacement inside the config after the container start as well as a test that dovecot starts with that configuration and that one can connect with a username that is present in ldap.
If you feel uncomfortable to merge this then I can create local test environment for this tough.

aendeavor:
> If you feel uncomfortable to merge this then I can create local test environment for this tough.

That'd be awesome!

moqmar:
I think the most interesting test would be LDAP with StartTLS, as I think that it hasn't been tested yet with the change.

wernerfred:
> If you feel uncomfortable to merge this then I can create local test environment for this tough.

If you could to this it would be great! If you do not find time then I would consider it safe to merge as we have the same opinion by just looking at the changes. But only a real test can show if we are right.

Thanks in advance!

aendeavor:
> If you could to this it would be great! If you do not find time then I would consider it safe to merge as we have the same opinion by just looking at the changes. But only a real test can show if we are right.

I agree.

aendeavor:
@wernerfred If you're comfortable merging this, I am too. I will update the branch first :)

wernerfred:
I would wait for a reply of @reneploetz :)
Depending on this we will merge or wait until he finished testing.

reneploetz:
I can confirm that the following configurations are working:
```
LDAP_SERVER_HOST=testing_ldap
DOVECOT_TLS=yes
```

```
DOVECOT_URIS=ldap://testing_ldap
DOVECOT_TLS=yes
```

All of them lead to this content in dovecot-ldap.conf.ext which allows dovecot to successfully connect to LDAP via TLS:
```
uris = ldap://testing_ldap
tls = yes
```

So as far as I'm concerned the configuration works with TLS.

As a side note in case somebody asks later: we currently do not support custom certificates for LDAP due to the lack of "tls_require_cert" or "tls_ca_cert_*" settings in the configuration.

aendeavor (approved):
Then I will approve these changes. @wernerfred if you're fine with this too, go ahead and merge this.

We can afterwards merge the PR about Gh actions

ENVIRONMENT.md
```diff
@@ -457,10 +457,12 @@ The following variables overwrite the default values for ```/etc/dovecot/dovecot
...
 - => Password for LDAP dn sepecifified in `DOVECOT_DN`.
 
-##### DOVECOT_HOSTS
+##### DOVECOT_URIS
 
 - **empty** => same as `LDAP_SERVER_HOST`
-- => Specify a space separated list of LDAP hosts.
+- => Specify a space separated list of LDAP uris.
+- Note: If the protocol is missing, `ldap://` will be used.
+- Note: This deprecates `DOVECOT_HOSTS` (as it didn't allow to use LDAPS), which is currently still supported for backwards compatibility.
```
wernerfred, line 465: This is actually a breaking change. @aendeavor another reason for `v10.0.0` 😄 
The note should be removed at some time in the future.

wernerfred (approved):

wernerfred:
Does this change require documentation updates?

wernerfred:
Ah @moqmar proposed to do this with a separate PR in https://github.com/docker-mailserver/docker-mailserver/pull/1902#issuecomment-821121746. 
Therefore I will merge

ENVIRONMENT.md
aendeavor: Yes - I said this earlier that this note will go with 10.0.0

PR #1909, opened 2021-04-14
Comments added

casperklein:
# Description

- Comments added
- ESMTP (implicit TLS) added

## Type of change

## Checklist:

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,10 +9,11 @@ services:
     env_file: mailserver.env
     # To avoid conflicts with yaml base-60 float, DO NOT remove the quotation marks.
     ports:
-      - "25:25"
-      - "143:143"
-      - "587:587"
-      - "993:993"
+      - "25:25"    # SMTP  (explicit TLS => STARTTLS)
+      - "143:143"  # IMAP4 (explicit TLS => STARTTLS)
+      - "465:465"  # ESMTP (implicit TLS)
+      - "587:587"  # ESMTP (explicit TLS => STARTTLS)
+      - "993:993"  # IMAP4 (implicit TLS)
     volumes:
       - maildata:/var/mail
       - mailstate:/var/mail-state
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,6 +8,8 @@ services:
     container_name: ${CONTAINER_NAME}
     env_file: mailserver.env
     # To avoid conflicts with yaml base-60 float, DO NOT remove the quotation marks.
+    # More information about the mailserver ports:
+    # https://docker-mailserver.github.io/docker-mailserver/edge/config/security/understanding-the-ports/
     ports:
       - "25:25"    # SMTP  (explicit TLS => STARTTLS)
       - "143:143"  # IMAP4 (explicit TLS => STARTTLS)


fbartels (approved):

wernerfred (changes requested):
Not sure If this is really needed as Ports are imho absolute basics If someone wants to Host a Mailserver. (You have to Provide Them in the Client as well). Here it messes Up the Compose File although the Ports are Well documented all over the documentation.

Either way i can live with this Change but please avoid comments in the Same Line and Change Them to the Line above the Ports. This way the File can be parsed properly.

polarathene (dismissed):
I don't mind it, it's an example config anyway and may be useful inline. What was the motivation for adding these into the config vs a comment linking to the docs which details the ports?

Do we not provide awareness well enough on this, is it just a quick refresher reference, or will a user unfamiliar with the ports wanting to setup mailserver make any sense of it? Whom does it benefit?

If the intention is to clarify the ports to someone who is wondering why the config has listed a bunch that someone might not know what they're for, add a 2nd comment linking to docs for the `ports`. The user is otherwise going to look these up anyway, or vaguely recalls them but unsure of which was for what protocol or implicit TLS vs STARTTLS.

casperklein:
> is it just a quick refresher reference

Exactly. I found it handy, having the port description inline and not to have to look it up once in a while (when I am unsure).

> This way the File can be parsed properly.

What do you mean. This is valid yaml. I intentionally took this form, because I found it more clear in that case. (I had your variant before, but found this approach better).

wernerfred:
> What do you mean. This is valid yaml. I intentionally took this form, because I found it more clear in that case. (I had your variant before, but found this approach better).

Sure it is valid. I just ran too often into issues that have their origin in inline comments. It is not easy to remove comments for example by deleting every line that starts with `#`. Also I had issues with inline comments when parsing yaml files with other scripts for example. 
That are surely "my" problems or the ones from ppl that do not just simply use the project but those problems can easily avoided by placing comments in separate lines.

Like I said, it is not a dealbreaker for me but if there is no real reason behind it I would welcome comments in separate lines. 😃

polarathene:
> Like I said, it is not a dealbreaker for me but if there is no real reason behind it I would welcome comments in separate lines.

If your ok with it, I think your request for changes review status is blocking a merge?

aendeavor (approved):

aendeavor:
I approve this as well:)

wernerfred (approved):

docker-compose.yml
```diff
@@ -8,11 +8,14 @@ services:
...
     ports:
-      - "25:25"
-      - "143:143"
-      - "587:587"
-      - "993:993"
+      - "25:25"    # SMTP  (explicit TLS => STARTTLS)
+      - "143:143"  # IMAP4 (explicit TLS => STARTTLS)
+      - "465:465"  # ESMTP (implicit TLS)
+      - "587:587"  # ESMTP (explicit TLS => STARTTLS)
+      - "993:993"  # IMAP4 (implicit TLS)
```
williamdes, line 18: Very nice change !
That said 3 ports are still missing
```yml
            - "110:110" # POP3
            - "995:995" # POP3
            - "4190:4190" # Sieve filters
```

wernerfred: No not really, `docker-compose.yml` is a **not** feature complete example of the mailserver.
Keep in mind that by default POP3 and SIEVE are not enabled https://github.com/docker-mailserver/docker-mailserver/blob/a0f4a375124b3b3402d5639a7d422501c2011b97/mailserver.env#L107-L108 and https://github.com/docker-mailserver/docker-mailserver/blob/a0f4a375124b3b3402d5639a7d422501c2011b97/mailserver.env#L80-L81

Therefore imho it is correct to not list these ports here

williamdes: Yes, but they could be added and commented ;)

Issue #1904, opened 2021-04-14
Encryption of maildata files and folders?


NorseGaud:
# Subject

I'm curious what is recommended for encryption of the emails/text files stores under the maildata volume. I'd like to back these up to S3, but want to prevent them from being read without a key. Can dovecot or postfix even use a key for decryption and encryption?

Thanks!

NorseGaud:
This looks promising: https://doc.dovecot.org/configuration_manual/mail_crypt_plugin/

NorseGaud:
Ok, I tried mail-crypt but it's problematic...

I ran:

```
cd $HOME
openssl ecparam -name prime256v1 -genkey | openssl pkey -out ecprivkey.pem
openssl pkey -in ecprivkey.pem -pubout -out ecpubkey.pem
```

I edited `$HOME/config/dovecot.cf` to:

```
mail_max_userip_connections = 120
  
mail_plugins = $mail_plugins mail_crypt
plugin {
  mail_crypt_global_private_key = </home/ec2-user/ecprivkey.pem
  mail_crypt_global_public_key = </home/ec2-user/ecpubkey.pem
  mail_crypt_save_version = 2
}

```

I see in the logs now:


```
deliver       | Apr 15 02:43:29 ip-172-31-0-35 postfix/submission/smtpd[19059]: warning: inet-MY IP-1.bos.netblazr.com[64.112.178.1]: SASL PLAIN authentication failed: generic failure
deliver       | Apr 15 02:43:29 ip-172-31-0-35 postfix/submission/smtpd[19059]: warning: SASL authentication failure: cannot connect to saslauthd server: No such file or directory
deliver       | Apr 15 02:43:29 ip-172-31-0-35 postfix/submission/smtpd[19059]: warning: inet.MY IP.bos.netblazr.com[64.112.178.1]: SASL LOGIN authentication failed: generic failure
. . .
deliver       | Apr 15 02:44:08 ip-172-31-0-35 postfix/smtpd[22684]: NOQUEUE: reject: RCPT from mail-qk1-f176.google.com[209.85.222.176]: 451 4.3.5 <MY EEMAIL>: Recipient address rejected: Server configuration problem; from=<connarpierce@gmail.com> to=<MY EMAIL> proto=ESMTP helo=<mail-qk1-f176.google.com>
[...]

PR #1913, opened 2021-04-16
[Documentation] mail_crypt plugin

NorseGaud:
# Description

Adding docs for mail_crypt

Related: #1904

## Type of change

## Checklist:

### DIFF
--- a/docs/content/config/advanced/override-defaults/dovecot.md
+++ b/docs/content/config/advanced/override-defaults/dovecot.md
@@ -32,6 +32,8 @@ This limit is quickly reached if users connect to the mail server with multiple 
 
 For major configuration changes it’s best to override the dovecot configuration files. For each configuration file you want to override, add a list entry under the `volumes` key.
 
+You will need to first obtain the configuration from the running container: `mkdir -p ./config/dovecot && docker cp <your-container-name>:/etc/dovecot/conf.d/10-master.conf ./config/dovecot/10-master.conf`
+
 ```yaml
 services:
   mail:
--- a/docs/content/config/advanced/override-defaults/dovecot.md
+++ b/docs/content/config/advanced/override-defaults/dovecot.md
@@ -32,7 +32,7 @@ This limit is quickly reached if users connect to the mail server with multiple 
 
 For major configuration changes it’s best to override the dovecot configuration files. For each configuration file you want to override, add a list entry under the `volumes` key.
 
-You will need to first obtain the configuration from the running container: `mkdir -p ./config/dovecot && docker cp <your-container-name>:/etc/dovecot/conf.d/10-master.conf ./config/dovecot/10-master.conf`
+You will need to first obtain the configuration from the running container: `mkdir -p ./config/dovecot && docker cp mailserver:/etc/dovecot/conf.d/10-master.conf ./config/dovecot/10-master.conf`
 
 ```yaml
 services:
@@ -47,7 +47,7 @@ services:
 To debug your dovecot configuration you can use:
 
 - This command: `./setup.sh debug login doveconf | grep <some-keyword>`
-- Or: `docker exec -it <your-container-name> doveconf | grep <some-keyword>`
+- Or: `docker exec -it mailserver doveconf | grep <some-keyword>`
 
 !!! note
     [`setup.sh`][github-file-setupsh] is included in the `docker-mailserver` repository. Make sure to grap the one matching your image version.
@@ -55,7 +55,7 @@ To debug your dovecot configuration you can use:
 The `config/dovecot.cf` is copied internally to `/etc/dovecot/local.conf`. To check this file run:
 
 ```sh
-docker exec -it <your-container-name> cat /etc/dovecot/local.conf
+docker exec -it mailserver cat /etc/dovecot/local.conf
 ```
 
 [github-file-setupsh]: https://github.com/docker-mailserver/docker-mailserver/blob/master/setup.sh
--- a/docs/content/config/security/mail_crypt.md
+++ b/docs/content/config/security/mail_crypt.md
@@ -1,0 +1,77 @@
+---
+title: 'Security | mail_crypt (email/storage encryption)'
+---
+
+!!! info
+ 
+    The Mail crypt plugin is used to secure email messages stored in a Dovecot system. Messages are encrypted before written to storage and decrypted after reading. Both operations are transparent to the user.
+
+    In case of unauthorized access to the storage backend, the messages will, without access to the decryption keys, be unreadable to the offending party.
+
+    There can be a single encryption key for the whole system or each user can have a key of their own. The used cryptographical methods are widely used standards and keys are stored in portable formats, when possible.
+
+Official Dovecot documentation: https://doc.dovecot.org/configuration_manual/mail_crypt_plugin/
+
+---
+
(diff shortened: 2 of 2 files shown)


docs/content/config/advanced/override-defaults/dovecot.md
```diff
@@ -30,15 +30,20 @@ This limit is quickly reached if users connect to the mail server with multiple
...
+1. Add a list entry under the `volumes` key for a temporary/empty folder:
+    ```yaml
+    services:
+      mail:
+        volumes:
+          - maildata:/var/mail
+          - ./config/dovecot:/etc/dovecot/conf.d-tmp
+    ```
+2. You can then start the container with `docker-compose up`
+3. Once started, execute `docker exec -it {id of container} bash -c "cp -rfp /etc/dovecot/conf.d/* /etc/dovecot/conf.d-tmp/"` and then `docker-compose down`.
```
casperklein, line 44: Instead of `{id of container}`, `mailserver` could be used, as this is our default container name.

NorseGaud: I don't use that as the name; I think it's more explicit for newbies this way.

casperklein:
What are the reasons for mounting the whole `conf.d` directory? There are currently 29 files and only two of them are modified. That means 27 of them are unchanged and don't need to be mounted from the host.

This could produce trouble in the future: Today you copy theses files (which got sane Debian default values) to the host. In some months there will be Debian 11. These files could change then. So you would have to keep that in mind and update your local files.
I think it's easier, to just copy and mount the two modified files.

Another and maybe the easiest approach could be, to put the content of the two files just in a new one, e.g. `conf.d/99-mail-crypt.conf`

What do you think?

Otherwise, the rest looks good to me.

docs/content/config/advanced/override-defaults/dovecot.md
```diff
@@ -30,15 +30,20 @@ This limit is quickly reached if users connect to the mail server with multiple
...
-For major configuration changes it’s best to override the dovecot configuration files. For each configuration file you want to override, add a list entry under the `volumes` key.
-
-```yaml
-services:
-  mail:
-    volumes:
-      - maildata:/var/mail
-      - ./config/dovecot/10-master.conf:/etc/dovecot/conf.d/10-master.conf
-```
+For major configuration changes it’s best to override the dovecot configuration files. If you do not yet have a full config files populated on the host, it's unfortunately not as easy as mounting an empty host location to `/etc/dovecot/conf.d` on the container as the config files in the container will be deleted with an empty folder. Fortunately, there is a multi-step process necessary to obtain the current configs:
```
wernerfred, line 33: Wouldn't overwriting only the files you need to modify (in your case `20-imap.conf` and `20-lmtp.conf`) the better approach? Then one could use the way which is already described [here in the docs](https://docker-mailserver.github.io/docker-mailserver/edge/config/advanced/override-defaults/dovecot/#override-configuration). The only thing you might need it obtaining the current file which can be done with `docker cp`.

By providing the whole configuration via volume you will not be able to receive updates to this files if something would change in future. Mounting only the files necessary for your change is imho the smarter solution.

What i don't like is that you deleted the current section which describes the way i mentioned. As your way works too it would be okay for me to add this as well but please keep the current approach in the docs. You can put it in another headline showing the way on how to work around the problem of mounting previously empty directory (which is more a general hint than a dovecot related thing). But overwriting or if implemented providing single files to overwrite current config is the preferred way. [List with currently supported files here](https://docker-mailserver.github.io/docker-mailserver/edge/config/advanced/optional-config/).

Your opinion on that? @NorseGaud @casperklein

docs/content/config/advanced/override-defaults/dovecot.md
aendeavor: It's not, since this project and its documentation builds on the premise that the container is called `mailserver`. This is found everywhere in the documentation, and it should be kept this way.

wernerfred (changes requested):

wernerfred:
`docs/content/config/security/mail_crypt.md` looks good. Thanks for this contribution 👍🏻

docs/content/config/security/mail_crypt.md
```diff
@@ -0,0 +1,54 @@
...
+title: 'Security | mail_crypt (email/storage encryption)'
+---
+
+Official Dovecot documentation: https://doc.dovecot.org/configuration_manual/mail_crypt_plugin/
+
+> The Mail crypt plugin is used to secure email messages stored in a Dovecot system. Messages are encrypted before written to storage and decrypted after reading. Both operations are transparent to the user.
+>
+> In case of unauthorized access to the storage backend, the messages will, without access to the decryption keys, be unreadable to the offending party.
+>
+> There can be a single encryption key for the whole system or each user can have a key of their own. The used cryptographical methods are widely used standards and keys are stored in portable formats, when possible.
```
wernerfred, line 11: You could use the `!!! info` admonitions stlye here. [See here for example](https://docker-mailserver.github.io/docker-mailserver/edge/config/setup.sh/)

wernerfred (changes requested):

docs/content/config/advanced/override-defaults/dovecot.md
NorseGaud: Good idea; docker cp totally slipped my mind. I'll switch to individual files and use docker cp to get the files we need out.

docs/content/config/advanced/override-defaults/dovecot.md
NorseGaud: @aendeavor , should I then update the rest of the documentation? https://github.com/docker-mailserver/docker-mailserver/blob/8ce3cca4244ed2adbf936571d62e2ee43ca8d68a/docs/content/config/advanced/override-defaults/dovecot.md#debugging

I'd rather not if it can be avoided

aendeavor: > @aendeavor , should I then update the rest of the documentation?
> I'd rather not if it can be avoided

No, this is something we as maintainers have to fix later:) But you will need to do it for the documentationyou introduced.

wernerfred (approved):
SGTM. The admonitions can be added later as it is only style related.

aendeavor (dismissed):
LGTM

If you want to, you can add the admonitions, but this should get merged either way

casperklein (dismissed):

docs/content/config/security/mail_crypt.md
NorseGaud: Perfect <3

NorseGaud:
Thanks guys, I believe this is final :)

aendeavor (approved):

aendeavor (approved):

casperklein (approved):

aendeavor:
@wernerfred feel free to merge when you approve this (again) :D

wernerfred (approved):

wernerfred:
Ah damn, i overlooked that we forgot to include the new file. I fixed that with #1916 
Looking forward to the preview feature @polarathene 👨🏻‍💻

polarathene:
> Looking forward to the preview feature @polarathene 👨🏻‍💻

Sorry about the delay! Been a busy week. I'll get that feature sorted ASAP :sweat_smile:

Issue #1915, opened 2021-04-17
Traefik v2 + letsencrypt wildcard-certificate, cant load certificate


DomenikIrrgang:
# Bug Report

## Context

I am using traefik v2 as a proxy for all my applications running on my server and it handles automatic renewal of my certificate. The idea here was to mount that certificate into the docker-mailserver container and us it here too. I followed all the steps in the documentation, except for the fact that the two of them are not in the same network (traefik is not supposed to route to the mailserver anyways). The certificate is a wildcard certificate. 

### Actual Behavior

The mailserver fails to load the certificates with the following warning:
```
warning: error loading private keys and certificates from: /etc/letsencrypt/live/syncup.de/key.pem /etc/letsencrypt/live/syncup.de/fullchain.pem: disabling TLS support
```

### Expected Behavior

TSL should work without any errors.

## Your Environment

available RAM: 8GB
Docker version 19.03.6-rc1, build 77739a9875
docker-compose version 1.28.6, build 5db8d86f

### docker-compose.yml

```
version: '3.8'

services:
  mailserver:
    image: docker.io/mailserver/docker-mailserver:latest
    hostname: ${HOSTNAME}
    domainname: ${DOMAINNAME}
    container_name: ${CONTAINER_NAME}
    env_file: mailserver.env
    ports:
      - "25:25"
      - "143:143"
      - "465:465"
      - "587:587"
      - "993:993"
    volumes:
      - maildata:/var/mail
      - mailstate:/var/mail-state
      - maillogs:/var/log/mail
      - /etc/localtime:/etc/localtime:ro
      - ./config/:/tmp/docker-mailserver/${SELINUX_LABEL}
      - /home/traefik/acme.json:/etc/letsencrypt/acme.json:ro
    environment:
      SSL_TYPE: letsencrypt
      SSL_DOMAIN: "*.syncup.de"
    restart: always
    stop_grace_period: 1m
    cap_add: [ "NET_ADMIN", "SYS_PTRACE" ]

volumes:
  maildata:
  mailstate:
  maillogs:
```

### Environment Variables

```
OVERRIDE_HOSTNAME=syncup.de
DMS_DEBUG=0
SUPERVISOR_LOGLEVEL=debug
ONE_DIR=0
SPOOF_PROTECTION=1
ENABLE_SRS=0
ENABLE_CLAMAV=0
ENABLE_AMAVIS=1
ENABLE_FAIL2BAN=0
[...]

PR #1918, opened 2021-04-18
Adjusted documentation for service name and Traefik certificate issuance

aendeavor:
# Description

- changed container name from `mail` to `mailserver` to streamline docs
- some braces around variables in Bash scripts were added
- changed Traefik's certificate section
  - removed docs for Traefik v1 - we want everyone to use Traefik v2 from `9.2.0`
  - **the wildcard certificate issue** in #1915 was addressed
  - proper indentation was added

I have drafted a prerelease for `v9.2.0`. Have a look :)

### TODO

1. Make sure the adjustments to `SSL_DOMAIN` are correct (by solving #1915)

## Type of change

## Checklist:

### DIFF
--- a/docs/content/config/security/ssl.md
+++ b/docs/content/config/security/ssl.md
@@ -363,72 +363,63 @@ no peer certificate available
 No client certificate CA names sent
 ```
 
-## Traefik
+## Traefik v2
 
-[Traefik](https://github.com/containous/traefik) is an open-source Edge Router which handles ACME protocol using [lego](https://github.com/go-acme/lego).
+[Traefik][traefik::github] is an open-source application proxy using the [ACME protocol][ietf::rfc::acme]. [Traefik][traefik::github] can request certificates for domains and subdomains, and it will take care of renewals, challenge negotiations, etc. We strongly recommend to use [Traefik][traefik::github]'s major version 2.
 
-Traefik can request certificates for domains through the ACME protocol (see [Traefik's documentation about its ACME negotiation & storage mechanism](https://docs.traefik.io/https/acme/)). Traefik's router will take care of renewals, challenge negotiations, etc.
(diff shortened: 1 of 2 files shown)


casperklein:
> I have drafted a prerelease for v9.2.0. Have a look :)

fyi: I've edited two fail2ban related things.

This release contains "breaking changes". Even if I can't imagine, what could break, the possibility is given. Per definition, breaking changes should only be done in major updates.

@docker-mailserver/maintainers Should we aim 9.2 or 10.0 in that case?

aendeavor:
> > I have drafted a prerelease for v9.2.0. Have a look :)
> 
> 
> 
> fyi: I've edited two fail2ban related things.
> 
> 
> 
> This release contains "breaking changes". Even if I can't imagine, what could break, the possibility is given. Per definition, breaking changes should only be done in major updates.
> 
> 
> 
> @docker-mailserver/maintainers Should we aim 9.2 or 10.0 in that case?

You may be right, it was written out of caution. I'd pledge against `10.0.0`, since these are many, but rather small changes. But I'd leave this up to a vote. Again, this is only a pre-release so we can have a look and decide.

wernerfred (approved):
LGTM 👍🏻

wernerfred:
> You may be right, it was written out of caution. I'd pledge against `10.0.0`, since these are many, but rather small changes. But I'd leave this up to a vote. Again, this is only a pre-release so we can have a look and decide.

Basically this have to be `v10.0.0` but this (rather fast) rising numbers are the cause many other software projects are shifting the semver and use: `core.major.minor`. As we commited to use Semver we should stick with it and use `10.0.0`. If we have trouble in the future with this numbers we could think of resetting the counter but I would keep it transparent to users that (even if the possibility is low) the next update could break their setup.

casperklein (approved):

aendeavor:
Alright. When #1901 is merged, we will release `v10.0.0` :)

PR #1947, opened 2021-05-06
introduce variable to control Amavis' loglevel

aendeavor:
# Description

Introduced `AMAVIS_LOGLEVEL` variable to control the log level. This way, one can control the verbositoy of Amavis. This can be of special importance when users post logs (for the first time) as they always post the (huge) Amavis log too. See #1944.

This is a "breaking" change, as the default of `0` (= warning) is new. We can merge this before releasing `10.0.0`. I can test this feature in my setup for a week if necessary,

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/startup/setup-stack.sh
+++ b/target/scripts/startup/setup-stack.sh
@@ -1501,6 +1501,10 @@ function _setup_security_stack
     then
       cp /tmp/docker-mailserver/amavis.cf /etc/amavis/conf.d/50-user
     fi
+
+    sed -i -E \
+      "s|(log_level).*|\1 = ${AMAVIS_LOGLEVEL};|g" \
+      /etc/amavis/conf.d/49-docker-mailserver
   fi
 }
 
--- a/ENVIRONMENT.md
+++ b/ENVIRONMENT.md
@@ -34,6 +34,15 @@ Amavis content filter (used for ClamAV & SpamAssassin)
 - 0 => Amavis is disabled
 - **1** => Amavis is enabled
 
+##### AMAVIS_LOGLEVEL
+
+[This page](https://lists.amavis.org/pipermail/amavis-users/2011-March/000158.html) provides information on Amavis' logging statistics.
+
+- -1 => log errors
+- **0** => log warnings
+- 1 => log info
+- 3 => log debug
+
 ##### ENABLE_CLAMAV
 
 - **0** => Clamav is disabled
--- a/ENVIRONMENT.md
+++ b/ENVIRONMENT.md
@@ -2,9 +2,8 @@
 
 ## Variables
 
-1. If an option doesn't work as documented here, check if you are running the latest image!
-2. Values in **bold** are the default values.
-3. Since `docker-mailserver v7.1.0`, comparisons for environment variables are executed differently. If you previously used `VARIABLE=''` as the `empty` value, **update** to now use `VARIABLE=`.
+1. Values in **bold** are the default values.
+2. If an option doesn't work as documented here, check if you are running the latest image. The current `master` branch corresponds to the image `mailserver/docker-mailserver:edge`.
 
 ### Assignments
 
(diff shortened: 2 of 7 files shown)


casperklein:
Please add the information provided in `ENVIRONMENT.md` also to `mailserver.env`.

Besides that, the rest looks good to me.

> We can merge this before releasing 10.0.0. 

Definitely. The more breaking changes, the better 😉 So we don't have to release that often major releases.

aendeavor:
Good point. I've added the variable to `mailserver.env`. This is then ready to be merged :)

casperklein (approved):

target/scripts/startup/setup-stack.sh
```diff
@@ -1501,6 +1501,10 @@ function _setup_security_stack
     then
       cp /tmp/docker-mailserver/amavis.cf /etc/amavis/conf.d/50-user
     fi
+
+    sed -i -E \
+      "s|(log_level).*|\1 = ${AMAVIS_LOGLEVEL};|g" \
+      /etc/amavis/conf.d/49-docker-mailserver
```
casperklein, line 1507: More precise:
`sed -i -E '/^\$log_level/'" s|(log_level).*|\1 = ${AMAVIS_LOGLEVEL};|g" /etc/amavis/conf.d/49-docker-mailserver`

Feel free to change it or leave it as it is, if you think that over complicate it 👍

aendeavor: I don't know this syntax. Can you explain what this does and why one would want to use this over the current command? :D

casperklein: It ensures, that only lines starting with "$log_level" are subject for the substitution.

`sed '/PATTERN/ OPERATION' FILE`
`sed '/match this lines only/ s|foo|bar|g' FILE`

You can try both methods with a test file like this to see the difference:

```
# some comment about $log_level functionality

$log_level = ;
```

Without PATTERN matching, the comment line is also substituted which isn't desired.

> apostrophies seem off to me

This didn't work, that's why I chose single quotes around the PATTERN:

```
sed -i -E \
  "/^\$log_level/ s|(log_level).*|\1 = ${AMAVIS_LOGLEVEL};|g" \
  /etc/amavis/conf.d/49-docker-mailserver
```

I suspect some sed escape black magic..

casperklein: Final attempt 🤓:

-> Match only lines starting with `$log_level`
-> Replace `log_level.*` with `log_level = ${AMAVIS_LOGLEVEL};`
-> We only expect/interested in the first match (`.*` will match the rest of the line), so no **g**lobal substitution is needed.

`sed -i -E "/^\\\$log_level/ s|log_level.*|log_level = ${AMAVIS_LOGLEVEL};|" /etc/amavis/conf.d/49-docker-mailserver`
or
`sed -i -E '/^\$log_level/'" s|log_level.*|log_level = ${AMAVIS_LOGLEVEL};|" /etc/amavis/conf.d/49-docker-mailserver`

aendeavor: I always knew `sed` could escalate quickly 😄 I agree with the statement that the current method is fine and easy to read (not that the latter is too important as we have some advanced Bash code anyways).

But thanks to you I learned something new - very much appreciated! Maybe I'll need this in the future:)

casperklein: What I don't like about `sed -i` in general is, that there is no easy way to detect, if e.g. a replacement took place. EXIT CODE is always 0, regardless if a substition took place or not. If a config file changes after an update or so, `sed` can silently fail to substitute something.

To avoid that, the sed substitute statement must be as tight as possible or a proper test must be written. Lot's of work if I look at the current script files 😄

aendeavor: That is correct, I agree.

PR #1948, opened 2021-05-07
Outsourcing Environment Variables to the Documentation

aendeavor:
# Description

Environment variables are now found in the documentation and `README` links to this documentation. As a consequence, `ENVIRONMENT.md` has been deleted.

Furthermore, I reworked the whole system for variables. In the past, users we're confused by two different `.env` files which serve two different purposes. Therefore, the `README` was adjusted and `compose.env` becomes history. This should make it more straight-forward and the setup easier.

## Type of change

## Checklist:


aendeavor:
@docker-mailserver/maintainers Contributions and discussions are welcome :D

casperklein:
Now that `compose.env --> .env` is going to be removed, I noticed, that at some point `env_file: .env` was removed from `docker-compose.yml`.

That leads to, that `$DOMAINNAME` is now no longer available in the container.

I did a quick grep. Some scripts do something like `DOMAINNAME=$(hostname -d)`, but others seem to depend on `$DOMAINNAME` already set?

IMO this can never be _true_:
https://github.com/docker-mailserver/docker-mailserver/blob/74068fef6cf129092014a7562f40eb5f36d8a31b/target/scripts/postsrsd-wrapper.sh#L11

Or do I overlook something?

README.md
```diff
@@ -144,12 +141,26 @@ When keys are generated, you can configure your DNS server by just pasting the c
...
 #### Custom user changes & patches
 
-If you'd like to change, patch or alter files or behavior of `docker-mailserver`, you can use a script. See the [documentation](https://docker-mailserver.github.io/docker-mailserver/edge/config/advanced/user-patches/) for a detailed explanation.
+If you'd like to change, patch or alter files or behavior of `docker-mailserver`, you can use a script. See the [documentation](https://docker-mailserver.github.io/docker-mailserver/edge/config/advanced/override-defaults/user-patches/) for a detailed explanation.
+
+#### Updating `docker-mailserver`
+
+``` BASH
+docker-compose down
+docker pull docker.io/mailserver/docker-mailserver:<VERSION TAG>
```
casperklein, line 150: It doesn't matter which version you pull. docker-compose will only use the configured version set in the yml configuration when starting a container.
Therefore I would change this to a simple: `docker-compose pull.`

To reduce the downtime, I would also change the order to:

```
docker-compose pull
docker-compose down
docker-compose up -d mailserver
```

aendeavor:
> Now that `compose.env --> .env` is going to be removed, I noticed, that at some point `env_file: .env` was removed from `docker-compose.yml`.
> 
> That leads to, that `$DOMAINNAME` is now no longer available in the container.
> 
> I did a quick grep. Some scripts do something like `DOMAINNAME=$(hostname -d)`, but others seem to depend on `$DOMAINNAME` already set?
> 
> IMO this can never be _true_:
> 
> https://github.com/docker-mailserver/docker-mailserver/blob/74068fef6cf129092014a7562f40eb5f36d8a31b/target/scripts/postsrsd-wrapper.sh#L11
> 
> Or do I overlook something?

No, you're quite correct. But according to blame, I think I removed this a rather long while back, but no complaints were raised. But we should definitely do something about this behavior.

I propose to get rid of this variable and just parse it from the `hostname` command of from the `OVERRIDE_HOSTNAME` variable or something like this.

README.md
aendeavor: Good point. Feel free to change this - I'm all for it.

README.md
```diff
@@ -190,9 +191,9 @@ version: '3.8'
 services:
   mailserver:
     image: docker.io/mailserver/docker-mailserver:latest
-    hostname: mail          # ${HOSTNAME}
-    domainname: domain.com  # ${DOMAINNAME}
-    container_name: mail    # ${CONTAINER_NAME}
+    hostname: mail
+    domainname: domain.com
```
casperklein, line 195: `domain.com` is a valid registered domain. We shouldn't use that as an example.

Instead, `example.com` should be used, which is defined exactly for that [use case](https://www.iana.org/domains/reserved)

README.md
```diff
@@ -231,9 +232,9 @@ version: '3.8'
...
```
casperklein, line 236: Same as above. Should be `example.com`

README.md
aendeavor: Agree.

polarathene (changes requested):
Thanks for kickstarting this! :smiley:  

Here's my feedback so far. Some of it's outdated now, was half way through it last night but got interrupted :grimacing:

README.md
```diff
@@ -77,10 +77,9 @@ All workflows are using the **tagging convention** listed below. It is subsequen
 
 ### Get the tools
 
-Download `docker-compose.yml`, `compose.env`, `mailserver.env`
+Download `docker-compose.yml` and `mailserver.env`
```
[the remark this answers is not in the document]
aendeavor: I prefer the current version without the `:`

aendeavor:
**OPEN TASKS**

- [ ] Check that all links are correct
- [ ]

aendeavor:
@casperklein I looked through it, and the `DOMAINNAME` variable is handled and set correctly. `start-mailserver.sh` sets this variable first here:

https://github.com/docker-mailserver/docker-mailserver/blob/master/target/scripts/start-mailserver.sh#L72

and if `OVERRIDE_HOSTNAME` is set, it will be overridden in the `setup-stack.sh`:

https://github.com/docker-mailserver/docker-mailserver/blob/master/target/scripts/startup/check-stack.sh#L22-L23

which is executed first.

aendeavor:
I cleared the last tasks and this PR is now ready for proper review and merging.

casperklein:
That's true. But scripts not called from from `start-mailserver.sh` or a Bash login shell cannot use that variable.

That's probably why some scripts did implement their own solution to get the DOMAINNAME:

https://github.com/docker-mailserver/docker-mailserver/blob/29d630f6954c96ed972e9bd1e7596a5b276d6704/target/scripts/check-for-changes.sh#L31-L36

My comment above about `postsrsd-wrapper.sh` remains true.

That said, it's IMO not related to this PR and can be fixed (or not) in a separate issue.

aendeavor:
Ah, I see. Yes, this should somehow be fixed. I'd propose another PR for this too, We could maybe put this into `/etc/environment` - would this help? Just to give an idea here that can be implemented in another PR.

casperklein:
> We could maybe put this into /etc/environment

Yes that would be one way. Or we source `/root/.bashrc` or we use a complete different approach: https://github.com/docker-mailserver/docker-mailserver/blob/c4cf3151930b4eeba19d30b09a745e19d08feb17/target/supervisor/conf.d/supervisor-app.conf#L142

That way, all env variables are usable within the script.

aendeavor:
> > We could maybe put this into /etc/environment
> 
> 
> 
> Yes that would be one way. Or we source `/root/.bashrc` or we use a complete different approach: https://github.com/docker-mailserver/docker-mailserver/blob/c4cf3151930b4eeba19d30b09a745e19d08feb17/target/supervisor/conf.d/supervisor-app.conf#L142
> 
> 
> 
> That way, all env variables are usable within the script.

How do `-i` and `-l` differ? The one seems to be interactive, the other one emulates it? How does that affect environment variables?

casperklein:
`-i` makes it _interactive_ and bind it to a terminal. It makes variables like $PS1 available and sources `.bashrc` etc.
`-l` make bash behave like a login shell (sourcing `.bashrc` etc.), but not providing some terminal features like $PS1.

In a nutshell, `-l` is a subset of `-i`.

docs/content/contributing/issues-and-pull-requests.md
```diff
@@ -37,14 +37,14 @@ The development workflow is the following:
 3. Add integration tests if necessary
 4. Get the linters with `make install_linters` and install `jq` with the package manager of your OS
 5. Use `make clean all` to build image locally and run tests (note that tests work on Linux **only**)
-6. Document your improvements if necessary (e.g. if you introduced new environment variables, write the description in [`ENVIRONMENT.md`][github-file-env])
+6. Document your improvements if necessary (e.g. if you introduced new environment variables, write the description on [`this page`][docs-environment])
```
polarathene, line 40:
```suggestion
6. Document your improvements if necessary (e.g. if you introduced new environment variables, describe those in the [ENV documentation][docs-environment])
```

polarathene (changes requested):
Almost there! :smiley:

README.md
```diff
@@ -144,12 +139,26 @@ When keys are generated, you can configure your DNS server by just pasting the c
...
+
+You're done! And don't forget to have a look at the remaining functions of the `setup.sh` script with `./setup.sh help`.
 
 #### Supported Operating Systems
 
 We are currently providing support for Linux. Windows is _not_ supported and is known to cause problems. Similarly, macOS is _not officially_ supported - but you may get it to work there. In the end, Linux should be your preferred operating system for this image, especially when using this mailserver in production.
 
+#### Bare Domains
+
+If you want to use a bare domain (host name = domain name), see [FAQ](https://docker-mailserver.github.io/docker-mailserver/edge/faq#can-i-use-nakedbare-domains-no-host-name).
```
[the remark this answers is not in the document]
aendeavor: That could cause confusion. Someone may be tempted to fill both fields with the same string now.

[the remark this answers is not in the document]
polarathene: I don't see how that differs from `host name = domain name` tbh. You're still implying one to equate to the other. The suggestion I've provided references a "bare domain" as `hostname` being equivalent to `domainname`, it still requires you to read the FAQ if you want that to happen which is what this line is all about directing the user to.

Making it clear that there is a relation to `hostname` and `domainname` fields as they are referenced elsewhere in the document and the `docker-compose.yml` configs only improves the discovery. If you feel there is risk for confusion, add some additional context to clarify:

```
bare domain (where `hostname` would be equivalent to `domainname`)
bare domain (eg `hostname` matching `domainname`)
bare domain (eg `example.com` instead of `mail.example.com`)
bare domain (without a subdomain aka `hostname`)
```

Or something along those lines? I might not be understanding the description well enough myself if I've described that wrong above.

README.md
```diff
@@ -103,20 +102,16 @@ chmod a+x ./setup.sh
...
-- Variables in `.env` are expanded in the `docker-compose.yml` file **only** and **not** in the container. The file `mailserver.env` serves this case where environment variables are used in the container.
-- If you want to use a bare domain (host name = domain name), see [FAQ](https://docker-mailserver.github.io/docker-mailserver/edge/faq#can-i-use-nakedbare-domains-no-host-name)
+1. [Install the latest docker-compose](https://docs.docker.com/compose/install/)
+2. Edit `docker-compose.yml` to your liking
+   - substitute `<HOSTNAME>` and `<DOMAINNAME>` according to your domain
+   - if you want to use SELinux for the `./config/:/tmp/docker-mailserver/` mount, append `-z` or `-Z`
+3. Configure the mailserver container to your liking by editing `mailserver.env` ([**Documentation**](https://docker-mailserver.github.io/docker-mailserver/edge/config/environment/))
+   - this files supports [_only_ simple `VAR=VAL`](https://docs.docker.com/compose/env-file/) (**don't** quote your values)
+   - variable substitution is **not** supported (e.g. :no_entry_sign:`OVERRIDE_HOSTNAME=$HOSTNAME.$DOMAINNAME` :no_entry_sign:)
```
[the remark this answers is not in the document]
aendeavor: I think this is fine.

wernerfred (dismissed):
LGTM 👍🏻 
Thanks

docs/content/config/environment.md
```diff
@@ -26,6 +30,23 @@ Here you can adjust the [log-level for Supervisor](http://supervisord.org/loggin
...
+- 1 => consolidate all states into a single directory (`/var/mail-state`) to allow persistence using docker volumes
+
+##### PERMIT_DOCKER
+
+Set different options for mynetworks option (can be overwrite in postfix-main.cf) **WARNING**: Adding the docker network's gateway to the list of trusted hosts, e.g. using the `network` or `connected-networks` option, can create an [**open relay**](https://en.wikipedia.org/wiki/Open_mail_relay), for instance if IPv6 is enabled on the host machine but not in Docker.
+
+- **empty** => localhost only
+- host => Add docker host (ipv4 only)
+- network => Add the docker default bridge network (172.16.0.0/12); **WARNING**: `docker-compose` might use others (e.g. 192.168.0.0/16) use `PERMIT_DOCKER=connected-networks` in this case
+- connected-networks => Add all connected docker networks (ipv4 only)
```
polarathene, line 46:
```suggestion
- **empty** => localhost only.
- host => Add docker host (ipv4 only).
- network => Add the docker default bridge network (172.16.0.0/12); **WARNING**: `docker-compose` might use others (e.g. 192.168.0.0/16) use `PERMIT_DOCKER=connected-networks` in this case.
- connected-networks => Add all connected docker networks (ipv4 only).
```

polarathene (changes requested):
- Absolute URLs  should not be used to reference internal docs, those should be relative file paths for error detection.
- 1 typo to correct.
- **Optional:** Punctuation consistency - `.` at end of lines for `environment.md` (I didn't review the whole document, just noticed inconsistent usage through highlighted diff).

polarathene (approved):

PR #1951, opened 2021-05-07
Update check

### LINKED ISSUE #1946: [FR] Include env variable or file containing the image version
# Feature Request

## Context

We should introduce an env variable or file that includes the image version. I see two main benefits from that:

- We could include the version information in the startup script, which makes it easier to identify the used version (e.g. when helping with issues).
- It should be easy to implement a cron based update check to see, if there is a new release available.

"Disadvantage":

- Each time we release a new version, this env variable/file must be updated.

Env variable or file containing the image version. I prefer the file solution. **What are your thoughts?**

The necessary adjustments.

casperklein:
# Description

This adds an update check, that will (if enabled) send a notification mail to `$POSTMASTER_ADDRESS` once an update is available.

The check runs on container start and then in a 24 hour interval.

The latest version information is fetched from https://raw.githubusercontent.com/docker-mailserver/docker-mailserver/master/VERSION

To test this feature, you can trigger a update notification with:

```bash
# "downgrade" version and restart update-check
docker exec mailserver bash -c "echo 9.0.0 >/VERSION && supervisorctl restart update-check"

# verify log
docker exec mailserver cat /var/log/supervisor/update-check.log
```

## Type of change

## Checklist:


target/scripts/update-check.sh
```diff
@@ -0,0 +1,46 @@
...
+
+VERSION=$(</VERSION)
+
+while true
+do
+  DATE=$(date '+%F %T')
+
+  # get remote version information
+  #LATEST=$(curl -sf https://raw.githubusercontent.com/docker-mailserver/docker-mailserver/master/VERSION) # doesn't exist yet
+  LATEST=$(curl -sf https://raw.githubusercontent.com/casperklein/archive_docker-mailserver/update-check/VERSION)
```
casperklein, line 11: This is for testing only. I'll correct this, right before the merge.

wernerfred (commented):
LGTM on a first sight 👍🏻 
What I don't like is that the approach with the `VERSION` file cannot really be automated in a CI/CD process. The event that would kick of CI/CD is either a new Release or the push of a tag. This could of course trigger a pipeline to update the `VERSION` to the one matching the tag for example but the docker build then would checkout the specific tag of the repo which would not contain the correct `VERISON` as this commit happened after the tag push.

One workaround would be to add a step that changes the `VERSION` after cloning but right before building the image but then it would never be changed in the repository.

No show-stopper but still something to take into consideration as forgetting to update the file will lead to floating update notifications.

casperklein:
Other idea: Why do we want to automate the update of this file?

Before tagging a new release, bumping the VERSION should be the last PR made. We can use that PR, so that @docker-mailserver/maintainers can confirm, that the to be released version is tested and works as desired.

polarathene:
You could tag releases via manual Github Action trigger that updates this VERSION file among other things if that's important.

You can also ignore the file approach and query the latest tag via Github API (requires `jq` installed):

```sh
curl -sL https://api.github.com/repos/docker-mailserver/docker-mailserver/releases/latest | jq -r ".tag_name"
```

casperklein:
> You can also ignore the file approach and query the latest tag via Github API (requires `jq` installed):

How does the script get then the current running version?

polarathene:
> How does the script get the current running version?

Oh I thought that was already available elsewhere like adding into an ENV in the Dockerfile. It could be added via arg during build on tagged release?

Then just query the ENV to compare against?

aendeavor:
I too would favor a solution without a file. I have no good solution myself, but the comments I read suggest that we could achieve this without a file.

casperklein:
Let's move the VERSION file discussion to the related issue #1946

fbartels:
I think the VERSION file could still be created during docker build as a last run step for example. But the comparison should instead be done either against the tags available on Docker Hub, or a git tag.

casperklein:
I've updated the description of the initial post.

@docker-mailserver/maintainers Please review and if possible also test this feature (see first post for an easy method.).

target/scripts/update-check.sh
```diff
@@ -0,0 +1,55 @@
...
+Hello ${POSTMASTER_ADDRESS}!
+
+There is a docker-mailserver update available on your host: $(hostname -f)
+
+Current version: ${DMS_VERSION}
+Latest  version: ${LATEST}
+
+Changelog: ${CHANGELOG}
+EOM
+      echo "${MAIL}" | mail -s "Update available! [ ${DMS_VERSION} --> ${LATEST} ]" "${POSTMASTER_ADDRESS}" && \
```
wernerfred, line 41: Maybe put "DMS" oder "Mailserver" into the subject: 

```suggestion
      echo "${MAIL}" | mail -s "Mailserver update available! [ ${DMS_VERSION} --> ${LATEST} ]" "${POSTMASTER_ADDRESS}" && \
```

target/scripts/update-check.sh
```diff
@@ -0,0 +1,55 @@
...
+      # only notify once
+      exit 0
+    else
+      echo "${DATE} Info: No update available"
+    fi
+  else
+    echo "${DATE} Error: Update check failed."
+  fi
+  # check again in one day
+  sleep 1d
```
wernerfred, line 54: Should we make the sleep interval configurable?

target/scripts/update-check.sh
```diff
@@ -0,0 +1,55 @@
...
+
+# DMS_VERSION == Commit hash?
+if [[ ${DMS_VERSION} =~ ^[a-f0-9]{40}$ ]]
+then
+  DATE=$(date '+%F %T')
+  echo "${DATE} Info: You are using an edge build. Update checks are not supported."
+  exit 0
+fi
+
+API="https://api.github.com/repos/docker-mailserver/docker-mailserver/releases/latest"
```
wernerfred, line 11: Does latest release always show the "most recent" release or the "greatest" one? So we had a moment where we had to rerun a job on a tag as 9.2 got published by accident and 9.1 was the lastest/most recent and 9.2 was the latest/greatest in this moment

casperklein: Good question, idk.

casperklein: I've tested it..

> Does latest release always show the "most recent" 

yes

I also encounter another problem using the API: There are rate limits in place. In case you already use the API for other things from the same host/IP as DMS runs on, you can hit this limit "easily" (at least in my case). This cost me some time yesterday figuring out, why the update check suddenly stopped working on my machine 😞 There are however no limits for requesting "raw" files from a repo.

I don't think I have mentioned it before^^ --> But I prefer a FILE based solution. The only valid concern IMO was, that we have to remember, to bump that FILE on a new release. I don't think that's a big deal, as we have the same situation with the CHANGELOG file. We also have to remember to adjust it in the last commit before release. One more file wouldn't hurt in my opinion. It's simple, it's flat and it just works. No rate limits and a lot of [big projects](https://github.com/home-assistant/version/blob/master/stable.json) do it successfully in a similar way.

Both mentioned problems would not relate to a file based solution 😉

PS: That was my last try to convince you all 😄 If you are unsure, take a look at my summary post [here](https://github.com/docker-mailserver/docker-mailserver/issues/1946#issuecomment-836905633), which covers all mentioned concerns from you.

target/scripts/update-check.sh
casperklein: I thought about the [same](https://github.com/docker-mailserver/docker-mailserver/issues/1946#issuecomment-833884441). However, for time reasons I skipped it in the first place.

target/scripts/update-check.sh
aendeavor: I think that since you're the creator of the PR and since you have taken your time to implement it, I will not reject a file-based solution if the benefits outweigh the disadvantages - which they seem to do.

Feel free to "revert" to the file-based approach, I consent :)

wernerfred: > The only valid concern IMO was, that we have to remember, to bump that FILE on a new release.

If we go the way I mentioned yesterday with the feature-freeze PR in front of every release we can cover the Update in that PR and multiple people can review if all final version changes (CHANGELOG, VERSION,..) are made. 👍🏻

wernerfred: Let's call this PR the last one before feature-freeze and we will directly see how it works out :)

Dockerfile
```diff
@@ -162,13 +165,7 @@ RUN \
   touch /var/log/auth.log && \
   update-locale && \
   rm /etc/postsrsd.secret && \
-  rm /etc/cron.daily/00logwatch && \
-  if [[ ${VCS_VER} =~ ^refs/tags/v.+ ]]; then \
-    echo "export DMS_VERSION='${VCS_VER:11}'" >>/root/.bashrc; \
-  else \
-    echo "export DMS_VERSION='${VCS_REF}'" >>/root/.bashrc; \
-  fi
```
polarathene, line 170: Is this not useful to have available? I thought the main benefit of the `VERSION` file was for getting a version to compare against?

Being able to query the `DMS_VERSION` env var may still be useful vs reading a generic named `VERSION` file? I'm thinking in regards to other scenarios unrelated to this update check feature, or is that likely to be such a niche case that they should just `cat /VERSION`?

casperklein:
At the moment $VERSION_URL points to my repo, so the version check can be tested. If this PR is ready for merge, I'll provide a final commit correcting that variable.

Dockerfile
casperklein: > or is that likely to be such a niche case that they should just cat /VERSION?

IMO yes.

Now that we have the VERSION file, I thought this is not needed any more. If some really needs it in a var --> `foo=$(<VERSION)`

target/scripts/update-check.sh
casperklein: https://github.com/docker-mailserver/docker-mailserver/pull/1951/commits/87dbb5650fbe1bb3d94e33e9612cd3c576351fa7

polarathene (approved):
LGTM.

May want to add a brief mention of the ENV vars and default behaviour in the [docs page regarding update/maintenance for the image](https://docker-mailserver.github.io/docker-mailserver/edge/config/advanced/maintenance/update-and-cleanup/)?

target/scripts/update-check.sh
```diff
@@ -0,0 +1,60 @@
...
+
+VERSION=$(</VERSION)
+#VERSION_URL="https://raw.githubusercontent.com/docker-mailserver/docker-mailserver/master/VERSION" # doesn't exist yet
+VERSION_URL="https://raw.githubusercontent.com/casperklein/archive_docker-mailserver/update-check/VERSION"
+CHANGELOG="https://github.com/docker-mailserver/docker-mailserver/blob/master/CHANGELOG.md"
+
+_log() {
+  DATE=$(date '+%F %T')
+  echo "${DATE} ${1}"
+}
```
aendeavor, line 11: We could in-line this function to be

```
_log { DATE=$(date '+%F %T'); echo "${DATE} ${1}"; }
```

which adheres to the styling guide in CONTRIBUTING.md ;)

aendeavor (dismissed):

target/scripts/start-mailserver.sh
```diff
@@ -64,6 +65,7 @@ VARS[SRS_SENDER_CLASSES]="${SRS_SENDER_CLASSES:=envelope_sender}"
 VARS[SSL_TYPE]="${SSL_TYPE:=}"
 VARS[SUPERVISOR_LOGLEVEL]="${SUPERVISOR_LOGLEVEL:=warn}"
 VARS[TLS_LEVEL]="${TLS_LEVEL:=modern}"
+VARS[UPDATE_CHECK_INTERVAL]="${UPDATE_CHECK_INTERVAL:=1d}"
```
wernerfred, line 68: Maybe we should use `7d` as the default value as the most problems that occurred after releases were fixed within a week?

docs/content/config/environment.md
```diff
@@ -193,6 +193,19 @@ Set the message size limit for all users. If set to zero, the size will be unlim
...
+Check for updates on container start and then once a day. If an update is available, a mail is send to POSTMASTER_ADDRESS.
+
+- 0 => Update check disabled
+- **1** => Update check enabled
+
+##### UPDATE_CHECK_INTERVAL
+
+Customize the update check interval. Number + Suffix. Suffix must be 's' for seconds, 'm' for minutes, 'h' for hours or 'd' for days.
+
+- **1d** => Check for updates once a day
```
wernerfred, line 207: Maybe we should use 7d as the default value as the most problems that occurred after releases were fixed within a week?

wernerfred (dismissed):
LGTM 👍🏻 
Thanks for reworking!

Only one small request from my side: Maybe it makes more sense to higher the default check interval? Now that I am writing this 7d seems a bit high as it would only check 4 times in a month but 1d is to short imho as new releases might need a hotfix? Anyways feel free to merge even with 1d, just a thought of mine and if i don't like it personally i can set it to a higher value by myself 😄

target/scripts/update-check.sh
casperklein: > which adheres to the styling guide in CONTRIBUTING.md

That functions have to be in one line is new for me? I did it for readability.

> maybe we want to add this to helper-functions.sh

It's a from case to case decision. In that case I thought, sourcing the complete helper file just for a little echo function would be overkill.
But I am open for discussion.

aendeavor: > That functions have to be in one line is new for me? I did it for readability.

To be precise, you're right. One-line functions do not appear in CONTRIBUTING.md. But the way the function looks ATM, it does not adhere to the style guide as well :)

casperklein:
I thought about the same, what would be a sane default check interval. In the end, I went with 1d.

1. Compared to 1h, someone does not get instant notified. That is good, if there are any (minor) hotfixes.
2. Compared to 1w, someone gets security updates pretty fast.

> the most problems that occurred after releases were fixed within a week?

If no one would update, because of waiting for a possible hotfix, we wouldn't get aware of any bugs. A vicious circle 😉

target/scripts/update-check.sh
casperklein: Ack. I'll fix the style with my last commit.

wernerfred (approved):
LGTM, thanks

NorseGaud:
@casperklein , was `apt-get -qq -y purge gpg gpg-agent &>/dev/null` in the Dockerfile intended to be removed? I've got It in a few of my branches and I wanted to confirm it was supposed to be removed (it seems useful).

polarathene:
> was `apt-get -qq -y purge gpg gpg-agent &>/dev/null` in the Dockerfile intended to be removed?

Yes.

That line was covered in [this commit](https://github.com/docker-mailserver/docker-mailserver/pull/1951/commits/9b369c5ccefb742632da55314996684bbdfc26a5) for the PR, the commit message references [this Fail2Ban PR](https://github.com/docker-mailserver/docker-mailserver/pull/1965#issuecomment-842475100), the linked comment and earlier discussion on that PR (which introduced it) should clarify why the line was removed.

The `&>dev/null` affected debugging and wasn't required, the `gpg` and `gpg-agent` are provided already via the earlier [`gnupg`](https://packages.debian.org/buster/gnupg) package.

georglauterbach:
@all-contributors please add @casperklein for maintenance

PR #1965, opened 2021-05-12
introduce F2B v0.11

### LINKED ISSUE #1810: Can't block Spambots
# Subject

I'd like some feedback regarding the Postscreen/Fail2ban configuration.

## Description
Hi guys please help me understand what is wrongly configured here with the Postscreen and Fail2ban modules. I've attached a log sample. Fail2ban is activated, it just doesn't kick in to ban the IP's retrying to connect. I can manually ban these IPs with fail2ban-client so the jails are working, however the ban doesn't occur automatically after the "maxretry" is reached. So what is going on here?

``` BASH
# CODE GOES HERE
mailserver       | Feb 15 19:31:21 mail postfix/postscreen[10972]: CONNECT from [87.246.7.226]:40982 to [192.168.176.2]:25
mailserver       | Feb 15 19:31:21 mail postfix/postscreen[10972]: cache btree:/var/lib/postfix/postscreen_cache full cleanup: retained=26 dropped=3 entries
mailserver       | Feb 15 19:31:22 mail postfix/dnsblog[10976]: addr 87.246.7.226 listed by domain list.dnswl.org as 127.0.10.3
mailserver       | Feb 15 19:31:22 mail postfix/dnsblog[10975]: addr 87.246.7.226 listed by domain bl.mailspike.net as 127.0.0.2
mailserver       | Feb 15 19:31:22 mail postfix/postscreen[10972]: PASS NEW [87.246.7.226]:40982
mailserver       | Feb 15 19:31:22 mail postfix/smtpd[10984]: warning: hostname net6-ip226.linkbg.com does not resolve to address 87.246.7.226: Name or service not known
mailserver       | Feb 15 19:31:22 mail postfix/smtpd[10984]: connect from unknown[87.246.7.226]
mailserver       | Feb 15 19:31:24 mail dovecot: auth: passwd-file(ns2@blkchaintech.de,87.246.7.226): unknown user (SHA1 of given password: e15e2f)
mailserver       | Feb 15 19:31:26 mail postfix/smtpd[10984]: warning: unknown[87.246.7.226]: SASL LOGIN authentication failed: UGFzc3dvcmQ6
mailserver       | Feb 15 19:31:26 mail postfix/smtpd[10984]: disconnect from unknown[87.246.7.226] ehlo=1 auth=0/1 rset=1 quit=1 commands=3/4
mailserver       | Feb 15 19:32:07 mail postfix/postscreen[10972]: CONNECT from [87.246.7.226]:49306 to [192.168.176.2]:25
mailserver       | Feb 15 19:32:07 mail postfix/postscreen[10972]: PASS OLD [87.246.7.226]:49306
mailserver       | Feb 15 19:32:07 mail postfix/smtpd[10984]: warning: hostname net6-ip226.linkbg.com does not resolve to address 87.246.7.226: Name or service not known
mailserver       | Feb 15 19:32:07 mail postfix/smtpd[10984]: connect from unknown[87.246.7.226]
mailserver       | Feb 15 19:32:11 mail dovecot: auth: passwd-file(darwin@blkchaintech.de,87.246.7.226): unknown user (SHA1 of given password: 0c0655)
mailserver       | Feb 15 19:32:13 mail postfix/smtpd[10984]: warning: unknown[87.246.7.226]: SASL LOGIN authentication failed: UGFzc3dvcmQ6
mailserver       | Feb 15 19:32:13 mail postfix/smtpd[10984]: disconnect from unknown[87.246.7.226] ehlo=1 auth=0/1 rset=1 quit=1 commands=3/4
mailserver       | Feb 15 19:32:54 mail postfix/postscreen[10972]: CONNECT from [87.246.7.226]:57546 to [192.168.176.2]:25
mailserver       | Feb 15 19:32:54 mail postfix/dnsblog[10977]: addr 87.246.7.226 listed by domain bl.mailspike.net as 127.0.0.2
mailserver       | Feb 15 19:32:54 mail postfix/dnsblog[10974]: addr 87.246.7.226 listed by domain list.dnswl.org as 127.0.10.3
mailserver       | Feb 15 19:32:54 mail postfix/postscreen[10972]: PASS OLD [87.246.7.226]:57546
mailserver       | Feb 15 19:32:54 mail postfix/smtpd[10984]: warning: hostname net6-ip226.linkbg.com does not resolve to address 87.246.7.226: Name or service not known
mailserver       | Feb 15 19:32:54 mail postfix/smtpd[10984]: connect from unknown[87.246.7.226]
mailserver       | Feb 15 19:32:59 mail dovecot: auth: passwd-file(net@blkchaintech,87.246.7.226): unknown user (SHA1 of given password: 976950)
mailserver       | Feb 15 19:33:01 mail postfix/smtpd[10984]: warning: unknown[87.246.7.226]: SASL LOGIN authentication failed: UGFzc3dvcmQ6
mailserver       | Feb 15 19:38:01 mail postfix/smtpd[10984]: timeout after AUTH from unknown[87.246.7.226]
mailserver       | Feb 15 19:38:01 mail postfix/smtpd[10984]: disconnect from unknown[87.246.7.226] ehlo=1 auth=0/1 rset=1 commands=2/3
mailserver       | Feb 15 19:41:21 mail postfix/anvil[10986]: statistics: max connection rate 2/60s for (smtpd:87.246.7.226) at Feb 15 18:32:07
mailserver       | Feb 15 19:41:21 mail postfix/anvil[10986]: statistics: max connection count 1 for (smtpd:87.246.7.226) at Feb 15 18:31:22
mailserver       | Feb 15 19:41:21 mail postfix/anvil[10986]: statistics: max cache size 1 at Feb 15 18:31:22
[87.246.7.226
```

aendeavor:
# Description

This PR replaces F2B 0.10 with the latest F2B 0.11.2. This is done as Debian 11 has still no ETA. This change is easily revertible. When Debian 11 launches with the latest F2B, this PR can be reverted in a manner of minutes.

Fixes #1810 

## Type of change

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,7 +51,7 @@ RUN \
   dovecot-managesieved dovecot-pop3d dovecot-sieve dovecot-solr \
   dumb-init \
   # # E - O
-  ed fetchmail file gamin gnupg gzip gpg gpg-agent inotify-tools \
+  ed fetchmail file gamin gnupg gzip gpg gpg-agent iproute2 \
   iptables locales logwatch lhasa libdate-manip-perl liblz4-tool \
   libmail-spf-perl libnet-dns-perl libsasl2-modules lrzip lzop \
   netcat-openbsd nomarch opendkim opendkim-tools opendmarc \
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,9 +4,8 @@ ARG VCS_VER
 ARG VCS_REF
 ARG DEBIAN_FRONTEND=noninteractive
 
-ARG FAIL2BAN_VERSION=0.11.2
-ARG FAIL2BAN_URL=https://github.com/fail2ban/fail2ban/releases/download/${FAIL2BAN_VERSION}/fail2ban_${FAIL2BAN_VERSION}-1.upstream1_all.deb
-ARG FAIL2BAN_URL_ASC=https://github.com/fail2ban/fail2ban/releases/download/${FAIL2BAN_VERSION}/fail2ban_${FAIL2BAN_VERSION}-1.upstream1_all.deb.asc
+ARG FAIL2BAN_URL=https://github.com/fail2ban/fail2ban/releases/download/0.11.2/fail2ban_0.11.2-1.upstream1_all.deb
+ARG FAIL2BAN_URL_ASC=https://github.com/fail2ban/fail2ban/releases/download/0.11.2/fail2ban_0.11.2-1.upstream1_all.deb.asc
 ARG FAIL2BAN_PGP_PUBLIC_KEY_ID=0x683BF1BEBD0A882C
 ARG FAIL2BAN_PGP_PUBLIC_KEY_SERVER=keys.gnupg.net
 ARG FAIL2BAN_GPG_FINGERPRINT="8738 559E 26F6 71DF 9E2C  6D9E 683B F1BE BD0A 882C"
--- a/Dockerfile
+++ b/Dockerfile
@@ -67,8 +67,8 @@ RUN \
   wget -q --no-check-certificate ${FAIL2BAN_URL} && \
   wget -q --no-check-certificate ${FAIL2BAN_URL_ASC} && \
   FINGERPRINT="$(LANG=C gpg --verify \
-    fail2ban_${FAIL2BAN_VERSION}-1.upstream1_all.deb.asc \
-    fail2ban_${FAIL2BAN_VERSION}-1.upstream1_all.deb 2>&1 \
+  fail2ban_0.11.2-1.upstream1_all.deb.asc \
+  fail2ban_0.11.2-1.upstream1_all.deb 2>&1 \
     | sed -n 's#Primary key fingerprint: \(.*\)#\1#p')" && \
   if [[ -z ${FINGERPRINT} ]]; then \
     echo "ERROR: Invalid GPG signature!" 2>&1; exit 1; fi && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -60,24 +60,23 @@ RUN \
   pax pflogsumm postgrey p7zip-full postfix-ldap postfix-pcre \
   postfix-policyd-spf-python postsrsd pyzor \
   razor rpm2cpio rsyslog sasl2-bin spamassassin supervisor \
-  unrar-free unzip wget=1.20.1-1.1 whois xz-utils >/dev/null && \
+  unrar-free unzip whois xz-utils >/dev/null && \
   # Fail2Ban
   gpg --keyserver ${FAIL2BAN_PGP_PUBLIC_KEY_SERVER} \
     --recv-keys ${FAIL2BAN_PGP_PUBLIC_KEY_ID} &>/dev/null && \
-  wget -q --no-check-certificate ${FAIL2BAN_URL} && \
-  wget -q --no-check-certificate ${FAIL2BAN_URL_ASC} && \
+  curl -Lso fail2ban.deb ${FAIL2BAN_URL} && \
+  curl -Lso fail2ban.deb.asc ${FAIL2BAN_URL_ASC} && \
   FINGERPRINT="$(LANG=C gpg --verify \
-  fail2ban_0.11.2-1.upstream1_all.deb.asc \
-  fail2ban_0.11.2-1.upstream1_all.deb 2>&1 \
+  fail2ban.deb.asc fail2ban.deb 2>&1 \
     | sed -n 's#Primary key fingerprint: \(.*\)#\1#p')" && \
   if [[ -z ${FINGERPRINT} ]]; then \
     echo "ERROR: Invalid GPG signature!" 2>&1; exit 1; fi && \
   if [[ ${FINGERPRINT} != "${FAIL2BAN_GPG_FINGERPRINT}" ]]; then \
     echo "ERROR: Wrong GPG fingerprint!" 2>&1; exit 1; fi && \
-  dpkg -i fail2ban_${FAIL2BAN_VERSION}-1.upstream1_all.deb &>/dev/null && \
+  dpkg -i fail2ban.deb &>/dev/null && \
   # cleanup
-  rm -rf fail2ban_${FAIL2BAN_VERSION}-1.upstream1_all.deb && \
-  apt-get -qq -y purge wget gpg gpg-agent &>/dev/null && \
+  rm -rf fail2ban.deb fail2ban.deb.asc && \
+  apt-get -qq -y purge gpg gpg-agent &>/dev/null && \
   apt-get -qq autoremove &>/dev/null && \
   apt-get -qq autoclean && \
   apt-get -qq clean && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,13 @@ ARG VCS_VER
 ARG VCS_REF
 ARG DEBIAN_FRONTEND=noninteractive
 
+ARG FAIL2BAN_VERSION=0.11.2
+ARG FAIL2BAN_URL=https://github.com/fail2ban/fail2ban/releases/download/${FAIL2BAN_VERSION}/fail2ban_${FAIL2BAN_VERSION}-1.upstream1_all.deb
+ARG FAIL2BAN_URL_ASC=https://github.com/fail2ban/fail2ban/releases/download/${FAIL2BAN_VERSION}/fail2ban_${FAIL2BAN_VERSION}-1.upstream1_all.deb.asc
+ARG FAIL2BAN_PGP_PUBLIC_KEY_ID=0x683BF1BEBD0A882C
+ARG FAIL2BAN_PGP_PUBLIC_KEY_SERVER=keys.gnupg.net
+ARG FAIL2BAN_GPG_FINGERPRINT="8738 559E 26F6 71DF 9E2C  6D9E 683B F1BE BD0A 882C"
+
 LABEL org.opencontainers.image.version=${VCS_VER}
 LABEL org.opencontainers.image.revision=${VCS_REF}
 LABEL org.opencontainers.image.title="docker-mailserver"
@@ -44,21 +51,35 @@ RUN \
   dovecot-core dovecot-imapd dovecot-ldap dovecot-lmtpd \
   dovecot-managesieved dovecot-pop3d dovecot-sieve dovecot-solr \
   dumb-init \
-  # E - O
-  ed fetchmail file gamin gnupg gzip iproute2 iptables \
-  locales logwatch lhasa libdate-manip-perl liblz4-tool \
+  # # E - O
+  ed fetchmail file gamin gnupg gzip gpg=2.2.12-1+deb10u1 \
+  gpg-agent=2.2.12-1+deb10u1 inotify-tools \
+  iptables locales logwatch lhasa libdate-manip-perl liblz4-tool \
   libmail-spf-perl libnet-dns-perl libsasl2-modules lrzip lzop \
   netcat-openbsd nomarch opendkim opendkim-tools opendmarc \
   # P - Z
   pax pflogsumm postgrey p7zip-full postfix-ldap postfix-pcre \
   postfix-policyd-spf-python postsrsd pyzor \
   razor rpm2cpio rsyslog sasl2-bin spamassassin supervisor \
-  unrar-free unzip whois xz-utils >/dev/null && \
+  unrar-free unzip wget=1.20.1-1.1 whois xz-utils >/dev/null && \
   # Fail2Ban
-  curl -Lso fail2ban.deb https://github.com/fail2ban/fail2ban/releases/download/0.11.2/fail2ban_0.11.2-1.upstream1_all.deb && \
-  dpkg -i fail2ban.deb && \
-  rm fail2ban.deb && \
+  gpg --keyserver ${FAIL2BAN_PGP_PUBLIC_KEY_SERVER} \
+    --recv-keys ${FAIL2BAN_PGP_PUBLIC_KEY_ID} &>/dev/null && \
+  wget -q --no-check-certificate ${FAIL2BAN_URL} && \
+  wget -q --no-check-certificate ${FAIL2BAN_URL_ASC} && \
+  FINGERPRINT="$(LANG=C gpg --verify \
+    fail2ban_${FAIL2BAN_VERSION}-1.upstream1_all.deb.asc \
+    fail2ban_${FAIL2BAN_VERSION}-1.upstream1_all.deb 2>&1 \
+    | sed -n 's#Primary key fingerprint: \(.*\)#\1#p')" && \
+  if [[ -z ${FINGERPRINT} ]]; then \
+    echo "ERROR: Invalid GPG signature!" 2>&1; exit 1; fi && \
+  if [[ ${FINGERPRINT} != "${FAIL2BAN_GPG_FINGERPRINT}" ]]; then \
+    echo "ERROR: Wrong GPG fingerprint!" 2>&1; exit 1; fi && \
+  dpkg -i fail2ban_${FAIL2BAN_VERSION}-1.upstream1_all.deb && \
   # cleanup
+  rm -rf fail2ban_${FAIL2BAN_VERSION}-1.upstream1_all.deb && \
+  apt-get purge -qq -y wget gpg gpg-agent && \
+  apt-get autoremove &>/dev/null && \
   apt-get -qq autoclean && \
   apt-get -qq clean && \
   rm -rf /var/lib/apt/lists/* && \
@@ -69,7 +90,7 @@ RUN \
 # –––––––––––––––––––––––––––––––––––––––––––––––
 
 RUN \
-  echo '0 */6 * * * clamav /usr/bin/freshclam --quiet' > /etc/cron.d/clamav-freshclam && \
+  echo '0 */6 * * * clamav /usr/bin/freshclam --quiet' >/etc/cron.d/clamav-freshclam && \
   chmod 644 /etc/clamav/freshclam.conf && \
   freshclam && \
   sed -i 's/Foreground false/Foreground true/g' /etc/clamav/clamd.conf && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,8 +4,8 @@ ARG VCS_VER
 ARG VCS_REF
 ARG DEBIAN_FRONTEND=noninteractive
 
-ARG FAIL2BAN_DEB=https://github.com/fail2ban/fail2ban/releases/download/0.11.2/fail2ban_0.11.2-1.upstream1_all.deb
-ARG FAIL2BAN_DEB_ASC=${FAIL2BAN_DEB}.asc
+ARG FAIL2BAN_DEB_URL=https://github.com/fail2ban/fail2ban/releases/download/0.11.2/fail2ban_0.11.2-1.upstream1_all.deb
+ARG FAIL2BAN_DEB_ASC_URL=${FAIL2BAN_DEB_URL}.asc
 ARG FAIL2BAN_GPG_PUBLIC_KEY_ID=0x683BF1BEBD0A882C
 ARG FAIL2BAN_GPG_PUBLIC_KEY_SERVER=keys.gnupg.net
 ARG FAIL2BAN_GPG_FINGERPRINT="8738 559E 26F6 71DF 9E2C  6D9E 683B F1BE BD0A 882C"
@@ -64,8 +64,8 @@ RUN \
   gpg gpg-agent >/dev/null && \
   gpg --keyserver ${FAIL2BAN_GPG_PUBLIC_KEY_SERVER} \
     --recv-keys ${FAIL2BAN_GPG_PUBLIC_KEY_ID} &>/dev/null && \
-  curl -Lso fail2ban.deb ${FAIL2BAN_DEB} && \
-  curl -Lso fail2ban.deb.asc ${FAIL2BAN_DEB_ASC} && \
+  curl -Lso fail2ban.deb ${FAIL2BAN_DEB_URL} && \
+  curl -Lso fail2ban.deb.asc ${FAIL2BAN_DEB_ASC_URL} && \
   FINGERPRINT="$(LANG=C gpg --verify \
   fail2ban.deb.asc fail2ban.deb 2>&1 \
     | sed -n 's#Primary key fingerprint: \(.*\)#\1#p')" && \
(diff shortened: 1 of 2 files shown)


aendeavor:
@casperklein Although unit tests fail, it is only one tests, which is version related:

```
 not ok 28 checking setup.sh: setup.sh debug fail2ban
# (from function `assert_output' in file test/test_helper/bats-assert/src/assert.bash, line 231,
#  in test file test/mail_fail2ban.bats, line 143)
#   `assert_output --partial "Unbanned IP from dovecot: 192.0.66.4"' failed
# 
# substring (1 lines):
#   Unbanned IP from dovecot: 192.0.66.4
# output (3 lines):
#   Unbanned IP from dovecot: 1
#   Unbanned IP from postfix: 0
#   Unbanned IP from postfix-sasl: 0
# --
# 
```

Can you have a look?

aendeavor:
@docker-mailserver/maintainers If someone is not fine with getting the latest Fail2Ban, please share your concerns here :)

wernerfred:
> @docker-mailserver/maintainers If someone is not fine with getting the latest Fail2Ban, please share your concerns here :)

If it doesn't solve the issue discussed in #1810 and still needs the "hacking" i see no real benefit. If there is an advantage then go for it (imho)

aendeavor:
@wernerfred You right: I think that

> I've tested 0.11 on the host and it flawlessly banned all the "auth failures"

is one good arguments. The other one may be that we should be at the forefront of security. I like having the latest versions :D

Dockerfile
```diff
@@ -54,6 +54,10 @@ RUN \
   postfix-policyd-spf-python postsrsd pyzor \
   razor rpm2cpio rsyslog sasl2-bin spamassassin supervisor \
   unrar-free unzip whois xz-utils >/dev/null && \
+  # Fail2Ban
+  curl -Lso fail2ban.deb https://github.com/fail2ban/fail2ban/releases/download/0.11.2/fail2ban_0.11.2-1.upstream1_all.deb && \
```
wernerfred, line 58: If we do this to be on forefront of security then we should check the downloaded files integrity/signing shouldn't we?

aendeavor: Yes, we should do that. Can you help?

wernerfred: Give me a minute

aendeavor: One easy method, without using GPG, would be to check the pre-computed SHA152 sum like this:

``` BASH
sha512sum --check --status <<< "09d91ad0f499b0825b9d6da1d175400aa32498aff5705624e88b3d7e02cf727403c49b967c732e42182e25d4f439de0faf681f9f6d3598364e3f2f2e3453f8bf  fail2ban_0.11.2-1.upstream1_all.deb"
``` 

which could be written directly into the Dockerfile.

casperklein: This should be enough, to ensure the file was not altered.

However, my checksum differs:

`echo "44fe3d67e7d24bba4fc81ee672059dcb67980258db6101bcd8740736a9d9110f *fail2ban.deb" | sha256sum --check --status`

or

`sha256sum --check --status <<< "44fe3d67e7d24bba4fc81ee672059dcb67980258db6101bcd8740736a9d9110f *fail2ban.deb"`

wernerfred: Here is an (truncated Dockerfile) example using pgp:

```
ARG FAIL2BAN_VERSION=0.11.2
ARG FAIL2BAN_URL=https://github.com/fail2ban/fail2ban/releases/download/${FAIL2BAN_VERSION}/fail2ban_${FAIL2BAN_VERSION}-1.upstream1_all.deb
ARG FAIL2BAN_URL_ASC=https://github.com/fail2ban/fail2ban/releases/download/${FAIL2BAN_VERSION}/fail2ban_${FAIL2BAN_VERSION}-1.upstream1_all.deb.asc

RUN apt-get install -q -y --no-install-recommends gpg=2.2.12-1+deb10u1 \
 && wget ${FAIL2BAN_URL} \
 && wget ${FAIL2BAN_URL_ASC} \
 && gpg --verify fail2ban_${FAIL2BAN_VERSION}-1.upstream1_all.deb.asc fail2ban_${FAIL2BAN_VERSION}-1.upstream1_all.deb
 && dpkg -i fail2ban_${FAIL2BAN_VERSION}-1.upstream1_all.deb
 # && apt cleanup of gpg etc
```

If someone can provide me the fail2ban public key and the fingerprint we could additionally check the fingerprint (like i did [here](https://github.com/wernerfred/docker-rainloop/blob/755a7776bb9fbc6bcdc427acc1d303a42dd84956/Dockerfile#L33))

aendeavor: @casperklein You used `sha256sum`, I used `sha512sum`.

@wernerfred Your solution looks very good to me. As you have made yourself the work of writing this up, we should use it (and of course, since it is better than just checking a hash sum). You're right, we need to get the pubkey and fingerprint. Shall I open an issue in the F2B repo and ask?

wernerfred: I did a short research but didn't find it directly. Maybe it is the easiest approach to ask directly

aendeavor: @wernerfred Have a look at the issue that is linked to the F2B repo. A contributor explained how to get pubkey and fingerprint.

wernerfred: This should be it:
```
ARG FAIL2BAN_VERSION=0.11.2
ARG FAIL2BAN_URL=https://github.com/fail2ban/fail2ban/releases/download/${FAIL2BAN_VERSION}/fail2ban_${FAIL2BAN_VERSION}-1.upstream1_all.deb
ARG FAIL2BAN_URL_ASC=https://github.com/fail2ban/fail2ban/releases/download/${FAIL2BAN_VERSION}/fail2ban_${FAIL2BAN_VERSION}-1.upstream1_all.deb.asc
ARG FAIL2BAN_PGP_PUBLIC_KEY_ID=0x683BF1BEBD0A882C
ARG FAIL2BAN_PGP_PUBLIC_KEY_SERVER=keys.gnupg.net
ARG FAIL2BAN_GPG_FINGERPRINT="8738 559E 26F6 71DF 9E2C 6D9E 683B F1BE BD0A 882C"

RUN apt-get install -q -y --no-install-recommends wget=1.20.1-1.1 gpg=2.2.12-1+deb10u1 gpg-agent=2.2.12-1+deb10u1\
 && apt-get clean \
 && rm -rf /var/lib/apt/lists/* \
 && gpg --keyserver ${FAIL2BAN_PGP_PUBLIC_KEY_SERVER} --recv-keys ${FAIL2BAN_PGP_PUBLIC_KEY_ID}
 && wget -q ${FAIL2BAN_URL} \
 && wget -q ${FAIL2BAN_URL_ASC} \
 && FINGERPRINT="$(LANG=C gpg --verify fail2ban_${FAIL2BAN_VERSION}-1.upstream1_all.deb.asc fail2ban_${FAIL2BAN_VERSION}-1.upstream1_all.deb 2>&1 \
    | sed -n "s#Primary key fingerprint: \(.*\)#\1#p")" \
 && if [ -z "${FINGERPRINT}" ]; then echo "ERROR: Invalid GPG signature!" && exit 1; fi \
 && if [ "${FINGERPRINT}" != "${FAIL2BAN_GPG_FINGERPRINT}" ]; then echo "ERROR: Wrong GPG fingerprint!" && exit 1; fi \
 && dpkg -i fail2ban_${FAIL2BAN_VERSION}-1.upstream1_all.deb
 && rm -rf fail2ban_${FAIL2BAN_VERSION}-1.upstream1_all.deb
 && apt-get purge -q -y wget gpg gpg-agent \
 && apt-get autoremove -y \
```

I have no time to test right now, might be that we need a bit more tweaking on cleanup or so. Also the `${FAIL2BAN_VERSION}` should be reworked bc other relases of F2B do not have the `-1.upstream` which is hard coded in my example.

I am able to test this today evening/night. If someone wants to do this earlier: do it 👍🏻

aendeavor: I have prepared the Dockerfile for you, and with latest commit it should suffice. You can give it a try locally, and check if you think everything is okay. I had to change some things here and there, but the overall construction stayed the same.

casperklein: - Is there a reason for the package pinning in that particular case (we don't do this for the other packages)? If we do it that way, we are probably going to install old versions of these packages maybe in some weeks, if some does not constantly check for new versions and adjust the Dockerfile.

- `ARG` should be `ENV`

- `ENV FAIL2BAN_URL_ASC=${FAIL2BAN_URL}.asc`

> ${FAIL2BAN_VERSION} [..] which is hard coded in my example.

IMO we don't need `FAIL2BAN_VERSION` at all. There are not that much releases of f2b. And as you said, when there is a new version, we have to adjust the Dockerfile and maybe f2b_url either way.

casperklein: In context of security, this feels wrong: `wget -q --no-check-certificate`.

Serious question: What kind of improved security do we get from this, compared to the very simple file hash comparison @aendeavor suggested? With both methods, we can assure, that the .deb file was not altered.

aendeavor: > In context of security, this feels wrong: `wget -q --no-check-certificate`.
> 
> 
> 
> Serious question: What kind of improved security do we get from this, compared to the very simple file hash comparison @aendeavor suggested? With both methods, we can assure, that the .deb file was not altered.

This was something I asked myself too. Maybe the hash sum is the best solution as it is simple.

The `--no-check-certificate` is needed in CI as wget behaves strangely in pipelines. I have seen this behaviour in GitLab CI too.

At the moment, I think that the hash may be the better solution in the end...

wernerfred: > IMO we don't need FAIL2BAN_VERSION at all. There are not that much releases of f2b. And as you said, when there is a new version, we have to adjust the Dockerfile and maybe f2b_url either way.

Agree 👍🏻 

> In context of security, this feels wrong: wget -q --no-check-certificate. 

True but doesn't matter if the signing is correct

> What kind of improved security do we get from this, compared to the very simple file hash compariso

We are not only checking if the file got altered but also if it was created by fail2ban or did i get it wrong?
Nevertheless the hash comparison adds that bit of security compared to having no check at all that i am fine with both ☑️

wernerfred: docker build passes without error or warning, i'm running the test suite right now

Dockerfile
```diff
@@ -44,21 +51,35 @@ RUN \
   dovecot-core dovecot-imapd dovecot-ldap dovecot-lmtpd \
   dovecot-managesieved dovecot-pop3d dovecot-sieve dovecot-solr \
   dumb-init \
-  # E - O
-  ed fetchmail file gamin gnupg gzip iproute2 iptables \
-  locales logwatch lhasa libdate-manip-perl liblz4-tool \
+  # # E - O
+  ed fetchmail file gamin gnupg gzip gpg=2.2.12-1+deb10u1 \
```
wernerfred, line 55:
```suggestion
  ed fetchmail file gamin gnupg gzip gpg\
```

Dockerfile
```diff
@@ -44,21 +51,35 @@ RUN \
...
+  gpg-agent=2.2.12-1+deb10u1 inotify-tools \
```
wernerfred, line 56:
```suggestion
  gpg-agent inotify-tools \
```

Dockerfile
```diff
@@ -44,21 +51,35 @@ RUN \
...
+  iptables locales logwatch lhasa libdate-manip-perl liblz4-tool \
   libmail-spf-perl libnet-dns-perl libsasl2-modules lrzip lzop \
   netcat-openbsd nomarch opendkim opendkim-tools opendmarc \
   # P - Z
   pax pflogsumm postgrey p7zip-full postfix-ldap postfix-pcre \
   postfix-policyd-spf-python postsrsd pyzor \
   razor rpm2cpio rsyslog sasl2-bin spamassassin supervisor \
-  unrar-free unzip whois xz-utils >/dev/null && \
+  unrar-free unzip wget=1.20.1-1.1 whois xz-utils >/dev/null && \
```
wernerfred, line 64: HADOLINT reports error to use curl or wget in a Dockerfile but not both, we should switch to curl

Dockerfile
```diff
@@ -4,6 +4,13 @@ ARG VCS_VER
 ARG VCS_REF
 ARG DEBIAN_FRONTEND=noninteractive
 
+ARG FAIL2BAN_VERSION=0.11.2
+ARG FAIL2BAN_URL=https://github.com/fail2ban/fail2ban/releases/download/${FAIL2BAN_VERSION}/fail2ban_${FAIL2BAN_VERSION}-1.upstream1_all.deb
+ARG FAIL2BAN_URL_ASC=https://github.com/fail2ban/fail2ban/releases/download/${FAIL2BAN_VERSION}/fail2ban_${FAIL2BAN_VERSION}-1.upstream1_all.deb.asc
```
wernerfred, line 9:
```suggestion
ARG FAIL2BAN_URL=https://github.com/fail2ban/fail2ban/releases/download/0.11.2/fail2ban_0.11.2-1.upstream1_all.deb
ARG FAIL2BAN_URL_ASC=https://github.com/fail2ban/fail2ban/releases/download/0.11.2/fail2ban_0.11.2-1.upstream1_all.deb.asc
```

Dockerfile
```diff
@@ -44,21 +51,35 @@ RUN \
...
-  curl -Lso fail2ban.deb https://github.com/fail2ban/fail2ban/releases/download/0.11.2/fail2ban_0.11.2-1.upstream1_all.deb && \
-  dpkg -i fail2ban.deb && \
-  rm fail2ban.deb && \
+  gpg --keyserver ${FAIL2BAN_PGP_PUBLIC_KEY_SERVER} \
+    --recv-keys ${FAIL2BAN_PGP_PUBLIC_KEY_ID} &>/dev/null && \
+  wget -q --no-check-certificate ${FAIL2BAN_URL} && \
+  wget -q --no-check-certificate ${FAIL2BAN_URL_ASC} && \
+  FINGERPRINT="$(LANG=C gpg --verify \
+    fail2ban_${FAIL2BAN_VERSION}-1.upstream1_all.deb.asc \
+    fail2ban_${FAIL2BAN_VERSION}-1.upstream1_all.deb 2>&1 \
```
wernerfred, line 72:
```suggestion
  fail2ban_0.11.2-1.upstream1_all.deb.asc \
  fail2ban_0.11.2-1.upstream1_all.deb 2>&1 \
```

Dockerfile
```diff
@@ -44,21 +51,35 @@ RUN \
...
+    | sed -n 's#Primary key fingerprint: \(.*\)#\1#p')" && \
+  if [[ -z ${FINGERPRINT} ]]; then \
+    echo "ERROR: Invalid GPG signature!" 2>&1; exit 1; fi && \
+  if [[ ${FINGERPRINT} != "${FAIL2BAN_GPG_FINGERPRINT}" ]]; then \
+    echo "ERROR: Wrong GPG fingerprint!" 2>&1; exit 1; fi && \
+  dpkg -i fail2ban_${FAIL2BAN_VERSION}-1.upstream1_all.deb && \
   # cleanup
+  rm -rf fail2ban_${FAIL2BAN_VERSION}-1.upstream1_all.deb && \
```
wernerfred:
```suggestion
  dpkg -i fail2ban_0.11.2-1.upstream1_all.deb && \
  # cleanup
  rm -rf fail2ban_0.11.2-1.upstream1_all.deb && \
```

Dockerfile
```diff
@@ -44,21 +51,35 @@ RUN \
...
+  apt-get purge -qq -y wget gpg gpg-agent && \
```
wernerfred: switch wget to curl or do we need curl later on?

wernerfred (commented):
I suggested changes to drop the `FAIL2BAN_VERSION` and removed the version pinning of packages

Dockerfile
aendeavor: We should stick with `curl`

Dockerfile
aendeavor: > docker build passes without error or warning, i'm running the test suite right now
> 
> 
> 
> image
> 
> 

I will commit some of the suggested changes, but you'd need to do final adjustments to this yourself:)

Dockerfile
aendeavor: > HADOLINT reports error to use curl or wget in a Dockerfile but not both, we should switch to curl

Yes, we should use `curl`

Dockerfile
casperklein: BTW: I like saving the package just to `fail2ban.deb`:  curl -Lso fail2ban.deb ${FAIL2BAN_URL}`

Then just  `dpkg -i fail2ban.deb` etc. is needed. Looks more clean IMO 😃

Dockerfile
casperklein: > We are not only checking if the file got altered but also if it was created by fail2ban or did i get it wrong?

Correct. I considered this only as a temporary addition (until Debian 11 is released). So manually sign checking once and then hard-coding the checksum would be enough (way less additions to Dockerfile needed).

If we are going this route long-term, then the gpg solution is more complete and convenient 👍

> We only need them during build an therefore ARG should be enough.

Good point. I had the wrong assumption, that using ARG in Dockerfile and not providing a build-arg for that during the build would throw a warning. That is not the case.
Maybe we should think about replacing most ENV with ARG in the whole Dockerfile. The ENVs are populated by the start-mailserver.sh script. So we don't need to bake them into the image.

wernerfred: > The ENVs are populated by the start-mailserver.sh script. So we don't need to bake them into the image.

Good point

aendeavor:
@casperklein @wernerfred I'm low on time today and tomorrow (Bachelor's thesis is waiting for me). Can you take care of the changes you suggested. I'm more than happy to review this in the end.

Dockerfile
wernerfred: Added with https://github.com/docker-mailserver/docker-mailserver/commit/3b56e841e64b0081c62c27248a1378ba235b6160

Dockerfile
wernerfred: Added with https://github.com/docker-mailserver/docker-mailserver/commit/3b56e841e64b0081c62c27248a1378ba235b6160

Dockerfile
wernerfred: Added with https://github.com/docker-mailserver/docker-mailserver/commit/3b56e841e64b0081c62c27248a1378ba235b6160

aendeavor:
LGTM

Anything left to do?

casperklein (approved):

wernerfred (approved):
I think this is fine to merge.
Do we need a documentation Update?

aendeavor:
> I think this is fine to merge.
> Do we need a documentation Update?

I don't think docs needs an update. I will go ahead and merge this.

casperklein:
Not doc, but we should mention it in the changelog as a breaking change.

casperklein:
I just wondered, why there is no new image @ dockerhub. --> [`build-multiarch-and-publish` failed](https://github.com/docker-mailserver/docker-mailserver/runs/2589892407?check_suite_focus=true) :confused:

wernerfred:
Do we have a arch problem?

casperklein:
Yes. But what exactly goes wrong is hard to tell, because all error output is suppressed.

wernerfred:
What are the next steps? How to debug this? Should we open a PR after #1970 got merged to see the build output on arm or do we revert this PR?

casperklein:
#1970 does not add additional output. It will only help for future PRs, so that we know in advance before merging, the build fails.

To get more output and track down which command exactly fails, we would have to remove all the `&>/dev/null`, `-qq` etc parts.

I did this locally, but unfortunately, my build for arm64/armv7 both succeed. I looked for the "exit code 60" in google and the first result was about a certificate problem of curl. @aendeavor mentioned:

> The --no-check-certificate is needed in CI as wget behaves strangely in pipelines

If I had to blindly guess, I would say, that curl behaves similar. `curl --insecure` should then fix it.

To track down this and future problems more easily, it's probably a good idea, to remove the `&>/dev/null` statements.

aendeavor:
> I did this locally, but unfortunately, my build for arm64/armv7 both succeed. I looked for the "exit code 60" in google and the first result was about a certificate problem of curl. @aendeavor mentioned:
> 
> > The --no-check-certificate is needed in CI as wget behaves strangely in pipelines
> 
> If I had to blindly guess, I would say, that curl behaves similar. `curl --insecure` should then fix it.

That is probably it.

> To track down this and future problems more easily, it's probably a good idea, to remove the `&>/dev/null` statements.

Yes.

casperklein:
Another problem: With the commit of #1970 `test_merge_requests.yml` got broken, because it's not using buildx compared to `default_on_push.yml`

As long this is not fixed, we cannot test #1971.

casperklein:
I can try to adjust `test_merge_requests.yml` to use buildx, but that will take some time as I am not that familiar with gh actions.

@wernerfred @aendeavor if one of you could do it, I wouldn't be sad :kissing_heart:

Or do you have any other suggestions how to proceed?

aendeavor:
> I can try to adjust `test_merge_requests.yml` to use buildx, but that will take some time as I am not that familiar with gh actions.
> 
> @wernerfred @aendeavor if one of you could do it, I wouldn't be sad :kissing_heart:
> 
> Or do you have any other suggestions how to proceed?

Still low on time to be honest 🙈 I have not done the build stuff at all. But I will help if I can. Remember, we've got time, no need to hurry. `9.1.0` is out there and solid :D

casperklein:
I had a look and will give it a try. Doesn't seem that hard as I thought.

wernerfred:
> If I had to blindly guess, I would say, that curl behaves similar. `curl --insecure` should then fix it.

But why does it work in the PR test then? This run is also in CI environment. Strange

casperklein:
Good point. That would indicate it's just related to arm builds? Or the curl guess is wrong.

wernerfred:
> To get more output and track down which command exactly fails, we would have to remove all the `&>/dev/null`, `-qq` etc parts.

We can still revert #1965 and restore the `f2b-0.11` branch. Then master will be clean again and we can use the recovered branch to create another PR that doesn't use `-qq` etc. to nail down the issue. If it passes (bc your PR #1970 will show it then) we can add back those arguments and merge.

aendeavor:
When #1972 passes, do we really need to revert this PR? We could just add another that removes the `-qq` and `[&]>/dev/null`, if want to.

wernerfred:
Absolute, will have the Same result. Just a Point of having a broken state commited to master for a certain amount of time

aendeavor:
> Absolute, will have the Same result. Just a Point of having a broken state commited to master for a certain amount of time

I can live with that, since `master` represents `:edge` and does not build stable images :D

casperklein:
Another downside of the `&>/dev/null` snippets is, that we missed apt-get info, that gnupg is already available and don't need to be installed for fail2ban 😉 

https://github.com/docker-mailserver/docker-mailserver/blob/225e21edb7978db6e16438f9f4c564c778f66ddd/Dockerfile#L54

PR #1970, opened 2021-05-15
Add ARM platforms

casperklein:
# Description

ARM platforms added to test action.

If we only test against amd64, later (multiarch) build & publish action can fail.

See:

* https://github.com/docker-mailserver/docker-mailserver/pull/1965
* https://github.com/docker-mailserver/docker-mailserver/runs/2589892407?check_suite_focus=true

## Type of change

## Checklist:

### DIFF
--- a/.github/workflows/test_merge_requests.yml
+++ b/.github/workflows/test_merge_requests.yml
@@ -26,7 +26,7 @@ jobs:
           build-args: |
             VCS_REF=${{ github.sha }}
             VCS_VER=${{ github.ref }}
-          platforms: linux/amd64
+          platforms: linux/amd64,linux/arm/v7,linux/arm64
           load: true
           tags: mailserver-testing:ci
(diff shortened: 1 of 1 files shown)


aendeavor:
I would only test against `arm64` to reduce time spent in testing. Otherwise, LGTM

casperklein:
It's not guaranteed, that if the arm64 build succeed, arm/v7 will also succeed.

We should test against all platforms we build images for. The tests run in parallel, so I don't think there will be much time saved, when excluding arm/v7.

wernerfred (approved):

aendeavor (approved):

casperklein:
@aendeavor In the end, you were right.

Building for amd64 platform alone takes about 2 minutes. However **emulated** (QEMU) arm builds are taking 18min. This adds ~16 minutes to the build process.

PR #1971, opened 2021-05-15
ARM build fix

casperklein:
# Description

Possible arm build fix

Fixes #1965

## Type of change

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -64,8 +64,8 @@ RUN \
   gpg gpg-agent >/dev/null && \
   gpg --keyserver ${FAIL2BAN_GPG_PUBLIC_KEY_SERVER} \
     --recv-keys ${FAIL2BAN_GPG_PUBLIC_KEY_ID} &>/dev/null && \
-  curl -Lso fail2ban.deb ${FAIL2BAN_DEB_URL} && \
-  curl -Lso fail2ban.deb.asc ${FAIL2BAN_DEB_ASC_URL} && \
+  curl -Lkso fail2ban.deb ${FAIL2BAN_DEB_URL} && \
+  curl -Lkso fail2ban.deb.asc ${FAIL2BAN_DEB_ASC_URL} && \
   FINGERPRINT="$(LANG=C gpg --verify \
   fail2ban.deb.asc fail2ban.deb 2>&1 \
     | sed -n 's#Primary key fingerprint: \(.*\)#\1#p')" && \


aendeavor:
The error shown by `docker buildx` is very cryptic indeed... I'd just restart the test.

aendeavor:
And why did the latest pipeline on `master` succeed? 🤣

casperklein:
See my https://github.com/docker-mailserver/docker-mailserver/pull/1965#issuecomment-841707214

aendeavor:
Wait. I may miss something here, but couldn't we:

1. Turn on the `--insecure` setting for `curl`, as this _could_/_should_ be the reason for CI to fail
2. Revert #1970 to just test against `amd64`
3. With a bit of luck, habe CI fixed and a lot less work :D

Did I miss something?

casperklein:
Yes, also possible. But I'll start now on adding buildx. Let's see how this goes 😄

casperklein:
Another CI adjustment is needed. I am just testing it. If it's fine, i'll provide a hopefully final PR.

casperklein:
The CI will be fine with #1974 👍

wernerfred (approved):

wernerfred:
oh i approved the changes but the workflow wasn't finished. Auto-merge went through tho so we need to keep an eye on the running CI

casperklein:
Tests are also fine 🎉

polarathene:
> oh i approved the changes but the workflow wasn't finished. Auto-merge went through tho so we need to keep an eye on the running CI

When I contributed to Gatsby they had a auto-merge on green (all tests pass), not sure if that's supported by default on Github for maintainers or was via some third-party integration they added.

Additionally with Github Actions, that may not always be reliable as some CI jobs can be related but triggered after the initial checks/actions complete. I've seen the UI only show 1 listing in the check suite as a result and treat it as "green" but then the 2nd workflow which manually manages it's status causing the UI to update with a 2nd check item. Having some delay timer might work around that concern.

wernerfred:
The problem is that we can define which workflows need to pass at least before a merge is possible. We have chosen the "linting" one here as this one runs every time. We could also add the "test_pr" but then this one also needs to run even if we only add a typo fix on readme for example.
I will double check if there are new settings available, otherwise we be especially cautious about the "enable auto-merge" feature if there is a test running currently (or to approve only if the test passes).

PR #1989, opened 2021-05-20
fix!: Allow protocol in SASLAUTHD_LDAP_SERVER & adjust SASLAUTHD_LDAP_ default values

### LINKED ISSUE #1983: [BUG] SASLAUTHD_LDAP_SSL=1 does not set ldaps://
# Bug Report

Setting the environmental variable SASLAUTHD_LDAP_SSL to "1" does not set "ldaps://" in /etc/saslauthd.conf.  This is breaking saslauthd from working as the parameter "ldap_servers" is not set to a proper URI.

### When does this occur?

When setting the environmental variable SASLAUTHD_LDAP_SSL to "1" 

Set the environmental variable SASLAUTHD_LDAP_SSL to "1".

### Actual Behavior

The parameter "ldap_servers" in the file /etc/saslauthd.conf is being set like the below when setting the environmental variable SASLAUTHD_LDAP_SSL to "1". 

```
ldap_servers: 10.10.10.10
```

I think the problem seems to be with the logic in https://github.com/docker-mailserver/docker-mailserver/blob/master/target/scripts/startup/setup-stack.sh:

```
if [[ -z ${SASLAUTHD_LDAP_SSL} ]] || [[ ${SASLAUTHD_LDAP_SSL} -eq 0 ]]
then
    SASLAUTHD_LDAP_PROTO='ldap://' || SASLAUTHD_LDAP_PROTO='ldaps://'
fi
```

When the environmental variable SASLAUTHD_LDAP_SSL to "1", the logic above will set SASLAUTHD_LDAP_PROTO to "" (i.e empty).  Later in the code when SASLAUTHD_LDAP_PROTO is used, this parameter will be "" (i.e empty) and so SASLAUTHD_LDAP_SERVER will only be printed.

```
if [[ ! -f /etc/saslauthd.conf ]]  then
    _notify 'inf' 'Creating /etc/saslauthd.conf'
    cat &gt; /etc/saslauthd.conf &lt;&lt; EOF
ldap_servers: ${SASLAUTHD_LDAP_PROTO}${SASLAUTHD_LDAP_SERVER}
```

If I am not mistaken, I think this logic needs to be updated so that it properly sets SASLAUTHD_LDAP_PROTO='ldaps://' when the environmental variable SASLAUTHD_LDAP_SSL to "1". 

### Expected Behavior

The parameter "ldap_servers" in the file /etc/saslauthd.conf should be set like the below when setting the environmental variable SASLAUTHD_LDAP_SSL to "1".

ldap_servers: ldaps://10.10.10.10

## Your Environment

- version: `v`
- available RAM: `GB`
- Docker version: `v`
- docker-compose version: `v`

### Environment Variables

```CFG
SASLAUTHD_LDAP_SSL=1 
```

### Relevant Stack Traces

```BASH
# BEGIN

# END
```

moqmar:
# Description

Deprecate SASLAUTHD_LDAP_SSL by allowing the protocol in SASLAUTHD_LDAP_SERVER & use more sane defaults for some SASLAUTHD_LDAP variables.

Fixes #1983

## Type of change

## Checklist:

- [x] My code follows the style guidelines of this project (I used the style of the code around my changes, I think there are some inconsistencies with the default values in the saslauthd section compared to the other code)
- [x] ~~If necessary I have added tests that prove my fix is effective or that my feature works~~
- [ ] New and existing unit tests pass locally with my changes (will have to test this tomorrow when I have a better internet connection)


aendeavor:
This PR is to be merged before the release of v10 as it contains breaking changes:) Just FYI

aendeavor:
@moqmar please reach back to us when this PR is ready to be reviewed, so we can take action:)

moqmar:
Alright, I'd say if the checks run fine now then this is done!

Note that there are currently no regression tests for this as there's only one setup_file, so that would require big changes to the testing methodology.

test/mail_with_ldap.bats
```diff
@@ -40,10 +40,8 @@ function setup_file() {
 		-e REPORT_RECIPIENT=1 \
 		-e ENABLE_SASLAUTHD=1 \
 		-e SASLAUTHD_MECHANISMS=ldap \
-		-e SASLAUTHD_LDAP_SERVER=ldap \
-		-e SASLAUTHD_LDAP_BIND_DN=cn=admin,dc=localhost,dc=localdomain \
-		-e SASLAUTHD_LDAP_PASSWORD=admin \
-		-e SASLAUTHD_LDAP_SEARCH_BASE=ou=people,dc=localhost,dc=localdomain \
+		-e SASLAUTHD_LDAP_SERVER=ldap://ldap \
+		-e SASLAUTHD_LDAP_SSL=0 \
```
polarathene, line 44: If this is being removed elsewhere, why is it added here now? The associated commit message wasn't very clear.

Presumably for testing the case in `setup-stack.sh`?

target/scripts/startup/setup-stack.sh
```diff
@@ -618,13 +618,23 @@ function _setup_saslauthd
...
+  [[ -z ${SASLAUTHD_LDAP_BIND_DN} ]] && SASLAUTHD_LDAP_BIND_DN="${LDAP_BIND_DN}"
+  [[ -z ${SASLAUTHD_LDAP_PASSWORD} ]] && SASLAUTHD_LDAP_PASSWORD="${LDAP_BIND_PW}"
+  [[ -z ${SASLAUTHD_LDAP_SEARCH_BASE} ]] && SASLAUTHD_LDAP_SEARCH_BASE="${LDAP_SEARCH_BASE}"
+
+  if [[ "${SASLAUTHD_LDAP_SERVER}" = *'://'* ]]
+  then
+    # %% removes longest matching suffix pattern; # removes shortest matching prefix pattern
+    SASLAUTHD_LDAP_PROTO="${SASLAUTHD_LDAP_SERVER%%://*}://"
+    SASLAUTHD_LDAP_SERVER="${SASLAUTHD_LDAP_SERVER#*://}"
+  elif [[ -z ${SASLAUTHD_LDAP_SSL} ]] || [[ ${SASLAUTHD_LDAP_SSL} -eq 0 ]]
```
wernerfred, line 633: Why is `SASLAUTH_LDAP_SSL` still used here? Thought you tried to eliminate it?

moqmar: It's still there for better backwards compatibility - shall I add a warning there or shall I remove it completely?

test/mail_with_ldap.bats
moqmar: `SASLAUTHD_LDAP_SSL=0` together with the protocol being specified would previously break, with this I basically check if it's ignored in favor of the protocol in `SASLAUTHD_LDAP_SERVER`.

target/scripts/startup/setup-stack.sh
wernerfred: Imho as we are about to release v10 - a major release - we warn about breaking changes and that people have to adapt their setup accordingly. As this is such a change i would vote for removing it completely as otherwise we will have the problem at a later point in time.

cc @aendeavor

aendeavor: > As this is such a change i would vote for removing it completely as otherwise we will have the problem at a later point in time.

I agree. We can safely remove it.

test/mail_with_ldap.bats
polarathene: If that's being kept there should be a comment about it for context when another contributor runs across it.

Seems like it's being asked to drop from `setup-stack.sh` though, so perhaps has no relevance here anymore either?

aendeavor: This should be removed as well, see my comment below :)

wernerfred (approved):
LGTM 👍🏻 

Anything else needed from your side @moqmar?

If not @docker-mailserver/maintainers the next one to approve pls merge this so we can start feature freeze and testing for `v10.0.0`

aendeavor:
> LGTM 👍🏻
> 
> Anything else needed from your side @moqmar?
> 
> If not @docker-mailserver/maintainers the next one to approve pls merge this so we can start feature freeze and testing for `v10.0.0`

I have opened #1996 which does not change anything else than the layer count. After this PR is merged, we can begin with the feature freeze. We should merge #1996 before releasing v10 since its a safe improvement, see the PR description.

aendeavor (approved):

wernerfred:
We currently have a failing CI on master. The merged PR check passed though. I retriggered the check, maybe wait with merge until it passes

aendeavor:
> We currently have a failing CI on master. The merged PR check passed though. I retriggered the check, maybe wait with merge until it passes

Just in time, I was about to merge this

wernerfred:
Tests passed. Now build and publish is in Progress. You can merge imho

aendeavor:
> Tests passed. Now build and publish is in Progress. You can merge imho.
> Seems Like the First Run had a Timing/CI Environment issue

I will merge this when CI has fully passed :D

Issue #1982, opened 2021-05-18
setup.sh new requirement? ssl_dh: Can't open file /etc/dovecot/dh.pem: No such file or directory


NorseGaud:
# Subject

Hello! I want to preface this by saying that I'm not using the default dovecot version with the docker-mailserver. I've upgraded to the latest ce-2.3 so that I can use a health-check endpoint to rely on AWS's Route53 failover DNS:

```
May 18 03:11:16 ip-172-32-77-44 postfix/master[909]: daemon started -- version 3.4.14, configuration /etc/postfix
May 18 03:11:21 ip-172-32-77-44 dovecot: master: Dovecot v2.3.14 (cee3cbc0d) starting up for imap, lmtp
```

This works in every way that I'm aware of, except, when I try to add a user:

```
[ec2-user@ip-172-31-3-247 ~]$ ./setup.sh email add mike@DOMAINAMEHERE temp1234
+ SCRIPT=setup.sh
+ set -euEo pipefail
+ trap '__log_err "${FUNCNAME[0]:-?}" "${BASH_COMMAND:-?}" "${LINENO:-?}" "${?:-?}"' ERR
+ trap '_unset_vars || :' EXIT
++ pwd
+ CDIR=/home/ec2-user
+ _get_current_directory
++ readlink -f ./setup.sh
+ dirname /home/ec2-user/setup.sh
+++ readlink -f ./setup.sh
++ dirname /home/ec2-user/setup.sh
+ CDIR=/home/ec2-user
+ CRI=
+ INFO=
+ IMAGE_NAME=
+ CONTAINER_NAME=
+ DEFAULT_CONFIG_PATH=/home/ec2-user/config
+ USE_CONTAINER=false
+ WISHED_CONFIG_PATH=
+ CONFIG_PATH=
+ VOLUME=
+ USE_TTY=
+ USING_SELINUX=
+ _main email add mike@DOMAINAMEHERE temp1234
+ command -v docker
+ CRI=docker
++ tail -1
++ docker ps --no-trunc --format '{{.Image}};{{.Names}}' --filter label=org.opencontainers.image.title=docker-mailserver
+ INFO='docker.io/xkremail/docker-mailserver:2.3.14;deliver'
+ IMAGE_NAME=docker.io/xkremail/docker-mailserver:2.3.14
+ CONTAINER_NAME=deliver
+ [[ -z docker.io/xkremail/docker-mailserver:2.3.14 ]]
+ test -t 0
[...]

PR #1995, opened 2021-05-22
issues/1982: default to using container, if it exists + more

NorseGaud:
# Description

From the various discussions and investigations I've done (https://github.com/docker-mailserver/docker-mailserver/issues/1982, etc), it seems like using the container, if available, should be the default.

Also, if not available, and if we use the default image, it should check if the config path exists or is empty.

I've also changed the docker_image/_container commands into one.

## Type of change

## Checklist:

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -55,7 +55,6 @@ CONTAINER_NAME=
 DEFAULT_CONFIG_PATH="${DIR}/config"
 IMAGE_NAME=
 INFO=
-USE_CONTAINER=false
 USE_TTY=
 USE_SELINUX=
 VOLUME=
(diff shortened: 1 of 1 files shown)


casperklein:
Some actions like `email delete` require a running container. From a quick look I would say, it's not guaranteed with your solution. 

If no container is running, a new one is created, with missing mounts (maildata.. etc.) ?

NorseGaud:
> Some actions like `email delete` require a running container. From a quick look I would say, it's not guaranteed with your solution.
> 
> If no container is running, a new one is created, with missing mounts (maildata.. etc.) ?

I could add logic that ensures those other directories are available when using the temporary container. Good catch

casperklein:
> I could add logic that ensures those other directories are available when using the temporary container. Good catch

That's not easy, as you don't know how some does mount the directorys. Some use volumes, some bind mounts.

Beside parsing `docker-compose.yml` which I think is a really bad idea, I don't have a solution for that.

Please see also my comment [here](https://github.com/docker-mailserver/docker-mailserver/pull/1874#issuecomment-809781531).

NorseGaud:
> > I could add logic that ensures those other directories are available when using the temporary container. Good catch
> 
> That's not easy, as you don't know how some does mount the directorys. Some use volumes, some bind mounts.
> 
> Beside parsing `docker-compose.yml` which I think is a really bad idea, I don't have a solution for that.
> 
> Please see also my comment [here](https://github.com/docker-mailserver/docker-mailserver/pull/1874#issuecomment-809781531).

I've added code that I believe covers volumes or host level paths.

Glad to hear you didn't like `docker-compose.yml` too :) I almost went that route but it seemed weird to me too.

PR #1996, opened 2021-05-22
Improve Layer Count

aendeavor:
# Description

Lowers the layer count (by 18 layers) just by using `\`. Does not change anything else. Plain and simple.

## Type of change

## Checklist:


Dockerfile
```diff
@@ -10,29 +12,27 @@ ARG FAIL2BAN_GPG_PUBLIC_KEY_ID=0x683BF1BEBD0A882C
...
-ENV ONE_DIR=0
-ENV POSTGREY_AUTO_WHITELIST_CLIENTS=5
-ENV POSTGREY_DELAY=300
-ENV POSTGREY_MAX_AGE=35
-ENV POSTGREY_TEXT="Delayed by Postgrey"
-ENV SASLAUTHD_MECHANISMS=pam
-ENV SASLAUTHD_MECH_OPTIONS=""
-ENV VIRUSMAILS_DELETE_DELAY=7
-
-SHELL ["/bin/bash", "-o", "pipefail", "-c"]
```
wernerfred, line 35: Not needed anymore?

wernerfred (commented):
Oh wow, didn't know that is possible tbh 🤣 

Did a quick research and found this page: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/

> In older versions of Docker, it was important that you minimized the number of layers in your images to ensure they were performant. The following features were added to reduce this limitation:
>
> Only the instructions RUN, COPY, ADD create layers. Other instructions create temporary intermediate images, and do not increase the size of the build

So do we really have a benefit or is this just a local number of layers that is decreased without any impact to the final image anyways?

aendeavor:
> Oh wow, didn't know that is possible tbh rofl
> 
> Did a quick research and found this page: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/
> 
> > In older versions of Docker, it was important that you minimized the number of layers in your images to ensure they were performant. The following features were added to reduce this limitation:
> > Only the instructions RUN, COPY, ADD create layers. Other instructions create temporary intermediate images, and do not increase the size of the build
> 
> So do we really have a benefit or is this just a local number of layers that is decreased without any impact to the final image anyways?

I was wondering this as well when I just stumbled upon this with a fellow student. After what you've posted, I don't this _actually_ adds a performance benefit, but looking at <https://hub.docker.com/layers/mailserver/docker-mailserver/edge/images/sha256-1caa65c0f7f85428a145922181093304b20baac57ca8afbdf4d0fa1115e031aa?context=explore> for example, they do appear as a layer? 

We don't need to merge this at all, I just looked at the link above and thought: "Hey, maybe we can reduce layers here :)"

Dockerfile
aendeavor: The `ENV`s can be ommited, the shell remains necessary and is now located at the very top of the file. We need `SHELL` as we're using some Bash features during the build, e.g. `[[ ... ]]`.

wernerfred:
Then this might be more a "style" question. Imho i see no problem with all `ENV` and `LABEL` to appear as a single layer on dockerhub although there is no change in fact.

Is this possible for `ARG` as well?

aendeavor:
> Then this might be more a "style" question. Imho i see no problem with all `ENV` and `LABEL` to appear as a single layer on dockerhub although there is no change in fact.
> 
> Is this possible for `ARG` as well?

You can apparently not do this for `ARG`. I had no problem with the style as is, but my fellow said he knows using `\` to concatenate the `LABEL` and `ENV` command to be good style.

Dockerfile
wernerfred: Ah i missed that `SHELL` at the very top

wernerfred (approved):
LGTM 👍🏻 
From my perspective go ahead 🚀

aendeavor:
I will merge #1989 first when CI passes, then I will merge this PR and then announce the feature freeze in the maintainers discussion 💯

casperklein:
I think, this is just a docker hub thing. LABEL, ENV etc. do not add additional layers. They just end up in the manifest.

You can verify this for example with [dive](https://github.com/wagoodman/dive). Or take a look at how many layers are pulled with `docker pull`. There will be no difference after this merge.

I don't open the style discussion 😄 That is just personal preference, if you prefer `\` syntax. What I do for example from time to time is `grep ENV Dockerfile` to get all ENVs. That would not be possible anymore (in that easy form). No show-stopper. My preference would be to leave it as it is.

aendeavor:
> I think, this is just a docker hub thing. LABEL, ENV etc. do not add additional layers. They just end up in the manifest.
> 
> You can verify this for example with [dive](https://github.com/wagoodman/dive). Or take a look at how many layers are pulled with `docker pull`. There will be no difference after this merge.
> 
> I don't open the style discussion smile That is just personal preference, if you prefer `\` syntax. What I do for example from time to time is `grep ENV Dockerfile` to get all ENVs. That would not be possible anymore (in that easy form). No show-stopper. My preference would be to leave it as it is.

I understand. As this was started as a real improvement when it comes to layer count, but now only is a style thing. I will close this:) I actually prefer the way it is ATM over this PR, but I figured in the beginning if there was a real benefit, it would be worth it.

casperklein:
What I do, to ultimately "optimize" the layer count to 1:

```
FROM docker.io/debian:buster-slim as build
[..]
# build final image
FROM scratch
COPY --from=build / /
```

The only disadvantage of this approach is, that you lose the ability to use a base-image (debian). Now you have one image (1 layer) without dependency to an upstream image (debian).

PR #2000, opened 2021-05-24
chore(ci): Linting Improved

aendeavor:
# Description

Improves and updates the linting procedures.

1. Bumped linter versions to their latest release as of today.
2. Streamlined `test/linting/lint.sh`
3. Added new configuration file for ECLint
4. Had to update `start-mailserver` and `setup-stack.sh` as ShellCheck was complaining with v0.7.2. So I revised `SPAMASSASIN_SPAM_TO_INBOX_SET` for the better and removed the ugly sub-shell usage which was totally unnecessary.

Fixes #1997

## Type of change

- [x] Improvement (non-breaking change that does improve existing functionality)

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [x] If necessary I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

---

PS: I got the PR marked with #2000 🎉

PR #1980, opened 2021-05-18
chore(scripts): Removing flock so NFS works

NorseGaud:
**BLOCKERS TO MERGING THIS PR (respectively ordered)**
1. https://github.com/docker-mailserver/docker-mailserver/pull/2001

# Description

https://github.com/docker-mailserver/docker-mailserver/issues/1979

- Using a temp file instead of flock to support NFS mounts
- `cmp` checks could return a failure for any reason so we should check specifically for exit code 1 as that indicates that it's actually a failure due to differences

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Improvement (non-breaking change that does improve existing functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [x] If necessary I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

Issue #1979, opened 2021-05-18
[BUG] flock doesn't work with NFS


NorseGaud:
# Bug Report

## Context

Several scripts use `flock` to lock various files and "prevent race conditions
![Screen Shot 2021-05-17 at 9 34 53 PM](https://user-images.githubusercontent.com/5896030/118577065-bd3f1d80-b757-11eb-84c9-e205e7e1bf80.png)

> https://linux.die.net/man/2/flock
> Notes
> flock() does not lock files over NFS. Use fcntl(2) instead: that does work over NFS, given a sufficiently recent version of Linux and a server which supports locking.

### What is affected by this bug?

Ability to use NFS (and EFS)

### When does this occur?

./setup.sh commands that use the binaries that use flock

### How do we replicate the issue?

Mount NFS4.1 and use it to store the configs.

## Behavior

```
+ docker run --rm -v /mnt/efs-us-west-2/config:/tmp/docker-mailserver -ti {tag} delmailuser {email}
Do you want to delete the mailbox as well (removing all mails) ? [Y/n] y
flock: 200: Bad file descriptor
```

FSTAB:
`XXX:/ /mnt/efs-us-west-2 nfs4 nfsvers=4.1,rsize=1048576,wsize=1048576,hard,timeo=600,retrans=2,noresvport 0 0`

PR #2001, opened 2021-05-24
MacOS linting & testing support + docs

NorseGaud:
# Description

macOS users cannot easily contribute as linting and testing doesn't work locally. Some code is linux/GNU only and breaks when running on mac for various reasons.

A key change: Linting no longer requires you to run things on your host.

## Type of change

## Checklist:


aendeavor:
Since this builds on #2000, it does not need to incorporate the same changes again. When #2000 is merged, this PR will possibly have unnecessary merge conflicts. Please remove the changes that are already present in #2000.

Not sure about #1980, but it seems to have these very changes too if I‘m not mistaken.

NorseGaud:
@aendeavor Hmm... I can't run 2001 without the changes in 2000. I also can't run 1979 without the changes from 2001. What do you suggest?

aendeavor:
> @aendeavor Hmm... I can't run 2001 without the changes in 2000. I also can't run 1979 without the changes from 2001. What do you suggest?

I see. Then leave the PRs as is for now. When #2000 is merged, you can rebase #2001 and already have the changes this way without introducing them twice. Same for #1979 and #2001 :)

Makefile
```diff
@@ -29,7 +39,7 @@ clean:
 # remove running and stopped test containers
 	-@ [[ -d config.bak ]] && { rm -rf config ; mv config.bak config ; } || :
 	-@ [[ -d testconfig.bak ]] && { sudo rm -rf test/config ; mv testconfig.bak test/config ; } || :
-	-@ docker ps -a | grep -E "mail|ldap_for_mail|mail_overri.*" | cut -f 1-1 -d ' ' | xargs --no-run-if-empty docker rm -f
+	-@ for cn in $$(grep -hr ".*docker run.*-name" ./* | xargs | sed 's/.*docker run.*-name//g'); do docker rm -f "$$cn" || true; done &>/dev/null
```
polarathene, line 42: I'm not sure about this change. The previous content was clear that it was for containers with specific names, but I'm not proficient enough with shell scripts to decipher how this loop is getting container names to remove?

There is no `docker ps` involved, it looks like some input source is parsed for `--name` options on `docker run` commands?

Makefile
```diff
@@ -50,7 +60,8 @@ tests:
 test/%.bats: ALWAYS_RUN
 	@ ./test/bats/bin/bats $@
 
-lint: eclint hadolint shellcheck
+lint:
+	$(call docker-execute,"[[ ! -e "$(TOOLS_DIR)/hadolint" ]] && make install_linters; make hadolint shellcheck eclint")
```
polarathene, line 64: I'm also not that familiar with `make` and it's config, I assume that's why you're not able to run the `docker run` command (that's been nested into `docker-execute`) here directly?

As this wasn't being done before (previously lint commands were called directly, or their makefile entries?), have I understood this correctly that the linting is now being performed inside a docker container?

Is the built `docker-mailserver` image providing some value to do that in vs a more generic container? (as in, was the `lint build` order for `all` swapped to `build lint` just to satisfy this change?)

docs/content/contributing/issues-and-pull-requests.md
```diff
@@ -35,8 +35,7 @@ The development workflow is the following:
    2. Run `git submodule update --init --recursive`
 2. Write the code that is needed :D
 3. Add integration tests if necessary
-4. Get the linters with `make install_linters` and install `jq` with the package manager of your OS
-5. Use `make clean all` to build image locally and run tests (note that tests work on Linux **only**)
```
polarathene, line 39: I assume the `clean` task was not required to be added prior to `all` as it runs at the end of `all`? Was there any other reason this has been dropped from the instructions?

I'm assuming that `clean` was advised this way in the event that something went wrong and broke during `make`, and that prevented `clean` from running at the end? If that's possible, it'd make sense to want to have a clean state before running `make` again? Should this advice really be dropped?

Additionally while this PR adds testing support for macOS, does that warrant dropping our compatibility notice? Should it remain explicit that Windows is not supported (or that only macOS and Linux work for testing?), assuming that is the case, eg WSL isn't supported or any other approach a contributor on Windows might try?

polarathene:
The linting could all be run via container?

Testing with bats might be a bit more tricky. We can run Docker-in-Docker with `docker:dind` image (requires the `--privileged` flag), or mounting the docker daemon socket (`-v /var/run/docker.sock:/var/run/docker.sock`) which seems better since you can use the docker CLI as usual within the container to run docker commands if you were doing so on the host.

It'd seem that on linux hosts we could use [sysbox](https://github.com/nestybox/sysbox) which is a bit more secure approach for running "system" containers, that can run nested containers, however that wouldn't help with this macOS support and I think the bind mount for `docker.sock` would work well to launch sibling containers on the host to run tests? (alternatively run tests via a VM which can be managed with Vagrant or similar)

I personally just ran tests in a temporary VPS instance for less than 1 cent. Dockerizing all that could come in future, along with the refactoring of `setup.sh` as has been discussed before.

polarathene (changes requested):
My review is mostly regarding the docs, but I have a few concerns with other changes that I'd appreciate some clarity on.

docs/content/contributing/issues-and-pull-requests.md
wernerfred: I've run this multiple times successfully on WSL - so the question is if we should keep the focus on Linux (only!) and just give some hints on what to do on other platforms, what might work?

polarathene: If there are no known issues, it's fine then. Can add back the disclaimer if someone finds a problem and we're unable to support that for some reason.

casperklein: @wernerfred Correct me, if I am wrong. WSL2 is basically a Ubuntu VM. So the tests still run on linux, even if the host is Windows.

wernerfred: Just wanted to respond to:
> eg WSL isn't supported or any other approach a contributor on Windows might try
imho we should stay with linux only BUT add hints on what do to on other plattforms. Here would be the hint to use WSL on windows as in fact then it runs on linux.

Makefile
NorseGaud:
```
    ~/docker-mailserver  on   macos-lint-and-test-support  for cn in $(grep -hr ".*docker run.*-name" ./* | xargs | sed 's/.*docker run.*-name//g'); do echo "$cn" || true; done 
mail_default_dhparams_one_dir
mail_default_dhparams_not_one_dir
mail_postfix_inet_default
mail_postfix_inet_all
mail_postfix_inet_ipv4
mail_postfix_inet_ipv6
mail_lmtp_ip
mail_manual_dhparams_not_one_dir
mail_pop3
mail_smtponly
mail_undef_spam_subject
mail_privacy
mail_fetchmail
mail_spam_bounced_defined
mail_spam_bounced_undefined
mail_fail2ban
fail-auth-mailer
mail_manual_dhparams_one_dir
mail_smtponly_second_network_sender
mail_fetchmail_parallel
mail_disabled_clamav_spamassassin
mail_spam_moved_junk
mail_spam_moved_new
mail_srs_domainname
mail_domainname
mail_with_default_relay
mail_postscreen
mail_postscreen_sender
mail_with_imap
mail_lets_domain
mail_lets_hostname
mail_lets_acme_json
ldap_for_mail
mail_with_ldap
mail
mail_with_relays
mail_special_use_folders
mail_with_mdbox_format
tls_test_cipherlists
mail_with_postgrey
mail_with_sdbox_format
mail_override_hostname
mail_no_quotas
mail_manual_ssl
```

```
    ~/docker-mailserver  on   macos-lint-and-test-support  grep -rl ".*docker run.*-name" ./*               2 ✘  at 21:50:33  
./Makefile
./test/mail_dhparams_default.bats
./test/mail_postfix_inet.bats
./test/mail_lmtp_ip.bats
./test/mail_dhparams_manual_not_one_dir.bats
./test/mail_pop3.bats
./test/mail_smtponly.bats
./test/mail_undef_spam_subject.bats
./test/mail_privacy.bats
./test/mail_fetchmail.bats
./test/mail_spam_bounced.bats
./test/mail_fail2ban.bats
./test/mail_dhparams_manual_one_dir.bats
./test/permit_docker.bats
./test/mail_fetchmail_parallel.bats
./test/mail_disabled_clamav_spamassassin.bats
./test/mail_spam_junk_folder.bats
./test/mail_srs_domainname.bats
./test/default_relay_host.bats
./test/mail_postscreen.bats
./test/mail_with_imap.bats
./test/mail_ssl_letsencrypt.bats
./test/mail_with_ldap.bats
./test/tests.bats
./test/mail_with_relays.bats
./test/mail_special_use_folders.bats
./test/mail_with_mdbox.bats
./test/security_tls_cipherlists.bats
./test/mail_with_postgrey.bats
./test/mail_with_sdbox.bats
./test/mail_override_hostname.bats
./test/mail_quotas_disabled.bats
./test/mail_ssl_manual.bats
```

This loop is futureproofing and eliminating the need for a human to remember to add the specific name/pattern to the old grep in the Makefile.

Makefile
NorseGaud: The lint.sh script used by eclint, hadolint, and shellcheck makefile functions do not work on macOS. It (macOS) does not support `realpath` and `readlink` flags that are only available on the GNU versions of those commands.

Secondly, we could use a generic container for lint testing, but I already build the Dockerfile container/tag for testing anyway. The switch of makefile functions `build` and `lint` so that build comes first solves this.

docs/content/contributing/issues-and-pull-requests.md
NorseGaud: Good catch @polarathene , I've updated the docs to include clean again.

docs/content/contributing/tests.md
```diff
@@ -2,5 +2,15 @@
...
+3. Install jq
+
+    > MacOS Specific (needed for tests):
+      ```bash
+      brew install coreutils
+      # bash >= 4.0 for associative arrays
+      brew install bash
+      ```
+
+4. Execute `make`
```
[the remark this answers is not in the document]
NorseGaud: Great point! I always forget this is possible 😅

Makefile
NorseGaud: Oh, additionally, I defined the docker-execute to make it easier to run other things inside of the built `mailserver-testing:ci` and keep things standardized. I can remove it if you're not liking it.

Makefile
polarathene: That needs to be explained via comment, I don't know how many are going to know what to expect from that line, whereas the previous one was far more clear with what was going on.

While neat, by being implicit like this it's far easier to gloss over that list and assume it's working correctly... but not all tests are adhering to such a convention to match on, eg [`open_dkim.bats`](https://github.com/docker-mailserver/docker-mailserver/blob/master/test/open_dkim.bats), [`permit_docker.bats`](https://github.com/docker-mailserver/docker-mailserver/blob/master/test/permit_docker.bats#L12), [this one doesn't even assign a name](https://github.com/docker-mailserver/docker-mailserver/blob/master/test/mail_with_postgrey_disabled_by_default.bats), etc.

While probably not entirely foolproof either, I believe we have a common image name at least for instances of the docker-mailserver image. There is support for customizing the image used, but querying containers using that image is probably better and more clear of intent?

I'm personally :-1: for this due to the above concern (alternatively, with the comment note included, all tests could be reviewed to conform to the convention pattern being matched and some documentation for test contributions about this convention being required so no future test contributions fail to match). I realize that the earlier approach also missed these examples I linked to, but at least it was more evident at a glance.

NorseGaud: Ok, I'm fine reverting it. It was just an idea I wanted to try out :)

Makefile
polarathene: > For example, it does not support the `readlink -f` flag that is only available on the GNU versions of those commands.

:+1: I'm in favor of using docker for running linting :)

> Why clutter my env and use disk space? The switch of makefile functions `build` and `lint` order, so that build comes first, solves this.

You are favoring disk space savings to avoid additional temporary images for running linting over project maintenance? As someone who's primary system is often lacking disk space I can understand wanting to (and often why I tend to prefer small images like Alpine), but at the same time.. I'd rather not introduce new ENV + ARG to the Dockerfile just to support adding a build variant with the lint tools to run.

Each of these lint tools do have an official Docker image. Those can be used for linting in general for various projects which imo is preferable vs each project DIY including their own with approaches like we have been. We don't have to include the new vars for kernel and arch either if that were the case? (since DockerHub takes care of that when pulling images published with different arch implicitly)

I'm not blocking this, it doesn't bother me that much and can always be addressed in future if desired. FWIW, all three official docker images is less than 30MB on disk, which makes the argument in favor difficult to justify imo. Single docker-compose can run all 3 in parallel and does not require keeping versions updated in both a Makefile and our Github Actions workflow.

---

> The switch of makefile functions `build` and `lint` order, so that build comes first, solves this.

I don't see much harm in this, however I still think it's better to pass lints before perfoming a build. Especially within CI where you can have quotas for bandwidth and build/CI minutes (I don't think that's a concern for this project), where failing early is preferable.

If we do adopt linting with official docker images instead, there's no need to prioritizing building first, and this order can be reverted.

---

> I defined the docker-execute to make it easier to run other things inside of the built `mailserver-testing:ci` while keeping it standardized/dry. I can remove it if you're not liking it.

I don't personally see a need or value in it when it's only used once and would be easier to grok without the indirection. I'm not opposed to it, thanks for the clarification.

Makefile
polarathene: That's all good, thank you. At least it drew attention that we're not covering all the containers due to lack of consistent convention :sweat_smile:

Makefile
NorseGaud: Good points. I agree too. I like the direction of running the lint docker containers and moving lint back before build. I'll try making some changes!

docs/content/contributing/tests.md
```diff
@@ -2,5 +2,15 @@
...
+1. Install docker
+2. Execute `git submodule update --init --recursive`
+3. Install jq
+
+    !!! info "MacOS Specific (needed for tests)"
+        ```bash
+        brew install coreutils
+        # bash >= 4.0 for associative arrays
+        brew install bash
+        ```
```
polarathene, line 14: Formatting nit-pick, blank new line after admonition declaration (there's some cases where the parser otherwise breaks which breaks rendering, so it's best to try keep this convention for admonitions).

Feel free to use the commit suggestion button below this time :stuck_out_tongue: 

```suggestion
    !!! info "MacOS Specific (needed for tests)"

        ```bash
        brew install coreutils
        # bash >= 4.0 for associative arrays
        brew install bash
        ```
```

polarathene (changes requested):
Looks like we just have the switch from the current docker image approach for linting to docker-compose using official lint images directly and we're good! :tada: 

For the other maintainers, note that the improvement to image name collection was reverted. It did [reveal that our current filter likewise isn't identifying all test images by name assigned](https://github.com/docker-mailserver/docker-mailserver/pull/2001#discussion_r645225913), mostly because there is no established/documented convention enforced which should be addressed at some point.

Makefile
NorseGaud: Ok! First I want to thank you for such thoughtful feedback and for taking the time to help me out. As you can tell I'm getting used to how things are done in the repo and really look forward to continued contributions and improvements!

Second, I've made what I think is the best way to lint using containers. There are no legit eclint or shellcheck dockerhub tags that have bash available, so I'm just using the same tag we build FROM. I also moved a lot of the logic into lint.sh and out of the Makefile. Not sure if that's ok or not. 

Thanks again!

setup.sh
```diff
@@ -34,20 +34,7 @@ the version / tag of docker-mailserver. Please read the
...
-  elif realpath -e -L "${0}" &>/dev/null
-  then
-    DIR="$(realpath -e -L "${0}")"
-    DIR="${DIR%/setup.sh}"
-  fi
-}
-
-DIR="$(pwd)"
-_get_absolute_script_directory
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
```
aendeavor, line 37: Any reason not to use `dirname “${BASH_SOURCE[0]}“` directly?

Makefile
```diff
@@ -52,21 +51,15 @@ test/%.bats: ALWAYS_RUN
 
 lint: eclint hadolint shellcheck
 
+HADOLINT_VERSION="2.4.1"
```
aendeavor, line 54: I prefer to have the variables on the top of the `Makefile`, but this may be fine. However, why do you quote the value(s)? This way, `""` becomes part of the value which is not desired in this case :)

setup.sh
```diff
@@ -225,7 +212,7 @@ function _docker_image
   if ${USE_CONTAINER}
   then
     # reuse existing container specified on command line
-    ${CRI} exec "${USE_TTY}" "${CONTAINER_NAME}" "${@}"
+    ${CRI} exec "${USE_TTY}" "${CONTAINER_NAME}" "$@"
```
aendeavor, line 215: Reason for changing `${@}` to `$@`? As CONTRIBUTING suggests, variables _always_ use braces. Is this related to macOS in some way?

If this is just a style preference, please revert this change as it does not adhere to our style guidelines.

aendeavor (changes requested):
Some small parts that still need clarifying IMHO. But the majority looks good!

Makefile
polarathene: > There are no legit eclint or shellcheck dockerhub tags that have bash available (required by the lint.sh for some reason), so I'm just using the same tag we build FROM.

Well, that's a little better I guess.

> I also moved a lot of the logic into lint.sh and out of the Makefile. Not sure if that's ok or not.

A major reason of wanting to push for the official images was to not require us to need the additional install code in the project :sweat_smile:

Sorry if that wasn't clear, I don't mean to waste your time or disrespect your contributions by dismissing the effort here. I'm just not sure that it's the right approach.

You're welcome to wait on feedback from the other maintainers btw, my maintenance role is minor in this area of the project. The others may have different opinions.

---

For better visibility, I'll share what I had in mind on the main discussion thread.

polarathene:
Follow up comment to [my earlier review comment](https://github.com/docker-mailserver/docker-mailserver/pull/2001#discussion_r645388870). Below shows my preferred approach with using docker images for linting, but they need to be further adapted into `lint.sh` and verified as working the same.

---

I have slightly improved/modified EC config below, and enabled `--check-sources` for shellcheck config (_which raises a single SC2153 warning that wasn't picked up_). I also believe I resolved the `readlink -f` issue that was being had on macOS, I am not entirely sure why the more complicated workaround was in place, the option for `--source-path` has been available for about 2 years or so.

Rough shell commands to run each official docker image with config like in `lint.sh`. If they're successful with no issues, they'll be blank in output when exiting (_should function just like error handling for existing command usage_). Each one seemed to run fine and as expected, although I've only run them individually manually, no `lint.sh`.

```
HADOLINT_VERSION=2.4.1
EC_VERSION=2.3.5
SC_VERSION=0.7.2

docker run --rm --tty \
  --volume "${PWD}:/ci:ro" \
  --workdir "/ci" \
  "hadolint/hadolint${HADOLINT_VERSION}-alpine" hadolint --config "test/linting/.hadolint.yaml" Dockerfile

docker run --rm --tty \
  --volume "${PWD}:/ci:ro" \
  --workdir "/ci" \
  "mstruebing/editorconfig-checker:${EC_VERSION}" ec -config "test/linting/.ecrc.json"

# File paths for shellcheck:
local F_SH="$(find . -type f -iname '*.sh' \
  -not -path './test/bats/*' \
  -not -path './test/test_helper/*' \
  -not -path './target/docker-configomat/*' \
)"
local F_BIN="$(find 'target/bin' -type f -executable)"
local F_BATS="$(find 'test/' -maxdepth 1 -type f -iname '*.bats')"

# This command is a bit easier to grok as multi-line. There is a `.shellcheckrc` file, but it's only supports half of the options below, thus kept as CLI:
local CMD_SHELLCHECK=(shellcheck --external-sources --check-sources \
  --severity=style \
  --color=auto \
  --wiki-link-count=50 \
  --enable=all \
  --exclude=SC2154 \
  --source-path=SCRIPTDIR \
  "${F_SH} ${F_BIN} ${F_BATS}" \
)

docker run --rm --tty \
  --volume "${PWD}:/ci:ro" \
  --workdir "/ci" \
  "koalaman/shellcheck-alpine:v${SC_VERSION}" ${CMD_SHELLCHECK[@]}
```

**`test/linting/.ecrc.json`:** (_modified to remove version as it's no longer required and the docker image was not compiled with it correctly, also moves the exclude config into the actual config file_)

```json
{
  "Verbose": false,
  "Debug": false,
  "IgnoreDefaults": false,
  "SpacesAftertabs": true,
  "NoColor": false,
  "Exclude": [
    "^test/",
    "\\.git.*",
    "\\.bats$",
    "\\.cf$",
    "\\.conf$",
    "\\.init$",
    "\\.md$"
  ],
  "AllowedContentTypes": [],
  "PassedFiles": [],
  "Disable": {
    "EndOfLine": false,
    "Indentation": false,
    "InsertFinalNewline": false,
    "TrimTrailingWhitespace": false,
    "IndentSize": false,
    "MaxLineLength": false
  }
}
```

aendeavor:
> @aendeavor added it in Oct 2020 for --external-sources, perhaps was just missing the --source-path=SCRIPTDIR config addition?

Yes, I missed this flag. When you get used to what you did you do not look for new flags :) Nice addition!

@polarathene about the `CMD_SHELLCHECK` array. Why do you used `\` at the end if it is an array anyways? I'd either use a string with `\` or (preferably) the array as so:

``` BASH
local CMD_SHELLCHECK=(
  shellcheck --external-sources --check-sources
  --severity=style
  --color=auto
  --wiki-link-count=50
  --enable=all
  --exclude=SC2154
  --source-path=SCRIPTDIR
  "${F_SH} ${F_BIN} ${F_BATS}"
)
```

and the rest can stay the same as you alrady use `${CMD_SHELLCHECK[@]}"` to get all elements. As of now, `${CMD_SHELLCHECK[@]}"` yields the first _and only_ field of the array because there is only one. (Correct me if I'm wrong here)

setup.sh
NorseGaud: Shellcheck was puking allover itself. That's not the case anymore since your fixes so I'll revert it back

Makefile
NorseGaud: You're right, sorry, I forgot to finalize this last night

setup.sh
NorseGaud: I'll add back the old code, good catch 👍

aendeavor (dismissed):

NorseGaud:
> Follow up comment to [my earlier review comment](https://github.com/docker-mailserver/docker-mailserver/pull/2001#discussion_r645388870). Below shows my preferred approach with using docker images for linting, but they need to be further adapted into `lint.sh` and verified as working the same.
> 
> I have slightly improved/modified EC config below, and enabled `--check-sources` for shellcheck config (_which raises a single SC2153 warning that wasn't picked up_). I also believe I resolved the `readlink -f` issue that was being had on macOS, I am not entirely sure why the more complicated workaround was in place, the option for `--source-path` has been available for about 2 years or so. @aendeavor [added it in Oct 2020 for `--external-sources`](https://github.com/docker-mailserver/docker-mailserver/pull/1654/commits/e1b34a0a347cf129de2fe277cd29ea7eba892fcd), perhaps was just missing the `--source-path=SCRIPTDIR` config addition?
> 
> Rough shell commands to run each official docker image with config like in `lint.sh`. If they're successful with no issues, they'll be blank in output when exiting (_should function just like error handling for existing command usage_). Each one seemed to run fine and as expected, although I've only run them individually manually, no `lint.sh`.
> 
> ```shell
> HADOLINT_VERSION=2.4.1
> EC_VERSION=2.3.5
> SC_VERSION=0.7.2
> 
> docker run --rm --tty \
>   --volume "${PWD}:/ci:ro" \
>   --workdir "/ci" \
>   "hadolint/hadolint${HADOLINT_VERSION}-alpine" hadolint --config "test/linting/.hadolint.yaml" Dockerfile
> 
> docker run --rm --tty \
>   --volume "${PWD}:/ci:ro" \
>   --workdir "/ci" \
>   "mstruebing/editorconfig-checker:${EC_VERSION}" ec -config "test/linting/.ecrc.json"
> 
> # File paths for shellcheck:
> local F_SH="$(find . -type f -iname '*.sh' \
>   -not -path './test/bats/*' \
>   -not -path './test/test_helper/*' \
>   -not -path './target/docker-configomat/*' \
> )"
> local F_BIN="$(find 'target/bin' -type f -executable)"
> local F_BATS="$(find 'test/' -maxdepth 1 -type f -iname '*.bats')"
> 
> # This command is a bit easier to grok as multi-line. There is a `.shellcheckrc` file, but it's only supports half of the options below, thus kept as CLI:
> local CMD_SHELLCHECK=(shellcheck --external-sources --check-sources \
>   --severity=style \
>   --color=auto \
>   --wiki-link-count=50 \
>   --enable=all \
>   --exclude=SC2154 \
>   --source-path=SCRIPTDIR \
>   "${F_SH} ${F_BIN} ${F_BATS}" \
> )
> 
> docker run --rm --tty \
>   --volume "${PWD}:/ci:ro" \
>   --workdir "/ci" \
>   "koalaman/shellcheck-alpine:v${SC_VERSION}" ${CMD_SHELLCHECK[@]}
> ```
> 
> **`test/linting/.ecrc.json`:** (_modified to remove version as it's no longer required and the docker image was not compiled with it correctly, also moves the exclude config into the actual config file_)
> 
> ```json
> {
>   "Verbose": false,
>   "Debug": false,
>   "IgnoreDefaults": false,
>   "SpacesAftertabs": true,
>   "NoColor": false,
>   "Exclude": [
>     "^test/",
>     "\\.git.*",
>     "\\.bats$",
>     "\\.cf$",
>     "\\.conf$",
>     "\\.init$",
>     "\\.md$"
>   ],
>   "AllowedContentTypes": [],
>   "PassedFiles": [],
>   "Disable": {
>     "EndOfLine": false,
>     "Indentation": false,
>     "InsertFinalNewline": false,
>     "TrimTrailingWhitespace": false,
>     "IndentSize": false,
>     "MaxLineLength": false
>   }
> }
> ```

I'd like to try implementing this into the lint.sh. My only concern is that the amount of files in the find results seems to be a problem:

```
docker: Error response from daemon: OCI runtime create failed: container_linux.go:367: starting container process caused: exec: "shellcheck --external-sources --check-sourced --severity=style --color=auto --wiki-link-count=50 --enable=all --exclude=SC2154 --source-path=SCRIPTDIR ./setup.sh ./test/linting/lint.sh ./test/config/user-patches/user-patches.sh ./test/duplicate_configs/test_helper.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_postfix_inet.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_spam_bounced_defined/user-patches/user-patches.sh ./test/duplicate_configs/mail_key_size_4096/user-patches/user-patches.sh ./test/duplicate_configs/mail_lmtp_ip.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_srs_domainname/user-patches/user-patches.sh ./test/duplicate_configs/mail_dhparams_manual_not_one_dir.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_lets_domain/user-patches/user-patches.sh ./test/duplicate_configs/mail_domainname/user-patches/user-patches.sh ./test/duplicate_configs/mail_pop3.bats/user-patches/user-patches.sh ./test/duplicate_configs/open-dkim/user-patches/user-patches.sh ./test/duplicate_configs/mail_default_dhparams_both_one_dir/user-patches/user-patches.sh ./test/duplicate_configs/test_-24-7bTEST-5fFILE-7dgenerator_creates_keys-2c_tables_and_TrustedHosts_without_postfix-2dvirtual-2ecf/user-patches/user-patches.sh ./test/duplicate_configs/mail/user-patches.sh ./test/duplicate_configs/mail_smtponly.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_undef_spam_subject.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_privacy.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_fetchmail.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_key_size_1024/user-patches/user-patches.sh ./test/duplicate_configs/mail_fail2ban.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_dhparams_manual_one_dir.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_with_postgrey_disabled_by_default.bats/user-patches/user-patches.sh ./test/duplicate_configs/test_-24-7bTEST-5fFILE-7dgenerator_creates_keys-2c_tables_and_TrustedHosts_using_manual_provided_selector_name/user-patches/user-patches.sh ./test/duplicate_configs/mail_fetchmail_parallel.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_dkim_generator_creates_keys_tables_TrustedHosts/user-patches/user-patches.sh ./test/duplicate_configs/mail_disabled_clamav_spamassassin.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_postscreen.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_spam_moved_new/user-patches/user-patches.sh ./test/duplicate_configs/mail_with_imap.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_with_ldap.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_spam_moved_junk/user-patches/user-patches.sh ./test/duplicate_configs/mail_lets_acme_json/user-patches/user-patches.sh ./test/duplicate_configs/mail_lets_hostname/user-patches/user-patches.sh ./test/duplicate_configs/mail_key_size_2048/user-patches/user-patches.sh ./test/duplicate_configs/mail_smtponly_second_network_sender/user-patches/user-patches.sh ./test/duplicate_configs/mail_undef_spam_subject_2/user-patches/user-patches.sh ./test/duplicate_configs/mail_special_use_folders.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_with_mdbox.bats/user-patches/user-patches.sh ./test/duplicate_configs/test_-24-7bTEST-5fFILE-7dgenerator_creates_keys-2c_tables_and_TrustedHosts_using_manual_provided_domain_name/user-patches/user-patches.sh ./test/duplicate_configs/security_tls_cipherlists.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_with_postgrey.bats/user-patches/user-patches.sh ./test/duplicate_configs/open_dkim.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_with_sdbox.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_override_hostname.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_default_key_size/user-patches/user-patches.sh ./test/duplicate_configs/mail_quotas_disabled.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_smtponly_second_network/user-patches/user-patches.sh ./test/duplicate_configs/mail_ssl_manual.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_default_dhparams_both_not_one_dir/user-patches/user-patches.sh ./testconfig.bak/user-patches/user-patches.sh ./target/scripts/fail2ban-wrapper.sh ./target/scripts/update-check.sh ./target/scripts/start-mailserver.sh ./target/scripts/helper-functions.sh ./target/scripts/quota-warning.sh ./target/scripts/check-for-changes.sh ./target/scripts/startup/misc-stack.sh ./target/scripts/startup/fixes-stack.sh ./target/scripts/startup/check-stack.sh ./target/scripts/startup/daemons-stack.sh ./target/scripts/startup/setup-stack.sh ./target/scripts/postsrsd-wrapper.sh ./target/scripts/postfix-wrapper.sh ./.github/workflows/scripts/docs/update-versions-json.sh ./.github/workflows/scripts/docs/build-docs.sh": stat shellcheck --external-sources --check-sourced --severity=style --color=auto --wiki-link-count=50 --enable=all --exclude=SC2154 --source-path=SCRIPTDIR ./setup.sh ./test/linting/lint.sh ./test/config/user-patches/user-patches.sh ./test/duplicate_configs/test_helper.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_postfix_inet.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_spam_bounced_defined/user-patches/user-patches.sh ./test/duplicate_configs/mail_key_size_4096/user-patches/user-patches.sh ./test/duplicate_configs/mail_lmtp_ip.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_srs_domainname/user-patches/user-patches.sh ./test/duplicate_configs/mail_dhparams_manual_not_one_dir.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_lets_domain/user-patches/user-patches.sh ./test/duplicate_configs/mail_domainname/user-patches/user-patches.sh ./test/duplicate_configs/mail_pop3.bats/user-patches/user-patches.sh ./test/duplicate_configs/open-dkim/user-patches/user-patches.sh ./test/duplicate_configs/mail_default_dhparams_both_one_dir/user-patches/user-patches.sh ./test/duplicate_configs/test_-24-7bTEST-5fFILE-7dgenerator_creates_keys-2c_tables_and_TrustedHosts_without_postfix-2dvirtual-2ecf/user-patches/user-patches.sh ./test/duplicate_configs/mail/user-patches.sh ./test/duplicate_configs/mail_smtponly.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_undef_spam_subject.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_privacy.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_fetchmail.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_key_size_1024/user-patches/user-patches.sh ./test/duplicate_configs/mail_fail2ban.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_dhparams_manual_one_dir.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_with_postgrey_disabled_by_default.bats/user-patches/user-patches.sh ./test/duplicate_configs/test_-24-7bTEST-5fFILE-7dgenerator_creates_keys-2c_tables_and_TrustedHosts_using_manual_provided_selector_name/user-patches/user-patches.sh ./test/duplicate_configs/mail_fetchmail_parallel.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_dkim_generator_creates_keys_tables_TrustedHosts/user-patches/user-patches.sh ./test/duplicate_configs/mail_disabled_clamav_spamassassin.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_postscreen.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_spam_moved_new/user-patches/user-patches.sh ./test/duplicate_configs/mail_with_imap.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_with_ldap.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_spam_moved_junk/user-patches/user-patches.sh ./test/duplicate_configs/mail_lets_acme_json/user-patches/user-patches.sh ./test/duplicate_configs/mail_lets_hostname/user-patches/user-patches.sh ./test/duplicate_configs/mail_key_size_2048/user-patches/user-patches.sh ./test/duplicate_configs/mail_smtponly_second_network_sender/user-patches/user-patches.sh ./test/duplicate_configs/mail_undef_spam_subject_2/user-patches/user-patches.sh ./test/duplicate_configs/mail_special_use_folders.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_with_mdbox.bats/user-patches/user-patches.sh ./test/duplicate_configs/test_-24-7bTEST-5fFILE-7dgenerator_creates_keys-2c_tables_and_TrustedHosts_using_manual_provided_domain_name/user-patches/user-patches.sh ./test/duplicate_configs/security_tls_cipherlists.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_with_postgrey.bats/user-patches/user-patches.sh ./test/duplicate_configs/open_dkim.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_with_sdbox.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_override_hostname.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_default_key_size/user-patches/user-patches.sh ./test/duplicate_configs/mail_quotas_disabled.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_smtponly_second_network/user-patches/user-patches.sh ./test/duplicate_configs/mail_ssl_manual.bats/user-patches/user-patches.sh ./test/duplicate_configs/mail_default_dhparams_both_not_one_dir/user-patches/user-patches.sh ./testconfig.bak/user-patches/user-patches.sh ./target/scripts/fail2ban-wrapper.sh ./target/scripts/update-check.sh ./target/scripts/start-mailserver.sh ./target/scripts/helper-functions.sh ./target/scripts/quota-warning.sh ./target/scripts/check-for-changes.sh ./target/scripts/startup/misc-stack.sh ./target/scripts/startup/fixes-stack.sh ./target/scripts/startup/check-stack.sh ./target/scripts/startup/daemons-stack.sh ./target/scripts/startup/setup-stack.sh ./target/scripts/postsrsd-wrapper.sh ./target/scripts/postfix-wrapper.sh ./.github/workflows/scripts/docs/update-versions-json.sh ./.github/workflows/scripts/docs/build-docs.sh: file name too long: unknown.

––– SHELLCHECK
  – type    = FAILURE
  – message = errors encountered
```

I'll keep digging into a way to fix this and follow up soon.

polarathene:
> Why do you use `\` at the end if it is an array anyways? I'd either use a string with `\` or (preferably) the array as such

> As of now, `${CMD_SHELLCHECK[@]}"` yields the first and only field of the array because there is only one. (Correct me if I'm wrong here)

I'm not sure if the `"` at the end there meant you were thinking of it wrapped in quotes, when I tried that it didn't work properly (_although I think shellcheck would complain about it if it linted that, I remember something similar with a prior contribution with this array syntax_).

As stated in my comment with the example, it wasn't properly tested as is, I had just been running the commands directly in a shell (_without the line splitting_).. so it's quite possible that I made a mistake there. I thought ending lines with `\` was required for multi-line values to collapse back into a single line that the shell accepts. I'm still for the most part a shell script noob :sweat_smile: I don't write scripts frequently enough to know better :weary: 

---

> FYI: The shellcheck flag is called `--check-sourced`, not `--check-sources` :)

Right you are! Sorry about that, I was naughty and only tested with the short flag `-a` in the terminal, then chose to use the long flag in the example as I appreciate the verbosity for maintenance.

setup.sh
```diff
@@ -34,7 +34,20 @@ the version / tag of docker-mailserver. Please read the
...
+    DIR="$(dirname "$(readlink -f "${0}")")"
+  elif realpath -e -L "${0}" &>/dev/null
+  then
+    DIR="$(realpath -e -L "${0}")"
+    DIR="${DIR%/setup.sh}"
+  fi
+}
+
+DIR="$(pwd)"
+_get_absolute_script_directory
```
polarathene: Does reverting this now mean `readlink -f` is not a compatibility issue on macOS due to installing some other dependencies instead?

NorseGaud: Yep, the brew install coreutils provides the GNU readlink on the machine. I need to clean my local env tomorrow when I finish the changes for this PR and double check. I want to make sure my local env is not modified somehow to make this work.

aendeavor:
> > Why do you use `\` at the end if it is an array anyways? I'd either use a string with `\` or (preferably) the array as such
> 
> > As of now, `${CMD_SHELLCHECK[@]}"` yields the first and only field of the array because there is only one. (Correct me if I'm wrong here)
> 
> I'm not sure if the `"` at the end there meant you were thinking of it wrapped in quotes, when I tried that it didn't work properly (_although I think shellcheck would complain about it if it linted that, I remember something similar with a prior contribution with this array syntax_).

The `"` is an error from my side. It should have read `${CMD_SHELLCHECK[@]}` without the hyphen (a typo). This would otherwise be incorrect syntax :)

NorseGaud:
I meant to have this wrapped up yesterday but got occupied with some life stuff :) Sorry for the delay! I've finished the changes necessary to do linting in docker containers and get tests to pass on MacOS. Ready for review!

polarathene (changes requested):
I've cloned the PR branch and run it on a fresh Ubuntu 21.04 install.

It errors on the `shellcheck` lint due to the `find` command usage of `-perm +111` being invalid (see [review comment](https://github.com/docker-mailserver/docker-mailserver/pull/2001#discussion_r646195402) for more details).

Some of my review requests input from other maintainers.

test/linting/lint.sh
```diff
@@ -115,77 +90,39 @@ function _hadolint
...
-      ERR=1
-    fi
-  done < <(find test/ -maxdepth 1 -type f -iname "*.bats")
-
-  if [[ ${ERR} -eq 1 ]]
+  # shellcheck disable=SC2086
+  if docker run --rm --tty \
+      --volume "${REPO_ROOT}:/ci:ro" \
+      --workdir "/ci" \
+      "koalaman/shellcheck-alpine:v${SHELLCHECK_VERSION}" "${CMD_SHELLCHECK[@]}" ${F_SH} ${F_BIN} ${F_BATS}
```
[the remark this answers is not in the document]
NorseGaud: > Presumably because you've wrapped that in quotes and when testing it wasn't working as intended otherwise?

yep, I'll move it back

test/linting/lint.sh
```diff
@@ -115,77 +90,39 @@ function _hadolint
...
-  fi
+  
+  # File paths for shellcheck:
+  F_SH="$(find . -type f -iname '*.sh' \
+    -not -path './test/bats/*' \
+    -not -path './test/test_helper/*' \
+    -not -path './target/docker-configomat/*' -print0 \
+    | xargs -0
+  )"
+  F_BIN="$(find 'target/bin' -perm +111 -type f -or -type l)"
```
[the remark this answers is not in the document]
NorseGaud: > When running this locally on the project I get an error "find: invalid mode ‘+111’".

Hm, that's odd... I'll check again.

---


Without the xargs/single line, it throws 

```
./setup.sh
./test/linting/lint.sh
./test/config/user-patches/user-patches.sh
./test/duplicate_configs/mail_dhparams_manual_not_one_dir.bats/user-patches/user-patches.sh
./test/duplicate_configs/mail_default_dhparams_both_one_dir/user-patches/user-patches.sh
./test/duplicate_configs/mail_dhparams_manual_one_dir.bats/user-patches/user-patches.sh
./test/duplicate_configs/mail_default_dhparams_both_not_one_dir/user-patches/user-patches.sh
./testconfig.bak/user-patches/user-patches.sh
./target/scripts/fail2ban-wrapper.sh
./target/scripts/update-check.sh
./target/scripts/start-mailserver.sh
./target/scripts/helper-functions.sh
./target/scripts/quota-warning.sh
./target/scripts/check-for-changes.sh
./target/scripts/startup/misc-stack.sh
./target/scripts/startup/fixes-stack.sh
./target/scripts/startup/check-stack.sh
./target/scripts/startup/daemons-stack.sh
./target/scripts/startup/setup-stack.sh
./target/scripts/postsrsd-wrapper.sh
./target/scripts/postfix-wrapper.sh
./.github/workflows/scripts/docs/update-versions-json.sh
./.github/workflows/scripts/docs/build-docs.sh: ./setup.sh
./test/linting/lint.sh
./test/config/user-patches/user-patches.sh
./test/duplicate_configs/mail_dhparams_manual_not_one_dir.bats/user-patches/user-patches.sh
./test/duplicate_configs/mail_default_dhparams_both_one_dir/user-patches/user-patches.sh
./test/duplicate_configs/mail_dhparams_manual_one_dir.bats/user-patches/user-patches.sh
./test/duplicate_configs/mail_default_dhparams_both_not_one_dir/user-patches/user-patches.sh
./testconfig.bak/user-patches/user-patches.sh
./target/scripts/fail2ban-wrapper.sh
./target/scripts/update-check.sh
./target/scripts/start-mailserver.sh
./target/scripts/helper-functions.sh
./target/scripts/quota-warning.sh
./target/scripts/check-for-changes.sh
./target/scripts/startup/misc-stack.sh
./target/scripts/startup/fixes-stack.sh
./target/scripts/startup/check-stack.sh
./target/scripts/startup/daemons-stack.sh
./target/scripts/startup/setup-stack.sh
./target/scripts/postsrsd-wrapper.sh
./target/scripts/postfix-wrapper.sh
./.github/workflows/scripts/docs/update-versions-json.sh
./.github/workflows/scripts/docs/build-docs.sh: openBinaryFile: does not exist (No such file or directory)

––– SHELLCHECK
  – type    = FAILURE
  – message = errors encountered
  ```
  
  maybe there is a better way to do this?

**Comment by polarathene** [comment]
Would there be interest in avoiding any macOS specific additions (at least regarding linting/testing) by instead having another docker container handle it?

That could provide bats, jq and anything else instead of requiring the contributor to have the dependencies on their system (_we may want to keep bats as a git submodule_).

Running docker commands from within a container is a bit more tricky to contribute, but I was looking into that. It should result in a simpler linting/testing experience for contributors and easier for maintainers not having to worry about macOS compatibility with shell scripts.

It doesn't need to block this PR, that can sorted out in future if preferred and revert the macOS compatibility changes in this PR (_excluding `setup.sh`, although it's desirable to move most of it's usefulness into the image itself at some point_).

File: test/linting/lint.sh [hunk 12]
```diff
@@ -104,73 +90,40 @@ function _hadolint
...
+    --external-sources 
+    --check-sourced
+    --severity=style
+    --color=auto
+    --wiki-link-count=50
+    --enable=all
+    --exclude=SC2154
+    --source-path=SCRIPTDIR
+    ${F_SH} ${F_BIN} ${F_BATS}
+  )
```
polarathene, line 116:
```suggestion
  CMD_SHELLCHECK=(shellcheck 
    --external-sources 
    --check-sourced
    --severity=style
    --color=auto
    --wiki-link-count=50
    --enable=all
    --exclude=SC2154
    --source-path=SCRIPTDIR
    "${F_SH} ${F_BIN} ${F_BATS}"
  )
```

test/linting/lint.sh
```diff
@@ -104,73 +90,40 @@ function _hadolint
...
-    then
-      ERR=1
-    fi
-  done < <(find test/ -maxdepth 1 -type f -iname "*.bats")
-
-  if [[ ${ERR} -eq 1 ]]
+  if docker run --rm --tty \
+      --volume "${REPO_ROOT}:/ci:ro" \
+      --workdir "/ci" \
+      "koalaman/shellcheck-alpine:v${SHELLCHECK_VERSION}" "${CMD_SHELLCHECK[@]}"
```
polarathene, line 121:
```suggestion
  # shellcheck disable=SC2068
  if docker run --rm --tty \
      --volume "${REPO_ROOT}:/ci:ro" \
      --workdir "/ci" \
      "koalaman/shellcheck-alpine:v${SHELLCHECK_VERSION}" ${CMD_SHELLCHECK[@]}
```

test/linting/lint.sh
```diff
@@ -104,73 +90,40 @@ function _hadolint
...
+  
+  # File paths for shellcheck:
+  F_SH="$(find . -type f -iname '*.sh' \
+    -not -path './test/bats/*' \
+    -not -path './test/test_helper/*' \
+    -not -path './target/docker-configomat/*' -print0 \
+    | xargs -0
+  )"
+  F_BIN="$(find 'target/bin' -perm +111 -type f -or -type l)"
+  F_BATS="$(find 'test' -maxdepth 1 -type f -iname '*.bats')"
```
polarathene, line 102: I prefer to be explicit about `test` being a directory. Was there a reason the `/` was dropped?

```suggestion
  F_BATS="$(find 'test/' -maxdepth 1 -type f -iname '*.bats')"
```

polarathene (changes requested):

test/linting/lint.sh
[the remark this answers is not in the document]
polarathene: See my recent code suggestion for this. At least on my Ubuntu 21.04 VPS this does not run into that error with the changes I proposed before.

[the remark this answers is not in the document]
polarathene: > Hm, that's odd... I'll check again.

You may have missed my edit which added a link explaining why it doesn't work.

test/linting/lint.sh
NorseGaud: The quotes variables doesn't work

test/linting/lint.sh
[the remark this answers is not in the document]
NorseGaud: > Please test the change suggestions on your system and let me know if they work as well.

Nope:

```
    ~/docker-mailserver  on   macos-lint-and-test-support  make clean shellcheck                  ✔  took 4s   at 19:30:10  
./setup.sh ./test/linting/lint.sh ./test/config/user-patches/user-patches.sh ./target/scripts/fail2ban-wrapper.sh ./target/scripts/update-check.sh ./target/scripts/start-mailserver.sh ./target/scripts/helper-functions.sh ./target/scripts/quota-warning.sh ./target/scripts/check-for-changes.sh ./target/scripts/startup/misc-stack.sh ./target/scripts/startup/fixes-stack.sh ./target/scripts/startup/check-stack.sh ./target/scripts/startup/daemons-stack.sh ./target/scripts/startup/setup-stack.sh ./target/scripts/postsrsd-wrapper.sh ./target/scripts/postfix-wrapper.sh ./.github/workflows/scripts/docs/update-versions-json.sh ./.github/workflows/scripts/docs/build-docs.sh target/bin/setquota
target/bin/updatemailuser
target/bin/debug-fetchmail
target/bin/open-dkim
target/bin/excluderelaydomain
target/bin/listmailuser
target/bin/fail2ban
target/bin/listalias
target/bin/addalias
target/bin/delmailuser
target/bin/setup-fetchmail
target/bin/addrelayhost
target/bin/virus-wiper
target/bin/delalias
target/bin/postfix-summary
target/bin/restrict-access
target/bin/addmailuser
target/bin/delquota
target/bin/addsaslpassword
target/bin/fetchmailrc_split
target/bin/report-pflogsumm-yesterday test/mail_dhparams_default.bats
test/test_helper.bats
test/mail_postfix_inet.bats
test/mail_lmtp_ip.bats
test/mail_dhparams_manual_not_one_dir.bats
test/mail_pop3.bats
test/mail_smtponly.bats
test/mail_undef_spam_subject.bats
test/mail_privacy.bats
test/mail_fetchmail.bats
test/mail_spam_bounced.bats
test/mail_fail2ban.bats
test/mail_dhparams_manual_one_dir.bats
test/mail_with_postgrey_disabled_by_default.bats
test/permit_docker.bats
test/mail_fetchmail_parallel.bats
test/mail_disabled_clamav_spamassassin.bats
test/mail_spam_junk_folder.bats
test/mail_srs_domainname.bats
test/default_relay_host.bats
test/mail_postscreen.bats
test/mail_with_imap.bats
test/mail_ssl_letsencrypt.bats
test/mail_with_ldap.bats
test/tests.bats
test/mail_with_relays.bats
test/helper_functions.bats
test/mail_special_use_folders.bats
test/mail_with_mdbox.bats
test/security_tls_cipherlists.bats
test/mail_with_postgrey.bats
test/open_dkim.bats
test/mail_with_sdbox.bats
test/mail_override_hostname.bats
test/mail_quotas_disabled.bats
test/mail_ssl_manual.bats: ./setup.sh ./test/linting/lint.sh ./test/config/user-patches/user-patches.sh ./target/scripts/fail2ban-wrapper.sh ./target/scripts/update-check.sh ./target/scripts/start-mailserver.sh ./target/scripts/helper-functions.sh ./target/scripts/quota-warning.sh ./target/scripts/check-for-changes.sh ./target/scripts/startup/misc-stack.sh ./target/scripts/startup/fixes-stack.sh ./target/scripts/startup/check-stack.sh ./target/scripts/startup/daemons-stack.sh ./target/scripts/startup/setup-stack.sh ./target/scripts/postsrsd-wrapper.sh ./target/scripts/postfix-wrapper.sh ./.github/workflows/scripts/docs/update-versions-json.sh ./.github/workflows/scripts/docs/build-docs.sh target/bin/setquota
target/bin/updatemailuser
target/bin/debug-fetchmail
target/bin/open-dkim
target/bin/excluderelaydomain
target/bin/listmailuser
target/bin/fail2ban
target/bin/listalias
target/bin/addalias
target/bin/delmailuser
target/bin/setup-fetchmail
target/bin/addrelayhost
target/bin/virus-wiper
target/bin/delalias
target/bin/postfix-summary
target/bin/restrict-access
target/bin/addmailuser
target/bin/delquota
target/bin/addsaslpassword
target/bin/fetchmailrc_split
target/bin/report-pflogsumm-yesterday test/mail_dhparams_default.bats
test/test_helper.bats
test/mail_postfix_inet.bats
test/mail_lmtp_ip.bats
test/mail_dhparams_manual_not_one_dir.bats
test/mail_pop3.bats
test/mail_smtponly.bats
test/mail_undef_spam_subject.bats
test/mail_privacy.bats
test/mail_fetchmail.bats
test/mail_spam_bounced.bats
test/mail_fail2ban.bats
test/mail_dhparams_manual_one_dir.bats
test/mail_with_postgrey_disabled_by_default.bats
test/permit_docker.bats
test/mail_fetchmail_parallel.bats
test/mail_disabled_clamav_spamassassin.bats
test/mail_spam_junk_folder.bats
test/mail_srs_domainname.bats
test/default_relay_host.bats
test/mail_postscreen.bats
test/mail_with_imap.bats
test/mail_ssl_letsencrypt.bats
test/mail_with_ldap.bats
test/tests.bats
test/mail_with_relays.bats
test/helper_functions.bats
test/mail_special_use_folders.bats
test/mail_with_mdbox.bats
test/security_tls_cipherlists.bats
test/mail_with_postgrey.bats
test/open_dkim.bats
test/mail_with_sdbox.bats
test/mail_override_hostname.bats
test/mail_quotas_disabled.bats
test/mail_ssl_manual.bats: openBinaryFile: does not exist (No such file or directory)

––– SHELLCHECK
  – type    = FAILURE
  – message = errors encountered

make: *** [shellcheck] Error 1
```

test/linting/lint.sh
polarathene: I wasn't able to suggest both changes as a single change. You did test this `CMD_SHELLCHECK` suggestion with the `docker run` change that removed quotes for `CMD_SHELLCHECK[@]` right? Otherwise no it would not work.

NorseGaud: Doesn't work on ubuntu either
Screen Shot 2021-06-06 at 9 03 56 PM

test/linting/lint.sh
NorseGaud:
```
In ./test/linting/lint.sh line 121:
      "koalaman/shellcheck-alpine:v${SHELLCHECK_VERSION}" ${CMD_SHELLCHECK[@]}
                                                          ^------------------^ SC2068: Double quote array expansions to avoid re-splitting elements.

For more information:
  https://www.shellcheck.net/wiki/SC2068 -- Double quote array expansions to ...

––– SHELLCHECK
  – type    = FAILURE
  – message = errors encountered

make: *** [shellcheck] Error 1
```

It seems like we have two options:

1. Quote wrap `${F_SH} ${F_BIN} ${F_BATS}`
2. or quota wrap `${CMD_SHELLCHECK[@]}`

test/linting/lint.sh
NorseGaud:
```
find 'test/' -maxdepth 1 -type f -iname '*.bats'
test//mail_dhparams_default.bats
test//test_helper.bats
test//mail_postfix_inet.bats
test//mail_lmtp_ip.bats
```

polarathene: It is, I was under the impression it might have matched other directories with `test` in the name, but I was mistaken. I don't see any harm in keeping the `/`, but I'm not going to insist on it.

test/linting/lint.sh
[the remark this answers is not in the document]
NorseGaud: New find command handles finding permission bit files properly for both mac and Linux now:

Screen Shot 2021-06-06 at 9 30 43 PM
Screen Shot 2021-06-06 at 9 30 29 PM

test/linting/lint.sh
polarathene: > But not both.

Correct. My change suggestions (_albeit unfortunately split into two, sorry for not adding context about them being required together_) did account for that. I prefer that we keep the args all in `CMD_SHELLCHECK` and keeping all those filepaths wrapped in quotes seems preferred imo.

@aendeavor may know of how to work around that issue if he wants to enforce the requirement for wrapping with quotes.

test/linting/lint.sh
polarathene: > Doesn't work on ubuntu either

It absolutely does when applying both of my change requests. Which I think you figured out based on your related comment?

NorseGaud: Yep, we're good :)

test/linting/lint.sh
```diff
@@ -104,73 +90,41 @@ function _hadolint
...
+  
+  # File paths for shellcheck:
+  F_SH="$(find . -type f -iname '*.sh' \
+    -not -path './test/bats/*' \
+    -not -path './test/test_helper/*' \
+    -not -path './target/docker-configomat/*' -print0 \
+    | xargs -0
+  )"
+  [[ "$(uname)" == "Darwin" ]] && PERM="+111" || PERM="/111"
+  F_BIN="$(find 'target/bin' -perm "${PERM}" -type f -or -type l)"
```
polarathene, line 102: This is only a suggestion if the other maintainers would prefer it. Since both variants produce 21 files, there is no parity concern at present:

> _It took me a while to understand the implications of "not identical to the `-executable` predicate" and "just tests if any execute permissions are set": It means that `-perm +111` **may yield false positives**, i.e., files that the current user cannot actually execute._
> _There's no way to emulate `-executable` by testing permissions alone, because what's needed is to relate the file's user and group identity to the current user's._

I don't see why we'd ever run into that difference, so this suggestion probably doesn't need to be applied.

```suggestion
  # macOS lacks parity for `-executable` but presently produces the same results: https://stackoverflow.com/a/4458361
  [[ "$(uname)" == "Darwin" ]] && FIND_EXEC="-perm +111 -type l -or" || FIND_EXEC="-executable"
  F_BIN="$(find 'target/bin' "${FIND_EXEC}" -type f)"
```

Makefile
```diff
@@ -29,8 +25,8 @@ clean:
...
+	-@ for container in $$(docker ps -a | grep -E "mail|ldap_for_mail|mail_overri.*|hadolint|eclint|shellcheck" | cut -f 1-1 -d ' '); do docker rm -f $$container; done
+	-@ sudo rm -rf tools test/onedir test/alias test/quota test/relay test/config/dovecot-lmtp/userdb test/config/key* test/config/opendkim/keys/domain.tld/ test/config/opendkim/keys/example.com/ test/config/opendkim/keys/localdomain2.com/ test/config/postfix-aliases.cf test/config/postfix-receive-access.cf test/config/postfix-receive-access.cfe test/config/dovecot-quotas.cf test/config/postfix-send-access.cf test/config/postfix-send-access.cfe test/config/relay-hosts/chksum test/config/relay-hosts/postfix-aliases.cf test/config/dhparams.pem test/config/dovecot-lmtp/dh.pem test/config/relay-hosts/dovecot-quotas.cf test/config/user-patches.sh test/alias/config/postfix-virtual.cf test/quota/config/dovecot-quotas.cf test/quota/config/postfix-accounts.cf test/relay/config/postfix-relaymap.cf test/relay/config/postfix-sasl-password.cf test/duplicate_configs/
```
polarathene, line 29: Is `tools` still a relevant directory to add here now?

docs/content/contributing/issues-and-pull-requests.md
```diff
@@ -35,8 +35,7 @@ The development workflow is the following:
...
+4. [Prepare your environment and run linting and tests](./tests.md)
```
polarathene, line 38:
```suggestion
4. [Prepare your environment and run linting and tests][docs-tests]
```

And add `[docs-tests]: ./test.md` to the bottom of the document, next to `docs-environment`. Makes it a little easier for us to maintain.

Also, I noticed the numbered list needs to be updated since you deleted the 5th entry. This won't show up on the rendered docs but we should probably fix that here :)

The rendered docs are rendered wrong however. This isn't your fault, but the not so great markdown parser our docs use. I can see that indents are only 3 spaces instead of 4. You don't have to bother with this, especially if that doesn't fix the rendering properly (may also need a blank line before the nested lists start).

polarathene (changes requested):
Almost there, thanks for your patience! :sweat_smile:

test/linting/lint.sh
```diff
@@ -104,73 +90,43 @@ function _hadolint
...
-
-  if [[ ${ERR} -eq 1 ]]
+  
+  # File paths for shellcheck:
+  F_SH="$(find . -type f -iname '*.sh' \
+    -not -path './test/bats/*' \
+    -not -path './test/test_helper/*' \
+    -not -path './target/docker-configomat/*' -print0 \
+    | xargs -0
+  )"
```
polarathene, line 100:
```suggestion
  F_SH="$(find . -type f -iname '*.sh' \
    -not -path './test/bats/*' \
    -not -path './test/test_helper/*' \
    -not -path './target/docker-configomat/*'
  )"
```

polarathene (approved):
Awesome work @NorseGaud ! Thanks for taking the time to contribute :grinning: 

---

### TL;DR for reviewers of PR thread thus far

Summary of anything possibly worth pointing out:

---

https://github.com/docker-mailserver/docker-mailserver/pull/2001#discussion_r645379310 notes that this should have been reverted, but has resulted in: `"${@:+$@}"`, apparently due to ["unbound variable problems"](https://github.com/docker-mailserver/docker-mailserver/pull/2001/commits/48fb61eeb046548cfcdb890bfccd5feaa2fe911a).

---

https://github.com/docker-mailserver/docker-mailserver/pull/2001#discussion_r646224368 notes that the change from `find` option `-executable` for compatibility reasons in `lint.sh` may result in potential false positives, but it's probably unlikely for this project:

```bash
[[ "$(uname)" == "Darwin" ]] &amp;&amp; PERM="+111" || PERM="/111" 
F_BIN="$(find 'target/bin' -perm "${PERM}" -type f -or -type l)"
```

---

https://github.com/docker-mailserver/docker-mailserver/pull/2001#discussion_r646204993 notes another minor `find` change with `test` instead of `test/` for matching the target directory to scan (_this only covers a limited depth mind you, not sure if that's intentional to skip some of the contents further down?_). Should otherwise be harmless?

---

https://github.com/docker-mailserver/docker-mailserver/pull/2001#discussion_r646196699 notes that logging the linter version has been dropped. If maintainers still desire that, now that we use docker images for linting, the versions in the variables should be sufficient to log with?

---

https://github.com/docker-mailserver/docker-mailserver/pull/2001#discussion_r646196300 notes `ERR` has been dropped in `lint.sh`, but is still referenced by this line:

`trap '__log_err "${FUNCNAME[0]:-?}" "${BASH_COMMAND:-?}" ${LINENO:-?} ${?:-?}' ERR`

---

https://github.com/docker-mailserver/docker-mailserver/pull/2001#pullrequestreview-675861543 notes unrelated to changes in this PR, our cleanup logic in Makefile needs to be addressed, tests aren't conforming to conventions preventing them from being matched.

---

https://github.com/docker-mailserver/docker-mailserver/pull/2001#discussion_r646231226 notes unrelated to this PR, the [issues and pull requests docs page](https://pullrequest-2001--dms-doc-previews.netlify.app/contributing/issues-and-pull-requests/) still has a rendering issue of nested lists.

test/linting/lint.sh
```diff
@@ -9,9 +9,9 @@ SCRIPT="lint.sh"
...
 
-KERNEL_NAME=$(uname -s)
-KERNEL_NAME_LOWERCASE=$(echo "${KERNEL_NAME}" | tr '[:upper:]' '[:lower:]')
-MACHINE_ARCH=$(uname -m)
+HADOLINT_VERSION=2.4.1
+ECLINT_VERSION=2.3.5
+SHELLCHECK_VERSION=0.7.2
 
 set -eEuo pipefail
 trap '__log_err "${FUNCNAME[0]:-?}" "${BASH_COMMAND:-?}" ${LINENO:-?} ${?:-?}' ERR
```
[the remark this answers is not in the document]
aendeavor: Might have been a bit misleading, but `ERR` at the end of line 17 refers to the _error signal_ that `trap` should handle. The variables later used in the script called `ERR` are _not_ related to the signals which has the same name. Dropping these variables is good thing as it avoids confusion :)

aendeavor (approved):

polarathene:
@aendeavor is the trap line with `ERR` at the end still valid?

aendeavor:
> @aendeavor is the trap line with `ERR` at the end still valid?

Yes :) In case of an unexpected error, it will print information via the `__log_err` function. This can help debugging or with issue reports. I'd leave it in there.

Makefile
```diff
@@ -29,7 +25,7 @@ clean:
 # remove running and stopped test containers
 	-@ [[ -d config.bak ]] && { rm -rf config ; mv config.bak config ; } || :
 	-@ [[ -d testconfig.bak ]] && { sudo rm -rf test/config ; mv testconfig.bak test/config ; } || :
-	-@ docker ps -a | grep -E "mail|ldap_for_mail|mail_overri.*" | cut -f 1-1 -d ' ' | xargs --no-run-if-empty docker rm -f
+	-@ for container in $$(docker ps -a | grep -E "mail|ldap_for_mail|mail_overri.*|hadolint|eclint|shellcheck" | cut -f 1-1 -d ' '); do docker rm -f $$container; done
```
casperklein, line 28: This is problematic/dangerous:

1. `ldap_for_mail|mail_overri.*` will never match, `mail` will match first.
2. All matched containers will be forcefully removed. Think about a container named `webmail` etc.

setup.sh
```diff
@@ -225,7 +231,7 @@ function _docker_image
   if ${USE_CONTAINER}
   then
     # reuse existing container specified on command line
-    ${CRI} exec "${USE_TTY}" "${CONTAINER_NAME}" "${@}"
+    ${CRI} exec "${USE_TTY}" "${CONTAINER_NAME}" "${@:+$@}"
```
casperklein, line 234: I don't understand the purpose of all these `"${@:+$@}"`.

`If $@ is not empty; then use $@`?

test/linting/lint.sh
```diff
@@ -6,6 +6,13 @@
 
 SCRIPT="lint.sh"
 
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
```
casperklein, line 9: `SCRIPT_DIR=$(dirname "$(readlink -f "$0")")`

setup.sh
NorseGaud: @casperklein https://stackoverflow.com/questions/44567534/work-around-for-unbound-variable-in-bash-4-0-0

casperklein: Did you encounter a problem or was this added in good faith?

BASH 4.0.0 was released more than 10 years ago. And the bug got fixed in later 4.x releases (just verified it with debian9 which included BASH 4.4).

NorseGaud: Even with the latest bash in macOS it was throwing the error for me. It's a macOS specific fix

wernerfred: Isn't Mac using Bash 3? :D

NorseGaud: MacOS users will install the brew bash which is 5:

```
    ~/docker-mailserver  on   issues/1979 !2  grep '_main "' setup.sh
_main "${@}"

    ~/docker-mailserver  on   issues/1979 !2  ./setup.sh
./setup.sh: line 415: @: unbound variable

    ~/docker-mailserver  on   issues/1979 !2  bash --version
GNU bash, version 5.1.8(1)-release (x86_64-apple-darwin20.3.0)
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

aendeavor: When brew installs Bash 5, I think we do not actually need this, right?

NorseGaud: See my code block above. Definitely needed

aendeavor: Oops. I had overlooked that :)

Makefile
NorseGaud: Fixed in https://github.com/docker-mailserver/docker-mailserver/pull/1980/commits/5084caf2da2dd20d7ecc741a8f6bfa7a5952962d

setup.sh
polarathene: @NorseGaud doesn't `setup.sh` say to use `#! /bin/bash`? Does that affect the behaviour at all? Is it actually using the v5 of Bash from brew install?

aendeavor: Good observation @polarathene! @NorseGaud try `bash setup.sh`

aendeavor:
> > Yes :) In case of an unexpected error, it will print information via the `__log_err` function. This can help debugging or with issue reports. I'd leave it in there.
> 
> @aendeavor I meant the `ERR` variable part as it no longer exists anywhere else in the file?

I understand. Then this may not be needed anymore and can be removed entirely I guess.

setup.sh
NorseGaud: @aendeavor @polarathene Derp... Can't believe I didn't think about that 😅.  How do we feel about using `#!/usr/bin/env bash`?

NorseGaud: Also, this all presumes the user has /usr/local/bin in their PATH before /bin... I personally think `${@:+$@}` is worth it.

wernerfred: On my mac i needed to `hash -r` after `brew install bash` and this only works bc `/usr/local/bin` ist listed first on default environments. So if it is environment safe as long as bash4+ is installed anywhere on the system we should go with  `${@:+$@}` i think

aendeavor: Not sure TBH. This construct just looks weird to me 😆 

But I guess we could make an effort to support macOS users:)

casperklein: > Isn't Mac using Bash 3 by default? 

Holy shit. YMMD 😄 Couldn't believe that's true.

polarathene: Is it worth leaving a comment with description or link about it for future maintainer reference? Or is the link to this PR with lengthy discussion thread to wade through sufficient?

PR #2014, opened 2021-05-29
Because it renders incorrectly in the terminal, what tf even is that character? It's character code is 8211, like wtf?

ihack2712:
image

No reason to be redicoulously fancy, I mean this works...

image

### DIFF
--- a/mailserver.env
+++ b/mailserver.env
@@ -1,13 +1,13 @@
-# –––––––––––––––––––––––––––––––––––––––––––––––
-# ––– Mailserver Environment Variables ––––––––––
-# –––––––––––––––––––––––––––––––––––––––––––––––
+# -----------------------------------------------
+# --- Mailserver Environment Variables ----------
+# -----------------------------------------------
 
 # DOCUMENTATION FOR THESE VARIABLES IS FOUND UNDER
 # https://docker-mailserver.github.io/docker-mailserver/edge/config/environment/
 
-# –––––––––––––––––––––––––––––––––––––––––––––––
-# ––– General Section –––––––––––––––––––––––––––
-# –––––––––––––––––––––––––––––––––––––––––––––––
+# -----------------------------------------------
+# --- General Section ---------------------------
+# -----------------------------------------------
 
 # empty => uses the `hostname` command to get the mail server's canonical hostname
 # => Specify a fully-qualified domainname to serve mail for.  This is used for many of the config features so if you can't set your hostname (e.g. you're in a container platform that doesn't let you) specify it in this environment variable.
@@ -227,9 +227,9 @@ REPORT_INTERVAL=daily
 # Note: More details in http://www.postfix.org/postconf.5.html#inet_protocols
 POSTFIX_INET_PROTOCOLS=all
 
-# –––––––––––––––––––––––––––––––––––––––––––––––
-# ––– SpamAssassin Section ––––––––––––––––––––––
-# –––––––––––––––––––––––––––––––––––––––––––––––
+# -----------------------------------------------
+# ----SpamAssassin Section ----------------------
+# -----------------------------------------------
 
 ENABLE_SPAMASSASSIN=0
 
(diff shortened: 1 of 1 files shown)


aendeavor:
**Like all othera too, you will stick to the PR template and give this PR a proper name**. Just because you find something _ridiculous does not mean rules do not apply to you.

---

The issue at hand is not related to a terminal, but to the font your terminal uses.

Screenshot from 2021-05-29 23-37-56

The font your terminal uses is not capable of displaying the _en dash_, a rather common character which is just a longer dash. I have put them there and prefer them over the normal dash, because, as you can see in the picture, **they form a straight line**.

I know where this is coming from, and it may be a bit fancy, true. I will leave this up to the other @docker-mailserver/maintainers to decide. If they think this really is a _valuable change_, they should merge it. But _I will not approve nor merge this PR_.

polarathene:
`–` en dash (Codepoint: U+2013, Decimal: 8211)](https://en.wikipedia.org/wiki/Dash#Common_dashes_and_their_Unicode_code_points). This glyph has been an [official part of Unicode since `1.1.0` (1993)](https://www.fileformat.info/info/unicode/char/2013/index.htm).

There's also the em dash, which is said to be a longer variant and to my understanding should always render as full width (1 em in length, as opposed to en which apparently lacks some consistency). There's also [other variants for this purpose that could be suitable](https://en.wikipedia.org/wiki/Dash#Similar_characters_and_their_Unicode_code-points).

Of note, it would seem the [Box Drawing](https://en.wikipedia.org/wiki/Box_Drawing_(Unicode_block)) glyphs cover this more appropriately than dash glyphs intended primarily for typography? These unicode glyphs have had support since Unicode `1.0.0` (1991) and almost certainly more likely to be present in terminal fonts. I believe this is what TUI programs also use when rendering lines for UI purposes.

`─` or the thicker `━` should provide adequate substitutes? @ihack2712 could confirm if whatever font they're using renders these just as well, then both parties will be satisfied :)

aendeavor:
@polarathene I actually wanted to say _em_ dash but hit the "n" instead of the "m" :D

The Box Drawings are actually a good idea – why didn't I think of them earlier. `─` (U+2500) should then be our choice.

casperklein:
I would stick to plain ASCII, so everyone can enjoy/extend this file, even without a proper font. If the majority is for UTF8 usage, it's fine for me. It doesn't bother me that much 😄 

PS: I didn't know, that UTF8 characters were used, until ihack2712 mentioned it. I use PuTTY / ViM and the characters looks like normal dashes for me (Font: consolas). You would only see the difference, if you add a regular dash for comparison.

putty

So when adding a new headline, I would have to keep in mind, to copy/paste an existing headline instead of using my keyboard dash button (to keep all uniformly).

polarathene:
> I would stick to plain ASCII, so everyone can enjoy/extend this file, even without a proper font.

I would think that you shouldn't have a problem with the Unicode Box Drawing glyphs from 1991. Most OS are meant to have fallback fonts for glyphs a font is missing as well, but perhaps that support is not as good in terminal programs for some reason.

> I use PuTTY / ViM and the characters looks like normal dashes for me (Font: consolas).

If they were instead em dash I think it would look like a solid line as it's meant to be full width. But the Box Drawing horizontal lines would be more appropriate in that case.

> So when adding a new headline, I would have to keep in mind, to copy/paste an existing headline instead of using my keyboard dash button (to keep all uniformly).

This is a valid point and I don't think most care much how these are styled, only that they render correctly and for contributors are easy to add/modify. I could definitely see potential for mixed glyphs or a need to copy/paste (because who else will know from a glance we're using U+2500 or whatever which also isn't something you'd usually type in a document either :stuck_out_tongue: 

Maybe we should just use the plain hyphen? (I do for a similar block heading style in my own config files)

aendeavor:
If you‘re going to change it, then please change it in all the documents :) That‘d be my only wish.

PR #2030, opened 2021-06-08
chore(dockerfile): Follow up on #2024 implementing @polarathene's suggestions

georglauterbach:
# Description

Follow up on #2024.

@polarathene Note that I enabled auto-merging again :D You may turn it off if there is more you'd like to discuss.

## Type of change

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,10 +40,10 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
 RUN \
   apt-get -qq update && \
-  apt-get -y -qq install apt-utils && \
-  apt-get -y -qq dist-upgrade && \
-  apt-get -y -q install postfix && \
-  apt-get -y -q --no-install-recommends install \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -40,10 +40,10 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
 RUN \
   apt-get -qq update && \
-  apt-get -y -qq install apt-utils && \
-  apt-get -y -qq dist-upgrade && \
-  apt-get -y -q install postfix && \
-  apt-get -y -q --no-install-recommends install \
+  apt-get -qq install apt-utils 2>/dev/null && \
```
casperklein, line 43: `2>/dev/nulll`

What's the rational behind this? It will hide error output (again)?

aendeavor: With `apt-utils`, the error message will always be the same: `debconf: delaying package configuration, since apt-utils is not installed` 😆 This is an inherently useless message, I figured. Hence, the error output from this command can be hidden.

PR #2031, opened 2021-06-09
Create Archives folder and subscribe by default

### LINKED ISSUE #2026: [FR] Enable "Archive" folder by default
# Feature Request

## Context
After first installation, on my mobile phone I am using Outlook for Android. And on my Linux desktop, I am using Thunderbird.

Thunderbird has an "archive" feature and automatically creates "Archives" folder in IMAP.
Outlook for Android also has an "archive" feature. But it does not use "Archives" folder created by Thunderbird. It uses a folder named "Archive" which is not even created on IMAP.

Kind of.

Add "Archives" folder by default.

I changed the config file and verified it works. And it's a simple change.

Should this config change has a feature toggle in the mailserv.env so it won't impact existing users?

I believe my usage, Outlook for Android and Thunderbird on desktop, is not that unusual.

I created the working config file and verified it in my installation.

hnws:
# Description
Clients should be able to share this "Archives" folder by default.
Usually email clients want to have an "Archives" folder.
By creating one following RFC-6154, all clients should be able to use it.

Tested on Linux Thunderbird, Outlook for Android and Spark for Android.

Fixes #2026

## Type of change

## Checklist:

### DIFF
--- a/target/dovecot/15-mailboxes.conf
+++ b/target/dovecot/15-mailboxes.conf
@@ -28,7 +28,7 @@ namespace inbox {
     auto = subscribe
     special_use = \Trash
   }
-  mailbox Archives {
+  mailbox Archive {
     auto = subscribe
     special_use = \Archive
   }
--- a/target/dovecot/15-mailboxes.conf
+++ b/target/dovecot/15-mailboxes.conf
@@ -28,6 +28,10 @@ namespace inbox {
     auto = subscribe
     special_use = \Trash
   }
+  mailbox Archives {
+    auto = subscribe
+    special_use = \Archive
+  }
 
   # For \Sent mailboxes there are two widely used names. We'll mark both of
   # them as \Sent. User typically deletes one of them if duplicates are created.


reneploetz:
Even tough RFC6154 does not specify a specific name, I think the folder should be named "Archive" instead of "Archives" for compatibility reasons. Some mail clients stick to the exact name and do not support the "SPECIAL-USE" flag in the LIST imap command. Most notably Windows Mail is using hardcoded localized values and does not support RFC6154.

aendeavor:
@reneploetz I‘ve been seeing you around here often lately, with very good answers and content. Curious to the join the moderator‘s team?

casperklein:
Does this change introduce any disadvantages? For example, in POP3 only setups?

NorseGaud:
> Even tough RFC6154 does not specify a specific name, I think the folder should be named "Archive" instead of "Archives" for compatibility reasons. Some mail clients stick to the exact name of the flag and do not support the "SPECIAL-USE" flag in the LIST IMAP command. Most notably Windows Mail is using hardcoded localized values and does not support RFC6154.
> 
> Note that the original bug report stated issues with Thunderbird which created "Archives" and Outlook using "Archive" which was not created on the server. Thunderbird does support RFC6154 and will then pick up the "Archive" as folder.

💯  I agree. Otherwise, I really like what this PR proposes

williamdes (approved):

hnws:
Updated the PR to use "Archive" instead of "Archives".
Thunderbird cannot use the new "Archive" folder automatically. It can be solved by moving all of the existing emails to new folder and change a setting.
On Spark for Android, I need to remove the accounts already setup to get rid of duplicated Archive folders.

How should I proceed with this? Should I learn and try to create a feature toggle?

reneploetz:
> @reneploetz I‘ve been seeing you around here often lately, with very good answers and content. Interested in joining the moderator‘s team?

I would be interested to help, yes.

@casperklein 
At least POP3 should be unaffected as only the INBOX is mapped.


@hnws 
I had a look into what Thunderbird is doing and it's somewhat tricky.
For new configurations, Thunderbird should pick up the folder name and use it accordingly. If the folder name has the "Archive" flag and is named "Archives", they will also localize it within their UI. Otherwise they will keep it named "Archive".
See also: https://github.com/mozilla/releases-comm-central/blob/master/mailnews/base/src/nsMsgDBFolder.cpp#L3089

If you have an existing IMAP configuration on both Spark and Thunderbird, this change will not update their configuration. You have to do this manually and potentially migrate.
Therefore I would say that this pull request currently is a breaking change as there might be other users that use different folder names for their current archive folders.

I had a look into how other projects are configuring dovecot:
- Some project are specifying a large number of translations and do not create/subscribe the translated names: https://github.com/mailcow/mailcow-dockerized/blob/master/data/conf/dovecot/dovecot.folders.conf#L75
- Some projects state that they leave it up to the user: https://gitlab.com/x-eysz7x/mailinabox/-/blob/master/conf/dovecot-mailboxes.conf#L19 (note: we are using the same configuration)

Maybe it's an idea to do a minimal test with one name subscribing and the other just referenced. We would need to test creating a new account as well as what happens with the current account if the change is applied.
```
  mailbox "Archive" {
    auto = subscribe
    special_use = \Archive
  }
  mailbox "Archives" {
    special_use = \Archive
  }
```


In general, there appears to be no consensus on the name of the folder. I see a couple of options:
- leave the name as "Archive" and let the user change the folder names manually - the server administrator can always override the default file using docker mounts or user-patches.sh
- use the "translate all" approach (which we *really* need to test first)
- create environment variables and startup scripts to allow the server administrator to configure the names - we can of course only do this with "Archive/Archives", but we are currently also commenting the "Sent/Sent Messages"

hnws:
How about this:
Instead of pushing this config to everyone, I will add this to documentation and provide the config file as an example?

aendeavor:
> How about this:
> Instead of pushing this config to everyone, I will add this to documentation and provide the config file as an example?

That's a very good idea!

georglauterbach:
Any progress on the subject?

hnws:
> Any progress on the subject?

Sorry for the reply. I think I will update this PR late this week to create a new document page for this purpose.

hnws:
Replaced by #2045

Issue #1906, opened 2021-04-14
The best way to backup my configuration, accounts and emails


e-fo:
# Subject
I have  questions about the best backup method for docker-mailserver

## Description
I searched your documents and I couldn't find anything about backup. I looking for a method that able me to backup my configs, accounts and their emails.

I appreciate you if can help me about that.

casperklein:
There is https://docker-mailserver.github.io/docker-mailserver/edge/faq/#what-about-backups. However, it's not complete.

Backup your volumes (maildata, mailstate, maillogs) and `config` directory should all you need to do.

aendeavor:
Just Backup the (bind-)mounted volumes, i.e. these directories on your drive, in a regular backup, nothing special:)

zzwxrchi:
https://docker-mailserver.github.io/docker-mailserver/edge/faq/#what-about-backups
Only backup config files, others folders are empty.

zzwxrchi:
Run for me with little change the command:

docker run --rm -ti --volumes-from ${ContainerName}\
  -v "${DOCKER-COMPOSE_ROOT_DIR}/config":/tmp/docker-mailserver \
  -v ${DIR_BACKUP}:/backup \
  alpine:latest \
  tar -czf "/backup/${FILE_NAME}_${DATESTAMP}.tar.gz" /var/mail /var/mail-state /var/log/mail /tmp/docker-mailserver

casperklein:
Check which sub directory makes it big.

zzwxrchi:
var/mail-state/lib-clamav folder, 440MB for a new installation with 2 email added.
With files: daily.cld 322.5MB and main.cvd 117.9 MB

casperklein:
You can exclude that directory, but then when you recreate the container, you'll not have the latest anti virus patterns. It will however update via cron job every 6 hours.

zzwxrchi:
Thanks @casperklein !!

Bluejanis:
The [solution in the FAQ](https://docker-mailserver.github.io/docker-mailserver/edge/faq/#what-about-backups) didnt work! It produced empty folders in the archive backup. 

Thanks @zzwxrchi for your snippet. That helped me to backup the whole mailserver successfully!

The documentation should be updated in this regard, so you can actually create complete backups with the available info. But for the most part the docker-mailserver documentation is extremely helpful!

casperklein:
[...]

PR #2035, opened 2021-06-14
Use bind mounts in sample config

casperklein:
# Description

If we switch to bind mounts in the sample config, that will simplify backup/restore of DMS.

See also: #1906

This would also make the not 100% correct working script at https://docker-mailserver.github.io/docker-mailserver/edge/faq/#what-about-backups obsolete.

Backup/restore would then just be simple host commands `cp / mv`..

What do you think?

## Type of change

## Checklist:

### DIFF
--- a/docs/content/faq.md
+++ b/docs/content/faq.md
@@ -54,16 +54,22 @@ Please do not use `CRLF`.
 
 ### What about backups?
 
+#### Bind mounts (default)
+
+```bash
+tar czf backup-$(date +%F).tar.gz config data
+```
+
+#### Volumes
+
 Assuming that you use `docker-compose` and data volumes, you can backup the configuration, emails and logs like this:
 
 ```sh
 # create backup
-docker run --rm -ti \
-  -v maildata:/var/mail \
-  -v mailstate:/var/mail-state \
-  -v maillogs:/var/logs/mail \
+docker run --rm -it \
   -v "$PWD/config":/tmp/docker-mailserver \
   -v /backup/mail:/backup \
+  --volumes-from mailserver \
   alpine:latest \
   tar czf "/backup/mail-$(date +%F).tar.gz" /var/mail /var/mail-state /var/logs/mail /tmp/docker-mailserver
 
--- a/docs/content/faq.md
+++ b/docs/content/faq.md
@@ -65,7 +65,7 @@ tar --gzip -cf "backup-$(date +%F).tar.gz" config data
 Then to restore `./config` and `./data` folders from your backup file:
 
 ```bash
-tar -xf backup-date.tar.gz
+tar --gzip -xf backup-date.tar.gz
 ```
 
 #### Volumes
--- a/docs/content/faq.md
+++ b/docs/content/faq.md
@@ -56,8 +56,16 @@ Please do not use `CRLF`.
 
 #### Bind mounts (default)
 
+From the location of your `docker-compose.yml`, create a compressed archive of your `./config` and `./data` folders:
+
 ```bash
-tar czf backup-$(date +%F).tar.gz config data
+tar --gzip -cf "backup-$(date +%F).tar.gz" config data
+```
+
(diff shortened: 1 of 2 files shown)


georglauterbach:
Actually a good idea. But this requires a documentation update.

casperklein:
For sure. But before wasting time, I wanted to know beforehand what your opinions are 😎

wernerfred:
I prefer bind mounts over volumes (personal opinion) so you have my vote

docs/content/faq.md
```diff
@@ -54,16 +54,22 @@ Please do not use `CRLF`.
 
 ### What about backups?
 
+#### Bind mounts (default)
+
+```bash
+tar czf backup-$(date +%F).tar.gz config data
+```
+
```
polarathene, line 62:
````suggestion
#### Bind mounts (default)

From the location of your `docker-compose.yml`, create a compressed archive of your `./config` and `./data` folders:

```bash
tar --gzip -cf "backup-$(date +%F).tar.gz" config data
```

Then to restore `./config` and `./data` folders from your backup file:

```bash
tar -xf backup-date.tar.gz
```

````

**Review by polarathene — CHANGES_REQUESTED** [review]
Might be worth adding to the FAQ how to restore?

I also wanted to be a bit more explicit about the option for compressing to gzip. I'd suggest `--zstd` with `tar.zst` but not sure how wide the support for that is.

**Comment by polarathene** [comment]
> Backup/restore would then just be simple host commands `cp / mv`..

Doesn't that mess with permissions / ownership metadata on files?

I prefer bind mounts myself, but a common issue I can run into with containers is the container writes the files with different uid/gid (mostly only an annoyance if you're non-root), or you configure some files from the host side and if the container lacks permission to read/write the file it gets upset.

Regarding the referenced issue, kinda seems odd to backup clamav bloat, depending on the age of the backup when it's restored, I'm not sure how worthwhile that would be to retain? This is still a nice improvement anyway :)

**Comment by georglauterbach** [comment]
> Might be worth adding to the FAQ how to restore?
> 
> I also wanted to be a bit more explicit about the option for compressing to gzip. I'd suggest `--zstd` with `tar.zst` but not sure how wide the support for that is.

I think creating a `.tar.gz` is the most widely compatible archive for Linux and we should use that:)

**Comment by polarathene** [comment]
> I think creating a `.tar.gz` is the most widely compatible archive for Linux and we should use that :)

Yeah, that's why I was reluctant with my preference to use zstd (_which is definitely better when available_).

I'd still prefer the more explicit option of `--gzip` however, since this is a FAQ item and the user is probably going to copy/paste anyway.

**Comment by wernerfred** [comment]
> I'd still prefer the more explicit option of `--gzip` however

I prefer verbosity, too. Always hard for future contributions to look up / follow along if a bunch of short options is thrown together using a tool a contributor might not know. As most will copy/paste it anyways it should not be a big deal

**Comment by georglauterbach** [comment]
Yes. We will use the long option `--gzip` :)

**Comment by casperklein** [comment]
@docker-mailserver/maintainers feel free to edit/add docu yourself to this PR 👍

**Review by georglauterbach — APPROVED** [review]

File: docs/content/faq.md [hunk 2]
```diff
@@ -65,7 +65,7 @@ tar --gzip -cf "backup-$(date +%F).tar.gz" config data
 Then to restore `./config` and `./data` folders from your backup file:
 
 ```bash
-tar -xf backup-date.tar.gz
+tar --gzip -xf backup-date.tar.gz
```
polarathene, line 68: [`--gzip` hint isn't needed](https://askubuntu.com/a/398336) for extraction.

casperklein: You are right. But it doesn't hurt if you ask me. I prefer calling tools with explicit parameters, especially in documentation. Feel free to remove it 👍

polarathene (approved):

NorseGaud:
<3

PR #2045, opened 2021-06-19
Adding instructions and discussion for folders

### LINKED ISSUE #2026: [FR] Enable "Archive" folder by default

hnws:
# Description

It tries to explain the process of creating typical folders and a few points for consideration.

Fixes #2026

## Type of change

## Checklist:

This replaces #2031


docs/content/config/best-practices/folders.md
```diff
@@ -0,0 +1,26 @@
...
+
+
+### Tag v.s. Folder name
+
+Certain email clients simply ignore the tag defined in `RFC 6154` and look of the exact name instead. So it might be necessary to create folder with name the same as tag. 
+
+### i18n 
+
+Users and email clients might want to have folder names in local language instead of English. Please test carefully. This might need to find a balance with the tag v.s. folder name issue.
+
```
polarathene, line 26:
```suggestion
## Adding a mailbox folder

See [`target/dovecot/15-mailboxes.conf`][gh-config-dovecot-mailboxes] for existing folder definitions you can modify or enable.
Dovecot also [provides their own example config][dovecot-config-mailboxes] for reference, with further information in [their documentation][dovecot-docs-mailboxes].

## Caution

### Adding folders to an existing setup

Handling of newly added mailbox folders can be inconsistent across mail clients:

- Users may experience issues such as archived emails only being available locally.
- Users may need to migrate emails manually between two folders.

### Tag Support

Not all mail clients support tags (_defined in [RFC 6154][rfc-6154]_), using the exact mailbox name instead. To support those clients, you may prefer to keep the tag and folder name identical.

### Internationalization (i18n)

Users and mail clients may prefer localized mailbox names instead of English. Take care to test localized names work well, keep in mind concerns such as tag support.

[gh-config-dovecot-mailboxes]: https://github.com/docker-mailserver/docker-mailserver/blob/master/target/dovecot/15-mailboxes.conf
[dovecot-config-mailboxes]: https://github.com/dovecot/core/blob/master/doc/example-config/conf.d/15-mailboxes.conf
[dovecot-docs-mailboxes]: https://doc.dovecot.org/configuration_manual/namespace/#mailbox-settings
[rfc-6154]: https://datatracker.ietf.org/doc/html/rfc6154

```

I have not read the RFC referenced, but it only has two `ctrl+f` matches for "tag" is that the correct word for that feature? Is this what a mail client refers to it as?

I also noticed the docs here focusing on "folder" name and creation, but is that the correct terminology or is it from the user perspective of what a mailbox is?

Honestly, I'm not sure about relevance for the i18n section? I had a brief search through some documents like Dovecot docs and the RFC mentioned, but I saw nothing about languages being mentioned. What is the actual feature or issue being described there? Is there an example scenario?

polarathene (changes requested):
I rephrased some content to sound more natural. I don't know too much about the actual topic being documented, so feel free to point out any misunderstandings I have.

Added some links and changed the link syntax so it's easier less noise to review markdown, and nicely groups the links for the docs page at the bottom which is preferred.

"Caution" perhaps isn't any better than "Consideration" for a heading. I thought of "Gotchas" but not sure how appropriate that is for the audience, especially if non-native english speakers? (_Is it a commonly understood word?_)

Some of the docs being contributed seem a bit iffy content wise. Without context I'm not quite sure why they're being mentioned, or how well they'd be understood (_eg tags and i18n_).

polarathene:
Note, as this is a new page you're contributing to the docs, you'll need to add a new entry into `mkdocs.yml`:

https://github.com/docker-mailserver/docker-mailserver/blob/80a0425adeee01f873b506bd9aa63eee2394ad95/docs/mkdocs.yml#L114-L118

I'm not sure if this really belongs to "Best Practices" personally, but that can be relocated at a later point if no maintainers want to suggest a more appropriate location.

wernerfred:
> I'm not sure if this really belongs to "Best Practices" personally, but this doc can be relocated at a later point if no maintainers want to suggest a more appropriate location.


I personally would Put it in the examples section. Not Sure wether use cases or examples fits netter though.

docs/content/examples/uses-cases/folders.md
```diff
@@ -0,0 +1,56 @@
+---
+title: 'Use Cases | Folders'
+hide:
+  - toc # Hide Table of Contents for this page
+---
+## Introduction
+
+This provides a solution letting different email clients sharing the same archive folder.
```
wernerfred, line 8: Does this really mean "share" the same folder (between accounts) or "share" the same configuration?

This wording might be confusing depending on what is meant indeed.

docs/content/examples/uses-cases/folders.md
```diff
@@ -0,0 +1,56 @@
...
+It might also help when troubleshooting sharing `Sent` and `Drafts` folders.
```
wernerfred, line 9: Same here

README.md
```diff
@@ -176,6 +176,21 @@ If you got any problems with SPF and/or forwarding mails, give [SRS](https://git
 
 See the [documentation](https://docker-mailserver.github.io/docker-mailserver/edge/config/security/understanding-the-ports/) for further details and best practice advice, **especially regarding security concerns**.
 
+#### Folders
+
+By default, `INBOX`, `Drafts`, `Sent`, `Junk`, `Trash` are created and automatically subscribed. Please take a look at [`target/dovecot/15-mailboxes.conf`][gh-config-dovecot-mailboxes] for existing folder definitions as reference. A useful example is to enable the `Archive` folder, allowing different mail clients to share the same archive folder automatically.
```
wernerfred, line 181: Same here

docs/content/examples/uses-cases/folders.md
```diff
@@ -0,0 +1,56 @@
...
+	If new email account is added after `Special Use Mailbox` enabled for archives, it will pick up the name assgined.
+
+*	Spark for Android
+	
+	If new email account is added without `Special Use Mailbox` enabled for archive, it archives on server folder named `Archive`.
+	If new email account is added after `Special Use Mailbox` enabled for archives, it will pick up the name assgined.
+
+* 	Windows Mail
+	
+	People suggested that it will look for the name instead of `Special Use Mailbox`.
```
polarathene, line 51:
```suggestion
### Email Clients Support

#### Thunderbird

- If a new email account is added without `Special Use Mailbox` enabled for archives, Thunderbird suggests and may create an `Archives` folder on the server.
- If a new email account is added after `Special Use Mailbox` is enabled for archives, it will pick up the name assigned.

#### Outlook for Android

- If a new email account is added without `Special Use Mailbox` enabled for archives, it archives locally.
- If a new email account is added after `Special Use Mailbox` is enabled for archives, it will pick up the name assigned.

#### Spark for Android

- If new email account is added without `Special Use Mailbox` enabled for archives, it archives on the server folder named `Archive`.
- If new email account is added after `Special Use Mailbox` is enabled for archives, it will pick up the name assigned.

#### Windows Mail

- People suggested that it will look for the name instead of `Special Use Mailbox`.
```

Why are there tabs used here? I suggest changing the style to sub-headings with bullet points added.

Alternatively, since there is repetition, this may be better:

```suggestion
### Email Clients Support

- If a new email account is added without `Special Use Mailbox` enabled for archives:
    - Thunderbird suggests and may create an `Archives` folder on the server.
    - _Outlook for Android_ archives locally.
    - _Spark for Android_ archives on the server folder named `Archive`.
- If a new email account is added after `Special Use Mailbox` is enabled for archives:
    - _Thunderbird_, _Outlook for Android_ and _Spark for Android_ will pick up the name assigned.

!!! caution "Windows Mail"

    _Windows Mail_ has been said to ignore `Special Use Mailbox` and look only at the name assigned.
```

I think this is better? I corrected grammar a bit and made some slight modifications, I hope it is still accurate.

It would be even better if there was a link to any resources for each statement, but it is not required for approval.

docs/content/examples/uses-cases/folders.md
```diff
@@ -0,0 +1,56 @@
...
+- Users may experience issues such as archived emails only being available locally.
+- Users may need to migrate emails manually between two folders.
+
+### Special Use Mailbox Support
+
+Not all mail clients support Special Use Mailbox (_defined in [RFC 6154][rfc-6154]_), using the exact mailbox name instead. To support those clients, you may prefer to keep the tag and folder name identical.
+
+### Internationalization (i18n)
+
+Users and mail clients may prefer localized mailbox names instead of English. Take care to test localized names work well, keep in mind concerns such as tag support.
```
polarathene, line 31: As you are making commits and resolving prior review with no discussion, "tag" is presumably still unclear? Please correct the naming or provide context what a "tag" refers to in the documentation.

I am placing trust in your knowledge on the topic or other reviewers here. My only involvement is QA for docs.

polarathene (changes requested):
Is the document title "Folders" accurate? Does this make sense to someone interested in this example?

RFC 6154 makes no mention of "folder" or "directory". Am I right to assume that "folders" is from the user perspective of a mail client?

Perhaps "Custom Mailbox Folders" or similar is better? (_maybe just "Mailbox Folders"?_)

docs/content/examples/uses-cases/folders.md
hnws: @polarathene I think I removed the reference of "tag" everywhere. And now it should only use "Special Use Mailbox" which is the term in RFC. I saw this "tag" usage when I search for the solution for this issue. But yes, it may cause confusion.

docs/content/examples/uses-cases/folders.md
hnws: I wanted to highlight behavior per email client. And also I was trying to tell readers that what would happen if they create this folder later. But since you are also a reader of this page, if you think this is more clear, I am fine with it.

README.md
hnws: I think folder is quite standard when referring to IMAP folders? At least when I was still working as sysadmin few years ago?

docs/content/examples/uses-cases/folders.md
polarathene: The 2nd suggestion just moves the email client names into the common conditions you repeated. The names are all still there and can be found with `ctrl+f` or reading the start of the lines that they apply to.

If italic style for the mail client names is not as good as bold, you could try that?

I think that the reader is someone who is interested in providing this feature with `docker-mailserver` to their users and wants to know the behaviour for different e-mail clients, but doesn't have any specific one in mind? If so reducing the repetition with the 2nd suggested version is better.

README.md
polarathene: Standard as in user / admin choice of word, or official terminology? "Folder" is not the technical term in the RFC correct?

It's fine to mention it as a commonly known name to those that are more used to the user perspective of it. eg: `Mailboxes (_aka Mail Folders_)`

docs/content/examples/uses-cases/folders.md
polarathene: "tag" is still referenced in the current document:

> _To support those clients, you may prefer to keep the tag and folder name identical._

> _keep in mind concerns such as tag support._

If "tag" is a common way to talk about it, like "folder" for mailbox, then just hint that to the reader. eg: `### Special Use Mailbox (_aka Tag_) Support`?

If you do that, then it is clear what "tag" refers to in the quotes to someone who may not know any of this from anywhere else, just reading the example page you are contributing and interested in the feature.

README.md
hnws: Mailbox should cause confusions. You can just google "IMAP folders" and "IMAP Mailbox". I think "folder" is a common and widely used wording in both user / admin.

docs/content/examples/uses-cases/folders.md
hnws: Updated PR with "tags" references removed. It's much less common than "folder".

README.md
polarathene: > Mailbox should cause confusions.

We're referencing an RFC about the feature you're contributing documentation for, there is no mention of "folder" or "tag". Just because a non-technical user may not be familiar with the mailbox term does not make it invalid or causing confusion.

We also link to official Dovecot wiki about what you're documenting, and config files that explicitly have mailbox in the name and file content. So no, this is not a poor choice to prefer.

I understand that some users may be more familiar with a "folder" concept for it instead, and that may be the first thing that comes to mind when performing a search query to learn how to configure such, that adds some bias, especially since they're more likely to use a keyword like "folder" instead of "mailbox" because they presumably don't know the correct term.

I don't mind you using "folder" or "tag" in the documentation, but there should be a very clear indication of the official terms. You can do this as advised by using official terminology like "Mailbox" in the headings and page name (title and filename + mkdocs entry), or "Mailbox folder" for example which covers both keywords.

The headings can provide an alternative user-friendly/familiar name like has been suggested between parentheses: `Mailbox (aka Folder)` or similar. Then the body content can use whatever word you think is better for the reader, so long as the communication is clear.

hnws: Mailbox usually refers to the whole IMAP structure. And folder refers to individual folders..
There is no official term. If you might not know, RFC does not necessarily mean "official" or "standard".

hnws:
I will leave this to my own fork.

README.md
wernerfred: My problem was not the wording of "folder" but what "shared" means by this topic:

> Does this really mean "share" the same folder (between accounts) or "share" the same configuration?

polarathene: > Mailbox usually refers to the whole IMAP structure. And folder refers to individual folders..

Like I said, this isn't something I'm personally familiar with. But between the RFC, the referenced Dovecot docs on the subject and the related config files as examples... The term "Mailbox"/"Mailboxes" is rather evident when referring to this.

- Microsoft seems to refer to them as [IMAP folders](https://support.microsoft.com/en-us/office/choose-which-imap-account-folders-appear-in-outlook-32103602-a568-4499-9d25-54fe236aee6a). Other parts of the docs reference "user mailboxes".
- Apple Mail refers to them as [mailboxes](https://support.apple.com/en-nz/guide/mail/mlhlp1021/mac). They even have the term [mailbox](https://support.apple.com/en-nz/guide/mail/aside/glos8e4eb14f/14.0/mac/11.0) defined with description in their support document.
- [Google references the RFC](https://developers.google.com/gmail/imap/imap-extensions#special-use_extension_of_the_list_command) Special-Use Mailboxes noting it's relevance to "special folders", this is developer documentation though. Looking through consumer docs suggest that it has no concept of mailboxes/folders, the user only manages via labels.. I believe some other clients call this [Virtual Folders](https://en.wikipedia.org/wiki/Virtual_folder#Email_clients).
- Courier seems to [only have a single INBOX mailbox](https://www.cyrusimap.org/imap/concepts/features/automatic-creation-of-mailboxes.html), and describes everything else as sub-folders belonging to it. [`imapd.conf` has a `specialuse_extra` field](https://www.cyrusimap.org/imap/reference/manpages/configs/imapd.conf.html) noting the attributes are per mailbox. A [Personal Namespace](https://www.cyrusimap.org/imap/concepts/features/namespaces.html#imap-features-namespaces-personal) depicts the expected default hierarchy though while mixing mailboxes and sub-folders of a mailbox in the description.

[Wikipedia further communicates](https://en.wikipedia.org/wiki/Internet_Message_Access_Protocol#Multiple_mailboxes_on_the_server) what I was trying to get across:

> _IMAP4 clients can create, rename, and/or delete mailboxes (usually presented to the user as folders) on the server, and copy messages between mailboxes._

LCN has a support article on [IMAP mailbox folders](https://www.lcn.com/support/articles/syncing-your-mailbox-folders-using-imap/), it ments "sent", "drafts", "trash" as standard set of folders for your mailbox. Then lists instructions for various clients, where you'll see different jargon like covered above:

- Windows Live Mail interestingly refers to these "standard folders" as "Special folders on IMAP server".
- Thunderbird has dedicated UI for Sent, Archives, Drafts and Templates and presumably Junk.

---

Having glanced over these resources I link you to, and the [RFC 6154](https://datatracker.ietf.org/doc/html/rfc6154.html#page-3) document again, it's rather clear that a mailbox can be commonly known as representing the Archive, Sent, Trash and Junk "folders".

> There is no official term. If you might not know, RFC does not necessarily mean "official" or "standard".

RFC documents are often what is referenced for implementing protocols that are standardized for interoperability between various software no? IMAP is one of those AFAIK. That does not mandate software use the same terminology in their service or GUI, but it's pretty official on the functional / technical level.

> _IMAP4rev1 permits manipulation of mailboxes (remote message folders) in a way that is functionally equivalent to local folders. - [Source (RFC 3501)](https://datatracker.ietf.org/doc/html/rfc3501)_

Quick look through the document is fairly clear with the mailbox term and it's description. There's also the [LIST command](https://datatracker.ietf.org/doc/html/rfc3501#section-6.3.8) (RFC 3501 - 6.3.8) also further supports that. There's also [section 5.1 on "Mailbox Naming"](https://datatracker.ietf.org/doc/html/rfc3501#section-5.1):

> _The case-insensitive mailbox name INBOX is a special name reserved to mean "the primary mailbox for this user on this server". The interpretation of all other names is implementation-dependent._

---

Do you disagree that the ambiguity to a user or mail client UI warrants being a little more specific than "folders", when that term is not clearly defined in the docs page?

All I've asked is to make it clear what "folder" relates to early on in the document or title.

polarathene: > My problem was not the wording of "folder" but what "shared" means by this topic

@wernerfred from the little reading I have done above and reviewing of this contribution, "shared" is probably not the right word to use.

I am under the impression the various mail clients are intended to implicitly recognize the "Archives" IMAP folder / mailbox and know it's purpose (from the special use attribute assigned to the mailbox).

polarathene:
> I will leave this to my own fork.

You're not going to complete the PR because of a disagreement with the review feedback?

It seemed like a nice contribution, I understand if it's been a hassle for you. I know I've made it more work to revise before accepting, but I have tried to make that easier for you with my own time.

Thank you for the contribution! :smile: 

---

As @hnws is not going to finish this, I think I can take over and get it merged.

hnws:
> > I will leave this to my own fork.
> 
> You're not going to complete the PR because of a disagreement with the review feedback?
> 
> It seemed like a nice contribution, I understand if it's been a hassle for you. I know I've made it more work to revise before accepting, but I have tried to make that easier for you with my own time.
> 
> Thank you for the contribution! 
> 
> As @hnws is not going to finish this, I think I can take over and get it merged.

Yes, disagreement creates different forks. And you need to learn more about sysadmin.

polarathene:
> Yes, disagreement creates different forks.

You're contributing upstream which is welcomed, but it is fair for the reviewers to ask questions and want answers before merging a PR.

You can create a fork and maintain that if you believe it's worth the trade-off for a documentation or config difference (_even when there is no need to do that for what you're doing_). It does not bother anyone here.

> And you need to learn more about sysadmin.

I don't follow? This seems to be a misunderstanding.

You've opted out of finishing the PR because I requested the document have a less vague title than "Folders", and to provide some context early on in the document. [I cited various sources where mailboxes is valid for the topic](https://github.com/docker-mailserver/docker-mailserver/pull/2045#discussion_r655292575) and used by some mail clients, while acknowledging that IMAP/special folders is also used elsewhere as an alternative jargon, mostly from a frontend/client interface.

I did not demand you to replace "folder" throughout the document :confused:

hnws:
Last word on this issue, from **my perspective**, from my **own experience and understanding**, I believe calling IMAP folders "folder" is something without any doubt and discussion. 
image
image

I do not have more time arguing with you here. I might have other changes to this project.
And as a critical infrastructure to my personal life, after the discussion here, I believe it's a better idea to keep my own fork.

casperklein:
> Mailbox usually refers to the whole IMAP structure. And folder refers to individual folders..

I thought the same way. Mail(account/box) is the whole thing. And in that mailbox there are several folders.

Today I learned, after looking into @polarathene answer, that some vendors call the "folders" mailboxes, which I personally find very confusing and never have heard of before.
On the other hand, most MUAs seem to use the folder term.

For me, the terms _folder_, _IMAP folder_ or _Mailbox folder_ are all ok. I think, the majority will know what is meant.

@hnws I can understand, that sometimes, such discussions can be exhausting - I had a few of these myself 😉 But don't get angry because of that. It's not meant bad in anyway.

> I did not demand you to replace "folder" throughout the document

I think this PR is then almost ready to merge. Can we find a compromise in using one of the three (or last two) term suggestions above?

@hnws I would really appreciate, if you reconsider your fork decision and instead contributing to this project directly (this PR and future things).

hnws:
@casperklein 
Thanks for the kind reply here.

For this PR and in general: 
My main frustration here is that as I do not want to explain something which is, again **from my own experience**, is a truth.
I learnt this from my work experience as sysadmin for a few years. And I thought I should find common ground here.

To be honest, maybe it's a bit offensive, but do not this this offensive, I really really put this here at good faith, as email is really a critical path of a person's daily life: If a person does not have better than average knowledge of details of how a email system works, but simply `docker-compose up -d`, this person should not setup self-hosted email system. 
This is a lesson I learnt the hard way myself.

reneploetz:
> > Mailbox usually refers to the whole IMAP structure. And folder refers to individual folders..

> Today I learned, after looking into @polarathene answer, that some vendors call the "folders" mailboxes, which I personally find very confusing and never have heard of before.

The correct term is mailbox as per all IMAP-related RFCs and the server configuration language of dovecot [1]. The RFCs give "remote folders" as alternative name, but from a specification level mailbox is the correct term.
It's also correct that many MDAs use "folder" as name because - I suspect - the name is less confusing for end-users and they generally do not care about using somewhat less well-defined names. Users usually understand what a folder is and less so that they can have multiple "mailboxes" at their (mail) address.

From a mail server adminstration perspective we can decide if we want to go for the dovecot configuration language of using "mailboxes" or use some alternative variant that might be better understandable for end users. Technically, I would prefer using the dovecot configuration language, but i think the whole discussion is really only about semantics and the question what kind of domain/technical knowledge of IMAP we expect users to have.

[1] https://doc.dovecot.org/configuration_manual/namespace/#mailbox-settings

wernerfred:
What i initially meant by my question: https://github.com/docker-mailserver/docker-mailserver/pull/2045#discussion_r654884421 was if "sharing" is the correct verb. 
I read it multiple times and always understood it that multiple clients (using multiple, different account) can share the same IMAP-folder (or however you will call it - imho folder is public spread the most).

But after reading through the discussion and reading it multiple times again i think i now know what @hnws meant: Instead of that every mail client of the _same_ acocount uses its own (local) archive IMAP-folder you can create one which is then kind of "shared" between all clients so comparable to "sent" or "inbox".

The fact that it took me a few readings and days and that "sharing" IMAP folders among all clients connected/logged-in with the same account is basically the definition of IMAP folders still brings me to the point that "sharing" might be completely missleading or am i the only one thinking in that direction?

Maybe a simple "using the same folder (on the server)" is the better choice.

----

All in all I can only say that i agree 100% with the answer of @casperklein. It would be very disappointing to loose you with your knowledge as a contributor. I am also 100% sure that @polarathene didnt intended to offend you in any way, just adding his 2 cents by throwing in his thoughts with the goal to improve the docs.

I would really be happy to see you finishing this PR 🚀

polarathene:
@wernerfred 

> Instead of that every mail client of the _same_ acocount uses its own (local) archive IMAP-folder you can create one which is then kind of "shared" between all clients so comparable to "sent" or "inbox".

My understanding is that these are intended as remote IMAP folders / mailboxes on the MDA (Dovecot for us), and that any MUA (mail client) should be aware of these. Compatibility with some features of IMAP extensions may vary, which the documentation page tries to make the reader aware of.

> brings me to the point that "sharing" might be completely missleading or am i the only one thinking in that direction?

Yes, I took some time to rephrase the document myself. I think it's good enough for merging and keeps both "mailboxes" and "folders" wording, making it clear early on that they're the same and these are IMAP folders.

The standard / common mailboxes are considered special IMAP folders on some clients due to the `SPECIAL-USE` attribute or functionality that differs them from regular folders a user may user for organizing their mail.

I did not touch on it, but there is also virtual folders which are like aliases AFAIK, eg for things like labels functionality, an example is for marking something as important (starred/favourite in some MUA).

> Maybe a simple "using the same folder (on the server)" is the better choice.

According to the information provided regarding various clients, some may behave differently and only use a folder locally (at least for archiving functionality).

polarathene:
> I am also 100% sure that @polarathene didnt intended to offend you in any way, just adding his 2 cents by throwing in his thoughts with the goal to improve the docs.

I absolutely had no intention to offend. I don't have the experience that @hnws has as a sysadmin, but tried to express that our docs should try avoid being vague on terms that vary based on context.

I tried to highlight that with the Dovecot docs choice of "mailboxes", which the RFC uses and Apple mail clients UI too. I've always personally thought of them as mail folders myself and from a user perspective I can understand the bias/preference to that usage.

However we reference it in relation to Dovecot and these are intended as technical docs for setting up an email service, I don't see any major reason to debate what is used, other than clearly defining they're the same thing early on. That's now taken care of.

---

@hnws

I'm sorry if I come off as difficult, I don't mean to! :sweat_smile: I am just providing feedback during review, I welcome to be corrected if you disagree, but I would appreciate adequate justification to back it up, or waiting on other maintainers to throw in their opinion.

Again, I was not trying to imply you were wrong about anything, but try to come to a common understanding based on the information available.

> And as a critical infrastructure to my personal life, after the discussion here, I believe it's a better idea to keep my own fork.

I wouldn't want to make you feel that you have to fork due to our interaction here.

I'm not representative of the other maintainers, I just care about the quality of the docs and tried to collaborate with you on your contribution. I hope you can understand and appreciate that, but if it would make you comfortable to contribute in future, I can avoid being a reviewer on any future PRs of yours?

polarathene:
Ready for review. I tried looking into the "build-and-test" failure, but that error message isn't too helpful.

polarathene (approved):

wernerfred:
@all-contributiors add @hnws for docs

wernerfred:
@all-contributors add @hnws for docs

PR #2051, opened 2021-06-22
Use keyserver that still returns keys with user IDs

### LINKED ISSUE #2050: [BUG] unstable build dependending on keys.gnupg.net
# Bug Report

## Context

Building the image on our CI fails some days when keys.gnupg.net doesn't return the key for 0x683BF1BEBD0A882C

Building the image

### When does this occur?

Sometimes (e.g. today, 2021-06-22), it also happened a few weeks ago but was resolved during the day

1. Run `gpg --keyserver keys.gnupg.net --recv-keys "0x683BF1BEBD0A882C"`

### Actual Behavior
gpg just returns 
`gpg: keyserver receive failed: No name`

### Expected Behavior
gpg should receive the key

## Your Environment

- version: `master`
- available RAM: `128GB`
- Docker version: `20.10.7`
- docker-compose version: `1.29.2`

ap-wtioit:
# Description

Use keyserver that still returns keys with user IDs

keys.gnupg.net doesn't return user IDs (without approval) and therefore gpg
doesn't accept keys from it.

Fixes #2050 

## Type of change

## Checklist:

FYI @wt-io-it

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ ARG DEBIAN_FRONTEND=noninteractive
 ARG FAIL2BAN_DEB_URL=https://github.com/fail2ban/fail2ban/releases/download/0.11.2/fail2ban_0.11.2-1.upstream1_all.deb
 ARG FAIL2BAN_DEB_ASC_URL=${FAIL2BAN_DEB_URL}.asc
 ARG FAIL2BAN_GPG_PUBLIC_KEY_ID=0x683BF1BEBD0A882C
-ARG FAIL2BAN_GPG_PUBLIC_KEY_SERVER=keys.gnupg.net
+ARG FAIL2BAN_GPG_PUBLIC_KEY_SERVER=hkps://keyserver.ubuntu.com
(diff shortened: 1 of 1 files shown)


wernerfred (approved):
LGTM if the pipeline passes.

Thanks for your contribution and the very detailed description of your approach in the linked Issue. Helps a lot keeping track and follow along with your thoughts.

We used keys.gnupg.net as the fail2ban maintainers referenced this server when we asked for their public key. Good finding that you discovered a (only sometimes appearing?) bug in this implementation.

Thanks

georglauterbach (approved):
Good catch!

casperklein:
> keys.gnupg.net doesn't return user IDs (without approval)

I stumbled upon the same yesterday and hoped that was only temporary. For completeness: It's not about a missing feature, the host doesn't even resolve to an IP address anymore.

casperklein:
I just found out, that the DNS record for the server pool was intentionally removed.

> 
> This service is deprecated. This means it is no longer maintained, and new HKPS certificates will not be issued. Service reliability should not be expected.
> 
> Update 2021-06-21: Due to even more GDPR takedown requests, the DNS records for the pool will no longer be provided at all.

Source: https://sks-keyservers.net/status/

`keys.gnupg.net` was a CNAME for `hkps.pool.sks-keyservers.net`

PR #2080, opened 2021-07-09
chore(encryption): support for passworded user email storage encryption keys

NorseGaud:
# Description

1. Allows users to `ENABLE_PER_USER_STORAGE_ENCRYPTION` in their docker-compose to enable the feature
2. Within the various password impacting scripts:
    - Notice mail_crypt is enabled and, by default, generated a key for the user on creation.
    - Notice mail_crypt is enabled and, by default, update encryption password when email password is changed.
    - Notice mail_crypt is enabled and, by default, when updating the user's password, create user keys/encryption if there aren't any yet.
2. Allow users to change the mail_crypt_curve with `PER_USER_STORAGE_ENCRYPTION_CURVE` and default to `secp521r1`
3. Allow users to change the scheme with `PER_USER_STORAGE_ENCRYPTION_SCHEME` and default to `CRYPT`

Related issue: https://github.com/docker-mailserver/docker-mailserver/issues/2058

Related: https://github.com/docker-mailserver/docker-mailserver/pull/2134

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Improvement (non-breaking change that does improve existing functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [x] If necessary I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #2097, opened 2021-07-23
Work around possible problems with init

### LINKED ISSUE #763: Image cannot be built after update & upgrade
Hi,

I'm using this image as base image for our own uses. Basically we just install few additional packages and add few files into the image.

After update & upgrade, the image fails to build because of this error:

```
Setting up postfix (3.1.0-3ubuntu0.2) ...
Installing new version of config file /etc/init.d/postfix ...

Postfix configuration was untouched.  If you need to make changes, edit
/etc/postfix/main.cf (and others) as needed.  To view Postfix configuration
values, see postconf(1).

After modifying main.cf, be sure to run '/etc/init.d/postfix reload'.

Running newaliases
postalias: fatal: unsupported dictionary type: texthash. Is the postfix-texthash package installed?
dpkg: error processing package postfix (--configure):
 subprocess installed post-installation script returned error exit status 1
dpkg: dependency problems prevent configuration of postfix-ldap:
 postfix-ldap depends on postfix (= 3.1.0-3ubuntu0.2); however:
  Package postfix is not configured yet.

dpkg: error processing package postfix-ldap (--configure):
 dependency problems - leaving unconfigured
Processing triggers for libc-bin (2.23-0ubuntu9) ...
Processing triggers for ca-certificates (20170717~16.04.1) ...
Updating certificates in /etc/ssl/certs...
17 added, 42 removed; done.
Running hooks in /etc/ca-certificates/update.d...
done.
Processing triggers for systemd (229-4ubuntu21) ...
Errors were encountered while processing:
 postfix
 postfix-ldap
E: Sub-process /usr/bin/dpkg returned an error code (1)
```

Affected tags: 2.1, 2.3, latest

This might have something to do with pull request #382

**How to reproduce**

Build using following dockerfile

```
FROM tvial/docker-mailserver:2.3

RUN DEBIAN_FRONTEND=noninteractive apt-get update -q  && \
	apt-get -y upgrade
```

Craeckie:
# Description

When building it I ran into [this](https://github.com/docker-mailserver/docker-mailserver/issues/763#issuecomment-343751893). It's some problem with init not working as expected. It seems to appear only for some people in some cases.

The code from [stackoverflow](https://stackoverflow.com/a/48782486) (almost) fixes the problem. Just a `-e` was missing.

So this PR also fixes #763

## Type of change

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,7 +46,7 @@ RUN \
   apt-get -qq dist-upgrade && \
   apt-get -qq install postfix && \
   # Work around possible problems with init
-  echo -e "#!/bin/sh\nexit 0" > /usr/sbin/policy-rc.d && \
+  echo -e '#! /bin/sh\nexit 0\n' >/usr/sbin/policy-rc.d && \
   apt-get -qq --no-install-recommends install \
   # A - D
   altermime amavisd-new apt-transport-https arj binutils bzip2 bsd-mailx \
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,6 +45,8 @@ RUN \
   apt-get -qq install apt-utils 2>/dev/null && \
   apt-get -qq dist-upgrade && \
   apt-get -qq install postfix && \
+  # Work around possible problems with init
+  echo -e "#!/bin/sh\nexit 0" > /usr/sbin/policy-rc.d && \
   apt-get -qq --no-install-recommends install \
   # A - D
   altermime amavisd-new apt-transport-https arj binutils bzip2 bsd-mailx \


casperklein:
Please provide some more information and how to re-produce that problem. Do you install additional packages as in the linked issue? 
There are currently no known issues about failing builds (on supported platforms).

georglauterbach:
> Please provide some more information and how to re-produce that problem. Do you install additional packages as in the linked issue?
> There are currently no known issues about failing builds (on supported platforms).

I agree. Please elaborate :)

Craeckie:
Sorry for the late reply, been quite busy the last few days.

Ok, after quite some time looking for the error it turns out it wasn't apt at all. The command `gpg --keyserver ${FAIL2BAN_GPG_PUBLIC_KEY_SERVER} --recv-keys ${FAIL2BAN_GPG_PUBLIC_KEY_ID}` fails, but since the output was suppressed I always assumed apt was causing it. I will further investigate when I have time for it.

What helped me a lot was activating the command output with `set -x` right at the beginning of RUN. This makes searching for the error much easier! You should really add that to the Dockerfile. Should I create a PR for this?

georglauterbach:
> [...] The command `gpg --keyserver ${FAIL2BAN_GPG_PUBLIC_KEY_SERVER} --recv-keys ${FAIL2BAN_GPG_PUBLIC_KEY_ID}` fails, but since the output was suppressed I always assumed apt was causing it. I will further investigate when I have time for it.

How was the output suppressed? The full command reads:

``` BASH
gpg --keyserver ${FAIL2BAN_GPG_PUBLIC_KEY_SERVER} --recv-keys ${FAIL2BAN_GPG_PUBLIC_KEY_ID} 2>&1

Craeckie:
> How was the output suppressed? The full command reads:
> 
> ```shell
> gpg --keyserver ${FAIL2BAN_GPG_PUBLIC_KEY_SERVER} --recv-keys ${FAIL2BAN_GPG_PUBLIC_KEY_ID} 2>&1
> ```
> 
> which only prints `stderr` on `stdout`, but does not remove any output AFAIK...

I was trying to build v10.0.0, where the output was suppressed:
https://github.com/docker-mailserver/docker-mailserver/blob/a0f4a375124b3b3402d5639a7d422501c2011b97/Dockerfile#L64-L65

georglauterbach:
I see.

@wernerfred maybe we should release `v10.0.1` as there were enough small fixes, including the F2B GPG server. What do you think?

wernerfred:
As far as releasing means feature freeze and testing :edge for at least 7 days I'm fine with it. 
Should i prepare the release PR like last time today?

georglauterbach:
> As far as releasing means feature freeze and testing :edge for at least 7 days I'm fine with it.
> Should i prepare the release PR like last time today?

SGTM 👍🏼

casperklein:
There were also changes that are not bugfixes/patches. Therefore 10.1.0 should be released.

wernerfred:
Then we need to decide whether to wait for #2080 or not ;)

georglauterbach:
I don’t think we need to wait. We can always release `v10.2.0` afterwards if you ask me.

Issue #2098, opened 2021-07-24
check-for-changes.sh performance concerns on NFS


NorseGaud:
# Subject

https://github.com/docker-mailserver/docker-mailserver/issues/2096 sparked my interest in this. The addmailuser script seems to be a bit slow for some users (NFS mostly):

> I notice that when calling addmailuser it becomes very slow when number of mailboxes increases.
>
> I did not notice slowness too much when using non-NFS e.g. local HDD or AWS EBS storage
>
> Now I am using NFS (actually AWS EFS) storage since deploying as docker in AWS ECS.
>
> For example I have 2500 mailboxes and a call to addmailuser takes 60 seconds.

Ok, so this is understandable due to the following code I added in one of my last PRs:

> https://github.com/docker-mailserver/docker-mailserver/blob/45345b2f49ba52b6ef9efe6492fa3b99bd4f89ae/target/bin/addmailuser#L68-L75

This code assists with NFS or other slow volumes in use by the mail server. We could debate why slow volumes are being used, but I'll save that for another ticket :)

So, with all of this said, it doesn't actually seem as if slow volumes are actually the main cause of the root problem. The root problem seems to be that `check-for-changes.sh` takes much too long to update what it needs to. I therefore did some digging in a forked branch: https://github.com/NorseGaud/docker-mailserver/commits/check-for-changes-performance

Here are the results while running a modified `check-for-changes.sh` that outputs run time and STDOUT/ERR:

---

1. Adding emails one by one using `for n in {1..10}; do ./usr/local/bin/addmailuser test${n}@pierce.us XXXXX; done` in the running container in my production setup using NFS. The `check-for-changes.sh` was only modified to show the run time, and no other changes were made:

```
root@ip-172-31-3-247:/# ./usr/local/bin/check-for-changes.sh
DONE TOTAL RUNTIME SECONDS: 0
DONE TOTAL RUNTIME SECONDS: 0
DONE TOTAL RUNTIME SECONDS: 0
DONE TOTAL RUNTIME SECONDS: 0
DONE TOTAL RUNTIME SECONDS: 0
DONE TOTAL RUNTIME SECONDS: 0
---------- Created test1@pierce.us ----------------------
[ WARNING ]  File not found for certificate in check_for_changes.sh
postfix: stopped
postfix: started
dovecot: stopped
dovecot: started
DONE TOTAL RUNTIME SECONDS: 6
---------- Created test2@pierce.us ----------------------
[...]

PR #2095, opened 2021-07-22
Update bats to latest stable version (1.4.1)

NorseGaud:
# Description

Updating bats to 1.4.0 (should be released in a few days); testing with 1.3.0 now. We currently use 1.2.1.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Improvement (non-breaking change that does improve existing functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [x] If necessary I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #2104, opened 2021-07-29
check-for-changes: performance improvements + wait for settle

NorseGaud:
# Description

See: https://github.com/docker-mailserver/docker-mailserver/issues/2098

This PR will:

- Have `check-for-changes.sh` wait for changes to "settle" (checksum comparison has to not have changed over two different checks with 5 second gaps between) before modifying and restarting anything
- Added backgrounding of processes and using wait throughout instead of sleeps for more efficiency
- Removed the while loop from addmailuser due to check-for-changes.sh logic changes

This is, in my opinion, critically needed for an API to be performing changes -- and lots of them all at once! See https://github.com/docker-mailserver/docker-mailserver-admin/issues/3 for more discussion about this.

## Type of change

## Checklist:

- [] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)


NorseGaud:
Forgot to draft this... again... Sorry... It's still a WIP

NorseGaud:
It looks like both the CI and my local machine are getting the following (running with new verbose flag):

```
✗ begin 186 checking tls: cipher list - rsa intermediate
   (from function `assert_success' in file test/test_helper/bats-assert/src/assert.bash, line 114,
    from function `collect_cipherlist_data' in file test/security_tls_cipherlists.bats, line 154,
    from function `check_ports' in file test/security_tls_cipherlists.bats, line 86,
    in test file test/security_tls_cipherlists.bats, line 47)
     `check_ports 'rsa' 'intermediate'' failed
   $ docker run -d --name tls_test_cipherlists --volume /Users/norsegaud/docker-mailserver/test/duplicate_configs/security_tls_cipherlists.bats/:/tmp/docker-mailserver/ --volume /Users/norsegaud/docker-mailserver/test/test-files/ssl/example.test/:/config/ssl/:ro --env DMS_DEBUG=0 --env ENABLE_POP3=1 --env SSL_TYPE=manual --env SSL_CERT_PATH=/config/ssl/cert.rsa.pem --env SSL_KEY_PATH=/config/ssl/key.rsa.pem --env TLS_LEVEL=intermediate --network test-network --network-alias example.test --hostname mail.example.test --tty mailserver-testing:ci
     f7d78f74fc3eedaa1277f63d9b7a2d06f22fbae7705f75b842b56bed71fce40b
   [ TASKLOG ]  mail.example.test is up and running
   $ docker run --rm --user 501:20 --network test-network --volume /Users/norsegaud/docker-mailserver/test/test-files/ssl/example.test/:/config/ssl/:ro --volume /var/folders/vb/pjyhgj4s491_k0t025_j56y00000gn/T//results/rsa/intermediate/:/output --workdir /output drwetter/testssl.sh:3.1dev --quiet --file /config/ssl/testssl.txt --mode parallel --overwrite --preference
     standard_init_linux.go:228: exec user process caused: permission denied
   
   -- command failed --
   status : 1
   output : standard_init_linux.go:228: exec user process caused: permission denied
   --
```

It seems like `--user 501:20` is the issue as it runs without it. The comment says 

>  # `--user "<uid>:<gid>"` is a workaround: Avoids `permission denied` write errors for json output, uses `id` to match user uid & gid.

@polarathene, I'm going to see if anything else fails after removing it but I wanted to see if it's ok if it's removed (as long as everything else passes).

polarathene:
> --user 501:20 --network test-network --volume /Users/norsegaud/

That is odd looking uid and gid values, and with that volume is this Windows/WSL? That context seems to be missing in the CI tests (due to lack of verbose flag I guess?)

But this other volume looks more of an issue:

> --volume /var/folders/vb/pjyhgj4s491_k0t025_j56y00000gn/T//results/rsa/intermediate/:/output

I'm not familiar with this mount, but that `T//results` looks a bit broken path wise? I don't think it should have two `/` in a row like that?

polarathene:
I believe whatever changes have been introduced have broken the functionality of `${BATS_TMPDIR}` providing a valid path prefix here:

https://github.com/docker-mailserver/docker-mailserver/blob/f33374023f94be9a6daf2093f3cf0bdbf8d1d7aa/test/security_tls_cipherlists.bats#L22-L23

NorseGaud:
Thanks @polarathene , appreciate your insite. So, I rolled bats back to 1.3.0 and it works. It looks like 1.4.0 introduced a change to BATS_TMPDIR: https://github.com/bats-core/bats-core/pull/410#discussion_r575864351

How the hell this didn't fail in https://github.com/docker-mailserver/docker-mailserver/pull/2095 is beyond me... Might be good to have @casperklein and @georglauterbach take a look if they get a few minutes. I'm unsure about how the CI is all set up TBH.

polarathene:
> How the hell this didn't fail in #2095 is beyond me...

I'd have to refresh my memory but I recall PRs can behave differently depending on what you're doing. Secrets aren't usually available for example, and in one case I think you end up with the default branch sha/ref depending on what you're doing. 

Perhaps the tests didn't get run on that PR with the updated submodule? And only after merging to the default branch does that apply? I know when setting up CI for the docs there were a few gotchas I ran into like this :\

NorseGaud:
Please ignore the bats submodule for now. I'm using that to gain visibility into the tests and will revert it soon.

georglauterbach:
Is this WIP?

NorseGaud:
@georglauterbach , no it's ready for review

georglauterbach (approved):
LGTM 👍🏼

georglauterbach:
@NorseGaud when you push yourself, my review will be dismissed I think. But @docker-mailserver/maintainers can update the branch via the webUI which _should not_ dismiss it (if that is somehow important) :D

test/open_dkim.bats
```diff
@@ -21,7 +21,7 @@ function setup_file
     --name "${CONTAINER_NAME}" \
 		--cap-add=SYS_PTRACE \
 		-v "${PRIVATE_CONFIG}":/tmp/docker-mailserver \
-		-v "${CDIR}/test/test-files":/tmp/docker-mailserver-test:ro \
+		-v "$(pwd)/test/test-files":/tmp/docker-mailserver-test:ro \
```
williamdes, line 24:
```suggestion
		-v "$PWD/test/test-files":/tmp/docker-mailserver-test:ro \
```

NorseGaud: We use $(pwd) in a ton of other tests. What's the reason to use $PWD here instead?

williamdes: Oh no, it was for me more practical then calling a binary
But if tests already use it then its perfect that way ;)

NorseGaud: Roger that -- Within the tests/* directory I see 60 occurrences of `$(pwd)` :)

georglauterbach:
LGTM 👍 

@casperklein or @wernerfred feel free to merge this PR if you approve it :)

georglauterbach (approved):

georglauterbach:
I will go ahead and merge this :)

casperklein:
I hadn't time to review this yet. Otherwise I had approved it..

wernerfred:
Dito ⌛

target/supervisor/conf.d/supervisor-app.conf
```diff
@@ -108,7 +108,7 @@ command=/usr/bin/fetchmail -f /etc/fetchmailrc -v --nodetach --daemon %(ENV_FETC
 
 [program:postfix]
 startsecs=0
-stopwaitsecs=55
+stopwaitsecs=20
```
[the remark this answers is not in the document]
NorseGaud: https://github.com/docker-mailserver/docker-mailserver/pull/2137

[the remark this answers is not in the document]
NorseGaud: If you find any others, let me know and I'll PR them. Maybe we could add a little comment in the supervisor file about the dependency? Or maybe add a test?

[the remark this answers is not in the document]
wernerfred: Add a test seems appropriate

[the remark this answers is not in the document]
casperklein: > If you find any others, let me know and I'll PR them

Yes, I think, I will fully review this PR until next week end. That was just something, that immediately kept my attention.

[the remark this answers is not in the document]
casperklein: A comment and/or test is good. Regarding a test, I don't know how to test this behaviour properly. Quick test would just be, to grep and see if "stopwaitsecs=55" is present or not.

target/scripts/postfix-wrapper.sh
```diff
@@ -20,11 +20,11 @@ trap "service postfix stop" SIGINT
 trap "service postfix stop" SIGTERM
 trap "service postfix reload" SIGHUP
 
-service postfix start
-sleep 5
+service postfix start &
+wait
```
casperklein, line 24: I think the `sleep` here was, because `service postfix start` exit (0) successfully pretty fast, however postfix needs some time to come up completely.

```
service postfix start &
wait
```

is equal to just `service postfix start` ?

I am not sure if it's related, but I read on a issue/PR this week, about our CI behaving "flacky" again. Maybe this is the reason?

target/scripts/check-for-changes.sh
```diff
@@ -50,195 +53,231 @@ do
...
   # get chksum and check it, no need to lock config yet
   _monitored_files_checksums >"${CHKSUM_FILE}.new"
-  cmp --silent -- "${CHKSUM_FILE}" "${CHKSUM_FILE}.new"
-  # cmp return codes
-  # 0 – files are identical
-  # 1 – files differ
-  # 2 – inaccessible or missing argument
-  if [ $? -eq 1 ]
+  CMP_RESULT=$(cmp -- "${CHKSUM_FILE}" "${CHKSUM_FILE}.new")
+  if [[ -n "${CMP_RESULT}" ]] # If there is a difference between checksum files
```
casperklein, line 57: I know it was there before, but do some know what the `--` are for? I couldn't find something in the man page.

NorseGaud: > A double-dash in a shell command signals the end of options and disables further option processing.

georglauterbach: Commonly used for example when you're trying to `grep` something that starts with a double dash, you'd write

``` BASH
grep -E -- "--to-grep" $FILE
```

casperklein: I knew that. I was wondering, if in that case, it has another purpose. Because it seems to me unnecessary here.

target/scripts/postfix-wrapper.sh
NorseGaud: Good points -- I'll play around with it and see what I find.

target/scripts/check-for-changes.sh
NorseGaud: I never use it personally :)

target/scripts/postfix-wrapper.sh
```diff
@@ -20,11 +20,11 @@ trap "service postfix stop" SIGINT
...
 
 # wait until postfix is dead (triggered by trap)
 while kill -0 "$(< /var/spool/postfix/pid/master.pid)"
 do
-  sleep 5
+  sleep 1
 done
```
casperklein, line 30: Inspired by your other changes, I wonder if this one liner would work too:

`wait $(</var/spool/postfix/pid/master.pid)`

casperklein: Unfortunately, `wait` only works on child processes of the current shell:

```
# wait $(</var/spool/postfix/pid/master.pid)
bash: wait: pid 3108 is not a child of this shell 
```

https://stackoverflow.com/a/41613532/568737 suggests `tail` instead (which works):

`tail -f --pid="$(</var/spool/postfix/pid/master.pid)" /dev/null`

We could replace the while loop with that.

NorseGaud: Sounds good -- including in https://github.com/docker-mailserver/docker-mailserver/pull/2134

casperklein: I am unsure. This could lead to problems on MacOS? I don't have a mac to test, but the stackoverflow posts suggest this only for linux.
On the other hand, we don't support MacOS officially. What do you think @docker-mailserver/maintainers?

NorseGaud: The code doesn't run on mac, it'll run inside of the linux contianer. Shouldn't be an issue.

casperklein: You are absolutely right.

Issue #1412, opened 2020-02-25
en.internet.nl says mail servers supports insufficiently secure parameters for DH key exchange


Rillke:
> At least one of your mail servers supports insufficiently secure parameters for Diffie-Hellman key exchange. 

I haven't manually changed any Ciphers or DH KEX Parameters, use the defaults the Image generated/ specified, however I use Let's Encrypt (if this matters).

Configuration:
```
OVERRIDE_HOSTNAME=
DMS_DEBUG=1
ONE_DIR=1
POSTMASTER_ADDRESS=
PERMIT_DOCKER=
# empty => modern
# modern => Enables TLSv1.2 and modern ciphers only. (default)
# intermediate => Enables TLSv1, TLSv1.1 and TLSv1.2 and broad compatibility ciphers.
# old => NOT implemented. If you really need it, then customize the TLS ciphers overriding postfix and dovecot settings
#        (https://github.com/tomav/docker-mailserver/wiki/)
TLS_LEVEL=
SPOOF_PROTECTION=1
ENABLE_SRS=0
ENABLE_POP3=
ENABLE_CLAMAV=1
ENABLE_FAIL2BAN=1
ENABLE_MANAGESIEVE=
POSTSCREEN_ACTION=enforce
SMTP_ONLY=
# Please read [the SSL page in the wiki](https://github.com/tomav/docker-mailserver/wiki/Configure-SSL) for more information.
#
# empty => SSL disabled
# letsencrypt => Enables Let's Encrypt certificates
# custom => Enables custom certificates
# manual => Let's you manually specify locations of your SSL certificates for non-standard cases
# self-signed => Enables self-signed certificates
SSL_TYPE=letsencrypt
VIRUSMAILS_DELETE_DELAY=
ENABLE_POSTFIX_VIRTUAL_TRANSPORT=
POSTFIX_DAGENT=
POSTFIX_MAILBOX_SIZE_LIMIT=512000000
POSTFIX_MESSAGE_SIZE_LIMIT=10240000
REPORT_RECIPIENT=0
REPORT_SENDER=
REPORT_INTERVAL=daily
ENABLE_SPAMASSASSIN=1
SA_TAG=2.0
SA_TAG2=6.31
SA_KILL=6.31
SA_SPAM_SUBJECT=***SPAM*****
ENABLE_FETCHMAIL=0
FETCHMAIL_POLL=300
ENABLE_LDAP=
LDAP_START_TLS=
[...]

PR #2116, opened 2021-08-11
chore(image): Update Debian base image to Debian 11 "Bullseye"

georglauterbach:
# Description

This PR updates the base image to Debian 11 "Bullseye". Debian 11 is going to be released on the 14th of August, 2021. We can already go ahead and see what our CI pipeline says about this change to prepare for the update.

## Type of change

- [x] Possibly breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:


williamdes (approved):

georglauterbach:
We are having a problem here:

``` BASH
georg@swift : ~ $ docker run --rm -e MAIL_USER=user1@localhost.localdomain -e MAIL_PASS=mypassword -t mailserver/docker-mailserver:10.0.0 /bin/sh -c 'echo "$MAIL_USER|$(doveadm pw -s SHA512-CRYPT -u $MAIL_USER -p $MAIL_PASS)"'
user1@localhost.localdomain|{SHA512-CRYPT}$6$q2KhCz6qOE3rMN4X$V62CfNAogbNnuxWFzEKngZsSgKrH2N01BccXda1HDJ1xVyCrrvoU/Rnw1gZWmU.v2S7ssddmPO33YCpYCC6o/.
georg@swift : ~ $ docker run --rm -e MAIL_USER=user1@localhost.localdomain -e MAIL_PASS=mypassword -t mailserver/docker-mailserver:10.1.0 /bin/sh -c 'echo "$MAIL_USER|$(doveadm pw -s SHA512-CRYPT -u $MAIL_USER -p $MAIL_PASS)"'
user1@localhost.localdomain|{SHA512-CRYPT}$6$zveXBRrFJ2I5/DZ6$rYr/wXx2xxRoeYLjtQNYIHFNWXdMc2lB8P7qsPJWUPnnEeb2TDiD3zQxJmrekTVBDtw6HzAK17TdhEJemJ2KN1
georg@swift : ~ $ docker run --rm -e MAIL_USER=user1@localhost.localdomain -e MAIL_PASS=mypassword -t mailserver-testing:ci /bin/sh -c 'echo "$MAIL_USER|$(doveadm pw -s SHA512-CRYPT -u $MAIL_USER -p $MAIL_PASS)"'
doveconf: Fatal: Error in configuration file /etc/dovecot/conf.d/10-ssl.conf line 50: ssl_dh: Can't open file /etc/dovecot/dh.pem: No such file or directory
user1@localhost.localdomain|
```

For some reason, `etc/dovecot/dh.pem` is missing when executing `make generate-accounts` when using Debian 11. This is most likely something to do with a new Dovecot version???

casperklein:
> PS: @casperklein I cannot find such a tag ATM at Docker Hub...

I meant in general. i haven't checked if there is already a tag for that. Probably once v11 is official released.

casperklein:
As I am always running the latest Dovecot version, I had to patch at least two scripts some time ago. See https://github.com/docker-mailserver/docker-mailserver/commit/0c4d61030cd9ad80201160c519e8422d2e602903.

georglauterbach:
Sadly, this does not circumvent the `Makefile`'s `generate-accounts` target's problem because it uses `doveadm` directly.

NorseGaud:
@casperklein, I could be wrong but here is how I solved that in my fork:

https://github.com/NorseGaud/docker-mailserver/pull/21/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557

casperklein:
> COPY target/shared/ffdhe4096.pem /etc/dovecot/dh.pem

I thought about the same back then, but for one reason (I can't remember right now) I decided to not fix it that way.

georglauterbach:
I thought about exactly the same, but I cannot figure out why this problem now comes up in the first place...

I mean, if this fixes it, why not. But finding the real root cause would be nice.

casperklein:
> I mean, if this fixes it, why not.

Of course. But fixing the CI could break other functionality. For a quick shot to see what happens, okay. But before finaly including the COPY in the Dockerfile, possible site effects should be examined. There are some condition checks in `setup-stack.sh`.

georglauterbach:
We see that CI is now running, but multiple tests are broken with the new image / new `COPY` command.

polarathene:
> For some reason, `etc/dovecot/dh.pem` is missing when executing `make generate-accounts` when using Debian 11. This is most likely something to do with a new Dovecot version???

I noticed in the [release notes for Dovecot 2.3](https://wiki2.dovecot.org/Upgrading/2.3) that dh params became optional. That release was end of 2017 though and AFAIK in Debian 10/Buster.

> ssl-parameters.dat file is now obsolete. You should use ssl_dh setting instead: ssl_dh=</etc/dovecot/dh.pem
> - You can convert an existing ssl-parameters.dat to dh.pem:
> dd if=/var/lib/dovecot/ssl-parameters.dat bs=1 skip=88 | openssl dhparam -inform der > /etc/dovecot/dh.pem
> ssl-params process has also been removed, as it is no longer used to generate these parameters.
> You are encouraged to create at least 2048 bit parameters. 4096 is industry recommendation.

> Since v2.3.3+ DH parameter usage is **optional** and can be omitted. You are invited to amend ciphers to disallow non-ECC based DH algorithms, but if you don't and someone does try to use them, error will be emitted.

That's also reflected for the config option in `10-ssl.conf` that we have:

https://github.com/docker-mailserver/docker-mailserver/blob/65622c56e934459beef7fba045f23ef897a6dcda/target/dovecot/10-ssl.conf#L47-L50

If we have been relying on it to exist previously, either it was removed in a later patch release or Buster package may have had some changes from upstream that provided it with Bullseye dropping that patch.

You'll also notice from that wiki on dovecot release linked earlier, that some defaults were changed. Our related configs maintained in this repo I imagine like the Fail2Ban config have fallen a bit out of date/sync. For example the `mdbox_rotate_size` default was changed from 2M to 10M, our config doesn't mess with that but the commented default does indicate that we might want to sync back with upstream? In this case LMTP variant is also a bit out of sync with our main config too IIRC.

https://github.com/docker-mailserver/docker-mailserver/search?q=mdbox_rotate_size

georglauterbach:
> we might want to sync back with upstream? 

Ideally, yes. I just don't have the time to cover everything :)

---

> AFAIK, there isn't any reason to keep the params configured by default, at least for the cipher suites the majority of clients (at least modern ones) should have support for ciphersuites that don't need DH params. Users could use a patch script like we usually advise if they require it, or the non-modern mode for SSL could enable using dhparams from the ones we ship in the repo.

I absolutely agree. We should therefore drop this, right?

casperklein:
> like the Fail2Ban config have fallen a bit out of date/sync. 

How do you come to that conclusion? Especially fail2ban is up to date, as we don't use the outdated version from the Debian repo.

polarathene:
> How do you come to that conclusion?

That was bad phrasing by me in the early hours of the morning before going to sleep :sweat_smile: 

I meant to reference prior to the PR you did, where we had used an old F2B config which had become outdated.

polarathene:
> I absolutely agree. We should therefore drop this, right?

Yeah, we can just comment it out in the dovecot config AFAIK. I'm not sure if it's used anywhere else in this repo, if it is we might want to keep it for legacy use?

georglauterbach:
I have commented it out. Let's see what happens.

georglauterbach:
Alright: We're now seeing

1. Check for changes does something weird. @NorseGaud could you please have a look? :)
2. Some cipher-list problems. @polarathene could you please have a look? :)

polarathene:
> 2\. Some cipher-list problems. @polarathene could you please have a look? :)

```
checking tls: cipher list - rsa intermediate
# -- output differs --
# expected : "ECDHE-RSA-AES128-SHA ECDHE-RSA-AES256-SHA DHE-RSA-AES128-SHA DHE-RSA-AES256-SHA"
# actual   : "ECDHE-RSA-AES128-SHA ECDHE-RSA-AES256-SHA"
# --

checking tls: cipher list - rsa modern
# -- output differs --
# expected : "ECDHE-RSA-AES128-GCM-SHA256 ECDHE-RSA-AES256-GCM-SHA384 ECDHE-RSA-CHACHA20-POLY1305 DHE-RSA-AES128-GCM-SHA256 DHE-RSA-AES256-GCM-SHA384"
# actual   : "ECDHE-RSA-AES128-GCM-SHA256 ECDHE-RSA-AES256-GCM-SHA384 ECDHE-RSA-CHACHA20-POLY1305"
# --

checking tls: cipher list - ecdsa intermediate
# -- output differs --
# expected : "ECDHE-ECDSA-AES128-SHA ECDHE-ECDSA-AES256-SHA"
# actual   : 
# --

ok 189 checking tls: cipher list - ecdsa modern

checking tls: cipher list - ecdsa intermediate, with rsa fallback
# -- output differs --
# expected : "ECDHE-ECDSA-AES128-SHA ECDHE-RSA-AES128-SHA ECDHE-ECDSA-AES256-SHA ECDHE-RSA-AES256-SHA DHE-RSA-AES128-SHA DHE-RSA-AES256-SHA"
# actual   : 
# --

checking tls: cipher list - ecdsa modern, with rsa fallback
# -- output differs --
# expected : "ECDHE-ECDSA-AES128-GCM-SHA256 ECDHE-RSA-AES128-GCM-SHA256 ECDHE-ECDSA-AES256-GCM-SHA384 ECDHE-RSA-AES256-GCM-SHA384 ECDHE-ECDSA-CHACHA20-POLY1305 ECDHE-RSA-CHACHA20-POLY1305 DHE-RSA-AES128-GCM-SHA256 DHE-RSA-AES256-GCM-SHA384"
# actual   : "ECDHE-ECDSA-AES128-GCM-SHA256 ECDHE-RSA-AES128-GCM-SHA256 ECDHE-ECDSA-AES256-GCM-SHA384 ECDHE-RSA-AES256-GCM-SHA384 ECDHE-ECDSA-CHACHA20-POLY1305 ECDHE-RSA-CHACHA20-POLY1305"
# --
```

All failures are due to `DHE-RSA-AES128-SHA DHE-RSA-AES256-SHA` (intermediate) and `DHE-RSA-AES128-GCM-SHA256 DHE-RSA-AES256-GCM-SHA384` (modern) no longer being supported when DH params are dropped.

Not sure why `ecdsa intermediate` and `ecdsa intermediate, with rsa fallback` output an empty "actual" result, presumably if we drop support for these 4 cipher suites (_and update the tests_) we'll have no failures. That will affect Postfix SMTP not just Dovecot though. Any old mail servers lacking ECC support for ECDHE would lose compatibility, these would be some servers from 2014 or earlier OS installs IIRC.

Do we want to keep these, or remove them with mention in release notes and documentation for enabling manually? The modern ones are still fine to use AFAIK (_provided DH params are available_). These 4 cipher suites were only kept because of OWASP recommendation which has since dropped that page and point readers to Mozilla's advice which is the same excluding those DHE suites (_albeit focused towards websites_).

georglauterbach:
> Do we want to keep these, or remove them with mention in release notes and documentation for enabling manually? The modern ones are still fine to use AFAIK (provided DH params are available). These 4 cipher suites were only kept because of OWASP recommendation which has since dropped that page and point readers to Mozilla's advice which is the same excluding those DHE suites (albeit focused towards websites).

This would likely be classified as a breaking change. I think we cannot circumvent possible breaking changes with this PR, and a new major version becomes inevitable. That being said, I'd try to update as much as possible (@polarathene you talked about getting in sync with upstream again, and I like that). We could then run this on `:edge` for a while, possibly quite a bit longer than a week. 

With version `v12.0.0`, we could remove support for old mail servers (@polarathene said something roughly about 2014 and earlier), get bigger Postfix and Dovecot version updates, etc. We should merge all other open PRs while we're at it and sync all config files with upstream again. The release of Debian 11 is the perfect point in time for that.

@wernerfred @casperklein what do you think?

NorseGaud:
> Check for changes does something weird. @NorseGaud could you please have a look? :)

Hey @georglauterbach , more than happy to help! What am I looking for? The tests don't seem to have any check-for-changes failure AFAICT

casperklein:
> With version v11.0.0, we could remove support for old mail servers

Hard to say. TBH, I've no idea, how many mail server these days do not support modern ciphers and would break due to this change.

Personally, if I had to choose between, receiving an email where the transport security is not state of the art vs. not receiving an email at all, I would choose the first one.
For me, transport security is just a nice to have in that case. There is no guarantee, the the whole email chain (transport) encrypt their traffic.
If some really wants to make sure, that no bad actor can read/alter your mails, then you have to use some kind of end to end sign & encrypt, like S/MIME or (P/G)PG.

casperklein:
I also cannot see a problem with `check-for-changes`. Instead, something with the letsencrypt certificate seems to be broken.

williamdes:
> I also cannot see a problem with `check-for-changes`. Instead, something with the letsencrypt certificate seems to be broken.

Is this PR to reload when certificates are renewed?

georglauterbach:
@NorseGaud Forgive me, I was mistaken. I misread the test results. It's not something with `check-for-changes.sh`, but with a (LE?) certificate... :)

polarathene:
> Edit: @polarathene Do you know of an easy method, how to log, which ciphers suits are used by connecting mail servers? I am interested to get a feeling, if there would be impact in my setup, if we remove old ciphers.

This is something I would appreciate information on, it's rare to find stats of this posted. I don't personally run a mailserver myself yet :sweat_smile: 

I believe Postfix details in their documentation a way to increase log verbosity which includes such information, though they advise against it in production. Not sure about Dovecot.

One alternative might be via proxy. nginx can reverse proxy StartTLS, there's a popular nginx-proxy docker-compose setup for web, but I don't think they ever got around to adding StartTLS/SMTP support. nginx is probably appropriate here if increasing log verbosity in Postfix for production setups is a concern.

Traefik can do TCP proxying but not StartTLS I think, @wernerfred has experience with Traefik and might be able to confirm, I think we could leverage TLS proxying on implicit ports (_465 submissions, 995 POP3, 993 IMAP_), the later two being Dovecot if you have users for that might be viable, I assume it's rare for third-parties to be using submission ports for SMTP.

Once that info is logged we can process it to know how a breakdown of what cipher suites are negotiated and % of each.

---

I don't see the lack of DHE as a concern for Dovecot users, we're the server and modern clients should be expected norm. I also would believe the big mail vendors and relays have no problem with ECC ciphersuites. Any problematic servers are likely spam or small business / enterprise with mail servers (_Running >5 years no upgrades: RedHat, CentOS, AWS EC2, etc_) they should really update. I would place more obligation on their end to upgrade by now.

I have also remembered that [Postfix SMTP cipher suites for port 25 are not part of our refined and explicit set](https://github.com/docker-mailserver/docker-mailserver/blob/c2fb98ea3c64e42f388e5016908f821fddba48ec/target/scripts/startup/setup-stack.sh#L873). We only prevent negotiation of ill advised cipher suites, so it has a wider support, and I don't think was involved in any of the test failures (_the tests do all ports for a configuration together, instead of port 25 separately, this could be improved.._)

---

This is the only DH params related Postfix line in the codebase AFAIK:

https://github.com/docker-mailserver/docker-mailserver/blob/c2fb98ea3c64e42f388e5016908f821fddba48ec/target/postfix/main.cf#L42

Which despite the name including `1024`, Postfix by default provides 2048-bit DH params, however in 2020 a user raised an issue about [some scanner complaining that 2048-bit wasn't sufficient](https://github.com/docker-mailserver/docker-mailserver/issues/1412), resulting in [this PR](https://github.com/docker-mailserver/docker-mailserver/pull/1463/files#diff-6d88c6c2abe074135b8efac1cf0dca97dd288a71720cdcd6ac48c2a6f85dcf08R1368) adding RFC7919 advised DH params for 3072-bit and 4096-bit, the latter of which overriding the Postfix DH params (_depending on configuration_):

https://github.com/docker-mailserver/docker-mailserver/blob/c2fb98ea3c64e42f388e5016908f821fddba48ec/target/scripts/startup/setup-stack.sh#L1313

That also handles DH params for Dovecot, thus the misconfiguration/missing pem file for those of you noticing the **issue with Debian 11 may need to investigate why that script logic failed**. It definitely looks like it could be refactored similar to my TLS functions for DRYing the logic up, especially avoiding repetition for both Postfix and Dovecot handling (_there is some inconsistencies with conditions between the two as well_).

**EDIT:** ~~Those methods aren't even called anymore?~~ :joy: The related tests([1](https://github.com/docker-mailserver/docker-mailserver/blob/master/test/mail_dhparams_default.bats),  [2](https://github.com/docker-mailserver/docker-mailserver/blob/master/test/mail_dhparams_manual_not_one_dir.bats), [3](https://github.com/docker-mailserver/docker-mailserver/blob/master/test/mail_dhparams_manual_one_dir.bats)) didn't fail despite this fact, so they should be fixed.

**EDIT2:** My bad, they are referenced here:
https://github.com/docker-mailserver/docker-mailserver/blob/e7b88d865b161d5ff82edd8f05edffa60f0d1b3c/target/scripts/start-mailserver.sh#L123
https://github.com/docker-mailserver/docker-mailserver/blob/e7b88d865b161d5ff82edd8f05edffa60f0d1b3c/target/scripts/start-mailserver.sh#L98

Not sure if those register functions are documented anywhere, I assume `_register_setup_function` does something that calls all the functions registered in sequence?

---

Keeping this dhparams handling around is probably useful. The TLS `intermediate` profile could be renamed to `legacy` and call these DH param methods, additionally enabling the postfix and dovecot config lines. If it remains as part of a TLS profile, then the tests for the profiles don't need to double for optional DH params, I can just drop those cipher suites from `modern`.

I do have interest in having SMTP cipher suites for port 25 match every other port. But as @casperklein mentions, that may reduce compatibility for older MTAs sending mail to us, Postfix likewise advises keeping it broad for the same reasons, some encryption tends to be preferable than none. In this case DH params (DHE cipher suites) are just slower performance vs ECC/ECDHE, as long as the modulus size is 2048-bit or higher it's still decently secure AFAIK.

polarathene:
@casperklein Here's some logging advice for getting cipher suites:

## TL;DR

- Logging for Postfix appears to be enabled in our image config by default.
- Dovecot needs an update to log the cipher suite.
- Then just run commands shown below to extract info. I'm not able to test this out yet.

If all works well, we update the configs to log this by default and mention it in the docs to save time for anyone seeking this info?

---

### Postfix

- `smtpd_tls_loglevel = 1` should be sufficient apparently. We [already have this enabled](https://github.com/docker-mailserver/docker-mailserver/blob/c2fb98ea3c64e42f388e5016908f821fddba48ec/target/postfix/main.cf#L27) by default for `smtp_` and `smtpd_`.
- There is a related `smtp_tls_loglevel` for outgoing negotiations if important. Likewise `lmtp_` and `tlsproxy_` variants.
- If logged details are insufficient, increasing loglevel to `2` and/or using `smtpd_tls_received_header` set to `yes` may add expected details. Details below in verbose response.

**Checking log** command: (_taken from postfix blog post quoted below_): `grep "TLS connection established from.*with cipher" /var/log/maillog | awk '{printf("%s %s %s %s\n", $12, $13, $14, $15)}' | sort | uniq -c | sort -n`

---

### Dovecot

[Dovecot config variables](https://doc.dovecot.org/configuration_manual/config_file/config_variables/) (usable in logging).

https://github.com/docker-mailserver/docker-mailserver/blob/c2fb98ea3c64e42f388e5016908f821fddba48ec/target/dovecot/10-logging.conf#L67

Append `%k`: `login_log_format_elements = user=<%u> method=%m rip=%r lip=%l mpid=%e %c %k`

**Checking log** command (_taken from blog post quoted below_): `zegrep ".*with cipher" /var/log/dovecot.log | awk '{printf("%s %s %s %s\n", $14, $15, $16, $17)}' | sort | uniq -c | sort -n`

---

<details>
<summary>Original verbose response</summary>

### Postfix

[Postfix docs `smtpd_tls_loglevel`](http://www.postfix.org/postconf.5.html) (default is `0`: Disable logging of TLS activity). Set to `2` for "Also log levels during TLS negotiation.", but note the caution from docs:

> *Do not use "smtpd_tls_loglevel = 2" or higher except in case of problems.*

Additionally [these other postfix docs](http://www.postfix.org/TLS_README.html) (_only warns of loglevel 3 and higher instead_) advise also setting `smtpd_tls_received_header` to `yes`:

> *Request that the Postfix SMTP server produces Received: message headers that include information about the protocol and cipher used, as well as the remote SMTP client CommonName and client certificate issuer CommonName.*
> 
> *This is disabled by default, as the information may be modified in transit through other mail servers. Only information that was recorded by the final destination can be trusted.*

There is a related `smtp_tls_loglevel` for outgoing negotiations if important. Likewise `lmtp_` and `tlsproxy_` variants.

#### Get stats

For analyzing there is [this 2017 answer](https://serverfault.com/a/879045) suggesting another tool for getting TLS stats, no idea how useful it is based on the discussion there and the tools docs.

I did come across [this blogpost](https://blog.tinned-software.net/harden-the-ssl-configuration-of-your-mailserver/) which only sets loglevel to 1 and describes how to extract stats:

> *The postfix log file gives a record of which ciphers were used. The logging enabled earlier will write a couple of lines to the log about the used protocol and cipher. The following command will extract these lines from the `/var/log/maillog` and count them for a quick summary.*

> ```
> grep "TLS connection established from.*with cipher" /var/log/maillog | awk '{printf("%s %s %s %s\n", $12, $13, $14, $15)}' | sort | uniq -c | sort -n
>       1 TLSv1.2 with cipher DHE-RSA-AES256-GCM-SHA384
>       1 TLSv1 with cipher ADH-DES-CBC-SHA
>       3 SSLv2 with cipher DES-CBC3-MD5
>       4 SSLv3 with cipher ECDHE-RSA-AES256-SHA
>       4 TLSv1.2 with cipher ECDHE-RSA-AES128-GCM-SHA256
>       6 TLSv1.2 with cipher ECDHE-RSA-AES256-GCM-SHA384
>       8 TLSv1 with cipher ECDHE-RSA-AES128-SHA
>      13 SSLv2 with cipher DES-CBC-MD5
>      55 TLSv1 with cipher ECDHE-RSA-AES256-SHA
>     323 TLSv1 with cipher DHE-RSA-AES256-SHA
> ```

That looks sufficient :+1: 

---

### Dovecot

I came across [this mail list advice from 2016](https://dovecot.org/pipermail/dovecot/2016-March/103535.html):

> Add %k to login_log_format_elements (in conf.d/10-logging.conf)
> 
> for example
> 
>  `login_log_format_elements = user=<%u> method=%m rip=%r lip=%l mpid=%e %c %k session=<%{session}>`

[This 2019 mail list response](https://dovecot.org/pipermail/dovecot/2019-December/117892.html):

> https://doc.dovecot.org/configuration_manual/config_file/config_variables/
> 
> You may use following log formatting
> 
> login_log_format_elements = user=<%u> method=%m rip=%r lip=%l mpid=%e %c %k
> 
> to get log content like
> 
> Dec 13 12:20:21 msg dovecot: imap-login: Login: user=<[foo at example.org](https://dovecot.org/mailman/listinfo/dovecot)>, 
method=PLAIN, rip=2003:e6:3a0c:7e12:106e:4562:cb72:566b, 
lip=2a01:1000:6:9d:0:dead:beef:cafe, mpid=6476, TLS, TLSv1.2 with cipher 
ECDHE-RSA-AES256-GCM-SHA384 (256/256 bits)
> 
> See what %c and %k translate to in that example.

[This blogpost](https://blog.schaal-24.de/mail/dovecot-und-postfix-mit-pfs-perfect-forward-secrecy):

> Dovecot (at least from 2.1.x) already used PFS, but the corresponding entries are not in the log files. Therefore, only  `%k`  has to be added in  `login_log_format_elements`. The entry is defined either in  `/etc/dovecot/dovecot.conf`  or  `/etc/dovecot/conf.d/10-logging.conf`:
>
> `login_log_format_elements = "user= method=%m rip=%r lip=%l mpid=%e %c  **%k**"`

#### Get stats (postfix and dovecot)

Taken from the end of the previous linked blog post:

> The logs of Dovecot and Postfix can then be searched using  `grep`  for DHE or ECDHE.
> 
> `zegrep ".*with cipher" /var/log/dovecot.log | awk '{printf("%s %s %s %s\n", $14, $15, $16, $17)}' | sort | uniq -c | sort -n`
> 
> `zegrep "TLS connection established from.*with cipher" /var/log/mail.log | awk '{printf("%s %s %s %s\n", $12, $13, $14, $15)}' | sort | uniq -c | sort -n`

The postfix one is the same as shown earlier, they're just using `zegrep` instead of `grep`.

</details>

georglauterbach:
> #### Debian 11 DH params missing
> 
> > For some reason, `etc/dovecot/dh.pem` is missing when executing `make generate-accounts` when using Debian 11.
> 
>     * **Short answer:** It seems we should be providing our own 4096-bit RFC7919 pem files for DH params, unclear why that broke on Debian 11. We can fix this handling and keep it for `intermediate` TLS profile, and drop DHE suites from `modern` TLS profile?
> 
>     * We have some logic in [setup-stack.sh](https://github.com/docker-mailserver/docker-mailserver/blob/c2fb98ea3c64e42f388e5016908f821fddba48ec/target/scripts/startup/setup-stack.sh#L1302) that is meant to set 4096-bit DH params (among other options) for both Postfix and Dovecot. For those finding no params file set in Debian 11, that may need to be investigated as to why.
> 
>     * That DH params logic is probably worth keeping around, and we could call it in `_setup_ssl` for `intermediate` TLS profile only? Presently the method is called in `start-mailserver.sh`. If we do this, I just need to update the tests for `modern` TLS to drop the DHE cipher suites and they should pass fine (_assuming the Debian 11 issue is identified, AFAIK the script should be providing DH params if none are provided_).

Agreed, I like this approach. Can you implement it @polarathene?

> #### Dropping DH params reduces compatibility, minimal security gain
> 
> > Personally, if I had to choose between, receiving an email where the transport security is not state of the art vs. not receiving an email at all, I would choose the first one.
> 
>     * **Short answer:** Agreed. Without stats or any valid security concerns it's probably something we should keep around, at least for `intermediate`.
> 
>     * I made a mistake with my info in prior post. Port 25 has a wider cipher suite, [we only set a min SSL/TLS protocol and exclude ciphers on top of what is already excluded by OpenSSL AFAIK](https://github.com/docker-mailserver/docker-mailserver/blob/c2fb98ea3c64e42f388e5016908f821fddba48ec/target/postfix/main.cf#L41), `tls_high_cipherlist` is our custom cipher suite setting for all other ports and shared with Dovecot.
> 
>     * DH params is still required for negotiating DHE cipher suites, but I think the tests are failing due to Dovecot not Postfix, unless Postfix is also missing DH params.
> 
>     * I would like the cipher suites for port 25 to match all others personally, but @casperklein makes a fair point about concerns for encryption compatibility, especially without any stats. Postfix docs advise the same with keeping support broad. Our cipher suites are secure / strong as-is, there isn't any notable security benefit in dropping DHE. ECDHE will be preferred by supporting clients/MTAs as it performs better.

Then we should keep compatibility broad, as @casperklein suggested, and not introduce breaking changes. This way, we may get this PR to be in `v10.2.x`, not updating to a new major. Is that possible?

polarathene:
> Agreed, I like this approach. Can you implement it @polarathene?

I can, perhaps during the following weekend.

> This way, we may get this PR to be in v10.2.x, not updating to a new major. Is that possible?

I'm not sure what is being asked here, was this regarding the implementation of your previous question?

georglauterbach:
> > This way, we may get this PR to be in v10.2.x, not updating to a new major. Is that possible?
> 
> I'm not sure what is being asked here, was this regarding the implementation of your previous question?

If the changes are not breaking anything, i.e. if there are no breaking changes in this PR, we would not need release a new major version (a circumstance I'd prefer).

polarathene:
##TL;DR

> If the changes are not breaking anything, i.e. if there are no breaking changes in this PR

Options to choose from:

- **A - Fix the script (_not breaking_):** Identifying and fixing the regression with the `setup-stack.sh` DH params handling would avoid a breaking change.
- **B - Drop official support entirely (_kinda breaking_):** Dropping DH params support for Dovecot alone (_assuming Postfix has it's own still_) just requires updating tests for Dovecot ports, not likely to affect anyone but should be mentioned in release notes (_requiring documentation with `user-patches.sh` if needed_).
- **C - ENV toggle (_breaking with fallback support_):** DH params support moved to ENV conditional. Release notes, anyone affected can enable the ENV. Not going to help with tests, since with/without DH params needs to be tested (tests require more logic) and we need to ensure supplying DH params doesn't fail (_actual problem faced with Bullseye_).
- **D - `intermediate` profile only (_breaking only for `modern` profile_):**DH params support only available in `intermediate` TLS profile. Less rework for tests than a separate ENV toggle support, but same mixed DH param suppport issue in needing to resolve the actual cause first. "Breaking" for `modern` TLS profile.

**B** is the least work. **A** is probably most ideal and technically required for **C** and **D** anyway.

---

<details>
<summary>Original message</summary>

> If the changes are not breaking anything, i.e. if there are no breaking changes in this PR

They'd be breaking if we change the cipher suite support? Proposal was to drop from `modern`, but makes no real difference here as DH params are either remaining by default or disabled.

If no one else checks before I get time to, I need to see if Postfix has the DH params pem file as configured and that Dovecot for some reason doesn't and why that is (since the PR that added script logic to handle it should be ensuring they're supplied).

If Postfix still has DH params, and we decide not to continue DH params for Dovecot, that'd still be a breaking change, just less concerning (_I doubt any Dovecot clients will have compatibility issues, MTAs sending mail to Postfix may, we need production setups to log the stats as described earlier_).

Partial or dropped DH param support would require updating the tests to reflect that, and should be announced in the release notes. I don't think it'll break anything in practice, but without sufficient stats I can't know for sure.

---

Basically, easiest solution is find out why the linked DH param logic in `setup-stack.sh` isn't working as expected, fix it and continue with DH param support as-is since test will pass again. Otherwise we drop it completely and expect users to use `user-patches.sh` themselves, also simple, but probably considered breaking.

Or we provide a means to enable via ENV or `intermediate` profile only, doing so requires reworking tests (complicated further if Postfix has DHE support but Dovecot drops it) potentially increasing the test time.

</details>

williamdes:
Option A for me
(upvote this comment for A?)

polarathene:
> can I help you with option A somehow?:)

I'm presently investigating it.

I have cloned the repo and built the Dockerfile(tagged `dms-test:buster`), then modified the base image from buster to bullseye to build a separate image (tagged `dms-test:bullseye`).

**Buster:**

```bash
docker run --rm -it -e MAIL_USER=user1@localhost.localdomain -e MAIL_PASS=mypassword -t "mailserver-testing:buster" bash

root@20b689d7c550:/# echo "${MAIL_USER}|$(doveadm pw -s SHA512-CRYPT -u ${MAIL_USER} -p ${MAIL_PASS})"
user1@localhost.localdomain|{SHA512-CRYPT}$6$97pU5x9MVBe62qts$bZb.SF.TQSuEgGAh/bhSpVqORXGmZwWkG2wQzFXTidfu8Lb43brUW8s5WOgFeeWBT2j4idotdV.oXYnIy7f.Q1

root@20b689d7c550:/# ls /etc/dovecot/dh.pem
ls: cannot access '/etc/dovecot/dh.pem': No such file or directory

root@20b689d7c550:/# cat /etc/dovecot/conf.d/10-ssl.conf | grep dh
# `docker-mailserver` is configured to use the recommended pre-defined 4096-bit DHE Group at https://github.com/internetstandards/dhe_groups
ssl_dh = </etc/dovecot/dh.pem

root@20b689d7c550:/# dovecot --version
2.3.4.1 (f79e8e7e4)
```

**Bullseye:**

```bash
docker run --rm -it -e MAIL_USER=user1@localhost.localdomain -e MAIL_PASS=mypassword -t "mailserver-testing:bullseye" bash

root@07d277d32cb4:/# echo "${MAIL_USER}|$(doveadm pw -s SHA512-CRYPT -u ${MAIL_USER} -p ${MAIL_PASS})"
doveconf: Fatal: Error in configuration file /etc/dovecot/conf.d/10-ssl.conf line 50: ssl_dh: Can't open file /etc/dovecot/dh.pem: No such file or directory user1@localhost.localdomain|

root@07d277d32cb4:/# ls /etc/dovecot/dh.pem
ls: cannot access '/etc/dovecot/dh.pem': No such file or directory

root@07d277d32cb4:/# cat /etc/dovecot/conf.d/10-ssl.conf | grep dh
# `docker-mailserver` is configured to use the recommended pre-defined 4096-bit DHE Group at https://github.com/internetstandards/dhe_groups
ssl_dh = </etc/dovecot/dh.pem

root@07d277d32cb4:/# dovecot --version
2.3.13 (89f716dc2)
```

I'm a little stumped with that... It's working fine on Buster, but `/etc/dovecot/dh.pem` doesn't exist, nor is the config path value for `ssl_dh` different..?

Perhaps some bug fix in-between these versions of Dovecot is the reason, or something to do with the `doveadm` command. I've not manually scanned the [Dovecot Changelog](https://dovecot.org/doc/NEWS) between versions, no luck with some keywords such as `dh` and `doveadm` though :\

---

I will continue to investigate, if you have any ideas why the behaviour above is happening despite the same config and lack of `dh.pem` file in both environments, that'd be helpful.

casperklein:
There was a change between dovecot versions. As I am using the community version, I encountered this problem earlier. See my comment here: https://github.com/docker-mailserver/docker-mailserver/issues/1982#issuecomment-844074143

polarathene:
> There was a change between dovecot versions.

I understand that, but there isn't anything that I saw in the CHANGELOG for Dovecot that indicates why. As shown from shell output commands above, there isn't anything notably different between the two.

I can only assume the `doveadm` command at some point decided to check for the dh file existing even if not necessary?

I ran the two images with the actual init scripts for docker-mailserver and both created a `dh.pem` file. So I can only assume that this is due to a validation introduced by `doveadm` since.

I don't think we need `dh.pem` enabled when calling `setup` methods that invoke `doveadm`, but disabling it by default where `docker-mailserver` scripts may conditionally enable it at a later point and persist it sounds like a hassle to work around. So we can just provide our own default in the Dockerfile build based on your solution :)

casperklein:
> I can only assume the doveadm command at some point decided to check for the dh file existing even if not necessary?

I think so too.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM docker.io/debian:buster-slim
+FROM docker.io/debian:bullseye-20210927-slim
```
casperklein, line 1:
```suggestion
FROM docker.io/debian:11-slim
```

georglauterbach: I will use `11.0` as it is a more robust tag:) But thanks for the hint!

georglauterbach:
> I'll make the changes after the updated tests are merged.

Nice 👍🏼

polarathene:
[Debian 11 Bullseye no longer ships `python` by default](https://unix.stackexchange.com/questions/609572/debian-bullseye-apt-full-upgrade-removes-whole-lot-of-packages), we use `python3` via packages which I found out when looking into why a package with "python" in the name wasn't pulling in the `python` package which fixed the issue as mentioned in my previous comment.

Python 2 is EOL, and I think the only reason this was an issue was because of the inline python code added for handling `acme.json` which called `python` instead of `python3`. That should resolve the failure better than pulling in `python` (Python 2).

target/scripts/helper-functions.sh
```diff
@@ -167,7 +167,7 @@ for key, value in acme.items():
 
   local CERT
   # shellcheck disable=SC2002
-  CERT=$(cat /etc/letsencrypt/acme.json | python -c "
+  CERT=$(cat /etc/letsencrypt/acme.json | python3 -c "
```
williamdes, line 170: Do we require python in our image only to do that?

polarathene: No, another dependency [`postfix-policyd-spf-python`](https://packages.debian.org/bullseye/postfix-policyd-spf-python) depends on it, I'm not sure if anything else does:

<details>
<summary>revere dependencies on bullseye upgrade for python3</summary>

```console
$ apt-cache rdepends --installed python3  
python3
Reverse Depends:
  postfix
```

```console
$ apt-rdepends python3
python3
  Depends: libpython3-stdlib (= 3.9.2-3)
  Depends: python3.9 (>= 3.9.2-0~)
  PreDepends: python3-minimal (= 3.9.2-3)
libpython3-stdlib
  Depends: libpython3.9-stdlib (>= 3.9.2-0~)
libpython3.9-stdlib
  Depends: libbz2-1.0
  Depends: libc6 (>= 2.28)
  Depends: libcrypt1 (>= 1:4.1.0)
  Depends: libdb5.3
  Depends: libffi7 (>= 3.3~20180313)
  Depends: liblzma5 (>= 5.1.1alpha+20120614)
  Depends: libmpdec3
  Depends: libncursesw6 (>= 6)
  Depends: libnsl2 (>= 1.0)
  Depends: libpython3.9-minimal (= 3.9.2-1)
  Depends: libreadline8 (>= 7.0~beta)
  Depends: libsqlite3-0 (>= 3.7.15)
  Depends: libtinfo6 (>= 6)
  Depends: libtirpc3 (>= 1.0.2)
  Depends: libuuid1 (>= 2.20.1)
  Depends: media-types
  Depends: mime-support
  Depends: tzdata
libbz2-1.0
  Depends: libc6 (>= 2.4)
libc6
  Depends: libcrypt1
  Depends: libgcc-s1
libcrypt1
  Depends: libc6 (>= 2.25)
libgcc-s1
  Depends: gcc-10-base (= 10.2.1-6)
  Depends: libc6 (>= 2.14)
gcc-10-base
libdb5.3
  Depends: libc6 (>= 2.17)
libffi7
  Depends: libc6 (>= 2.14)
liblzma5
  Depends: libc6 (>= 2.17)
libmpdec3
  Depends: libc6 (>= 2.14)
  Depends: libgcc-s1 (>= 3.0)
  Depends: libstdc++6 (>= 5.2)
libstdc++6
  Depends: gcc-10-base (= 10.2.1-6)
  Depends: libc6 (>= 2.23)
  Depends: libgcc-s1 (>= 4.2)
libncursesw6
  Depends: libc6 (>= 2.14)
  Depends: libtinfo6 (= 6.2+20201114-2)
libtinfo6
  Depends: libc6 (>= 2.16)
libnsl2
  Depends: libc6 (>= 2.14)
  Depends: libtirpc3 (>= 1.0.2)
libtirpc3
  Depends: libc6 (>= 2.16)
  Depends: libgssapi-krb5-2 (>= 1.17)
  Depends: libtirpc-common (>= 1.3.1-1)
libgssapi-krb5-2
  Depends: libc6 (>= 2.27)
  Depends: libcom-err2 (>= 1.43.9)
  Depends: libk5crypto3 (>= 1.16)
  Depends: libkrb5-3 (= 1.18.3-6+deb11u1)
  Depends: libkrb5support0 (>= 1.15~beta1)
libcom-err2
  Depends: libc6 (>= 2.17)
libk5crypto3
  Depends: libc6 (>= 2.25)
  Depends: libkrb5support0 (>= 1.16)
libkrb5support0
  Depends: libc6 (>= 2.25)
libkrb5-3
  Depends: libc6 (>= 2.25)
  Depends: libcom-err2 (>= 1.43.9)
  Depends: libk5crypto3 (>= 1.15~beta1)
  Depends: libkeyutils1 (>= 1.5.9)
  Depends: libkrb5support0 (= 1.18.3-6+deb11u1)
  Depends: libssl1.1 (>= 1.1.0)
libkeyutils1
  Depends: libc6 (>= 2.14)
libssl1.1
  Depends: debconf (>= 0.5)
  Depends: debconf-2.0
  Depends: libc6 (>= 2.25)
debconf
  PreDepends: perl-base (>= 5.20.1-3~)
perl-base
  PreDepends: dpkg (>= 1.17.17)
  PreDepends: libc6 (>= 2.29)
  PreDepends: libcrypt1 (>= 1:4.1.0)
dpkg
  Depends: tar (>= 1.28-1)
  PreDepends: libbz2-1.0
  PreDepends: libc6 (>= 2.15)
  PreDepends: liblzma5 (>= 5.2.2)
  PreDepends: libselinux1 (>= 3.1~)
  PreDepends: zlib1g (>= 1:1.1.4)
tar
  PreDepends: libacl1 (>= 2.2.23)
  PreDepends: libc6 (>= 2.28)
  PreDepends: libselinux1 (>= 3.1~)
libacl1
  Depends: libc6 (>= 2.14)
libselinux1
  Depends: libc6 (>= 2.14)
  Depends: libpcre2-8-0 (>= 10.22)
libpcre2-8-0
  Depends: libc6 (>= 2.14)
zlib1g
  Depends: libc6 (>= 2.14)
debconf-2.0
libtirpc-common
libpython3.9-minimal
  Depends: libc6 (>= 2.4)
  Depends: libssl1.1 (>= 1.1.1)
libreadline8
  Depends: libc6 (>= 2.15)
  Depends: libtinfo6 (>= 6)
  Depends: readline-common
readline-common
  Depends: dpkg (>= 1.15.4)
  Depends: install-info
install-info
  Depends: libc6 (>= 2.14)
  PreDepends: dpkg (>= 1.16.1)
libsqlite3-0
  Depends: libc6 (>= 2.29)
libuuid1
  Depends: libc6 (>= 2.25)
media-types
mime-support
  Depends: mailcap
  Depends: media-types
mailcap
  Depends: media-types
  Depends: perl
perl
  Depends: libperl5.32 (= 5.32.1-4+deb11u2)
  Depends: perl-base (= 5.32.1-4+deb11u2)
  Depends: perl-modules-5.32 (>= 5.32.1-4+deb11u2)
  PreDepends: dpkg (>= 1.17.17)
libperl5.32
  Depends: libbz2-1.0
  Depends: libc6 (>= 2.29)
  Depends: libcrypt1 (>= 1:4.1.0)
  Depends: libdb5.3
  Depends: libgdbm-compat4 (>= 1.18-3)
  Depends: libgdbm6 (>= 1.18-3)
  Depends: perl-modules-5.32 (>= 5.32.1-4+deb11u2)
  Depends: zlib1g (>= 1:1.2.2.3)
libgdbm-compat4
  Depends: libc6 (>= 2.14)
  Depends: libgdbm6 (>= 1.16)
libgdbm6
  Depends: libc6 (>= 2.14)
perl-modules-5.32
  Depends: perl-base (>= 5.32.1-1)
  PreDepends: dpkg (>= 1.17.17)
tzdata
  Depends: debconf (>= 0.5)
  Depends: debconf-2.0
python3.9
  Depends: libpython3.9-stdlib (= 3.9.2-1)
  Depends: media-types
  Depends: mime-support
  Depends: python3.9-minimal (= 3.9.2-1)
python3.9-minimal
  Depends: libexpat1 (>= 2.1~beta3)
  Depends: libpython3.9-minimal (= 3.9.2-1)
  Depends: zlib1g (>= 1:1.2.0)
  PreDepends: libc6 (>= 2.29)
libexpat1
  Depends: libc6 (>= 2.25)
python3-minimal
  Depends: dpkg (>= 1.13.20)
  PreDepends: python3.9-minimal (>= 3.9.2-0~)
```

```console
$ apt-rdepends python3.9-minimal
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
python3.9-minimal
  Depends: libexpat1 (>= 2.1~beta3)
  Depends: libpython3.9-minimal (= 3.9.2-1)
  Depends: zlib1g (>= 1:1.2.0)
  PreDepends: libc6 (>= 2.29)
libexpat1
  Depends: libc6 (>= 2.25)
libc6
  Depends: libcrypt1
  Depends: libgcc-s1
libcrypt1
  Depends: libc6 (>= 2.25)
libgcc-s1
  Depends: gcc-10-base (= 10.2.1-6)
  Depends: libc6 (>= 2.14)
gcc-10-base
libpython3.9-minimal
  Depends: libc6 (>= 2.4)
  Depends: libssl1.1 (>= 1.1.1)
libssl1.1
  Depends: debconf (>= 0.5)
  Depends: debconf-2.0
  Depends: libc6 (>= 2.25)
debconf
  PreDepends: perl-base (>= 5.20.1-3~)
perl-base
  PreDepends: dpkg (>= 1.17.17)
  PreDepends: libc6 (>= 2.29)
  PreDepends: libcrypt1 (>= 1:4.1.0)
dpkg
  Depends: tar (>= 1.28-1)
  PreDepends: libbz2-1.0
  PreDepends: libc6 (>= 2.15)
  PreDepends: liblzma5 (>= 5.2.2)
  PreDepends: libselinux1 (>= 3.1~)
  PreDepends: zlib1g (>= 1:1.1.4)
tar
  PreDepends: libacl1 (>= 2.2.23)
  PreDepends: libc6 (>= 2.28)
  PreDepends: libselinux1 (>= 3.1~)
libacl1
  Depends: libc6 (>= 2.14)
libselinux1
  Depends: libc6 (>= 2.14)
  Depends: libpcre2-8-0 (>= 10.22)
libpcre2-8-0
  Depends: libc6 (>= 2.14)
libbz2-1.0
  Depends: libc6 (>= 2.4)
liblzma5
  Depends: libc6 (>= 2.17)
zlib1g
  Depends: libc6 (>= 2.14)
debconf-2.0
```

</details>

I doubt removing python will save much image size ratio-wise if that's what you were thinking. Your welcome to replace the `acme.json` parsing/extraction script and the SPF policy server for Postfix, but I doubt it's worth the effort :sweat_smile: 

I've sussed out the test failure (_the python code for acme.json is presently Python 2 apparently, so it needed to be updated to v3_).

georglauterbach: I don't think removing python is a good idea too. We should leave it in there, but we should also use Python 3 :D

polarathene:
Additional Debian 11 changes to consider:

- [`yescrypt` has been adopted for user password hashing](https://www.debian.org/releases/stable/amd64/release-notes/ch-information.en.html#pam-default-password), this would be better than SHA512-CRYPT in my opinion. It's similar to `scrypt` but more flexible. That said, AFAIK `bcrypt` is also available, not sure about argon2id which would be nicer.
- [Possible issue for `fail2ban` in known severe bugs](https://www.debian.org/releases/stable/amd64/release-notes/ch-information.en.html#rc-bugs):
  > fix for CVE-2021-32749 breaks systems with mail from `bsd-mailx`

  The [Bullseye release notes have a little section about it here](https://www.debian.org/releases/bullseye/amd64/release-notes/ch-information.en.html#fail2ban):

  > Users of `fail2ban` in combination with `bsd-mailx` who wish `fail2ban` to send out e-mail should either switch to a different provider for **mail** or manually unapply [the upstream commit](https://github.com/fail2ban/fail2ban/commit/410a6ce5c80dd981c22752da034f2529b5eee844) (which inserted the string `-E 'set escape'` in multiple places under `/etc/fail2ban/action.d/`).

  [Github `fail2ban` repo issue](https://github.com/fail2ban/fail2ban/issues/3059). I'm not sure exactly how that affects us, the current configs at the `fail2ban` repo do have the `-E` usage and our Dockerfile does add the `bsd-mailx` package.

casperklein:
I am not so familiar with the different hash types. As long as they are not deprecated, I think all of them should be fine for us.
As we also build for ARM arch, some might run DMS on a raspberry. Performance could be take into consideration when choosing a new hash algo.

---

> Possible issue for fail2ban in known severe bugs:
> I'm not sure exactly how that affects us

If I understand correctly, this breaks the possibility to let fail2ban send mails about bans. If that is a useful feature at all is a different question 😆 
To fix it in v11, some could undo the [patch](https://github.com/fail2ban/fail2ban/commit/410a6ce5c80dd981c22752da034f2529b5eee844.patch) via `config/user-patches.sh`

polarathene:
> As we also build for ARM arch, some might run DMS on a raspberry. Performance could be take into consideration when choosing a new hash algo.

Security trumps performance when it comes to choice :sweat_smile: when security doesn't matter you can favor performance, eg for checksums. For small deployments (eg: on a rpi), you're not likely to have any performance concerns with login anyway.

Many of these have a difficulty based on computation effort. You do want to balance it to not be too heavy on your hardware/server and negatively impact users with long login delays, but at the same time you don't want the computation to be tuned "for performance" where on weak hardware it enables an attacker to be more successful sooner.

I see advise ranging between 50-250ms usually for interactive login when tuning, where 50ms would handle 20 auths/logins per second.

Raising that too high and you need rate-limiting or other means to detect/handle abuse where a malicious user puts the server under load wasting resources on hashing rather than doing useful things. Using cgroups to limit where resources are spent helps (_probably not as much with `docker-mailserver` as I think the container is considered a single unit, unless auth is delegated externally_).

So a balance is important. This is more applicable to `bcrypt` and the other kinds like `SHA512-CRYPT` that we currently use, however the latter has additional weaknesses IIRC (_I was writing up an issue about it months ago, but I believe I lost the content to a browser crash before submitting_).

`scrypt` (_and the variant `yescrypt` that distros are adopting lately_), tries to address the issue a bit better by raising the memory requirements to perform the processing. `bcrypt` only requires 4KB (_still enough to bottleneck parallelization on GPUs local memory cores, but not FPGA rigs which can attack it more affordably_), but `scrypt` had memory scaling tied to the computational difficulty, so it didn't work much better in practice IIRC. I don't recall too much about `yescrypt` other than it's meant to be an improvement upon that and is more configurable receiving praise.

`scrypt` has previously not seen wide adoption or standardization however AFAIK, but Linux distros have been looking into adopting it for login passwords over the `SHA512-CRYPT` that's been long used (_`bcrypt` was preferable in the past, but the maintainer of some part of linux and author of `SHA512-CRYPT` kept downplaying it despite concerns expressed about `SHA512-CRYPT`, `bcrypt` also wasn't NIST approved while `SHA512-CRYPT` used NIST approved SHA2 primitive_).

These days a more advised solution is `argon2id` which scales well and can configure memory and cpu resource demands, allowing you to balance the distribution to suit your server hardware and security needs. It's lower bound though if your hardware is too weak seems to make that flexibility a bit moot (_as in the due to minimal computation for the lowest practical config values don't really allow lowering the cpu time further by upping memory required_).

We can stick with `SHA512-CRYPT` for now, I'm not sure how changing that would affect users upgrading to a new release, but I think we should look into changing to something better endorsed by professionals in future. As long as users have strong passwords (eg: diceware based entropy), it's not practical to attack password hashes. Raising the difficulty and thus time/resources required to process is effectively equivalent to adding an X bits of entropy as a security margin against poor passwords.

polarathene:
> If I understand correctly, this breaks the possibility to let fail2ban send mails about bans. If that is a useful feature at all is a different question :laughing: 

Not something I'm familiar with much myself, I leave up to the other maintainers to know if it's an issue with upgrading and if any mention needs to be added to release notes or documented.

I just wanted to make us all aware of it.

williamdes:
> If I understand correctly, this breaks the possibility to let fail2ban send mails about bans. If that is a useful feature at all is a different question 😆

I use this feature in production and want the fail2ban mails to work btw :)

casperklein:
Out of interest, how do you use this feature? What are the benefits for you.

georglauterbach:
About the hashes: I'd love to see `argon2id` being used, but I also think that this could end up in a lot of (as of now, potentionally unnecessary) work. I know that Dovecot supports `argon2id`, and I guess Postfix does too. One would need to dig deeper there whether other applications needs these hashes too, and how this affects compatibility.

williamdes:
> Out of interest, how do you use this feature? What are the benefits for you.

I use it to know what ip is banned because clearly I already blocked most of bad actors on my router using ip block lists
And also to know when the service stops and starts 

Screenshot_2021-11-01-13-00-06-892_com fsck k9

georglauterbach:
@docker-mailserver/maintainers please leave your review :D

This is marked for `v11`, so no immediate merges. But it'd be good to have it ready :)

polarathene (approved):

casperklein (approved):

williamdes (approved):

georglauterbach:
Now that `v10.3.0` was released, I think we can go ahead an merge this PR as well. Are there currently any objections?

williamdes (approved):

casperklein:
I propose, changing `SSL_TYPE=letsencrypt` to e.g. `SSL_TYPE=certbot`. I think, that is more clear. You don't have to use SSL_TYPE=letsencrypt, when using a LE certificate. That have confused some people in the past. Also in the meantime, there are new cert providers supporting free certificates via ACME, like ZeroSSL. So certbot can also use other provides instead of LE.

williamdes:
> I propose, changing `SSL_TYPE=letsencrypt` to e.g. `SSL_TYPE=certbot`. I think, that is more clear. You don't have to use `SSL_TYPE=letsencrypt`, when using a LE certificate. That have confused some people in the past. Also in the meantime, there are new cert providers supporting free certificates via ACME, like ZeroSSL. So certbot can also use other provides instead of LE.

We should add "certbot" and keep the other options IMO. having "certbot" would mean, pick any provider you want.
Another value than "certbot" would also be nice, maybe "automatic-ca" or "automatic".
Ref: https://github.com/acmesh-official/acme.sh#supported-ca

georglauterbach:
The discussion about LE vs certbot is very valid, but I'd rather like to keep this PR as simple as possible. The discussion could very well be solved with a future PR so we do not mix up concerns here.

casperklein:
Why would we want to keep a outdated/not clear option naming? This is a major release, expected to introduce breaking changes.
`SSL_TYPE=certbot` nails it, because that's exactly what happens internally (handling a certbot generated json file).

PS: I don't know if that json file is certbot specific or a standardized acme json file.

georglauterbach:
> Why would we want to keep a outdated/not clear option naming? This is a major release, expected to introduce breaking changes. `SSL_TYPE=certbot` nails it, because that's exactly what happens internally (handling a certbot generated json file).

I agree, but we do not immediately release `v11` after this PR is merged, right? This is on `:edge`, and we can add other PRs afterwards as far as I understood it.

casperklein:
I haven't read the whole thread again, but I was the opinion, there were some other breaking changes mentioned, when introducing debian 11 🤷

polarathene:
> This is only my thoughts, let's discuss about all this in a dedicated GH discussion ;)

As we don't enable "Discussions" here and our internal GH teams area splits moderators and maintainers it seems I can't pick it up over there. That'd leave creating a new issue, but I think I'll just weigh in here since my response probably resolves it.

---

> I propose, changing `SSL_TYPE=letsencrypt` to e.g. `SSL_TYPE=certbot`. I think, that is more clear. You don't have to use `SSL_TYPE=letsencrypt`, when using a LE certificate.

While I agree with the concerns about `SSL_TYPE=letsencrypt`, I don't agree with `certbot` for reasons similar to @williamdes . 

In the past I have considered adding `certbot` or `acme.sh` internally and making that available as a mode to potentially simplify cert management/setup for some users that would prefer it, and that would be relevant as an `SSL_TYPE` then.

`certbot` would be incorrect for replacing `letsencrypt` however as nothing about this type internally is related to `certbot` really. We also document the `letsencrypt` mode for other providers such as `nginx-proxy` which uses `acme.sh` internally to provision it's LE certs.

> Another value than `certbot` would also be nice, maybe `automatic-ca` or `automatic`.

Yeah, at one point I considered `acme` but since working on the refactor recently of `letsencrypt` support and getting familiar with [that part of the support we provide via that mode](https://github.com/docker-mailserver/docker-mailserver/blob/b11e5ffd1a36244fa90416df49c43fe93a4a013a/target/scripts/startup/setup-stack.sh#L847-L914), it's really just "automatic" discovery.

We just check a hard-coded mount path for a few variations of expected cert folders and then for a few hard-coded pem filenames (cert: `fullchain.pem`, key: `key.pem` / `privkey.pem`). You'll see from the comments there, that we already maintain for compatibility beyond `certbot` as a provisioner, which while they seem to standardize on the cert filename, there is a little variance for the private key filename.

`automatic` would be reasonably more accurate, but it still requires meeting certain expectations, hence our documentation groups a few providers under the "Let's Encrypt" mode section for what we know is supported by us officially.

> because that's exactly what happens internally (handling a certbot generated json file).
> PS: I don't know if that json file is certbot specific or a standardized acme json file.

It's neither. You can view the linked lines of code to the `letsencrypt` handling or the letsencrypt tests and both have been documented by me to clarify that `acme.json` is specific to Traefik support (_which effectively falls under the `letsencrypt` mode, but only because we extract the files to the expected location and filenames to be compatible_).

> People need to understand that certbot is a tool and LE one of many CAs.

Absolutely agree.

---

## TL;DR

:-1: to `SSL_TYPE=certbot`, that would not be any better IMO. The mode will become irrelevant in future after a refactoring (late Dec or sometime in Jan), see below for details.

---

FWIW, refactoring this particular feature has been on my TODO for quite a while, but I've been busy outside of the project as well as spending time for this project on assisting with other PRs / issues.

With the refactoring that was started in `v10.3.0` the linked TLS logic from `setup-stack.sh` will be extracted out into a helper as `check-for-changes.sh` needs it for an edge case bug, and then afterwards a larger refactoring of the handling will take place, likely to unify all the modes as there is a lack of feature parity and generally no need for them to be distinct modes, rather separate complimentary ENVs will address those needs better and the current `letsencrypt` support will just become a fallback for discovery if explicit paths are not provided.

As for internally using `certbot` or `acme.sh`, for the time being, it may be better supported via documentation examples of using `user-patches.sh` as I can't see us officially supporting in issues or documentation beyond that (_they've both got massive docs and all we care about is the end result files_).

polarathene:
> I haven't read the whole thread again, but I was the opinion, there were some other breaking changes mentioned, when introducing debian 11 :shrug:

Those IIRC were related to actual changes from the Debian upgrade, such as the default hashing for user accounts changing and if we should consider adopting that for Dovecot (_which spurred some discussion about alternatives like `argon2id`_).

I've tried to manage the noise in the discussion by hiding no longer relevant messages, while keeping some (mostly mine) that were still deemed of value for someone going over the PR, some which are buried now until you expand the collapsed entries to get the full PR discussion listed such as:
- [This one](https://github.com/docker-mailserver/docker-mailserver/pull/2116#issuecomment-955615529) relevant to release notes
- And [this one](https://github.com/docker-mailserver/docker-mailserver/pull/2116#issuecomment-955800544) regarding notable changes to discuss about password hashing and CVE/bugs like `bsd-mailx` with `fail2ban`.

Regarding those changes, password hashing won't be touched anytime soon due to effort required to investigate implementation and compatibility concerns (_it was discussed in the WIP encryption feature thread too_). And for the `bsd-mailx` concern where @williamdes expresses interest in that continuing to work, so someone (perhaps @williamdes ) will need to verify that works fine and the CVE/bug concerns (that I linked to) aren't applicable to keep the feature around.

---

> Now that `v10.3.0` was released, I think we can go ahead an merge this PR as well. Are there currently any objections?

We may want to wait a week, one of the open issues about `letsencrypt` renewal bug is going to be confirmed if fixed or not. It'd be a bit easier to address from feedback via a hotfix (_useful in general for holding off on merging this PR after a new fresh release_) or new minor release if such a release doesn't have to dance around changes like this one would add, since I don't anticipate v11 is being released until late Dec or in Jan? Or are you wanting to release v11 with minimal changes beyond this PR?

No objections from me, other than that concern.

casperklein:
> certbot would be incorrect for replacing letsencrypt however as nothing about this type internally is related to certbot really. 

I agree. I admit, I should have take a deeper look before. My point was, that `SSL_TYPE=letsencrypt` is not suitable and we should choose a better wording.

> The mode will become irrelevant in future after a refactoring (late Dec or sometime in Jan), see below for details.

Ack. I am fine with not changing the wording for now.

georglauterbach:
We can with merging this PR for a week, sure. There is no rush with releasing v11.

georglauterbach:
@polarathene please reach back to me whe you're fine with merging this (when the issues seem resolved, I guess), Orga ahead and merge this yourself by all accounts :)

PR #1988, opened 2021-05-20
docs(ci): Deploy Previews

polarathene:
# Description

**Feature:** When a PR contributes to our docs, a preview link will be added as a comment via Github Actions.

---

Each PR will have a unique prefix alias of `pullrequest-N--` (_where N is the PR number_) to our docs preview site at [dms-doc-previews.netlify.app](https://dms-doc-previews.netlify.app/) (_404 page is intentional, only the preview aliases for PRs will have actual content_).

Example: `https://pullrequest-42--dms-doc-previews.netlify.app`

---

Details of interested are covered below. The workflows have some inlined documentation via comments, and each commit should be fairly clean with useful commit messages.

## Sponsored plan + requirements to retain

<details>
<summary>Netlify has sponsored the docker-mailserver organization with an OSS team account.</summary>

To keep this sponsorship, we must comply with the [plan policy](https://www.netlify.com/legal/open-source-policy); we qualified for sponsorship so currently meet the requirements. The Netlify branding mention is injected into build config for PRs only and can be found in the footer section on all pages.

We continue to deploy the production versions of documentation to Github Pages, but will utilize their service for deploying preview builds of PRs touching the documentation.

</details>

## Billing / Plan limits

<details>
<summary>There are two billing concerns that still apply if exceeded</summary>

- [Bandwidth usage](https://docs.netlify.com/accounts-and-billing/billing/#bandwidth-usage) (400GB monthly)
  - This is regarding bandwidth transferring data to viewers. Our site is under 2MB in total, but this is considerably less over the wire as Netlify compresses responses/assets with gzip and brotli where appropriate. Caching is also leveraged on the client side.
  - As it's only being used for previewing PR contributions for docs, this is expected to be very low and a non-concern.
- [Build Minutes](https://docs.netlify.com/accounts-and-billing/billing/#builds-usage) (1000 minutes per month)
  - This shouldn't be a concern. Our builds don't take long, but it could more easily be abused maliciously or through very high frequency of pushing commits.
  - Github Actions is handling builds instead, and then transferring the build to Netlify via their deploy API. This bandwidth is not part of the billing limit, and does not impact the build time.
  - To further guard against high frequency commits to a PR, Github Actions `concurrency` feature is in use to cancel any existing instance of the workflow for the same PR.

[...]

PR #2153, opened 2021-08-28
fix: broken documentation workflow

### LINKED ISSUE #2152: Pipeline for documentation currently broken
# Bug Report

## Context

We should see the versioned docs for every release in `<major>.<minor>`
Currently there is only 3 versions available:
- `v9.1`
- `v10.0`
- `edge`

Latest release on the main repo is `v10.1.1` so we are missing at least `v10.1` 

The documentation

### When does this occur?

When a new release gets published

Publish a release (should be avoided due to public visibility)

### Actual Behavior

No version `v10.1` of docs

### Expected Behavior

Docs with version `v10.1`

### Relevant Stack Traces

<details>
<summary>Action log</summary>

```
Run ../.github/workflows/scripts/docs/update-versions-json.sh
  ../.github/workflows/scripts/docs/update-versions-json.sh
  shell: /usr/bin/bash -e {0}
  env:
    DOCS_VERSION: edge
    GIT_USER: github-actions[bot]
    GIT_EMAIL: 41898282+github-actions[bot]@users.noreply.github.com
../.github/workflows/scripts/docs/update-versions-json.sh: line 60: _update-versions-json: command not found
Error: Process completed with exit code 127.
```

</details>

The documentation pipeline is broken at some point: 
https://github.com/docker-mailserver/docker-mailserver/runs/3451215273?check_suite_focus=true#step:4:9

Further investigation needed why this happened and when.
I will look into that and encourage @polarathene to help me out finding the root that caused that and why it didn't attract our attention over 3 releases (Pipeline should fail).

wernerfred:
# Description

Fixes #2152 

Problems that need to be adressed:
- [x] Fix error in workflow script that causes current runs to fail
- [ ] ~Fix issue with actions showing state "passed" although it failed~

## Type of change

## Checklist:


wernerfred:
Seems like this issue was introduced by @polarathene during your work on #1988 
We need to investiagate if this function is called from another workflow as well

https://github.com/docker-mailserver/docker-mailserver/blame/master/.github/workflows/scripts/docs/update-versions-json.sh#L6

polarathene:
> Seems like this issue was introduced by @polarathene during your work on #1988

Whoops! Sorry about that :(

polarathene (approved):
> We need to investiagate if this function is called from another workflow as well

I cloned master and did a ripgrep search (with `--hidden`) and searched for "update", nothing else is calling it. I did find it odd that we have commands like `updatemailuser` instead of `update-mailuser` or `update_mailuser` though :sweat_smile: (no functional issues, just readability).

So only the docs production deploy workflow calls this update method as would be expected. It does so only via the script name, since I chose to wrap it in a function and call the method immediately afterwards:

https://github.com/docker-mailserver/docker-mailserver/blob/784ec60612476350242a2ead1bfb2e804a38cf2d/.github/workflows/docs-production-deploy.yml#L91

wernerfred:
No problem! But glad to hear that you did not change this on purpose for further dependencies because that means this PR is ready to merge i think? I will leave it up to you to double check and merge if you agree :v:

I would propose not to "test" this via a release but to wait for the next release and doublecheck then.

wernerfred:
The `edge` deployment is unaffected isn't it?

polarathene:
> The `edge` deployment is unaffected isn't it?

Correct. Details of flow below if useful:

---

Default docs deploy target is `edge`:

https://github.com/docker-mailserver/docker-mailserver/blob/4755b16b312d9ac7f814ef7a8b62dba5251663c1/.github/workflows/docs-production-deploy.yml#L17-L19

Conditional task that will update target from `edge` to tagged version if appropriate:

https://github.com/docker-mailserver/docker-mailserver/blob/4755b16b312d9ac7f814ef7a8b62dba5251663c1/.github/workflows/docs-production-deploy.yml#L31-L32

Then later deploy:

https://github.com/docker-mailserver/docker-mailserver/blob/4755b16b312d9ac7f814ef7a8b62dba5251663c1/.github/workflows/docs-production-deploy.yml#L59-L60

2nd job waits on 1st deploy job, if tagged commit then runs the script:

https://github.com/docker-mailserver/docker-mailserver/blob/4755b16b312d9ac7f814ef7a8b62dba5251663c1/.github/workflows/docs-production-deploy.yml#L70-L75

Github Pages branch switched to and `versions.json` updated:

https://github.com/docker-mailserver/docker-mailserver/blob/4755b16b312d9ac7f814ef7a8b62dba5251663c1/.github/workflows/docs-production-deploy.yml#L86-L91

And finally committed and pushed if no issues:

https://github.com/docker-mailserver/docker-mailserver/blob/4755b16b312d9ac7f814ef7a8b62dba5251663c1/.github/workflows/docs-production-deploy.yml#L93-L96

---

So the tagged releases are still deployed:

https://github.com/docker-mailserver/docker-mailserver/tree/gh-pages

Just the bug here prevented the `versions.json` being updated for the version selector dropdown UI on the docs to know about it as an option. It can be manually updated if needed.

But in this case, tagged releases are only Major and Minor as the workflow mentions, any patch version just overrides the current `<Major>.<Minor>` (10.1). As that presently doesn't exist in `versions.json`, the next tagged release after this fix is merged will just add `10.1` to `versions.json` and any Patch release will update the `10.1` deployed docs folder:

https://github.com/docker-mailserver/docker-mailserver/blob/c06e379244910a57fff4a30e7c4af0bc141f533a/versions.json

https://github.com/docker-mailserver/docker-mailserver/tree/gh-pages/v10.1

polarathene (approved):

Issue #2154, opened 2021-08-28
[BUG] [10.1.1] High CPU usage caused by check-for-changes.sh


KowalskiPiotr98:
# Bug Report

After updating to v10.1.1 I'm seeing heavy CPU usage caused by `check-for-changes.sh`

## Context

![image](https://user-images.githubusercontent.com/56399446/131232335-87169b39-bc50-4a96-8bf2-70536d3510fa.png)

![image](https://user-images.githubusercontent.com/56399446/131232337-551d84d4-8fd4-4453-b323-d654869f4b6e.png)

This is what I'm seeing. The CPU usage graph shows a significant increase after an update to 10.1.1 and then a sudden drop once I've downgraded back to 10.1.0. Htop reported `check-for-changes.sh` as the process using the most CPU on the server at the time.

This is how `check-for-changes.sh` behaves on 10.1.0:
![image](https://user-images.githubusercontent.com/56399446/131232353-1ca20cf7-f225-4733-a420-eb5c310f0ed3.png)

There was no increase in mail traffic during that period of time so I don't believe it was caused by external factors.

### What is affected by this bug?

High CPU usage.

### When does this occur?

All the time on v10.1.1

### How do we replicate the issue?

1. Start the server on v10.1.1
2. Done

## Your Environment

- version: `v10.1.1`
- available RAM: `4GB`
- Docker version: `v20.10.8`
- docker-compose version: `v1.28.5`

### Environment Variables

```CFG
ENABLE_UPDATE_CHECK=1
UPDATE_CHECK_INTERVAL=1d
AMAVIS_LOGLEVEL=0
FAIL2BAN_BLOCKTYPE=drop
ONE_DIR=1
SPOOF_PROTECTION=1
ENABLE_CLAMAV=1
ENABLE_FAIL2BAN=1
POSTFIX_MAILBOX_SIZE_LIMIT=10737420000
PFLOGSUMM_TRIGGER=daily_cron
LOGWATCH_INTERVAL=daily
REPORT_RECIPIENT=1
ENABLE_SPAMASSASSIN=1
SPAMASSASSIN_SPAM_TO_INBOX=1
MOVE_SPAM_TO_JUNK=1
```
Above are the env variables that were not empty that I've either changed or am not sure what the default was. I haven't changed any settings before upgrading to v10.1.1 as well as before downgrading back to v10.1.0 so this isn't caused by a settings change.

### Relevant Stack Traces

[...]

Issue #2096, opened 2021-07-22
addmailuser is slow when number of mailboxes increases (especially with NFS)


davidcallen:
# Bug Report

## Context

Firstly thanks for a great container. The email software stack must be one of the most complex to implement !

I notice that when calling addmailuser it becomes very slow when number of mailboxes increases.

I did not notice slowness too much when using non-NFS e.g. local HDD or AWS EBS storage

Now I am using NFS (actually AWS EFS) storage since deploying as docker in AWS ECS.

For example I have 2500 mailboxes and a call to addmailuser takes 60 seconds.

I have removed some old inactive mailboxes and now have 500 mailboxes but a call to addmailuser now still takes 30 seconds. Seems too slow to add a single user.

I think the slowness is in the design of check-for-changes.sh.  It detects change in postfix-accounts.cf and then loops through **all** lines in this file, trying to add each user or update its password for each line. This may give good consistency of user data across dovecot etc... but is very inefficient. If it saved a copy of the last successful processing of file postfix-accounts.cf, then it could diff the new file against that copy and detect only one line is new and then create the new user.  That would probably be 0.5 secs or less. 

There is a slight risk that the copy of postfix-accounts.cf becomes incorrect, through script crashing.  To help fix this the start on the container could process the entire postfix-accounts.cf (current behaviour) to fix any config "drift".

Arguably this is not a Bug but a Feature Request.

### What is affected by this bug?

Speed of addmailuser command.

### When does this occur?

When number of mailboxes is high (>=500) and storage is NFS.

### How do we replicate the issue?

1. NFS storage
2. create > 500 mailboxes
3. time execution of addmailuser 

## Your Environment

Using latest EDGE docker image from dockerhub.
AWS ECS docker on EC2 nodes.
EFS storage from all volumes e.g. /var/mail and /tmp/docker-mailserver

### Environment Variables

From a docker ECS service spec :
```CFG
     { "name" : "SSL_TYPE", "value" : "manual" },
      { "name" : "SSL_CERT_PATH", "value" : "/tmp/ssl/mail.qa-mail.tradingapps.com.crt" },
      { "name" : "SSL_KEY_PATH", "value" : "/tmp/ssl/mail.qa-mail.tradingapps.com.key" },
      { "name" : "DMS_DEBUG", "value" : "0" },
      { "name" : "ENABLE_CLAMAV", "value" : "0" },
      { "name" : "ONE_DIR", "value" : "1" },
[...]

PR #1966, opened 2021-05-12
Reduce unnecessary Disk IO

aendeavor:
# Description

See #1958

@nukedupe please double-check this PR too.

Fixes #1958

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Improvement (non-breaking change that does improve existing functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [ ] If necessary I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #2157, opened 2021-08-28
check-for-changes: performance improvements + wait for settle v2

NorseGaud:
Original PR that was [reverted](https://github.com/docker-mailserver/docker-mailserver/pull/2155) : https://github.com/docker-mailserver/docker-mailserver/pull/2104
# Description

See: https://github.com/docker-mailserver/docker-mailserver/issues/2098

This PR will:

- Have `check-for-changes.sh` wait for changes to "settle" (checksum comparison has to not have changed over two different checks with 5 second gaps between) before modifying and restarting anything
- Added backgrounding of processes and using wait throughout instead of sleeps for more efficiency
- Removed the while loop from addmailuser due to check-for-changes.sh logic changes
- Fixed a bug in `_monitored_files_checksums`
- TODO: Add resource usage tests of default setup and ensure the minimum requirements for the container/host specs stay the same (or as close as possible)

This is, in my opinion, critically needed for an API to be performing changes -- and lots of them all at once! See https://github.com/docker-mailserver/docker-mailserver-admin/issues/3 for more discussion about this.

## Type of change

## Checklist:

- [] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)


NorseGaud:
Per https://github.com/docker-mailserver/docker-mailserver/issues/2154 we need to check why so much CPU is used before merging

NorseGaud:
Ok, I was able to reproduce the high CPU usage from the script. For an AWS t2.micro, I was seeing a load average of ~1.00 while running and doing absolutely nothing.

Removing the WAIT_FOR_PIDS and backgrounding from almost everything seems to have fixed it. Load average is ~0.01 again.

NorseGaud:
On boot of the docker container/starting services:

```
 12:39:32 up 23 min,  3 users,  load average: 1.04, 0.59, 0.30
USER     TTY      FROM             LOGIN@   IDLE   JCPU   PCPU WHAT
ec2-user pts/0    inet-64-112-178- 12:17   21:38   1.45s  1.43s top -c
ec2-user pts/1    inet-64-112-178- 12:17    1:32   0.89s  0.65s docker-compose up
ec2-user pts/2    inet-64-112-178- 12:17    0.00s  0.07s  0.00s w
```

Once booted, it drops down but hovers around `0.50`

I'm trying to determine what specifically is causing this.

NorseGaud:
Ok, it looks to be fine after I remove 

```
      for FILE in ${CHANGED}
      do
        case "${FILE}" in
          "/etc/letsencrypt/acme.json" )
              for CERTDOMAIN in ${SSL_DOMAIN} ${HOSTNAME} ${DOMAINNAME}
              do
                {
                  _extract_certs_from_acme "${CERTDOMAIN}" && break
                } &
                WAIT_FOR_PIDS+=($!)
              done
            ;;

          * )
            _notify 'warn' 'File not found for certificate in check_for_changes.sh'
            ;;

        esac
      done
```

Continuing to confirm...

NorseGaud:
Supervisor logs were showing spam:
```
cmp: EOF on /tmp/docker-mailserver-config-chksum.new after byte 596, line 4
```

Turns out the chksum.new file was being written too slowly and the cmp command was trying to read it when it was partially written to. 

it looks like we've wrapped _monitored_files_checksums's code in ( ) and that's likely causing it to execute and continue when we really want it to wait.

NorseGaud:
Odd... check for changes uses a helper that doesn't work...

`set -x` and a few modifications show:

```
+ _monitored_files_checksums
+ echo 'HOSTNAME: deliver'
+ cd /tmp/docker-mailserver
+ exec sha512sum -- postfix-accounts.cf postfix-virtual.cf postfix-aliases.cf dovecot-quotas.cf /etc/letsencrypt/acme.json /etc/letsencrypt/live/deliver/key.pem /etc/letsencrypt/live/deliver/privkey.pem /etc/letsencrypt/live/deliver/fullchain.pem
```

```
root@deliver:/# ls -alht /etc/letsencrypt/live/deliver/
ls: cannot access '/etc/letsencrypt/live/deliver/': No such file or directory
root@deliver:/# ls -alht /etc/letsencrypt/live/
total 20K
drwxr-xr-x 9 root root 6.0K Aug 31 00:00 ..
drwxr-xr-x 2 root root 6.0K Aug 16 00:01 xkr.email
drwxr-xr-x 2 root root 6.0K Aug 16 00:01 pierce.us
drwx------ 4 root root 6.0K Mar 30 02:26 .
-rw-r--r-- 1 root root  740 Aug 11  2020 README
```

Why would anyone generate a letsencrypt with just the hostname and not the domain?

target/scripts/check-for-changes.sh
```diff
@@ -28,217 +28,255 @@ fi
...
 # ? --------------------------------------------- Actual script begins
 
-# determine postmaster address, duplicated from start-mailserver.sh
+# determine postmaster address
 # this script previously didn't work when POSTMASTER_ADDRESS was empty
-if [[ -n ${OVERRIDE_HOSTNAME} ]]
+# used in helper-functions.sh/_monitored_files_checksums
+if [[ -n "${OVERRIDE_HOSTNAME}" ]]
 then
-  DOMAINNAME="${OVERRIDE_HOSTNAME#*.}"
```
NorseGaud, line 35: Maybe I don't understand this, but the old DOMAINNAME we're creating here is wrong:
```
cat docker-mailserver/docker-compose.yml
version: '3.8'

services:
  mailserver:
    build:
      context: .
    hostname: xkr.email
    domainname: xkr.email
    container_name: deliver
    env_file: mailserver.env
```

```
head mailserver.env
# –––––––––––––––––––––––––––––––––––––––––––––––
# ––– Mailserver Environment Variables ––––––––––
# –––––––––––––––––––––––––––––––––––––––––––––––

# empty => uses the `hostname` command to get the mail server's canonical hostname
# => Specify a fully-qualified domainname to serve mail for.  This is used for many of the config features so if you can't set your hostname (e.g. you're in a container platform that doesn't let you) specify it in this environment variable.
OVERRIDE_HOSTNAME=xkr.email
```

```
HOSTNAME: xkr.email
DOMAINNAME: email
```

polarathene (commented):
I've completed my personal notes on this PR, it's prior PRs and other related history in the past regarding the change detection logic. **This PR remains open as a reminder for me to tackle, but it could be closed in favor of a new issue instead.**

The current state of this PR has fallen a fair bit out of sync with the changes/refactoring for the `v10.3.0` release, with more changes to occur again in the near future as `check-for-changes.sh` is refactored some more. Some of the changes this PR covered have also since become part of the `v10.3.0` release, **I don't think it's worth reviving this PR**.

---

Personally, I'm not fond of the approach proposed here with `wait`.

Changes can be detected via FS event notifications, with fallback to polling when that is not supported (_eg: remote FS like NFS_):

[Issue #1958](https://github.com/docker-mailserver/docker-mailserver/issues/1958) was about reducing the rate of calculating checksums for the `changedetector` service, with [PR #1966](https://github.com/docker-mailserver/docker-mailserver/pull/1966/files#diff-05663f5e492c94bf9eb2d1cbc0c2ec5c59b16f79330c44b2ff072c043228d5d7R44) attempting implementing `inotifywait` where it [was noted something regarding volume mounts (_insufficient details provided_) were having trouble](https://github.com/docker-mailserver/docker-mailserver/pull/1966#discussion_r631177614) with `acme.json` but not `postfix-accounts.cf` (_may not have been a bind mount?_) and making a suggestion to swap the `cmp` command for `sha512sum -c --status /path/to/existing/checksum`.

That user also had [a separate PR attempt](https://github.com/docker-mailserver/docker-mailserver/pull/1967/files) with little details that branched PR-1966 to swap `inotifywait` with `fswatch` for some reason. Neither PR was merged, nor a comment regarding their rejection.

---

Additionally when looking into all of this I did a deep dive into file locking and believe the changes made to replace `flock` introduced regressions.

<details>
<summary>Switch back to `flock`</summary>

Both in regards to:
- The **intent of locking files** in the first place - `flock` locked the configs being interacted with, while the replacement applied locks to individual scripts instead, thus locks only blocked a specific script from being concurrently called, but not concurrent access to the actual config files being modified which was the original intent with `flock` being introduced.
- And **performance*/timing* - `flock` had no lockfile to cleanup or timeout on, timeout could be configured via option, as could immediate exit, otherwise it would cause a process to block/wait until a lock was released and could be acquired to continue processing, all with no sleep or similar loop involved..

`flock` was removed only for the benefit of remote filesystems that it was deemed unavailable (NFS) which was inaccurate and referencing outdated information, the error experienced appears to be for different reasons. I additionally looked into `lockf` and `fcntl` as the other common locking alternatives:
- `lockf` is a wrapper around `fcntl` and doesn't seem to offer much benefit.
- `fcntl` has some pro/cons, but is what NFS will use to emulate `flock` with and will need some testing to verify correct behaviour with multiple clients involved.
- And some utilities like `lockfile` (from `procmail` package) that provide a more convenient command and API to use for actual lockfiles if other options like `flock` aren't viable.

File based locks are generally not ideal however as they're not atomic (checking for the lockfile and creating one if it doesn't exist is **not a single operation** and thus prone to some race conditions), `mkdir` is often cited as an atomic variant for that, but using syscalls with `flock` and `fcntl` is preferable from what I came across.

</details>

---

<details>
<summary>Avoiding background processes with wait for parallelization due to concerns</summary>

Using locks properly should resolve that `addmailuser` loop removal just as well AFAIK. The background of processes and wait if I understood correctly was to try parallelize the logic, but this seems unwise as IIRC there is some synchronous flow between some of the logic and you'd be complicating that flow for maintainers for minimal time savings as most of that is related to the service restarts and any sleep delays (_which can be better reduced / optimized via FS event notifications when supported, few users will use remote network file systems like NFS where polling should work just as well as current sleep delays_).

</details>

<details>
<summary>Change Detection (consider `watchexec`)</summary>

Settling on changes is a good improvement for `check-for-changes.sh`, and I think we can leverage a file watcher daemon like [`watchexec`](https://github.com/watchexec/watchexec) for this. It supports both FS event notifications and polling modes, and can debounce (_wait for no changes until X time has passed before running_) although I'm not sure if it can ensure we force processing if debounce stalls processing due to sustained high frequency of changes, which should be unlikely and better served by such users abstracting changes via queue to batch apply rather than directly modifying or calling other utilities (_eg: what the approach by the DMS API project is currently taking_). At least I think `watchexec` would cover everything we'd do and more/better and allow us to offload maintenance for that functionality.

</details>

## TL;DR

- This PR (_and it's history in prior PRs and issues involved_) serves as a good reference but probably not worth resuming or the correct approach to merge.
- File locking appears to have regressed and should bring back `flock`, NFS does actually support it and the failure just needs to be investigated/resolved.
- Detecting changes can avoid the outer loop and it's frequency by using something like `watchexec` with FS event notifications for file changes (or polling mode for NFS), using the debounce feature to get settled changes. I'm unsure currently how beneficial the checksum functionality is to keep (_existing in master, not additions from this PR_) if adopting `watchexec` (_it can inject ENV with values of affected files, but checksum is probably nicer to work with, we can switch from sha512 to blake3 for optimization_).

Additionally (_improvements, but can come later in separate PRs_):
- While I recently de-duplicated the `check-for-changes.sh` logic shared with `start-mailserver.sh` startup scripts, we probably can refactor that some more to diff accordingly as [issue #2096](https://github.com/docker-mailserver/docker-mailserver/issues/2096) notes. (_Note: They are an LDAP + NFS user, noting the `chown -R` perf issue on `/var/mail` which is likely compounding together, not that `check-for-changes.sh` presently runs that for LDAP users however..they also cite adding many users via API service and proposing for a queue in the container as well to batch apply_)
- I also recall a PR some years ago moved away from `hash` type config DB files with Postfix, these could still use text configs but were possible to update at runtime without restarting Postfix IIRC, I'd need to look up that discussion again for why the change was made and if it makes sense to revert.

---

## Actionable Tasks

If I find time for these, it should be around Dec or Jan (excluding "extras" tasks), or someone else is welcome to attempt these before me. I've built up a rather verbose list of notes for when I tackle this, but I'm just as happy to review/collaborate (_just give me a heads up in advance, if you the reader are going to tackle this :sweat_smile: )

- [ ] Use [`watchexec`](https://github.com/watchexec/watchexec) (with debounce) instead for change detection, possibly requiring an ENV for NFS/polling.
- [ ] Look at replacing the existing file locking logic with `lockfile` (from `procmail` package) or better `flock` (_will require reproducing @NorseGaud multi-instance of DMS containers with NFS volumes that produce the bad file descriptor error and resolving that_). Probably more importantly address that files being locked are the actual relevant ones, not the scripts themselves..
- [ ] If keeping the checksum logic around (_eg: the monitor function_), consider using blake3 instead of sha512 for checksums as minor optimization.

Extras:

- [ ] Investigate viability / complexity of only responding to diff changes and/or changing from `text:` to `hash:` based config records for Postfix. Users primary issue may be more related to NFS in combination with `chown -R` (_being investigated and resolved in [PR #2256](https://github.com/docker-mailserver/docker-mailserver/pull/2256)_), but they appear to be an LDAP user with usage for spinning up throwaway images for testing, `check-for-changes.sh` doesn't apply to LDAP at present,  it could be startup related but their API to add users at runtime shouldn't be involved with that only `check-for-changes.sh` :man_shrugging: 
- [ ] Investigate possibility to support batch change requests, potentially via a queue service, though at present this is likely better deferred to the API project, it may be related to file locking concerns depending on implementation. File based queues may be similarly unreliable to in-memory queue system with different failure risks, but the core project is also meant to be DB averse :man_shrugging:

polarathene:
This shouldn't be too important anymore. Many changes were done since to improve how `check-for-changes.sh` works. It is much faster at processing now, with minimal down time.

However, it may not be sufficient for bulk adding of users. Especially for supporting encrypted mailbox creation.

For this I still think debouncing with throttle is better than the current polling approach. My advice in previous comment is still valid for approaching that with `watchexec`. I also still recommend switching the file locking approach back to `flock`.

PR #2158, opened 2021-08-29
sed wrapper

casperklein:
# Description

This introduces `sedfile`, a wrapper for `sed -i` commands.

Problem: When `sed -i` is used to manipulate files, sed exit with error code 0, even when the file was not altered:

```sh
$ cat foo
foo

$ sed -i 's/hello/world/g' foo
$ echo $?
0

$ cat foo
foo
```

`sedfile` compares the file hash before and after the sed operation and fails, if the file was not altered (hashes do match).

```sh
$ sedfile -i 's/hello/world/g' foo
Error: sed -i s/hello/world/g foo

$ echo $?
1
```

This helps to prevent issues caused by failed file modifications, e.g. a package is upgraded and the config file structure changes.

### imap_idle_notify_interval

`Error: sed -i s/#imap_idle_notify_interval = 2 mins/imap_idle_notify_interval = 29 mins/ /etc/dovecot/conf.d/20-imap.conf`
https://github.com/docker-mailserver/docker-mailserver/pull/2158/checks?check_run_id=3455478612#step:6:2236

**Reason:** `imap_idle_notify_interval` is not present in `/etc/dovecot/conf.d/20-imap.conf`.
**Action:** [Line commented out and FIXME label added](https://github.com/docker-mailserver/docker-mailserver/pull/2158/commits/e5a23266f43f897957c5c58e7f864e4569e9a304)

⚠️ Is this entry needed, should it be re-added?
Edit: https://github.com/docker-mailserver/docker-mailserver/pull/554
Edit2: Broken since commit: [added dovecot quota feature](https://github.com/docker-mailserver/docker-mailserver/commit/2527ebfaf2b3eee11b11c4ee589010e8c7a61a99#diff-65abdb9c452354c7d05561295902e8069062b9b5fb42b02633d7deded6f3ce3e)

### adapt mkcert for Dovecot community repo (old leftover)

`Error: sed -i s/CERTDIR=.*/CERTDIR=\/etc\/dovecot\/ssl/g /usr/share/dovecot/mkcert.sh`
`Error: sed -i s/KEYDIR=.*/KEYDIR=\/etc\/dovecot\/ssl/g /usr/share/dovecot/mkcert.sh`
`Error: sed -i 's/KEYFILE=.*/KEYFILE=\$KEYDIR\/dovecot.key/g' /usr/share/dovecot/mkcert.sh`
https://github.com/docker-mailserver/docker-mailserver/runs/3455521780?check_suite_focus=true#step:6:2225

**Reason:** `CERTDIR`, `KEYDIR`, `KEYFILE` already have the correct values in `/usr/share/dovecot/mkcert.sh`.
**Fix:** Affected lines [removed](https://github.com/docker-mailserver/docker-mailserver/pull/2158/commits/9511ef4a52263f663bb96c900212bd5b4bb89fd7) from Dockerfile

## Future to-do

Replace existing `sed -i` with `sedfile -i` in scripts and tests.

## Type of change

## Checklist:


georglauterbach (approved):
LGTM 👍🏼

casperklein:
@egavrilov Your addition from #554 broke some time ago. Do you know, if it's still needed nowadays?

wernerfred (dismissed):

egavrilov:
@casperklein hi there. It's not really a hard requirement to have idle time set from default to whatever value, but just a recomendation to have it higher than default because of battery drain on non-optimised client. I guess it is even possible to remove it because the problem is on the client actually.

casperklein:
> I guess it is even possible to remove it because the problem is on the client actually.

It is already removed (unintentionally) [since ~1.5 years](https://github.com/docker-mailserver/docker-mailserver/commit/2527ebfaf2b3eee11b11c4ee589010e8c7a61a99#diff-65abdb9c452354c7d05561295902e8069062b9b5fb42b02633d7deded6f3ce3e). My question just was, if it's worth to re-add it. Now, I don't think so.
If someone needs that, it can still be configured in a custom dovecot config.

georglauterbach (approved):

polarathene (approved):
LGTM.

Although the sed lines look a bit difficult to read some times, especially that dovecot one with all the path escaping `\/` mixed in with the `/` sed delimiter. In other scripts we switched to a consistent `|` sed delimiter instead.

Not required for this PR but probably a worthwhile change :)

PR #2159, opened 2021-08-29
Update all docker-compose files to use the same version and examples

williamdes:
# Description

Update all docker-compose files to use the same version and examples

## Type of change

## Checklist:


docs/content/examples/tutorials/mailserver-behind-proxy.md
```diff
@@ -54,13 +54,16 @@ Feel free to add your configuration if you archived the same goal using differen
     [...]
     ```
 
-    Truncated list of neccessary labels on the mailserver container:
+    Truncated list of necessary labels on the mailserver container:
```
williamdes, line 57: found a typo

docs/content/examples/tutorials/basic-installation.md
```diff
@@ -25,37 +25,37 @@ We are going to use this docker based mailserver:
     !!! example
 
         ```yaml
-        version: '2'
+        version: '3.8'
 
         services:
-        mail:
+          mailserver:
```
williamdes, line 31: this was an indent mistake

docs/content/examples/tutorials/mailserver-behind-proxy.md
```diff
@@ -54,13 +54,16 @@ Feel free to add your configuration if you archived the same goal using differen
...
 
     ```yaml
-    version: '2'
+    version: '3.8'
     services:
-      mail:
-        image: mailserver/docker-mailserver:release-v7.2.0
```
williamdes, line 63: this was strange fixed tag

wernerfred (commented):
All in all thanks for syncing up all those examples.
But i really dislike inline comments (no benefit except for extra work when automating and scripting stuff)
So if we **really** need those comments at all then please put them in the line above. As most of the occurences are examples imho it is fine to keep `mail` and `example.com` . I think we can expect this level of knowledge transfer capacity of people that want to run their own mailserver.

williamdes:
> So if we really need those comments at all then please put them in the line above. As most of the occurences are examples imho it is fine to keep mail and example.com . I think we can expect this level of knowledge transfer capacity of people that want to run their own mailserver.

Well, let me know what example if best so I can sync all of them
I would like to let some of the comments as users are not always aware values need to be changed even if it is obvious for us having the knowledge

casperklein:
I think this is fine:

```
    hostname: mail
    domainname: example.com
```

README.md
```diff
@@ -197,9 +197,9 @@ version: '3.8'
 
 services:
   mailserver:
-    image: docker.io/mailserver/docker-mailserver:latest
```
georglauterbach, line 200: I would like to keep this as this makes clear that only with DockerHub, we use `mailserver/docker-mailserver`. Another possible address would be `ghcr.io/docker-mailserver/docker-mailserver`.

docs/content/config/security/ssl.md
```diff
@@ -380,24 +380,22 @@ This setup only comes with one caveat: The domain has to be configured on anothe
...
         environment:
           SSL_TYPE: letsencrypt
           SSL_DOMAIN: mail.example.com"
           # for a wildcard certificate, use
           # SSL_DOMAIN: example.com
-      
+
       traefik:
-        image: docker.io/traefik:v2.4.8
+        image: docker.io/traefik:v2.4
```
georglauterbach, line 398: As of now, you can already use `traefik:v2.5`

georglauterbach (changes requested):
Two small changes I proposed. Other than that, LGTM 👍🏼

README.md
williamdes: > Another possible address would be ghcr.io/docker-mailserver/docker-mailserver.

For now, it needs auth..
It's a big blocker

Will update all examples to docker.io then ;)

casperklein (dismissed):

williamdes:
All good now, sorry for the dismissed review
I updated the man example that was not quite up to date

casperklein (approved):

docs/content/examples/tutorials/basic-installation.md
```diff
@@ -25,37 +25,37 @@ We are going to use this docker based mailserver:
...
-            - SPOOF_PROTECTION=0
-            - REPORT_RECIPIENT=1
-            - ENABLE_SPAMASSASSIN=0
-            - ENABLE_CLAMAV=0
-            - ENABLE_FAIL2BAN=1
-            - ENABLE_POSTGREY=0
+                - PERMIT_DOCKER=network
+                - SSL_TYPE=letsencrypt
+                - ONE_DIR=1
+                - DMS_DEBUG=0
```
williamdes, line 49: changed to 0, it's a basic example
First start users do not need debug

docs/content/examples/tutorials/basic-installation.md
```diff
@@ -25,37 +25,37 @@ We are going to use this docker based mailserver:
...
+                - SPOOF_PROTECTION=0
+                - REPORT_RECIPIENT=1
```
williamdes, line 51: By the way, this is deprecated, what should it be changed to ?
The docs are not clear about that

casperklein: What exactly is deprecated?

williamdes: https://docker-mailserver.github.io/docker-mailserver/v10.0/config/environment/#report_recipient-deprecated

`REPORT_RECIPIENT` :)

casperklein:
I've [aligned](https://github.com/docker-mailserver/docker-mailserver/pull/2159/commits/fc3d5e12796cf4476cefe834bb6e592860bbdd06) the _volumes_ in `docs/content/examples/tutorials/basic-installation.md` with our default `docker-compose.yml`.

casperklein (dismissed):

docs/content/config/setup.sh.md
```diff
@@ -66,7 +66,7 @@ OPTIONS
 
         -p PATH
             Provides the config folder path. The default is
-            /home/maxwell/Dokumente/github/docker-mailserver/config/
+            /home/maxwell/Documents/github/docker-mailserver/config/
```
wernerfred, line 69: Is this true? Does `-p` have a default value and why is it prefixed with maxwell 😆

docs/content/examples/tutorials/basic-installation.md
wernerfred: https://docker-mailserver.github.io/docker-mailserver/v10.0/config/environment/#pflogsumm_recipient

`PFLOGSUMM_RECIPIENT` or leave it blank

README.md
wernerfred: the ghcr needs auth for pulling? Never noticed 🙃

docs/content/config/setup.sh.md
georglauterbach: Because I copied it from a VM with a user called `maxwell` 😂 I will change that...

README.md
georglauterbach: @williamdes yes, we should stick to `docker.io` :) But GHCR does not need auth AFAIK as @wernerfred noticed too :D

georglauterbach (approved):

docs/content/config/setup.sh.md
```diff
@@ -66,7 +66,7 @@ OPTIONS
 
         -p PATH
             Provides the config folder path. The default is
-            /home/maxwell/Documents/github/docker-mailserver/config/
+            /${HOME}/Documents/github/docker-mailserver/config/
```
wernerfred: This is still a personalized "default" isn't it?
Is this somewhere in the code or just a "hint" on usage?

I would prefer something like `$(pwd)/config` **but only if** one maps the `./config` in the volumes section.

The current state is a bit misleading imho. Or do i overlook smth?

georglauterbach: You are correct, I did not think about this. The solution using `${PWD}` is better suited, but I guess at best, we should describe it with text (only)?

wernerfred: Text only would be the "most correct" form for every setup imho. Maybe we have a link to the docs at this point (which need to be covered on changes so not optimal as well)

georglauterbach: I'd use plain old text only. This way, we  can avoid confusion and the mistakes spotted above.

williamdes: Lets that that the user is running the command in `<example>` and then after print out the result with a full path

polarathene (changes requested):
This looks great @williamdes ! Thanks for addressing this :grinning: 

---

An observation that I couldn't easily add a review comment on, was some inconsistencies in volumes. The main concern with that being older data volumes mapping vs newer bind mount volumes that we changed to by default this year.

Any of these mappings should take precedence over the previous data volumes usage I think?:

```yaml
    volumes:
      - ./data/maildata:/var/mail
      - ./data/maillogs:/var/log/mail
```

May be easiest to search based on the the internal mapped location, there is bind mount examples using `./data/` prefix followed by `mail`,`state`,logs` as a variant too. Perhaps we should make that consistent?

`./data/mail*` seems redundant vs `./data/*` for the prefix, alternatively unify the config volume under the same hierarchy, ie: `./data/config/`.

Since some compose examples have other services doing local mounts, might want to group those volumes to the context of the service (dms):

- `./dms/config/` and `./dms/data/*` (mail, state, logs).
- or `./data/dms/config/` and `./data/dms/*`.

Some examples only mount some of those volumes, I assume because the others aren't required for the example.

In the `ssl.md` doc, nginx related services are using absolute volume paths like `/mnt/data/nginx`, should probably match relative path as well (except for `/var/run/docker.sock`). The related dms container just has `./mail` and `./mail-state` and mixes in the absolute nginx path to share the cert.

On `basic-installation.md`, I believe @wernerfred might have created this doc from splitting a wiki article of his during the migration to new docs. It needs to be reviewed from the looks of it as it includes the volume mount: `/var/ds/wsproxy/letsencrypt/:/etc/letsencrypt/`, this is described further down [on the page](https://docker-mailserver.github.io/docker-mailserver/edge/examples/tutorials/basic-installation/), but the referenced docker image isn't included in the compose example above, I don't think the absolute path is required? If updating this keep in mind the content on the document outside the config snippet referencing it.

---

As the example `docker-compose.yml` at the root of the git repo has comments to provide guidance on ports, should there also be one above `hostname` to clarify that the two properties define the FQDN `mail.example.com`? I think when I first encountered that (not something I was used to configuring before docker-mailserver), I was initially confused with what to expect.

williamdes:
Hi,
I am too much overbooked now, could other maintainers apply the minor changes that are needed ? 🙏🏻

polarathene:
> I am too much overbooked now

No immediate rush for this AFAIK.

Thanks for letting us know, if someone has the time to spare we'll leave a comment upon picking it up :)

polarathene:
I am now wrapping up this PR.

polarathene (approved):

polarathene:
My additions are too large to mix into this PR. I have some questions aimed at my changes and due to the noise it would add, I think it's better addressed via separate PR than to dirty the review state that reached approval here.

Unfortunately it seems there's some conflicts already, I'll address those and we can approve/merge?

polarathene (approved):
Resolved the conflicts. `setup.sh.md` and  `mailserver-behind-proxy.md` updated to main branch.

docs/content/config/advanced/kubernetes.md
```diff
@@ -176,7 +176,7 @@ spec:
       hostname: mailserver
       containers:
         - name: mailserver
-          image: ghcr.io/docker-mailserver/docker-mailserver:10.0.0
+          image: ghcr.io/docker-mailserver/docker-mailserver:latest
```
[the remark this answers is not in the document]
casperklein: Why not change this to `docker.io`?

docs/content/examples/tutorials/basic-installation.md
casperklein: The space count seems wrong (4 instead of 2) for the ports, volumes and environment items.

docs/content/config/advanced/kubernetes.md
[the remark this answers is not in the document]
polarathene: I have done so on my additions. I marked this as resolved when I committed it.

Originally I was going to push here, but other changes piled up. If you would like I can commit the suggestion here and rebase my changes on top?

docs/content/examples/tutorials/basic-installation.md
polarathene: I have already addressed indent issues in my commits. This file in particular gets it's own separate PR due to it's history and issues with it in general I've resolved, that will follow after my bulk changes PR.

georglauterbach (approved):

wernerfred (approved):

PR #2172, opened 2021-09-05
Update delmailuser to allow mailboxes without TLD

### LINKED ISSUE #2171: [BUG] No valid mail address in remove, after making email.
# Bug Report

## Context

It's possible to use ```addmailuser daniel@test```, but not possible to use ```delmailuser daniel@test```.

addmailuser and delmailuser

### When does this occur?

When forgetting the .com after a mail address.

CLI into container (I'm using this as a swarm service so not using setup.sh.)
1. ```addmailuser daniel@test```
2. ```delmailuser daniel@test```

### Actual Behavior

Addmailuser works, delmailuser fails

### Expected Behavior

*Either* addmailuser shouldn't work *or* delmailuser should work.

## Your Environment

As swarm service
- version: `v10.1.0`
- available RAM: `8GB`

danielvandenberg95:
# Description

Don't allow invalid email addresses in addmailuser.

Fixes #2171

## Type of change

## Checklist:


NorseGaud:
So simple and yet so powerful 👍🏼😎😉

casperklein:
Thank you for your contribution. You can create mailboxes without a TLD. So fixing `delmailuser` instead seems more appropriate to me.

danielvandenberg95:
Changed. (Please squash merge when merging, to prevent commits showing up on addmailuser.)

casperklein (approved):

target/bin/delmailuser
```diff
@@ -91,9 +91,6 @@ for EMAIL in "${@}"
 do
   ERROR=false
 
-  # very simple plausibility check
-  [[ ${EMAIL} != *@*.* ]] && errex "No valid email address: ${EMAIL}"
```
NorseGaud, line 95: Could we instead mimic what addmailuser does so that there is still some validation in this file?

NorseGaud (changes requested):

target/bin/delmailuser
georglauterbach: I already wanted to agree, but then, think about it: If you were able to add addresses, you should be able to remove them, and restrictions would only hinder you here. We have a "closed context" in this case, i.e. we only remove lines in one file.

I'm fine with it either way though:)

georglauterbach (approved):

target/bin/delmailuser
danielvandenberg95: I don't fully get the script below here, is there currently a warning if the mailbox couldn't be found / doesn't exist? If not, I think that should be added, and if so I think this suggestion is covered by that warning?

casperklein: > is there currently a warning if the mailbox couldn't be found / doesn't exist? 

There should be a warning, however the check is currently broken. I've already fixed it locally and will provide a PR once this got merged.

georglauterbach: @NorseGaud are you fine with this change anyway?

danielvandenberg95: In that case I'm resolving this conversation, as that warning will inform users. I'll request a re-review from @NorseGaud.

NorseGaud: @georglauterbach go ahead

georglauterbach:
@casperklein You can go ahead and provide the PR you mentioned in the comments :)

PR #2134, opened 2021-08-14
setup.sh: docker_container first, then fall back to docker_image

NorseGaud:
# Description

Here is an original discussion about this: https://github.com/docker-mailserver/docker-mailserver/pull/1874#issuecomment-809781531

Then there is https://github.com/docker-mailserver/docker-mailserver/pull/2080#discussion_r667359915: As seen in the comment thread, a user experienced confusion and didn't realize `USE_CONTAINER=true` or `-c` was necessary. With my current knowledge I can't see any reason why we wouldn't use the container by default and if it exists, fall back to using the default image/tag when it doesn't.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Improvement (non-breaking change that does improve existing functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [ ] If necessary I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

Issue #2140, opened 2021-08-18
[FR] default user from env variable for unit tests


cyberb:
I am am trying to use maileserver in CI for testing IMAP connectivity. For CI I use Drone CI services which only allows environment variable based configuration.
It would be very useful to allow adding a user using env var for such use case.

NorseGaud:
There is really no way to execute something inside of the container? If so, you can just do what we do here in the tests: https://github.com/docker-mailserver/docker-mailserver/blob/master/test/tests.bats#L43

NorseGaud:
Also, what about pre-creating the users/config locally and then using the config directory by mounting it in your CI system?

polarathene:
A custom entry point for your container (or image extending the official one) should allow you to run any logic needed if your CI service provides no other means beyond ENV.

cyberb:
Right, service can either run default entrypoint and detach or custom commands and then detach. In either case it can also provide envs.

As I understand now I can run 'addmailuser' command and then start server. If that is correct can you tell me what is the right command to start the server in this case?

NorseGaud:
> Right, service can either run default entrypoint and detach or custom commands and then detach. In either case it can also provide envs.
> 
> As I understand now I can run 'addmailuser' command and then start server. If that is correct can you tell me what is the right command to start the server in this case?

No, sorry for the confusion, You must run the command after it is started. At the moment this is no ENV support for creating a user.

polarathene:
> As I understand now I can run 'addmailuser' command and then start server. If that is correct can you tell me what is the right command to start the server in this case?

Looking at one of my projects that does this, I add the following to docker-compose for the `docker-mailserver` container:

```yml
volumes:
  - $PWD/data/mailserver/entrypoint.sh:/entrypoint.sh:ro
entrypoint: /entrypoint.sh
```

And this is the entrypoint override:

```sh
#!/bin/sh
set -e

# Container needs an account by default to run, docker-mailserver CI automates this for testing a bit differently:
# https://github.com/docker-mailserver/docker-mailserver/blob/df54750d02a9a30c472226d1b1030f21d490b704/.github/workflows/test_merge_requests.yml#L35
# That CI approach doesn't use the convenience method here, but more direct generation perhaps test-specific:
# https://github.com/docker-mailserver/docker-mailserver/blob/df54750d02a9a30c472226d1b1030f21d490b704/Makefile#L39
[...]

PR #2174, opened 2021-09-06
Move setup process via script into container

georglauterbach:
# Description

Decoupling setup process from `setup.sh` script by introducing a setup script _inside_ the container that coordinates the setup process. **This is not a breaking change**. This way, we do not have to keep track of versions of `setup.sh`, which was said to be a no-no [here](https://www.ifthenel.se/self-hosted-mail-server/). This is a blog post we're linking to in our documentation. This change brings the additional benefit for Kubernets users to be able to make use of `setup` now, without the need for `setup.sh`.

The maintainers discussion referencing / talking about the current `setup.sh` versioning-problem can be found [here](https://github.com/orgs/docker-mailserver/teams/maintainers/discussions/24).

This PR was really made possible by @NorseGaud with #2134 

## Type of change

## Checklist:


georglauterbach:
MkDocs does not seem to render \~\~ (crossed out) correctly. @wernerfred @polarathene do you have any idea how to fix this? I mean, it's just a joke in the docs and we can remove it, but it makes me curious.

docs/content/config/setup.sh.md
```diff
@@ -13,13 +13,9 @@ wget https://raw.githubusercontent.com/docker-mailserver/docker-mailserver/maste
...
 ```
 
-!!! info
-
-    Make sure to get the `setup.sh` that comes with the release you're using. Look up the release and the git commit on which this release is based upon by selecting the appropriate tag on GitHub. This can done with the "Switch branches/tags" button on GitHub, choosing the right tag. This is done in order to rule out possible inconsistencies between versions.
-
 ## Usage
 
-Run `./setup.sh help` and you'll get some usage information:
+Run `./setup.sh help` and you'll get ~~all you have ever wanted~~ some usage information:
```
polarathene, line 18: This should work according to [upstream docs](https://squidfunk.github.io/mkdocs-material/reference/formatting/#highlighting-text).

It uses [this plugin](https://facelessuser.github.io/pymdown-extensions/extensions/tilde/ AFAIK, ah scrolling up on the original upstream docs link [mentions we have to enable it](https://squidfunk.github.io/mkdocs-material/reference/formatting/#caret-mark-tilde) in `mkdocs.yml` config:

```yaml
markdown_extensions:
  - pymdownx.caret
  - pymdownx.mark
  - pymdownx.tilde
```

Add `- pymdownx.tilde` into this part of the config:

https://github.com/docker-mailserver/docker-mailserver/blob/0da66ccb34dcaf0663b6b5635ba97f990b761881/docs/mkdocs.yml#L68

README.md
```diff
@@ -76,28 +76,17 @@ All workflows are using the tagging convention listed below. It is subsequently
...
-
 ``` BASH
-# if you're using :edge as the image tag
-wget https://raw.githubusercontent.com/docker-mailserver/docker-mailserver/master/setup.sh
-# if you're using :latest (= :10.1.1) as the image tag
-wget https://raw.githubusercontent.com/docker-mailserver/docker-mailserver/v10.1.2/setup.sh
+DMS_GITHUB_URL='https://raw.githubusercontent.com/docker-mailserver/docker-mailserver/master'
+wget "${DMS_GITHUB_URL}/docker-compose.yml"
+wget "${DMS_GITHUB_URL}/mailserver.env"
+wget "${DMS_GITHUB_URL}/setup.sh"
```
polarathene, line 83: Is this last step with the `setup.sh` necessary still?

Can you run `docker exec ...` as an alternative now? (seem to be missing docs example on that if an option).

Is the main intent to still interact with this external script but persist version specific changes internally with releases?

georglauterbach: Yes, you can run `docker exec .. setup ...` and it should work fine. The mail intend of `setup.sh` in my eyes is to wrap the interaction with the container for many users so they do not have to know the name of the container, etc.

But I agree with the fact that we should note that one could easily run `docker exec ... setup ...` as well.

polarathene: While tackling this, do you think it's worth supporting [this issue request](https://github.com/docker-mailserver/docker-mailserver/issues/2140#issuecomment-902296722) to add a default user when an ENV is set?

Their use case was for CI where they apparently could not easily initialize a user via invoking an explicit script/command. I shared how I added an entry point to automate this for my local testing, we could do similar by adopting an entrypoint with that logic?

If we use an entrypoint in the Dockerfile, then no args can keep the current behaviour, otherwise we could take args provided in a `docker run` and pass them to this internal script now? (I have had some experience with that a while back, [example shared here](https://github.com/gatsbyjs/gatsby-docker/pull/19#issuecomment-449576344))

May be worth it, as that recent blogpost you linked to also mentioned setting up a user as a gripe, having users provide an explicit ENV in their compose file for an admin or dummy user can simplify getting started without needing to get into `setup.sh` as early on?

georglauterbach: To be honest, I'd rather not like to implement this behaviour in this PR for two reasons:

1. I'm not sure what the solution should look like in the end
2. I'd like to not mix up concerns with PRs. This PR can (and should in my opinion) the base for the suggested changes, but implement them too:)

docs/content/config/setup.sh.md
georglauterbach: Done. [20502294be003363e7f324a8e49166c7085324e4]

README.md
georglauterbach: Added the note with a534bf6f02f522ffd229ce1444d08e205988f9de

README.md
```diff
@@ -76,28 +76,21 @@ All workflows are using the tagging convention listed below. It is subsequently
...
 
 chmod a+x ./setup.sh
-
-# and make yourself familiar with the script
 ./setup.sh help
 ```
 
-**Make sure to get the `setup.sh` that comes with the release you're using**. Look up the release and the git commit on which this release is based upon by selecting the appropriate tag on GitHub. This can done with the "Switch branches/tags" button on GitHub, choosing the right tag. This is done in order to rule out possible inconsistencies between versions.
+Running `./setup.sh help` will also pull the `:latest` image (which is the latest stable release), if this has not been done beforehand.
```
polarathene, line 91: Is this accurate?

Looking at the file (with my stale shell skills) any command used seems to call the container method which will use an existing running container (including`help`)?

It's only when a container isn't running that it pulls the latest `:latest` tag from a registry?

The last part "if this has not been done beforehand", would it not remove the container after satisfying the command in this case, so repeat the same flow by checking the registry again to pull?

polarathene (changes requested):

README.md
georglauterbach: You're quite right.

> The last part "if this has not been done beforehand", would it not remove the container after satisfying the command in this case, so repeat the same flow by checking the registry again to pull?

Could you rephrase this question for me. Not I can follow here.

polarathene: > Accurate.

Ok, so `./setup.sh help` may not actually pull `:latest` if it detects an existing DMS container running. That should be rephrased then.

> Could you rephrase this question for me. Not sure I can follow here.

If no container is running, and I run `./setup.sh help`, then a docker registry is checked for newer `:latest` tag to pull, container is created to run the command, and then exits.

Calling it again will play out the same sequence, unless the digest of `:latest` on the registry is updated.

I guess "if this has not been done beforehand" is hinting that the pull download could potentially be skipped? (`:latest` has been not updated since the last registry check).

---

Could just be me, I'm not fussed if you think the text is fine as-is :)

Two attempts at re-phrasing it:

```suggestion
If no `docker-mailserver` container is running, any `./setup.sh` command will check online for the `:latest` image tag (the current stable release), performing a `pull` if necessary followed by running the command in a temporary container. 
```

or

```suggestion
`./setup.sh help` (_or any supported command_) with no active `docker-mailserver` container will _pull_ the latest stable release (_`:latest` image tag_) before running the command.
```

georglauterbach: Went with the former :)

polarathene (approved):

target/bin/setup
```diff
@@ -0,0 +1,171 @@
...
+function _main
+{
+  case ${1:-} in
+
+    email )
+      case ${2:-} in
+        add      ) shift 2 ; addmailuser "${@:+$@}" ;;
+        update   ) shift 2 ; updatemailuser "${@:+$@}" ;;
+        del      ) shift 2 ; delmailuser "${@:+$@}" ;;
+        restrict ) shift 2 ; restrict-access "${@:+$@}" ;;
```
casperklein, line 109:
```suggestion
        add      ) shift 2 ; addmailuser "${@}" ;;
        update   ) shift 2 ; updatemailuser "${@}" ;;
        del      ) shift 2 ; delmailuser "${@}" ;;
        restrict ) shift 2 ; restrict-access "${@}" ;;
```

Remove MacOS specific quirks, intruduced in #2001

target/bin/setup
```diff
@@ -0,0 +1,171 @@
...
+        del      ) shift 2 ; delalias "${1}" "${2}" ;;
+        list     ) shift 2 ; listalias ;;
+        *        ) _invalid_command "${1}" "${2}" ;;
+      esac
+      ;;
+
+    quota )
+      case ${2:-} in
+        set      ) shift 2 ; setquota "${@:+$@}" ;;
+        del      ) shift 2 ; delquota "${@:+$@}" ;;
```
casperklein, line 127:
```suggestion
        set      ) shift 2 ; setquota "${@}" ;;
        del      ) shift 2 ; delquota "${@}" ;;
```

target/bin/setup
```diff
@@ -0,0 +1,171 @@
...
+        *        ) _invalid_command "${1}" "${2}" ;;
+      esac
+      ;;
+
+    config )
+      case ${2:-} in
+        dkim     ) shift 2 ; open-dkim "${@:+$@}" ;;
```
casperklein, line 134:
```suggestion
        dkim     ) shift 2 ; open-dkim "${@}" ;;
```

target/bin/setup
```diff
@@ -0,0 +1,171 @@
...
+        *        ) _invalid_command "${1}" "${2}" ;;
+      esac
+      ;;
+
+    relay )
+      case ${2:-} in
+        add-domain     ) shift 2 ; addrelayhost "${@:+$@}" ;;
+        add-auth       ) shift 2 ; addsaslpassword "${@:+$@}" ;;
+        exclude-domain ) shift 2 ; excluderelaydomain "${@:+$@}" ;;
```
casperklein, line 143:
```suggestion
        add-domain     ) shift 2 ; addrelayhost "${@}" ;;
        add-auth       ) shift 2 ; addsaslpassword "${@}" ;;
        exclude-domain ) shift 2 ; excluderelaydomain "${@}" ;;
```

target/bin/setup
```diff
@@ -0,0 +1,171 @@
...
+        *              ) _invalid_command "${1}" "${2}" ;;
+      esac
+      ;;
+
+    debug )
+      case ${2:-} in
+        fetchmail      ) debug-fetchmail ;;
+        fail2ban       ) shift 2 ; fail2ban "${@:+$@}" ;;
```
casperklein, line 151:
```suggestion
        fail2ban       ) shift 2 ; fail2ban "${@}" ;;
```

target/bin/setup
```diff
@@ -0,0 +1,171 @@
...
+        show-mail-logs ) cat /var/log/mail/mail.log ;;
+        login          )
+          shift 2
+          if [[ -z ${1:-} ]]
+          then
+            /bin/bash
+          else
+            /bin/bash -c "${@:+$@}"
```
casperklein, line 159:
```suggestion
            /bin/bash -c "${@}"
```

target/bin/setup
```diff
@@ -0,0 +1,171 @@
...
+        * ) _invalid_command "${1}" "${2}" ;;
+      esac
+      ;;
+
+    help ) _usage ;;
+    *    ) _invalid_command "${*}" ;;
+  esac
+}
+
+_main "${@:+$@}"
```
casperklein, line 171:
```suggestion
_main "${@}"
```

casperklein (changes requested):
See my suggestions to remove MacOS specific quirks, intruduced in #2001

target/bin/setup
georglauterbach: Ah, I did not think of this myself. Thanks!

target/bin/setup
```diff
@@ -0,0 +1,171 @@
...
+        add      ) shift 2 ; addalias "${1}" "${2}" ;;
+        del      ) shift 2 ; delalias "${1}" "${2}" ;;
+        list     ) shift 2 ; listalias ;;
+        *        ) _invalid_command "${1}" "${2}" ;;
+      esac
+      ;;
+
+    quota )
+      case ${2:-} in
+        set      ) shift 2 ; setquota "${@}" ;;
```
NorseGaud, line 126: I seem to remember that "${@:+$@}" was required here for mac support. I can run a full set of the tests before you to make sure if you're not worried about it for now

georglauterbach: But isn't this running inside the container? It should therefore not be dependent on macOS, right?

casperklein (approved):

polarathene (approved):

setup.sh
```diff
@@ -288,33 +187,29 @@ function _main
...
-      i ) IMAGE_NAME="${OPTARG}" ;;
-      z ) USE_SELINUX=":z"     ;;
-      Z ) USE_SELINUX=":Z"     ;;
-      c )
-        # container specified, connect to running instance
-        CONTAINER_NAME="${OPTARG}"
-        ;;
+      ( i ) IMAGE_NAME="${OPTARG}"       ;;
+      ( z | Z ) USE_SELINUX=":${OPTARG}" ;;
+      ( c ) CONTAINER_NAME="${OPTARG}"   ;;
```
casperklein, line 195: @georglauterbach

Just noticed this: Why are there all these '(' added in the case statement?

georglauterbach: It seems like no one knows this syntax :D I have seen it a little while ago by accident and my brace highlighting extensions liked it too. What do you think of it? I personally like the braces on both sidesy

NorseGaud: I'm ok with it as long as it's not a requirement for the rest of us. I've never seen or used it before myself, so I'd like to not have to remember and change my existing habits :)

casperklein: TIL that this is [valid syntax](https://www.gnu.org/software/bash/manual/bash.html#index-case) 🎉
IMO we should stick to well-known syntax patterns, as this will probably confuse everyone on a first look.

On the other hand, we can see that as a security feature. Bad guys analyzing our scripts to find vulnerabilities will immediately think: "WTF is this. I probably should not mess with guys that use such a syntax.". 😆

polarathene: > "WTF is this. I better should not mess with guys that use such a syntax."

:joy: 

> What do you think of it?

I like it. Shell syntax has always been weird looking to me though with stuff like `case esac` and `if then fi`, but I prefer parenthesis on both ends to wrap a case condition, little less foreign looking.

georglauterbach: I'm fine with everyone using his/her own style in this special instance:) We're not linting this, and I and @polarathene as well as @casperklein and @NorseGaud can use their preferred syntax too :D

casperklein: Yeah. What confuse me, in BASH `()` means subshell. So for me it feels not consistent, using () for other things. Apart from that, `()` looks nicer.

matrixes:
I have updated the blog post to cover the new conditions regarding setup. I'd also like to take the opportunity to thank you for your efforts!

PR #2183, opened 2021-09-11
Lock file create and remove improvements

NorseGaud:
# Description

Problem: When using multiple containers all sharing the same network volume for `/tmp/docker-mailserver`, I notice that locking doesn't work properly. This is due to flawed logic (my logic) when creating the lock files. I also noticed that orphaned locks can be left if the container is interrupted by a human or crashes.

This PR introduces:

- A unique ID in the lock file so that `remove_lock` can't remove something that was not created earlier in the same process/instance.
- A while-loop inside of the `create_lock`, making it easier to use and preventing it from exiting the entire check-for-changes (leading to heavy CPU usage to re-boot the script).
- Tests for the locking used in check-for-changes.sh
- Locking is only needed when there is a chance, so I've moved the function calls inside of the `[ $? -eq 1 ]`'s if

## Type of change

## Checklist:


target/scripts/check-for-changes.sh
```diff
@@ -7,6 +7,8 @@ LOG_DATE=$(date +"%Y-%m-%d %H:%M:%S ")
 _notify 'task' "${LOG_DATE} Start check-for-changes script."
 
 SCRIPT_NAME="$(basename "$0")"
+LOCK_ID="${RANDOM}" # Required for create_lock uniqueness and cleanup
+trap "remove_lock ${SCRIPT_NAME} ${LOCK_ID}" EXIT
```
polarathene: This is only case where a `create_lock` line is avoided from the prior pattern used. I understand that is because unlike other cases, `create_lock` is being called in a loop below instead and only if a change is detected.

The comment for `LOCK_ID` is not on the other definitions for it: `# Required for create_lock uniqueness and cleanup`. To clarify this is for uniqueness of the `create_lock` below from other locks right? The `create_lock` below does not need to be unique if called more than once in that loop, it is intended to reuse the same `LOCK_ID` defined here?

Your commit messages and PR description do not clarify why the `create_lock` and `remove_lock` were moved into the conditional `[$? -eq 1 ]`?

target/bin/addmailuser
```diff
@@ -44,7 +44,11 @@ PASSWD="${*}"
 
 # Protect config file with lock to avoid race conditions
 touch "${DATABASE}"
-create_lock "$(basename "$0")"
+SCRIPT_NAME="$(basename "$0")"
+LOCK_ID="${RANDOM}"
```
casperklein, line 48: Why not move this two lines into the create_lock function?

NorseGaud: Both are unique to the script executing the create and remove_lock functions. Maybe I misunderstand but I think if I move them into the function SCRIPT_NAME would be `helper-functions.sh` and the LOCK_ID wouldn't be able to be set with the ID for the specific instance of the script.

NorseGaud:
I'll remove the bats changes tomorrow. Forgot about those and I'm done coding for the day :)

polarathene (commented):
This sort of thing is why I'd prefer we didn't rely on shell scripts so much :sweat_smile:

I'd much rather use a language that's easier to grok and [provides such functionality in an easy to use manner](https://crates.io/crates/file-lock). I don't have too much input, just some questions.

`check-for-changes.sh` should probably adopt `|` with `sed`, and I assume switch to `sedfile`? Same for `helper-functions.sh` although I'm not sure if `sedfile` is relevant for `sed -n` operations, @casperklein would know :)

[`sed -n`](https://linux.die.net/man/1/sed):

> -n, --quiet, --silent
> suppress automatic printing of pattern space

target/bin/addmailuser
casperklein: Your statement would be true, if `helper-functions.sh` would be called directly.
But in this case, `helper-functions.sh` is sourced. You can consider its content as part of the current file, therefore no problems with `$0` etc.

polarathene: > Both are unique to the script executing the create and remove_lock functions.

I believe the question was due to these vars not being used anywhere else in the script, they're serving this purpose for locking only and the repetition in other files establishes a pattern that could be DRY instead.

> SCRIPT_NAME would be `helper-functions.sh`

Not if you pass `"$0"` as the arg, or `"$(basename "$0")"` if that's not the case?

> LOCK_ID wouldn't be able to be set with the ID for the specific instance of the script.

It's a random value that is used on both `create_lock` and the trap for `remove_lock`, if those are both wrapped into a helper function as well, that should be a non-issue?

There is one other usage so far where you avoid `create_lock` early and use the values later that would not utilize this helper/wrapper method I guess.

polarathene:
From original PR message:

> It also introduces

Did you have more to say?

target/bin/addmailuser
```diff
@@ -44,7 +44,10 @@ PASSWD="${*}"
...
+create_lock "${SCRIPT_NAME}" "${LOCK_ID}"
+trap "remove_lock ${SCRIPT_NAME} ${LOCK_ID}" EXIT
```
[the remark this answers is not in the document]
casperklein: I noticed the same. See my comment above.

`basename` strips of the path, leaving just the file name.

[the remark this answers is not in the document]
casperklein: `$RANDOM` --> integer in the range 0 - 32767 (it uses /dev/urandom).

BTW: If a unique value is wanted, `uuid` command would be a better choise.

target/bin/addmailuser
NorseGaud: > Your statement would be true, if `helper-functions.sh` would be called directly.
> But in this case, `helper-functions.sh` is sourced. You can consider its content as part of the current file, therefore no problems with `$0` etc.

Oh DERP! You're right. Silly mistake. Thanks for catching that

NorseGaud: > > Both are unique to the script executing the create and remove_lock functions.
> 
> I believe the question was due to these vars not being used anywhere else in the script, they're serving this purpose for locking only and the repetition in other files establishes a pattern that could be DRY instead.
> 
> > SCRIPT_NAME would be `helper-functions.sh`
> 
> Not if you pass `"$0"` as the arg, or `"$(basename "$0")"` if that's not the case?
> 
> > LOCK_ID wouldn't be able to be set with the ID for the specific instance of the script.
> 
> It's a random value that is used on both `create_lock` and the trap for `remove_lock`, if those are both wrapped into a helper function as well, that should be a non-issue?
> 
> There is one other usage so far where you avoid `create_lock` early and use the values later that would not utilize this helper/wrapper method I guess.

I don't remember why, but I thought I removed the trap from create_lock because it was breaking something. We should set it in there and not have to call it from the various scripts. I'll add it back and test everything.

target/bin/addmailuser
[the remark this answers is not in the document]
NorseGaud: Thanks! Reverted a bunch of code and added uuid

target/scripts/check-for-changes.sh
NorseGaud: > Your commit messages and PR description do not clarify why the create_lock and remove_lock were moved into the conditional [$? -eq 1 ]?

Added, thanks!

> The comment for LOCK_ID is not on the other definitions for it: # Required for create_lock uniqueness and cleanup. To clarify this is for uniqueness of the create_lock below from other locks right? The create_lock below does not need to be unique if called more than once in that loop, it is intended to reuse the same LOCK_ID defined here?

I don't know what I was thinking actually... Moved the lock ID into the helper

NorseGaud:
Do y'all know if creating a trap inside of a while loop would cause memory or something to eventually fill up?
We use `trap remove_lock EXIT` inside of the `create_lock` and if check-for-changes.sh runs for a loooong time, the trap will never be triggered. I'm hoping that trapping with the same function and signal just overwrites existing traps.

polarathene:
> if check-for-changes.sh runs for a loooong time

What would cause it to run for a long duration?

---

> I'm hoping that trapping with the same function and signal just overwrites existing traps.

Glancing over [this](https://unix.stackexchange.com/questions/170409/bash-issuing-trap-twice) and [this](https://stackoverflow.com/questions/3338030/multiple-bash-traps-for-the-same-signal) seem to suggest that is the case.

In your earlier commits where you supplied args I'm not sure, but I think that would override/conflict with any other traps on the same signal, the last trap would apply with it's specific args to the function.

target/bin/updatemailuser
```diff
@@ -27,8 +27,9 @@ fi
 
 HASH="$(doveadm pw -s SHA512-CRYPT -u "${USER}" -p "${PASSWD}")"
 
-# Protect config file with lock to avoid race conditions
 touch "${DATABASE}"
-create_lock "$(basename "$0")"
+# Protect config file with lock to avoid race conditions
+LOCK_ID="$(uuid)"
```
polarathene, line 32: I assume `LOCK_ID` is no longer meant to be here and was missed for removal?

target/scripts/helper-functions.sh
```diff
@@ -1,6 +1,8 @@
 #! /bin/bash
 
 DMS_DEBUG="${DMS_DEBUG:=0}"
+SCRIPT_NAME="$(basename "$0")"
```
polarathene, line 4: Probably add a comment that it refers to the script that sourced `helper-functions.sh`, not `helper-functions.sh`, which was believed to be the case by yourself earlier (and for maintainers like me that are shell-phobic :sweat_smile: ).

target/scripts/helper-functions.sh
```diff
@@ -1,6 +1,8 @@
...
+LOCK_ID="$(uuid)" # Required for create_lock uniqueness and cleanup
```
polarathene, line 5: I was going to say this had become a global now, but due to the clarification of how `SCRIPT_NAME` above works with it's value, I guess this is locally scoped to each file sourcing `helper-functions`, for the duration of that scripts execution? (_same `LOCK_ID` for all `create_lock` calls by that parent script_)

```suggestion
LOCK_ID="$(uuid)" # Restricts the scope of create_lock and remove_lock calls to the sourcing parent script
```

Probably a better way to phrase that :sweat_smile:

polarathene (changes requested):

target/bin/updatemailuser
NorseGaud: 100%

target/scripts/helper-functions.sh
NorseGaud: Can do

target/scripts/helper-functions.sh
NorseGaud: I added something that makes it even clearer if that's ok

georglauterbach:
> This sort of thing is why I'd prefer we didn't rely on shell scripts so much :sweat_smile:
> 
> I'd much rather use a language that's easier to grok and [provides such functionality in an easy to use manner](https://crates.io/crates/file-lock). I don't have too much input, just some questions.

I was so hoping for this link to be one to a Rust-related website - and indeed, made me happy. If someone decides to try and implement something in Rust, let me know 👍🏼

> `check-for-changes.sh` should probably adopt `|` with `sed`, and I assume switch to `sedfile`? Same for `helper-functions.sh` although I'm not sure if `sedfile` is relevant for `sed -n` operations, @casperklein would know :)

Having uniform delimiters would be very nice :D

polarathene (approved):
LGTM.

Would be good to [address style consistency with the `sed` delimiter](https://github.com/docker-mailserver/docker-mailserver/pull/2183#pullrequestreview-752037735) for the files you're touching.

That's just:

1. `s/A/B/` to `s|A|B|`
2. Any `sed` using the `-i` option that always expects to change file content to use `sedfile -i` to prevent silent failures. If a modification of a file is conditional/optional it should not use `sedfile`.

NorseGaud:
Well, this is odd... We can't run `uuid` in `helper-functions.sh` because it's being executed on the host. Looking for a solution... If anyone has any recommendations, I'm all-ears

casperklein:
> I'm not sure if sedfile is relevant for sed -n operations, @casperklein would know :)

No it isn't. Only when `-i` is involved, using `sedfile` makes sense.

test/helper-functions.bats
```diff
@@ -0,0 +1,43 @@
...
+    PRIVATE_CONFIG="$(duplicate_config_for_container .)"
+    docker run -d --name mail_helper_functions \
+		-v "${PRIVATE_CONFIG}":/tmp/docker-mailserver \
+		-v "$(pwd)/test/test-files":/tmp/docker-mailserver-test:ro \
+		-e ENABLE_FETCHMAIL=1 \
+		--cap-add=NET_ADMIN \
+		-e DMS_DEBUG=0 \
+		-h mail.my-domain.com -t "${NAME}"
+    wait_for_finished_setup_in_container mail_helper_functions
+}
```
polarathene, line 22: Seemed a bit weird having `--cap-add` between the two env options. Main concern was mixed indentation (spaces and tabs), fixed:

```suggestion
function setup_file() {
    local PRIVATE_CONFIG
    PRIVATE_CONFIG="$(duplicate_config_for_container .)"
    docker run -d --name mail_helper_functions \
        --cap-add=NET_ADMIN \
        -v "${PRIVATE_CONFIG}":/tmp/docker-mailserver \
        -v "$(pwd)/test/test-files":/tmp/docker-mailserver-test:ro \
        -e ENABLE_FETCHMAIL=1 \
        -e DMS_DEBUG=0 \
        -h mail.my-domain.com -t "${NAME}"
    wait_for_finished_setup_in_container mail_helper_functions
}
```

test/helper-functions.bats
```diff
@@ -0,0 +1,43 @@
...
+
+function teardown_file() {
+    docker rm -f mail_helper_functions
+}
+
+@test "first" {
+  skip 'this test must come first to reliably identify when to run setup_file'
+}
```
polarathene, line 30: Match same indentation of level of 4 spaces per tab in this file.

Yes I'm aware the other bats file in this PR is using 2 spaces instead, if someone cares to normalize that via a lint tool later great, but for now I only care about file level consistency :)

```suggestion
@test "first" {
    skip 'this test must come first to reliably identify when to run setup_file'
}
```

test/helper-functions.bats
```diff
@@ -0,0 +1,43 @@
...
+    assert_output "0.0.0.0/0"
+    run docker exec mail_helper_functions bash -c ". /usr/local/bin/helper-functions.sh; _sanitize_ipv4_to_subnet_cidr 192.168.255.14/20"
+    assert_output "192.168.240.0/20"
+    run docker exec mail_helper_functions bash -c ". /usr/local/bin/helper-functions.sh; _sanitize_ipv4_to_subnet_cidr 192.168.255.14/32"
+    assert_output "192.168.255.14/32"
+}
+
+@test "last" {
+  skip 'this test is only there to reliably mark the end for the teardown_file'
+}
```
polarathene, line 43: Indent spaces change from 2 to 4.

```suggestion
@test "last" {
    skip 'this test is only there to reliably mark the end for the teardown_file'
}
```

polarathene (changes requested):

polarathene:
Same test failure:

```
not ok 329 setup.sh :: exit with error when no arguments provided
# (from function `assert_line' in file test/test_helper/bats-assert/src/assert.bash, line 444,
#  in test file test/tests.bats, line 946)
#   `assert_line --index 0 --partial "The command '' is invalid."' failed
# 
# -- line does not contain substring --
# index     : 0
# substring : The command '' is invalid.
# line      : SETUP(1)
# --
# 
not ok 330 setup.sh :: exit with error when wrong arguments provided
# (from function `assert_line' in file test/test_helper/bats-assert/src/assert.bash, line 444,
#  in test file test/tests.bats, line 952)
#   `assert_line --index 0 --partial "The command 'lol troll' is invalid."' failed
# 
# -- line does not contain substring --
# index     : 0
# substring : The command 'lol troll' is invalid.
# line      : SETUP(1)
# --
# 
```

That is the [same failure with the recent PR merged to revert `setup.sh`](https://github.com/docker-mailserver/docker-mailserver/pull/2188#issuecomment-917629292). Seems that they can be ignored as the [prior PR with `setup.sh` v10.2 migrated into the container had it's tests adjusted](https://github.com/docker-mailserver/docker-mailserver/pull/2174/files#diff-c65d41b8ad46ee079a476545e7816994ee96c554eeac057e58c9f430f568b243L946-L952) (and these weren't reverted).

NorseGaud:
Thanks @polarathene. @casperklein, do you have plans to fix the tests?

polarathene:
> do you have plans to fix the tests?

The mentioned failures are non-issue. It's due to recent `setup.sh` PR being partially reverted temporarily because of user complaint. The affected tests were modified for the `setup.sh` change, but were not reverted back. It will resolve itself when preparing for v10.2 release, can safely be ignored.

polarathene (approved):

georglauterbach:
@polarathene is right; I had to adjust tests for the new command:) Test failures are fine in this case :)

georglauterbach (approved):

wernerfred (approved):

target/scripts/helper-functions.sh
```diff
@@ -17,17 +19,33 @@ function escape
...
+  echo "${LOCK_ID}" > "${LOCK_FILE}"
 }
 
 function remove_lock
 {
-  SCRIPT_NAME=${SCRIPT_NAME:-$1}
-  rm -f "/tmp/docker-mailserver/${SCRIPT_NAME}.lock"
+  LOCK_FILE="${LOCK_FILE:-"/tmp/docker-mailserver/${SCRIPT_NAME}.lock"}"
+  [[ -z "${LOCK_ID}" ]] && errex "Cannot remove ${LOCK_FILE} as there is no LOCK_ID set"
+  if [[ -e "${LOCK_FILE}" && $(grep -c "${LOCK_ID}" "${LOCK_FILE}") -gt 0 ]] # Ensure we don't delete a lock that's not ours
```
casperklein, line 44: Instead of `$(grep -c "${LOCK_ID}" "${LOCK_FILE}") -gt 0`

`grep -q "${LOCK_ID}" "${LOCK_FILE}"`

NorseGaud: @casperklein , should we add it to the 10.2.0 branch?

casperklein: No veto from me 😃

PR #2187, opened 2021-09-12
docs: Improve FAQ entry for `mail-state` folder

polarathene:
# Description

- Links to relevant script logic.
- Better list of services data moved to `mail-state`.

## Type of change

## Checklist:


casperklein (approved):
Maybe add further information why mail-state is useful: It preserves the service data between container restarts. Otherwise fail2ban information (blocks) are lost and clamav patterns need to be downloaded again and may not up2date for a time period.

polarathene:
> why mail-state is useful: It preserves the service data between container restarts.

"this folder will store the data from internal services so that you can more easily persist state." should communicate that?

> Otherwise fail2ban information (blocks) are lost and clamav patterns need to be downloaded again and may not up2date for a time period.

Sure some examples can be added I guess :+1:

docs/content/faq.md
```diff
@@ -85,10 +85,11 @@ docker run --rm -it \
 find /backup/mail -type f -mtime +30 -delete
 ```
 
-### What about `mail-state` folder?
+### What is the `mail-state` folder for?
 
-This folder consolidates all data generated by the server itself to persist when you upgrade.
-Example of data folder persisted: lib-amavis, lib-clamav, lib-fail2ban, lib-postfix, lib-postgrey, lib-spamassasin, lib-spamassassin, spool-postfix, ...
+When you run `docker-mailserver` with the ENV var `ONE_DIR=1` (_default since v10.2_), this folder will store the data from internal services so that you can more easily persist state.
```
polarathene, line 90: How about this?

```suggestion
When you run `docker-mailserver` with the ENV var `ONE_DIR=1` (_default since v10.2_), this folder will store the data from internal services so that you can more easily persist state to disk (via `volumes`).

This has the advantage of fail2ban blocks, ClamAV anti-virus updates and the like being kept across restarts for example.
```

casperklein (approved):

polarathene:
Quick question, are we lacking any brief description/docs on the other directories `maildata` (`/var/mail`) and `maillogs` (`/var/log/mail`)?

---

<details>
<summary>Resolved misunderstanding</summary>

**TL;DR:** Didn't pay attention to the nested flow and shell variable names properly.

---

This script only has `POSTSRSD_STATE_DIR` as `/var/mail-state`, thus only works with `ONE_DIR`? (it's only used in the conditional)
https://github.com/docker-mailserver/docker-mailserver/blob/d5a188c509f37e5f78d15bac310cd1e2293903cb/target/scripts/postsrsd-wrapper.sh#L20

There is one `mkdir -p` command for `/var/mail-state` that is run in `setup-stack.sh` if `ONE_DIR=1`:

https://github.com/docker-mailserver/docker-mailserver/blob/d5a188c509f37e5f78d15bac310cd1e2293903cb/target/scripts/startup/setup-stack.sh#L186

I'm not sure where `postsrsd` is configured to use the `/var/mail-state` location instead of `/etc` location? `postsrsd` is not handled in `misc-stack.sh`:

https://github.com/docker-mailserver/docker-mailserver/blob/d5a188c509f37e5f78d15bac310cd1e2293903cb/target/scripts/startup/misc-stack.sh#L24-L33

`SRS_SECRET` in `target/postsrsd/postsrsd` needs to be modified, identified that is renamed to `/etc/default/postsrsd` in the Dockerfile:

https://github.com/docker-mailserver/docker-mailserver/blob/d5a188c509f37e5f78d15bac310cd1e2293903cb/Dockerfile#L173

Realized that the state file uses `install` to place it in `/etc/postsrsd.secret` from `/var/mail-state/etc-postsrsd/postsrsd.secret`...:

https://github.com/docker-mailserver/docker-mailserver/blob/d5a188c509f37e5f78d15bac310cd1e2293903cb/target/scripts/postsrsd-wrapper.sh#L35

</details>

docs/content/config/environment.md
```diff
@@ -715,3 +715,5 @@ you to replace both instead of just the envelope sender.
 
 - **empty** => no default
 - password for default relay user
+
+[docs-faq-onedir]: ../../faq.md#what-about-mail-state-folder
```
polarathene, line 719:
```suggestion
[docs-faq-onedir]: ../faq.md#what-about-mail-state-folder
```

docs/content/config/environment.md
```diff
@@ -715,3 +715,5 @@ you to replace both instead of just the envelope sender.
 
 - **empty** => no default
 - password for default relay user
+
+[docs-faq-onedir]: ../faq.md#what-about-mail-state-folder
```
wernerfred, line 719:
```suggestion
[docs-faq-onedir]: ../faq/#what-is-the-mail-state-folder-for

**Review by wernerfred — CHANGES_REQUESTED** [review]
Fixed a broken link

File: docs/content/config/environment.md (hunk 3 above)
Comment (polarathene) [reply docs/content/config/environment.md:719]: IIRC, it expects to be given the `.md` extension, it'll convert to `faq/` during build. It'd be good if I ever find time to implement the lint test for `#anchor` references in the link URI :sweat_smile: 

```suggestion
[docs-faq-onedir]: ../faq.md#what-is-the-mail-state-folder-for
```

Thanks for spotting this one! :heart:

wernerfred (approved):
LGTM 📄

georglauterbach (approved):
LGTM

PR #2192, opened 2021-09-14
fix: Enable DH parameters (ffdhe4096) by default

polarathene:
# Description

### Summary

- DH params were only provided during `setup-stack.sh`.
- We now just do the copy operation in Dockerfile to always have DH params file by default.
- `doveadm` in newer versions of Dovecot (community/bullseye) seems to validate Dovecot configs and will fail if `setup-stack.sh` is not run first (eg when using `setup.sh` and no existing running container).
- Some feature disparity, but this was never officially supported / documented, thus not a breaking change IMO.

### Problem

`setup-stack.sh` currently copies the `ffdhe4096.pem` DH param file when none exists and no custom DH params file is provided by a user. This only applies when initializing a container instance via the default command however..

When you provide another command, such as the `setup.sh` management ones, and a container instance needs to be started instead of using an existing one, this will not run `setup-stack.sh` and thus default DH params are not guaranteed.

With our current Dovecot in `buster-slim` images, this has gone unnoticed as no error is raised. With the community edition of Dovecot that some users replace our shipped version with, or with the newer Dovecot in the `bullseye-slim` image we want to upgrade to, it seems that `doveadm` will validate the `10-ssl.conf` config, which involves `ssl_dh` pointing to a file that doesn't exist. `doveadm` throws an error and refuses to run as a result, even though this file is unnecessary.

### Solution

[As the cause has been identified](https://github.com/docker-mailserver/docker-mailserver/pull/2116#issuecomment-918705492), this PR resolves the issue by copying the `ffdhe4096.pem` to Dovecot and Postfix locations during a build via the Dockerfile.

The DH param handling in `setup-stack.sh` has been heavily refactored to remove no longer relevant logic, and make it easier to maintain. Commit history of this PR has carefully iterated on this process with relevant commit messages for anyone interested in following the changes made in an easier to grok manner.

Likewise, the related tests were updated and I've opted to combine them all into a single new refactored DRY variant. As `ONE_DIR` handling was removed from the relevant `setup-stack.sh` function, the test could also drop testing both states if desired. There is minimal increase in test time from the `ONE_DIR` handling, so I've left it as-is for now.

**Note:** While this resolves a failure with `bullseye-slim` image, the TLS cipher suite test continues to fail on `bullseye-slim` for other reasons AFAIK (_presently investigating_).

I do not consider this a breaking change despite some of the changes. The [original PR that added the feature](https://github.com/docker-mailserver/docker-mailserver/pull/1463) added no documentation, and I did several searches through the git repo files to double check. If we don't advertise official support for providing custom DH params, this seems fine.

AFAIK most/all users would have just been using the default supplied DH params (`ffdhe4096.pem`). Custom DH params is still supported and that has only changed in removing the variant `ONE_DIR=1` (`mail-state/lib-shared/`) directory. The `ONE_DIR=0` approach (add custom file to `/tmp/docker-mailserver/dhparams.pem`) is now the the approach regardles of `ONE_DIR` setting.

I've not yet added documentation of the feature (eg providing custom params, what are the default params). If you would like that as well with this PR let me know.

## Type of change

## Checklist:


georglauterbach:
I'd love to see a bit of documentation on this :) Other than that, LGTM 👍🏼

polarathene:
> I'd love to see a bit of documentation on this

Sure, I'll sort that out after I figure out the other bullseye failure with TLS tests.

polarathene:
Test failure can be ignored, fixed by the [v10.2 release `setup.sh` change](https://github.com/docker-mailserver/docker-mailserver/pull/2189).

---

Lint errors to address:

1. `DH_DEFAULT_CHECKSUM="$(sha512sum ${DH_DEFAULT_PARAMS} | awk '{print $1}')"` => "SC2086: Double quote to prevent globbing and word splitting."
2. `DMS_ONE_DIR=0` => "SC2030: Modification of DMS_ONE_DIR is local (to subshell caused by @bats test)."
3. `-v "${PRIVATE_CONFIG}:/tmp/docker-mailserver"` => "SC2031: PRIVATE_CONFIG was modified in a subshell. That change might be lost."

These 3 ShellCheck failures seem like they should be ignored?

- 1st one I have no clue how to fix since it's already wrapping the `$()` (subshell?) with double quotes.
- The last two are due to how bats works AFAIK and non-issue.

target/scripts/startup/setup-stack.sh
```diff
@@ -1301,92 +1301,30 @@ function _setup_postfix_relay_hosts
...
-      else
-        _notify 'inf' "Use ffdhe4096 for dhparams (dovecot)"
-        cp /etc/postfix/shared/ffdhe4096.pem /etc/dovecot/dh.pem
-      fi
-    else
-      _notify 'inf' "Use existing dovecot dhparams"
-      _notify 'warn' "Using self-generated dhparams is considered as insecure."
-      _notify 'warn' "Unless you known what you are doing, please remove /etc/dovecot/dh.pem."
-    fi
+    cp -f ${DH_CUSTOM} "${DH_DEST}"
```
georglauterbach, line 1325:
``` SUGGESTION
    cp -f "${DH_CUSTOM}" "${DH_DEST}"
```

test/mail_tls_dhparams.bats
```diff
@@ -0,0 +1,139 @@
...
+  export DMS_ONE_DIR=1 # default
+
+  local DH_DEFAULT_PARAMS
+  export DH_DEFAULT_CHECKSUM
+  export DH_CUSTOM_PARAMS
+  export DH_CUSTOM_CHECKSUM
+
+
+  DH_DEFAULT_PARAMS="$(pwd)/target/shared/ffdhe4096.pem"
+  DH_DEFAULT_CHECKSUM="$(sha512sum ${DH_DEFAULT_PARAMS} | awk '{print $1}')"
```
georglauterbach, line 43: The first linting actually makes sense issue:) Here is the change you'd need to do:

``` SUGGESTION
  DH_DEFAULT_CHECKSUM="$(sha512sum "${DH_DEFAULT_PARAMS}" | awk '{print $1}')"
```

test/mail_tls_dhparams.bats
```diff
@@ -0,0 +1,139 @@
...
+@test "testing tls: DH Parameters - Default [ONE_DIR=1]" {
+    PRIVATE_CONFIG="$(duplicate_config_for_container . mail_dhparams_default_1)"
+
+    common_container_setup
+    should_have_valid_checksum "${DH_DEFAULT_CHECKSUM}"
+}
+
+@test "testing tls: DH Parameters - Custom [ONE_DIR=0]" {
+    PRIVATE_CONFIG="$(duplicate_config_for_container . mail_dhparams_custom_0)"
+    DMS_ONE_DIR=0
```
georglauterbach, line 87: The linting checks that apply to this line tell you that this value will not be propagated to the outside of this function. But this is expected, and can be ignored with

``` SUGGESTION
    # shellcheck disable=SC2031
    DMS_ONE_DIR=0
```

You can add `# shellcheck disable=SC2031` to the other affected lines above too.

georglauterbach:
LGTM. I addressed the linting problems with comments above. Note that GitHub now has a "batch" feature with which you can commit multiple suggestions at once.

casperklein:
Very nice simplification of our DH parameter stuff! I was not able to find anything to complain 😉 Just waiting till Georgs  suggestion are applied, to approve this PR 👍

wernerfred (commented):
LGTM 👍🏻 well done

test/mail_tls_dhparams.bats
polarathene: Ah ok, I wasn't sure if that was the right way to handle it, as I thought that became `"quoted"unquoted"quoted"` which seemed like what the linter tries to avoid happening with vars?

polarathene (commented):
I'll need to apply the other lint corrections (same lint issues raised on similar lines). Then docs will follow.

georglauterbach (approved):

wernerfred (approved):
LGTM 🚀

polarathene:
> The failing action is due to missing setup.sh changes right?

Correct :)

Almost got another PR ready that is intended for 10.2 release. So that issue should be taken care of soon enough.

PR #492, opened 2017-01-24
Hardening TLS ciphers

tyranron:
To follow the principle: __Secure by default__

## Problem
Currently Dovecot does support TLS Medium Strength Cipher Suites.
```
$ nmap --script ssl-enum-ciphers -p 993 127.0.0.1

Starting Nmap 7.40 ( https://nmap.org ) at 2017-01-24 16:31 EET
Nmap scan report for localhost (127.0.0.1)
Host is up (0.0047s latency).
PORT    STATE SERVICE
993/tcp open  imaps
| ssl-enum-ciphers:
|   TLSv1.0:
|     ciphers:
|       TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA (secp384r1) - A
|       TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA (secp384r1) - A
|       TLS_DHE_RSA_WITH_AES_128_CBC_SHA (dh 1024) - A
|       TLS_DHE_RSA_WITH_AES_256_CBC_SHA (dh 1024) - A
|       TLS_ECDHE_RSA_WITH_3DES_EDE_CBC_SHA (secp384r1) - C
|       TLS_DHE_RSA_WITH_3DES_EDE_CBC_SHA (dh 1024) - D
|       TLS_RSA_WITH_AES_128_CBC_SHA (rsa 2048) - A
|       TLS_RSA_WITH_AES_256_CBC_SHA (rsa 2048) - A
|       TLS_RSA_WITH_3DES_EDE_CBC_SHA (rsa 2048) - C
|     compressors:
|       NULL
|     cipher preference: server
|     warnings:
|       64-bit block cipher 3DES vulnerable to SWEET32 attack
|       Key exchange (dh 1024) of lower strength than certificate key
|   TLSv1.1:
|     ciphers:
|       TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA (secp384r1) - A
|       TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA (secp384r1) - A
|       TLS_DHE_RSA_WITH_AES_128_CBC_SHA (dh 1024) - A
|       TLS_DHE_RSA_WITH_AES_256_CBC_SHA (dh 1024) - A
|       TLS_ECDHE_RSA_WITH_3DES_EDE_CBC_SHA (secp384r1) - C
|       TLS_DHE_RSA_WITH_3DES_EDE_CBC_SHA (dh 1024) - D
[...]

PR #352, opened 2016-10-12
Add ldap auth for postfix and dovecot

dennis95stumm:

PR #2196, opened 2021-09-15
fix: Remove `mkcert.sh` usage + `_setup_ssl` refactor.

polarathene:
# Description

This PR does some refactoring and fixes within the `_setup_ssl` method. Due to the amount of changes, it may be easier to review via individual commits (each has commit message detailing changes).This is easier through apps like GitKraken.

If not reviewing via commits, **I strongly advise disabling white-space noise from the diff.** Other noise near the end with some test files from the [housekeeping commit](https://github.com/docker-mailserver/docker-mailserver/pull/2196/commits/828d6354c79fd92852acaff7205b3eff6a23ece4)

One notable commit is the [removal of `mkcert.sh` usage from Dovecot](https://github.com/docker-mailserver/docker-mailserver/pull/2196/commits/5c1012b11478dea3eb3cca928cc51e3588cf713c). There was discussion to replace it with a smallstep binary, but presently I don't think there is a need for such, I've opted to just use the same dummy "snakeoil" certs that Postfix is configured to use.

This turned out being larger than expected. If desirable I could cherry pick out `mkcert.sh` removal. Should all be fine for v10.2.

---

`dms_panic` lacks tests. It looks like I implemented something similar to `errex` or `_shutdown`, both in `helper-functions.sh`? Is it worth keeping around or should I drop it?

---

### Change to default `SSL_TYPE` (partially affects tests)

<details>
<summary>Click to expand</summary>

I initially replaced `''` (empty / unset) case for `SSL_TYPE` to be `disabled` value instead and panic if `SSL_TYPE` had not been set.. However this is a bit of a hassle with tests atm so I've reverted that case value while keeping the modified `disabled` behaviour. No one should be using such a value in production regardless.

For tests that do require SSL/TLS, it's only because of a test with `openssl` making a `starttls` connection. For that I've just added a `snakeoil` case that doesn't panic or adjust the SSL/TLS settings. In `tests.bats`, this should only be the IMAP related test but as most tests use the same container I've just added `snakeoil` to that container for all tests.

Both `SSL_TYPE` values differ from the original `''` used for tests which would set `ssl=yes, disable_plaintext_auth=no`:

- `''` => `ssl=no, disable_plaintext_auth=no`, additionally disables for Postfix as well, previously only Dovecot was adjusted.
- `snakeoil` => `ssl=required, disable_plaintext_auth=yes` (defaults)

As no certs are provided/configured for either `SSL_TYPE`, when SSL / TLS is available, it uses the "snakeoil" certs provided from upstream. These are self-signed / insecure but fine for testing, actual tests that care about such use more appropriate certs.

</details>

---

### Changes related to tests

<details>
<summary>Click to expand</summary>

#### `tests.bats`

**There were 2 remaining `checking ssl:` tests in this file, they have been dropped.** These are from 2016, one is already covered in my `mail_ssl_manual.bats` test, while the other checks a lets-encrypt CA cert that was issued March 2016, but has expired since March 2021. This was required back in 2016 when LetsEncrypt was new and needing broader adoption of compatible CA certs in OS trust stores IIRC.

```
$ step-cli certificate inspect lets-encrypt-x3-cross-signed.pem

...

        Issuer: O=Digital Signature Trust Co.,CN=DST Root CA X3
        Validity
            Not Before: Mar 17 16:40:46 2016 UTC
            Not After : Mar 17 16:40:46 2021 UTC
        Subject: C=US,O=Let's Encrypt,CN=Let's Encrypt Authority X3

...
```

`git blame`:

- https://github.com/docker-mailserver/docker-mailserver/issues/109 ([relevant direct commit referencing issue])(https://github.com/docker-mailserver/docker-mailserver/issues/109)
- https://github.com/docker-mailserver/docker-mailserver/pull/420

**nmap tests have been dropped.** These were added about 4-5 years ago via these two PRs:

- https://github.com/docker-mailserver/docker-mailserver/pull/492
- https://github.com/docker-mailserver/docker-mailserver/pull/709

I have since made these redundant with the `testssl.sh` tests. Additionally the `--link` option is deprecated and IIRC these grades were a bit misleading when I initially used nmap in my own TLS cipher suite update PRs in the past.

**`mail_with_ldap.bats`** - Migrated `--link` _docker run_ option to use `--network` + `--network-alias`. [`--link` is deprecated](https://docs.docker.com/network/links/), future breakage is expected. No other tests use `--link`.

**`mail_privacy.bats`** - Removed irrelevant lines from. When this test was extracted from `tests.bats`, it copied over some excess comment headings. The test itself looks a bit vague too.

- [Original PR introducing the `LDAP` section header comment to `tests.bats`](https://github.com/docker-mailserver/docker-mailserver/pull/352) ([specific line referenced](https://github.com/docker-mailserver/docker-mailserver/pull/352/files#diff-c65d41b8ad46ee079a476545e7816994ee96c554eeac057e58c9f430f568b243R755)).
- [Original PR introducing the privacy test](https://github.com/docker-mailserver/docker-mailserver/pull/708) ([specific line referenced](https://github.com/docker-mailserver/docker-mailserver/pull/708/files#diff-c65d41b8ad46ee079a476545e7816994ee96c554eeac057e58c9f430f568b243R1104)).

</details>

## Type of change

## Checklist:


target/scripts/helper-functions.sh
```diff
@@ -12,6 +12,37 @@ function errex
   exit 1
 }
 
+PANIC_NO_ENV='no-env'
+PANIC_NO_FILE='no-file'
+PANIC_INVALID_VALUE='invalid-value'
+function dms_panic
```
georglauterbach, line 18: I'd like to see some documentation for this function :)

georglauterbach (approved):
LGTM

target/scripts/helper-functions.sh
```diff
@@ -12,6 +12,37 @@ function errex
...
+  local PANIC_INFO=$3
+
+  function _exit_with_message
+  {
+    _notify 'fatal' "${PANIC_SCOPE} | ${1} Exiting.."
+    exit 1
+  }
+
+  case "${PANIC_TYPE}" in
+    ( "${PANIC_NO_ENV}" )
```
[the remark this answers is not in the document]
polarathene: > why not use the value directly

It's self-explanatory in the var name that is defined right above the function?

The var name is the one that most would use if calling the panic function, so presumably the reader would find that easier to follow.

It's more like an enum/const, the values could be numbers or anything really as they're not really meant to have more meaning or direct usage without the vars.

target/scripts/helper-functions.sh
polarathene: Sure! Did you want inline documentation in this file, or did you mean our contributor docs?

georglauterbach: Inline or above the function:)

casperklein: We already have and use `_shutdown()`.

Shouldn't we enhance this function, instead of introducing a new one, that is basically doing the same?

E.g. afaik `_shutdown()` is never called without a prior `_notify()` call. So why not move the `_notify` call inside `_shutdown()` and make the notify message a parameter for `_shutdown()`? That's basically what _exit_with_message() in this PR does.

This leads to, why do we need `dms_panic()` when there is already `_shutdown()` (which could be extended if needed).

polarathene: > This leads to, why do we need `dms_panic()`

It just provides convenience messages for certain types of errors to avoid repetition (which'd likely diverge a bit).

> when there is already `_shutdown()`

Shutdown is doing a `kill 1` instead of `exit 1`, what is the difference? Panic seems more similar to `errex` (error exit?), as it's meant to stop/exit immediately, whereas shutdown usually implies more of a process with graceful shutdown orchestration.

I chose "panic" as akin to kernel panic, or panic in Rust. It additionally uses the `_notify` with "fatal" to indicate something has gone quite wrong/broke, as in show stopping not a recoverable/ignorable error.

Happy to rework it if you like or integrate with `_shutdown` or `errex`.

polarathene: > Inline or above the function:)

Done, above the function, a tad verbose though :sweat_smile:

casperklein: > Shutdown is doing a kill 1 instead of exit 1, what is the difference?

`kill 1` will shutdown the container, by sending a SIGTERM signal to PID 1 (init process).
`exit 1` in the `start-mailserver.sh` context, will just exit the script. Supervisor will [not restart](https://github.com/docker-mailserver/docker-mailserver/blob/2bf24e4c08dd652f9f4d2bdef639978ad7406d30/target/supervisor/conf.d/supervisor-app.conf#L11) that service. This leads to a container, where only the init process + supervisor (without any services) is running.

So `exit 1` is probably not what you want.

I would go with `_shutdown()` and enhance it (or "merge" with your current solution _dms_panic), to fit also your needs.

casperklein:
I finished reviewing. That seems to be a lot of work, especially the detailed description of each commit - awesome! Also congratulation to your sed wizard achievement 😄 I didn't know about `-e` too. I used `-e` alot with `grep`, but not with `sed` so far 👍 

All in all it LGTM. Just the panic/shutdown function should be merged IMO, as they both try to achieve the same: shutting down the container.

polarathene:
@casperklein thanks for taking the time to review it! Glad the commit messages were helpful :)

I've replaced `exit 1` with `_shutdown`. I think shutdown could be called without the intent of it being fatal, so I've retained the inner function for adding context. When this type of functionality is desired it's probably better handled/listed in this `dms_panic` method?

If someone finds a reason not to need `dms_panic` but want to use the the inner function `_shutdown_with_message`; it could be made more generic and available to others to call directly instead at that time.

casperklein:
> I think shutdown could be called without the intent of it being fatal

Any example? _shutdown is never called under any "normal" circumstances, just in case of hard errors. I found this three occurrences:

```
check-stack.sh-    _notify 'err' 'Setting hostname/domainname is required'
check-stack.sh:    _shutdown
--
setup-stack.sh-    _notify 'err' "Detecting the container IP address failed. Check if NETWORK_INTERFACE is correctly configured."
setup-stack.sh:    _shutdown
--
setup-stack.sh-    _notify 'err' "${POSTFIX_DAGENT} not set."
setup-stack.sh:    _shutdown
```

These could be converted from 'err' to 'fatal'.

We should move `_notify` inside `_shutdown()` and modify it to something like:

function _shutdown()
{
   _notify 'fatal' "$1"
   ....
}

target/scripts/startup/setup-stack.sh
```diff
@@ -1118,8 +1190,7 @@ function _setup_docker_permit
 
   if [[ -z ${CONTAINER_IP} ]]
   then
-    _notify 'err' "Detecting the container IP address failed. Check if NETWORK_INTERFACE is correctly configured."
-    _shutdown
+    _shutdown "Detecting the container IP address failed. Check if NETWORK_INTERFACE is correctly configured."
```
casperklein, line 1193: Do you think, converting that to `dms_panic()` would make sense?

polarathene: Possibly.

Do you want the same/similar message or is there an appropriate generic one that could benefit from a new `PANIC_TYPE`?

Could emit the "Detecting the container IP address failed." via `_notify 'err'` prior to the shutdown/panic fatal error: "Possible misconfiguration for NETWORK_INTERFACE" (new panic type, `PANIC_MISCONFIGURED`)?

casperklein: I am not sure if I understand correctly. But the information what is misconfigured should in any case be given and not just a generic "something is misconfigured" message.

polarathene: Hopefully the new commit makes my intent more clear :)

I did it in two stages:
- 1st commit is about what was discussed here.
- 2nd commit is a refactor based on better developer experience.

casperklein: Ok, I was already a bit tired yesterday. I get now what you meant.

I understand your dms_panic approach, but I still have the feeling, that this is a bit over-engineered. There are hopefully not that much situation, where we need to "shutdown". But I don't want to be the judge here. In the end, if it works, it's okay for me 👍 
The whole time, I was more thinking of a much simpler solution (yet not so professional as yours):

Just one function `_shutdown($message)` that is called in the SSL context (you reworked in that PR + ), as well as in the three other occurrences mentioned earlier.
To keep it DRY, repeating $messages can still be put in constants:
```

readonly PANIC_NO_IP="no IP ..."
readonly PANIC_NO_HOST="Setting hostname/domainname is required'
readonly PANIC_SSL_KEY="Certificate key not found"
readonly PANIC_SSL_CERT="Certificate not found"
```

Then just call it like this:

`_shutdown($PANIC_NO_HOST)`

Yes, this is not as flexible as your solution, but do we really need that flexibility for that few "shutdown/panic" events?

Again, don't get me wrong. This is not something I say we must do it this way, just an option to discuss 😉

polarathene: > `_shutdown($PANIC_NO_HOST)`

I removed the vars so anyone that wants to panic/shutdown doesn't have to copy/paste a specific one, if they're going to lookup the options to copy/paste might as well just make it a method name and/or prevent any typo mistakes?

In my editor it also seems to autocomplete with a suggestion if I type the function name, but that might be because I've opened `helper-functions.sh` and was working on the method :sweat_smile: 

You're right that it's over-engineered for what it is and our needs atm. I wanted to panic within `_setup_tls` under some common conditions and wasn't aware of `_shutdown` at the time, so I wrote a method to reduce copy/paste if I need it in future and avoid typing similar error messages for the same thing months later :sweat_smile: 

I personally prefer the current method as it's more maintainable / flexible, than adding new string vars for more specific failures (_which has less value in co-location vs inline?_).

`_shutdown` still works directly but expects a message arg (I could make it optional if desired), which is to encourage invoking it with some context of where it was called (less of an issue with us and having a review process though :stuck_out_tongue: ).

---

I'm fine with whatever is agreed on. The current approach might be a bit much, but it works and shouldn't cause any maintenance woes, probably best to not bike shed on it too much :smile:

casperklein (approved):

georglauterbach:
LGTM. We will wait for #2189 to be merged so we can rebase and see what the tests do.

target/scripts/helper-functions.sh
```diff
@@ -252,8 +303,13 @@ function _obtain_hostname_and_domainname
...
 
+# Call this method when you want to panic (emit a 'FATAL' log level error, and exit uncleanly).
+# `dms_panic` methods should be preferred if your failure type is supported.
 function _shutdown
 {
+  local FATAL_ERROR_MESSAGE=$1
+
+  _notify 'fatal' "${FATAL_ERROR_MESSAGE}"
   _notify 'err' "Shutting down.."
   kill 1
```
georglauterbach, line 314: This is fine with me, but we could use

``` suggestion
  kill -TERM 1
```

instead to make clear we're sending `SIGTERM` here. Otherwise, I'd have to always look it up (which is okay though. Just a small hint)

georglauterbach (approved):
LGTM 👍🏼

target/scripts/helper-functions.sh
casperklein: To keep it uniform with other `kill` calls in DMS, `-SIGTERM` should be used instead of `-TERM`.

On the other hand, why providing a functionless option (because it's default), which should be well known by everyone, that ever used kill? This will add 5-8 extra bytes to our image!!°1^1^1 😄

casperklein:
We should also get rid of the broken `function _defunc()`, which aims the same functionality as `_shutdown()`.

@polarathene If you would like to fix that also within this PR, I found it referenced here:

```
target/scripts/startup/check-stack.sh:    ${FUNC} || _defunc
target/scripts/startup/fixes-stack.sh:    ${FUNC} || _defunc
target/scripts/startup/setup-stack.sh:      _defunc
target/scripts/startup/daemons-stack.sh:    ${FUNC} || _defunc
target/scripts/startup/misc-stack.sh:    ${FUNC} || _defunc
target/scripts/start-mailserver.sh:function _defunc
```

Otherwise, I can provide a separate PR for that.

polarathene:
> Otherwise, I can provide a separate PR for that.

I would prefer that to be handled via a separate PR to avoid requiring two approvals again :stuck_out_tongue: 

I don't mind doing the follow up PR, but if you're happy to take care of that I appreciate it! I've got a large update to docs normalizing paths/volumes I've been working on to finish an existing PR, then wanting to help @NorseGaud tackle his long-lived PR.

target/scripts/helper-functions.sh
polarathene: I'll add this in a small follow up PR or if @casperklein takes on the other improvement replacing `_defunc` I guess it could go with that.

casperklein:
Done in #2199.

PR #2199, opened 2021-09-19
function _defunc removed

casperklein:
# Description

Follow up PR as discussed in #2196 

This removes the function `_defunc`, which probably never worked as intended.

`_defunc` calls were replaced by `_shutdown`.

Fixes #

## Type of change

## Checklist:


georglauterbach (dismissed):
LGTM 👍🏼

target/scripts/startup/fixes-stack.sh
```diff
@@ -35,21 +33,18 @@ function _fix_var_amavis_permissions
...
-  rm -f /etc/cron.d/clamav-freshclam
+  _notify 'task' 'Cleaning up disabled ClamAV'
+  rm /etc/logrotate.d/clamav-* /etc/cron.d/clamav-freshclam || _shutdown 'Failed to remove ClamAV configuration'
 }
 
 function _fix_cleanup_spamassassin
 {
   _notify 'task' 'Cleaning up disabled SpamAssassin'
-  rm -f /etc/cron.daily/spamassassin
+  rm /etc/cron.daily/spamassassin || _shutdown 'Failed to remove SpamAssassin configuration'
```
polarathene, line 49: Is there a reason for dropping `-f` on `rm` calls?

[`rm` docs](https://man7.org/linux/man-pages/man1/rm.1.html):

> `-f`, `--force`
> 
> ignore nonexistent files and arguments, never prompt

I understand that usually `docker-mailserver` isn't run with an interactive terminal, and prompting probably doesn't happen in that case, but is it desirable to be prompted when using something like `docker run -it`? We usually only do `docker run -t` if not using compose commands.

Without `-f` that also implies it'll error if the file does not exist, which in this context doesn't seem like it should be a concern to throw such an error?

target/scripts/startup/setup-stack.sh
```diff
@@ -373,8 +373,7 @@ function _setup_dovecot_local_user
   then
     if [[ ${ENABLE_LDAP} -eq 0 ]]
     then
-      _notify 'fatal' 'Unless using LDAP, you need at least 1 email account to start Dovecot.'
-      _defunc
+      _shutdown 'Unless using LDAP, you need at least 1 email account to start Dovecot.'
```
polarathene, line 376: This LDAP error handling is a good case for preferring `_shutdown` as that kind of message is less generic :+1:

target/scripts/startup/fixes-stack.sh
```diff
@@ -35,21 +33,18 @@ function _fix_var_amavis_permissions
   [[ ${ONE_DIR} -eq 0 ]] && AMAVIS_STATE_DIR="/var/lib/amavis"
   [[ ! -e ${AMAVIS_STATE_DIR} ]] && return 0
 
-  _notify 'inf' 'Checking and fixing Amavis permissions'
-  chown -hR amavis:amavis "${AMAVIS_STATE_DIR}"
-
-  return 0
```
polarathene, line 41: I'm not sure what the `return 0` was for here, I see you removed it elsewhere too, but not earlier in this function which calls `return 0` if conditional was true. Was that missed for removal, or is that particular case still valid for `return 0`?

polarathene (dismissed):
LGTM :+1: 

I see a common pattern though with the "Failed to start.." lines. In some cases the original `|| _defunc` handled in the loop could have been kept to call `_shutdown`, although I guess that's more complicated/hassle in shell scripts to reference the service index/key :/

Could still be a good case for adding a new panic type to replace the `_shutdown` calls as this is kind of what it's intended for (_repetition and re-use across scripts with common error messages_):

```sh
function dms_panic
{
# ...
( 'fail-init' ) # PANIC_INFO == <name of service or process that failed to start / initialize>
      SHUTDOWN_MESSAGE="Failed to start ${PANIC_INFO}!"
    ;;
# ...
}

function dms_panic__fail_init { dms_panic 'fail-init' "${1}" "${2}"; }
```

The other two variants "Failed to fix" / "Failed to remove" may not benefit as much atm :man_shrugging: 

I don't mind either way with `_shutdown` as is, or adopting `dms_panic` like above.

polarathene:
Did you also want to add `-SIGTERM` to the `_shutdown` kill command as discussed at the end of https://github.com/docker-mailserver/docker-mailserver/pull/2196 ?

target/scripts/startup/fixes-stack.sh
casperklein: When using `-f`, `rm` will never fail (exit 1), which makes testing for success pointless.

casperklein:
> In some cases the original `|| _defunc` handled in the loop could have been kept to call `_shutdown`

In most cases, `${FUNC} || _defunc` was pretty pointless. This checked only, if the **last** command in the called function was successful or not. For example:

```
_importantFunction() {
  doImportantThing1
  doImportantThing2
  echo "useless info"
}

_imortantFunction || echo "something went wrong"
```

You'll only see the "something went wrong" output, if the last echo command in that functions fails. Pretty pointless if you ask me.

Therefore I've added the extra _shutdown calls explicitly to operations, where I thought they are important.

Ideally, we should `set -ueo pipefail` or similar on top, so that the scripts immediately fails if something went wrong.

This PR should only obsolete `_defunc`. I agree, with a further PR the whole thing could be reworked and improved.

target/scripts/startup/fixes-stack.sh
casperklein: See my other comment. Return 0 was pretty pointless. In that case for example, `chown` could fail, but return 0 ensures, that the function never the less "reports back", everything is ok.

casperklein: The mistake in thinking here is, that the function stops on the first error. That is not the case. Even when a command fails, all following commands are executed.

casperklein:
> Did you also want to add `-SIGTERM` to the `_shutdown` kill command as discussed at the end of #2196 ?

Nope. If someone want to add it - okay. But I don't see a reason for it (see my comment there). We shouldn't start, providing default settings on the command line.

target/scripts/startup/fixes-stack.sh
casperklein: > it'll error if the file does not exist

This should never be the case when using the images we provide?

georglauterbach (approved):

polarathene:
> Ideally, we should `set -ueo pipefail` or similar on top, so that the scripts immediately fails if something went wrong.

This sounds good (_the functionality, my shell-fu doesn't recall that incantation that well_), I suppose it might make sense to do that as early as possible during init scripts?

polarathene (approved):

target/scripts/startup/fixes-stack.sh
casperklein: `rm <file>` never prompts for confirmation. However, many Linux distributions ship with some default aliases like:

```sh
alias rm='rm -i' # prompt before every removal
alias cp='cp -i' # prompt before overwrite
```

casperklein:
`set -ueo pipefail` That is btw the unofficial BASH strict mode 😉 

> I suppose it might make sense to do that as early as possible during init scripts?

Yes. But keep in mind, it's not just putting the line on top of the script. The script itself may need some adjustments.

Every command that is expected to fail for whatever reasons, must be catched or the script will fail.

```sh
this-command-may-fail-or-not-we-dont-care || true
```

This allows the rest of the script to continue. Otherwise, in "strict mode" the script will exit on the first failed command.

georglauterbach:
At the moment, I'd advise against using `set -euEo pipefail` in the start-stack. This needs seperate PR:)

target/scripts/startup/fixes-stack.sh
georglauterbach: To give my two cents here as well, especially when using `set -euEo pipefail` (mentioned in another comment today or yesterday), you may need a `return 0` because Bash will return from a function with the return value of the last command of not explicit `return 0` is provided. Therefore, when running Bash in "strict mode", you may actually _need_ a `return 0` in some places.

casperklein:
The test [failed](https://github.com/docker-mailserver/docker-mailserver/pull/2199/checks?check_run_id=3686946671#step:8:300) 😞 

```
 not ok 294 checking system: postfix should not log to syslog
# (from function `assert_failure' in file test/test_helper/bats-assert/src/assert.bash, line 140,
#  in test file test/tests.bats, line 538)
#   `assert_failure' failed
# 
# output : Sep 23 12:24:31 mail postfix/postfix-script[2647]: fatal: the Postfix mail system is already running
```

casperklein:
The failed test does not seem to be related to this PR, as it was successful in the [past](https://github.com/docker-mailserver/docker-mailserver/actions/workflows/test_merge_requests.yml?query=branch%3Aremove-defunc++).

georglauterbach:
I was seeing multiple issues with this test too, unrelated to any of my changes. Disable it too?

PR #2179, opened 2021-09-10
feat: rootless Podman with documentation

OrvilleQ:
# Description

Setup.sh : Add ability to operate rootless podman.

Doc: Add guide for Podman and Root-less Podman

Fixes #2178

## Type of change

- [x] Improvement (non-breaking change that does improve existing functionality)
- [x] This change requires a documentation update

## Checklist:

- [x]  My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [x] If necessary I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #2201, opened 2021-09-19
make setup.sh completely non-interactive

georglauterbach:
# Description

Cleans up `setup.sh` and makes it possible to use the script in a completely non-interactive way. Added `-R` flag. Moved ENV variables to the top, removed unnecessary functions and renamed functions.

This PR _should_ be the last one added to the `10.2.0` milestone. This PR is not a breaking change. Fixes a regression for SELinux (used `OPTARG` where `OPT` was actually correct).

Follow up on #2179

## Type of change

## Checklist:

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -54,8 +54,8 @@ function _show_local_usage
             Provides the name of the running container.
 
         -p PATH
-            Provides the config folder path to the temporary container
-            Does not work if docker-mailserver container already exists.
+            Provides the local path of the config folder to the temporary container instance.
+            Does not work if an existing docker-mailserver container is already running.
 
     ${LBLUE}SELinux${RESET}
         -z
--- a/setup.sh
+++ b/setup.sh
@@ -4,19 +4,20 @@
 # executed  manually / via Make
 # task      wrapper for various setup scripts
 
-SCRIPT='setup.sh'
 DIR="$(pwd)"
-CRI=
 CONFIG_PATH=
 CONTAINER_NAME=
+CRI=
 DEFAULT_CONFIG_PATH=
+DESIRED_CONFIG_PATH=
+DMS_CONFIG='/tmp/docker-mailserver'
 IMAGE_NAME=
 INFO=
-USE_TTY=
-USE_SELINUX=
-VOLUME=
-WISHED_CONFIG_PATH=
 PODMAN_ROOTLESS=false
+SCRIPT='setup.sh'
+USE_SELINUX=
+USE_TTY=
+VOLUME=
 
 set -euEo pipefail
 trap '__err "${FUNCNAME[0]:-?}" "${BASH_COMMAND:-?}" "${LINENO:-?}" "${?:-?}"' ERR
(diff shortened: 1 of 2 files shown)


setup.sh
```diff
@@ -41,7 +54,8 @@ function _show_local_usage
             Provides the name of the running container.
 
         -p PATH
-            Provides the config folder path to the temporary container (does not work if docker-mailserver container already exists).
+            Provides the config folder path to the temporary container
+            Does not work if docker-mailserver container already exists.
```
polarathene, line 58:
```suggestion
            Provides the local path of the config folder to the temporary container instance.
            Does not work if an existing docker-mailserver container is already running.
```

setup.sh
```diff
@@ -77,58 +101,40 @@ function _get_absolute_script_directory
...
+function _handle_config_path
 {
-  if [[ -n ${CONTAINER_NAME} ]]
+  if [[ -z ${WISHED_CONFIG_PATH} ]]
   then
-    VOLUME=$(${CRI} inspect "${CONTAINER_NAME}" \
-      --format="{{range .Mounts}}{{ println .Source .Destination}}{{end}}" | \
-      grep "/tmp/docker-mailserver$" 2>/dev/null || :)
-  fi
+    # no wished config path
```
polarathene, line 112: Desired seems more appropriate? Change throughout file:

`WISHED`/`wished` => `DESIRED`/`desired`

setup.sh
```diff
@@ -77,58 +101,40 @@ function _get_absolute_script_directory
...
+    if [[ -n ${CONTAINER_NAME} ]]
+    then
+      VOLUME=$(${CRI} inspect "${CONTAINER_NAME}" \
+        --format="{{range .Mounts}}{{ println .Source .Destination}}{{end}}" | \
+        grep "/tmp/docker-mailserver$" 2>/dev/null || :)
```
polarathene, line 117: Might be a bit OCD of me, but do you see any value in lifting `/tmp/docker-mailserver` up into a var at the top? (`DMS_CONFIG`?)

polarathene (changes requested):
Nice improvements! :heart:

setup.sh
georglauterbach: I will have a look at all occurrences and decide :)

polarathene: > at all occurrences

Do not make a change to many files for that string. It'll cause big conflicts for a PR I need to push that modifies lines often with that value in it.

georglauterbach: I'll only do it for `setup.sh`.

georglauterbach: Done. [c006a977c55ff3e5fe90a540cd153372165e9363]

setup.sh
georglauterbach: Done. [c006a977c55ff3e5fe90a540cd153372165e9363]

georglauterbach:
@polarathene I removed one line from the `.ecrc` file for the EditorConfig lint, but the line with `"^test/"` will still block the whole test directory. I will take care of that with another PR.

polarathene (approved):

setup.sh
```diff
@@ -52,7 +67,16 @@ function _show_local_usage
             Allows container access to the bind mount content that is private and
             unshared with other containers on a SELinux-enabled host.
 
-${ORANGE}EXIT STATUS${RESET}
+    ${LBLUE}Podman${RESET}
+        -R
```
casperklein, line 71: No show-stopper. But I was wondering, why you chose an upper-case 'R'. With exception of `-Z`, all other parameters are lower-case.

casperklein (approved):

setup.sh
georglauterbach: The full story is this: I war originally trying to use `--rootless`, but since `getopts` does not care about double-dashed flags, and I did not want to increase the complexity of the script, I figured I should go with a single-letter-option. When I used `r` -- and I did in the beginning -- I always found myself reminded of `--recursive` and I knew some scripts where an uppercase letter denoted that there is a long version available, so I chose `-R` :)

PR #2207, opened 2021-09-23
docs(deps): bump mkdocs-material to 7.3.0

wernerfred:
# Description

bump mkdocs-material to 7.3.0

it got released today and we now (theoretically) have the option to:
- Added support for sticky navigation tabs
- Added support for section index pages
- Added support for removing generator notice

## Type of change

## Checklist:

### DIFF
--- a/.github/workflows/scripts/docs/build-docs.sh
+++ b/.github/workflows/scripts/docs/build-docs.sh
@@ -6,7 +6,7 @@ docker run \
   --rm \
   --user "$(id -u):$(id -g)" \
   -v "${PWD}:/docs" \
-  squidfunk/mkdocs-material:7.2.8 build --strict
+  squidfunk/mkdocs-material:7.3.0 build --strict
 
 # Remove unnecessary build artifacts: https://github.com/squidfunk/mkdocs-material/issues/2519
 cd site || exit


georglauterbach:
> 1. Added support for sticky navigation tabs
> 2. Added support for section index pages
> 3. Added support for removing generator notice

I would not like to enable 1 and 2 and I think we should keep the notice about MkDocs with 3 to help the MkDocs project.

georglauterbach (approved):

casperklein (approved):

polarathene:
> I would not like to enable 1 and 2

Sometimes sticky navigation tabs would be useful, although I'd prefer them revealing on scroll up, or better a reveal button toggle. We have the "Back to top" button that reveals on scroll, bit clunky but I guess it's better than permanently visible tabs.

Agree with avoiding section index pages, that's confusing UX. We use the header tabs for this purpose.

> I think we should keep the notice about MkDocs with 3 to help the MkDocs project.

That's fine. It might be worth disabling though, that way we can move the MkDocs mention/link into the same line as our injected footer line. Slims down the footer a bit.

georglauterbach:
> > I think we should keep the notice about MkDocs with 3 to help the MkDocs project.
>
> That's fine. It might be worth disabling though, that way we can move the MkDocs mention/link into the same line as our injected footer line. Slims down the footer a bit.

That would be fine with me :)

PR #2206, opened 2021-09-22
docs(chore): Normalize for consistency

polarathene:
# Description

Follow up to @williamdes PR: https://github.com/docker-mailserver/docker-mailserver/pull/2159

---

This PR only applies to the `docs/content/` directory (_with a few minor exceptions_). `target/` and `test/` can be normalized at a later date.

To try make review easier on maintainers, I have tried to scope specific normalizations across files to individual commits per normalization applied. It's not perfectly organized but may be preferable than the noise of the full PR diff, since it should be **quicker to scan through the changes per commit intent**.

I have also covered the most relevant changes in each commits message as an overview. The larger normalization commits have had their messages included below in a comment for convenience/reference.

I have tried to be thorough, cautious and review the changeset multiple times, but **I may have made some errors accidentally** (_this PR took a fair amount of time to prepare, I have probably been looking at it too long!_). Some docs where I especially lack configuration knowledge, such as **LDAP or K8S may need closer review of changes**, see the "Commit Overview" section below for more details.

**Follow-up PRs (_queued up, dependent upon this PR merging_):**

- **`ssl.md`:** letsencrypt section only has partial modifications here. That section has been revised as it seems rather outdated with current upstream projects it references.
- **`basic-installation.md`:** This is an old community contributed guide from the wiki prior to docs migration. Also heavily revised enough to warrant a separate PR, instead of adding unnecessary noise to the focus of this PR.

**TODO:**

- Possibly merge in `mail_crypt.md` changes from the [active encryption PR](https://github.com/docker-mailserver/docker-mailserver/pull/2080). File avoided for now to reduce chance of merge conflicts affecting it.

## Type of change

- [x] Improvement (non-breaking change that does improve existing functionality)
- [x] This change requires a documentation update

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [x] New and existing unit tests pass locally with my changes

PR #2209, opened 2021-09-24
docs: `ssl.md` - Revise letsencrypt section

polarathene:
# Description

This is a follow-up to https://github.com/docker-mailserver/docker-mailserver/pull/2206

The _Let's Encrypt_ section has been heavily revised. It includes the missing normalization for this section from the linked PR, and brings the section content up to date.

It is an initial pass and can probably be refined further. I've not yet fully verified the instructions work, only the `nginx-proxy` + `acme-companion` docker-compose example with staging certs works (_however `docker-mailserver` isn't compatible with this as FQDN directory name is prefixed by `_test_`_).

[Link to the TLS docs preview page for this PR](https://pullrequest-2209--dms-doc-previews.netlify.app/config/security/ssl/).

## Type of change

## Checklist:

- [x]  I have performed a self-review of my own code

### DIFF
--- a/docs/content/config/security/ssl.md
+++ b/docs/content/config/security/ssl.md
@@ -6,9 +6,9 @@ There are multiple options to enable SSL (via [`SSL_TYPE`][docs-env::ssl-type]):
 
 - Using [letsencrypt](#lets-encrypt-recommended) (recommended)
 - Using [Caddy](#caddy)
-- Using [Traefik](#traefik)
-- Using [self-signed certificates](#self-signed-certificates-testing-only)
-- Using [your own certificates](#custom-certificate-files)
+- Using [Traefik](#traefik-v2)
+- Using [self-signed certificates](#self-signed-certificates)
+- Using [your own certificates](#bring-your-own-certificates)
 
 After installation, you can test your setup with:
 
--- a/docs/content/config/security/ssl.md
+++ b/docs/content/config/security/ssl.md
@@ -25,22 +25,35 @@ After installation, you can test your setup with:
 
 ## Let's Encrypt (Recommended)
 
-To enable Let's Encrypt for `docker-mailserver`, you have to:
+To enable _Let's Encrypt_ for `docker-mailserver`, you have to:
 
-- Get your certificate using [letsencrypt client](https://github.com/letsencrypt/letsencrypt)
-- Add an environment variable `SSL_TYPE` with value `letsencrypt` (see [`docker-compose.yml`][github-file-compose])
-- Mount your whole `letsencrypt` folder to `/etc/letsencrypt`
-- The certs folder name located in `letsencrypt/live/` must be the `fqdn` of your container responding to the `hostname` command. The `fqdn` (full qualified domain name) inside the docker container is built combining the `hostname` and `domainname` values of the `docker-compose` file, eg:
+1. Get your certificate using the _Let's Encrypt_ client [Certbot][certbot::github].
+2. For your `docker-mailserver` container:
+
+    1. Add the environment variable `SSL_TYPE=letsencrypt`.
+    2. Mount [your local `letsencrypt` folder][certbot::certs-storage] as a volume to `/etc/letsencrypt`.
+
+You don't have to do anything else. Enjoy!
+
+!!! note
+
+    `/etc/letsencrypt/live` stores provisioned certificates in individual folders named by their FQDN (_Fully Qualified Domain Name_). `docker-mailserver` looks for it's certificate folder via the `hostname` command. The FQDN inside the docker container is derived from the `--hostname` and `--domainname` options.
+
+!!! example
+
+    Add these additions to the `mailserver` service in your [`docker-compose.yml`][github-file-compose]:
 
     ```yaml
     services:
       mailserver:
+        # For the FQDN 'mail.example.com':
         hostname: mail
         domainname: example.com
-        fqdn: mail.example.com
+        environment:
+          - SSL_TYPE=letsencrypt
+        volumes:
+          - /etc/letsencrypt:/etc/letsencrypt
     ```
-
-You don't have anything else to do. Enjoy.
 
 ### Example using Docker for Let's Encrypt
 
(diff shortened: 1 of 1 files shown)


polarathene:
**NOTE:** On a Ubuntu 21.04 VPS instance, docker-compose is version 1.25 (from 2020), this fails to run `docker-compose.yml` examples versioned `3.8`, requiring the version be dropped down to `3.7`.

## Commit Overview:

- Added a warning to make users aware that using a public CA like _Let's Encrypt_ will publicly log information that may be somewhat sensitive, or undesirable to have historic records made public which cannot be redacted.
- The `letsencrypt` repo that was linked early in this guide now redirects to the [Certbot repo](https://github.com/certbot/certbot).
- More explicit volume mount instruction for CertBot; the local location was a tad vague.
- Better clarified `/etc/letsencrypt/live` contents structure, as well as FQDN info. Removed the misleading `fqdn:` from `docker-compose.yml` example snippet.
- General rewrite of the Certbot section with additional tips, and heavy rewrite of `nginx-proxy` sections.
- Links extracted to the bottom of the file, and many new supporting links added.

---

- `letsencrypt-nginx-proxy-companion` has [changed project name to `acme-companion`](https://github.com/nginx-proxy/acme-companion/pull/777), and [transferred to new maintainers and the `nginx-proxy` organization](https://github.com/nginx-proxy/nginx-proxy/issues/1304#issuecomment-811851949). This also affects the DockerHub image references.
- `acme-companion` has [switched from using `simp_le` to `acme.sh`](https://github.com/nginx-proxy/acme-companion/pull/719) for provisioning certificates. This requires mounting an additional volume for persisting provisioner state.
- The dummy container (_webmail_) is no longer `library/nginx`, just [`nginx`](https://hub.docker.com/_/nginx). This container also doesn't appear to be required. I've verified that the ENV can be given to the `mailserver` service container directly.
- This setup implies `SSL_TYPE=letsencrypt` (_due to the `/etc/letsencrypt/live` mount_), ~~but was never compatible with this mode (_[git blame from `target/start-mailserver.sh` letsencrypt handling in 2016](https://github.com/docker-mailserver/docker-mailserver/blame/f0defda2a19f69eed44297df0836f331c4a306c3/target/start-mailserver.sh#L1044) to [current letsencrypt handling](https://github.com/docker-mailserver/docker-mailserver/blob/c7e9dd2badd2daebbc0dd8f676d25312059a976f/target/scripts/startup/setup-stack.sh#L914) - both expect the FQDN as the directory name, not [as the filename](https://github.com/nginx-proxy/acme-companion/blame/main/app/letsencrypt_service#L39)_)..~~ **EDIT:** Nevermind, it appears to provision to a directory with the correct filenames (_although `acme.sh` docs warn against using these internal contents directly as it is subject to change_), which are symlinked to from `certs/`. It's only with the _Let's Encrypt_ staging endpoint that it would seem incompatible as that prefixes the the directory names with `_test_`. I'll need to test this a bit more.

---

Example - `nginx-proxy` + `acme-companion` with `docker-compose.yml`:
- The bulk of the `mailserver` service has been removed, users are advised to have an existing `docker-compose.yml` config for `docker-mailserver` and update only what is relevant to integrate with the cert provisioner.
- [`DEBUG` is _false_ by default](https://github.com/nginx-proxy/acme-companion/blob/main/docs/Container-configuration.md).
- The `htpasswd` volume is unnecessary, only relevant if using _"Basic Authentication"_ to protect access to web service endpoints. `conf.d/` is also not required by default, it can be useful for the `standalone` mode (_documented as a `tip`_). Remaining volumes have inline-comments to document their purpose.
- The `networks:` portion of the example appears to be taken from upstream, [which that has since dropped it](https://github.com/nginx-proxy/nginx-proxy/commit/478ad17adb59808fecdaf1b7076d938460ba0108). While we could continue to document this, I consider it more of an advanced config detail that we don't need to touch on in our docs.
- `volumes_from:` is [not supported in v3 Compose format](https://docs.docker.com/compose/compose-file/compose-file-v3/#volumes), [only v2](https://docs.docker.com/compose/compose-file/compose-file-v2/#volumes_from) and the Docker CLI. I did not want to advise v2, so I've duplicated the volumes between the two containers instead. Internally `acme-companion` would rely on `volumes_from:` to identify the `nginx-proxy` container, it [provides alternative discovery methods](https://github.com/nginx-proxy/acme-companion/blob/main/docs/Getting-containers-IDs.md), the label is outdated and refers the legacy label (_their script logic is the same_); using the ENV `NGINX_PROXY_CONTAINER` seemed most appropriate and has been added.
- Upstream `acme-companion` docs only cover support for v2 Compose format. [There is a note regarding `nginx-proxy`](https://github.com/nginx-proxy/acme-companion/blob/main/docs/Docker-Compose.md) having [volumes configured in it's Dockerfile](https://github.com/nginx-proxy/nginx-proxy/blob/b1334033317566fd58858c1d43e60d810ed6502b/Dockerfile#L72). Providing a volume for `/etc/nginx/dhparam` is required to avoid creating anonymous volumes each run of `nginx-proxy`. I've used a named data volume here to make it stick out more, it's not desirable and upstream should fix this, then we can drop it.
- I considered [disabling DH param generation](https://github.com/nginx-proxy/nginx-proxy#ssl-support-using-an-acme-ca), as upstream is pointlessly generating DH params regardless of needing them, they have not yet adopted standardized DHE groups. `acme-companion` differs slightly in generation (a bit more secure by not using `-DSAPARAM`), but lacks an ENV to disable generation. Neither should be a concern for us, but I might open a PR to both projects to address it.
- I've also opted to only demonstrate the [Two Container (Basic) setup](https://github.com/nginx-proxy/acme-companion/blob/main/docs/Basic-usage.md) that upstream documents. Previously our docs have been showing [`docker-gen` with the Three Container (Advanced) setup](https://github.com/nginx-proxy/acme-companion/blob/main/docs/Advanced-usage.md), which allows for not having the Docker API socket attached as a volume to a container exposed to the web. This reduces the security a bit, and I have not mentioned that on our docs. I could caution the reader with a link to upstream about the risk, but I don't think we should maintain the `docker-gen` setup.
- `acme-companion` does support restarting a container [when certificates are provisioned/renewed](https://github.com/nginx-proxy/acme-companion/blob/main/docs/Let's-Encrypt-and-ACME.md#container-restart-on-cert-renewal), but I assume that's not something we want to encourage vs having `supervisorctl` internally restart services? I've mentioned it along with other config ENV in a `tip` admonition.

georglauterbach (dismissed):

casperklein:
These links on the top do not work:

- Using Traefik
- Using self-signed certificates
- Using your own certificates

polarathene:
> These links on the top do not work

Good spotting! Thanks for that. I really need to get that link checking lint added to the CI :sweat_smile:

casperklein (approved):

georglauterbach (approved):

georglauterbach (approved):

georglauterbach:
Nice PR! Ready to be merged I think :)

PR #2210, opened 2021-09-24
docs: Revise `basic-install.md`

polarathene:
# Description

This is a follow-up to #2206

It is an initial pass and can probably be refined further. The main intention was to re-structure the content and enumerate steps. Besides re-arranging content order, some of it was tweaked or dropped as appropriate.

Due to the noisy PR diff not aligning well, I've split the PR into several commits to resolve the diff boundaries being misaligned. Unlike my other PRs, these have no message content, no changes worth highlighting :)

[Direct preview link - Basic Installation](https://pullrequest-2210--dms-doc-previews.netlify.app/examples/tutorials/basic-installation/).

- Should the default ENVs in the `docker-compose.yml` ENV be removed?
- `ds` and `wsproxy` software referenced has been removed. So has the Moodle software specific content at the end.

<details>
<summary>History of this community guide</summary>

The [original content was from the old wiki](https://github.com/docker-mailserver/docker-mailserver/commit/fb17b3e81f2900997f2cd7a9951ef3cc5661d691), and that was imported from a third-party repo that the author has since [rewritten their guide](https://github.com/docker-scripts-archived/moodle/commit/9abb2f5837ca2f0a45b3e63e1d4008e033fd25ed#diff-e38fb25a70662eb7953330dec7ebb8e015b180f2d481b2caf0f1e5c4d1a6718e) to not reference `docker-mailserver` (_they don't seem to use our image anymore for some reason_). That repo is archived, [they migrated to Gitlab](https://gitlab.com/docker-scripts/moodle/-/blob/master/docs/email-setup.md).

This wiki content is where the `ds` (_script utility the contributor developed_) and `wsproxy` (_an `apache` reverse proxy_) were introduced into our docs. The guide appears originally tailored towards supporting [`moodle` LMS software](https://moodle.org/). The author only contributed a link to their README docs, then someone copied the content directly over into the wiki.

Likewise, the same author contributed the [`forward-only-mailserver-with-ldap-auth.md` guide in the wiki too](https://github.com/docker-mailserver/docker-mailserver/commit/cfd78782611bbcb1dbd151b914a8035d433f50e2).

</details>

## Type of change

## Checklist:

- [x]  I have performed a self-review of my own code

### DIFF
--- a/docs/content/examples/tutorials/basic-installation.md
+++ b/docs/content/examples/tutorials/basic-installation.md
@@ -18,7 +18,7 @@ In this setup `docker-mailserver` is not intended to receive email externally, s
 
 2. Pull the docker image: `docker pull docker.io/mailserver/docker-mailserver:latest`.
 
-- Create the file `docker-compose.yml` with a content like this:
+3. Create the file `docker-compose.yml` with a content like this:
 
     !!! example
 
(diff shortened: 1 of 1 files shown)


docs/content/examples/tutorials/basic-installation.md
```diff
@@ -2,25 +2,23 @@
...
-    curl -o setup.sh \
-        https://raw.githubusercontent.com/docker-mailserver/docker-mailserver/master/setup.sh
-    chmod a+x ./setup.sh
-    ```
+    Adding the docker network's gateway to the list of trusted hosts (_eg: using the `network` or `connected-networks` option_), can create an [**open relay**](https://en.wikipedia.org/wiki/Open_mail_relay). For instance [if IPv6 is enabled on the host machine, but not in Docker][github-issue-1405-comment].
+
+1. If you're running a version of `docker-mailserver` earlier than v10.2, [you'll need to get `setup.sh`][docs-setup-script]. Otherwise you can substitute `./setup.sh <command>` with `docker exec mailserver setup <command>`.
 
-- Create the file `docker-compose.yml` with a content like this:
+2. Pull the docker image: `docker pull docker.io/mailserver/docker-mailserver:latest`.
```
casperklein, line 19: We should advice to use versioned tags instead of :latest.
Or at least mention, that using :latest can lead to unwanted updates.

polarathene: > Or at least mention, that using :latest can lead to unwanted updates.

It's usually the opposite?

When relying on the `:latest` tag, you may not get the actual latest version of that tag if a local one is found. This is different with `pull` command which will fetch the latest, but in this case for the guide, I think it would be assumed that's intended? (as in starting fresh)

We normalized on `:latest` because some docs content was not being maintained as actively, the 10.1 version of docs reference a tag of v7.x IIRC for `docker-mailserver` on one page. The warnings for relying on `:latest` in production is fairly widespread in the Docker communities last I knew?

I don't mind making a change on this if more than one maintainer would like to, however I don't think `:latest` can be considered as a immutable tag. If a user expects to avoid unwanted updates, they should use a tag to pin the release. `:latest` is default right? Would you prefer this tag was just omitted and used implicitly?

polarathene:
Ugh wrong PR, rewriting history with force push, one moment..

docs/content/examples/tutorials/basic-installation.md
casperklein: You are absolutely right, in all what you said. Users should know about :latest tag and what this means in terms of updates.
It's good the docs were normalized. But just in that case, I thought, people tend to be lazy. And as this is a basic example, I bet, many will just copy/paste without knowing exactly what each line means. Especially when they are new to Docker.
Also me 😄: People shouldn't start running this mail-server, when they don't have basic knowledge about Docker/SMTP.

> This is different with pull command

It's not only the pull command. I am personally against auto-updates, where images are independently updated and containers restarted. However we give a manual [here](https://pullrequest-2210--dms-doc-previews.netlify.app/config/advanced/maintenance/update-and-cleanup/) to do so.
So using the basic example + auto updates, can lead to unexpected automated failures in the future.

Conclusion: If others also approve this, I am also fine to leave it as it is 👍

casperklein (approved):

georglauterbach (approved):

docs/content/examples/tutorials/basic-installation.md
polarathene: Have you noticed it as a recurring problem that occurs in issues raised for the project? If so, that'd be a good reason to ensure the user is better informed.

On the TLS page I have been expanding on details a little more as I think most readers are lacking knowledge and awareness in that area.

To be fair, we have been having issues raised recently including duplicates from users that aren't bothering to properly read the README, and then deciding `:edge` tag resolves their problem without thinking that through..

PR #2224, opened 2021-09-28
ci: Fix lint check status update

polarathene:
# Description

The lint workflow is not important for this PR, but a fixed requirement to pass for merging.

As this workflow is triggered by `schedule` or `workflow_dispatch`, it will not trigger other events such as `pull_request` for other workflows to respond to. The [earlier attempt was partially successful](https://github.com/docker-mailserver/docker-mailserver/pull/2220), the lint workflow did run, but it incorrectly associated the lint workflow run to the latest commit on `master` branch. 

Since the linting workflow is not important for this type of PR, we can pretend it was "skipped" and set the check status to "success". This is simpler than running the actual Lint workflow redundantly, which would require `workflow_dispatch` event triggering AFAIK to do correctly, but that has additional auth constraints via token requirements.

Resolves: https://github.com/docker-mailserver/docker-mailserver/pull/2218

## Type of change

## Checklist:

### DIFF
--- a/.github/workflows/contributors.yml
+++ b/.github/workflows/contributors.yml
@@ -43,3 +43,22 @@ jobs:
           PATH: '/CONTRIBUTORS.md'
           COMMIT_MESSAGE: 'docs(CONTRIBUTORS): update contributors'
           AVATAR_SHAPE: 'round'
+
+      # This workflow will not trigger a `pull_request` event without a PAT.
+      # The lint workflow is not important for this type of PR, skip it and pretend it was successful:
+      - name: 'Get the latest commit hash from the contributors-update branch'
+        id: commit-data
+        run: 'echo "::set-output name=head_sha::$(git rev-parse contributors-update)"'
+
+      - name: 'Commit Status: Set Lint status to success (skipped)'
+        uses: myrotvorets/set-commit-status-action@1.1.4
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          # Skipped workflows are still assigned a "success" status:
+          status: success
+          # This should be the correct commit SHA on the contributors-update branch:
+          sha: ${{ steps.commit-data.outputs.head_sha }}
+          # Name of status check to add/update:
+          context: 'lint'
+          # Optional message/note we can inline to the right of the context name in the UI:
+          description: "Lint skipped. Not relevant."
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -6,11 +6,6 @@ on:
   push:
     branches:
       - master
-  # These workflows when done will trigger this workflow too:
-  workflow_run:
-    workflows: ['Update contributors']
-    types:
-      - completed
 
 jobs:
   lint:


georglauterbach (approved):

wernerfred (commented):
Woa there happened a lot in the like 8h i was not online 😄 
I'm not sure if i would approve that. While it works and it is fine from a linting perspective in my eyes all that is a lot of workaround only because this one contributor workflow triggers a PR that cannot be linted. 
Maybe we will find another solution/action das can create PRs in a different way? It might be okay in the meantime but we should look into different solutions imho for the near future. I don't like workflows that went from workaround to workaround. I mean, the contributors workflow has half of its LOC by deleting branches and pretending success for linting job.

What do you think?

polarathene:
If you view the discussion for the resolve link, I have detailed alternative solutions to trigger the lint workflow correctly, but I can't say that they're any better. If adopting the `create-pull-request` referenced action, it can create branches IIRC and delete before/after among other features. Triggering another workflow from it making a pull request is possible, but it requires one of:

- SSH deploy key
- Github App generated token
- PAT

I don't see an issue with the approach proposed here. It's the simplest "fix", and it makes it clear that the check was skipped (_remember that these check status will show "success" if a workflow skips internally IIRC?_) via explicit message beside the check.

If someone wants to resolve it a different way that's fine :+1: I just lack the time to put towards alternatives approaches. This should at least work in the meantime :sweat_smile:

wernerfred:
Not against it - don't get me wrong. Just my incentive to get it working like it should. I will see if i can reproduce this on my repo and if i will find a solution working without PAT,..

casperklein:
1. If the goal is to skip the lint-test, why not just exclude `CONTRIBUTORS.md` in `linting.yml`?
2. What are your concerns with using a PAT? It seems to be the simplest solution and security wise as secure as any other SECRET we use.

wernerfred:
> * If the goal is to skip the lint-test, why not just exclude `CONTRIBUTORS.md` in `linting.yml`?

Because linting.yml is set in the repository settings as a mandatory check to run on every interaction. We could discuss on removing this but imho its good practice to force as many workflows as possible to run to cover all possible circumstances.

casperklein:
I understand. What is the downside of "scoped to an individual person", beside it's ugly not having a org PAT?

casperklein:
> As mentioned earlier i will try to rebuild the environment/setting in my scope and see if i find a solution that "works" but it will take some time.

So in the meantime, we merge this one or should we wait?

wernerfred:
Merge it in the meantime. Lets See If it is working. 
If not we need to fix it anyways otherwise i have time for improvements ;)

wernerfred (approved):

polarathene:
PAT will also make the action/workflow appear to be performed by you. Eg, we won't have a bot opening the PR, it'll be your account. With the SSH deploy keys you can make this work with the bot IIRC, and it allows better restricting the scope of permission to single repo. I can't recall how the Github App generated tokens compare, but if you follow that link I provided it covers all pro/cons well.

If you'd like an example of using SSH Key deployment, I came across [this AWS project PR](https://github.com/async-aws/aws/pull/271/files) which implements it, could be good reference? I also advise looking into that `create-pull-request` action as it probably simplifies the workflow and can take care of the PR/branch management while the contributors action just does what it's good at.

casperklein:
This doesn't seem to work, see [here](https://github.com/barthheiko/docker-mailserver/runs/3737606830?check_suite_focus=true#step:7:11)

`git rev-parse contributors-update` does not return a "40 character SHA1".

polarathene:
> does not return "a 40 character SHA1".

It should...

On my system:

```console-session
$ git rev-parse main
b1334033317566fd58858c1d43e60d810ed6502b

$ git log -n 1 main --pretty=format:"%H"
b1334033317566fd58858c1d43e60d810ed6502b
```

In the action the step gets the hash and sets it to an output variable correctly AFAIK:

```yml
id: commit-data
run: 'echo "::set-output name=head_sha::$(git rev-parse contributors-update)"'
```

Although the workflow report you linked shows this error output for that command:

```
fatal: ambiguous argument 'contributors-update': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

Is this branch not locally available at the time the command is run? I thought you had checked out the repo and created the branch with a commit pushed in the earlier steps?

I unfortunately don't have time to setup a throwaway repo to iterate on this workflow so that the required info is available. Running either of the two commands I showed above, but giving an invalid branch name results in the same error output, so I'm rather positive that's the cause of the failure.

I can create another PR to add a step that I think would resolve it:

https://github.com/docker-mailserver/docker-mailserver/blob/7b4ce69c3f55c316e273acf14b973d0645983351/.github/workflows/docs-production-deploy.yml#L80-L84

That does a shallow checkout of the `ref` branch, and if `path` is omitted should checkout to the active location instead of a sub-directory.

polarathene:
The current workflow earlier steps:

https://github.com/docker-mailserver/docker-mailserver/blob/7b4ce69c3f55c316e273acf14b973d0645983351/.github/workflows/contributors.yml#L21-L34

Looks like it checks out master branch, then creates the `contributors-update` branch but presumably that's where the local change is pushed to rather than creating the same branch locally? For the fix I added, we need the relevant commit locally that matches the SHA for the PR.

---

I am reminded that a PR also has a "merge" commit that's separate from the PR commits which is to provide the diff from the PR and the target branch, this isn't publicly viewable as a commit IIRC, but I do recall the preview docs workflow when testing different configs could target that and add comments to it like any other commit. I'm not sure if this special type of PR commit has relevance to the check status, I don't think it does.. just mentioning it if it turns out I'm mistaken about the correct commit SHA to retrieve.

casperklein:
> Is this branch not locally available at the time the command is run?

That was the problem 👍

polarathene:
I was writing a response and doing some investigation, then decided I could put out a PR again for this workflow (just not properly test it unfortunately). Looks like you bet me to it with a PR though :sweat_smile: 

---

<details>
<summary>Old WIP message</summary>

If I understand the contributors action correctly, it will commit the update locally and you must push it yourself afterwards? (unless using the create PR feature):

- [Relevant action logic](https://github.com/BobAnkh/add-contributors/blob/f3ad4d979bb184926682f9b7bf80b52941b5816c/main.py#L111).
- [PyGithub `update_file` docs](https://pygithub.readthedocs.io/en/latest/github_objects/Repository.html#github.Repository.Repository.update_file).

The branch `contributors-update` is remote only, we're pushing a modified master branch with the updated contributors file to that remote `contributors-update` branch? Therefore we should be querying the commit hash without expecting that branch name locally, which is my mistake.

We should be able to fix this by changing the branch to refer to the HEAD with `@`, alternatively the [checkout action can create a local branch which isn't a detached head state](https://github.com/actions/checkout#whats-new)

</details>

wernerfred:
The `actions/checkout@v2` puts the working dir in a detached HEAD state. So simply `git checkout` to the ref/branch we want should fix the isssue

Issue #2232, opened 2021-10-01
[NOT A BUG] Fresh 10.2.0 container crashes while booting: Failed to remove ClamAV configuration


MohammedNoureldin:
### Miscellaneous first checks

- [X] I checked that all ports are open and not blocked by my ISP / hosting provider.
- [X] I know that SSL errors are likely the result of a wrong setup on the user side and not caused by DMS itself. I'm confident my setup is correct.

### Affected Component(s)

ClamAV

### What happened and when does this occur?

```Markdown
When running a fresh container (no old mounted folders or files), the container crashes with the error:


[[  INF  ]]  Fixing Amavis permissions
[[ TASKS ]]  Cleaning up disabled ClamAV
rm: cannot remove '/etc/logrotate.d/clamav-*': No such file or directory
rm: cannot remove '/etc/cron.d/clamav-freshclam': No such file or directory
[  FATAL  ]  Failed to remove ClamAV configuration
[  ERROR  ]  Shutting down..
```
```


### What did you expect to happen?

```Markdown
The container should start normally
```

### How do we replicate the issue?

```Markdown
Just run the container using the settings below.
```

### DMS version

v10.2.0 (edge)

### How much RAM is available to DMS explicitly?

less than 2GB

### How many CPU cores are available?

less than 1 Core

### Is DMS running in a virtualized environment?

... a virtual private server (VPS) (with virtual CPU cores)

### What operating system is DMS running on?

Linux

### What instruction set architecture is DMS running on?

x86_64 / AMD64

### I/O - Persistent memory

-

### What container orchestration tool are you using?

Docker Compose

### Docker version

Docker version 20.10.8, build 3967b7d

### Docker Compose version

docker-compose version 1.29.2, build 5becea4c

### The output of `uname -a`

Linux srv02 5.4.0-84-generic #94-Ubuntu SMP Thu Aug 26 20:27:37 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

### Important environment variables

```bash
DMS_DEBUG=1

POSTMASTER_ADDRESS=postmaster@MYDOMAIN.com

SSL_TYPE=manual
SSL_CERT_PATH=/etc/ssl/pki/mail.MYDOMAIN.com.crt
SSL_KEY_PATH=/etc/ssl/pki/mail.MYDOMAIN.com.key

ENABLE_LDAP=1
LDAP_START_TLS=yes
[...]

PR #2233, opened 2021-10-02
Relaxed error handling

casperklein:
# Description

As discussed in #2232, this prevents DMS from shutdown and shows an error message instead, when not so fatal errors occurs.

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/startup/fixes-stack.sh
+++ b/target/scripts/startup/fixes-stack.sh
@@ -34,7 +34,7 @@ function _fix_var_amavis_permissions
   [[ ! -e ${AMAVIS_STATE_DIR} ]] && return 0
 
   _notify 'inf' 'Fixing Amavis permissions'
-  chown -hR amavis:amavis "${AMAVIS_STATE_DIR}" || _notify 'err' 'Failed to fix Amavis permissions'
+  chown -hR amavis:amavis "${AMAVIS_STATE_DIR}" || _shutdown 'Failed to fix Amavis permissions'
 }
 
 function _fix_cleanup_clamav
--- a/target/scripts/startup/fixes-stack.sh
+++ b/target/scripts/startup/fixes-stack.sh
@@ -21,7 +21,7 @@ function _fix_var_mail_permissions
   if find /var/mail -maxdepth 3 -a \( \! -user 5000 -o \! -group 5000 \) | read -r
   then
     _notify 'inf' 'Fixing /var/mail permissions'
-    chown -R 5000:5000 /var/mail || _notify 'err' 'Failed to fix /var/mail permissions'
+    chown -R 5000:5000 /var/mail || _shutdown 'Failed to fix /var/mail permissions'
   else
     _notify 'inf' 'Permissions in /var/mail look OK'
   fi
--- a/target/scripts/startup/fixes-stack.sh
+++ b/target/scripts/startup/fixes-stack.sh
@@ -21,7 +21,7 @@ function _fix_var_mail_permissions
   if find /var/mail -maxdepth 3 -a \( \! -user 5000 -o \! -group 5000 \) | read -r
   then
     _notify 'inf' 'Fixing /var/mail permissions'
-    chown -R 5000:5000 /var/mail || _shutdown 'Failed to fix /var/mail permissions'
+    chown -R 5000:5000 /var/mail || _notify 'err' 'Failed to fix /var/mail permissions'
   else
     _notify 'inf' 'Permissions in /var/mail look OK'
   fi
@@ -34,17 +34,17 @@ function _fix_var_amavis_permissions
   [[ ! -e ${AMAVIS_STATE_DIR} ]] && return 0
 
   _notify 'inf' 'Fixing Amavis permissions'
-  chown -hR amavis:amavis "${AMAVIS_STATE_DIR}" || _shutdown 'Failed to fix Amavis permissions'
+  chown -hR amavis:amavis "${AMAVIS_STATE_DIR}" || _notify 'err' 'Failed to fix Amavis permissions'
 }
 
 function _fix_cleanup_clamav
 {
   _notify 'task' 'Cleaning up disabled ClamAV'
-  rm /etc/logrotate.d/clamav-* /etc/cron.d/clamav-freshclam || _shutdown 'Failed to remove ClamAV configuration'
+  rm /etc/logrotate.d/clamav-* /etc/cron.d/clamav-freshclam || _notify 'err' 'Failed to remove ClamAV configuration'
 }
 
 function _fix_cleanup_spamassassin
 {
   _notify 'task' 'Cleaning up disabled SpamAssassin'
-  rm /etc/cron.daily/spamassassin || _shutdown 'Failed to remove SpamAssassin configuration'
+  rm /etc/cron.daily/spamassassin || _notify 'err' 'Failed to remove SpamAssassin configuration'
 }


target/scripts/startup/fixes-stack.sh
```diff
@@ -21,7 +21,7 @@ function _fix_var_mail_permissions
   if find /var/mail -maxdepth 3 -a \( \! -user 5000 -o \! -group 5000 \) | read -r
   then
     _notify 'inf' 'Fixing /var/mail permissions'
-    chown -R 5000:5000 /var/mail || _shutdown 'Failed to fix /var/mail permissions'
+    chown -R 5000:5000 /var/mail || _notify 'err' 'Failed to fix /var/mail permissions'
```
casperklein, line 24: Not sure about this one. If the permissions cannot be fixed, normal DMS operation is questionable?

target/scripts/startup/fixes-stack.sh
```diff
@@ -34,17 +34,17 @@ function _fix_var_amavis_permissions
   [[ ! -e ${AMAVIS_STATE_DIR} ]] && return 0
 
   _notify 'inf' 'Fixing Amavis permissions'
-  chown -hR amavis:amavis "${AMAVIS_STATE_DIR}" || _shutdown 'Failed to fix Amavis permissions'
+  chown -hR amavis:amavis "${AMAVIS_STATE_DIR}" || _notify 'err' 'Failed to fix Amavis permissions'
```
casperklein, line 37: Same here..

target/scripts/startup/fixes-stack.sh
georglauterbach: I agree. This should remain as a fatal error if you ask me:)

target/scripts/startup/fixes-stack.sh
georglauterbach: In this case, I don't think this should be a fatal error because peration of DMS is "fine" even without Amavis.

target/scripts/startup/fixes-stack.sh
georglauterbach, line 24: As mentioned, this line (and if you ask me, only this line,) should remain as a fatal error.

georglauterbach (changes requested):
LGTM, just the one line change :)

target/scripts/startup/fixes-stack.sh
MohammedNoureldin: @georglauterbach this will be executed only if some process that requires writing was not successfully, and changing the owner to 5000 failed? So in this case, changing the owner is the recovering mechanism if I see correctly?

target/scripts/startup/fixes-stack.sh
casperklein: Sure? Do you know, how postfix behaves when amavis is enabled, but not working due to wrong permissions?

target/scripts/startup/fixes-stack.sh
polarathene: @MohammedNoureldin I believe the container runs as the `docker` user (that is, not the host `docker` user if one exists, but the [internal one we create with the UID 5000](https://github.com/docker-mailserver/docker-mailserver/blob/e4eca00b252ce59e0e3c1e7f59ebccd11b2461c1/Dockerfile#L162)). These checks are run during startup, so the container is ensuring it owns and manages the `/var/mail` folder.

polarathene (dismissed):

target/scripts/startup/fixes-stack.sh
georglauterbach: I don't know. Feel free to make it a fatality as well:) My gut tells me that jut logging the error is fine though:)

georglauterbach (approved):

target/scripts/startup/fixes-stack.sh
casperklein: @polarathene What's your take on this?

georglauterbach: Wow, quick response :D I'm fine with you changing it to a fatal error, really. When I think about it, having a fatal error may be the best option after all - fail fast is a good approach!

target/scripts/startup/fixes-stack.sh
MohammedNoureldin: @polarathene I am not sure if this is really important, because I used to use another directory for the email, that is why I did not really care about */var/mail*. Yes it is the default directory, but I find it again not as fatal error that deserves shutting down, especially if the user knows that he is doing it on purpose. Anyway, at least this should not happen normally and should not cause the container to crash even if the user decided to use another directory.

polarathene: > I used to use another directory to store the emails, that is why I did not really care about _/var/mail_.

This is an internal container location. The host system can volume mount any location to the containers `/var/mail`. I assume you are talking about the local host location being different from `/var/mail` for you? That's fine.

target/scripts/startup/fixes-stack.sh
polarathene: > @polarathene What's your take on this?

No opinion as I'm not too familiar of the impact :)

If it's important to do this for functionality to be correct, then do so. We can always revert later with a patch release if it's problematic.

Alternatively, if just raising an error in the logs is sufficient... that should still surface in any bug reports?

target/scripts/startup/fixes-stack.sh
MohammedNoureldin: @polarathene actually I meant the internal location in the container, but you are right, this makes not much sense to change it (though we can). Especially that it can be counted on the system somewhere else.

georglauterbach: I disagree as having the correct permissions on this directory is rather important. It should therefore end the container start if something went wrong when changing permissions.

MohammedNoureldin: @georglauterbach I agree with you in case the user is using that directory, but this directory can be configured in Dovecot to somewhere else, and by this the user still can use a different directory with invalid permissions. But as you said, this makes sense when using the default directory, thus it can be left as fatal error at the moment.

georglauterbach: You can of course overwrite all the configuration of this project, but we need to assure correct functionality for _the defaults provided by us_. When you change the defaults, you'll have to make sure all works well for you.

MohammedNoureldin: @georglauterbach you are right. As long as it does not prevent changing the directory and shutdown the container, this is a nice check. And it seems to only have benefits.

target/scripts/startup/fixes-stack.sh
casperklein: > No opinion as I'm not too familiar of the impact :)

Same as Georg and me 😆

I'll revert to `_shutdown` for another reason. Even when postfix doesn't care if amavis is not working, I see a kind of security risk.
Spamassassin & ClamAV will not scan emails. Therefore potential dangerous content can be delivered to user mailboxes.

MohammedNoureldin: Let us try it like this. As long as it has no unwanted effects, it should be good to stay as a fatal error.

MohammedNoureldin: > > No opinion as I'm not too familiar of the impact :)
> 
> Same as Georg and me 😆
> 
> I'll revert to `_shutdown` for another reason. Even when postfix doesn't care if amavis is not working, I see a kind of security risk. Spamassassin & ClamAV will not scan emails. Therefore potential dangerous content can be delivered to user mailboxes.

Let us try it like this. As long as it has no unwanted effects, it should be good to stay as a fatal error.

georglauterbach (approved):
LGTM

polarathene (approved):

PR #2234, opened 2021-10-02
Update README to reflect changes of `v10.2.0`

georglauterbach:
# Description

Because the container will now exit on startup failure if no accounts exits, we have to update the README's section for beginners. Furthermore, I added another section containing information about not using `docker start / stop` and about the help users can get by running the `setup help` command inside the container.

This issue is related, but only an example: #2232

## Type of change

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,7 @@ You can then proceed by creating the postmaster alias and by creating DKIM keys.
 docker-compose up -d mailserver
 
 # you may add some more users
+# for SELinux, use -Z
 ./setup.sh [-Z] email add <user@domain> [<password>]
 
 # and configure aliases, DKIM and more
--- a/README.md
+++ b/README.md
@@ -134,7 +134,6 @@ docker-compose up -d mailserver
 ./setup.sh [-Z] email add <user@domain> [<password>]
 
 # and configure aliases, DKIM and more
-# for SELinux, use -Z
 ./setup.sh [-Z] alias add postmaster@<domain> <user@domain>
 ./setup.sh [-Z] config dkim
 ```
--- a/README.md
+++ b/README.md
@@ -108,16 +108,35 @@ If you're using `docker-mailserver` version `v10.1.x` or below, you will need to
 
 ### Get up and running
 
+#### First Things First
+
+**Use `docker-compose up / down`, not `docker-compose start / stop`**. Otherwise, the container is not properly destroyed and you may experience problems during startup because of inconsistent state.
+
+You are able to get a full overview of how the configuration works by either running:
+
+1. `./setup.sh help` which includes the options of `setup.sh`.
+2. `docker run --rm docker.io/mailserver/docker-mailserver:latest setup help` which provides you with all the information on configuration provided "inside" the container itself.
+
+#### Starting for the first time
+
+On first start, you will likely see an error stating that there are no mail accounts and the container will exit. You must now do one of two things:
+
+1. Use `setup.sh` to help you: `./setup.sh email add <user@domain> <password>`. You may need to provide the correct configuration path (to the directory mounted to `/tmp/docker-mailserver` inside the container yourself) with the `-c` option. This will spin up a new container, mount your configuration volume, and create your first account.
+2. Execute the complete command yourself: `docker run --rm -v ./docker-data/dms/config/:/tmp/docker-mailserver/ docker.io/mailserver/docker-mailserver setup email add <user@domain> <password>`. Make sure to mount the correct configuration directory.
+
+You can then proceed by creating the postmaster alias and by creating DKIM keys.
+
 ``` BASH
 docker-compose up -d mailserver
 
-# for SELinux, use -Z 
+# you may add some more users
 ./setup.sh [-Z] email add <user@domain> [<password>]
+
+# and configure aliases, DKIM and more
+# for SELinux, use -Z
 ./setup.sh [-Z] alias add postmaster@<domain> <user@domain>
 ./setup.sh [-Z] config dkim
 ```
-
-If you're seeing error messages about unchecked errors, please **verify that you're using the right version of `setup.sh`**. Refer to the [Get the tools](#get-the-tools) section and / or execute `./setup.sh help` and read the `VERSION` section.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -108,17 +108,36 @@ If you're using `docker-mailserver` version `v10.1.x` or below, you will need to
...
+
+1. `./setup.sh help` which includes the options of `setup.sh`.
+2. `docker run --rm docker.io/mailserver/docker-mailserver:latest setup help` which provides you with all the information on configuration provided "inside" the container itself.
+
+#### Starting for the first time
+
+On first start, you will likely see an error stating that there are no mail accounts and the container will exit. You must now do one of two things:
+
+1. Use `setup.sh` to help you: `./setup.sh email add <user@domain> <password>`. You may need to provide the correct configuration path (to the directory mounted to `/tmp/docker-mailserver` inside the container yourself) with the `-c` option. This will spin up a new container, mount your configuration volume, and create your first account.
+2. Execute the complete command yourself: `docker run --rm -v ./docker-data/dms/config/:/tmp/docker-mailserver/ docker.io/mailserver/docker-mailserver setup email add <user@domain> <password>`. Make sure to mount the correct configuration directory.
```
casperklein, line 125: @polarathene At first, I thought there is an error: `-v ./docker-data/dms/config/:/tmp/docker-mailserver`

But after looking in our docker-compose.yml this seems to be correct. I was wrongly assuming, this was an absolute path: `-v /docker-data/dms/config/` (dot removed). E.g. Some kind of global `/docker-data` directory, with a subdirectory for our DMS data.

Given you have a directory for your DMS stuff. Then you create a subdirectory for your "data". Why is there in the example another directory level "dms"?

This is from our old docker-compose file and how my config looks like:

```
      - ./data/maildata:/var/mail
      - ./data/maillogs:/var/log/mail
```

I see no benefit in changing it to

```
      - ./data/dms/maildata:/var/mail
      - ./data/dms/mailstate:/var/mail-state
      - ./data/dms/maillogs:/var/log/mail
```

I hope you understand what I mean 😉

README.md
```diff
@@ -108,17 +108,36 @@ If you're using `docker-mailserver` version `v10.1.x` or below, you will need to
...
+
+You can then proceed by creating the postmaster alias and by creating DKIM keys.
+
 ``` BASH
 docker-compose up -d mailserver
 
-# for SELinux, use -Z 
+# you may add some more users
```
casperklein, line 132:
```suggestion
# you may add some more users
# for SELinux, use -Z
```

README.md
```diff
@@ -108,17 +108,36 @@ If you're using `docker-mailserver` version `v10.1.x` or below, you will need to
...
 ./setup.sh [-Z] email add <user@domain> [<password>]
+
+# and configure aliases, DKIM and more
+# for SELinux, use -Z
```
casperklein, line 136:
```suggestion
```

casperklein (changes requested):

README.md
```diff
@@ -108,17 +108,36 @@ If you're using `docker-mailserver` version `v10.1.x` or below, you will need to
...
+You are able to get a full overview of how the configuration works by either running:
+
+1. `./setup.sh help` which includes the options of `setup.sh`.
+2. `docker run --rm docker.io/mailserver/docker-mailserver:latest setup help` which provides you with all the information on configuration provided "inside" the container itself.
+
+#### Starting for the first time
+
+On first start, you will likely see an error stating that there are no mail accounts and the container will exit. You must now do one of two things:
+
+1. Use `setup.sh` to help you: `./setup.sh email add <user@domain> <password>`. You may need to provide the correct configuration path (to the directory mounted to `/tmp/docker-mailserver` inside the container yourself) with the `-c` option. This will spin up a new container, mount your configuration volume, and create your first account.
```
polarathene, line 124:
```suggestion
1. Use `setup.sh` to help you: `./setup.sh email add <user@domain> <password>`. You may need  the `-c` option to provide the local path for persisting configuration (_a directory that mounts to `/tmp/docker-mailserver` inside the container_). This will spin up a new container, mount your configuration volume, and create your first account.
```

polarathene (changes requested):

README.md
polarathene: > I see no benefit in changing it to

This was a part of the normalization for docs. The layout is not required, but it was chosen to have all examples mount volumes into `docker-data/` dir locally. Then within that dir for volumes, a service dir contains the volumes as sub-directories.

In most cases, perhaps you'll only have a `docker-mailserver` and then this additional server dir `dms/` seems unnecessary. Otherwise it contains your `mail-*/` and `config/` volume dirs, and keeps them separate/organized from other services.

In the docs especially this was an area that wasn't as easy to grok or search, since it was quite inconsistent or referring to `config/` in different ways. You can look over the [scoped commit for this change](https://github.com/docker-mailserver/docker-mailserver/pull/2206/commits/9a3d105898c0ca64fa6a6041cd88a0cd5b0f5a0b?diff=split&w=1). It was mainly to address any ambiguity and provide context of a common layout/hierarchy when other services with volume mounts have examples.

You can use anything you like or remove the `docker-data/dms/` prefix entirely. For an experienced Docker user, they should be able to realize that, it only matters that the volume mounts are made with distinct mappings (local folders to internal ones), where they're located is up to the user.

It is easier for our maintenance though, knowing that we can find any reference to these volumes in our docs and related files by searching `docker-data/dms/`, especially for instances of `config/` which isn't always related to the same local config volume path.

---

> At first, I thought there is an error: `-v ./docker-data/dms/config/:/tmp/docker-mailserver`
> 
> I was wrongly assuming, this was an absolute path

I think it's still an error. In `docker-compose.yml` you use the relative `./`, but CLI doesn't like that IIRC, requires: `"${PWD}/docker-data/dms/config/:/tmp/docker-mailserver"`.

casperklein: > I think it's still an error.

You are right. Overlooked that 👍

casperklein:
```suggestion
2. Execute the complete command yourself: `docker run --rm -v "${PWD}/docker-data/dms/config/:/tmp/docker-mailserver/" docker.io/mailserver/docker-mailserver setup email add <user@domain> <password>`. Make sure to mount the correct configuration directory.
```

casperklein (approved):

polarathene (approved):

Issue #2239, opened 2021-10-03
[BUG] Let's encrypt certificate renew not recognised. mailserver uses old one and not updated one. 


zaphoodb:
### Miscellaneous first checks

- [X] I checked that all ports are open and not blocked by my ISP / hosting provider.
- [X] I know that SSL errors are likely the result of a wrong setup on the user side and not caused by DMS itself. I'm confident my setup is correct.

### Affected Component(s)

mailserver does not use renewed let's encrypt certificate

### What happened and when does this occur?

```Markdown
I'm using traefik v2 to get and renew certificates. The traefik json which holds the certificates has been mounted as volume into the mailserver docker container. When the certificate gets renewed by traefik (~4 weeks before expiry), the mailserver does not recognise this certificate and keeps using the old one. 

I checked certificate with https://www.checktls.com/TestReceiver and it still shows the old one. Opening the Website that uses the same hostname and certificate, it shows the renewed one. 

When I do a restart of the docker container, the new certificate is recognized and from there on used.
```

### What did you expect to happen?

```Markdown
docker mailserver should recognise the updated / renewed certificate (by traefik2) without restarting the container.
```

### How do we replicate the issue?

```Markdown
1. Setup traefik2, nginx webserver and docker mailserver as described in the docu
2. wait until let's encrypt certificate get's renewed by traefik2
3. check that mailserver does not pick renewed certificate
...
```

### DMS version

v10.1.2

### How much RAM is available to DMS explicitly?

more than 4GB

### How many CPU cores are available?

less than 4 Cores

### Is DMS running in a virtualized environment?

... a virtual private server (VPS) (with virtual CPU cores)

### What operating system is DMS running on?

Linux

### What instruction set architecture is DMS running on?

x86_64 / AMD64

### I/O - Persistent memory

docker volume mount ext4

### What container orchestration tool are you using?

Docker Compose

### Docker Compose version

docker-compose version 1.28.5

### The output of `uname -a`

Linux 4.19.0-17-amd64 #1 SMP Debian 4.19.194-3 (2021-07-18) x86_64 GNU/Linux

### Important environment variables

```bash
Mount point within my docker-compose file: 

      - /data/acme/:/etc/letsencrypt/

I tried 

      - /data/acme/acme.json:/etc/letsencrypt/acme.json:ro

last time but same result.
```

### What level of experience do you have with Docker and mail servers?

Trust me, I'm a (computer) engineer! [expert]

### Code of conduct

[...]

PR #2245, opened 2021-10-09
Improved logic to handle HOSTNAME and DOMAINNAME + Improving handling of Letsencrypt changes + build docs WARNINGs should throw exit code

NorseGaud:
# Description

Original PR that may have broken this: https://github.com/docker-mailserver/docker-mailserver/pull/2175/files

In https://github.com/docker-mailserver/docker-mailserver/issues/2239, the user has:

```
hostname: mail
domainname: somedomain.com

environment:
- SSL_TYPE=letsencrypt
- SSL_DOMAIN=mail.somedomain.com
```

And the certificate was triggering changedetector: `[[  INF  ]]  Cert found in /etc/letsencrypt/acme.json for mail.somedomain.com`

Yet, they had two different directories under `/etc/letsencrypt`:

```
root@mail:/etc/letsencrypt# ls -al live/
total 16
drwxr-xr-x 4 root root 4096 Oct  6 10:20 .
drwxr-xr-x 3 root root 4096 Oct  5 20:15 ..
drwxr-xr-x 2 root root 4096 Oct  6 10:20 mail
drwxr-xr-x 2 root root 4096 Oct  5 20:15 mail.somedomain.com
```

They indicated that they found a solution by:

> I copied the files from mail to mail.somedomain.com and restarted postfix/dovecot via supervisorctl restart dovecot / postfix. The renewed certificate is in use now

This means that `mail` is getting the renewal (`_extract_certs_from_acme`), but `mail.somedomain.com` is being watched. 

The user has explicitly set `SSL_DOMAIN` to `mail.somedomain.com` and yet the extraction seems to be happening into a folder named `mail`. Under the `setup-stack.sh` I see:

```
      if [[ -f /etc/letsencrypt/acme.json ]]
      then
        if ! _extract_certs_from_acme "${SSL_DOMAIN}"
        then
          if ! _extract_certs_from_acme "${HOSTNAME}"
          then
            _extract_certs_from_acme "${DMS_DOMAIN}"
          fi
        fi
      fi

      # first determine the letsencrypt domain by checking both the full hostname or just the domainname if a SAN is used in the cert
      if [[ -e /etc/letsencrypt/live/${HOSTNAME}/fullchain.pem ]]
      then
        LETSENCRYPT_DOMAIN=${HOSTNAME}
      elif [[ -e /etc/letsencrypt/live/${DMS_DOMAIN}/fullchain.pem ]]
      then
        LETSENCRYPT_DOMAIN=${DMS_DOMAIN}
      else
        _notify 'err' "Cannot access '/etc/letsencrypt/live/${HOSTNAME}/fullchain.pem' or '/etc/letsencrypt/live/${DMS_DOMAIN}/fullchain.pem'"
        return 1
      fi
```

There are a few problems here:

1. Extraction of their acme is happening into `mail` and not `mail.somedomain.com`
2. The if statements after that don't set `LETSENCRYPT_DOMAIN` to `SSL_DOMAIN`

***For 1***, we see `[[  INF  ]]  Cert found in /etc/letsencrypt/acme.json for mail.somedomain.com`, so that's not the issue. HOWEVER, the code shows that ${1} is being used instead of ${HOSTNAME} so we aren't seeing accurately what is being used:

```
  if [[ -n "${KEY}${CERT}" ]]
  then
    mkdir -p "/etc/letsencrypt/live/${HOSTNAME}/"

    echo "${KEY}" | base64 -d >/etc/letsencrypt/live/"${HOSTNAME}"/key.pem || exit 1
    echo "${CERT}" | base64 -d >/etc/letsencrypt/live/"${HOSTNAME}"/fullchain.pem || exit 1
    _notify 'inf' "Cert found in /etc/letsencrypt/acme.json for ${1}"
```

To fix this, we shouldn't be using HOSTNAME and instead use `${1}`.

***For 2***, I have added the required code

---

It also looked as if we were testing acme with a "changed" cert that was for `example.com` and not `mail.my-domain.com`. I'm unsure how this worked... I've expanded the tests to test the regular situation of mai.my-domain.com and then also the wildcard acme for example.com.

---

Seems as if the docs generation script wasn't properly exiting on WARNING of broken links. I fixed that too.

## Type of change

## Checklist:


polarathene:
> Maybe I don't understand this, but the old DOMAINNAME we're creating here is wrong

Just in-case anyone isn't aware... `HOSTNAME` and `DOMAINNAME` are system ENV, not custom ENV that we define. Docker additionally supports both as options for the CLI or docker-compose config, but has a spotty history of how they were handled (it's changed over time and at one point IIRC CLI and docker-compose behaved differently).

Our tests also I think only set `hostname` while docker-compose configs would set both, plus we have a separate ENV introduced to further mess around with that, in addition to any other related logic in our startup scripts. There's also different ways for software to check the hostname and domainname, several CLI utilities that output different FQDN splits to `HOSTNAME` and `DOMAINNAME`, and a kernel interface. The outputs also can differ across different distro base images (eg Alpine vs Debian), and other configurations with docker (user-ns causes issues IIRC).

So I'm not surprised if reliance on these values may be wonky :sweat_smile:

NorseGaud:
Tests are failing
```
❯ ./test/bats/bin/bats --timing test/mail_disabled_clamav_spamassassin.bats 
 ✗ File mail_disabled_clamav_spamassassin.bats []
   (from function `repeat_until_success_or_timeout' in file test/test_helper/common.bash, line 37,
    from function `wait_for_tcp_port_in_container' in file test/test_helper/common.bash, line 80,
    from function `wait_for_smtp_port_in_container' in file test/test_helper/common.bash, line 85,
    from function `setup_file' in test file test/mail_disabled_clamav_spamassassin.bats, line 24)
     `wait_for_smtp_port_in_container mail_disabled_clamav_spamassassin' failed
   0dae13b3d64d9074575e5f60185841f14ed1a26ad5ad1ace73be88da26c4e58d
   Timed out on command: docker exec mail_disabled_clamav_spamassassin /bin/sh -c nc -z 0.0.0.0 25
   mail_disabled_clamav_spamassassin
   bats warning: Executed 1 instead of expected 7 tests

7 tests, 1 failure, 6 not run in 123 seconds

```

NorseGaud:
@polarathene , you can wait to review this. It's still a WIP. I one again forgot to mark it as such.

NorseGaud:
Anyone know why we set `hostname` in the docker-compose.yml? The help output for `docker run` says:

```
      --domainname string              Container NIS domain name
```
 
 Do our users even use NIS?

polarathene:
I'll look at doing a review tomorrow or so. I think I've already been heavily verbose (and a tad repetitive) in this PR discussion already today :sweat_smile: 

Apologies if my responses are a bit overwhelming in size, this is something I want to be thorough about and understand well enough going forward. I've tried to condense/revise the information (_at least the investigation insights comment prior to this one_).

Nice catch with the docs fixes btw! :grinning:

NorseGaud:
Hey @polarathene , can you finish up the changes for the PR to be approved? I don't have any time left to invest in researching how this all works and what you mean in your tasks for changes ( you're on a different level, man 😉  ) It seems you're far more knowledgeable than I am at this point that you could do a better job.

I believe I have edit permissions allowed for maintainers.

polarathene:
@NorseGaud yeah no worries :+1: 

Thanks for putting all the time and effort into this so far, it's great that we're addressing this concern.

Issue #2238, opened 2021-10-03
[BUG] The created folders for the different users miss +x permissions


MohammedNoureldin:
### Miscellaneous first checks

- [X] I checked that all ports are open and not blocked by my ISP / hosting provider.
- [X] I know that SSL errors are likely the result of a wrong setup on the user side and not caused by DMS itself. I'm confident my setup is correct.

### Affected Component(s)

Mail home folder permissions (tested with v10.1.2 and 10.2.0)

### What happened and when does this occur?

```Markdown
When trying to use the mailserver, some permission error related to mail home directory is shown.
```

### What did you expect to happen?

```Markdown
To not get the error
```

### How do we replicate the issue?

```Markdown
Try the attached configuration with LDAP server.
```

### DMS version

Tested with 10.1.2 and 10.2.0

### How much RAM is available to DMS explicitly?

less than 2GB

### How many CPU cores are available?

less than 1 Core

### Is DMS running in a virtualized environment?

... a virtual private server (VPS) (with virtual CPU cores)

### What operating system is DMS running on?

Linux

### What instruction set architecture is DMS running on?

x86_64 / AMD64

### I/O - Persistent memory

-

### What container orchestration tool are you using?

Docker Compose

### Docker version

Docker version 20.10.8, build 3967b7d

### Docker Compose version

docker-compose version 1.29.2, build 5becea4c

### The output of `uname -a`

Linux srv02 5.4.0-84-generic #94-Ubuntu SMP Thu Aug 26 20:27:37 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

### Important environment variables

```bash
DMS_DEBUG=1

POSTMASTER_ADDRESS=postmaster@MYDOMAIN.com

SSL_TYPE=manual
SSL_CERT_PATH=/etc/ssl/pki/mail.MYDOMAIN.com.crt
SSL_KEY_PATH=/etc/ssl/pki/mail.MYDOMAIN.com.key

ENABLE_LDAP=1
LDAP_START_TLS=yes
LDAP_SERVER_HOST=dc1.internal.MYDOMAIN.com
LDAP_SEARCH_BASE=dc=internal,dc=MYDOMAIN,dc=com
LDAP_BIND_DN=cn=Administrator,cn=users,dc=internal,dc=MYDOMAIN,dc=com
LDAP_BIND_PW=Ch@ngeMe
# Specifies how ldap should be asked for users.
[...]

PR #2256, opened 2021-10-19
Don't change owners in /var/mail recursively

MohammedNoureldin:
# Description

Changing the whole owners of folders inside */var/mail* ruins the whole LDAP users permissions on their own mail folders. Thus, only */var/mail* owner should be fixed if necessary (not recursively).

Fixes: https://github.com/docker-mailserver/docker-mailserver/issues/2238

## Type of change

## Checklist:

### DIFF
--- a/test/mail_with_ldap.bats
+++ b/test/mail_with_ldap.bats
@@ -51,8 +51,8 @@ function setup_file() {
     -e DOVECOT_TLS=no \
     -e DOVECOT_PASS_FILTER="(&(objectClass=PostfixBookMailAccount)(uniqueIdentifier=%n))" \
     -e DOVECOT_USER_FILTER="(&(objectClass=PostfixBookMailAccount)(uniqueIdentifier=%n))" \
-    -e DOVECOT_USER_ATTRS==uid=%{ldap:UID},=gid=5000,=home=/var/mail/%Ln,=mail=maildir:~/Maildir \
-    -e DOVECOT_PASS_ATTRS=UID=user,userPassword=password \
+    -e DOVECOT_USER_ATTRS="=uid=%{ldap:UID},=gid=5000,=home=/var/mail/%Ln,=mail=maildir:~/Maildir" \
+    -e DOVECOT_PASS_ATTRS="UID=user,userPassword=password" \
     -e REPORT_RECIPIENT=1 \
     -e ENABLE_SASLAUTHD=1 \
     -e SASLAUTHD_MECHANISMS=ldap \


polarathene:
> With this "fix" applied, what happens when files in subdirectorys of `/var/mail` have the wrong permission? They won't be fixed and therefore not accessible by dovecot for example.

I think the assumption was that it was not clear what exactly required `5000:5000` ownership within `/var/mail` to function correctly. The PR author found that functionality for them worked well with their LDAP user permissions, as have other non-LDAP users apparently in the past... while others haven't.

Based on what information I could lookup via `git blame` and searching issues, it wasn't clear what `chown -R` on `/var/mail` was "Fixing permissions" for. Deferred to tests, which then provided a variety of failures, but I don't have time to look into them and delegate that to @MohammedNoureldin  to resolve if they want to fix their issue via this PR.

Are you saying that dovecot requires `/var/mail` to have ownership of `5000:5000` to function correctly?

---

> You could for example only turn off recursive changes when LDAP is enabled.

I don't know if that's an appropriate "fix" if the tests would still fail with LDAP settings. I'm reluctant to approve the PR if it's prone to subtle failures elsewhere that the author hasn't run into themselves.

I haven't looked at the failing tests, so perhaps the risk isn't there for LDAP :man_shrugging:

MohammedNoureldin:
Hello again @polarathene @georglauterbach @casperklein,

I took a look on the tests, I don't fully understand why it is not able to find a specific folder under */var/mail*, probably it is not able now to write these folders, but I am not sure why.

So here I need some help from you please, what would you suggest either to fix the tests (I am not sure if fixing the tests is a good idea here), or do you suggest anything else to change in the code to make the tests no fail? What would you suggest? Will only checking if LDAP is activated and execute the code otherwise sufficient?

I still have the opinion that this recursive owner changing is not required, and is rather not good, but I don't want to trigger a big refactoring here. Any small fix that gets things to work properly for both LDAP and non-LDAP is fine.

Any suggestions how I should go further?

polarathene:
> @polarathene you're right, I wouldn't approve this either as long as tests fail.

@georglauterbach 

I was more referring to not hacking around the tests (eg: a feature flag that applies this PR only to LDAP users, when the failing tests are not configured for LDAP, so the tests pass), or modifying the tests to pass in a manner that hides the issue as well.

---

> If you migrate to a new server/storage etc, or forget to backup permissions as well, the permissions in the data volume may be incorrect. I guess, this fix ensures that all permissions are as expected and the files are accessible by the various services.

@casperklein 

I already pointed out that case as one of a few related to the issue this PR is addressing [in this reveiw comment](https://github.com/docker-mailserver/docker-mailserver/pull/2256#discussion_r732315322). Although it's not clear why that's happening.

> I think so, but cannot confirm this 100%. This was just an assumption based on the current script. I suppose, no one would introduce a fix for permissions, when there aren't any issues that could arise.

See the same link I just provided. This "fix" was introduced in the very first commit of the project by the original author.

They've not always made sound choices and didn't seem to explain it's purpose during a relevant discussion I linked, I can't say I'm confident in the choice based on the history I dug up. They likely went with it because "it works for me" as a fix at the time.

---

> But for me it seems like it can be removed safely, as it is not needed to recusively change all permissions, only changing /var/mail should be sufficient. Or what do you think?

@MohammedNoureldin 

Without looking into the failing tests, I would like to know that they would pass for an LDAP setup if you applied your PR via a feature toggle (doesn't have to be LDAP specific).

My concern is that if tests are failing, just because you don't notice any issues doesn't mean that there are none, or that it won't affect other LDAP users as a result. This would be less of an issue by using an opt-in ENV, but I am not particularly fond of that either without a proper understanding of why it's needed (not your requirement, but why we can't use it by default due to tests failing).

---

> I took a look on the tests, I don't fully understand why it is not able to find a specific folder under /var/mail, probably it is not able now to write these folders, but I am not sure why.

Permissions....? That's literally what this PR is changing.

The good news is that they were all `ls` and `grep` commands. Review the tests and how the filepaths that are checked are added/created. Try identify where the permissions are coming from, presumably it's the host system rather than the container. If the UID and GID make the files inaccessible to `docker-mailserver`, that's an issue.

What might work is just recursively adding the GID :man_shrugging: But then that should work for you if your LDAP users belonged to the 5000 group within `docker-mailserver` (should any UID from that group try to interact with the files). One of the linked issues in my review was about GID, but the user wanted to use a different GID within the container to match the one they already had host users sharing, I don't consider that an issue to worry about though.

When you have identified the failure cause for tests and can detail it here, we can discuss further.

> What would you suggest? Will only checking if LDAP is activated and execute the code otherwise sufficient?

You may consider adding an ENV to ignore applying the `chown -R`, it should be opt-in by default.

Personally I would prefer you still look into the test failures for more information as to why they're failing, as if you add that ENV it should probably come with supporting tests anyway.

> I still have the opinion that this recursive owner changing is not required, and is rather not good, but I don't want to trigger a big refactoring here. Any small fix that gets things to work properly for both LDAP and non-LDAP is fine.

While I agree, we need to understand why that is not the case with the tests first.

I was not able to dig up anything in project history that demonstrates what the permissions are being fixed for. We do thankfully have failing tests that should help enlighten us on that, but I do not have the time to do so myself.

If you want to resolve this, you'll need to investigate and find this information.

---

> Any suggestions what I should do further?

Yes. If you do not want to pursue the above, you can choose to volume mount your fix over the related file within the container on your end instead of upstreaming it here. I assume `user-patches.sh` wouldn't work, otherwise I'd advise that and adding to our documentation the workaround.

The other maintainers may be fine with the opt-in ENV to disable the `chown -R` as well, but it would be great to at least try look into the test failures a little for insights.

polarathene:
> Because it was not really clear for me why the file is not being written to the expected folder and why is this happening.

What did you try so far?

I kindly provided all the failing tests with the common errors and full failure logged output [in this earlier comment](https://github.com/docker-mailserver/docker-mailserver/pull/2256#issuecomment-947240208).

Failing bats test files: `mail_hostname`, `mail_lmtp_ip`, `mail_privacy`, `mail_spam_junk_folder`, `mail_special_use_folders`, `mail_with_mdbox`, `mail_with_sdbox`. All can be [found in the project `test/` folder](https://github.com/docker-mailserver/docker-mailserver/tree/master/test).

Here is a small test, [`mail_with_mdbox.bats`](https://github.com/docker-mailserver/docker-mailserver/blob/master/test/mail_with_mdbox.bats):
- It only has one test: ["checking dovecot mailbox format: mdbox file created"](https://github.com/docker-mailserver/docker-mailserver/blob/34ba3c2a5b0f709a1680a779fda3da338535a063/test/mail_with_mdbox.bats#L37-L43).
- It's [container setup is here](https://github.com/docker-mailserver/docker-mailserver/blob/34ba3c2a5b0f709a1680a779fda3da338535a063/test/mail_with_mdbox.bats#L11-L26) in the `setup_file()` method.
- The failure of the test is due to `ls /var/mail/localhost.localdomain/user1/storage/m.1` not being able to access the file. That is presumably created when permissions enable it to, by the earlier line in the test that sends a mail from the content of `/tmp/docker-mailserver-test/email-templates/existing-user1.txt`. If you look at `setup_file()`, you can see this file is mounted from our test files, and the content is [here](https://github.com/docker-mailserver/docker-mailserver/blob/master/test/test-files/email-templates/existing-user1.txt).
- This failure is probably related to [the account being created here in the Makefile](https://github.com/docker-mailserver/docker-mailserver/blob/34ba3c2a5b0f709a1680a779fda3da338535a063/Makefile#L35-L40), which is required to run the test. The user is added via `doveadm` running from the mail container.
- We don't have any volume mount related to `/var/mail` from what I can see, and the account generation itself has not caused problems in the past, which AFAIK will write a file to the config volume. So when does `/var/mail` get populated normally?
- It must be some other UID/GID created during startup that we would normally correct to `5000:5000`? By not doing that, either that mail that is sent and received in the test doesn't get stored, or permissions are preventing the `ls` command to access that location if it otherwise exists.

There should be enough information there to follow the test and understand how to reproduce the failure locally. You want to manually do so, and then investigate inside the container `docker exec -it container_name bash`. You'll likely find that something in `/var/mail` is not `5000:5000` and that it may be the cause of the problem as was pointed out by casper earlier with Dovecot.

MohammedNoureldin:
Hi Casper,

> If this breaks LDAP, it should be fixed of course. I personally like the "chown -R" approach for non LDAP setups. It does not hurt, but provides an additional layer of protection against possible permission problems.

I have the opposite opinion :D if the user does something wrong with permissions, then he should be responsible for fixing them himself, not automatically overwrite all existing one just because some users may do mistakes, which will hurt other uses like me who do need other permissions/onwer.

> What makes me wonder or I didn't fully understand it yet, why this problem was not noticed earlier. LDAP was introduced in 2016. It sounds to me like a serious issue for LDAP users.

I think it is because most users used hard-coded uid=5000 in `DOVECOT_USER_ATTRS`, and no body kept in mind that it is much better to set the LDAP user as the real owner of each folder instead of assigning a fixed unified owner for all mail-boxes, which looks like a security issue.

polarathene:
> **Unfortunately, I cannot test this soon**. Can anyone of you test this? At least the unit tests says to a negative answer. While the tests may also be written for a specific case which is not totally correct. But testing is the only reliable judge.

@MohammedNoureldin 

The questions were aimed at Casper regarding being in favor of keeping the `chown -R 5000:5000`, but thanks for chiming in :sweat_smile: 

The tests are all failing for the same reason AFAIK, I just haven't had time to investigate myself, so I've tried to delegate this to you with some guidance.

I suspect that the startup `chown -R 5000:5000` for `/var/mail` masks an actual problem it's intended to fix as a broad stroke, while not clearly defining it (_it probably wasn't investigated when introduced, but used since it resolved the issue without having to investigate further_). We just need to verify this is the case and the more appropriate way to address it. Then your original PR commits that I reviewed should be sufficient.

> I will try to test it as soon as possible, nevertheless, if anyone has available infrastructure now to test it would be great.

You can follow the walk-through guidance I provided on the referenced test case earlier. The docker command is there and is pretty much copy/paste, just replace the config directory with an empty one like a normal config setup, but make the test files volume available. Then reproduce by following the same shell commands in the failing test case.

---

> I personally like the "chown -R" approach for non LDAP setups. It does not hurt, but provides an additional layer of protection against possible permission problems.

@casperklein 

It sounds like it would not need to be run more than once in practice. Unless `/var/mail` is constantly being fed data with incorrect UID/GID ownership, which is assumed not to happen for `docker-mailserver` out of the box (LDAP aside).

In that sense, this is a troubleshooting feature better suited to a FAQ entry in our docs and a possible `setup.sh` method for convenience. Likewise it could be added as a `user-patches.sh` script to run at startup just the same.

I would prefer we had actual issues raised to justify it being commonly required/used that it should be put back into our startup scripts, which if I understand the issue correctly, isn't going to happen.

> What makes me wonder or I didn't fully understand it yet, why this problem was not noticed earlier. LDAP was introduced in 2016. It sounds to me like a serious issue for LDAP users.

This is fair, but I'm not sure how well LDAP is used in practice for `docker-mailserver`. Just some observations:

We had a few contributions this year by one user, and then pretty much nothing for 4-5 years IIRC. Our tests for LDAP likewise use an LDAP docker image that is pinned to a very old release version.

Some users may have tried to use LDAP with `docker-mailserver` in the past, potentially raised some issues, and failing to get a satisfactory setup working, migrated to an alternative project instead. There's also been quite a bit of major releases, especially over the past year where we may have surfaced problems through changes or fixes. Some issues by less observant users may also go unnoticed :man_shrugging: 

This particular one based on description only becomes apparent after restarting the container once LDAP users have added data since startup AFAIK. Some deployments may not experience the issue until restarting the container for maintenance or update (and we know that some updates can lag behind for some time).

@MohammedNoureldin also made some good points that are probably more applicable as to why, thanks!

test/mail_with_ldap.bats
```diff
@@ -51,6 +51,8 @@ function setup_file() {
     -e DOVECOT_TLS=no \
     -e DOVECOT_PASS_FILTER="(&(objectClass=PostfixBookMailAccount)(uniqueIdentifier=%n))" \
     -e DOVECOT_USER_FILTER="(&(objectClass=PostfixBookMailAccount)(uniqueIdentifier=%n))" \
+    -e DOVECOT_USER_ATTRS="=uid=%{ldap:UID},=gid=5000,=home=/var/mail/%Ln,=mail=maildir:~/Maildir" \
```
Murazaki, line 54: I think that by adding Maildir here, the path to the mail folder changed to `/var/mail/<domain>/<user>/Maildir` which led to the test suite error. (I'm using this scheme too but the test suite might not have been doing it)

MohammedNoureldin: Thank you for contributing in this, @Murazaki! Something still missing apparently, could you check it? I have been working on another important issue in the last few days, maybe we can close this permission issue also if we work on this together.

Murazaki: Hi Mohammed. I tried here but couldn't finish either. There might be a file permission issue: 
https://github.com/Murazaki/docker-mailserver/tree/patch-1

georglauterbach:
This has stalled significantly. As part of #3289, I will close this (see my rationale here: <https://github.com/docker-mailserver/docker-mailserver/issues/3289#issuecomment-1520172545>). 

@MohammedNoureldin this does not mean these changes are not worthwhile; I just cannot maintainer PRs that are years old. When you have the time, please propose a working set of changes on the most recent version of `master` with a new PR :)

polarathene:
This was intended to be investigated once the test suite was fully migrated to the new structure. LDAP did not get to that point for v12.

This has not been an issue raised by many other LDAP users AFAIK, there has been little involvement since the issue / PR was opened. If someone were to be faced by this problem and wants to resolve it, my advice is to do the following:

- Remove the `chown -R`
- Add that removed logic into a `setup ...` command with troubleshooting documentation.
- IIRC there was a scenario where `chown -R` here was necessary (_at least for non-LDAP_). It may have been related to a feature or initial creation of mailboxes. DMS upgrades should be fine as the UID + GID for `docker` is hard-coded to 5000.

Seeing as it doesn't affect all LDAP users, I assume it's config specific or affected by choice of LDAP provider.

MohammedNoureldin:
Hi, thanks for notifying me @georglauterbach, I will check it again for sure once I have more time again.

@polarathene it is kind of configuration-specific, but not fully. If the administrator is OK to get all emails owned by the user 5000, then the issue will not appear. On the other hand, if every single LDAP user (with unique ID) should own only his own emails on the Linux system (`chown`) then the issue will appear.

So it will not appear in the less-secure option I would say, but it should be definitely be checked somewhen.

AntiBlueQuirk:
Specifically, I think this is caused by including `uid` and `gid` in `DOVECOT_USER_ATTRS`. I'm a little fuzzy on the details, but Dovecot actually tries to switch the process to the uid and gid of a user when accessing mail. Normally, DMS gets around this by just setting all users to use the same uid and gid: `5000:5000`. When using a manual user database, you can see this is set here in [accounts.sh](https://github.com/docker-mailserver/docker-mailserver/blob/b001f5a14056fa4df5bf50466247f3f4e4d8000e/target/scripts/helpers/accounts.sh#L73). I'm guessing similar permission problems are caused if you change the uid and gid of users in the manual user database.

When using LDAP, you can tell Dovecot to get the uid:gid from the database, which it will then switch to when accessing mail. Since the directories are all chowned to 5000:5000, this causes Dovecot to not actually be able to access the user's mail. You can manually correct the permissions, which makes Dovecot work correctly, but I was still having delivery problems.

I was using `uidNumber=uid,gidNumber=gid` for `DOVECOT_USER_ATTRS`, but right now I've replaced it with `=uid=5000,=gid=5000`, which makes all LDAP users use 5000:5000. It seems to be working now. I would prefer to keep the proper user permissions, but I like having working mail more.

polarathene:
> I would prefer to keep the proper user permissions, but I like having working mail more.

The discussion above had attempted to only apply the `chown -R 5000:5000` for non-LDAP config IIRC (_or remove it completely, I think both were tried_), but this [caused test failures](https://github.com/docker-mailserver/docker-mailserver/pull/2256#issuecomment-947240208). So while we're happy to allow such logic for a better LDAP experience, someone has to spend time looking into why tests were failing.

We apply this `chown` at container startup, but also with the change-detection service. Presently that is only valid for non-LDAP config but there is a desire to enable change-detector (at least partially) for LDAP users as well.

A lot has happened with the project codebase since this PR was attempted, quite a bit of refactoring in our test suite and code related to this PR/issue.

PR #2260, opened 2021-10-25
fix: `check-for-changes.sh` should not fall out of sync with shared logic

polarathene:
# Description

Removes duplicate logic from `check-for-changes.sh` that is used/maintained elsewhere to avoid risk of problems, as this code is already starting to diverge / rot.

---

Previously the change detection support has had code added for rebuilding config upon change detection which is the same as code run during startup scripts. Unfortunately over time this has fallen out of sync. Mostly the startup scripts would get maintenance and the contributor and reviewers may not have been aware of the duplicate code handled by `check-for-changes.sh`.

That code was starting to diverge in addition to some changes in structure (_eg: relay host logic seems interleaved here vs separated out in startup scripts_). I wanted to address this before it risks becoming a much bigger headache.

Rather than bloat `helper-functions.sh` further, I've added a `helpers/` folder extracting relevant common logic between startup scripts and `changedetector`. If you want to follow that process I've kept scoped commits to make those diffs easier. Some minor changes/improvements were added but nothing significant.

I've not given too much thought to naming conventions at present, that should still be fairly flexible. A similar PR will be handled for TLS functionality ~~and additional changes to `check-for-changes.sh` once https://github.com/docker-mailserver/docker-mailserver/pull/2245 is merged. I want to minimize any merge conflicts due to related PRs open.~~

~~`accounts.sh` contains logic that https://github.com/docker-mailserver/docker-mailserver/pull/2248 is updating, so I've reverted my `setup-stack.sh` changes for that affected method to avoid merge conflicts. Once it is merged, a merge commit into this PR should go smoothly and I'll sync changes to `accounts.sh`, followed by bringing back the minimal `setup-stack.sh:_setup_dovecot_local_user` method this PR introduces.~~

---

Ran into a little confusion with `shellcheck` lint, that's been resolved and some extra maintainer comment docs in the `lint.sh` for `shellcheck` to clarify it to others. Probably will migrate to maintainer docs at a later stage.

A recent PR `shellcheck` linting behaviour/config came up and regarding `SCRIPTDIR` was also [identified here](https://github.com/docker-mailserver/docker-mailserver/pull/2281#issuecomment-956631050).

## Type of change

## Checklist:


target/scripts/helpers/index.sh
```diff
@@ -0,0 +1,15 @@
+#! /bin/bash
+# shellcheck source-path=target/scripts/helpers
```
georglauterbach, line 2: 👍🏼

georglauterbach (dismissed):
If I see and understand all of this correctly, there is no "new" functionality in this PR, but this PR cleans up the whole `check-for-changes.sh` mess plus additional cleanup?

Exceptionally well done, LGTM!

polarathene:
> If I see and understand all of this correctly, there is no "new" functionality in this PR

Correct :+1: 

> but this PR cleans up the whole `check-for-changes.sh` mess plus additional cleanup?

Yeah, an initial pass. It still causes a large diff so I didn't want to introduce any broader changes, those can be addreessed in a follow-up PR at a later point :)

I made sure to use commit history for easier to follow transition of changes if helpful.

casperklein:
Sorry for the delay. Sometimes I am bit scared when I see large PR/changes 😄. I'll try to review it within the next week.

polarathene:
> Sometimes I am bit scared when I see large PR/changes :smile: 

It's really not as large as it looks in the diff. The commit history is intended to help provide you with that confidence as I was careful to break down the transition for better review by diffs.

Even without that, you can bring up your own diff app (eg: meld) and take the `setup-stack.sh` deleted sections to find they're effectively 1:1 copied to functions in new files. Only minor differences might be splitting into multiple functions and some conditionals.

`check-for-changes.sh` roughly is the same I believe I kept the order intact and just call the extracted functions. I used a diff against `setup-stack.sh` initially while extracting the common logic out to the new helpers.

> I'll try to review it within the next week.

No worries, take your time reviewing it, no rush :+1:

georglauterbach (approved):

target/scripts/helper-functions.sh
```diff
@@ -349,6 +316,12 @@ function _obtain_hostname_and_domainname
   DOMAINNAME="${DOMAINNAME:-${HOSTNAME}}"
 }
 
+# Remove string input with only whitespace or `#` as the first non-whitespace character.
```
casperklein, line 319:
```suggestion
# Remove string input with empty line, only whitespace or `#` as the first non-whitespace character.
```

target/scripts/helpers/relay.sh
```diff
@@ -0,0 +1,136 @@
+#! /bin/bash
+# Support for Relay Hosts
+
+function _relayhost_default_port_fallback
+{
+  [[ -z ${RELAY_PORT} ]] && RELAY_PORT=25
```
casperklein, line 6: That's, how we set defaults in other places:

```suggestion
  : ${RELAY_PORT:=25}
```

```suggestion
  RELAY_PORT=${RELAY_PORT:-25}
```

target/scripts/helpers/sasl.sh
```diff
@@ -0,0 +1,23 @@
...
+  then
+    # create SASL password
+    echo "${SASL_PASSWD}" > /etc/postfix/sasl_passwd
+    _sasl_passwd_chown_chmod
+  else
+    rm -f /etc/postfix/sasl_passwd
+  fi
+}
+
+function _sasl_passwd_chown_chmod
```
casperklein, line 16: Apply or discard:

```suggestion
function _sasl_set_passwd_permissions
```

target/scripts/helpers/relay.sh
```diff
@@ -0,0 +1,136 @@
...
+    "smtp_tls_security_level = encrypt" \
+    "smtp_tls_note_starttls_offer = yes" \
+    "smtp_tls_CAfile = /etc/ssl/certs/ca-certificates.crt" \
+    "sender_dependent_relayhost_maps = texthash:/etc/postfix/relayhost_map" \
+    "smtp_sender_dependent_authentication = yes"
+}
+
+# ? --------------------------------------------- Callers
+
+# setip-stach.sh:
```
casperklein, line 99:
```suggestion
# setup-stach.sh:
```

target/scripts/helpers/postfix.sh
```diff
@@ -0,0 +1,52 @@
...
+# texthash type does not require running `postmap` after changes are made, other types might.
+#
+# Examples of different types actively used:
+# setup-stack.sh:_setup_spoof_protection uses texthash + hash + pcre, and conditionally unionmap
+# main.cf:
+# - alias_maps and alias_database both use hash:/etc/aliases
+# - virtual_mailbox_maps and virtual_alias_maps use texthash
+# - `alias.sh` may append pcre:/etc/postfix/regexp to virtual_alias_maps in `main.cf`
+#
+# /etc/aliases is handled by `alias.sh` and uses `postalias` to update the Postfix alias database. No need for `postmap`
```
casperklein, line 51:
```suggestion
# /etc/aliases is handled by `alias.sh` and uses `postalias` to update the Postfix alias database. No need for `postmap`.
```

casperklein (changes requested):

casperklein:
One other thing to consider is, to conditionally source some helpers, e.g.:

source SASL stuff only, if SASL is used

target/scripts/helpers/relay.sh
polarathene: This was the original line already in use, I started making some minor changes with relay logic which was the first to extract, but then realized it was better at least for review to minimize introducing changes :sweat_smile: 

I can apply this though no worries :+1:

target/scripts/helpers/sasl.sh
polarathene: I'll apply this change later today, either via two commits in web UI or when I can free enough RAM on my local system (_because of laziness avoiding CLI for git_ :sweat_smile: )

polarathene:
> One other thing to consider is, to conditionally source some helpers, e.g.:
> 
> source SASL stuff only, if SASL is used

Are you referring to the `check-for-changes.sh` part?

That is something I'd like to do, but deferred from this PR to ease review since I wanted it to be reasonably clear in the diff that the bulk of it is just shifting the logic out to new files, especially with `check-for-changes.sh` having differed a bit in order and some lines.

casperklein:
I was referring to `helpers/index.sh`. It was just a (future) suggestion, no need to change/improve it now.

target/scripts/helpers/sasl.sh
casperklein: Don't forget to update also the corresponding function calls 👍

polarathene:
> I was referring to `helpers/index.sh`. It was just a (future) suggestion, no need to change/improve it now.

Ah, well that's already handled either within the function or via the caller IIRC. The SASL one has two functions, one that checks the ENV and another that checks for a file. It seems that those conditions are best suited where they are and that conditionally sourcing the functions/file would be a bit more messier and prone to timing? (especially at setup)

casperklein:
> prone to timing? 

A bit. But it will also save memory 😉 In that particular case, I admit, it's not worth 👍 
It was more a general thinking: Source only those files/function that are needed.

polarathene:
> Edit: I misunderstood the timing thing. There should be no race condition or what did you mean?

- `aliases.sh:_handle_postfix_aliases_config` checks if `/tmp/docker-mailserver/postfix-aliases.cf` exists, otherwise creates it, not really anything to conditionally source there, but if anything else expected `/tmp/docker-mailserver/postfix-aliases.cf` when invoked later on, and we didn't originally source it because the file didn't exist, that'd fail silently? (_a quick grep through the project for `postfix-aliases.cf` suggests this currently isn't used for anything, so it's unclear why we create this as an empty file if it doesn't exist, other than possibly because it's expected by the file monitor?_)
- `postfix.sh:_create_postfix_vhost` likewise relies on `/tmp/vhost.tmp` which is populated at runtime by `setup-stack.sh:_setup_ldap`, `accounts.sh:_create_accounts` and `aliases.sh:_handle_postfix_virtual_config`.
- `relay.sh:_populate_relayhost_map` checks for `postfix-accounts.cf` and `postfix-virtual.cf` files, non-issue if you don't have relay enabled, and the two methods in `relay.sh`: `_setup_relayhost` (`setup-stack.sh`) and `_rebuild_relayhost` (`check-for-changes.sh`) handle the conditional checks for what to run.. you could lift that up into `index.sh` to only source if `RELAY_HOST` exists, I didn't investigate enough to check why `DEFAULT_RELAY_HOST` was handled separately though which may defeat the benefit of sourcing that way.
- Additionally with `relay.sh` there's a SASL section, if that's conditionally sourced, it needs to add logic to conditionally not run that method when `sasl.sh` is not sourced.

Timing wise, I was thinking more about the time that `helper-functions.sh` is sourced and these conditional sourcing would apply, vs changes that occur after that point onwards. Besides the concerns above that need to be accounted for, `check-for-changes.sh` might later have a config file created/added but if we sourced conditionally based on that, the related scripts wouldn't have been sourced.. perhaps you only had ENV in mind for conditionally sourcing?

Just seems like more maintenance burden for minimal benefit vs just sourcing it all and guarding with conditionals within the respective files like we presently have. There is some conditionals external still, such as `SMTP_ONLY` IIRC.

---

Personally I prefer the requirements to run the logic to be clearly defined within these files themselves, so conditionally sourcing at `index.sh` will for the most part duplicate that, or enforce a hard requirement of sourcing only via `index.sh` (_or rather implicitly via `helper-functions.sh`_).

After this is merged, I'd like to move some of the `helper-functions.sh` out to separate files too, and noticed that the new files in this PR rely on functions from `helper-functions.sh` already and dependency management via sourcing in each file would be messy in bash :sweat_smile:

So at best I might leave a comment to reference them at the top, but rely on importing via `index.sh` anyway, so this will probably be the advised way to use any method regardless (_via sourcing `helper-functions.sh` as we currently do_), so lifting the conditions where appropriate is probably fine...but then complicates it a bit further for minimal benefit when calling the functions that may not be sourced (_eg: the relay ones in `setup-stack.sh` and `check-for-changes.sh`, it's more convenient to have the conditional execution wrapped in the associated `relay.sh` file and just source unconditionally_)

target/scripts/helpers/relay.sh
```diff
@@ -0,0 +1,136 @@
...
+  fi
+
+  # add default relay
+  if [[ -n ${RELAY_USER} ]] && [[ -n ${RELAY_PASSWORD} ]]
+  then
+    # white-space separates value pairs (any length is valid)
+    echo "[${RELAY_HOST}]:${RELAY_PORT} ${RELAY_USER}:${RELAY_PASSWORD}" >> /etc/postfix/sasl_passwd
+  fi
+
+  _sasl_passwd_chown_chmod
```
polarathene, line 45:
```suggestion
  _sasl_set_passwd_permissions
```

target/scripts/helpers/sasl.sh
```diff
@@ -0,0 +1,23 @@
+#! /bin/bash
+# Support for SASL
+
+function _sasl_passwd_create
+{
+  if [[ -n ${SASL_PASSWD} ]]
+  then
+    # create SASL password
+    echo "${SASL_PASSWD}" > /etc/postfix/sasl_passwd
+    _sasl_passwd_chown_chmod
```
polarathene, line 10:
```suggestion
    _sasl_set_passwd_permissions
```

georglauterbach (approved):

casperklein (approved):

Issue #2263, opened 2021-10-28
Mailserver behind Traefik


dimitrimestdag:
# Subject

I would like to configure a not documented mail server use case

## Description

Hello,

I have configured your mail server with rainloop.
I receive emails well and when I configure an alias to an external address and I receive an email on my address, the email is transferred to the external address.
The concern is that when I try to send an email from rainloop I get the message "Authentication failed":
![image](https://user-images.githubusercontent.com/18188598/139282284-d528ab19-8ccf-40f7-9b7a-3358cd846b98.png)

Here are the container logs regarding this error : 

``` BASH
Oct 28 16:09:23 mail dovecot: imap(user@domain.fr)<18024><BXryQWrPUMtVqGs9>: Logged out in=90 out=982 deleted=0 expunged=0 trashed=0 hdr_count=0 hdr_bytes=0 body_count=0 body_bytes=0,
Oct 28 16:13:38 mail dovecot: imap(user@domain.fr)<19420><UynoUGrPAs9VqGs9>: Logged out in=90 out=982 deleted=0 expunged=0 trashed=0 hdr_count=0 hdr_bytes=0 body_count=0 body_bytes=0,
Oct 28 16:13:38 mail dovecot: imap-login: Login: user=<user@domain.fr>, method=PLAIN, rip=my.ip, lip=traefik ip, mpid=19420, TLS, session=<UynoUGrPAs9VqGs9>,
Oct 28 16:13:34 mail dovecot: imap(user@domain.fr)<19384><wkHLUGrP7s5VqGs9>: Logged out in=11 out=449 deleted=0 expunged=0 trashed=0 hdr_count=0 hdr_bytes=0 body_count=0 body_bytes=0,
Oct 28 16:13:34 mail postfix/smtps/smtpd[19385]: disconnect from my-ip.rev.numericable.fr[my.ip] ehlo=1 auth=0/1 quit=1 commands=2/3,
Oct 28 16:13:34 mail postfix/smtps/smtpd[19385]: warning: my-ip.rev.numericable.fr[my.ip]: SASL PLAIN authentication failed: ,
Oct 28 16:13:32 mail dovecot: auth: passwd-file(user,my.ip): unknown user (SHA1 of given password: ca7fbf),
[...]

PR #2264, opened 2021-10-29
fix(listmailuser): Don't query quota, if ENABLE_QUOTAS is not 1

casperklein:
# Description

Fixes https://github.com/docker-mailserver/docker-mailserver/issues/2263#issuecomment-954701535 and introduces a new file `/etc/dms-settings`, which contains all DMS specific settings.

## Type of change

## Checklist:

### DIFF
--- a/target/bin/listmailuser
+++ b/target/bin/listmailuser
@@ -2,6 +2,7 @@
 
 # shellcheck source=../scripts/helper-functions.sh
 . /usr/local/bin/helper-functions.sh
+. /etc/dms-settings
 
 function dovecot_quota_to_hr()
 {
@@ -15,18 +16,6 @@ function dovecot_quota_to_hr()
 
 DATABASE="/tmp/docker-mailserver/postfix-accounts.cf"
 ALIASES="/tmp/docker-mailserver/postfix-virtual.cf"
-
-if [[ -z ${ENABLE_QUOTAS} ]]
-then
-  # container was started without the env ENABLE_QUOTAS set.
-  # get default value from .bashrc
-  if grep -qE "^export ENABLE_QUOTAS='0'$" /root/.bashrc
-  then
-    ENABLE_QUOTAS=0
-  else
-    ENABLE_QUOTAS=1
-  fi
-fi
 
 [[ -f ${DATABASE} ]] || errex "Error: No postfix-accounts.cf file"
 [[ -s ${DATABASE} ]] || errex "Error: Empty postfix-accounts.cf - no accounts have been added"
--- a/target/bin/listmailuser
+++ b/target/bin/listmailuser
@@ -24,15 +24,15 @@ while read -r LINE
 do
   USER=$(echo "${LINE}" | cut -d'|' -f1)
 
-  if [[ "${ENABLE_QUOTAS}" -eq 0 ]]
+  if [[ "${ENABLE_QUOTAS}" -eq 1 ]]
   then
-    echo "* ${USER}"
-  else
     # ${QUOTA[0]} => current size
     # ${QUOTA[1]} => configured size limit
     # ${QUOTA[2]} => usage in percent
     IFS=' ' read -r -a QUOTA <<< "$(doveadm quota get -u "${USER}" | tail +2 | awk '{ if ($3 == "STORAGE") { print $4" "$5" "$6 } }')"
     echo "* ${USER} ( $(dovecot_quota_to_hr "${QUOTA[0]}") / $(dovecot_quota_to_hr "${QUOTA[1]}") ) [${QUOTA[2]}%]"
+  else
+    echo "* ${USER}"
   fi
 
   if [[ -f ${ALIASES} ]] && grep -q "${USER}" "${ALIASES}"
--- a/target/bin/listmailuser
+++ b/target/bin/listmailuser
@@ -23,12 +23,16 @@ while read -r LINE
 do
   USER=$(echo "${LINE}" | cut -d'|' -f1)
 
-  # ${QUOTA[0]} => current size
-  # ${QUOTA[1]} => configured size limit
-  # ${QUOTA[2]} => usage in percent
-  IFS=' ' read -r -a QUOTA <<< "$(doveadm quota get -u "${USER}" | tail +2 | awk '{ if ($3 == "STORAGE") { print $4" "$5" "$6 } }')"
-
-  echo "* ${USER} ( $(dovecot_quota_to_hr "${QUOTA[0]}") / $(dovecot_quota_to_hr "${QUOTA[1]}") ) [${QUOTA[2]}%]"
+  if [[ "${ENABLE_QUOTAS}" -eq 0 ]]
+  then
+    echo "* ${USER}"
+  else
+    # ${QUOTA[0]} => current size
+    # ${QUOTA[1]} => configured size limit
+    # ${QUOTA[2]} => usage in percent
+    IFS=' ' read -r -a QUOTA <<< "$(doveadm quota get -u "${USER}" | tail +2 | awk '{ if ($3 == "STORAGE") { print $4" "$5" "$6 } }')"
+    echo "* ${USER} ( $(dovecot_quota_to_hr "${QUOTA[0]}") / $(dovecot_quota_to_hr "${QUOTA[1]}") ) [${QUOTA[2]}%]"
+  fi
 
   if [[ -f ${ALIASES} ]] && grep -q "${USER}" "${ALIASES}"
   then
--- a/target/bin/listmailuser
+++ b/target/bin/listmailuser
@@ -16,6 +16,18 @@ function dovecot_quota_to_hr()
 DATABASE="/tmp/docker-mailserver/postfix-accounts.cf"
 ALIASES="/tmp/docker-mailserver/postfix-virtual.cf"
 
+if [[ -z ${ENABLE_QUOTAS} ]]
+then
+  # container was started without the env ENABLE_QUOTAS set.
+  # get default value from .bashrc
+  if grep -qE "^export ENABLE_QUOTAS='0'$" /root/.bashrc
+  then
+    ENABLE_QUOTAS=0
+  else
+    ENABLE_QUOTAS=1
+  fi
+fi
+
 [[ -f ${DATABASE} ]] || errex "Error: No postfix-accounts.cf file"
 [[ -s ${DATABASE} ]] || errex "Error: Empty postfix-accounts.cf - no accounts have been added"
 
--- a/target/bin/listmailuser
+++ b/target/bin/listmailuser
@@ -2,7 +2,9 @@
 
 # shellcheck source=../scripts/helper-functions.sh
 . /usr/local/bin/helper-functions.sh
-. /etc/dms-settings
+
+# suppress error output, e.g. when listmailuser runs in a fresh container (DMS not running)
+. /etc/dms-settings 2> /dev/null
 
 function dovecot_quota_to_hr()
 {
--- a/target/scripts/startup/setup-stack.sh
+++ b/target/scripts/startup/setup-stack.sh
@@ -75,8 +75,8 @@ function _setup_default_vars
   VARS[REPORT_RECIPIENT]="${REPORT_RECIPIENT}"
   VARS[REPORT_SENDER]="${REPORT_SENDER}"
 
-  : >/root/.bashrc
-  : >/etc/dms-settings
+  : >/root/.bashrc     # make DMS variables available in login shells and their subprocesses
+  : >/etc/dms-settings # this file can be sourced by other scripts
   local VAR
   for VAR in "${!VARS[@]}"
   do
--- a/target/scripts/startup/setup-stack.sh
+++ b/target/scripts/startup/setup-stack.sh
@@ -334,6 +334,7 @@ function _setup_dovecot_local_user
 
     # creating users ; 'pass' is encrypted
     # comments and empty lines are ignored
+    local LOGIN PASS USER_ATTRIBUTES
     while IFS=$'|' read -r LOGIN PASS USER_ATTRIBUTES
     do
       # Setting variables for better readability
@@ -344,24 +345,31 @@ function _setup_dovecot_local_user
       if [[ -f /tmp/docker-mailserver/dovecot-quotas.cf ]]
       then
         declare -a USER_QUOTA
-        IFS=':' ; read -r -a USER_QUOTA < <(grep "${USER}@${DOMAIN}:" -i /tmp/docker-mailserver/dovecot-quotas.cf)
-        unset IFS
+        IFS=':' read -r -a USER_QUOTA < <(grep "${USER}@${DOMAIN}:" -i /tmp/docker-mailserver/dovecot-quotas.cf)
 
-        [[ ${#USER_QUOTA[@]} -eq 2 ]] && USER_ATTRIBUTES="${USER_ATTRIBUTES} userdb_quota_rule=*:bytes=${USER_QUOTA[1]}"
+        if [[ ${#USER_QUOTA[@]} -eq 2 ]]
+        then
+          USER_ATTRIBUTES="${USER_ATTRIBUTES:+${USER_ATTRIBUTES} }userdb_quota_rule=*:bytes=${USER_QUOTA[1]}"
+        fi
       fi
 
-      # Let's go!
-      _notify 'inf' "user '${USER}' for domain '${DOMAIN}' with password '********', attr=${USER_ATTRIBUTES}"
+      if [[ -z ${USER_ATTRIBUTES} ]]
+      then
+        _notify 'inf' "Creating user '${USER}' for domain '${DOMAIN}'"
+      else
+        _notify 'inf' "Creating user '${USER}' for domain '${DOMAIN}' with attributes '${USER_ATTRIBUTES}'"
+      fi
 
       echo "${LOGIN} ${DOMAIN}/${USER}/" >> /etc/postfix/vmailbox
-      # User database for dovecot has the following format:
+      # Dovecot's userdb has the following format
       # user:password:uid:gid:(gecos):home:(shell):extra_fields
-      # Example :
-      # ${LOGIN}:${PASS}:5000:5000::/var/mail/${DOMAIN}/${USER}::userdb_mail=maildir:/var/mail/${DOMAIN}/${USER}
-      echo "${LOGIN}:${PASS}:5000:5000::/var/mail/${DOMAIN}/${USER}::${USER_ATTRIBUTES}" >> /etc/dovecot/userdb
+      echo \
+        "${LOGIN}:${PASS}:5000:5000::/var/mail/${DOMAIN}/${USER}::${USER_ATTRIBUTES}" \
+        >>/etc/dovecot/userdb
+
       mkdir -p "/var/mail/${DOMAIN}/${USER}"
 
-      # Copy user provided sieve file, if present
+      # copy user provided sieve file, if present
       if [[ -e "/tmp/docker-mailserver/${LOGIN}.dovecot.sieve" ]]
       then
         cp "/tmp/docker-mailserver/${LOGIN}.dovecot.sieve" "/var/mail/${DOMAIN}/${USER}/.dovecot.sieve"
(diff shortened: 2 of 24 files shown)


polarathene (changes requested):
Any logic that's trying to apply defaults that should have been applied already is a bit of a red flag, especially for maintenance.

For now, maintain this as a separate file (eg: `/target/scripts/init-default-env.sh`) and source that. `helper-functions.sh` is commonly sourced by utilities (_and pretty much everything else_), so is probably an appropriate place to import this workaround.

We should be able to avoid the issue entirely via an entrypoint script in future.

target/bin/listmailuser
```diff
@@ -16,19 +16,35 @@ function dovecot_quota_to_hr()
...
+  # container was started without the env ENABLE_QUOTAS set.
+  # get default value from .bashrc
+  if grep -qE "^export ENABLE_QUOTAS='0'$" /root/.bashrc
+  then
+    ENABLE_QUOTAS=0
+  else
+    ENABLE_QUOTAS=1
+  fi
+fi
+
```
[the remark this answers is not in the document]
casperklein: > or better yet via an entrypoint script that will ensure we apply ENV defaults upfront?

😄 As of now, there is no entrypoint script. I wanted to quickly fix an error and not introducing a complete new bigger feature.

> Any subshell or subprocess invoked afterwards should inherit that, unless invoked via a non-interactive shell (-c)?

No. It's the other way around. Only when starting bash as an interactive shell, `.bashrc` is automatically sourced.

We could just source that file manually in this case for example. But in general, `.bashrc` can contain other things like functions, aliases and normal commands. That's why I tried to avoid sourcing it. In my setup for example, there are some handy aliases and other things in that file.

> For now, maintain this as a separate file (eg: /target/scripts/init-default-env.sh) and source that

I tried to avoid introducing another file. But if everyone is fine with that, I can implement it. That would probably the cleanest/best solution 👍 

> Any defaults/adjustments that the mailserver service makes and stores to .bashrc won't be visible to all other services?

Correct.

> An entrypoint script will run when starting a container with a command such as setup, which would not run all of our initialization startup scripts (including the ones that create .bashrc here). Presently our entrypoint is just dumb-init.

Using an entrypoint script is not a good idea, especially when running multiple services in one container as we do. An (dumb) init system [takes care of a few things](https://github.com/Yelp/dumb-init#why-you-need-an-init-system).

BTW: I wanted to link to the dumb-init PR where I thought there were already discussions about pro/cons. I found the [commit](https://github.com/docker-mailserver/docker-mailserver/commit/6e9f6152becf10938c41934014fd877768d871c2), but not the PR?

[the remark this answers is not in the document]
polarathene: > I wanted to quickly fix an error and not introducing a complete new bigger feature.

I understand, that's why I suggested extracting it to a centralized file so this sort of thing doesn't pop up elsewhere in future.

> No. It's the other way around. Only when starting bash as an interactive shell, `.bashrc` is automatically sourced.

That's what I meant :sweat_smile: 

`-c` is non-interactive shell, and doesn't source `.bashrc`.

> I tried to avoid introducing another file. But if everyone is fine with that, I can implement it.

I think that would be best, feel free to see how others weigh in :)

> Using an entrypoint script is not a good idea, especially when running multiple services in one container as we do.

Isn't it effectively what `start-mailserver.sh` / `mailserver` (supervisor service) is?

We call the `dumb-init` that takes the `supervisord` CMD and that begins the `mailserver` service. Only difference is that if you use `docker run ... setup ...` instead of `docker exec` on a running container, the CMD will override `supervisord` instead of doing any preprocessing via ENTRYPOINT (_which can optionally start the mailserver service along with dumb-init still_).

> An (dumb) init system takes care of a few things.

For Docker at least, [an alternative `tini` is used by default](https://docs.docker.com/engine/reference/run/#specify-an-init-process) if using `--init`/`init:` ([docker docs init page](https://docs.docker.com/config/containers/multi-service_container/)), but for `docker-mailserver` I guess it makes more sense to bake it into the Dockerfile as part of the entrypoint :sweat_smile: .

> BTW: I wanted to link to the dumb-init PR where I thought there were already discussions about pro/cons. I found the commit, but not the PR?

That looks like roughly the time the project transferred to an org, where PR history seemed to be a bit iffy and @georglauterbach just carried over some changes manually I think. I don't see the PR looking back around Jan 2021, looks like it might have just been lost and the change just directly committed to master (_I don't think we had any branch protection back then), issues and PR references weren't accurate as a result.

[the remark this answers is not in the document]
casperklein: > Isn't it effectively what start-mailserver.sh / mailserver (supervisor service) is?

No. Supervisor is not an init system. See my link above for the differences and what an init process additionally does compared to any script.

> For Docker at least, an alternative tini is used by default if using --init/init

There isn't much difference between those two solutions in what they do. However, `--init` depends on docker. While our solution is build-in and works everywhere. Only "downside" is a few more used bytes in our image 😄

As you and Georg are fine with a new file, I am going to implement it 🎉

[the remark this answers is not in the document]
polarathene: > No. Supervisor is not an init system. See my link above for the differences and what an init process additionally does compared to any script.

Misunderstanding. I was referring to how this differs:

a. `ENTRYPOINT=dumb-init` + `CMD=supervisord` (or any override for CMD)
b. `ENTRYPOINT=entrypoint.sh` (that calls `dumb-init` with CMD) + `CMD=supervisord` (or any override for CMD)

Both still use `dumb-init`, one just allows running some conditional logic with the initial shell that runs the entrypoint (_that currently is equivalent to an `entrypoint.sh` that just calls `dumb-init` with `$@` to pass the CMD to_).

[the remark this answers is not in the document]
casperklein: Entrypoint = PID1. If entrypoint is a bash script, it will not take care of signal forwarding, reaping processes etc. An init-system must be PID1.

But I think we can skip this discussion. If a script like `listmailuser` needs ENVs, it can just source `/etc/dms-settings` now 😎

target/scripts/startup/setup-stack.sh
```diff
@@ -67,20 +67,20 @@ function _setup_default_vars
...
+  VARS[POSTMASTER_ADDRESS]="${POSTMASTER_ADDRESS}"
+  VARS[REPORT_RECIPIENT]="${REPORT_RECIPIENT}"
+  VARS[REPORT_SENDER]="${REPORT_SENDER}"
+
   local VAR
   for VAR in "${!VARS[@]}"
   do
-    echo "export ${VAR}='${VARS[${VAR}]}'" >>/root/.bashrc
+    echo "export ${VAR}='${VARS[${VAR}]}'" >/root/.bashrc
+    echo "${VAR}='${VARS[${VAR}]}'"        >/etc/dms-settings
```
georglauterbach, line 82: Should be

``` suggestion
    echo "export ${VAR}='${VARS[${VAR}]}'" >>/root/.bashrc
    echo "${VAR}='${VARS[${VAR}]}'"        >>/etc/dms-settings
```

Maybe then even clear `/etc/dms-settings` before the loop runs? (`: >/etc/dms-settings`)

georglauterbach (commented):
LGTM, just the one change I pointed to with the comment above. Not sure whether we want to clear the `/etc/dms-settings` files beforehand - that's why I did not directly commit the changes :D

target/scripts/startup/setup-stack.sh
casperklein: In a fresh container, `.bashrc` does only contain some comments.

I chose to overwrite the file, to avoid possible duplicate entrys, when for example, the container is restarted.

casperklein: Holy shit. That makes no sense. You are absolutly right.

casperklein:
> Not sure whether we want to clear the `/etc/dms-settings` files beforehand - that's why I did not directly commit the changes :D

I think clearing the file is a good practice to avoid duplicate entrys, when the container is restarted.

georglauterbach (approved):
LGTM 👍🏼

polarathene:
> Entrypoint = PID1. If entrypoint is a bash script, it will not take care of signal forwarding, reaping processes etc. An init-system must be PID1.

```diff
-ENTRYPOINT ["/usr/bin/dumb-init", "--"]
+ENTRYPOINT ["/usr/bin/dumb-init", "/entrypoint.sh"]

CMD ["supervisord", "-c", "/etc/supervisor/supervisord.conf"]
```

Supposedly that would work fine, but allow conditional logic based on the CMD used?

> But I think we can skip this discussion. If a script like `listmailuser` needs ENVs, it can just source `/etc/dms-settings` now :sunglasses:

I only bring it up because of `docker run ... setup email list` would not know about `/etc/dms-settings` or `/root/.bashrc` as neither exist.

I guess if `docker run` is never used with a CMD override it's a non-issue and any other commands should be run via `docker exec` on an existing instance instead.

I have had interest with adding an entrypoint script in the past, but `start-mailserver.sh` being started via `supervisord` via the `mailserver` service is for the most part equivalent, and anything I would add via an entrypoint should be ok there instead? (_eg: there's been interest to support initializing with a dummy account instead of requiring separate external script/files to provide an account up front_)

---

I can definitely understand how always setting up ENV could be an issue (_eg: the `open-dkim` utility script has generic var existence checks like `DOMAINS` which aren't scoped to local vars/unset upfront_), but there's also stuff like this in our tests:

https://github.com/docker-mailserver/docker-mailserver/blob/fb72f3ad52769ad470cd1f7fc36f76b53853f3e8/test/tests.bats#L478-L482

That's just running the container and echoing the ENV that was set on it. The `echo` CMD overrides the default `supervisord` command since this is `docker run` and not `docker exec`, thus none of the startup scripts are run. I guess that's fine, as `VIRUSMAILS_DELETE_DELAY` is only:

- `Dockerfile`: ENV default to 7
- `start-mailserver.sh`: default to 7 (makes the Dockerfile ENV a bit redundant?)
- `setup-stack.sh`: writes it to `/etc/environment`

Likewise two tests that use `delmailuser` and `addmailuser` also call [themselves via `docker run`](https://github.com/docker-mailserver/docker-mailserver/blob/fb72f3ad52769ad470cd1f7fc36f76b53853f3e8/test/tests.bats#L669-L689), if they ever need to rely on something like `/etc/dms-settings` that won't exist.

Just a caveat to be aware of, doesn't seem like it would cause any problems for the project atm though :+1:

test/tests.bats
```diff
@@ -660,7 +661,13 @@ EOF
   assert_success
 }
 
-@test "checking accounts: listmailuser" {
+@test "checking accounts: listmailuser (quotas disabled)" {
+  run docker exec mail /bin/sh -c "ENABLE_QUOTAS=0 listmailuser | head -n 1"
+  assert_success
+  assert_output '* user1@localhost.localdomain'
+}
```
polarathene, line 668: I was expecting one of these to fail, and sure enough the tests show that happened :stuck_out_tongue: :

```
not ok 306 checking accounts: listmailuser (quotas disabled) in 964ms
#  in test file test/tests.bats, line 667)
#   `assert_output '* user1@localhost.localdomain'' failed
# 
# -- output differs --
# expected : * user1@localhost.localdomain
# actual   : * user1@localhost.localdomain ( 12K / ~ ) [0%]
# --
# 
```

<details>
<summary>Stuff you already know</summary>

You're calling `sh -c` so `/root/.bashrc` is not loaded as expected. You originally added this ENV here to provide an explicit value and had conditional logic in `listmailuser` to use a default if it wasn't set explicitly.

Now with the `/etc/dms-settings` approach, you're dumping a copy of the `/root/.bashrc` contents for the vars but without the export (_not sure why you've opted for this difference vs sourcing the `/etc/dms-settings` and using `export` there_).

This overrides the ENV you provide always now with whatever `/etc/dms-settings` has, which I think was your intention.

</details>

Just need to tweak the test?

target/scripts/startup/setup-stack.sh
```diff
@@ -67,20 +67,22 @@ function _setup_default_vars
...
+  VARS[REPORT_RECIPIENT]="${REPORT_RECIPIENT}"
+  VARS[REPORT_SENDER]="${REPORT_SENDER}"
+
+  : >/root/.bashrc
+  : >/etc/dms-settings
   local VAR
   for VAR in "${!VARS[@]}"
   do
     echo "export ${VAR}='${VARS[${VAR}]}'" >>/root/.bashrc
+    echo "${VAR}='${VARS[${VAR}]}'"        >>/etc/dms-settings
```
polarathene, line 84: I didn't see in the discussion what the point of dropping the `export` prefix was? Couldn't you just dump these to `/etc/dms-settings` as before instead of `/root/.bashrc`? Then have `/root/.bashrc` source `/etc/dms-settings`?

Or was there some reason to want to temporarily set/override ENV without making it available to subprocesses?

I'm fine with whatever you go with, you have a better understanding of this stuff than me :sweat_smile: It seems better than maintaining a bunch of conditional default assignments like the previous approach :+1: (_except for loss of compatibility with `docker run` or the `docker exec ... sh -c ...` overrride_)

polarathene (changes requested):
LGTM :+1: 

Just need to update the test to work with the new `/etc/dms-settings` approach :grinning:

target/scripts/startup/setup-stack.sh
casperklein: The reason is pretty simple: The file should only contain the current container settings and not any further commands/logic. So if a script needs them (afaik this is the first time), the file can just be sourced. There is no need for "export". If a script needs **our** variables exported to a subprocess, it can do it on its own.

test/tests.bats
casperklein: I am not sure right now, why the latest test fail, but the test before didn't. Will investigate later.

casperklein:
@georglauterbach 

Some time back, you wanted us to use `[[ expresssion ]]`, instead of `[ expression ]`, which was used until then. I thought, there were only benefits and no drawbacks. However, today I noticed something that did not work as expected for me:

```sh
     1  #!/bin/bash
     2
     3  if [ "$SOME_UNSET_VARIABLE" -eq 0 ]; then
     4          echo "foo"
     5  fi
     6
     7  if [[ "$SOME_UNSET_VARIABLE" -eq 0 ]]; then
     8          echo "bar"
     9  fi
```

```
./test.sh: line 3: [: : integer expression expected
bar
```

I stumpled upon this, because I wondered, why listmailuser didn't fail, when $ENABLE_QUOTA was unset (called via setup.sh and no container running).

target/scripts/startup/setup-stack.sh
polarathene: Just seemed more consistent in behaviour/expectation if we kept the `export` as if `.bashrc` isn't loaded (eg: because of `bash -c`), then we're still able to make the expected environment available to our scripts. And I say that in the sense that sometimes calling a command may have the environment already made available to it, while the other context is sourcing it without the `export`.

I think we're more likely to run into troubleshooting issues from that inconsistency more than we would be from issues of having the env exported when it's one of the expected contexts and the whole point of this PR is a workaround for when that hasn't happened to ensure defaults are correctly fallen back to.

---

Not my forte here, so if your more confident in avoiding `export` as the better approach go for it :sunglasses:

test/tests.bats
polarathene: > why the latest test fail, but the test before didn't.

Expand the "Stuff you already know" section in my previous comment.

I thought that was pretty explanatory for why it failed now and didn't before.

polarathene:
> However, today I noticed something that did not work as expected for me

That works how I'd expect tbh.

The value is unset/null/false, so I expect the it to evaluate to equivalent of 0 (falsey). Some programming languages at least behave that way. Although `${SOME_UNSET_VARIABLE} == 0` will evaluate to false.. :man_shrugging:

georglauterbach:
I guess this is some kind of Bashism, but it also kind of makes sense to me. What you brought up a `sh` specific error, and bash evaluates variables whose values are null to falsy and I guess therefore equal to zero. I wouldn't see this as a drawback with Bash to be honest, because the type system (haha, as if Bash had a real type system) is somewhat weird with shells in general.

If this was a programming language like C or Rust or something else with a strict typing discipline, I'd consider this a drawback too, but with Bash, I've learned to just accept it :D

test/tests.bats
casperklein: Thanks. I was in hurry this morning and didn't see your explanation. ATM I've no idea how to fix the test in an elegant way. Do you?
Otherwise, I think I'll remove the test.

polarathene: It wouldn't be an issue with the original approach I thought was going to be taken, with the conditional check handling or via an entrypoint :stuck_out_tongue: 

- You could probably just copy/paste [this small test file](https://github.com/docker-mailserver/docker-mailserver/blob/master/test/mail_with_mdbox.bats) and adjust it for the two quota tests. Each can just spin up their own container.
- You can comment the failing one out but keep it there with a TODO to fix it instead of outright removing it.
- You can sed the `/etc/dms-settings` file to set the ENV before calling the utility script.

That last suggestion is probably the way to go? I know the current approach won't work with `docker run ... setup ...` but that's not really an issue, users should run additional commands via `docker exec` instead.

target/bin/listmailuser
```diff
@@ -23,12 +24,16 @@ while read -r LINE
 do
   USER=$(echo "${LINE}" | cut -d'|' -f1)
 
-  # ${QUOTA[0]} => current size
-  # ${QUOTA[1]} => configured size limit
-  # ${QUOTA[2]} => usage in percent
-  IFS=' ' read -r -a QUOTA <<< "$(doveadm quota get -u "${USER}" | tail +2 | awk '{ if ($3 == "STORAGE") { print $4" "$5" "$6 } }')"
-
-  echo "* ${USER} ( $(dovecot_quota_to_hr "${QUOTA[0]}") / $(dovecot_quota_to_hr "${QUOTA[1]}") ) [${QUOTA[2]}%]"
+  if [[ "${ENABLE_QUOTAS}" -eq 1 ]]
```
polarathene, line 29:
```suggestion
  if [[ "${ENABLE_QUOTAS}" == 0 ]]
```

Would have worked instead of `-eq 0` btw. Your earlier finding with an expression failing because the value was not an integer to compare with `-eq` when ENV values are never really integers but strings that bash at some cases decides can qualify as integers if a number (with the unset case being coerced to NaN presumably).

I think it's good though with this change to check if it's enabled rather than disabled :+1:

target/scripts/startup/setup-stack.sh
```diff
@@ -67,20 +67,22 @@ function _setup_default_vars
...
+  VARS[LOGWATCH_RECIPIENT]="${LOGWATCH_RECIPIENT}"
+  VARS[PFLOGSUMM_RECIPIENT]="${PFLOGSUMM_RECIPIENT}"
+  VARS[PFLOGSUMM_SENDER]="${PFLOGSUMM_SENDER}"
+  VARS[PFLOGSUMM_TRIGGER]="${PFLOGSUMM_TRIGGER}"
+  VARS[POSTMASTER_ADDRESS]="${POSTMASTER_ADDRESS}"
+  VARS[REPORT_RECIPIENT]="${REPORT_RECIPIENT}"
+  VARS[REPORT_SENDER]="${REPORT_SENDER}"
+
+  : >/root/.bashrc
+  : >/etc/dms-settings
```
polarathene, line 79: Might be worth adding some inline documentation comment about why vars are being stored slightly differently in two different files.

`git blame` history should work technically, so up to you I'm just thinking forward with future maintainers that come across this, especially if any future refactorings shift code around (_eg: when `start-mailserver.sh` gutted out into multiple files and heavy revisions early this year, it makes tracing the origin of some things via `git blame` a bit more of a hassle_ :sweat_smile: ).

polarathene (approved):

test/tests.bats
casperklein: I think I fixed the test with [`b3c6af9` (#2264)](https://github.com/docker-mailserver/docker-mailserver/pull/2264/commits/b3c6af90f6b2369bfd8b12a0819f00577656f802). This is similar to your `sed` proposal.

There should be no problem when no container is running, as the source error is silent now.

target/scripts/startup/setup-stack.sh
casperklein: > as if .bashrc isn't loaded

For clarification: the DMS variable stuff in .bashrc is not used in any way. It's just there (I think), to quickly check what ENVs the running container uses. So when you start a bash session, you can easily do something like `echo $SOME_DMS_VAR`.
That said, I don't think the exports in .bashrc aren't needed for anything DMS rely on.

The only script that need to access a DMS var (ENABLE_QUOTA) is `listmailuser`. I don't think introducing a entrypoint script is worth it. If there is more demand/scripts in the future, we can still implement it.

target/bin/listmailuser
casperklein: > == 0 would have worked instead of -eq 0 btw?

Yes. But as we work with "integers" 0/1, a string comparison feels wrong.
I like the -eq 1 approach from a logical point of view. Only if ENABLE_QUOTA has the value of 1 and is therefore set, query the dovecot quota. In all other cases, hide the quota information.

target/scripts/startup/setup-stack.sh
casperklein: The whole .bashrc thing started with [this](https://github.com/docker-mailserver/docker-mailserver/commit/08dc28e3041550646cf8b45ccfccada2a0ffde3c) for the `VIRUSMAILS_DELETE_DELAY`.

casperklein: > start-mailserver.sh: default to 7 (makes the Dockerfile ENV a bit redundant? When the Dockerfile ENV was added in 2017 this var was removed, but then added back in 2018 to resolve an issue)

I think there is a room for improvements 😉 I am wondering, if any of the ENVs in Dockerfile nowadays are needed.

casperklein: >  Edit: The only exception is VIRUSMAILS_DELETE_DELAY.

Or maybe not 😆 You also wrote, that `VIRUSMAILS_DELETE_DELAY` is also in `/etc/environment`.

target/scripts/startup/setup-stack.sh
casperklein: [`f014754` (#2264)](https://github.com/docker-mailserver/docker-mailserver/pull/2264/commits/f014754300d31a199be1966d856b06d3146d4dd8)

polarathene (approved):

georglauterbach (approved):

PR #2266, opened 2021-10-30
ci(docs): Fail when build aborts from broken links

polarathene:
# Description

This is a change originally from @NorseGaud in [his PR commit here](https://github.com/docker-mailserver/docker-mailserver/pull/2245/commits/77036f4510f7d4ba010461fb67d408a429a7d552). To ease review of that PR I'm pulling out anything that doesn't need to be in it.

---

Using `set -ex` will exit the script as soon as a non-zero exit code is returned, such as when the docker image fails building the docs due to `build --strict` catching broken links. This also removes the need for `|| exit` when changing directory.

This seems fine for a small script, but AFAIK an alternative fix is just adding `|| exit` to the end of the `docker run` command too? There appears to be advice [against adopting `-e` carelessly](http://mywiki.wooledge.org/BashFAQ/105), while others [encourage `-e`](http://redsymbol.net/articles/unofficial-bash-strict-mode/). I know that several maintainers here have preference towards `set -e` so I've kept the original PR solution.

Additionally:

- `-x` is used to improve command visibility when reviewing the workflow log output.
- `--name` isn't necessary, but was part of the original PR.
- I've chosen not to include `-o pipefail`, only because no pipes are used in this script.

## Type of change

## Checklist:

### DIFF
--- a/.github/workflows/scripts/docs/build-docs.sh
+++ b/.github/workflows/scripts/docs/build-docs.sh
@@ -1,15 +1,17 @@
 #!/bin/bash
+set -ex
 
 # `--user` is required for build output file ownership to match the CI user,
 #  instead of the internal root user of the container.
 docker run \
   --rm \
   --user "$(id -u):$(id -g)" \
-  -v "${PWD}:/docs" \
+  --volume "${PWD}:/docs" \
+  --name "build-docs" \
   squidfunk/mkdocs-material:7.3.5 build --strict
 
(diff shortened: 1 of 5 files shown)


casperklein (approved):

casperklein (approved):

wernerfred (approved):

NorseGaud:
❤️

NorseGaud (approved):

PR #2275, opened 2021-10-31
tests(feat): Refactor `test_helper/common.bash` common_container methods

polarathene:
# Description

This has been extracted from https://github.com/docker-mailserver/docker-mailserver/pull/2245 as it does not need to be part of that PR.

---

These are improvements for better supporting the requirements of other tests.

- Opted for passing an array reference instead of an ENV file. This seems to be a better approach and supports more than just ENV changes.
- Likewise, shifted to a `create` + `start` approach, instead of `docker run` for added flexibility.
- Using `TEST_TMP_CONFIG` instead of `PRIVATE_CONFIG` to make the difference in usage with config volume in tests more clear.
- Changed the config volume from read-only volume mount to be read-write instead, which seems required for other tests.
- Added notes about logged failures from a read-only config volume during container startup.
- Added `TEST_CA_CERT` as a default CA cert path for the test files volume. This can be used by default by openssl methods.

## Type of change

## Checklist:


test/test_helper/common.bash
```diff
@@ -210,30 +210,52 @@ function wait_for_empty_mail_queue_in_container() {
...
     "${NAME}"
   assert_success
+}
+
+function common_container_start() {
+  run docker start "${TEST_NAME}"
+  assert_success
 
   wait_for_finished_setup_in_container "${TEST_NAME}"
 }
```
georglauterbach, line 261: No newline at the end of file. Interesting :)

test/test_helper/common.bash
```diff
@@ -210,30 +210,52 @@ function wait_for_empty_mail_queue_in_container() {
...
+  common_container_start
+}
 
-  run docker run -d --name "${TEST_NAME}" \
-    --volume "${TEST_FILES_VOLUME}" \
-    --volume "${TEST_CONFIG_VOLUME}" \
+# Common docker setup is centralized here,
+# can pass an array as a reference for adding extra config
+function common_container_create() {
+  local -n X_EXTRA_ARGS=${1}
```
georglauterbach, line 244: What does `local -n` do exactly? Never seen this before. Does ot create an array here (I assume because the variable is used as such later)?

test/mail_spam_bounced.bats
```diff
@@ -44,10 +48,13 @@ function setup_file() {
...
-  local TEST_ENV_FILE="${PRIVATE_CONFIG}/undefined.env"
-  echo 'ENABLE_SPAMASSASSIN=1' > "${TEST_ENV_FILE}"
 
-  common_container_setup "${TEST_ENV_FILE}"
+  # shellcheck disable=SC2034
+  local TEST_DOCKER_ARGS=(
+    --env ENABLE_SPAMASSASSIN=1
+  )
+
+  common_container_setup TEST_DOCKER_ARGS
```
georglauterbach, line 57: `TEST_DOCKER_ARGS` is not expanded here, but just a string. Indented? I mean, it's a local variable, so it cannot be used by other functions.

georglauterbach (commented):
Just one or two questions; not that good at bats myself TBH.

test/test_helper/common.bash
casperklein: `help local` gives this:

`      -n        make NAME a reference to the variable named by its value`

test/mail_spam_bounced.bats
polarathene: > I mean, it's a local variable, so it cannot be used by other functions.

`local` vars are scoped to the function bounds, they are visible to any function called by them IIRC, I can assure you this works :)

As an arg, it is passing just the string identifier, should I make that more clear by quote wrapping it? It's effectively a reference to the local array var.

test/test_helper/common.bash
polarathene: Yeah, I needed a way to pass the array over to the method. I see that I also export `TEST_DOCKER_ARGS` in a related `common.bash` function while doing a `local` declaration in the test.

There was probably a point to that, maybe with the updated `letsencrypt` test which required the updates made in this PR. :thinking:

test/test_helper/common.bash
polarathene: Would you like one?

casperklein: Each line (also the last one) in a text file, should end with a line-break.

See also our [.editorconfig](https://github.com/docker-mailserver/docker-mailserver/blob/c7dec1e33086bbaf8609b0db6bc8799da5d5bb5b/.editorconfig): `insert_final_newline = true`

test/mail_spam_bounced.bats
casperklein: Maybe I missunderstood what you are trying to achieve here, but this might what you are looking for:

```sh
function common_container_setup() {
	docker run \
	--option1  \
	--option2  \
	"$@" # additional options
}

OPTIONS=("--env" "ENABLE_SPAMASSASSIN=1")
common_container_setup "${OPTIONS[@]}"
```

polarathene: I believe I considered that at the time, but wanted to keep flexibility as this functionality is still being iterated on as I rewrite tests to adopt it.

Using the array reference I can still manipulate the array on either end IIRC, but I probably chose the approach to make it a single arg, should I need to change what inputs it accepts again (_switched away from an env filepath arg in this PR_).

Looking at my rewritten `letsencrypt` test I don't see anything requiring that, so I could switch to expanding the array for now if preferred?

test/test_helper/common.bash
georglauterbach: But it's good we're talking about it; as we can see, there's a bit of confusion about this:) There should be a short comment explaining what's going on :D

test/mail_spam_bounced.bats
georglauterbach: Using a real string with `'...'` here may be a good idea to make it explicit. And a bit of documentation as well, because some questions come up on what this code does. Don't get me wrong, seems like good code to me, just difficult for me to understand because I'm not fluent in BATS.

test/test_helper/common.bash
casperklein: I don't understand what exactly `local -n X_EXTRA_ARGS=${1}` try to solve her. Why is a named reference needed?

polarathene: > Why is a named reference needed?

`test => ( (method1 => method2) ||  method2 ) => X_EXTRA_ARGS[@]`

It just allows a single arg/var to expand where desired while allowing adding other args and not having to deal with that mess if just passing `$@` and changing the function to work with multiple inputs later?

I'd like to convert all the tests eventually to share the common config, but some like the other SSL tests have some more involved setups, and I'd like some flexibility in configuring a test matrix run (eg run the test suite with a different set of defaults like `example.com` vs `mail.example.com` without having to manually change / alter all tests individually.

So it's mostly for convenience while iterating on it.

polarathene:
- Quoted the array var name being passed to the arg to avoid confusion.
- Added more verbose documentation to the common helper methods. Hopefully not as confusing now :sweat_smile: 

I'd prefer to keep the current approach for now. Can change to `${TEST_DOCKER_ARGS[@]}` and `$@` at a later point if worthwhile.

georglauterbach (approved):
LGTM 👍

test/test_helper/common.bash
casperklein: I am still not getting it 100%. If you like, you can give a generic code example, maybe that enlightens me 😆 All what you wrote seem to me perfectly doable without named references. That's what confusing me right know. Also I don't understand how this actually works.

Here, an array with two elements is defined:

https://github.com/polarathene/docker-mailserver/blob/d8644fbb09bfaaa3dec03fc611646adcbd67e140/test/mail_spam_bounced.bats#L53


Here, function common_container_setup is called with a string argument 'TEST_DOCKER_ARGS'
https://github.com/polarathene/docker-mailserver/blob/d8644fbb09bfaaa3dec03fc611646adcbd67e140/test/mail_spam_bounced.bats#L57

common_container_setup passes the string 'TEST_DOCKER_ARGS' to the function common_container_create.
https://github.com/polarathene/docker-mailserver/blob/d8644fbb09bfaaa3dec03fc611646adcbd67e140/test/test_helper/common.bash#L248


Assumption: Here some magic happens and X_EXTRA_ARGS now contains the two elements from defined array defined at the beginning.
https://github.com/polarathene/docker-mailserver/blob/d8644fbb09bfaaa3dec03fc611646adcbd67e140/test/test_helper/common.bash#L260

I am wondering, how a function can access other local variables.

Still not understanding, why in that case `local -n X_EXTRA_ARGS=${1}` is needed and the docker run command is not just using `"$@" \`, instead of `"${X_EXTRA_ARGS[@]}" \`

Thank you for your patience 👍

polarathene: > Assumption: Here some magic happens and X_EXTRA_ARGS now contains the two elements from defined array defined at the beginning.

```bash
local -n X_EXTRA_ARGS=${1}
```

- Take the 1st parameter received (a string `TEST_DOCKER_ARGS`) and via `-n` reference the variable from the current scope.
- `local` scopes it the declaration/value within the bounds of the function with the `local` keyword, it's value is visible to sub-function calls until the function with the `local` scope has finished and the value is lost or changed back to what it was previously defined as.
- When the BATS tests run, IIRC, `setup_file()` is run once, and can use `export` to make variables available to the individual test cases. `local` will not pass over to these, I'm not sure what's going on under the hood, BATS perhaps is calling them as subprocesses/subshells or something where `export` works? While the test itself runs (which is scoped within the `@test "description" { ... }`, it can define `local` and regular bash functions will be called afterwards so scoping works fine without using `export`.
- Any functions called that define their own `local` keyworded vars will also have those limited scope to those functions bounds, thus not be available to anything further up the call chain.
- I chose the name `X_EXTRA_ARGS` here instead to try emphasize it was temporary, which although the `local` probably does that, this var is working a bit differently as a reference with `-n` so I chose that `X_` prefix convention :sweat_smile: 
- I don't think I could provide the same `TEST_DOCKER_ARGS` var name as it might not work well with the reference feature (I haven't confirmed that), and I'm not requiring that the test use the array var name `TEST_DOCKER_ARGS` (_otherwise we could probably avoid parsing in a parameter at all and hard-code the string I guess?_), the `X_` prefix also sort of safeguards against someone using an variable `EXTRA_ARGS` which might not be passed in as a string but if it would conflict, this was just an easy way to not worry about it :sweat_smile:
- Anything prefixed with `TEST_` here is usually intended for these `common_container` methods and can be overrided at any stage of the test where needed (in the letsencrypt test PR open atm, I do that with `TEST_CA` for one test case.

> Still not understanding, why in that case `local -n X_EXTRA_ARGS=${1}` is needed and the docker run command is not just using `"$@" \`, instead of `"${X_EXTRA_ARGS[@]}" \`

Presently that'd require each test case calling `common_container_setup "${TEST_DOCKER_ARGS[@]}"` as well. Then passing that through the method to `common_container_create` (unless calling this directly instead), and then expanding with just `"$@" as you mention.

The current approach is more terse in that sense, and I kind of prefer seeing the descriptive `X_EXTRA_ARGS` as an arg where it's being used, even if that might still be obvious what it's doing with just `"$@" \` :man_shrugging: 

More importantly though, what happens if you want to have additional parameters for `common_container_setup` or `common_container_create` that aren't these array of values to expand in the `docker create` command? Then it seems to get more messy to handle, vs defining a new local var to catch `${2}` or whatever and using that where needed.

There is a lot of tests to adapt to these helper methods, I'm not sure yet if just an array of values for `docker create` will be sufficient. Hopefully that clarifies things better :)

casperklein: > Hopefully that clarifies things better :)

Absolutely, thank you! Especially `local TEST_DOCKER_ARGS` confused me, cause of the `local`.

> it's value is visible to sub-function calls

Didn't know that. I thought, this was only for a function and not sub-function calls. But it makes sense.

casperklein (approved):

test/test_helper/common.bash
polarathene: I removed the `export` in `common.bash:init_with_defaults()` for `TEST_DOCKER_ARGS` with the previous commit since I don't think I have a need for it atm, it was probably for declaring common extra args for an entire test file in `setup_file()` but as it's not referenced anywhere globally and just passed as a reference var to `common_container_setup()`, that could just be exported in `setup_file()` when needed :sweat_smile: 

If you think I could document anything better let me know :+1:

georglauterbach:
I think you @polarathene and I just updated the branch at the same time. My commit does the same, but has no changes 😂

PR #2274, opened 2021-10-31
refactor: acme.json extraction

polarathene:
# Description

This has been extracted from https://github.com/docker-mailserver/docker-mailserver/pull/2245 as it does not need to be part of that PR.

It has additionally extracted the python logic to an actual python file that I've refactored and made compatible with Python 3, as Python 2 is EOL and no longer in newer versions of Debian. Required for https://github.com/docker-mailserver/docker-mailserver/pull/2116 to pass tests.

Split into scoped commits with messages if it helps with review :)

**Commit Summary:**

**`check-for-changes.sh`**

- Prevent `SSL_DOMAIN` silently skipping when value has wildcard prefix `*.` (_at least this was known as a bugfix when originally committed in linked PR_).
- Improved inlined docs for maintainers.
- Additional logging for debugging.

**`helper-functions.sh:_extract_certs_from_acme`**:

- Fail if the input arg (_`$CERT_DOMAIN`, aka the FQDN_) provided for extraction is empty.
- Use `$CERT_DOMAIN` in place of `$HOSTNAME` and `$1` for a consistent value (_previously could mismatch, eg with `SSL_DOMAIN` defined_).
- The conditional is now only for handling extraction failure (_key or cert value is missing from extraction_).
- Log an actual warning or success (debug) based on outcome.
- Don't use `SSL_DOMAIN` with wildcard value for the `mkdir` letsencrypt directory name (_wildcard prefix `*.` is first stripped instead_).

**`acme_extract`** (_new python utility for `acme.json` handling_):

- Extracted out into a python script that can be treated as a utility in the `$PATH` like other helper scripts. It can now be used and optionally tested directly instead of via `helper-functions.sh`.

<details>
<summary>Interacting with the python script has following output:</summary>

```console
$ acme_extract --help
usage: acme_extract [-h] (--key | --cert) <filepath> <FQDN>

Traefik acme.json key and cert extractor utility.

positional arguments:
  <filepath>  Path to acme.json
  <FQDN>      FQDN to match in a certificate 'main' or 'sans' field

optional arguments:
  -h, --help  show this help message and exit
  --key       Output the key data to stdout
  --cert      Output the cert data to stdout
```

`--key` or `--cert` is actually required, so minor mistake by the `argparser` Python module with it's generated help text above:

```console
$ acme_extract ./acme.json mail.example.com
usage: testy.py (--key | --cert) <filepath> <FQDN>
acme_extract: error: one of the arguments --key --cert is required

$ acme_extract
[...]

PR #2278, opened 2021-11-01
chore: Refactor `setup-stack.sh` case `SSL_TYPE=letsencrypt`

polarathene:
# Description

This has been extracted from https://github.com/docker-mailserver/docker-mailserver/pull/2245 as it does not need to be part of that PR.

To function correctly, it depends on a `helper-function.sh` method added in https://github.com/docker-mailserver/docker-mailserver/pull/2274 , tests will pass once that is merged.

---

Mostly cleans up the code and documents it better, although there are some minor fixes for handling `SSL_DOMAIN` ENV and additional logging added for spotting issues related to it in future when troubleshooting.

Commits are scoped with context messages for easing review if necessary. Overview of changes:

Traefik specific:

- Logic extracted out into it's own function.
- Conditional reworked to assist with debugging.
- `SSL_DOMAIN` must not be empty when attempting to extract.
- Added additional notes.

`SSL_TYPE=letsencrypt` case:

- Revised top note block.
- Correct handling for `SSL_DOMAIN`.
- Removed some unnecessary nesting.
- Less repetitive error message for `LETSENCRYPT_DOMAIN`.
- Added use of panics where appropriate (kept `return 1` so failures still exit functionality early).
- Improved inline docs.

## Type of change

## Checklist:


georglauterbach (approved):
LGTM 👍🏼

georglauterbach:
Please label your PRs  to make it more obvious what they are about even when not having them opened :)

target/scripts/startup/setup-stack.sh
```diff
@@ -986,68 +1018,66 @@ function _setup_ssl
...
-      local LETSENCRYPT_KEY=""
+      local LETSENCRYPT_DOMAIN
+      local LETSENCRYPT_KEY
 
-      # 2020 feature intended for Traefik v2 support only:
-      # https://github.com/docker-mailserver/docker-mailserver/pull/1553
-      # Uses `key.pem` and `fullchain.pem`
-      if [[ -f /etc/letsencrypt/acme.json ]]
+      # Identify a valid letsencrypt FQDN folder to use.
+      if [[ -n ${SSL_DOMAIN} && -e /etc/letsencrypt/live/$(_strip_wildcard_prefix "${SSL_DOMAIN}")/fullchain.pem ]]
```
casperklein, line 1040: Just stylistic, because we use this style in (all?) other places:

```suggestion
      if [[ -n ${SSL_DOMAIN} ]] && [[ -e /etc/letsencrypt/live/$(_strip_wildcard_prefix "${SSL_DOMAIN}")/fullchain.pem ]]
```

casperklein (changes requested):

target/scripts/startup/setup-stack.sh
georglauterbach: Your suggestion is good style and should be preferred, yes :)

polarathene:
> Please label your PRs to make it more obvious what they are about even when not having them opened :)

Oh sure! Sorry I never thought about that :sweat_smile:

georglauterbach (approved):

casperklein (approved):

PR #2280, opened 2021-11-01
refactor: Internal HOSTNAME and DOMAINNAME configuration

polarathene:
# Description

This has been extracted from https://github.com/docker-mailserver/docker-mailserver/pull/2245 to simplify review.

These changes are the main focus of the original PR where the bulk of the discussion and investigation was oriented around.

Originally reworked by @NorseGaud with heavy refactoring by me afterwards.

---

Commits have been scoped with context via commit messages for ease of reviewing.

Summary of changes:

- Better logical flow, handling and inline documentation.
- Panic when `$HOSTNAME` is misconfigured.
- Heavy inline docs for maintainers to understand and be aware of related functionality.
- `SRS_DOMAINNAME` test should use `--domainname` not ENV `DOMAINNAME` (_reverted this fallback that was added from the [PR that originally introduced `_obtain_hostname_and_domainname`](https://github.com/docker-mailserver/docker-mailserver/pull/2175)_)

There was discussion to consider dropping config ENV `OVERRIDE_HOSTNAME` and the internal `DOMAINNAME` in future, and rename our `HOSTNAME` var to `DMS_FQDN`. Before that happens tests need to be overhauled to more easily perform matrix testing (_different base configs where we can verify tests pass for various scenarios_) to establish confidence in what may likely be a breaking change when introduced. As such that's delayed to a future PR.

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/helper-functions.sh
+++ b/target/scripts/helper-functions.sh
@@ -288,27 +288,39 @@ export -f _monitored_files_checksums
 
 # ? --------------------------------------------- General
 
+# Outputs the DNS label count (delimited by `.`) for the given input string.
+# Useful for determining an FQDN like `mail.example.com` (3), vs `example.com` (2).
+function _get_label_length
+{
+  local INPUT=${1}
+  awk -F '.' '{ print NF }' <<< "${INPUT}"
+}
+
+# Sets HOSTNAME and DOMAINNAME globals used throughout the scripts,
+# and any subprocesses called that intereact with it.
 function _obtain_hostname_and_domainname
 {
-  if [[ -n "${OVERRIDE_HOSTNAME}" ]]
+  # Normally this value would match the output of `hostname` which mirrors `/proc/sys/kernel/hostname`,
+  # However for legacy reasons, the system ENV `HOSTNAME` was replaced here with `hostname -f` instead.
+  export HOSTNAME="${OVERRIDE_HOSTNAME:-"$(hostname -f)"}"
+
+  # If the `HOSTNAME` is more than 2 labels long (eg: mail.example.com),
+  # We take the FQDN from it, minus the 1st label (aka _short hostname_, `hostname -s`).
+  if [[ $(_get_label_length "${HOSTNAME}") -gt 2 ]]
   then
-    export HOSTNAME="${OVERRIDE_HOSTNAME}"
-    export DOMAINNAME="${DOMAINNAME:-${HOSTNAME#*.}}"
-    # Handle situations where the hostname is name.tld and hostname -d ends up just showing "tld"
-    if [[ ! "${DOMAINNAME}" =~ .*\..* ]]
+    if [[ -n ${OVERRIDE_HOSTNAME} ]]
     then
-      DOMAINNAME="${HOSTNAME}"
-    fi
-  else
-    # These hostname commands will fail with "hostname: Name or service not known"
-    # if the hostname is not valid (important for tests)
-    HOSTNAME="$(hostname -f)"
-    DOMAINNAME="${DOMAINNAME:-$(hostname -d)}"
-    if [[ ! "${DOMAINNAME}" =~ .*\..* ]]
-    then
-      DOMAINNAME="${HOSTNAME}"
+      # Emulates the intended behaviour of `hostname -d`
+      DOMAINNAME="$(echo "${HOSTNAME}" | cut -d '.' -f2-99)"
+    else
+      # Operates on the FQDN returned from querying `/etc/hosts` or fallback DNS
+      DOMAINNAME="$(hostname -d)"
     fi
   fi
+
+  # Otherwise we assign the same value (eg: example.com):
+  # Not an else statement in the previous conditional in the event that `hostname -d` fails.
+  DOMAINNAME="${DOMAINNAME:-"${HOSTNAME}"}"
 }
 
 # Call this method when you want to panic (emit a 'FATAL' log level error, and exit uncleanly).
--- a/target/scripts/helper-functions.sh
+++ b/target/scripts/helper-functions.sh
@@ -329,7 +329,7 @@ function _obtain_hostname_and_domainname
     if [[ -n ${OVERRIDE_HOSTNAME} ]]
     then
       # Emulates the intended behaviour of `hostname -d`:
-      DOMAINNAME="$(echo "${HOSTNAME}" | cut -d '.' -f2-)"
+      DOMAINNAME="$(echo "${HOSTNAME}" | cut -d '.' -f2-99)"
     else
       # Operates on the FQDN returned from querying `/etc/hosts` or fallback DNS:
       #
--- a/target/scripts/helper-functions.sh
+++ b/target/scripts/helper-functions.sh
@@ -329,7 +329,8 @@ function _obtain_hostname_and_domainname
     if [[ -n ${OVERRIDE_HOSTNAME} ]]
     then
       # Emulates the intended behaviour of `hostname -d`:
-      DOMAINNAME="$(echo "${HOSTNAME}" | cut -d '.' -f2-)"
+      # Assign the HOSTNAME value minus everything up to and including the first `.`
+      DOMAINNAME=${HOSTNAME#*.}
     else
       # Operates on the FQDN returned from querying `/etc/hosts` or fallback DNS:
       #
--- a/target/scripts/helper-functions.sh
+++ b/target/scripts/helper-functions.sh
@@ -324,7 +324,7 @@ function _obtain_hostname_and_domainname
   # `hostname -d` was probably not the correct command for this intention either.
   # Needs further investigation for relevance, and if `/etc/hosts` is important for consumers
   # of this variable or if a more deterministic approach with `cut` should be relied on.
-  if [[ $(_get_label_length "${HOSTNAME}") -gt 2 ]]
+  if [[ $(_get_label_count "${HOSTNAME}") -gt 2 ]]
   then
     if [[ -n ${OVERRIDE_HOSTNAME} ]]
     then
--- a/target/scripts/helper-functions.sh
+++ b/target/scripts/helper-functions.sh
@@ -329,7 +329,7 @@ function _obtain_hostname_and_domainname
     if [[ -n ${OVERRIDE_HOSTNAME} ]]
     then
       # Emulates the intended behaviour of `hostname -d`:
-      DOMAINNAME="$(echo "${HOSTNAME}" | cut -d '.' -f2-99)"
+      DOMAINNAME="$(echo "${HOSTNAME}" | cut -d '.' -f2-)"
     else
       # Operates on the FQDN returned from querying `/etc/hosts` or fallback DNS:
       #
--- a/target/scripts/helper-functions.sh
+++ b/target/scripts/helper-functions.sh
@@ -302,6 +302,13 @@ function _obtain_hostname_and_domainname
 {
   # Normally this value would match the output of `hostname` which mirrors `/proc/sys/kernel/hostname`,
   # However for legacy reasons, the system ENV `HOSTNAME` was replaced here with `hostname -f` instead.
+  #
+  # TODO: Consider changing to `DMS_FQDN`; a more accurate name, and removing the `export`, assuming no
+  # subprocess like postconf would be called that would need access to the same value via `$HOSTNAME` ENV.
+  #
+  # TODO: `OVERRIDE_HOSTNAME` was introduced for non-Docker runtimes that could not configure an explicit hostname.
+  # k8s was the particular runtime in 2017. This does not update `/etc/hosts` or other locations, thus risking
+  # inconsistency with expected behaviour. Investigate if it's safe to remove support. (--net=host also uses this as a workaround)
   export HOSTNAME="${OVERRIDE_HOSTNAME:-"$(hostname -f)"}"
 
   # If the container is misconfigured.. `hostname -f` (which derives it's return value from `/etc/hosts` or DNS query),
@@ -313,14 +320,26 @@ function _obtain_hostname_and_domainname
 
   # If the `HOSTNAME` is more than 2 labels long (eg: mail.example.com),
   # We take the FQDN from it, minus the 1st label (aka _short hostname_, `hostname -s`).
+  #
+  # TODO: For some reason we're explicitly separating out a domain name from our FQDN,
+  # `hostname -d` was probably not the correct command for this intention either.
+  # Needs further investigation for relevance, and if `/etc/hosts` is important for consumers
+  # of this variable or if a more deterministic approach with `cut` should be relied on.
   if [[ $(_get_label_length "${HOSTNAME}") -gt 2 ]]
   then
     if [[ -n ${OVERRIDE_HOSTNAME} ]]
     then
-      # Emulates the intended behaviour of `hostname -d`
+      # Emulates the intended behaviour of `hostname -d`:
       DOMAINNAME="$(echo "${HOSTNAME}" | cut -d '.' -f2-99)"
     else
-      # Operates on the FQDN returned from querying `/etc/hosts` or fallback DNS
+      # Operates on the FQDN returned from querying `/etc/hosts` or fallback DNS:
+      #
+      # Note if you want the actual NIS `domainname`, use the `domainname` command,
+      # or `cat /proc/sys/kernel/domainname`.
+      # Our usage of `domainname` is under consideration as legacy, and not advised
+      # going forward. In future our docs should drop any mention of it.
+
+      #shellcheck disable=SC2034
       DOMAINNAME="$(hostname -d)"
     fi
   fi
--- a/target/scripts/helper-functions.sh
+++ b/target/scripts/helper-functions.sh
@@ -304,6 +304,13 @@ function _obtain_hostname_and_domainname
   # However for legacy reasons, the system ENV `HOSTNAME` was replaced here with `hostname -f` instead.
   export HOSTNAME="${OVERRIDE_HOSTNAME:-"$(hostname -f)"}"
 
+  # If the container is misconfigured.. `hostname -f` (which derives it's return value from `/etc/hosts` or DNS query),
+  # will result in an error that returns an empty value. This warrants a panic.
+  if [[ -z ${HOSTNAME} ]]
+  then
+    dms_panic__misconfigured 'obtain_hostname' '/etc/hosts'
+  fi
+
   # If the `HOSTNAME` is more than 2 labels long (eg: mail.example.com),
   # We take the FQDN from it, minus the 1st label (aka _short hostname_, `hostname -s`).
   if [[ $(_get_label_length "${HOSTNAME}") -gt 2 ]]
--- a/target/scripts/helper-functions.sh
+++ b/target/scripts/helper-functions.sh
@@ -290,10 +290,9 @@ export -f _monitored_files_checksums
 
 # Outputs the DNS label count (delimited by `.`) for the given input string.
 # Useful for determining an FQDN like `mail.example.com` (3), vs `example.com` (2).
-function _get_label_length
+function _get_label_count
 {
-  local INPUT=${1}
-  awk -F '.' '{ print NF }' <<< "${INPUT}"
+  awk -F '.' '{ print NF }' <<< "${1}"
 }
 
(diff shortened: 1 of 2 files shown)


polarathene:
Current test failure:

```
not ok 48 checking SRS: DOMAINNAME is handled correctly in 15369ms
# (from function `repeat_until_success_or_timeout' in file test/test_helper/common.bash, line 37,
#  in test file test/mail_hostname.bats, line 76)
#   `repeat_until_success_or_timeout 15 docker exec mail_domainname grep "SRS_DOMAIN=my-domain.com" /etc/default/postsrsd' failed
# Timed out on command: docker exec mail_domainname grep SRS_DOMAIN=my-domain.com /etc/default/postsrsd
```

That is resolved by the updated `mail_hostname.bats` test in the original PR.

Technically this is a false-positive, `DOMAINNAME` is meant to be only used internally, we don't support configuring it via ENV (_previously, prior to this change it would use `DOMAINNAME` from ENV if it was already set, we now explicitly avoid that_).

I will look into history of the test and project issues to confirm, then update the tests here with minimum to pass. Updated hostnames test can come via separate PR as it's mostly re-arranging the tests IIRC. It did adjust the tests to use proper `--domainname` with `--hostname` CLI option which was probably the original intention by the test author.

georglauterbach:
Anything I can still help you with here?

polarathene:
> Anything I can still help you with here?

Nope, I got it thanks! :+1: 

Just had to allocate the past week elsewhere. I should hopefully get this sorted over the weekend :)

polarathene:
I don't consider the following as blocking for merge of this PR (_no regressions are introduced AFAIK_), but they are relevant context wise, referencing my comments from the original PR discussion thread/review.

Including here for traceability for myself and future maintainers:

- **Not handled yet:**
  - **Bare domains** (_two DNS labels only, for the FQDN_) will still result in misconfiguration with Postfix. This should be handled similar to how LDAP does to avoid issues ([More details](https://github.com/docker-mailserver/docker-mailserver/pull/2245#discussion_r735188897)).
  - **`--network=host`** is a case where `/etc/hosts` may not be configured correctly. Present advice is to use `OVERRIDE_HOSTNAME`. I don't consider that sufficient as it does not update `/etc/hosts`, only a workaround for `HOSTNAME` usage. _Needs investigation + documentation_ ([More details](https://github.com/docker-mailserver/docker-mailserver/pull/2245#discussion_r735239248)).
- **Test `mail_hostname.bats`:** Still needs to be extracted from original PR (_now closed_), it was mostly a refactor (_but included some `/etc/hosts` checks_).
  - I have a [review comment to go over again](https://github.com/docker-mailserver/docker-mailserver/pull/2245#discussionr726768146) relevant to that. I will handle that with a follow-up PR that additionally converts it to using the new common container template approach.
- **v11:** Should probably _deprecate support for `domainname`_, possibly with some related ENV, but more thorough testing is required ([More details](https://github.com/docker-mailserver/docker-mailserver/pull/2245#discussion_r728729281)).
  - `SSL_DOMAIN` and `OVERRIDE_HOSTNAME` history [detailed here](https://github.com/docker-mailserver/docker-mailserver/pull/2245#discussion_r730317941) and [here](https://github.com/docker-mailserver/docker-mailserver/pull/2245#discussion_r729449401) (_also details `SRS_DOMAINNAME`, and a 2017 switch for `HOSTNAME` from using the command `hostname`, to `hostname -f` as a workaround fix when using `--link`.._), along with [this comment](https://github.com/docker-mailserver/docker-mailserver/pull/2245#discussion_r730319205) about cert provisioning for a bare domain but configuring a hostname with a subdomain scenario.
  - **Reference:** `HOSTNAME` and `DOMAINNAME` handling in `_obtain_hostname_and_domainname`, with comparison to before/after changes in original PR, and other insights worth looking over again are [detailed here](https://github.com/docker-mailserver/docker-mailserver/pull/2245#discussion_r729593921).

---

Future work:

- `DOMAINNAME` => `DMS_DOMAINNAME` (or similar), potentially with deprecation notice for the `domainname` user config.
- `OVERRIDE_HOSTNAME` and `SSL_DOMAIN` => Dropping support with deprecation notice once test matrix is implemented and confirms safe for removal (_with original scenarios requiring these better documented_). `SRS_DOMAINNAME` may warrant consideration as well.
- `HOSTNAME` => `DMS_FQDN`, after the prior two changes are able to ensure it's safe to make this change. Potentially switching back to `hostname` instead of `hostname -f` or considering other alternatives.

When looking into these changes, this comment itself provides ample reference for insights in doing so. In addition, `hostname` and `domainname` config options along with command output was tested with a variety of configurations on both `Debian` and `Alpine` base images (_just the base images, without DMS additions_), along with interaction of `/etc/hosts` content is detailed [here](https://github.com/docker-mailserver/docker-mailserver/pull/2245#issuecomment-942919301). That resource can be revised with the next iteration and added to maintainer docs, which in combination with an improved test suite will help avoid this area becoming problematic in future contributions.

target/scripts/helper-functions.sh
```diff
@@ -288,27 +288,65 @@ export -f _monitored_files_checksums
 
 # ? --------------------------------------------- General
 
+# Outputs the DNS label count (delimited by `.`) for the given input string.
+# Useful for determining an FQDN like `mail.example.com` (3), vs `example.com` (2).
+function _get_label_length
```
casperklein, line 293: I think, count is more suitable:

```suggestion
function _get_label_count
```

As `mail` (third level domain), `example` (second level domain) and `com` (top level) are all "domains", `domain_count` could also be considered:

```suggestion
function _get_domain_count
```

target/scripts/helper-functions.sh
```diff
@@ -288,27 +288,65 @@ export -f _monitored_files_checksums
...
+{
+  local INPUT=${1}
+  awk -F '.' '{ print NF }' <<< "${INPUT}"
```
casperklein, line 296: One-liner:
```suggestion
  awk -F '.' '{ print NF }' <<< "${1}"
```

target/scripts/helper-functions.sh
```diff
@@ -288,27 +288,65 @@ export -f _monitored_files_checksums
...
+  # `hostname -d` was probably not the correct command for this intention either.
+  # Needs further investigation for relevance, and if `/etc/hosts` is important for consumers
+  # of this variable or if a more deterministic approach with `cut` should be relied on.
+  if [[ $(_get_label_length "${HOSTNAME}") -gt 2 ]]
+  then
+    if [[ -n ${OVERRIDE_HOSTNAME} ]]
     then
-      DOMAINNAME="${HOSTNAME}"
+      # Emulates the intended behaviour of `hostname -d`:
+      DOMAINNAME="$(echo "${HOSTNAME}" | cut -d '.' -f2-99)"
```
casperklein, line 333: No "99" limit:

```suggestion
      DOMAINNAME="$(echo "${HOSTNAME}" | cut -d '.' -f2-)"
```

target/scripts/helper-functions.sh
```diff
@@ -288,27 +288,65 @@ export -f _monitored_files_checksums
...
+      # going forward. In future our docs should drop any mention of it.
+
+      #shellcheck disable=SC2034
+      DOMAINNAME="$(hostname -d)"
     fi
   fi
+
+  # Otherwise we assign the same value (eg: example.com):
+  # Not an else statement in the previous conditional in the event that `hostname -d` fails.
+  DOMAINNAME="${DOMAINNAME:-"${HOSTNAME}"}"
```
casperklein, line 349: Quotes unnecessary:

```suggestion
  DOMAINNAME="${DOMAINNAME:-${HOSTNAME}}"
```

target/scripts/helper-functions.sh
```diff
@@ -288,27 +288,65 @@ export -f _monitored_files_checksums
...
+  # Normally this value would match the output of `hostname` which mirrors `/proc/sys/kernel/hostname`,
+  # However for legacy reasons, the system ENV `HOSTNAME` was replaced here with `hostname -f` instead.
+  #
+  # TODO: Consider changing to `DMS_FQDN`; a more accurate name, and removing the `export`, assuming no
+  # subprocess like postconf would be called that would need access to the same value via `$HOSTNAME` ENV.
+  #
+  # TODO: `OVERRIDE_HOSTNAME` was introduced for non-Docker runtimes that could not configure an explicit hostname.
+  # k8s was the particular runtime in 2017. This does not update `/etc/hosts` or other locations, thus risking
+  # inconsistency with expected behaviour. Investigate if it's safe to remove support. (--net=host also uses this as a workaround)
+  export HOSTNAME="${OVERRIDE_HOSTNAME:-"$(hostname -f)"}"
```
casperklein, line 312: Quotes unnecessary:

```suggestion
  export HOSTNAME="${OVERRIDE_HOSTNAME:-$(hostname -f)}"
```

casperklein (changes requested):

target/scripts/helper-functions.sh
polarathene: Thanks, I just carried that over from the original PR, would it also be helpful here to add an inline comment that it's taking everything after the 1st label? Or do you have a better approach instead of `cut` for this?

I don't know how many maintainers are as familiar/comfortable with lines like this, I know I definitely didn't know much shell syntax when I started helping maintain the project, and different utilities like `cut` parameters would confuse me even more so (_although I haven't tried to look up if `-f2-` would make sense to me reading docs, I know I could probably find `-f<number>` at least_).

target/scripts/helper-functions.sh
polarathene: I usually prefer a named variable myself for clarity but in this case I guess it's not too important :sweat_smile:

target/scripts/helper-functions.sh
polarathene: > I think, count is more suitable:

Absolutely :+1: 

> domain_count could also be considered:

I find `domain` as a bit overloaded and easy for others to misinterpret :sweat_smile: 

FQDN does describe itself with domains like you mention, but domain names themselves and hostname, also as DNS labels. I prefer to go with DNS labels (as per RFC on DNS) for this. I feel that `label` is more clear about the count/separation than `domain` is.

target/scripts/helper-functions.sh
```diff
@@ -288,27 +288,64 @@ export -f _monitored_files_checksums
...
+  # `hostname -d` was probably not the correct command for this intention either.
+  # Needs further investigation for relevance, and if `/etc/hosts` is important for consumers
+  # of this variable or if a more deterministic approach with `cut` should be relied on.
+  if [[ $(_get_label_length "${HOSTNAME}") -gt 2 ]]
+  then
+    if [[ -n ${OVERRIDE_HOSTNAME} ]]
     then
-      DOMAINNAME="${HOSTNAME}"
+      # Emulates the intended behaviour of `hostname -d`:
+      DOMAINNAME="$(echo "${HOSTNAME}" | cut -d '.' -f2-)"
```
polarathene, line 332: Presumably this change-request wasn't accurate?

```suggestion
      DOMAINNAME="$(echo "${HOSTNAME}" | cut -d '.' -f2-99)"
```

target/scripts/helper-functions.sh
```diff
@@ -288,27 +288,64 @@ export -f _monitored_files_checksums
...
+  fi
+
+  # If the `HOSTNAME` is more than 2 labels long (eg: mail.example.com),
+  # We take the FQDN from it, minus the 1st label (aka _short hostname_, `hostname -s`).
+  #
+  # TODO: For some reason we're explicitly separating out a domain name from our FQDN,
+  # `hostname -d` was probably not the correct command for this intention either.
+  # Needs further investigation for relevance, and if `/etc/hosts` is important for consumers
+  # of this variable or if a more deterministic approach with `cut` should be relied on.
+  if [[ $(_get_label_length "${HOSTNAME}") -gt 2 ]]
```
polarathene, line 327: I'm tired :sweat_smile: forgot to change the called method when changing the method name..

```suggestion
  if [[ $(_get_label_count "${HOSTNAME}") -gt 2 ]]
```

target/scripts/helper-functions.sh
casperklein: First: I am never against comments, even if I don't would add them myself in some cases. `cut` is heavily used in this project, so the usage should be clear/familiar. There is also no prevention, to look up unknown things in the man pages or do a web search 😉

>  I know I definitely didn't know much shell syntax

And every time you stumble upon something like this, you learn something new. Same here. When I first saw this (in another code), I didn't know the meaning. Now I do and use it myself 👍 So this is never an excuse for me, not to use something. Just because someone wouldn't understand it at first..

> Or do you have a better approach instead of cut for this?

You asked for it, you get it 😆 

`DOMAINNAME=${HOSTNAME#*.}`

Explained:

```
#     remove string from the beginning (non-greedy) --> ## (greedy)
*     wildcard
.     dot

--> Remove all strings beginning from the left, until (including) the first dot.
```

But someone not familiar with bash variable expansion will "complain" about this too. This solution is more efficient (no sub-process needed). I admit, more readable on **first** sight, is probably the `cut` command.
I would use the variable expansion.

target/scripts/helper-functions.sh
```diff
@@ -288,27 +288,64 @@ export -f _monitored_files_checksums
...
+  # Normally this value would match the output of `hostname` which mirrors `/proc/sys/kernel/hostname`,
+  # However for legacy reasons, the system ENV `HOSTNAME` was replaced here with `hostname -f` instead.
+  #
+  # TODO: Consider changing to `DMS_FQDN`; a more accurate name, and removing the `export`, assuming no
+  # subprocess like postconf would be called that would need access to the same value via `$HOSTNAME` ENV.
+  #
+  # TODO: `OVERRIDE_HOSTNAME` was introduced for non-Docker runtimes that could not configure an explicit hostname.
+  # k8s was the particular runtime in 2017. This does not update `/etc/hosts` or other locations, thus risking
+  # inconsistency with expected behaviour. Investigate if it's safe to remove support. (--net=host also uses this as a workaround)
+  export HOSTNAME="${OVERRIDE_HOSTNAME:-$(hostname -f)}"
```
georglauterbach, line 311: I would not like to see `OVERRIDE_HOSTNAME` being removed as is without introducing another variable, maybe for K8s, that does an equal job. I'd be very hesitant to remove this. Maybe adjust the comment, or remove it at all?

georglauterbach (approved):
LGTM, just this one comment :)

target/scripts/helper-functions.sh
polarathene: >  Maybe adjust the comment, or remove it at all?

The comment stays. It makes it rather clear what the concern is and what needs to be done for it to be considered for removal:

- Introduced for non-Docker runtimes that could not configure an explicit hostname (_k8s was the particular runtime in 2017_). `--net=host` also uses this as a workaround.
- This does not update `/etc/hosts` or other locations, thus risking inconsistency with expected behaviour.
- Investigate if it's safe to remove support.

I'm likely one of the only ones who will investigate it, and that will be a while off as it requires big adjustments to the test suite that I'm slowly chipping away at.

If you can confirm that in almost 2022, this remains a requirement for k8s that is helpful :)

---

> I would not like to see `OVERRIDE_HOSTNAME` being removed as is without introducing another variable

I know of one case with `--network=host` where it is currently required, but I don't consider the workaround satisfactory due to the lack of `/etc/hosts` parity.

What is likely to happen is the tests should show if the `/etc/hosts` addition matters at all, other than `hostname -f` or similar commands or if we can shift away from `hostname`/`domainname` config and have all scenarios unify on a common ENV instead.

If that makes sense to do, it can't be done confidently AFAIK without the test matrix run I want to be able to perform :sweat_smile:

The only way `OVERRIDE_HOSTNAME` will get removed or replaced is via PR anyway, I see no harm in the comment, only additional clarification to future maintainers :)

target/scripts/helper-functions.sh
polarathene: Ah awesome :)

You make some good points so I'm fine with whatever you advise :+1: 

Cheers for all the insights!

casperklein: Take the variable expansion 👍

target/scripts/helper-functions.sh
polarathene:
```suggestion
      # Assign the HOSTNAME value minus everything up to and including the first `.`
      DOMAINNAME=${HOSTNAME#*.}
```

target/scripts/helper-functions.sh
NorseGaud: If it means anything: It took me half an hour to figure out what the heck `#*.}` did... I highly highly recommend against anything that wouldn't be obvious to anyone of any bash skill level.

polarathene: > If it means anything: It took me half an hour to figure out what the heck `#*.}` did...

Does it help that I added a comment about it in the commit, and that anyone who uses `git blame` should be able to arrive on this PR and come across this review sub-thread about it?

NorseGaud: > > If it means anything: It took me half an hour to figure out what the heck `#*.}` did...
> 
> 
> 
> Does it help that I added a comment about it in the commit, and that anyone who uses `git blame` should be able to arrive on this PR and come across this review sub-thread about it?

Sure, the comment is fine. 👍🏼Just seems like overkill/clutter.

target/scripts/helper-functions.sh
georglauterbach: Alright, we keep the comment, this is fine with me :)

georglauterbach (approved):

casperklein (approved):

jpduyx:
i would like my mailserver to pass all the mxtoolbox (and other's) tests including the "reverse DNS name" that is announced with EHLO / HELO stuff) but my reverse DNS name is: 

isphostname.ISPDOMAIN.com not mymailserver.mydomain.com

how to do this? 

if I configure my hostname in docker-compose as the reverse DNS name mxtoolbox tests are all green ... but I get into problems that the reports are send to postmaster@ispdomain.com (and while trying to reduce the spam points of my server, i became a spammer myself)  perhaps also other things will go wrong if ispdomain is then used instead of mydomain.

Is there a way to automatically set the EHLO / HELO response matching the reverse DNS to pass those tests correctly to avoid spam points with my mailserver without breaking other out of the box settings the refactoring of internal HOSTNAME and DOMAINNAME configuration ... ? 

Thank you ... for now I choose to go back to the standard settings and accept the extra spam points and not pass the Reverse DNS tests, to avoid breaking other stuff.

polarathene:
Not relevant to this PR, open a discussion for personal support requests.

Short answer is you can't do anything about that. The IP provider controls rDNS entries, you need a static IP that they agree to change rDNS entry for.

Changing your hostname to the rDNS record makes no sense when you don't own that domain.

You don't need the rDNS, if you're concerned about delivery reputation / success relay through a service like SendGrid.

jpduyx:
Thank for the reply. I'm sorry to put this question here and was not looking for personal support. 

But I would like to know if it would be technally wrong if postfix would could somehow be configured to announce itself with the reverse dns name to avoid spam points .... wouldnt' this be possible to automate  during the docker container setup, without breaking anything else?

Met vriendelijke groet, 

Jean-Paul 

    Op maandag 12 februari 2024 om 13:13:50 EET schreef Brennan Kinney ***@***.***>:  
 
 


Not relevant to this PR, open a discussion for personal support requests.

Short answer is you can't do anything about that. The IP provider controls rDNS entries, you need a static IP that they agree to change rDNS entry for.

Changing your hostname to the rDNS record makes no sense when you don't own that domain.

You don't need the rDNS, if you're concerned about delivery reputation / success relay through a service like SendGrid.

polarathene:
> Thank for the reply. I'm sorry to put this question here and was not looking for personal support. But I would like to know if it would be technally wrong if postfix would could somehow be configured to announce itself with the reverse dns name to avoid spam points .... wouldnt' this be possible to automate  during the docker container setup, without breaking anything else?

This is still the kind of support questions that belong in Discussions, not our pull requests.

The check you're referring to varies by MTA AFAIK, with Postfix it is likely [`reject_unknown_client_hostname`](https://www.postfix.org/postconf.5.html#reject_unknown_client_hostname), I have written [this comment](https://github.com/docker-mailserver/docker-mailserver/issues/2650#issuecomment-1160004457) to explain it better.

It is not as strict of a requirement as you are seeking with rDNS exact match. I don't think that needs to be that strict, so you should be fine.

PR #2281, opened 2021-11-01
Improve VIRUSMAILS_DELETE_DELAY usage

casperklein:
# Description

The `VIRUSMAILS_DELETE_DELAY` variable is set in multiple places. With the introduction of `/etc/dms-settings`, this can be simplified. 

Discussed/Discovered partly here: https://github.com/docker-mailserver/docker-mailserver/pull/2264#issuecomment-955603197#

## Type of change

## Checklist:


casperklein:
Does anyone knows, why shellcheck doesn't care about sourcing here:

https://github.com/docker-mailserver/docker-mailserver/blob/c7dec1e33086bbaf8609b0db6bc8799da5d5bb5b/target/bin/listmailuser#L7

but complaining [here](https://github.com/docker-mailserver/docker-mailserver/runs/4068102148?check_suite_focus=true)?

georglauterbach:
> Does anyone knows, why shellcheck doesn't care about sourcing here:
> 
> https://github.com/docker-mailserver/docker-mailserver/blob/c7dec1e33086bbaf8609b0db6bc8799da5d5bb5b/target/bin/listmailuser#L7
> 
> but complaining [here](https://github.com/docker-mailserver/docker-mailserver/runs/4068102148?check_suite_focus=true)?

A very good question, I have no idea :D

georglauterbach (approved):

polarathene (approved):

polarathene:
> Does anyone knows, why shellcheck doesn't care about sourcing here:

It seems this was missed during review:

https://github.com/docker-mailserver/docker-mailserver/blob/c7dec1e33086bbaf8609b0db6bc8799da5d5bb5b/test/linting/lint.sh#L121

`--source-path="${SCRIPT_DIR}"` is probably what was intended as that's defined earlier in the script. No reference of `SCRIPTDIR` anywhere.

What I do find odd though is why the errors are reduced with that, but substituting the value with say `x` instead did not have the same effect but should be just as invalid :confused: 

---

**EDIT:** Nope, nevermind [`SCRIPTDIR` is a special value for `shellcheck`](https://github.com/koalaman/shellcheck/wiki/Directive#source-path). Without that you will get the failure for `listmailuser` as well among others such as the last var in `start-mailserver.sh` VARS list/section.

Which might hint that it's a similar behaviour with `. /etc/dms-settings`, it only reports the last occurrence for whatever reason.. In [my `common.bash` PR](https://github.com/docker-mailserver/docker-mailserver/pull/2275), you'll see that the bounced spam test update adds a shellcheck disable comment only to the 2nd test case, without needing to do it for the first one despite no difference apart from declaration/source order in the file.

casperklein:
>  among others such as the last var in start-mailserver.sh VARS list/section.

I was wondering about this also some time ago 😆

PR #2282, opened 2021-11-01
Document aliases that forward to multiple addresses

samtuke:
# Description

Document the fact that aliases pointing to multiple email addresses also works.

## Type of change


casperklein (approved):

docs/content/config/user-management/aliases.md
```diff
@@ -7,7 +7,8 @@ Please read the [Postfix documentation on virtual aliases](http://www.postfix.or
 You can use [`setup.sh`][docs-setupsh] instead of creating and editing files manually. Aliases are managed in `/tmp/docker-mailserver/postfix-virtual.cf`. An alias is a _full_ email address that will either be:
 
 * delivered to an existing account registered in `/tmp/docker-mailserver/postfix-accounts.cf`
-* redirected to one or more other email addresses
+* redirected to another email address
+* redirected to multiple email addresses
```
polarathene, line 11: How is that different than the previous content?

This change also is a bit buddy with the previous line. You were perhaps trying to make a distinction between internal accounts/addresses that `docker-mailserver` is responsible for, and external accounts/addresses?

What happens is Postfix expands an alias into one or more recipients, it must accept the mail first before doing this however. Dovecot is only aware of real accounts (_and more recently simple aliases that map to one real address, internal or external, but not another alias, and only if the quota feature is enabled_).

We identified a problem with aliases and Dovecot quotas (_currently enabled by default, in consideration for being changed to opt-in in a future release_). As aliases are not expanded until Postfix accepts the incoming mail, it causes a problem when checking with Dovecot about quota status for the recipient as Dovecot until recently would only report on quota being exceeded for internal accounts, with aliases it gets complicated and if the quota has been exceeded would result in bouncing the mail back to the sender address... which can get your mail server blacklisted for backscatter when spammers forge the sender address to make you send mail to others on their behalf.

We only recently added the fix for basic aliases (_single aliased address to a real internal account only_). And haven't yet updated documentation about this caveat.

---

Also, AFAIK our `setup alias add ...` method doesn't have support to add multiple addressses :man_shrugging: 

Not an issue if editing `postfix-virtual.cf` directly (_quota issue above is still a thing though_).

---

Maybe..?

```suggestion
* redirected to an external email address
* redirected to multiple email addresses that can be real accounts and/or additional aliases 
```

polarathene (changes requested):

georglauterbach:
@samtuke please have a look at the maintainers' questions / suggestions.

docs/content/config/user-management/aliases.md
georglauterbach: @polarathene Is it fine for you if I or you just commit your suggestions and merge this? It's likely we do not get a response in any meaningful time...

polarathene: Sure!

Sorry I haven't been too active lately, I still intend to wrap up the work I've been doing here, I've been caught up with work elsewhere :sweat_smile: (_just in case you're wondering about the status of the stuff I've got pending as WIP PRs_)

polarathene: > Also, AFAIK our `setup alias add ...` method doesn't have support to add multiple addressses man_shrugging
> 
> Not an issue if editing `postfix-virtual.cf` directly (_quota issue above is still a thing though_).

Considering the above though, is that something we want to commit to the documentation in the current state? I don't have time to spare revising this PR further myself right now :sweat_smile:

polarathene:
As this PR is minimal, and the contributor has not responded to feedback, I am deciding to reject the PR for the time being.

AFAIK, while multiple aliases are supported, our `setup` tooling doesn't provide that support at current? And there are other caveats raised in review that should be better documented, without which we are likely to open ourselves to more issues raised about those problems.

If more users express interest in documenting this officially we can consider improving the docs here at that time. As it stands right now, this is just something supported by Postfix when you edit the config file directly and can be found on their docs.

Thank you for the contribution, feel free to discuss further and revise it if you'd like to see it merged into the docs.

PR #2284, opened 2021-11-02
tests(chore): Adjust supported FQDNs in test certs + add wildcard and `acme.json` configs

polarathene:
# Description

These are changes extracted from https://github.com/docker-mailserver/docker-mailserver/pull/2245 which are intended for a rewritten `mail_ssl_letsencrypt.bats` test.

These files will replace the existing `test/config/letsencrypt` content which has some random provisioned FQDN for letsencrypt that doesn't match the FQDN tested, `acme.json` files with FQDNs that don't match those certs FQDNs and changes to certs that won't expire until 2031. `test/config/letsencrypt` will be removed with the associated test update PR.

The changes amount to:

- Re-configuring the FQDN values that some certs were created for (_needed for flexibility in testing_).
- Adding an `*.example.test` wildcard (_both RSA and ECDSA_).
- Adding `acme.json` encoded versions (_traefik extraction support will use these instead_).
- Updated / new internal docs for maintainers of this content.

For more detailed information on those changes, please see the associated commits message.

<details>
<summary>Commit messages added here for convenience</summary>

**chore: Modify `example.test` cert FQDNs + add wildcards:**
- `with_ca/ecdsa` (ECDSA root CA) certs were recreated to restrict to specific FQDNs for testing. The ECDSA cert is now only for `mail.example.test`, and the RSA cert only for `example.test`.
- `with_ca/rsa` (RSA root CA) retains both FQDN for ECDSA and RSA certs to support tests that expect FQDN consistency between both types.
- README removes the `--san` fields from root CA creation as that shouldn't be there (_AFAIK, no effect on the created cert, but I had already recreated all certs by this point and base64 encoded for traefik `acme.json`, so I kept the modified root cert_).
- Updated the README accordingly, along with matching the output to updated file content, and adding a wildcard section.
- Both `RSA` and `ECDSA` root CA folders contain a wildcard cert for `*.example.test` that is of the same type as the root CA cert. These will be used for wildcard testing support.

**chore: Add `acme.json` files for `example.test`:**
- These versions of `acme.json` add base64 encoded certs from the `example.test` directory as documented in the new `traefik.md`.
- Since the `acme.json` files are dependent upon these test certs and this traefik feature is to remain, it makes sense to co-locate these files together.
</details>

## Type of change

- [x] This change requires a documentation update (done)

## Checklist:


wernerfred (approved):
Thumbs up 👍🏻 for the PR description!

LGTM. I checked the changes and read the documentation and wondered why the bot didn't post a Documentation link. Then i realized that these .md files are not in the Documentation. Should we think about moving them to the Documentation in a section about testing?

test/test-files/ssl/example.test/README.md
```diff
@@ -6,6 +6,8 @@ These certificates for usage with TLS have been generated via the [Smallstep `st
 
 `Certificate Details` sections are the output of: `step certificate inspect cert.<key type>.pem`.
 
+Each certificate except for the wildcard one, have the SANs(Subject Alternative Name) `example.test` and `mail.example.test`.
```
casperklein, line 9: Why `example.test` and not `example.com` etc. ?

polarathene: Purely for testing purposes, as the TLD is intended for.

Because `example.com` will resolve via external DNS queries, but `external.test` will not, it's a reserved TLD for internal use among others. My own DNS resolver can be configured to accept it, but public ones will reject it.

I find that helpful for testing as when it does try to query a public resolver for whatever reason I know that it'll fail there too and not mislead me in the logs when troubleshooting :sweat_smile:

polarathene:
> Should we think about moving them to the Documentation in a section about testing?

Our docs do briefly mention the main README here in the [SSL self-signed generation section](https://docker-mailserver.github.io/docker-mailserver/edge/config/security/ssl/#generating-a-self-signed-certificate).

Yeah they could be migrated at a later point, might help with discovery/awareness if contributors prefer browsing that part of our docs. I find a copy co-located with these files makes them fairly visible for anyone who'd be looking at them.

They're just covering what the contents are and how to create them, as at some point I'll be less active here and I don't want to leave maintainers in the dark :sweat_smile:

casperklein (approved):

PR #2295, opened 2021-11-12
make Bash start user-patches.sh

georglauterbach:
# Description

This addition is somewhat important for K8s users and _changes nothing_ in terms of functionality. We currently execute the `user-patches.sh` script with a shebang, making it executable beforehand if it wasn't to begin with.

Now, Bash will take the script and execute it. This change has therefore no implications on the actual execution of the `user-patches.sh` script.

**But for Kubernetes** users, this way of executing the script is the most comfortable one because all others are somewhat tiresome or involve unnecessary work! The explanation:

You would mount this script via a `ConfigMap` into the container. These files are by default not executable and one cannot change the permissions. One would need to specify the permissions for this `ConfigMap` but then you'd do this for all files of this `ConfigMap`, which is not necessarily desirable either. By using Bash and supplying it the script, we can solve this problem with ease!

I marked this as high priority because I need this change to be in `:edge` ASAP :D Please review!

Fixes permissions error for Kubernetes users (`user-patches.sh`).

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/check-for-changes.sh
+++ b/target/scripts/check-for-changes.sh
@@ -115,12 +115,12 @@ do
 
     if [[ -f /tmp/docker-mailserver/postfix-accounts.cf ]] && [[ ${ENABLE_LDAP} -ne 1 ]]
     then
-      sed -i 's/\r//g' /tmp/docker-mailserver/postfix-accounts.cf &>/dev/null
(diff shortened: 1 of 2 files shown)


georglauterbach:
Manually cancelled the first two CI pipelines, no worries there :)

polarathene (approved):

casperklein (approved):

PR #2298, opened 2021-11-13
improve: sort environment variables

casperklein:
# Description

Among other things, this improves readability of `print-environment`, by sorting the ENVs.

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/startup/setup-stack.sh
+++ b/target/scripts/startup/setup-stack.sh
@@ -77,12 +77,16 @@ function _setup_default_vars
 
   : >/root/.bashrc     # make DMS variables available in login shells and their subprocesses
   : >/etc/dms-settings # this file can be sourced by other scripts
+
   local VAR
   for VAR in "${!VARS[@]}"
   do
     echo "export ${VAR}='${VARS[${VAR}]}'" >>/root/.bashrc
     echo "${VAR}='${VARS[${VAR}]}'"        >>/etc/dms-settings
   done
+
+  sort -o /root/.bashrc     /root/.bashrc
+  sort -o /etc/dms-settings /etc/dms-settings
 }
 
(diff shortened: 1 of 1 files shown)


georglauterbach (approved):

target/scripts/startup/setup-stack.sh
```diff
@@ -77,12 +77,16 @@ function _setup_default_vars
...
   : >/etc/dms-settings # this file can be sourced by other scripts
+
   local VAR
   for VAR in "${!VARS[@]}"
   do
     echo "export ${VAR}='${VARS[${VAR}]}'" >>/root/.bashrc
     echo "${VAR}='${VARS[${VAR}]}'"        >>/etc/dms-settings
   done
+
+  sort -o /root/.bashrc     /root/.bashrc
```
polarathene, line 88: I thought this can contain more than ENV? Wouldn't that be bad/risky to sort?

If you did echo `export` to the `/etc/dms-settings` you could just sort that and append to `/root/.bashrc` instead?

polarathene (approved):

target/scripts/startup/setup-stack.sh
casperklein: > I thought this can contain more than ENV? Wouldn't that be bad/risky to sort?

Yes, but potential additions (via (user-patches.sh`) are applied later. Also both files are emptied right above:

https://github.com/docker-mailserver/docker-mailserver/blob/cac0647250fe6e0422c2787745699f8be2d6c813/target/scripts/startup/setup-stack.sh#L78-L79

georglauterbach: Though about exactly the same, but came to the same conclusion as @casperklein

PR #2301, opened 2021-11-15
Revise workflow for stalled objects

georglauterbach:
# Description

~~Follow up on #2300. Sorry for the extra work.~~

Revises the workflow for stale issues **and** PRs. One can now add the label `bot/ignore` to issues and PRs to indicate that the stale action should ignore this object. Furthermore, configuration for PRs was added explictly.

## Type of change

## Checklist:

### DIFF
--- a/.github/workflows/close_stale_issues.yml
+++ b/.github/workflows/close_stale_issues.yml
@@ -13,6 +13,7 @@ jobs:
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
           exempt-issue-labels: "issue/not-stale"
+          exempt-pr-labels: "pr/not-stale"
           days-before-stale: 20
           days-before-close: 10
           stale-issue-label: "meta/stale"
--- a/.github/workflows/handle_stalled.yml
+++ b/.github/workflows/handle_stalled.yml
@@ -28,7 +28,7 @@ jobs:
           stale-pr-label: "meta/stale"
           stale-pr-message: >
             This PR has become stale because it has been open for 20 days without
-            activity. Remove the label and comment or this issue will be closed in 10 days.
+            activity. Remove the label and comment or this PR will be closed in 10 days.
           close-pr-label: "meta/closed due to age or inactivity"
           closed-pr-message: >
             This PR was closed due to inactivity.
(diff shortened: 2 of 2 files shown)


.github/workflows/close_stale_issues.yml
```diff
@@ -13,6 +13,7 @@ jobs:
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
           exempt-issue-labels: "issue/not-stale"
+          exempt-pr-labels: "pr/not-stale"
```
wernerfred, line 16:
```suggestion
          exempt-pr-labels: "bot/ignore"
```

casperklein (changes requested):
I wasn't aware, that PRs are also considered by stale bot, as we have no explicit config for PRs.

The workflow file and title should be renamed, to reflect also PRs.
Also the close message, tags, etc.  should be streamlined. ATM for PRs the label `stale` is used instead of `meta/stale` like in issues.

wernerfred (changes requested):
Maybe i am a bit late to the party as #2300 got already merged but i would highly prefer a solution like suggested (`bot/ignore` or variations like `stalebot/ignore`) over the now introduced `[issue|pr]/not-stale`.

Imho we should avoid making statements on the actual state of an issue or pr (what `not-stale` implies as we cannot guarantee it - its state could change in future or other PRs or might be totally different depending of the individual working on this - but only maintainers can edit the labels not the contributor) but rather highlight that certain automation does not apply here.

georglauterbach:
I will take care.

.github/workflows/close_stale_issues.yml
georglauterbach: 👍🏼

casperklein (approved):
LGTM. The only thing to consider is maybe, to write out "pull requests" instead of PR.

wernerfred (approved):
Thanks

georglauterbach:
If "PR" turns out to be confusing in the future, I will change it :)

polarathene:
Might not matter much, especially since this got merged already, but `bot/ignore` implies any bot should ignore it rather than something like `stale-bot/ignore`.

georglauterbach:
As we only have one bot at the moment, all is well. If we implement a new bot, we can change this too.

PR #2302, opened 2021-11-16
Housekeeping

georglauterbach:
# Description

This is a housekeeping PR. This PR does not and must not contain changes that affect functionality!

This PR does not block the release of `v10.3.0`. **Feel free to add your own small additions** if you think a section could use a bit of cleanup (nothing big, just small things).

If this is merged before `v10.3.0`, please add the corresponding milestone.

## Type of change

## Checklist:

### DIFF
--- a/.github/workflows/handle_stalled.yml
+++ b/.github/workflows/handle_stalled.yml
@@ -1,4 +1,4 @@
-name: "Handled Stalled Issues & PRs"
+name: "Handled Stalled Issues & Merge Requests"
 
 on:
   schedule:
@@ -15,20 +15,22 @@ jobs:
           days-before-stale: 20
           days-before-close: 10
 
-          exempt-issue-labels: "bot/ignore"
-          stale-issue-label: "meta/stale"
+          exempt-issue-labels: stale-bot/ignore
+          stale-issue-label: meta/stale
           stale-issue-message: >
             This issue has become stale because it has been open for 20 days without
-            activity. Remove the label and comment or this issue will be closed in 10 days.
+            activity. Remove the label (maintainer) or comment on this issue (assignee).
+            Otherwise, this issue will be closed in 10 days automatically.
           close-issue-label: "meta/closed due to age or inactivity"
           close-issue-message: >
             This issue was closed due to inactivity.
 
-          exempt-pr-labels: "bot/ignore"
-          stale-pr-label: "meta/stale"
+          exempt-pr-labels: stale-bot/ignore
+          stale-pr-label: meta/stale
           stale-pr-message: >
-            This PR has become stale because it has been open for 20 days without
-            activity. Remove the label and comment or this PR will be closed in 10 days.
+            This merge request has become stale because it has been open for 20 days without activity
+            Remove the label 'meta/stale' (if you're a maintainer) or comment on this merge request
+            (as an assignee). Otherwise, this merge request will be closed in 10 days automatically.
           close-pr-label: "meta/closed due to age or inactivity"
           closed-pr-message: >
             This PR was closed due to inactivity.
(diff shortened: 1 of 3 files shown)


.github/workflows/handle_stalled.yml
```diff
@@ -15,20 +15,22 @@ jobs:
...
-          exempt-pr-labels: "bot/ignore"
-          stale-pr-label: "meta/stale"
+          exempt-pr-labels: stale-bot/ignore
+          stale-pr-label: meta/stale
           stale-pr-message: >
-            This PR has become stale because it has been open for 20 days without
-            activity. Remove the label and comment or this PR will be closed in 10 days.
+            This merge request has become stale because it has been open for 20 days without activity
+            Remove the label 'meta/stale' (if you're a maintainer) or comment on this merge request
+            (as an assignee). Otherwise, this merge request will be closed in 10 days automatically.
```
polarathene, line 33: I'm only familiar with MR being used in Gitlab, while PR for Github.

I don't mind either way, but if you change from "pull request" wording to "merge request", it should be consistent. The `closed-pr-message` value still uses "PR".

Likewise the maintainer/assignee mentions are inconsistent with the phrasing for similar messages in this PR, that should be addressed. Might be worth mentioning the ignore label as another option for maintainers too?:

```suggestion
            This merge request has become stale because it has been open for 20 days without activity. This merge request will be closed in 10 days automatically unless:
            - A maintainer removes the label `meta/stale` label or adds the `stale-bot/ignore` label.
            - New activity occurs, such as a new comment.
```

polarathene (approved):

.github/workflows/handle_stalled.yml
georglauterbach: Had been using GitLab before (and find the term "merge request" more fitting). Since GitHub uses pull requests, I will adjust this, right.

wernerfred (dismissed):

.github/workflows/handle_stalled.yml
```diff
@@ -1,4 +1,4 @@
-name: "Handled Stalled Issues & PRs"
+name: "Handled Stalled Issues & Merge Requests"
```
polarathene, line 1: Missed changing "Merge Requests" here to "Pull Requests", also not sure if you meant "Stalled" or "Stale". I assume "Handle" is more appropriate than "Handled" for the job name too?

```suggestion
name: "Handle Stale Issues & Pull Requests"
```

polarathene (approved):

wernerfred (approved):

polarathene (approved):

target/scripts/startup/setup-stack.sh
```diff
@@ -1742,7 +1742,7 @@ function _setup_user_patches
   if [[ -f ${USER_PATCHES} ]]
   then
     _notify 'tasklog' 'Applying user patches'
-    bash "${USER_PATCHES}"
+    /bin/bash "${USER_PATCHES}"
```
casperklein, line 1745: Why? To be consequent, we should then also use, e.g. `/bin/sed` across our scripts.

georglauterbach: Because all our scripts use the shebang

```
#! /bin/bash
```

and the user patches scrips should also be executed this way.

casperklein: `/bin/bash` is also used, when not providing the full path. You can verify this, with `which bash`.

I find it odd, to prefix the full path to some commands, where not necessary. That's whats the $PATH variable is for 😉

georglauterbach: Right, but I could have also used `/usr/bin/env bash`, but this way, I make it explicit and my intentions clear sich program to call how:)

casperklein: To be clear: I was not talking about the shebang line. I refered to the quoted script line above.

PR #2306, opened 2021-11-27
Rename function and adjust comment

casperklein:
# Description

After a second view on the function, I think this makes more sense.

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/helper-functions.sh
+++ b/target/scripts/helper-functions.sh
@@ -316,8 +316,8 @@ function _obtain_hostname_and_domainname
   DOMAINNAME="${DOMAINNAME:-${HOSTNAME}}"
 }
 
-# Remove string input with empty line, only whitespace or `#` as the first non-whitespace character.
-function _strip_comments
+# Check if string input is an empty line, only whitespaces or `#` as the first non-whitespace character.
+function _is_comment
 {
   grep -q -E "^\s*$|^\s*#" <<< "${1}"
 }
--- a/target/scripts/helpers/relay.sh
+++ b/target/scripts/helpers/relay.sh
@@ -28,7 +28,7 @@ function _relayhost_sasl
     # add domain-specific auth from config file:
     while read -r LINE
     do
-      if ! _strip_comments "${LINE}"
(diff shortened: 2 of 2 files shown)


polarathene (commented):
Probably not necessary due to the rename of the function (good change btw!), although I wasn't immediately sure what the actual output returned was and verified it's `0` or `1`, previously I thought it was returning the original string or an empty string (if it was detected as a comment) :sweat_smile:

casperklein:
Functions cannot return strings. They return the exit status of the last command. In case of grep, the man page states 3 options:

> Normally the exit status is 0 if a line is selected, 1 if no lines were selected, and 2 if an error occurred.

If you want a string to be returned, you can use something like:

```sh
foo() {
        echo "some string"
}
output=$(foo)
echo $output
```

The comment adjustment is correct, as the function does not actually remove anything. Or did you mean to remove the comment entirely? I would keep it.

polarathene:
> Or did you mean to remove the comment entirely? I would keep it.

Oh no the current changes are all great. I was meaning that I wasn't sure what to expect the return value to be, but assumed it was something truthy/falsey, tried it and noted it was `0` or `1` which made sense :sweat_smile: 

I wasn't aware about that with grep or functions, but probably because in the past I've used echo with functions that I called via subshells :)

I think it's fine as is, even for those less confident with bash/scripts the new name is clear it should be a true/false type of value, as well as the usage in an if statement. `git blame` for the curious is probably sufficient for the added clarification you've covered here :+1:

polarathene (approved):
LGTM! (_sorry, my previous review was meant to be an approval_ :sweat_smile: )

georglauterbach (approved):

wernerfred (approved):

georglauterbach:
Just FYI: Nothing will break, but this was marked as `meta/feature-freeze` - it should not have been merged :) But no worries there, just for the next time.

casperklein:
> this was marked as meta/feature-freeze

Sorry, I overlooked that. I added the v10.3.0 milestone and got approval. That's why I thought it was okay to merge.

Issue #2321, opened 2021-12-11
Best practice for pointing at letsencrypt cert maintained by another image


andrewlow:
# Subject
I would like some feedback concerning a use case

## Description

I'm running version `10.3.0`

I use a volume map to point at the host filesystem for my let's encrypt certificate
```
 -v /path/to/config/etc/letsencrypt:/etc/letsencrypt \
```
This works well for the most part.

The let's encrypt certificate is maintained by another image - https://docs.linuxserver.io/general/swag
[as an aside, I think the [linuxserver.io ](https://fleet.linuxserver.io/) folks provide a lot of great images]

This is a nginx server with let's encrypt built in for acquiring your certificates. However, I believe this problem would happen with any other let's encrypt maintenance - possibly even if it was not in a container.

This morning I was messing around with creating a new subdomain, this would have caused the swag image to cause the let's encrypt certificate to be re-issued.  I noticed that mail was subtly busted and looked into the logs

```
Dec 11 10:26:32 mail postfix/smtpd[1766]: warning: error opening chain file: /etc/letsencrypt/live/lowtek.ca/privkey.pem: No such file or directory
```

This was easy to remedy by restarting the mail server image. However, it makes me think there is a problem with how I have this configured. 

I'm assuming that for whatever reason - the mail server was holding on to a path that had been deleted.
Even though a new path was created, the mailserver didn't re-read the path / file - and continued to reference the deleted file? Or maybe this is a docker / host file sharing problem?

Given that others will probably be also using let's encrypt for certificates I figured it was worth mentioning. Either I've done something bad in my configuration - or this is just a corner case on certificate renewal/change that might bite others.

It seems hard to believe that in the months I've been running this, I haven't had a certificate / mailserver update issue - but the certificates don't update all that often, and my server is configured to automatically reboot for security patches - so it will reboot weekly in general - an uptime of 30 days is rare. So it is possible this is a bug that will catch others with the right timing - and regular reboots/restarts avoids it.

casperklein:
If you use `SSL_TYPE=manual`, you don't depend on any DMS logic for detecting the correct path/certificate. Could that be an option for you?

andrewlow:
My `SSL_TYPE` setting is not manual

```
$ grep SSL_TYPE mailserver.env 
SSL_TYPE=letsencrypt
```

[...]

PR #2279, opened 2021-11-01
fix: Monitor for changes in supported letsencrypt locations

polarathene:
# Description

This has been extracted from https://github.com/docker-mailserver/docker-mailserver/pull/2245 as it does not need to be part of that PR. It is related to addressing the problem, but I'm opting to piece-meal extract the bulk of the original PR out for easier review.

This fix is all @NorseGaud with minor fix by quote wrapping the vars and adding comments by me.

---

Previously we only monitored for `$HOSTNAME` in `/etc/letsencrypt/live` and only for hard-coded `.pem` filenames.

This ensures we check the locations of other locations that may not match `$HOSTNAME`, which we also support. Ideally in future at least the directory to look in would be better known in advance..

---

**Current test failure** is due to the lack of quote wrapping vars in `check-for-changes.sh` loop when `SSL_DOMAIN` contains a wildcard prefix (`*.`) for the FQDN. That was fixed in https://github.com/docker-mailserver/docker-mailserver/pull/2274 , once that's merged the test will pass.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Improvement (non-breaking change that does improve existing functionality)

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] New and existing unit tests pass locally with my changes

PR #2326, opened 2021-12-16
fix: Restore detection of letsencrypt certificate file changes

polarathene:
# Description

The `DYNAMIC_FILES` var was quote wrapped, treating all filepaths to create checksums for as a single string that would be ignored instead of processed individually.

Removed the quotes, and changed the for loop to an array which accomplishes the same goal.

This was discovered while investigating [#2321](https://github.com/docker-mailserver/docker-mailserver/issues/2321#issuecomment-994491231), but it's unclear if this will fix that issue.

---

This bug was introduced in https://github.com/docker-mailserver/docker-mailserver/pull/2279 with the `v10.3.0` release.

~~We may want to branch that release tag and cherry pick this commit from master to publish a bugfix (_as we've already merged `v11` changes to master, the bugfix branch is the only way to backport it I think?_).~~ (**RESOLVED:** Decision was to release in `v10.4.0` with the Debian upgrade)

## Type of change

## Checklist:

It'd probably be a good idea to add a test that verifies the cert path is in the checksum file, I'm just a bit pressed on time and resources currently but wanted to get this fix raised while I could spare the time.

If there's no rush to merge I can sort out tests within 7 days depending on how events pan out on my end :sweat_smile:

### DIFF
--- a/target/scripts/helper-functions.sh
+++ b/target/scripts/helper-functions.sh
@@ -187,9 +187,9 @@ function _extract_certs_from_acme
 }
 export -f _extract_certs_from_acme
 
-# Remove the `*.` prefix if it exists
+# Remove the `*.` prefix if it exists, else returns the input value
 function _strip_wildcard_prefix {
-  [[ "${1}" == "*."* ]] && echo "${1:2}"
+  [[ "${1}" == "*."* ]] && echo "${1:2}" || echo "${1}"
 }
 
 # ? --------------------------------------------- Notifications
--- a/target/scripts/helper-functions.sh
+++ b/target/scripts/helper-functions.sh
@@ -224,7 +224,8 @@ export -f _notify
 # shellcheck disable=SC2034
 CHKSUM_FILE=/tmp/docker-mailserver-config-chksum
 
-# Compute checksums of monitored files.
+# Compute checksums of monitored files,
+# returned output is lines of hashed content + filepath pairs.
 function _monitored_files_checksums
 {
   # If a wildcard path pattern (or an empty ENV) would yield an invalid path
@@ -232,13 +233,11 @@ function _monitored_files_checksums
   shopt -s nullglob
 
   # React to any cert changes within the following letsencrypt locations:
-  local DYNAMIC_FILES
-  for FILE in /etc/letsencrypt/live/"${SSL_DOMAIN}"/*.pem \
-              /etc/letsencrypt/live/"${HOSTNAME}"/*.pem   \
-              /etc/letsencrypt/live/"${DOMAINNAME}"/*.pem
-  do
-    DYNAMIC_FILES="${DYNAMIC_FILES} ${FILE}"
-  done
+  local CERT_FILES=(
+    /etc/letsencrypt/live/"${SSL_DOMAIN}"/*.pem
+    /etc/letsencrypt/live/"${HOSTNAME}"/*.pem
+    /etc/letsencrypt/live/"${DOMAINNAME}"/*.pem
+  )
 
   (
     cd /tmp/docker-mailserver || exit 1
@@ -248,7 +247,7 @@ function _monitored_files_checksums
       postfix-aliases.cf \
       dovecot-quotas.cf \
       /etc/letsencrypt/acme.json \
-      "${DYNAMIC_FILES}"
+      ${CERT_FILES[@]}
   )
 }
 export -f _monitored_files_checksums
--- a/target/scripts/helper-functions.sh
+++ b/target/scripts/helper-functions.sh
@@ -239,6 +239,9 @@ function _monitored_files_checksums
     /etc/letsencrypt/live/"${DOMAINNAME}"/*.pem
   )
 
+  # CERT_FILES should expand to separate paths, not a single string;
+  # otherwise fails to generate checksums for these file paths.
+  #shellcheck disable=SC2068
   (
     cd /tmp/docker-mailserver || exit 1
     exec sha512sum 2>/dev/null -- \
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## `v10.3.0`
 
+**WARNING:** This release had a small regression affecting the detection of changes for certificates provisioned in `/etc/letsencrypt` with the config ENV `SSL_TYPE=letsencrypt`, unless you use Traefik's `acme.json`. If you rely on this functionality to restart Postfix and Dovecot when updating your cert files, this will not work and it is advised to upgrade to `v10.4.0` or newer prior to renewal of your certificates.
+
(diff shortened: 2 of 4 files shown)


georglauterbach:
I feel like it would be a good idea for you to run

``` BASH
cat >.git/hooks/pre-commit << "EOM"
#! /bin/bash

set -euEo pipefail
make lint
EOM
```

to setup a pre-commit hook that does the linting :D

georglauterbach:
I've never done a cherry pick but I'd like to learn it :D From what you said, this looks like a valuable solution for a bugfix backport.

wernerfred:
> We may want to branch that release tag and cherry pick this commit from master to publish a bugfix (_as we've already merged `v11` changes to master, the bugfix branch is the only way to backport it I think?_)

IIRC the release CI doesn't trigger on other branches than master so when cherry picked you might need to build and push the image manually to the registries.

This problem follows up the discussion in the maintainer team on the git strategy. Our current setup is not set up to support backports. Implementing such support should be weigh up against the need of backports (didn't happen often in the past, at least since i participate). Another (but ugly) way would be to revert the `v11` commit merge and publish this fix and then add the `v11` changes again

casperklein:
> other (but ugly) way would be to revert the v11 commit merge and publish this fix and then add the v11 changes again

Ugly yes. But easy to do and resulting in very small changes to revert.
I've some PR on my to-do, but wanted to wait, until it's clear, that the next release will definitely be v11 and not another 10.x. 

I am a bit confused wether we treat Debian11 as breaking change and therefore a requirement for our v11 release or not. My last impression was, it's not related to our v11 release.

wernerfred:
> My last impression was, it's not related to our v11 release.

If thats the case we need no revert and can merge straight away... 🚀

polarathene:
> IIRC the release CI doesn't trigger on other branches than master so when cherry picked you might need to build and push the image manually to the registries.

My recall is that tags are not associated to branches, so as long as we have CI that works on the version tags like the docs deployment does, that should work fine via CI?

> This problem follows up the discussion in the maintainer team on the git strategy. Our current setup is not set up to support backports.

I remember that discussion and what I proposed here is exactly what I thought I had suggested was the agreed approach for handling this sort of fix?

> Another (but ugly) way would be to revert the v11 commit merge and publish this fix and then add the v11 changes again

We have done that in the past, the only issue without the bugfix is that anyone running `10.3.0` without `acme.json` for cert management will not get their certs automatically detected upon renewal.

In my own local testing, this did not seem to prevent Postfix from using the renewed cert, even though Postfix was not restarted. I did not test Dovecot however, and I'm not sure why the linked issue was experiencing a problem, so it may depend on how the cert files are provided too.

> I am a bit confused wether we treat Debian11 as breaking change and therefore a requirement for our v11 release or not. My last impression was, it's not related to our v11 release.

Well it did cause breaking changes originally that needed to be fixed. It also upgrades Postfix/Dovecot and has that potential Fail2Ban CVE. Seems enough to warrant being part of a major release, although I wanted to delay that release until sometime in Jan as I was wanting to make other notable changes (_eg: restoring `flock()` file locking, and refactoring SSL/TLS logic to unify modes, rename `intermediate` TLS profile to `legacy`_).

---

If you think it's better to skip a `10.3.1` to merge the Debian 11 as `10.4.0` that's fine, but we should update release notes/changelog about `10.3.0` with a warning that the change detection doesn't react to certificate changes in `/etc/letsencrypt`?

I don't think the Debian 11 upgrade warrants being part of a patch release, only a minor or major.

polarathene:
> to setup a pre-commit hook that does the linting :D

This wouldn't work when making a commit via the Github UI right? My computer isn't presently able to do much work without risking an OOM event lately, so I just contribute small PRs via the web interface and do any testing/verification via a VPS instance with SSH.

Pre-commit wouldn't help with the VPS unless I get into the habit of committing and pushing in the VPS too, but I could at least get into the habit of running the lint checks before opening a PR :sweat_smile: The CI check just lets me be lazy and address it afterwards (since we squash merge commit history :stuck_out_tongue: )

> I've never done a cherry pick but I'd like to learn it :D From what you said, this looks like a valuable solution for a bugfix backport.

I don't know if I do it right whenever I've done it. I have only done so via GitKraken UI, I can right click on a commit from another branch and cherry pick it over, I'm not entirely sure if that is any different than just copy/pasting the change (_just with a fancy git command to do copy/paste via commits instead_).

---

It looks like this will trigger CI just fine via `v10.3.1` tag, doesn't need to be on master branch:

https://github.com/docker-mailserver/docker-mailserver/blob/6ad9dd30635a6e81dfe8dcecb933ee445a1cb145/.github/workflows/default_on_push.yml#L1-L15

wernerfred:
> 1. as a Christmas gift for sysadmins to complain about

😆 😆 😆

polarathene:
> 1. just merging this on top, (waiting one more week and) releasing 10.4 (as a Christmas gift for sysadmins to complain about) and calling it a day (but feeling like Santa himself)

I guess this is the easiest choice if we're not going to try the cherrypick.

I should add the changelog update for `10.3.0` in this PR and update the release page to both add a warning about the regression?

---

It looks like this change also broke some related tests, joy.. :sweat_smile: 

I'll need to investigate and resolve those first.

polarathene:
Tests resolved, fixed another bug (_didn't previously affect any tests as the method was only ever used with `SSL_DOMAIN` to test against documented wildcard prefix_).

Refactored the `check-for-changes.sh` logic for `acme.json` a bit by removing the unnecessary for loop and switch statement. Just uses a regex conditional now. View diff with white-space hidden to reduce noise.

- Cert files extracted for `acme.json` now have their hashes re-calculated afterwards to avoid the redundant change event that they otherwise triggered. Using `sha512sum` is run without subshell + exec + stderr redirect, I think that's fine?
- The helper method `_strip_wildcard_prefix` now returns the original input if it didn't have a wildcard prefix to remove. Without that, the usage with `CERT_DOMAIN` and no `SSL_DOMAIN` filtering caused the logic to hang, likewise if `SSL_DOMAIN` with a wildcard was not stripped. I assume something went wrong with `sed` or `sha512sum` with the `*.`.
- Change detection test increased the amount of service log checked, the amount (3000 bytes) is matching an existing value used elsewhere in that test file which is more than enough.

---

The two tests passed locally, I expect all tests to be good now :+1: 

I'll add the warning for `10.3.0` release not reacting to cert file changes and prep a `10.4.0` release.

target/scripts/helper-functions.sh
```diff
@@ -187,9 +187,9 @@ function _extract_certs_from_acme
 }
 export -f _extract_certs_from_acme
 
-# Remove the `*.` prefix if it exists
+# Remove the `*.` prefix if it exists, else returns the input value
 function _strip_wildcard_prefix {
-  [[ "${1}" == "*."* ]] && echo "${1:2}"
+  [[ "${1}" == "*."* ]] && echo "${1:2}" || echo "${1}"
```
georglauterbach, line 192: More than surprised ShellCheck did not pick this up. IIRC in Bash, this is not an if-then-else. But I will look that up.

georglauterbach (approved):
LGTM 👍 

I will have a look at the if-then-else thing today.

target/scripts/helper-functions.sh
casperklein: The second `echo `is not executed, if the first `echo `fails. That's why you thought, it's not a real if-statement.

I think in that case, the shorthand is fine, as echo is unlikely to fail.

casperklein (approved):

target/scripts/helper-functions.sh
georglauterbach: I guess it's fine in this case, right.

casperklein: I've edited my comment above. It's the opposite of what I wrote 😖

PR #2337, opened 2021-12-20
ci(fix): Do not trust user controlled input

### LINKED ISSUE #2332: [Security] Workflow `docs-preview-deploy.yml` is using vulnerable action `nwtgck/actions-netlify`
The workflow docs-preview-deploy.yml is referencing action nwtgck/actions-netlify using references v1.2. However this reference is missing the commit [223b5b5981680adfe8ec4c9f471620cbbdbfadec](https://github.com/nwtgck/actions-netlify/commits/223b5b5981680adfe8ec4c9f471620cbbdbfadec) which may contain fix to the some vulnerability. 
The vulnerability fix that is missing by actions version could be related to: 
(1) CVE fix 
(2) upgrade of vulnerable dependency 
(3) fix to secret leak and others. 
Please consider to update the reference to the action.

polarathene:
# Description

The prepare workflow runs in an untrusted context already and thus should not have anything worthwhile to exploit.

However care should still be taken to avoid interpolating expressions into shell scripts directly that is data a user can control the value of. Especially to avoid any maintainer referencing an existing workflow from copying a risky snippet unaware of different security contexts for workflows.

In this case, as per Github Documentation and referenced issue comment, the PR title is user controllable data, which if directly interpolated into the shell script being run (as it previously was), allows for injecting commands to execute.

---

Additionally bumps the checkout action from `2.3` to `2.4` (_I only did this for the docs workflows, all the others still keep the checkout action at the current `@v2.3.4`_). Let me know if you want me to handle that here or via separate PR (_I thought we had some automation regarding that?_)

The referenced issue regarding the netlify deployment action for our previews does not have a vulnerability that concerns us, and our version of `1.2` now will also use the just released `1.2.3` update that contains the mentioned commits of interest by the user reporting the vulnerability concern, so we're good on that front regardless :+1: 

Fixes #2332

## Type of change

## Checklist:

### DIFF
--- a/.github/workflows/docs-preview-prepare.yml
+++ b/.github/workflows/docs-preview-prepare.yml
@@ -26,7 +26,7 @@ jobs:
       NETLIFY_SITE_PREFIX: pullrequest-${{ github.event.pull_request.number }}
       NETLIFY_SITE_NAME: dms-doc-previews
     steps:
-      - uses: actions/checkout@v2.3.4
+      - uses: actions/checkout@v2.4
 
       - name: 'Build with mkdocs-material via Docker'
         working-directory: docs
@@ -53,12 +53,17 @@ jobs:
       # Minimize risk of upload failure by bundling files to a single compressed archive (tar + zstd).
       # Bundles build dir and env file into a compressed archive, nested file paths will be preserved.
       - name: 'Prepare artifact for transfer'
+        env:
+          # As a precaution, reference this value by an interpolated ENV var;
+          # instead of interpolating user controllable input directly in the shell script..
+          # https://github.com/docker-mailserver/docker-mailserver/issues/2332#issuecomment-998326798
+          PR_TITLE: ${{ github.event.pull_request.title }}
         run: |
           # Save ENV for transfer
           {
             echo "PR_HEADSHA=${{ github.event.pull_request.head.sha }}"
             echo "PR_NUMBER=${{ github.event.pull_request.number }}"
-            echo "PR_TITLE=${{ github.event.pull_request.title }}"
+            echo "PR_TITLE=${PR_TITLE}"
             echo "NETLIFY_SITE_PREFIX=${{ env.NETLIFY_SITE_PREFIX }}"
             echo "BUILD_DIR=${{ env.BUILD_DIR }}"
           } >> pr.env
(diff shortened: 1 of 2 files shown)


.github/workflows/docs-preview-prepare.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       NETLIFY_SITE_PREFIX: pullrequest-${{ github.event.pull_request.number }}
       NETLIFY_SITE_NAME: dms-doc-previews
     steps:
-      - uses: actions/checkout@v2.3.4
+      - uses: actions/checkout@v2.4
```
polarathene, line 29:
```suggestion
      - uses: actions/checkout@v2.4.0
```

.github/workflows/docs-production-deploy.yml
```diff
@@ -26,7 +26,7 @@ jobs:
     name: 'Deploy Docs'
     runs-on: ubuntu-20.04
     steps:
-      - uses: actions/checkout@v2.3.4
+      - uses: actions/checkout@v2.4
```
polarathene, line 29:
```suggestion
      - uses: actions/checkout@v2.4.0
```

.github/workflows/docs-production-deploy.yml
```diff
@@ -75,10 +75,10 @@ jobs:
     needs: deploy
     steps:
       - name: 'Checkout the tagged commit (shallow clone)'
-        uses: actions/checkout@v2.3.4
+        uses: actions/checkout@v2.4
```
polarathene, line 78:
```suggestion
        uses: actions/checkout@v2.4.0
```

.github/workflows/docs-production-deploy.yml
```diff
@@ -75,10 +75,10 @@ jobs:
...
 
       - name: 'Checkout the docs deployment branch to a subdirectory'
-        uses: actions/checkout@v2.3.4
+        uses: actions/checkout@v2.4
```
polarathene, line 81:
```suggestion
        uses: actions/checkout@v2.4.0
```

polarathene (commented):
Fixing the version since only `@v2` or `@v2.4.0` is accepted.

georglauterbach (approved):

georglauterbach:
> Additionally bumps the checkout action from 2.3 to 2.4 (I only did this for the docs workflows, all the others still keep the checkout action at the current @v2.3.4). Let me know if you want me to handle that here or via separate PR (I thought we had some automation regarding that?)

Yes, this would not have been necessary, but it does not hurt either. Dependabot should sooner or later pick it up and open a PR to update these actions :)

---

LGTM 👍🏼

wernerfred (approved):
> I thought we had some automation regarding that?

Dependabot checks for changes once a week i think, so this change will follow for the other files in at least 7 days.

Issue #1927, opened 2021-04-21
550 5.7.1 Service unavailable; client [xxx.xxx.xxx.xxx] blocked using zen.spamhaus.org


ghnp5:
Today I got an absolutely ridiculous problem, after noticing that I've not been receiving emails at all, since the morning...

I decided to do a test, by sending an email to my mailbox, from an external address, and here is the result:

> 550 5.7.1 Service unavailable; client [xxx.xxx.xxx.xxx] blocked using zen.spamhaus.org

After investigating, I see the problem is similar to #1781

And gladly, someone's comment drove me to the solution, which is to overwrite the following configs in "postfix-main.cf":

`postscreen_dnsbl_sites`
`smtpd_recipient_restrictions`

Removing the offender instructions for `zen.spamhaus.org`.

Now, I have a few questions:

1) What does it mean that emails are "blocked using zen.spamhaus.org" ?
Does this mean my domain/IP was blocked somehow?
Is this something that only affects me, or nothing to do with "me" ("my domain") ?

2) Why are these "sites" in my configurations? Who decides to put these sites as default?

I'm just **very concerned** with the fact that these 3rd party sites, that I never heard about before, can so easily "break" the receiving of emails on my server...

3) Is there a way to recover all the emails that were lost during the day?

This is just absolutely ridiculous...

I'm not blaming you at all. I understand you don't control "zen.spamhaus.org".
But I definitely want to prevent this from happening in the future.

I'm on the latest 7.x version.

Thank you.

casperklein:
> What does it mean that emails are "blocked using zen.spamhaus.org" ?  Does this mean my domain/IP was blocked somehow?

The sending system is on a block list.

> Why are these "sites" in my configurations? Who decides to put these sites as default?

They were added by this project in the past and are not part of the postfix default configuration.

> I'm just very concerned with the fact that these 3rd party sites, that I never heard about before, can so easily "break" the receiving of emails on my server...

I have the same opinion. I don't like the fact, that mails are silently blocked/discard. I want to receive all mails, but marked as spam. Thats why I also disabled all DNS block lists and use Spamassassin only.

fyi: The same goes for Antivirus (clamav). If an email is considered harmful (e.g. containing "dangerous" attachments), it will be discarded.

ghnp5:
Thank you very much.

> The sending system is on a block list.

[...]

PR #2342, opened 2021-12-25
Introduce ENABLE_DNSBL env

casperklein:
# Description

This PR makes the usage of the postfix/postscreen DNS block lists optional.

Fixes issues like #1927

## Type of change

## Checklist:

### DIFF
--- a/docs/content/config/environment.md
+++ b/docs/content/config/environment.md
@@ -116,8 +116,8 @@ In the majority of cases, you want `letsencrypt` or `manual`.
 - **empty** => SSL disabled.
 - letsencrypt => Support for using certificates with _Let's Encrypt_ provisioners. (Docs: [_Let's Encrypt_ Setup][docs-tls-letsencrypt])
 - manual => Provide your own certificate via separate key and cert files. (Docs: [Bring Your Own Certificates][docs-tls-manual])
-  - Requires: `SSL_CERT_PATH` and `SSL_KEY_PATH` ENV vars to be set to the location of the files within the container.
-  - Optional: `SSL_ALT_CERT_PATH` and `SSL_ALT_KEY_PATH` allow providing a 2nd certificate as a fallback for dual (aka hybrid) certificate support. Useful for ECDSA with an RSA fallback. _Presently only `manual` mode supports this feature_.
+    - Requires: `SSL_CERT_PATH` and `SSL_KEY_PATH` ENV vars to be set to the location of the files within the container.
+    - Optional: `SSL_ALT_CERT_PATH` and `SSL_ALT_KEY_PATH` allow providing a 2nd certificate as a fallback for dual (aka hybrid) certificate support. Useful for ECDSA with an RSA fallback. _Presently only `manual` mode supports this feature_.
 - custom => Provide your own certificate as a single file containing both the private key and full certificate chain. (Docs: `None`)
 - self-signed => Provide your own self-signed certificate files. Expects a self-signed CA cert for verification. **Use only for local testing of your setup**. (Docs: [Self-Signed Certificates][docs-tls-selfsigned])
 
--- a/docs/content/config/environment.md
+++ b/docs/content/config/environment.md
@@ -64,9 +64,10 @@ Amavis content filter (used for ClamAV & SpamAssassin)
 
 ##### ENABLE_DNSBL
 
-This enables the [zen.spamhaus.org](https://www.spamhaus.org/zen/) DNS block list.
+This enables the [zen.spamhaus.org](https://www.spamhaus.org/zen/) DNS block list in postfix
+and various [lists](https://github.com/docker-mailserver/docker-mailserver/blob/f7465a50888eef909dbfc01aff4202b9c7d8bc00/target/postfix/main.cf#L58-L66) in postscreen.
 
-Note: Emails will be rejected, if they don't pass the spamhaus check!
+Note: Emails will be rejected, if they don't pass the block list checks!
 
 - **0** => DNS block list is disabled
 - 1     => DNS block list is enabled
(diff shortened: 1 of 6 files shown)


docs/content/config/environment.md
```diff
@@ -62,6 +62,15 @@ Amavis content filter (used for ClamAV & SpamAssassin)
...
 - 3/4/5    => log debug information (very verbose)
 
+##### ENABLE_DNSBL
+
+This enables the [zen.spamhaus.org](https://www.spamhaus.org/zen/) DNS block list.
+
+Note: Emails will be rejected, if they don't pass the spamhaus check!
+
+- **0** => DNS block list is disabled
+- 1     => DNS block list is enabled
```
georglauterbach, line 72: We _could_ make `1` the default, and with an upcoming version, that is not a patch version, change the default. This would provide a "grace period" until, let's say, `v10.5` is released. We would therefore circumvent a real change with a patch version. What do you think?

georglauterbach:
Not sure whether Postscreen should be affected by this too or not, but you could think about these lines in Postfix's configuration as well:

https://github.com/docker-mailserver/docker-mailserver/blob/701037d3b2909db5b268ae4dbc171d70921600c3/target/postfix/main.cf#L58-L66

docs/content/config/environment.md
casperklein: I've not considered this for the patch release, see the milestone label 😉 Patch releases should only contain patches and not introduce new stuff.
Another option would be, to skip the patch release and directly release `10.5`.

casperklein:
Good catch. I've overlooked that, because in my setup `postscreen_dnsbl_sites` is empty due a custom `postfix-main.cf`.

docs/content/config/environment.md
georglauterbach: That is one way, and certainly a valid one. Are the changes / patches enough to justify a new release though?

casperklein: Sorry, I was quiet busy the last days. It depends on how urgent the ldap patch is. If urgent, then a patch release should be made asap. If it's not that critical, we can wait until some more changes are there for a minor update.

georglauterbach:
Just a short question: Why is the variable called `ENABLE_DNSBL` and the function that is executed disables the variables? Wouldn't it make much more sense to adjust the remove the lists from the configuration files and _add_ them if `ENABLE_DNSBL=1`?

casperklein:
It did it this way, because currently there is no DISABLE_VAR. This was just done, to align with the existing namings.

georglauterbach:
> It did it this way, because currently there is no DISABLE_VAR. This was just done, to align with the existing namings.

I'm very much fine with the naming of the variable, just the implementation is logic-wise implemented in exactly the opposite way compared to how I would have done it :)

casperklein:
The result should be the same. Background: I mainly migrated my personal "remove blocklists" stuff from my user-patches. That's why it's not the other way around.

georglauterbach (dismissed):

georglauterbach (dismissed):

docs/content/config/environment.md
```diff
@@ -106,8 +116,8 @@ In the majority of cases, you want `letsencrypt` or `manual`.
 - **empty** => SSL disabled.
 - letsencrypt => Support for using certificates with _Let's Encrypt_ provisioners. (Docs: [_Let's Encrypt_ Setup][docs-tls-letsencrypt])
 - manual => Provide your own certificate via separate key and cert files. (Docs: [Bring Your Own Certificates][docs-tls-manual])
-    - Requires: `SSL_CERT_PATH` and `SSL_KEY_PATH` ENV vars to be set to the location of the files within the container.
-    - Optional: `SSL_ALT_CERT_PATH` and `SSL_ALT_KEY_PATH` allow providing a 2nd certificate as a fallback for dual (aka hybrid) certificate support. Useful for ECDSA with an RSA fallback. _Presently only `manual` mode supports this feature_.
+  - Requires: `SSL_CERT_PATH` and `SSL_KEY_PATH` ENV vars to be set to the location of the files within the container.
+  - Optional: `SSL_ALT_CERT_PATH` and `SSL_ALT_KEY_PATH` allow providing a 2nd certificate as a fallback for dual (aka hybrid) certificate support. Useful for ECDSA with an RSA fallback. _Presently only `manual` mode supports this feature_.
```
polarathene, line 120: Why was the 4 space indent changed to 2 space? 4 space indent is required for the python markdown parser the docs generation tools use. If you view the docs preview you can see this change no longer nests these two items under the `manual` type list item.

Revert.

```suggestion
    - Requires: `SSL_CERT_PATH` and `SSL_KEY_PATH` ENV vars to be set to the location of the files within the container.
    - Optional: `SSL_ALT_CERT_PATH` and `SSL_ALT_KEY_PATH` allow providing a 2nd certificate as a fallback for dual (aka hybrid) certificate support. Useful for ECDSA with an RSA fallback. _Presently only `manual` mode supports this feature_.
```

target/scripts/startup/setup-stack.sh
```diff
@@ -1512,3 +1512,13 @@ function _setup_fail2ban
     echo -e "[Init]\nblocktype = DROP" > /etc/fail2ban/action.d/iptables-common.local
   fi
 }
+
+function _setup_dnsbl_disable
+{
+  _notify 'task' 'Disabling postfix DNS block list (zen.spamhaus.org)'
+  sedfile -i '/^smtpd_recipient_restrictions = / s/, reject_rbl_client zen.spamhaus.org//' /etc/postfix/main.cf
```
polarathene, line 1519: Will this fail similar to the current TLS logic with `sedfile` during a container restart? (_as opposed to recreating the container from scratch_)

May want to just use `sed` or have `sedfile` not fail if no match is found to alter?

test/mail_dnsbl.bats
```diff
@@ -0,0 +1,64 @@
+load 'test_helper/common'
+
+function setup() {
+    run_setup_file_if_necessary
+}
+
+function setup_file() {
+  local PRIVATE_CONFIG CONTAINER CONTAINER2
+  PRIVATE_CONFIG="$(duplicate_config_for_container . "${CONTAINER}")"
```
polarathene, line 9: Does this actually work? `CONTAINER` is declared local var prior but no value assigned, only after this line do you set `CONTAINER` to a string... shouldn't that come before defining `PRIVATE_CONFIG`?

polarathene (changes requested):

docs/content/config/environment.md
casperklein: > python markdown parser the docs generation tools use.

I didn't know that. Markdown linter complaint about it. That's why I've changed it. I will revert it.

target/scripts/startup/setup-stack.sh
casperklein: Yes, this behaves the same as the TLS logic at the moment. I am unsure how we should handle this in general.
Errors on a container restart are nasty. On the other I think there is value in getting to know when this replacement does not take place for some reason on regular container starts.
We can open a separate issue to discuss/vote how to proceed with `sedfile` usage outside of `Dockerfile`.

test/mail_dnsbl.bats
casperklein: This is fine and works. I never saw it the other way. Any example?

polarathene: > Any example?

The lines I've commented on here, and the ones that follow directly after. Is the shell script hoisting or is `${CONTAINER}` at this point empty and misleading in usage?

```bash
  local PRIVATE_CONFIG CONTAINER CONTAINER2
  PRIVATE_CONFIG="$(duplicate_config_for_container . "${CONTAINER}")"
  CONTAINER="mail_dnsbl_enabled"
  CONTAINER2="mail_dnsbl_disabled"
```

polarathene (approved):
Other than concerns raised earlier, LGTM :+1:

georglauterbach (dismissed):

test/mail_dnsbl.bats
casperklein: Now I get it! 😆 

I thought you meant something like this, which confused me:

```
PRIVATE_CONFIG="$(duplicate_config_for_container . "${CONTAINER}")"
local PRIVATE_CONFIG CONTAINER CONTAINER2
```

You are right. CONTAINER should be empty at this point 😞 I'll fix this. Hopefully the last fix in this PR 😆

casperklein: Fixed with the (hopefully) last commit for that PR 😄

georglauterbach (approved):

polarathene:
TLS cipherlist test was failing similar to what @casperklein was experiencing in a separate issue I think, might be an issue with upstream `testssl` image.

polarathene (approved):

Issue #2360, opened 2022-01-06
How to set Logwatch sender address?


craue:
# Subject

I would like some feedback concerning a use case.

## Description

By default, when enabling logwatch and pflogsumm reports, pflogsumm sender address is "mailserver-report@`hostname`.`domainname`" and logwatch sender address is "root@`hostname`.`domainname`". Now I'd like to change both to be sent from "whatever@`domainname`" but couldn't find out how.

I tried various environment variables, but none actually changes the sender e-mail address for the logwatch report ...

- `REPORT_SENDER`: sets sender for pflogsumm only, not logwatch
- `POSTMASTER_ADDRESS`: sets recipient addresses (for pflogsumm and logwatch) only, not sender
- `PFLOGSUMM_SENDER`: sets sender for pflogsumm only, as documented (and expected)
- `LOGWATCH_SENDER`: undocumented and obviously not working (but I wanted to be sure)

So is there a way to set the sender address for logwatch as well?

casperklein:
You can try putting something like this in your [`user-patches.sh`](https://docker-mailserver.github.io/docker-mailserver/edge/faq/#how-to-adjust-settings-with-the-user-patchessh-script) (not tested):

`echo "MailFrom = foo@example.com" >> /etc/logwatch/conf/logfiles/maillog.conf`

craue:
Thank you for the quick response. This led me in the right direction.

`echo "MailFrom = foo@example.com" >> /etc/logwatch/conf/logfiles/maillog.conf` indeed doesn't make a difference.

Taking a look at `/usr/share/logwatch/dist.conf/logwatch.conf` revealed the entry `MailFrom = root`, so I changed this instead and it's working: `sed -i 's|^MailFrom = root$|MailFrom = foo@example.com|' /usr/share/logwatch/dist.conf/logwatch.conf`

Interestingly, other directives put into `/etc/logwatch/conf/logfiles/maillog.conf` are ignored as well, like `Detail = High`. I also needed to add it in `/usr/share/logwatch/dist.conf/logwatch.conf`.

Having an env var at least for the sender address would be much more convenient, though. 😏

georglauterbach:
@craue If you'd like to give it a try, you may regret to #2358 which provided a bit of structure for someone new to the project.

PR #2362, opened 2022-01-06
add env var `LOGWATCH_SENDER`

craue:
# Description

Follow-up to #2360, but using `/etc/logwatch/conf/logwatch.conf` as config file.

## Type of change

This *might* be considered to be a breaking change if someone relied on the fact that `REPORT_SENDER` wouldn't influence the sender address for logwatch reports. But when I started using this project, I was actually expecting it would.

## Checklist:

### DIFF
--- a/docs/content/config/environment.md
+++ b/docs/content/config/environment.md
@@ -266,9 +266,9 @@ Recipient address for pflogsumm reports.
 
 ##### PFLOGSUMM_SENDER
 
-From address for pflogsumm reports.
+Sender address (`FROM`) for pflogsumm reports if pflogsumm reports are enabled.
 
-- **not set** => Use REPORT_SENDER or POSTMASTER_ADDRESS
+- **not set** => Use REPORT_SENDER
 - => Specify the sender address
 
 ##### LOGWATCH_INTERVAL
@@ -288,9 +288,9 @@ Recipient address for logwatch reports if they are enabled.
 
 ##### LOGWATCH_SENDER
 
-From address for logwatch reports if they are enabled.
+Sender address (`FROM`) for logwatch reports if logwatch reports are enabled.
 
-- **not set** => Use REPORT_SENDER or POSTMASTER_ADDRESS
+- **not set** => Use REPORT_SENDER
 - => Specify the sender address
 
 ##### REPORT_RECIPIENT (deprecated)
--- a/docs/content/config/environment.md
+++ b/docs/content/config/environment.md
@@ -286,6 +286,13 @@ Recipient address for logwatch reports if they are enabled.
 - **not set** => Use REPORT_RECIPIENT or POSTMASTER_ADDRESS
 - => Specify the recipient address(es)
 
+##### LOGWATCH_SENDER
+
+From address for logwatch reports if they are enabled.
+
+- **not set** => Use REPORT_SENDER or POSTMASTER_ADDRESS
+- => Specify the sender address
+
 ##### REPORT_RECIPIENT (deprecated)
 
 Enables a report being sent (created by pflogsumm) on a regular basis.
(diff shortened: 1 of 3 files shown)


craue:
I've fixed the issue's description because I just noticed that `POSTMASTER_ADDRESS` wouldn't bubble to `LOGWATCH_SENDER` or `PFLOGSUMM_SENDER`. The documenation for `PFLOGSUMM_SENDER` was already wrong and I just copied it. Shall I fix the comments for both sender vars?

docs/content/config/environment.md
```diff
@@ -286,6 +286,13 @@ Recipient address for logwatch reports if they are enabled.
 - **not set** => Use REPORT_RECIPIENT or POSTMASTER_ADDRESS
 - => Specify the recipient address(es)
 
+##### LOGWATCH_SENDER
+
+From address for logwatch reports if they are enabled.
```
georglauterbach, line 291: I'd rephrase this:

This variable sets the sender address (`FROM`) for logwatch reports if logwatch reports are enabled.

georglauterbach:
> I've fixed the issue's description because I just noticed that `POSTMASTER_ADDRESS` wouldn't bubble to `LOGWATCH_SENDER` or `PFLOGSUMM_SENDER`. The documenation for `PFLOGSUMM_SENDER` was already wrong and I just copied it.

What do you mean by "bubble"?

> Shall I fix the comments for both sender vars?

Sure :)

docs/content/config/environment.md
craue: I also just copied these lines from `PFLOGSUMM_SENDER`. 😏

How about a bit shorter form?
**Sender address (`FROM`) for logwatch reports if logwatch reports are enabled.**

craue:
> > I've fixed the issue's description because I just noticed that `POSTMASTER_ADDRESS` wouldn't bubble to `LOGWATCH_SENDER` or `PFLOGSUMM_SENDER`. The documenation for `PFLOGSUMM_SENDER` was already wrong and I just copied it.
> 
> What do you mean by "bubble"?

I mean that the value of a variable will also set another variable if it's not explicitly set otherwise.

> Sure :)

Will do.

casperklein:
Don't forget to adjust `start-mailserver.sh` as well. As of now, the new env can't be used.

docs/content/config/environment.md
georglauterbach: Fine with me :)

docs/content/config/environment.md
```diff
@@ -286,6 +286,13 @@ Recipient address for logwatch reports if they are enabled.
...
+
+- **not set** => Use REPORT_SENDER or POSTMASTER_ADDRESS
```
georglauterbach, line 293: Maybe we should make sure which one it actually is :)

georglauterbach:
> Don't forget to adjust `start-mailserver.sh` as well. As of now, the new env can't be used.

Isn't it enough that they're set up in `_setup_default_vars` in `setup-stack.sh`?

docs/content/config/environment.md
craue: Only `REPORT_SENDER` is used, at least from what I can tell from the code. I've fixed the docs.

docs/content/config/environment.md
craue: Done.

casperklein:
I think you are right. I find it always confusing, when default vars are defined in two places 😕

georglauterbach:
> I think you are right. I find it always confusing, when default vars are defined in two places confused

Very true. I guess the variables in `_setup_default_vars` need some setup beforehand (like hostname and stuff). That#s why they're located in that specific place.

craue:
Not sure about the failing test. Is it related to my changes?

polarathene (approved):

georglauterbach (approved):

wernerfred (approved):

georglauterbach:
> Not sure about the failing test. Is it related to my changes?
> 
> 
> 
> Edit: Interesting... now the failure is gone. 😮

Sometimes, although not too often, a test is flakey and does not behave well. Restarting the jobs solves these problems.

PR #2364, opened 2022-01-08
Remove sudo from Makefile

casperklein:
# Question

What is the reason for `sudo` here? I think, the user that clones this repo and runs the tests (which create files), is also able to delete those files.


georglauterbach:
I think Docker places configuration file (from the container) and they belong to root - therefore the `sudo`.

polarathene:
That said it does make me a little uneasy, `sudo rm -rf` and one typo with `/` by itself could do damage no? :sweat_smile:

Hopefully not easy to miss with diffs :pray:

georglauterbach:
> That said it does make me a little uneasy, `sudo rm -rf` and one typo with `/` by itself could do damage no? :sweat_smile:
> 
> 
> 
> Hopefully not easy to miss with diffs :pray: 

Modern Linux prevents you from easily `sudo rm -rf /`-ing without having to confirm that you know what you're doing:)

casperklein:
We should add `--no-preserve-root` to satisfy @polarathene's fear. 😆

Issue #2381, opened 2022-01-26
SSL documentation contains a small bug


Twist235:
# Subject

I would like to contribute to the project

## Description
I propose to perform a documentation update:

Section: 
Let's Encrypt (Recommended)
Subsection:
Example

Add these additions to the mailserver service in your docker-compose.yml:

services:
  mailserver:
    # For the FQDN 'mail.example.com':
    hostname: mail
    domainname: example.com
    environment:
      - SSL_TYPE=letsencrypt
    volumes:
      **- /etc/letsencrypt:/etc/letsencrypt**

Change bold part to 
      **- ./docker-data/certbot/certs/:/etc/letsencrypt**

Description of problem:
In this example, a different path is used than in the following Example "using Docker for Let's Encrypt". I propose to use the same path both times. This cost me some time to find, when I follow the examples the docker restarts and I did not understand why.

I would be happy to update it myself (and maybe add some sentences for DNS cert which I used) but I do not have the rights (new to github).

georglauterbach:
Thank you for reporting this.

> I would be happy to update it myself (and maybe add some sentences for DNS cert which I used) but I do not have the rights (new to github).

Please just have a look at [our documentation on how to contribute to this project](https://docker-mailserver.github.io/docker-mailserver/edge/contributing/issues-and-pull-requests/#pull-requests). Quick read and you should be able to open a PR (pull request) which we can then review and merge if appropriate.

PR #2383, opened 2022-01-26
 SSL documentation contains a small bug #2381

Twist235:
Update of example to be more consistent with the next one

# Description

Update of examples

Fixes https://github.com/docker-mailserver/docker-mailserver/issues/2381

## Type of change

## Checklist:

- [x ] My code follows the style guidelines of this project
- [x ] I have performed a self-review of my own code
- [x ] I have commented my code, particularly in hard-to-understand areas
- [x ] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [x ] If necessary I have added tests that prove my fix is effective or that my feature works
- [N/A no functional change ] New and existing unit tests pass locally with my changes

### DIFF
--- a/docs/content/config/security/ssl.md
+++ b/docs/content/config/security/ssl.md
@@ -134,7 +134,8 @@ You don't have to do anything else. Enjoy!
         volumes:
           - ./docker-data/certbot/certs/:/etc/letsencrypt
     ```
-4. The certificate setup is complete, but remember _it will expire_. Consider automating renewals.
+    
+3. The certificate setup is complete, but remember _it will expire_. Consider automating renewals.
 
 !!! tip "Renewing Certificates"
 
--- a/docs/content/config/security/ssl.md
+++ b/docs/content/config/security/ssl.md
@@ -97,7 +97,7 @@ You don't have to do anything else. Enjoy!
         environment:
           - SSL_TYPE=letsencrypt
         volumes:
-          - /etc/letsencrypt:/etc/letsencrypt
+          - ./docker-data/certbot/certs/:/etc/letsencrypt
     ```
 
 #### Example using Docker for _Let's Encrypt_ { data-toc-label='Certbot with Docker' }
--- a/docs/content/config/security/ssl.md
+++ b/docs/content/config/security/ssl.md
@@ -97,7 +97,7 @@ You don't have to do anything else. Enjoy!
         environment:
           - SSL_TYPE=letsencrypt
         volumes:
-          - ./docker-data/certbot/certs/:/etc/letsencrypt
+          - /etc/letsencrypt:/etc/letsencrypt
     ```
 
 #### Example using Docker for _Let's Encrypt_ { data-toc-label='Certbot with Docker' }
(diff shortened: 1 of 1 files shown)


polarathene (commented):
Hey, thanks for taking the time to raise an issue and even contribute a PR to resolve it! Very much appreciated :grinning: 

However in this case there appears to be a misunderstanding, and perhaps we could better adjust the docs to clarify that instead?

---

The path to a local `/etc/letsencrypt` system path is correct. On a Linux system this is where certbot run _locally without docker_ would install certs to, as per the link to certbot docs in that section.

The example was showing how you could mount your local certs to your docker container this way. Whereas the examples that follow after it are provisioned via separate docker containers (one being certbot). I can understand why that might seem confusingly similar and trip a reader up, especially if they're less familiar with either Linux or running certbot outside of a container.

---

That said, since we're a docker focused project, perhaps we don't need to cover this difference and could just focus on the dockerized certbot example that follows?

I did not write the original docs, only revised them. Did part of the confusion also come from the switching between `docker-compose` and the `docker` CLI examples?

There is a feature that [our docs generator provides for tabbed content](https://squidfunk.github.io/mkdocs-material/reference/content-tabs/) ([additional related docs](https://facelessuser.github.io/pymdown-extensions/extensions/tabbed/)), it could probably be updated at a later point to include both, while defaulting to `docker-compose` config? No need to worry about that for this PR, but your input would be appreciated :)

---

**TL;DR:** I am reluctant to merge this given the above context, the change is not correct. I am open towards a revision that better clears up any confusion in our docs however.

Twist235:
I followed the example and used the cerbot docker image, did not store the data at /etc/letsencrypt 

Maybe it could make sense to change the docs to add a placeholder in the local path and explain how this might be changed?
e.g.
volumes:
- <local path>:/etc/letsencrypt

where <local path> is either
/etc/letsencrypt if cerbot is used directly
./docker-data/certbot/certs/ if cerbot docker image is used

And on a different topic:
My setup uses automated letsencrypt DNS challenges because http ports were already in use and forwarding the letsencrypt port via nginx did not work (for sure some config mistake I made).
I could add a short description with links to docs used for my setup, maybe this could also help someone? (of course the DNS provider need to have an API for it)

polarathene:
> Maybe it could make sense to change the docs to add a placeholder in the local path and explain how this might be changed?

I think it'd make more sense just to drop the non-docker certbot example. It's not going to really be lost due to version control in case we ever need to restore it.

> Or we could use the following parameters for the local certbot call

I suppose? I imagine the example was intended for users already using certbot elsewhere and familiar with it, or as a quick option if not using the other provisioners.

I haven't looked through the history for when each guide was introduced to the prior Github Wiki before migration and revising to our current docs generator. Most of the content was community contributed and had fallen out of date, I've trimmed some of it away since.

Should be safe to trim away the non-docker certbot portion, or make it a collapsed by default section.

---

> I could add a short description with links to docs used for my setup, maybe this could also help someone? (of course the DNS provider need to have an API for it)

Contributions are welcome. If you can fit it in nicely for the SSL docs that's great, otherwise we seem to allow general tips/guides in the [examples section](https://docker-mailserver.github.io/docker-mailserver/edge/examples/tutorials/basic-installation/).

Twist235:
I made an update to my comit based on comment from polarathene: For me, it could also make sense to create a second example for the letsencrypt with docker and keep normal letsencrypt as it was before? Or we could also hide the other example or re-order it?

polarathene (approved):
I'm not too fond of the mostly duplicate/repeated example here where the only change is the local folder being mounted into the container.

The initial LE section I think is meant to be less about Certbot and a general overview of what's required to use LE certs (_set the hostname and optional domainname, set the `SSL_TYPE` env and mount your certs to the expected location_).

After that were examples of popular LE provisioners (not too sure about the NAS one).

---

If you feel this reduces confusion, I'm happy to approve it. I've been short of time lately, but I do recall my backlog tasks cover refactoring our SSL/TLS support to simplify configuration, at which point I'll revise the docs again too.

georglauterbach (approved):

georglauterbach:
Oops, figured this is fine to merge because of the approval... I will take care of the rendering if necessary to clean up my premature merge.

PR #2389, opened 2022-02-01
`README.md` and Documentation Update

georglauterbach:
# Description

Follow up on #2383 and update of the `README.md`. Please refer to the commit messages for further information on the "why".

## Type of change

- [x] This change _is_ a documentation update

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,8 +77,7 @@ All workflows are using the tagging convention listed below. It is subsequently 
 
 Since Docker Mailserver `v10.2.0`, **`setup.sh` functionality is included within the container image**. The external convenience script is no longer required if you prefer using `docker exec <CONTAINER NAME> setup <COMMAND>` instead. **If you're new to `docker-mailserver`**, it is recommended to use the script `setup.sh` for convenience.
 
-
-``` BASH 
+``` BASH
 DMS_GITHUB_URL='https://raw.githubusercontent.com/docker-mailserver/docker-mailserver/master'
 wget "${DMS_GITHUB_URL}/docker-compose.yml"
 wget "${DMS_GITHUB_URL}/mailserver.env"
@@ -100,7 +99,7 @@ chmod a+x ./setup.sh
 
 #### First Things First
 
-**Use `docker-compose up / down`, not `docker-compose start / stop`**. Otherwise, the container is not properly destroyed and you may experience problems during startup because of inconsistent state. 
+**Use `docker-compose up / down`, not `docker-compose start / stop`**. Otherwise, the container is not properly destroyed and you may experience problems during startup because of inconsistent state.
 
 You are able to get a full overview of how the configuration works by either running:
 
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ If you have issues, read the full `README` **and** the [documentation][documenta
 - [LetsEncrypt](https://letsencrypt.org/) and self-signed certificates
 - [Setup script](https://docker-mailserver.github.io/docker-mailserver/edge/config/setup.sh) to easily configure and maintain your mail-server
 - Basic [Sieve support](https://docker-mailserver.github.io/docker-mailserver/edge/config/advanced/mail-sieve) using dovecot
-- SASLauthd with LDAP auth
+- SASLauthd with LDAP auth (please see the note [down below](#ldap-setup))
 - Persistent data and state
 - [CI/CD](https://github.com/docker-mailserver/docker-mailserver/actions)
 - [Extension Delimiters](http://www.postfix.org/postconf.5.html#recipient_delimiter) (`you+extension@example.com` go to `you@example.com`)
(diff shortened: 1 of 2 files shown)


wernerfred (dismissed):
LGTM

README.md
```diff
@@ -243,7 +256,9 @@ services:
     restart: always
 ```
 
-#### LDAP setup
+### LDAP Setup
+
+**Note** There are currently no LDAP maintainers mainting LDAP related code. If you encounter issues, please rease them in the issue tracker, but be aware that the core maintainers team will most likely not be able to help you. **We would appreciate and we encourage people to actively participate in maintaining LDAP-related code by becoming a maintainer!**
```
casperklein, line 261:
```suggestion
**Note** There are currently no LDAP maintainers. If you encounter issues, please raise them in the issue tracker, but be aware that the core maintainers team will most likely not be able to help you. **We would appreciate and we encourage everyone to actively participate in maintaining LDAP-related code by becoming a maintainer!**
```

README.md
```diff
@@ -158,12 +171,12 @@ If you'd like to change, patch or alter files or behavior of `docker-mailserver`
...
+**Make sure to read the [CHANGELOG](https://github.com/docker-mailserver/docker-mailserver/blob/master/CHANGELOG.md)** before updating to new versions, to be prepared for possible breaking changes.
 
-``` BASH
-docker-compose pull
-docker-compose down
-docker-compose up -d mailserver
+``` CONSOLE
+$ docker-compose pull
+$ docker-compose down
+$ docker-compose up -d mailserver
```
casperklein, line 179: 1. Whats the difference between using BASH / CONSOLE here? I've never seen CONSOLE used before.
2. Adding `$` here makes easy copy/paste harder.

polarathene: I think it might have been to distinguish between commands and output, it's just for syntax highlighting, so bash is probably fine too.

README.md
```diff
@@ -110,13 +100,30 @@ If you're using `docker-mailserver` version `v10.1.x` or below, you will need to
 
 #### First Things First
 
-**Use `docker-compose up / down`, not `docker-compose start / stop`**. Otherwise, the container is not properly destroyed and you may experience problems during startup because of inconsistent state.
-
-You are able to get a full overview of how the configuration works by either running:
```
polarathene, line 115: Why was the line break removed? The two sentences aren't exactly related and the 2nd sentence has an association to the ordered list that follows.

README.md
```diff
@@ -126,30 +133,36 @@ On first start, you will likely see an error stating that there are no mail acco
...
-# and configure aliases, DKIM and more
-./setup.sh [-Z] alias add postmaster@<domain> <user@domain>
-./setup.sh [-Z] config dkim
-```
+``` CONSOLE
+$ docker-compose up -d mailserver
 
-In case you're using LDAP, the setup looks a bit different as you do not add user accounts directly. Postfix doesn't know your domain(s) and you need to provide it when configuring DKIM:
+$ # you may add some more users
+$ # for SELinux, use -Z
```
polarathene, line 140: This is a bit weird, in that while it's not output from the previous command, it's a bit weird annotating as a command with `$` as well (which is included in the rendered markdown still).

This also occurs elsewhere in the fenced blocks that have switched to `CONSOLE`.

docs/content/config/security/ssl.md
```diff
@@ -119,41 +118,41 @@ You don't have to do anything else. Enjoy!
...
+        # This will need access to port 443 from the internet, adjust your firewall if needed.
+        docker run --rm -it \
+          -v "${PWD}/docker-data/certbot/certs/:/etc/letsencrypt/" \
+          -v "${PWD}/docker-data/certbot/logs/:/var/log/letsencrypt/" \
+          -p 80:80 \
+          -p 443:443 \
+          certbot/certbot renew
+        ```
+    
+        This process can also be [automated via _cron_ or _systemd timers_][certbot::automated-renewal].
```
polarathene, line 155: While similar, this admonition is not associated to the third ordered list item prior to it.

This is discussing renewal in general with a brief mention at the end about automating the otherwise manually invoked renewal.

The ordered list item is only about advising to consider automating renewal. Although I guess you could associate the admonition to it..?

---

The admonition prior was indented as it's associated via an example, rather than a tip.

polarathene (changes requested):

README.md
georglauterbach: `CONSOLE` is for displaying "an interactive shell", i.e. when you want to show to people that these commands are to be run in a shell. I disagree that this makes copy-pasting harder - someone who cannot copy a line properly (be it here or elsewhere in the docs) should probably not operate a mail server :D

README.md
georglauterbach: Good catch - thanks!

README.md
georglauterbach: I will have a look and revert it :)

README.md
casperklein: > someone who cannot copy a line properly

I didn't say it's not possible or so, but it makes it harder. You cannot copy / paste these three lines at once. You'll have to paste them somewhere else and remove the `$` or you'll have to copy each line alone.
No big deal in the end, but I would remove them. Leave it as it is if you disagree, no reason to discuss it to death 😉

georglauterbach: I will revert to `BASH` in this case, as there is no output from the individual commands. The rendering with `CONSOLE` shows a clear distinction between the command and the output, which is preferred in some situations - but not in this one particularly.

README.md
georglauterbach: I agree. Will see what I can do

docs/content/config/security/ssl.md
georglauterbach: Alright. Figured this belongs together.

casperklein (approved):

polarathene (approved):
LGTM :+1: 

Nice improvement to the README and thanks for fixing the docs rendering issue :grinning:

PR #2393, opened 2022-02-05
Rootless Podman security update

### LINKED ISSUE #2377: Security concerns using podman in rootless mode

### Affected Component(s)

Podman creates open postfix relay

```Markdown
As you state in your [mailserver.env](https://github.com/docker-mailserver/docker-mailserver/blob/b2b701a4786f91fe266cca4d077a088345fdff5d/mailserver.env#L53), the `network` option might open up an open relay when specifying the wrong network. The default is `localhost`.

When running Podman in rootless mode, all source IP's get translated to localhost by default, so the relay is publicly accessible to the whole WAN side and there is no mention/warning about this anywhere in the wiki.
```

```Markdown
I expected this to be an `out-of-the-box secure` solution. I did not know about Podmans behaviour earlier and I think users should be warned or ideally the default config should be updated to secure Podman in rootless mode.
```

```Markdown
1. Launch DMS with Podman in rootless mode
2. Expose port 25
3. The gateway is now open to the public -> Watch your mailserver get used as a spam-distributor :)
```

### DMS version

v10.4.0

Linux

x86_64 / AMD64

Podman (not officially supported)

### docker-compose.yml

```yaml
It's the default config provided by the master branch
```

### Relevant log output

_No response_

```Markdown
This was tested on AlmaLinux 8.5 using Podman 3.3.1 in rootless mode with podman-compose v1.0.2.
Unprivileged port bindings start at 25, I have not configured firewalld redirects to unprivileged ports (but I guess this has no impact on the issue?)
```

- [X] I am inexperienced with mail servers

### Code of conduct

_No response_

p-fruck:
# Description

This PR includes security related documentation to run docker-mailserver with podman in rootless mode. See the linked issue for more details on the security concerns.

Personally I would like to set `PERMIT_DOCKER=none` to the new default in `mailserver.env` because it enforces authentication even from localhost and therefore is more secure, but this would result in a breaking change for users that have authentication-less setups for their localhost. The decision is up to you :)

Fixes #2377

## Type of change

## Checklist:

### DIFF
--- a/mailserver.env
+++ b/mailserver.env
@@ -46,9 +46,12 @@ UPDATE_CHECK_INTERVAL=1d
 # **WARNING**: Adding the docker network's gateway to the list of trusted hosts, e.g. using the `network` or
 # `connected-networks` option, can create an open relay
 # https://github.com/docker-mailserver/docker-mailserver/issues/1405#issuecomment-590106498
-# empty => localhost only
-# host => Add docker host (ipv4 only)
-# network => Add all docker containers (ipv4 only)
+# The same can happen for rootless podman. To prevent this, set the value to "none" or configure slirp4netns
+# https://github.com/docker-mailserver/docker-mailserver/issues/2377
+# <empty> => localhost and container ip only
+# none => Explicitly force authentication
+# host => Add docker container network (ipv4 only)
+# network => Add all docker container networks (ipv4 only)
 # connected-networks => Add all connected docker networks (ipv4 only)
 PERMIT_DOCKER=
 
--- a/mailserver.env
+++ b/mailserver.env
@@ -48,7 +48,7 @@ UPDATE_CHECK_INTERVAL=1d
 # https://github.com/docker-mailserver/docker-mailserver/issues/1405#issuecomment-590106498
 # The same can happen for rootless podman. To prevent this, set the value to "none" or configure slirp4netns
 # https://github.com/docker-mailserver/docker-mailserver/issues/2377
-# <empty> => localhost and container ip only
+# <empty> => container ip only
 # none => Explicitly force authentication
 # host => Add docker container network (ipv4 only)
 # network => Add all docker container networks (ipv4 only)
--- a/target/scripts/startup/setup-stack.sh
+++ b/target/scripts/startup/setup-stack.sh
@@ -1155,7 +1155,7 @@ function _setup_docker_permit
 
   case "${PERMIT_DOCKER}" in
     "none" )
-      _notify 'inf' "Clearing my networks"
+      _notify 'inf' "Clearing Postfix's 'mynetworks'"
       postconf -e "mynetworks ="
       ;;
 
--- a/target/scripts/startup/setup-stack.sh
+++ b/target/scripts/startup/setup-stack.sh
@@ -1154,6 +1154,11 @@ function _setup_docker_permit
   done < <(ip -o -4 addr show type veth | grep -E -o '[0-9\.]+/[0-9]+')
 
   case "${PERMIT_DOCKER}" in
+    "none" )
+      _notify 'inf' "Clearing my networks"
+      postconf -e "mynetworks ="
+      ;;
+
     "host" )
       _notify 'inf' "Adding ${CONTAINER_NETWORK}/16 to my networks"
       postconf -e "$(postconf | grep '^mynetworks =') ${CONTAINER_NETWORK}/16"
--- a/README.md
+++ b/README.md
@@ -95,6 +95,8 @@ chmod a+x ./setup.sh
    - [_only_ basic `VAR=VAL`](https://docs.docker.com/compose/env-file/) is supported (**do not** quote your values!)
    - variable substitution is **not** supported (e.g. :no_entry_sign: `OVERRIDE_HOSTNAME=$HOSTNAME.$DOMAINNAME` :no_entry_sign:)
 
+**Note:** If you're using podman, make sure to read the related [documentation](https://docker-mailserver.github.io/docker-mailserver/edge/config/advanced/podman/)
+
 ### Get up and running
 
 #### First Things First
--- a/docs/content/config/advanced/podman.md
+++ b/docs/content/config/advanced/podman.md
@@ -14,6 +14,11 @@ Podman is a daemonless container engine for developing, managing, and running OC
 
     This guide was tested with Fedora 34 using `systemd` and `firewalld`. Moreover, it requires Podman version >= 3.2. You may be able to substitute `dnf` - Fedora's package maneger - with others such as `apt`.
 
+!!! warning "About Security"
+
+    Running podman in rootless mode requires additional modifications in order to keep your mailserver secure.
+    Make sure to read the related documentation.
+
 ## Installation in Rootfull Mode
 
 While using Podman, you can just manage docker-mailserver as what you did with Docker. Your best friend `setup.sh` includes the minimum code in order to support Podman since it's 100% compatible with the Docker CLI.
@@ -56,6 +61,7 @@ Running rootless containers is one of Podman's major features. But due to some r
 
 - a rootless container is running in a user namespace so you cannot bind ports lower than 1024
 - a rootless container's systemd file can only be placed in folder under `~/.config`
+- a rootless container can result in an open relay, make sure to read the [security section](#security-in-rootless-mode).
 
 Also notice that Podman's rootless mode is not about running as a non-root user inside the container, but about the mapping of (normal, non-root) host users to root inside the container.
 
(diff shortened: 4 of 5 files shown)


georglauterbach:
I recently merged a PR that changed the README. You'll need to resolve the (small) conflict there. I'd also like you to relocate the information in the documentation, at the end of the `### Create a docker-compose Environment` section, above the `### Get up and running` heading. I think this place is more fitting.

p-fruck:
I've rebased to the latest master branch, missed the last commit :sweat_smile:

casperklein:
Not much time right now, so just two quick thoughts:

1. This could break the update-check https://github.com/docker-mailserver/docker-mailserver/blob/4cb617f7f90bc8716601f2a07cbae8a001da30e0/target/scripts/update-check.sh#L46
2. You might want to add a test for this in https://github.com/docker-mailserver/docker-mailserver/blob/master/test/permit_docker.bats

target/scripts/startup/setup-stack.sh
```diff
@@ -1154,6 +1154,11 @@ function _setup_docker_permit
   done < <(ip -o -4 addr show type veth | grep -E -o '[0-9\.]+/[0-9]+')
 
   case "${PERMIT_DOCKER}" in
+    "none" )
+      _notify 'inf' "Clearing my networks"
```
georglauterbach, line 1158: I'd rather use `_notify 'inf' "Clearing Postfix's 'mynetworks'"

georglauterbach (commented):
Other than the one comment, @casperklein mentioned the update check. If these two concerns are resolved, I think this PR is fine :D

p-fruck:
I've applied the suggestion and added a simply testcase. @casperklein you mentioned the issue with the updatescript, how should we overcome that? Authenticate against the server using credentials supplied as environment variables, or simply output to the terminal and leave it up to the user to handle the output if `PERMIT_DOCKER` is explicitly set to `none`?

mailserver.env
```diff
@@ -46,9 +46,12 @@ UPDATE_CHECK_INTERVAL=1d
 # **WARNING**: Adding the docker network's gateway to the list of trusted hosts, e.g. using the `network` or
 # `connected-networks` option, can create an open relay
 # https://github.com/docker-mailserver/docker-mailserver/issues/1405#issuecomment-590106498
-# empty => localhost only
-# host => Add docker host (ipv4 only)
-# network => Add all docker containers (ipv4 only)
+# The same can happen for rootless podman. To prevent this, set the value to "none" or configure slirp4netns
+# https://github.com/docker-mailserver/docker-mailserver/issues/2377
+# <empty> => localhost and container ip only
```
casperklein, line 51:
```suggestion
# <empty> => container ip only
```

casperklein:
I revoked my comment about the update-check (see my edited comment above). But to be sure, you can test it with something like this:

```bash
docker exec -it mailserver bash -c 'echo "10.0.0" > /VERSION && supervisorctl restart update-check'
```

p-fruck:
I see, just did a quick test and it works as you described, thanks :+1:

casperklein (approved):
Excellent first contribution 👌

wernerfred (approved):
LGTM, thanks

georglauterbach (approved):

casperklein:
Two more things that came to my mind are `LOGWATCH` and `PFLOGSUMM`. Do you have a config like that and can check if you still get those mails daily?

```
- LOGWATCH_INTERVAL=daily
- LOGWATCH_RECIPIENT=foo@example.com
- PFLOGSUMM_TRIGGER=daily_cron
- PFLOGSUMM_RECIPIENT=foo@example.com
```

p-fruck:
I don't have at the moment, but I could set it up within the next few days if you like

casperklein:
That would be great.

casperklein:
@docker-mailserver/maintainers Is there a way to detect within the container, if it's run by podman? If that is possible, we should print a open-relay warning, if `PERMIT_DOCKER` is empty.

We could also make `PERMIT_DOCKER=none` as default in the future to provide a more secure out-of-the-box solution, despite podman isn't officially supported. But then `LOGWATCH` and `PFLOGSUMM` needs to be adjusted to use authentication (if neccesary, let's wait for p-frucks feedback).

wernerfred:
Just did a short Research and found the following:

The best check If the runtime being Docker or Podman would be to look for /run/.containerenv (Podman) vs /run/.dockerenv (Docker)

Haven't tested it myself though

polarathene:
> .dockerenv

That is old advice IIRC, I remember looking into detecting if operating in a Docker container a while back and that approach no longer works (it was never officially documented by Docker either) as the file is no longer present AFAIK.

The easiest approach of course would be to just add a `podman` prefix/suffix tag variant that sets an ENV to detect, or very early on in README/docs explictly direct Podman users to a relevant docs page that details anything they should be aware of (might be missed if following third-party guides elsewhere though).

Changing defaults is probably fine too, but should wait until a major release right?

casperklein:
>  as the file is no longer present AFAIK.

confirm 😞

polarathene:
There was some other approaches I recall being advised, one regarding cgroups but it was flakey IIRC (_there's v1 and v2, along with differences between container runtimes and possibly kernel versions I think.._) and was specifically for detecting running in a containerized environment.

If Podman does still make that file available, that's all you'd really need I guess to pull off the detection, but if it's not officially documented anywhere, it might disappear one day like the Docker one did. We could also just add a `PODMAN=1` ENV or similar for users running podman to use, but that seems kinda redundant since awareness of our docs for podman (_with `PERMIT_DOCKER` warning added_) would be just as effective.

p-fruck:
In case you want to consider a rather hacky solution, you might give this command a try
```sh
if grep 'host.containers.internal' /etc/hosts > /dev/null; then echo podman; else echo docker; fi
```

Also I just confirmend @polarathene's guess that podman (tested with `v3.4.4`) still creates the `/run/.containerenv` file

polarathene:
When running with host mode networking, at least with Docker it will use the host system `/etc/hosts` instead IIRC, so that approach may not be reliable for some users (maybe podman handles that differently though?).

p-fruck:
You are right, running podman in host mode does not add the `host.containers.internal` entry to the hosts file. On the other hand I'm not sure if the IP rebinding issues exist when running the container in host mode, because from what I understood the IP rebinding happens in podmans rootless overlay network, which is unused in host mode. Either way the `/run/.containerenv` might be the more reliable option

p-fruck:
Another think that came to my mind during the setup of my new mailserver: Do you think it's reliable to check for the present network interfaces? That way also the default network interface could be dynamically changed and podman users would not have to specify `NETWORK_INTERFACE=tap0` in their environment file.

Regarding the `LOGWATCH` and `PFLOGSUMM` checks: I need to wait for my current email provider to update their DNS cache so I can send mails to my other address :unamused: How would you test such a setup? Do you have two local mail servers that communicate with each other?

casperklein:
I am not sure if I understand you correctly. Why would you need a second DMS instance?

Another podman question: Can you confirm, that [this](https://github.com/docker-mailserver/docker-mailserver/issues/2411#issuecomment-1043299434) solution does also prevent an open-relay?
That might be a solution, if `LOGWATCH` doesn't work anymore with ``PERMIT_DOCKER=none` set.

p-fruck:
> people not reading the documentation properly

One of them being me, I guess :sweat_smile: 

> Why would you need a second DMS instance?

In case I wanted to test inter server communication. But you are right, it's not required for the current usecase

> Can you confirm, that https://github.com/docker-mailserver/docker-mailserver/issues/2411#issuecomment-1043299434 solution does also prevent an open-relay

I can, but I didn't expect the result. Using the [slirp4netns](https://docker-mailserver.github.io/docker-mailserver/edge/config/advanced/podman/#use-the-slip4netns-network-driver) approach (compatible with v10.4.0 and older) I get a  client host rejected message:
```
HELO example.com
250 mail.example.com
mail from: admin@example.com          
250 2.1.0 Ok
rcpt to: a.b@c.de
554 5.7.1 <mail.example.com[10.88.2.2]>: Client host rejected: Access denied
```

The solution shown in die issue produces a relay access denied error:
```
HELO example.com
220 mail.example.com ESMTP
250 mail.example.com
mail from: admin@example.com
250 2.1.0 Ok
rcpt to: a.b@c.de
554 5.7.1 <a.b@c.de>: Relay access denied
```

Not what I've expected, but both solutions seem to work

casperklein:
> We could also make PERMIT_DOCKER=none as default in the future to provide a more secure out-of-the-box solution, despite podman isn't officially supported. But then LOGWATCH and PFLOGSUMM needs to be adjusted to use authentication (if neccesary, let's wait for p-frucks feedback).

I did my own tests with `PERMIT_DOCKER=none` and the mails from `LOGWATCH & PFLOGSUMM` were send successful.

PR #2361, opened 2022-01-06
consistently make 1 the default value for `SPAMASSASSIN_SPAM_TO_INBOX`

craue:
According to https://github.com/docker-mailserver/docker-mailserver/blob/0c31f713587ab590af40552a1e3be07be273ff81/mailserver.env#L252 the default value is actually `1`. I switched the order of both lines to have the default on top which seems to be the usual order. Let me know if further changes are needed.

PR #2397, opened 2022-02-07
fix: double occurence of `/etc/postfix/regexp`

### LINKED ISSUE #2394: [BUG]  virtual_alias_maps contains the regexp entry multiple times

### Affected Component(s)

postfix

```Markdown
virtual_alias_maps ends up with multiple `pcre:/etc/postfix/regexp` entries
```

```Markdown
Only a single entry
```

```Markdown
1. Start dms
2. create data/config/postfix-regexp.cf
2. edit the alias file; wait one second
3. edit the alias file; again
```

### DMS version

edge

Linux

x86_64 / AMD64

Docker

### docker-compose.yml

_No response_

### Relevant log output

_No response_

```Markdown
The sed line in `_handle_postfix_regexp_config` that adds the config to `virtual_alias_maps` is not guarded by a check to see if it is already present in the file. 

I think it may be possible to just always include the `pcre:/etc/postfix/regexp` as `_handle_postfix_virtual_config` always creates that file as empty.
```

- [X] I am inexperienced with docker
- [X] I am uncomfortable with the CLI

### Code of conduct

_No response_

georglauterbach:
# Description

`/etc/postfix/regexp` appeared ~~twice~~ in Postfix's `virtual_alias_maps`. This should be corrected now.

PS: It turns out, without this PR, every time a change is detected, `/etc/postfix/regexp` will be added again. So this PR is indeed somewhat important :D

Fixes #2394

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/startup/setup-stack.sh
+++ b/target/scripts/startup/setup-stack.sh
@@ -342,7 +342,7 @@ function _setup_dovecot_local_user
 
 function _setup_ldap
 {
-  _notify 'task' 'Setting up Ldap'
+  _notify 'task' 'Setting up LDAP'
   _notify 'inf' 'Checking for custom configs'
 
   for i in 'users' 'groups' 'aliases' 'domains'
@@ -404,7 +404,7 @@ function _setup_ldap
   # add domainname to vhost
   echo "${DOMAINNAME}" >>/tmp/vhost.tmp
 
-  _notify 'inf' "Enabling dovecot LDAP authentification"
+  _notify 'inf' 'Enabling dovecot LDAP authentification'
 
   sed -i -e '/\!include auth-ldap\.conf\.ext/s/^#//' /etc/dovecot/conf.d/10-auth.conf
   sed -i -e '/\!include auth-passwdfile\.inc/s/^/#/' /etc/dovecot/conf.d/10-auth.conf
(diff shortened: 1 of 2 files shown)


georglauterbach:
Seems like I broke something. I will have a look again.

target/scripts/startup/setup-stack.sh
```diff
@@ -426,7 +426,7 @@ function _setup_ldap
   if [[ -f /etc/postfix/ldap-aliases.cf ]] && [[ -f /etc/postfix/ldap-groups.cf ]]
   then
     postconf -e "virtual_alias_maps = ldap:/etc/postfix/ldap-aliases.cf, ldap:/etc/postfix/ldap-groups.cf" || \
-    _notify 'inf' "==> Warning: /etc/postfix/ldap-aliases.cf or /etc/postfix/ldap-groups.cf not found"
+    _notify 'war' "/etc/postfix/ldap-aliases.cf or /etc/postfix/ldap-groups.cf not found"
```
casperklein, line 429: I am confused. First there is a check, if the two files exist. But if `postconf` fails, a warning that one of those files is missing is printed 😮

georglauterbach: Haha, I just noticed that it is weird that `notify 'inf' 'warning:'` was used, so I changed it to `notify 'war'` instead without looking what is being checked at all 😂 I think this line needs to go into an `else` statement...

casperklein (dismissed):

target/scripts/startup/setup-stack.sh
```diff
@@ -413,20 +413,23 @@ function _setup_ldap
 
   if [[ -f /etc/postfix/ldap-users.cf ]]
   then
-    postconf -e "virtual_mailbox_maps = ldap:/etc/postfix/ldap-users.cf" || \
-    _notify 'inf' "==> Warning: /etc/postfix/ldap-user.cf not found"
+    postconf -e "virtual_mailbox_maps = ldap:/etc/postfix/ldap-users.cf"
+  else
+    _notify 'war' "'/etc/postfix/ldap-user.cf' not found"
```
polarathene, line 418:
```suggestion
    _notify 'war' "'/etc/postfix/ldap-users.cf' not found"
```

polarathene (changes requested):
LGTM :+1: 

Just one typo (not from this PR but that existed before with the warning file path being incorrect).

casperklein (approved):

georglauterbach:
I'd like to release `v10.5.0` when this PR and #2361 are merged :D

wernerfred (approved):
Thx

polarathene (approved):

Issue #2348, opened 2021-12-27
check-for-changes.sh uses a lot of resources


ontheair81:
# Subject

One day after upgrading to `10.4.0` I realized the process `check-for-changes.sh` consumes a lot of memory and cpu resources.
I did not see this behaviour with `10.3.0`.

## Description

The next day after the upgrade the container used 2.5G of memory (this usually was less than 800M) and the four cpu cores have been very busy. After restarting `docker-mailserver` everything was as usual again.

Some hours later I realized that memory and cpu usage was growing again. It seems to be caused by the process `bin/bash /usr/local/bin/check-for-changes.sh`.
After starting `docker-mailserver`, the memory consumption of this process is around 6M, but growing 800k per minute (ca. 50M per hour). The higher the memory usage goes, the higher the cpu load.

## Question

It would be nice to hear if there are other people experiencing the same issue, or if it is just me ;-) Help is very appreciated!
I am not an expert in docker (yet), but I am very willing to learn as much as possible and to provide more information as needed and requested.

Thank you very much in advance!
(I did not label this issue as "bug" as I am not sure if this is a general issue or just happening in my configuration.)

## Environment

Debian Bullseye, latest patches
Let'sEncrypt-certificates are generated from [nginx-reverse-proxy](https://hub.docker.com/r/nginxproxy/nginx-proxy) with [acme-companion](https://hub.docker.com/r/nginxproxy/acme-companion).

``` docker-compose-yml
version: "3.7"

services:

    mailserver:
        image: "docker.io/mailserver/docker-mailserver:10.4.0"
        container_name: "mailserver"
        hostname: "mail"
        domainname: "[removed].de"
        env_file: "/home/user/docker_compose/mailserver/mailserver.env"
        ports:
           - "25:25"     # SMTP  (explicit TLS => STARTTLS)
           - "143:143"   # IMAP4 (explicit TLS => STARTTLS)
           - "465:465"   # ESMTP (implicit TLS)
           - "587:587"   # ESMTP (explicit TLS => STARTTLS)
           - "993:993"   # IMAP4 (implicit TLS)
           - "4190:4190" # ManageSieve
        volumes:
           - "data:/var/mail/"
           - "state:/var/mail-state/"
           - "logs:/var/log/mail/"
           - "config:/tmp/docker-mailserver/"
           - "/etc/localtime:/etc/localtime:ro"
[...]

PR #2398, opened 2022-02-07
relieve some pain: restart `check-for-changes.sh` regularly

georglauterbach:
# Description

The script does not need to run every single second. This is not to say
that there is huge resource usage due to running every second, but every
~~ten~~ two seconds is more than enough. ~~This _should_ act like a plaster when
it comes to #2348 -~~ I KNOW THIS IS **NOT** A FIX. ~~But this way, people
can first of all stay on Debian 11 and DMS 10.4.0, secondly, we may get
some more feedback on this mysterious issue due to people staying in DMS
10.4.0.~~ Restarting the script will provide a new PID and should in theory stop the resource leakage. Not sure whether we want to go through with it though, as this will also disguise the error at hand to some degree.

Independently of #2348, increasing the timeout is a very valid concern I had, and on my deployment, I already raised it.

Fixes nothing, but _should_ increase the time before DMS starts using all the RAM on a system.

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/startup/fixes-stack.sh
+++ b/target/scripts/startup/fixes-stack.sh
@@ -58,5 +58,8 @@ function _fix_cleanup_spamassassin
 function _fix_restart_changedetector_daily
 {
   _notify 'task' 'Making sure the changedetector is restarted daily'
-  echo 'supervisorctl restart changedetector' >/etc/cron.daily/restart_changedetector
+  local CRON_CHANGEDETECTOR_FILE=/etc/cron.daily/restart_changedetector
+
+  printf '#! /bin/bash\n\nsupervisorctl restart changedetector\n' >"${CRON_CHANGEDETECTOR_FILE}"
+  chmod +x "${CRON_CHANGEDETECTOR_FILE}"
 }
--- a/target/scripts/startup/fixes-stack.sh
+++ b/target/scripts/startup/fixes-stack.sh
@@ -54,3 +54,9 @@ function _fix_cleanup_spamassassin
     [[ ! -f /CONTAINER_START ]] && _notify 'err' 'Failed to remove SpamAssassin configuration'
   }
 }
+
+function _fix_restart_changedetector_daily
+{
+  _notify 'task' 'Making sure the changedetector is restarted daily'
+  echo 'supervisorctl restart changedetector' >/etc/cron.daily/restart_changedetector
+}
--- a/target/scripts/check-for-changes.sh
+++ b/target/scripts/check-for-changes.sh
@@ -131,7 +131,7 @@ do
   # mark changes as applied
   mv "${CHKSUM_FILE}.new" "${CHKSUM_FILE}"
 
-  sleep 10
+  sleep 2
 done
 
 exit 0
--- a/target/scripts/check-for-changes.sh
+++ b/target/scripts/check-for-changes.sh
@@ -60,12 +60,7 @@ do
     create_lock # Shared config safety lock
     CHANGED=$(grep -Fxvf "${CHKSUM_FILE}" "${CHKSUM_FILE}.new" | sed 's/^[^ ]\+  //')
 
-    # Bug alert! This overwrites the alias set by start-mailserver.sh
-    # Take care that changes in one script are propagated to the other
-
     # ! NEEDS FIX -----------------------------------------
-    # TODO FIX --------------------------------------------
-    # ! NEEDS EXTENSIONS ----------------------------------
     # TODO Perform updates below conditionally too --------
     # Also note that changes are performed in place and are not atomic
     # We should fix that and write to temporary files, stop, swap and start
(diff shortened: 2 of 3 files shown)


georglauterbach:
@polarathene would providing the one failing test with more time help here, or do you think the issue is somewhere else entirely?

oblitum:
This is probably related to PR #2157.

casperklein:
> Fixes nothing, but should increase the time before DMS starts using all the RAM on a system.

I think a better approach would be, to restart the changedetector service once a day or so.
Your solution does only slow down a potential memory leak, but will not prevent it. Restarting the service should however "fix" it.

georglauterbach:
Good idea. Will `supervisorctl changedetector restart` suffice here?

casperklein:
Yes, I think so. But haven't tested it. If the PID of the script changes afterwards, it works.

casperklein:
BTW: While reviewing the check-for-changes script, I stumbled upon https://github.com/docker-mailserver/docker-mailserver/blob/7b21db77cc5c2723374131a6f9f1ff21fc7f1859/target/scripts/helper-functions.sh#L245-L254

@georglauterbach @polarathene Do you have an idea, why there is a subshell used + `exec` statement? IMO it should work without those too. I cannot imagine a reason for that.

georglauterbach:
> BTW: While reviewing the check-for-changes script, I stumbled upon
> 
> https://github.com/docker-mailserver/docker-mailserver/blob/7b21db77cc5c2723374131a6f9f1ff21fc7f1859/target/scripts/helper-functions.sh#L245-L254
> 
> @georglauterbach @polarathene Do you have an idea, why there is a subshell used + `exec` statement? IMO it should work without those too. I cannot imagine a reason for that.

I see no reason why the subshell is there either... maybe the writer wanted to be able to not run the SHA-sum when he/she could not change directories, but this is an awful way of doing it. I think this can be done in a better way :D

target/scripts/startup/fixes-stack.sh
```diff
@@ -54,3 +54,9 @@ function _fix_cleanup_spamassassin
     [[ ! -f /CONTAINER_START ]] && _notify 'err' 'Failed to remove SpamAssassin configuration'
   }
 }
+
+function _fix_restart_changedetector_daily
+{
+  _notify 'task' 'Making sure the changedetector is restarted daily'
+  echo 'supervisorctl restart changedetector' >/etc/cron.daily/restart_changedetector
```
casperklein, line 61: 1. No shebang line?
2. `restart_changedetector` is not executable?

georglauterbach: 1. I forgot
2. I forgot

Thanks for pointing this out 👌🏽

oblitum:
> I think this can be done in a better way :D

TBH, this spot in the code is the strange part that makes me feel it's the culprit. I wish that, if you folks could refactor this to something less astonishing, we could then test whether that actually removes the issue, before having a PR applied that could daily kill the problem.

georglauterbach:
> > I think this can be done in a better way :D
> 
> 
> 
> TBH, this spot in the code is the strange part that makes me feel it's the culprit. I wish that, if you folks could refactor this to something less astonishing, we could then test whether that actually removes the issue, before having a PR applied that could daily kill the problem.

I will see where I can provide something here unless @casperklein beats me to it :D

casperklein:
>  function _fix_restart_changedetector_daily

One last thing: I think it's a good idea, to add a small comment referencing the issue, so it's clear, why this function exists.
Otherwise it LGTM 👍

oblitum:
Just to be clear, I suspect of sub shell part because it does a full clone of shell script state (without due reason afaik), and this state cloning may be increasing each time it happens, accumulating over previous state (although I couldn't yet realize what could be linearly increasing).

NorseGaud:
> Just to be clear, I suspect of sub shell part because it does a full clone of shell script state (without due reason afaik), and this state cloning might be increasing each time it happens, accumulating over previous state (although I couldn't yet realize what could be linearly increasing), while leaving leftovers behind in the parent shell process. This can explain why both RAM and CPU usage increases linearly.

I think that's a really good point.

NorseGaud:
I've seen exec used instead of eval but both for the same purpose: allowing interpolation of dynamic variables to happen before execution. Likely why because of the array expansion/interpolation happening. Though, it should work just fine without exec afaik

georglauterbach:
I will close this in favor of #2401.

polarathene:
Just to add to the discussion. It was maybe related to syscall change with the newer Debian image, perhaps not due to Bash itself but another core package (**EDIT:** glibc AFAIK).

I don't recall the specifics but Alpine has a similar problem I became aware of a week or so ago (_**EDIT:** [this answer explains it](https://unix.stackexchange.com/questions/671246/why-is-bash-not-evaluating-the-executable-bit-correctly-in-alpine-3-14-2/671279#671279), for Debian 10 vs 11 it seems to be related to the glibc update, notably with kernels prior to 5.8_), where a change since their 3.14 release resulted in a syscall that older kernels didn't support properly.. I was surprised as it broke the `hash` command from bash from working correctly (_the Sentry.io CLI installer script was failing because of that_).

georglauterbach:
Very interesting, so it was as I suspected earlier in the issue discussion (a kernel thing...). Thank you for sharing the information here, very much appreciated!

oblitum:
@georglauterbach tbh I'm not sure how that relates to a leak due to subshell'ing, I couldn't connect the dots.

PR #2401, opened 2022-02-08
improvement: get rid of subshell + `exec` in `helper-functions.sh`

georglauterbach:
# Description

The new way of executing `sha512sum` should work as well as the old way
but without the clutter and possible problems the usage of subshells +
exec incurs.

Moreover, there was a misconception about array expansion. Using `""`
around an expanding array (`${ARRAY[@]}`) is quite fine (and actually
the preffered way), not because it makes the expansion _one_ string
(this would be `${ARRAY[*]}`), but it makes sure when elements are
expanded, each element has `""` around them so to speak, i.e. there is
no re-splitting of these elements.

This PR is trying to a be step in uncluttering the changedetector process to find out what is causing the resource leak described in #2348.

## Type of change

- [x] Possible bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/target/scripts/helper-functions.sh
+++ b/target/scripts/helper-functions.sh
@@ -231,7 +231,7 @@ function _monitored_files_checksums
   # If a wildcard path pattern (or an empty ENV) would yield an invalid path
   # or no results, `shopt -s nullglob` prevents it from being added.
   shopt -s nullglob
-  local CERT_FILES
+  declare -a CERT_FILES
 
   # React to any cert changes within the following letsencrypt locations:
   CERT_FILES=(
--- a/target/scripts/helper-functions.sh
+++ b/target/scripts/helper-functions.sh
@@ -225,33 +225,33 @@ export -f _notify
 CHKSUM_FILE=/tmp/docker-mailserver-config-chksum
 
 # Compute checksums of monitored files,
-# returned output is lines of hashed content + filepath pairs.
+# returned output on `stdout`: hash + filepath tuple on each line
 function _monitored_files_checksums
 {
   # If a wildcard path pattern (or an empty ENV) would yield an invalid path
   # or no results, `shopt -s nullglob` prevents it from being added.
   shopt -s nullglob
+  local CERT_FILES
 
   # React to any cert changes within the following letsencrypt locations:
-  local CERT_FILES=(
+  CERT_FILES=(
     /etc/letsencrypt/live/"${SSL_DOMAIN}"/*.pem
     /etc/letsencrypt/live/"${HOSTNAME}"/*.pem
     /etc/letsencrypt/live/"${DOMAINNAME}"/*.pem
   )
 
-  # CERT_FILES should expand to separate paths, not a single string;
-  # otherwise fails to generate checksums for these file paths.
-  #shellcheck disable=SC2068
-  (
-    cd /tmp/docker-mailserver || exit 1
-    exec sha512sum 2>/dev/null -- \
-      postfix-accounts.cf \
-      postfix-virtual.cf \
-      postfix-aliases.cf \
-      dovecot-quotas.cf \
-      /etc/letsencrypt/acme.json \
-      ${CERT_FILES[@]}
-  )
+  if [[ ! -d /tmp/docker-mailserver ]]
+  then
+    return 1
+  fi
+
+  sha512sum 2>/dev/null --                     \
+    /tmp/docker-mailserver/postfix-accounts.cf \
+    /tmp/docker-mailserver/postfix-virtual.cf  \
+    /tmp/docker-mailserver/postfix-aliases.cf  \
+    /tmp/docker-mailserver/dovecot-quotas.cf   \
+    /etc/letsencrypt/acme.json                 \
+    "${CERT_FILES[@]}"
 }
 export -f _monitored_files_checksums
 
--- a/target/scripts/check-for-changes.sh
+++ b/target/scripts/check-for-changes.sh
@@ -132,7 +132,7 @@ do
   # mark changes as applied
   mv "${CHKSUM_FILE}.new" "${CHKSUM_FILE}"
 
-  sleep 1
+  sleep 2
 done
 
 exit 0
--- a/target/scripts/check-for-changes.sh
+++ b/target/scripts/check-for-changes.sh
@@ -50,6 +50,7 @@ do
   # get chksum and check it, no need to lock config yet
   _monitored_files_checksums >"${CHKSUM_FILE}.new"
   cmp --silent -- "${CHKSUM_FILE}" "${CHKSUM_FILE}.new"
+
   # cmp return codes
   # 0 – files are identical
   # 1 – files differ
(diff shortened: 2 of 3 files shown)


georglauterbach:
When this is merged, #2398 can possibly be closed.

target/scripts/helper-functions.sh
```diff
@@ -225,33 +225,33 @@ export -f _notify
...
+    return 1
+  fi
+
+  sha512sum 2>/dev/null --                     \
+    /tmp/docker-mailserver/postfix-accounts.cf \
+    /tmp/docker-mailserver/postfix-virtual.cf  \
+    /tmp/docker-mailserver/postfix-aliases.cf  \
+    /tmp/docker-mailserver/dovecot-quotas.cf   \
+    /etc/letsencrypt/acme.json                 \
+    "${CERT_FILES[@]}"
```
casperklein, line 254: I think, the quotes must be removed. See https://github.com/docker-mailserver/docker-mailserver/pull/2326

```suggestion
    ${CERT_FILES[@]}
```

casperklein: Update: Now I think the quotes are fine. Initially, no array `[@]` was used, but just `${CERT_FILES}`. Quoting this, leads to a problem. Hower using a quoted array like above should be fine.

@polarathene Agree?

georglauterbach: I'd like to wait until I see what tests are saying because right now, I do not see why re-splitting of elements is needed here. Because of

> `# CERT_FILES should expand to separate paths, not a single string;`

it look like to me that there was a misconception that the array is expanded but is a single string, which is not the case. I could be wrong though - let's see.

casperklein (approved):

target/scripts/helper-functions.sh
casperklein: See my comment above 😉

georglauterbach: > See my comment above 😉 

You beat me to it 🙈

wernerfred (approved):

NorseGaud (approved):

georglauterbach:
Alright, we have one failing check (which is related as well):

``` TXT
not ok 10 checking changedetector: can detect changes & between two containers using same config in 15732ms
# (from function `assert_output' in file test/test_helper/bats-assert/src/assert.bash, line 231,
#  in test file test/mail_changedetector.bats, line 58)
#   `assert_output --partial "postfix: stopped"' failed
# 
# substring (1 lines):
#   postfix: stopped
# output (12 lines):
#   [[ TASKS ]]  2022-02-08 12:06:26 Start check-for-changes script.
#   [[  INF  ]]  2022-02-08 12:06:26 Using postmaster address postmaster@my-domain.com
#   [[  INF  ]]  2022-02-08 12:06:37 check-for-changes is ready
#   [[  INF  ]]  2022-02-08 12:06:37 Change detected
#   [ WARNING ]  Lock file /tmp/docker-mailserver/check-for-changes.sh.lock exists. Another check-for-changes.sh execution is happening. Trying again shortly...
#   [[  INF  ]]  Checking file line endings
#   [[  INF  ]]  Regenerating postfix user list
#   [[  INF  ]]  Creating user 'user1' for domain 'localhost.localdomain'
#   [[  INF  ]]  Creating user 'user2' for domain 'otherdomain.tld'
#   [[  INF  ]]  Creating user 'user3' for domain 'localhost.localdomain' with attributes 'userdb_mail=mbox:~/mail:INBOX=~/inbox'
#   [[  INF  ]]  Adding regexp alias file postfix-regexp.cf
#   [[  INF  ]]  Configuring root alias
# --
#
```

To me, it seems as if the restart of Postfix is about to happen because we see that the script says: `Change detected`. Maybe it's too slow at the moment because ealier, the `exec` would somehow interplay with the file lock (that is now being waited for, I guess (`[ WARNING ]  Lock file /tmp/docker-mailserver/check-for-changes.sh.lock exists. Another check-for-changes.sh execution is happening. Trying again shortly...`)). @NorseGaud @casperklein @wernerfred @polarathene any idea? Maybe the weird `exec` had a bad tendency to masquerade a problem with locks in the changedetector?

georglauterbach:
Yes, just tested it - increasing the timeout (by a mere 10s) does the job. I think this is absolutely fine. I will push a fix, and increase the whole script `sleep` duration from 1s to 2s as well.

target/scripts/helper-functions.sh
georglauterbach: Confirmed - this is not an issues :D

NorseGaud:
Something to consider: sleep values are dependent on a certain state. If heavy load is happening, are these sleep values valid? It's typically much better to check for some sort of state instead of relying on sleep. Especially when you find yourself increasing or decreasing sleep values.

georglauterbach:
> Something to consider: sleep values are dependent on a certain state. If heavy load is happening, are these sleep values valid? It's typically much better to check for some sort of state instead of relying on sleep. Especially when you find yourself increasing or decreasing sleep values.

You're very right. We should consider removing sleeps as much as possible, but I have to admit, my main focus was on removing the `exec` :)

casperklein (approved):

NorseGaud (approved):

wernerfred (approved):

georglauterbach:
I will leave this open overnight in case @polarathene has something to add here. If you don't, by all accounts, go ahead and merge this PR @polarathene :)

target/scripts/helper-functions.sh
```diff
@@ -225,33 +225,33 @@ export -f _notify
...
     /etc/letsencrypt/live/"${HOSTNAME}"/*.pem
     /etc/letsencrypt/live/"${DOMAINNAME}"/*.pem
   )
 
-  # CERT_FILES should expand to separate paths, not a single string;
-  # otherwise fails to generate checksums for these file paths.
-  #shellcheck disable=SC2068
-  (
-    cd /tmp/docker-mailserver || exit 1
-    exec sha512sum 2>/dev/null -- \
```
polarathene, line 247: I take it we don't care about the removed directory check and directory change because `sha512sum` will handle invalid paths fine by ignoring them, and we're returning the output instead of redirecting to a file right? (which original may have done, I haven't gone through history to confirm).

The subshell was only added because of pairing it with the directory change failure case?

target/scripts/helper-functions.sh
polarathene: I think previously it was a for loop that appended to the `CERT_FILES` variable, and the quoted string was preventing the paths being treated separately for input... then I refactored to an array but didn't know if it'd behave different so stuck with what worked? :sweat_smile: 

I trust you two since the you both know shell-fu better than me and the tests I wrote are still passing :+1:

polarathene (approved):
LGTM :+1: 

Thanks for the improvement and cleanup!

target/scripts/helper-functions.sh
casperklein: There is still a directory check. See line 243-246.

>and we're returning the output instead of redirecting to a file right?

The logic haven't changed. See line 51.

PR #2418, opened 2022-02-18
Add SpamAssassin KAM

### LINKED ISSUE #2412: SpamAssassin - KAM channel auto-update
Hey

For keeping `KAM.cf` up to date, they have a channel:
https://mcgrail.com/template/kam.cf_channel

`wget https://mcgrail.com/downloads/kam.sa-channels.mcgrail.com.key`
`sa-update --import kam.sa-channels.mcgrail.com.key`

Then in a cron:

`sa-update --gpgkey 24C063D8 --channel kam.sa-channels.mcgrail.com`

I know I can run some custom commands on start-up of the mailserver,
but,

1) Do you have already anything that makes this easy to do?

2) Would you consider making this native of the mailserver, as a setting in the env file? e.g. `SPAMASSASSIN_KAM_CHANNEL=1`

Thank you very much.

georglauterbach:
# Description

Adds KAM (<https://mcgrail.com/template/projects#KAM1>) to SpamAssassin (SA) if SA is enabled. If not, there is no effect.

Closes #2412

## Type of change

## Checklist:


casperklein:
I can only speak for myself, but I never heared about KAM before the mentioned issue. So a quick explanation what KAM is or at least a link to the homepage/documentation should be in `environment.md` and '`mailserver.env`.

docs/content/config/environment.md
```diff
@@ -345,16 +345,25 @@ Finally the logrotate interval **may** affect the period for generated reports.
...
 - 0 => Spam messages will be bounced (_rejected_) without any notification (_dangerous_).
+- **1** => Spam messages will be delivered to the inbox and tagged as spam using `SA_SPAM_SUBJECT`.
 
-##### MOVE_SPAM_TO_JUNK
+##### ENABLE_SPAMASSASSIN_KAM=1
 
-- **1** => Spam messages will be delivered in the `Junk` folder.
-- 0 => Spam messages will be delivered in the mailbox.
+If SpamAssassin is enabled, KAM can be individually enabled or disabled as well. If
+SpamAssassin is disabled, this environment variable will have no effect.
```
casperklein, line 354:
```suggestion
[KAM](https://mcgrail.com/template/projects#KAM1) is a 3rd party SpamAssassin ruleset, provided by the McGrail Foundation. If SpamAssassin is enabled, KAM can be used in addition to the default ruleset.
```

docs/content/config/environment.md
```diff
@@ -345,16 +345,25 @@ Finally the logrotate interval **may** affect the period for generated reports.
...
+
+- 0 => KAM will be disabled even when SpamAssassin is enabled
+- **1** => KAM will be enabled when SpamAssassin is enabled
```
casperklein, line 357:
```suggestion
- 0 => KAM disabled
- **1** => KAM enabled
```

It's already explained directly above, that KAM will only be used if SA is enabled.

casperklein:
> I will add this.

Oh, I didn't see that you answered meanwhile 😉

casperklein:
> I will add this. I can, if you like, change the default to 0 too, if you think this is more appropriate?

I think that's a good idea. I don't use KAM and don't see a reason to do so at the moment. I am very satisfied, how SA performs in the current state. Enabling KAM per default, could lead to new problems / more false positives.

mailserver.env
```diff
@@ -260,7 +260,10 @@ ENABLE_SPAMASSASSIN=0
 # deliver spam messages in the inbox (eventually tagged using SA_SPAM_SUBJECT)
 SPAMASSASSIN_SPAM_TO_INBOX=1
 
-# spam messages will be moved in the Junk folder (SPAMASSASSIN_SPAM_TO_INBOX=1 required)
+# only has an effect if `ENABLE_SPAMASSASSIN=1`
```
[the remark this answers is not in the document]
georglauterbach: I was thinking about how to do this as well. I think ideally, only the real documentation should contain the descriptions, etc. and `mailserver.env` should have only the variables and the default values.

But I agree, this should not be a blocker for this PR. Anything else that comes to mind when reviewing this PR? :)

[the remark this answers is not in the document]
casperklein: I haven't time yet to review the whole PR properly. That's just one thing that I immediately saw. I will do it later or tomorrow.

> I think ideally, only the real documentation should contain the descriptions

I slightly disagree, we should stick to one decision. My "thats not something this PR must solve" relates only the "PS:" part.

General options for the future could be:

1. Include the description (pro: you don't have to open the documentary).
2. Or purge all description, and link to the documentation.

But we should try to avoid a mixed setup. In the last PRs I always put the same description in the env file as in the documentation. That said, I'd like to see here the description as well, until we decide to go another route.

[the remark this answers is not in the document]
georglauterbach: I will add a description here as well. We can open a maintainers discussion about this if want to have a decision on the topic.

docs/content/config/environment.md
```diff
@@ -345,16 +345,24 @@ Finally the logrotate interval **may** affect the period for generated reports.
 
 ##### SPAMASSASSIN_SPAM_TO_INBOX
 
-- **1** => Spam messages will be delivered to the inbox and tagged as spam using `SA_SPAM_SUBJECT`.
 - 0 => Spam messages will be bounced (_rejected_) without any notification (_dangerous_).
+- **1** => Spam messages will be delivered to the inbox and tagged as spam using `SA_SPAM_SUBJECT`.
 
-##### MOVE_SPAM_TO_JUNK
+##### ENABLE_SPAMASSASSIN_KAM=1
```
casperklein, line 351:
```suggestion
##### ENABLE_SPAMASSASSIN_KAM
```

target/scripts/startup/setup-stack.sh
```diff
@@ -960,6 +960,22 @@ function _setup_security_stack
...
+    if [[ ${ENABLE_SPAMASSASSIN_KAM} -eq 1 ]]
+    then
+      _notify 'inf' 'Configuring Spamassassin KAM'
+      local SPAMASSASSIN_KAM_CRON_FILE=/etc/cron.daily/10-spamassassin_kam
+
+      sa-update --import /etc/spamassassin/kam/kam.sa-channels.mcgrail.com.key
+      cat >"${SPAMASSASSIN_KAM_CRON_FILE}" <<"EOM"
+#! /bin/bash
+
+sa-update --gpgkey 24C063D8 --channel kam.sa-channels.mcgrail.com -v 2>&1
```
casperklein, line 973:
```suggestion
sa-update --gpgkey 24C063D8 --channel kam.sa-channels.mcgrail.com 2>&1
```

Otherwise there will be output (an email) every day the cronjob runs.

target/scripts/startup/setup-stack.sh
```diff
@@ -960,6 +960,22 @@ function _setup_security_stack
       sed -i "s|\$final_spam_destiny.*=.*$|\$final_spam_destiny = D_BOUNCE;|g" /etc/amavis/conf.d/49-docker-mailserver
       sed -i "s|\$final_bad_header_destiny.*=.*$|\$final_bad_header_destiny = D_BOUNCE;|g" /etc/amavis/conf.d/49-docker-mailserver
     fi
+
+    if [[ ${ENABLE_SPAMASSASSIN_KAM} -eq 1 ]]
+    then
+      _notify 'inf' 'Configuring Spamassassin KAM'
+      local SPAMASSASSIN_KAM_CRON_FILE=/etc/cron.daily/10-spamassassin_kam
```
casperklein, line 967:
```suggestion
      local SPAMASSASSIN_KAM_CRON_FILE=/etc/cron.daily/spamassassin_kam
```

The other files in that directory don't use a number prefix. Unless there is a reason for that, we should keep it uniform.

casperklein (changes requested):

mailserver.env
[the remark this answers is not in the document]
polarathene: I haven't looked over the PR to review it, but glanced over this conversation.

Presently the ENV file and docs aren't in sync, some have information or vars that the other lacks.

There is a syntax feature for the docs which would be suitable for documenting these vars and should generate semantic HTML for a parser tool to inspect and extract out the env details fairly easily from that point that we could generate our `mailserver.env` from it if desired.

I'd offer to tackle that but already have a lot on my backlog for this project to handle that's dependent upon me restoring my PC (_waiting on a new monitor to be delivered from overseas, as current one seemed to have gotten damaged from powercut or being a decade old (still used DVI instead of HDMI / DP)_). Happy to take care of the env generation once I clear out my existing backlog for the project though :sweat_smile:

target/scripts/startup/setup-stack.sh
polarathene: Would it be better to just extract this to it's own script file in a `target/cron/` or similar location? (_we also have `/etc/dms-settings` (file) and `/etc/dms` (directory)_)

I know it's small, but it would seem a little cleaner to centralize anything cron specific? Or do you feel inline generation like this is better suited? Personally the syntax and indentation to support it vs copying a file over doesn't seem like a great tradeoff?

We have a few different ways in this file alone for cron jobs. Multiple echo calls (1 per line for written file), `mv` for third-party and `sed` (adjustment fix), `chmod` differing permissions across the cron jobs too. That could be addressed by a separate PR, just raising awareness that this approach muddies the already inconsistent handling.

target/scripts/startup/setup-stack.sh
```diff
@@ -960,6 +960,22 @@ function _setup_security_stack
...
+
+      sa-update --import /etc/spamassassin/kam/kam.sa-channels.mcgrail.com.key
+      cat >"${SPAMASSASSIN_KAM_CRON_FILE}" <<"EOM"
+#! /bin/bash
+
+sa-update --gpgkey 24C063D8 --channel kam.sa-channels.mcgrail.com -v 2>&1
+
+EOM
+
+      chmod +x "${SPAMASSASSIN_KAM_CRON_FILE}"
```
polarathene, line 977: This shouldn't be required either if the file is committed with `+x` set, or via Dockerfile build as extra precaution? (_not sure how reliable `+x` is in if building image on platform like windows_)

mailserver.env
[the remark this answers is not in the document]
georglauterbach: That sounds great @polarathene! There is, as always, no timing constraint of course - you know that, just wanted to say that again :) I will resolve this conversation as we have reached a plan for the future.

target/scripts/startup/setup-stack.sh
georglauterbach: Probably a good idea, handling cron in a separate function maybe. I will keep this in mind and when I come to it, provide a PR that takes care of uniform `cron`-handling.

target/scripts/startup/setup-stack.sh
georglauterbach: There is a reason: this should run before the normal Spamassassin cron script runs. When I tend to uniform cron-handling, I will see that all scripts get a number prefix.

target/scripts/startup/setup-stack.sh
georglauterbach: @casperklein made me aware in one of my last PRs that this was missing, so I added it here not sure if it was necessary. I don't see any harm here though, and because we do not support Windows, we do not need to care about whether `chmod +x` work in Windows, right?

polarathene: > because we do not support Windows, we do not need to care about whether `chmod +x` work in Windows, right?

I was referring to not being sure if the `+x` metadata is retained if you checkout/clone from git onto a Windows machine and build the Dockerfile from that.

Which was regarding if this `chmod` line was dropped because we could commit the executable `+x` with a separate `.sh` file. Alternatively, that separate `.sh` file could get a more explicit `chmod +x` line added to the Dockerfile which would resolve any edge case concerns about that issue :sweat_smile: 

---

I'm not fussed either way, but from a maintenance viewpoint with the other approaches to handing cron files elsewhere in `setup-stack.sh`, I don't quite get why the PR does it this way (heredoc?).

If a future PR extracts out related cron specific logic, perhaps the `echo` statements will be converted to this approach with heredoc as well if that seems more appropriate?

target/scripts/startup/setup-stack.sh
casperklein: `crond` does not care about naming and will run all `@daily` jobs in parallel (with some ms delay of course). It will not run the jobs sequential and wait for each to be finished). That's why no script in that directory uses the naming scheme.

> This should run before the normal Spamassassin cron script

What's the exact reason? I guess it's because SA cron reloads the service? Two suggestions to solve this:

1. Add reload function to the KAM cronjob (stolen from `/etc/cron.daily/spamassassin`):	
```
# Tell a running spamd to reload its configs and rules.
reload() {
    # Reload
    if which invoke-rc.d >/dev/null 2>&1; then
        invoke-rc.d --quiet spamassassin status > /dev/null && \
          /etc/init.d/spamassassin reload > /dev/null
    else
        /etc/init.d/spamassassin reload > /dev/null
    fi
    if [ -d /etc/spamassassin/sa-update-hooks.d ]; then
        run-parts --lsbsysinit /etc/spamassassin/sa-update-hooks.d
    fi
}
```

2. Move `/etc/cron.daily/spamassassin` to `/etc/cron.d/spamassassin` and source it in the KAM cronjob after the KAM update.

The second solution seems more elegant on first sight to me, but then I realized, the SA cronjob will only reload SA, if there where SA updates. So if there are KAM updates, but no SA updates, this will not work reliable.

target/scripts/startup/setup-stack.sh
casperklein: What I do in other projects is, to build the exact directory structure under "rootfs" and put the files there:

```
rootfs/etc/cron
rootfs/etc/supervisord
rootfs/bin
```

That way you have a clear overview about the files and how they will be in the image + only one COPY instruction is needed in Dockerfile --> `COPY rootfs /`

target/scripts/startup/setup-stack.sh
georglauterbach: I didn't know that cron executes "in parallel" :D Nice to know! I think it is not strictly required that one runs after another because there is at most a delay of one day (if KAM updates are received after the normal SA updates run, the next day, SA will also update the KAM signatures). I will renamed the file accordingly.

target/scripts/startup/setup-stack.sh
georglauterbach: > Which was regarding if this chmod line was dropped because we could commit the executable +x with a separate .sh file. Alternatively, that separate .sh file could get a more explicit chmod +x line added to the Dockerfile which would resolve any edge case concerns about that issue

Can you elaborate on this again, because I don't get it :) What do you mean by "separate .sh file"?

---

I like the heredoc because you can immediate see the content of the file. With `echo` you need `\n` etc. which I find to not be suitable. In the future PR, I will tend to this issue and make all cron additions heredocs.

casperklein: > Can you elaborate on this again, because I don't get it :) What do you mean by "separate .sh file"?

I think he meant something like [this](https://github.com/docker-mailserver/docker-mailserver/blob/master/target/dovecot/dovecot-purge.cron).

georglauterbach: Ah, now I see. I actually like the heredoc and `chmod +x` better, but if a majority votes for the extra-script-approach, I'd implement it.

casperklein: I think both attempts are valid and have their pro/cons. But as always: We should keep it uniform 😆

georglauterbach: Well, at least this is something a future PR will solve :D

target/scripts/startup/setup-stack.sh
casperklein: I don't know how often the rules are updated and if one day delay is "acceptable".
But by adding `&& /etc/init.d/spamassassin reload` we should be on the safe side?

target/scripts/startup/setup-stack.sh
```diff
@@ -960,6 +960,22 @@ function _setup_security_stack
...
+    if [[ ${ENABLE_SPAMASSASSIN_KAM} -eq 1 ]]
+    then
+      _notify 'inf' 'Configuring Spamassassin KAM'
+      local SPAMASSASSIN_KAM_CRON_FILE=/etc/cron.daily/spamassassin_kam
+
+      sa-update --import /etc/spamassassin/kam/kam.sa-channels.mcgrail.com.key
+      cat >"${SPAMASSASSIN_KAM_CRON_FILE}" <<"EOM"
+#! /bin/bash
+
+sa-update --gpgkey 24C063D8 --channel kam.sa-channels.mcgrail.com 2>&1
```
casperklein, line 973:
```suggestion
sa-update --gpgkey 24C063D8 --channel kam.sa-channels.mcgrail.com
```

`2>&1` is not needed. Cron handles both file descriptors.

casperklein (approved):

target/scripts/startup/setup-stack.sh
```diff
@@ -960,6 +960,22 @@ function _setup_security_stack
...
+    if [[ ${ENABLE_SPAMASSASSIN_KAM} -eq 1 ]]
+    then
+      _notify 'inf' 'Configuring Spamassassin KAM'
+      local SPAMASSASSIN_KAM_CRON_FILE=/etc/cron.daily/spamassassin_kam
+
+      sa-update --import /etc/spamassassin/kam/kam.sa-channels.mcgrail.com.key
+      cat >"${SPAMASSASSIN_KAM_CRON_FILE}" <<"EOM"
+#! /bin/bash
+
+sa-update --gpgkey 24C063D8 --channel kam.sa-channels.mcgrail.com
```
casperklein, line 973:
```suggestion
sa-update --gpgkey 24C063D8 --channel kam.sa-channels.mcgrail.com && \
/etc/init.d/spamassassin reload
```

https://github.com/docker-mailserver/docker-mailserver/pull/2418#discussion_r810538291

casperklein (approved):

polarathene (approved):

ghnp5:
Hey  @georglauterbach 
Thanks for this!

Did you test `sa-update` on this?

For some reason, `--gpgkey 24C063D8` doesn't work for me.
When there is an actual update, `sa-update` it fails with:

     The update downloaded successfully, but it was not signed with a trusted GPG
     key.  Instead, it was signed with the following keys:
         24C063D8 

So I had to change the command to `--nogpg`.

`sa-update --nogpg --channel kam.sa-channels.mcgrail.com --gpghomedir /var/lib/spamassassin/sa-update-keys -v 2>&1`

georglauterbach:
> When there is an actual update, `sa-update` it fails with:
> 
>      The update downloaded successfully, but it was not signed with a trusted GPG
> 
>      key.  Instead, it was signed with the following keys:
> 
>          24C063D8 
>
> So I had to change the command to `--nogpg`.
> 
> `sa-update --nogpg --channel kam.sa-channels.mcgrail.com --gpghomedir /var/lib/spamassassin/sa-update-keys -v 2>&1`

I experienced a similar issue, but (as I did with this PR) I did not use `--nogpt`, but I removed the `--gpghomedir /var/lib/spamassassin/sa-update-keys` option which was messing with the command. Without this option, all should work well. This PR makes use of that as well.

ghnp5:
ah nice one! Thanks very much. 👍👍

target/scripts/startup/setup-stack.sh
```diff
@@ -960,6 +960,23 @@ function _setup_security_stack
...
+    if [[ ${ENABLE_SPAMASSASSIN_KAM} -eq 1 ]]
+    then
+      _notify 'inf' 'Configuring Spamassassin KAM'
+      local SPAMASSASSIN_KAM_CRON_FILE=/etc/cron.daily/spamassassin_kam
+
+      sa-update --import /etc/spamassassin/kam/kam.sa-channels.mcgrail.com.key
+      cat >"${SPAMASSASSIN_KAM_CRON_FILE}" <<"EOM"
+#! /bin/bash
+
+sa-update --gpgkey 24C063D8 --channel kam.sa-channels.mcgrail.com && \
```
williamdes, line 973: the GPG key is a bit small, could a full one be used ?

georglauterbach: If you find the complete key, please, feel free to open a PR :D

williamdes: Yeah, but for some reason they do use a short version.

PR #2420, opened 2022-02-18
refactoring: split helper functions into smaller scripts

georglauterbach:
# Description

This PR does **not** change any functionality of DMS, it just refactors the `helper-function.sh` script into smaller scripts in the earlier created `helpers/` directory. `helper-function.sh` was a mess and all over the place. Now, a single script `index.sh` (which was already created before this PR) sources all helper functions, or one can individually source smaller helpers, somewhat sorted. Again, this is just a refactoring without functionality changes :)

One more thing: I changed `.` to `source`. As described in the commit message, `source` is idiomatic to Bash, and I find it easier to understand that a script is sourced here. I hope this is not a major problem for someone :)

## Type of change

- [x] Improvement (non-breaking change that does improve existing functionality)

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [x] If necessary I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #2419, opened 2022-02-18
fix: do not add accounts that already exists to account files

georglauterbach:
# Description

Includes a fix with which users are not added to the accounts files of Postfix and Dovecot if they are already present.

Fixes #2416

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Improvement (non-breaking change that does improve existing functionality)

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [ ] If necessary I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #2424, opened 2022-02-20
PERMIT_DOCKER=none as new default value

casperklein:
# Description

Using DMS with podman can lead to security issues (e.g. an open relay), if not done properly. This PR aims a more secure out-of-the-box setup for podman users.

The following changes have been done:

1. New default value for PERMIT_DOCKER is now `none`.
2. An empty PERMIT_DOCKER value is now considered as `none`.
3. To keep the old behaviour (adding the container IP address to Postfix's 'mynetworks'), `PERMIT_DOCKER=container` must be used.

Related to: https://github.com/docker-mailserver/docker-mailserver/pull/2393

## Type of change

## Checklist:


wernerfred:
> 3\. To keep the old behaviour

Wouldn't that qualify for a breaking change?

casperklein:
My opinion: definitely. But see our maintainer discussion from 19. Dec and the provided example, which also dropped functionality.

tl;dr. if the container comes up without interaction, it's not a breaking change and people should read the changelog + adjust their config.

polarathene:
Breaking change should probably wait until next major?

What's the impact of existing users that upgrade and don't adjust the ENV accordingly? How soon would they realize something like this happened? (_assuming they don't read changelog and expect things not to break from a minor release_)

casperklein:
> What's the impact of existing users that upgrade and don't adjust the ENV accordingly?

Two things I can think of:

1. Some customization, that runs a script inside the container, that relies on sending unauthenticated mails.
2. Running an open-relay will probably be broken after this 😆

polarathene:
> 1. Some customization, that runs a script inside the container, which sends mails unauthenticated.

Sounds like it warrants waiting on major version then?

I know we're a little less strict on that if the change isn't likely to affect many users in practice and our release notes clearly cover the change. I'm fine with this being a minor version change :+1:

target/scripts/startup/setup-stack.sh
```diff
@@ -768,22 +786,10 @@ function _setup_docker_permit
...
-      ;;
-
     * )
-      _notify 'inf' 'Adding container ip in my networks'
-      postconf -e "$(postconf | grep '^mynetworks =') ${CONTAINER_IP}/32"
-      echo "${CONTAINER_IP}/32" >> /etc/opendmarc/ignore.hosts
-      echo "${CONTAINER_IP}/32" >> /etc/opendkim/TrustedHosts
+      _notify 'warn' "Invalid value for PERMIT_DOCKER: ${PERMIT_DOCKER}"
+      _notify 'inf' "Clearing Postfix's 'mynetworks'"
+      postconf -e "mynetworks ="
```
polarathene, line 792: This happens with the default `none` I take it? So anyone who was relying on previous behaviour will see the warning at least in their logs?

casperklein: No. `PERMIT_DOCKER=` now defaults to `PERMIT_DOCKER=none`.

This warning will only be shown, if something unexpected is set, e.g. a typo like `PERMIT_DOCKER=hosts`

polarathene (approved):
LGTM, may want to adjust some log messages though?

target/scripts/startup/setup-stack.sh
polarathene: Yup, edited my comment while reviewing and realized I misunderstood :grimacing: 

Users will have to rely on checking release notes before upgrading or raising issues to find out if the minor update impacts them. Shouldn't be a problem, this is a project you shouldn't blindly update AFAIK :+1:

target/scripts/startup/setup-stack.sh
```diff
@@ -754,6 +754,24 @@ function _setup_docker_permit
...
+        NETWORK=$(_sanitize_ipv4_to_subnet_cidr "${NETWORK}")
+        _notify 'inf' "Adding docker network ${NETWORK} in my networks"
+        postconf -e "$(postconf | grep '^mynetworks =') ${NETWORK}"
+        echo "${NETWORK}" >> /etc/opendmarc/ignore.hosts
+        echo "${NETWORK}" >> /etc/opendkim/TrustedHosts
+      done
+      ;;
+
+    "container" )
+      _notify 'inf' 'Adding container IP address to my networks'
```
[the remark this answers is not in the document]
casperklein: Done.

polarathene (approved):

georglauterbach:
I have an upcoming PR as well that changes some default values (unifies ENV setup). These changes would technically be identical in impact to the changes in this PR. IMO we should wait for #2419 ad #2420 to be merged, release `10.5.0` first and then see whether `10.6.0` or `11.0.0` should be the next successor after this PR and my next PR are merged.

casperklein:
If this PR can go into 10.6.0, why not straight into 10.5.0? Whats the benefit in waiting?

georglauterbach:
> If this PR can go into 10.6.0, why not straight into 10.5.0? Whats the benefit in waiting?

Fair point, and very much fine with me :)

georglauterbach (approved):

target/scripts/startup/setup-stack.sh
```diff
@@ -754,38 +754,44 @@ function _setup_docker_permit
...
-
-    "network" )
-      _notify 'inf' "Adding docker network in my networks"
-      postconf -e "$(postconf | grep '^mynetworks =') 172.16.0.0/12"
-      echo 172.16.0.0/12 >> /etc/opendmarc/ignore.hosts
-      echo 172.16.0.0/12 >> /etc/opendkim/TrustedHosts
-      ;;
-
     "connected-networks" )
       for NETWORK in "${CONTAINER_NETWORKS[@]}"
```
georglauterbach, line 772: Could the generation of `CONTAINER_NETWORKS` be "inlined" to this `case` statement or is it needed somewhere else?

casperklein: I think that's possible.

But the same goes for `$CONTAINER_IP` and `$CONTAINER_NETWORK`. The generation of these values are also handled outside the case statements. I think that was made for better readability of the case blocks 🤷

georglauterbach: It's not mandatory for this PR, just wanted to have an opinion. We can leave this as is.

wernerfred:
I still See it as a breaking Change and therefore a Major Version bump. If we wajt to avoid a Lot of Major Version bumps we could:

1. Wait with merge of breaking Changes until we have a bunch and Release them all together or 
2. Bring in More and deeper thinking to the reviewal/Implementation process and Pick the "correct" Default straight away instead of changing it Afterwards many Times. 

Ideally we should also have a grace period to warn Users already 1-2 Version ahead about upcoming breaking Changes behaviors.

georglauterbach:
> I still See it as a breaking Change and therefore a Major Version bump. If we wajt to avoid a Lot of Major Version bumps we could:
> 
>     1. Wait with merge of breaking Changes until we have a bunch and Release them all together or
> 
>     2. Bring in More and deeper thinking to the reviewal/Implementation process and Pick the "correct" Default straight away instead of changing it Afterwards many Times.
> 
> 
> Ideally we should also have a grace period to warn Users already 1-2 Version ahead about upcoming breaking Changes behaviors.

While this would be optimal, we're maintaining this in our free time and we do not have the resources to plan one or two (even minor) versions ahead. I've got no problem with releasing `10.5.0` "now" (after feature freeze), and merging this into `master` to be on `:edge` first. The other PR of mine could similarly wait.

polarathene:
We could also have a separate PR that is long-lived open which is a `next`/`major` branch that PRs like this could be merged into, then when ready that could merge to master before making a new major release.

We'd just have the long-lived PR frequently syncing up with `master` branch to ensure tests pass and all commits cleanly carry over to master without any conflicts?

georglauterbach:
> We could also have a separate PR that is long-lived open which is a `next`/`major` branch that PRs like this could be merged into, then when ready that could merge to master before making a new major release.
> 
> We'd just have the long-lived PR frequently syncing up with `master` branch to ensure tests pass and all commits cleanly carry over to master without any conflicts?

This sounds like a possible solution, but we'd need to make sure no `:edge` images are built from this branch (haven't checked whether this would happen at the moment (we could introduce a `:bleeding` tag or something 😆). We'd manually need to sync with master, but this _should_ not be a big issue as not too many PR's well container breaking changes.

polarathene:
It would also need some extra care with merging as we'd want to rebase the commits instead of squash merge for such a branch AFAIK when merging to master for a release. Might also be important to make that branch protected as well.

wernerfred:
Or Just Release New Major Versions If a breaking Change occurs 😅
No big Deal imho, i have No Problem with high Version Numbers. Thats semver ;)

polarathene:
> i have No Problem with high Version Numbers.

Does make our docs version selector less pleasant to use though :sweat_smile: 

The separate branch approach is only useful for queuing up changes for a major release, which isn't that common of a need. Maybe you're right though and major version releases don't need to be few and far between, treated just like minor/patch releases..

georglauterbach:
So, according to the maintainers' discussion, this should be tagged `v11.0.0`, right? I will tag it as such, if someone objects, let me know :)

georglauterbach:
@casperklein we can now go ahead and start merging the open PRs. I'd start with the numerically lowest, i.e. this one and work our way forward.

casperklein:
> I'd start with the numerically lowest,

I good plan. However, I just updated the branch in https://github.com/docker-mailserver/docker-mailserver/pull/2434 to merge next 😆

Issue #2411, opened 2022-02-15
How to check that my mailserver is not sending spam?


billy-mar:
# Subject

Other

## Description

Dear developers and maintainers,

Docker Mailserver is fantastic project, thank you very much for it! ;-)

Could you please help me with one question I didn't find an answer to? I searched in Docs and in github issues...

A few days ago I installed Docker Mailserver and I sent just a few e-mails from it to my own address to verify everything is working fine. Yesterday I received a concerning e-mail from linode.com where I'm hosting the server.

```
We have received a notification from an FBL service that a message you have sent has been marked as spam.
```

I think there are two options.

a) The e-mail it's wrong and I should not worry.

b) Some bot searching for 25/tcp open ports hacked into the mailserver. When I set up my accounts I used a very strong password but I'm slightly worried anyway.

Could you please tell me the best way to monitor all outgoing messages? It's not a privacy concern since I'm the only user of the server...

Thank you.

Billy

casperklein:
Your server could be an _open relay_. You can use external services like [this](https://mxtoolbox.com/diagnostic.aspx) to check that.

If you use the default `docker-compose.yml`, you should find the mail logs in `./docker-data/dms/mail-logs/` on the host. Every mail that was send, is logged there.

billy-mar:
Yes, it is a `SMTP Open Relay`. Thank you for the tip about the MXtoolbox!

I must admit, I don't use the recommended docker-compose.yml. I run it as a "pod" using Podman but I did configured `PERMIT_DOCKER=none` as suggested in the Docs. I'll investigate further. Thank you.

billy-mar:
Fixed by:

```
sed -E -i "s|^mynetworks.*|mynetworks = 127.0.0.0/8 [::1]/128 [fe80::]/64|" /etc/postfix/main.cf
supervisorctl restart postfix
```

(from inside of the container)

At this moment workaround - I need to figure out how to make it permanent.

casperklein:
In the next version or the current edge build, you can set the [`PERMIT_DOCKER=none`](https://github.com/docker-mailserver/docker-mailserver/blob/54f218137990553faa4fed8a29ee8e9d6a68af45/mailserver.env#L52).

billy-mar:
Ouh, I didn't realize that this setting is not available in `latest` / `10.4.0`. Thank you!

billy-mar:
[...]

PR #2429, opened 2022-02-21
docs(podman): Fix broken links and add version hints

p-fruck:
# Description

This is a follow up of #2393, where I accidentally added broken links to the podman documentation.

In addition to this fix, I've added some version hints to the changes introduced in the aforementioned PR to circumvent issues like [this](https://github.com/docker-mailserver/docker-mailserver/issues/2411#issuecomment-1043542552)

## Type of change

## Checklist:

### DIFF
--- a/docs/content/config/advanced/podman.md
+++ b/docs/content/config/advanced/podman.md
@@ -128,7 +128,7 @@ You must also add the ENV `NETWORK_INTERFACE=tap0`, because Podman uses a [hard-
 
 !!! note
 
-    `podman-compose` is not compatible with configuration.
+    `podman-compose` is not compatible with this configuration.
 
 ### Self-start in Rootless Mode
 
--- a/docs/content/config/advanced/podman.md
+++ b/docs/content/config/advanced/podman.md
@@ -170,3 +170,6 @@ firewall-cmd --reload
 ```
 
 Just map all the privilege port with non-privilege port you set in docker-compose.yml before as root user.
+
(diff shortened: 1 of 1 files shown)


wernerfred (commented):
The Version warning should afaik not be necessary as the docs are Versioned as well and released Always together with a new Container Version. 
Looking at the "Edge" Docs is valid for the "Edge" build of the Container. But switching to Docs Version "10.4" (which counts for the 10.4 builds) will not contain the newest additions. 

In fact we implemented the Versioned Docs for exactly that reason, to avoid updating references and Versions throughout the Docs. 

If it Happens to be the Case that "newer" Docs are in a "older" Release we messed Up the Release/merge process but not the Docs (content)

p-fruck:
Hmm, seems like I should enable JS every once in a while... Now I see the versioning dropdown :smile: I'll revert the versioning related changes from the docs. Should I also revert [this line](https://github.com/p-fruck/docker-mailserver/blob/e22b3205f26663366e0ace35ae3a598536508abf/mailserver.env#L52) in the env file?

polarathene (approved):

georglauterbach:
> Should I also revert the change in [this line](https://github.com/p-fruck/docker-mailserver/blob/e22b3205f26663366e0ace35ae3a598536508abf/mailserver.env#L52) in the env file?

Yes. The `none` case is not be visible for the current docs and will only appear from `10.5.0` if I'm not mistaken.

p-fruck:
Alright, I've reverted all changes except for the typo and the broken links :+1:

polarathene (approved):

wernerfred (approved):
Thanks

PR #2436, opened 2022-02-24
Fix doc

jmccl:
Based on where the file is actually put when running through the tutorial.

### DIFF
--- a/docs/content/examples/tutorials/basic-installation.md
+++ b/docs/content/examples/tutorials/basic-installation.md
@@ -95,7 +95,7 @@ In this setup `docker-mailserver` is not intended to receive email externally, s
 
 5. [Generate DKIM keys][docs-dkim] for your domain via `./setup.sh config dkim`.
 
-    Copy the content of the file `docker-data/dms/config/opendkim/keys/example.com/mail.txt` and add it to your DNS records as a TXT like SPF was handled above.
+    Copy the content of the file `config/opendkim/keys/example.com/mail.txt` and add it to your DNS records as a TXT like SPF was handled above.
 
     I use [bind9](https://github.com/docker-scripts/bind9) for managing my domains, so I just paste it on `example.com.db`:
 


georglauterbach:
I'm not quite sure on this, as I did not manage consistency for volume mounts, but the current value seems to be correct to me. It is the location on the host file system within our volume example directories. @polarathene can you confirm?

polarathene:
> but the current value seems to be correct to me. It is the location on the host file system within our volume example directories.

Current docs are correct, the host volume path is a placeholder but reflects what we use in our example `docker-compose` volumes:

https://github.com/docker-mailserver/docker-mailserver/blob/2de3340d2c58a02b4daced192f8662492d055195/docker-compose.yml#L23

@jmcci probably has upgraded an existing install perhaps, before we normalized the volume paths across the docs?

---

> Based on where the file is actually put when running through the tutorial.

@jmccl you say where the file is when running through the tutorial you're updating in the PR? But this volume path is correctly matching the example `docker-compose` volume path earlier in that tutorial:

https://github.com/docker-mailserver/docker-mailserver/blob/2de3340d2c58a02b4daced192f8662492d055195/docs/content/examples/tutorials/basic-installation.md?plain=1#L43

I think the actual correction would be to update this line:

https://github.com/docker-mailserver/docker-mailserver/blob/2de3340d2c58a02b4daced192f8662492d055195/docs/content/examples/tutorials/basic-installation.md?plain=1#L96

We removed the need for `setup.sh` as we now include this file with the relevant container image release. Instead the command would be `docker exec -it mailserver setup config dkim`. This is mentioned early on, but it's probably a good time for our current docs to remove the transition by dropping mention of `setup.sh` / `v10.2`?:

https://github.com/docker-mailserver/docker-mailserver/blob/2de3340d2c58a02b4daced192f8662492d055195/docs/content/examples/tutorials/basic-installation.md?plain=1#L17

Likewise step 8 has some examples with many `setup.sh` commands that would need to be converted (_or shell into the container to run`setup` without the `docker exec mailserver` prefix_). That said [using the external `setup.sh` is still a valid alternative AFAIK](https://github.com/docker-mailserver/docker-mailserver/blob/master/docs/content/config/setup.sh.md), just no longer requiring keeping it in sync with the correct image release.

At least it seems that the support for both caused some confusion for this tutorial. `setup.sh` is setup to use the new path, but still supports the legacy one, so I'm not sure what happened for @jmccl , if they didn't have `./config` already created, `setup.sh` should not have created it...?:

https://github.com/docker-mailserver/docker-mailserver/blob/2de3340d2c58a02b4daced192f8662492d055195/setup.sh#L109-L118

georglauterbach:
I see. I think it is a good idea to remove the calls to `setup.sh` and replace them with the `docker exec -ti mailserver setup ...` alternative and make the reader aware that `setup.sh` can do this as well.

Upgrading to DMS `>10.2.0` likely caused the confusion here, I agree. The docs should be updated

casperklein:
Another thing I run into myself some time ago. When you are not using

```bash
DMS_GITHUB_URL='https://raw.githubusercontent.com/docker-mailserver/docker-mailserver/master'
wget "${DMS_GITHUB_URL}/docker-compose.yml"
wget "${DMS_GITHUB_URL}/mailserver.env"
wget "${DMS_GITHUB_URL}/setup.sh"
chmod a+x ./setup.sh
```

but clone the repo instead, you'll have an `config` directory. Then `setup.sh` falls back using the `config` directory. Newly accounts will then be added under `config`. When you then start DMS with the default compose file, `docker-data/dms/config` is used. An error, that at least one email account must exist, will be printed and DMS does not start.

jmccl:
This was not an upgrade; it was (an attempt at) a clean install. This is my first time trying to use this project.

I started from this page and I generally prefer to start from source rather than rely on build processes I don't understand.

https://github.com/docker-mailserver/docker-mailserver/blob/master/docs/content/examples/tutorials/docker-build.md

I got the source from head yesterday with this command, found on that page.

git clone --recurse-submodules https://github.com/docker-mailserver/docker-mailserver

I then used this page to twiddle the config as the page is about what I'm trying to do.

https://github.com/docker-mailserver/docker-mailserver/blob/master/docs/content/examples/use-cases/forward-only-mailserver-with-ldap-authentication.md

After changing those settings, I went to this page and followed the instructions.

https://github.com/docker-mailserver/docker-mailserver/blob/master/docs/content/examples/tutorials/basic-installation.md

=========================

Things that surprised me:

* It's not clear to me that you can build from source. It's what I thought I was doing, but setup.sh pulls an image anyway.
* It wasn't clear to me there were two ways of doing things. I now see the main README is mostly docker-compose based, but most of the rest of the documentation seems to be setup.sh based.

georglauterbach:
There really is a discrepancy between how one would now do it vs. the way it was done with `setup.sh`. This needs some work indeed. However, at no point do we state that `setup.sh` is used to build DMS.

jmccl:
I agree, I misunderstood. I was just trying to explain what I did and what I was thinking when I did it.

polarathene:
> * It's not clear to me that you can build from source. It's what I thought I was doing, but setup.sh pulls an image anyway.

Building from source would be our [Dockerfile](https://github.com/docker-mailserver/docker-mailserver/blob/master/Dockerfile), just as the example doc you referenced detailed.

`setup.sh` is post-setup, perhaps more like `config.sh` or `util.sh` really.. Originally it was required to use this external helper script file, but in the 10.x release series we've since integrated the bulk of the functionality into the image.

The external script just provides some extras like pulling the image if it's not locally available. By default it's set to out `:latest` release tag:

https://github.com/docker-mailserver/docker-mailserver/blob/2de3340d2c58a02b4daced192f8662492d055195/setup.sh#L15

`setup.sh help` (_also covered in [our docs](https://github.com/docker-mailserver/docker-mailserver/blob/master/docs/content/config/setup.sh.md#usage)_) will also show that you can override that default with your own locally built image, or running container (when appropriate):

https://github.com/docker-mailserver/docker-mailserver/blob/2de3340d2c58a02b4daced192f8662492d055195/setup.sh#L52-L61

> It wasn't clear to me there were two ways of doing things. I now see the main README is mostly docker-compose based, but most of the rest of the documentation seems to be setup.sh based.

You can use `docker-compose.yml`, Docker CLI or other alternatives (Podman / kubernetes). `setup.sh <command>` or `docker exec mailserver setup <command>` are the two different approaches for calling the same commands (_`email`, `alias`, `quota`, `config`, `relay`, `debug`_), the docs thus use `setup.sh` for that since that was previously the only way and there isn't much difference beyond preference.

For you it was a little jarring, since you strayed from our "happy" path by choosing to begin with your own local build (_which is fine, but mostly documented for contributors_). You're welcome to improve our docs for that experience if you'd like :)

---

> This was not an upgrade; it was (an attempt at) a clean install. This is my first time trying to use this project.
> 
> I started from this page as I generally prefer to start from source rather than rely on build processes I don't understand.

Totally fine :+1: 

> I then used this page to twiddle the config as the page is about what I'm trying to do.

Just  a little warning (_although I think we've made it more clear recently_), LDAP support is fully community contributed, we don't have any maintainers that use it or know much to maintain it well, and users of LDAP have raised some issues that are yet to be resolved if you run into those.

> After changing those settings, I went to this page and followed the instructions.

You'd probably want to start off with the basic installation first, then repeat for extras like LDAP I'd have thought? :sweat_smile: 

Still I don't see anything there that would have caused you to end up with the old config volume path that you're contributing here? Were you not using `docker-compose`? How did you configure your `config` volume?

polarathene:
> but clone the repo instead, you'll have an `config` directory. Then `setup.sh` falls back using the `config` directory.

Oops missed this comment! That explains it then :sweat_smile: 

Perhaps we could fix that? Either by dropping the legacy support from `setup.sh` in a future release (v11?) or renaming the repo `config` directory?

georglauterbach:
> > but clone the repo instead, you'll have an `config` directory. Then `setup.sh` falls back using the `config` directory.
> 
> @casperklein Oops I missed this comment! That explains it then sweat_smile
> 
> Perhaps we could fix that? Either by dropping the legacy support from `setup.sh` in a future release (v11?) or renaming the repo `config` directory?

I vote for dropping support of the legacy configuration in `v11`. If you agree, 👍🏼 , if not, 👎🏼  :D

casperklein:
PS: I find the new directory structure still weird. `docker-data/dms` suggest, that `docker-data` is a general docker directory, with DMS as one of many projects. In reality, I doubt, that someone will ever use this directory to place other docker-data in, beside DMS. Why would anyone want to mix-up his DMS stack with data not related to DMS.
While I understand the reasons for the renaming (easier handling of our docs), personally I find it strange to have the proposed directory structure in this repo.

However, no matter what we do, we should do it consequently. And at the moment, this means to rename `config` to `docker-data/dms/config`.

georglauterbach:
I agree with @casperklein that the directory structure proposed in our docs is, at least IMHO, not likely to be found out there in the wild (i.e. a bit weird).

casperklein:
I've created a PR. Let's continue to discuss in #2438.

jmccl:
BTW, thanks to all of you for looking at this so quickly.

PR #2438, opened 2022-02-26
Rename config examples directory

casperklein:
# Description

To align with the documentation, this PR relocates the `config` directory.

See also discussion [here](https://github.com/docker-mailserver/docker-mailserver/pull/2436#issuecomment-1051296031).

## Type of change

## Checklist:


polarathene (dismissed):

polarathene:
Since there was a request to continue discussion here..

@casperklein [said](https://github.com/docker-mailserver/docker-mailserver/pull/2436#issuecomment-1052115148):

> PS: I find the new directory structure still weird. `docker-data/dms` suggest, that `docker-data` is a general docker directory, with DMS as one of many projects. In reality, I doubt, that someone will ever use this directory to place other docker-data in, beside DMS. Why would anyone want to mix-up his DMS stack with data not related to DMS.
> 
> While I understand the reasons for the renaming (easier handling of our docs), personally I find it strange to have the proposed directory structure in this repo.

As stated, it was an improvement to our docs with the normalization effort for consistency. If you grep the docs, you'll find other directories beyond `dms` used in `docker-data`.

This provides a root data/volume directory with sub-folder for each container, thus related data, but still isolated by associated container. I use this approach for storing data with multi-container projects as littering the volume dirs one level higher in a project structure isn't helpful.

---

@georglauterbach [said](https://github.com/docker-mailserver/docker-mailserver/pull/2436#issuecomment-1052116937):

> I agree with @casperklein that the directory structure proposed in our docs is, at least IMHO, not likely to be found out there in the wild (i.e. a bit weird).

Like I replied above to casper, this is a structure I use and makes sense to me, and our docs to avoid any vagueness.

I would assume that our users are familiar enough with Docker that if they did not like the `docker-data` dir and had no need for the extra layer due to no additional containers or isolation needs, that they'd adjust it accordingly. If our docs are not clear enough on that freedom to do so, and you feel that should be clarified better, go for it :)

Personally, I don't see much benefit in changing the paths used, especially since they are more useful in the current state for maintaining the docs vs before when paths were inconsistent or overlapping (_containers from different images using same/similar paths like `config/`, or unclear if referencing directory path from host or container_) across various doc pages.

---

I'm open to it being handled better too. I just remember the state of the docs prior to normalization due to various contributors over time and this sort of thing slipping through where it was harder to correct.

I'm not going to defend the current choice more than that.

If maintainers disagree with the justification and think it's fine/better to simplify the path, that's all good too :+1:

polarathene:
> Why the file mode changes (`chmod -x`) on the helper files?

See the commit message associated to them:

> _remove executable bit from scripts that are only sourced _

casperklein:
> If you grep the docs, you'll find other directories beyond dms used in docker-data.

Okay, I wasn't aware of that. I took only the 4 paths from `docker-compose.yml` into consideration, when making my comment.

Going back to the previous approach (just using `config` and reverting the docs) is IMHO not an option. All the hard work on the docs from you would be useless.

But to avoid problems mentioned in #2436, the directory must be renamed. However, I am still not happy with the new long path.

What about renaming the `config` directory to `config-samples`? This way, you can also clone this repo and use `setup.sh`, but without headache like in #2436 + we avoid to have the long path in our repo.
People that want to use some or all of the example files, can easily copy them to their `config` directory.

polarathene:
<details>
<summary>Redundant</summary>

> However, I am still not happy with the new long path

That's only really an option with the `config/` folder in the repo rather than the docs right? Isn't that folder more of an example? 

It doesn't have to match the docs path, in fact it may be more clearer and less surprising if it's clearly `example/config/` instead?

Is there a benefit to implicitly using that folder when a user clones the repo and uses `setup.sh` at the project root? I don't think the user expected that either. `example/config/` should make it clear that it's an example only and not actually required to run. Some users do seem to get confused about that, similar to `mailserver.env`.

</details>

..I should really read the full message before replying :sweat_smile:

---

> What about renaming the `config` directory to `config-samples`?



Yes that works just as well as `example/config/` :+1:

casperklein:
As long as there are no other subdirectorys, I would like to keep it flat.

What is better: `config-samples` or `config-examples`? I tend to the second one.

polarathene:
> What is better

`config-examples` seems good :+1: 

Or if it's more of a single example rather `config-example`?

casperklein:
> Or if it's more of a single example rather `config-example`?

There are various files in that [directory](https://github.com/docker-mailserver/docker-mailserver/tree/master/config). Plural is fine.

polarathene (approved):

georglauterbach (approved):
LGTM 👍

Due to the file adjustments this PR has to wait until `10.5.0` is released.

casperklein:
See milestone 😉

wernerfred (approved):

PR #2439, opened 2022-02-27
Update 'build' doc

jmccl:
'setup.sh' expects an image named 'mailserver/docker-mailserver:latest' or else it will pull a remote image.

(To some degree whether this change makes sense depends on whether the audience for this page is intended to go on to use the image themselves. If it's for somebody who will use some downstream publishing process perhaps it's not correct.)


casperklein (approved):
Good first contribution, thank you.

georglauterbach (approved):

wernerfred (approved):
LGTM, thx

PR #2447, opened 2022-03-01
macOS bash 5 & zsh doesn't support shopt -s inherit_errexit

NorseGaud:
# Description

macOS/ZSH doesn't support shopt -s inherit_errexit

## Type of change

## Checklist:

### DIFF
--- a/test/linting/lint.sh
+++ b/test/linting/lint.sh
@@ -1,5 +1,4 @@
 #! /bin/bash
-set -x
 # version   v0.2.0 unstable
 # executed  by Make during CI or manually
 # task      checks files against linting targets
@@ -105,7 +104,7 @@ function _shellcheck
     -not -path './target/docker-configomat/*'
   )"
   # macOS lacks parity for `-executable` but presently produces the same results: https://stackoverflow.com/a/4458361
-  [[ "$(uname)" == "Darwin" ]] && FIND_EXEC="-perm +111 -type l -or" || FIND_EXEC="-executable"
+  [[ "$(uname)" == "Darwin" ]] && FIND_EXEC="-perm -711" || FIND_EXEC="-executable"
   # shellcheck disable=SC2248
   F_BIN="$(find 'target/bin' ${FIND_EXEC} -type f)"
   F_BATS="$(find 'test' -maxdepth 1 -type f -iname '*.bats')"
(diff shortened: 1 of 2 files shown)


casperklein:
I am wondering why you use ZSH with BASH scripts, when you encounter problems. Using BASH instead would be the obvious solution.

NorseGaud:
It's default for macOS

casperklein:
But BASH is available and there is nothing preventing you to use it? (just curious, not using MacOS myself).

casperklein:
Okay, afaik that feature was introduced in BASH 4.4

casperklein:
Not directly related to this At the best, linting/building etc. should take place in a container. That way, we could get rid of all OS/Shell version specific adjustments.

test/linting/lint.sh
```diff
@@ -105,7 +104,7 @@ function _shellcheck
     -not -path './target/docker-configomat/*'
   )"
   # macOS lacks parity for `-executable` but presently produces the same results: https://stackoverflow.com/a/4458361
-  [[ "$(uname)" == "Darwin" ]] && FIND_EXEC="-perm +111 -type l -or" || FIND_EXEC="-executable"
+  [[ "$(uname)" == "Darwin" ]] && FIND_EXEC="-perm -711" || FIND_EXEC="-executable"
```
casperklein, line 107: This change isn't mentioned in the PR description. Can you add a short description for future reference?

casperklein (approved):

casperklein:
```
bash-5.1$ make lint
./test/linting/lint.sh: line 22: shopt: inherit_errexit: invalid shell option name
```

I just recognized, that bash 5.1 > 4.4 😆 So the feature should be available.

What happens, when you run `shopt -s inherit_errexit` directly in your bash shell?

NorseGaud:
Hi @casperklein, I figured it out. In our contributor guide, I see that we're telling users to install brew's bash (5.x). This is installed into /usr/local/bin and works when doing `bash --version`. However, Apple still has `/bin/bash` available and that's version 3.x. 

This is because of `#! /bin/bash` in the top of the lint script. We should use `#!/usr/bin/env bash` instead so I'll PR for that.

casperklein:
> Not directly related to this: At the best, linting/building etc. should take place in a container. That way, we could get rid of all OS/Shell version specific adjustments.

For everyone interested in this, I hacked something together. Take a look at https://github.com/docker-mailserver/docker-mailserver/issues/2449 and give feedback, if that works for you.

PR #2448, opened 2022-03-02
macos support: lint.sh doesn't find proper bash under /usr/local/bin + acme_extract (python) is trying to be shellchecked

NorseGaud:
# Description

1. `shopt -s inherit_errexit` is only a bash >= 4.x feature. Apple still hasn't updated bash from 3.x, but you can (and we tell contributors to in our docs) install 5.x. This installs into /usr/local/bin. However, when I go to run `make lint`, it fails at the shopt line.

2. acme_extract is being scanned by shellcheck due to permissions differences and throwing 

```
In target/bin/acme_extract line 1:
#!/usr/bin/env python3
^-- SC1071 (error): ShellCheck only supports sh/bash/dash/ksh scripts. Sorry!
```

## Type of change

## Checklist:


test/linting/lint.sh
```diff
@@ -1,4 +1,4 @@
-#! /bin/bash
+#!/usr/bin/env bash
```
casperklein, line 1: IMO this is the more common syntax. However all of our scripts have a space before the path, so we should do the same here..

```suggestion
#! /usr/bin/env bash
```

wernerfred (approved):
LGTM

casperklein (approved):

PR #2456, opened 2022-03-03
fix: remove SA reload for KAM

georglauterbach:
# Description

Reading https://github.com/docker-mailserver/docker-mailserver/issues/2412#issuecomment-1041529047 I should have intervened when the SA reload line was added. The comments
says:

``` QUOTE
Note: SpamAssassin runs under Amavis  (so there isn't a "spamd" process)
/etc/spamassassin/sa-update-hooks.d/amavisd-new   takes care of reloading SpamAssassin after rules are updated
So, there is no point running "/etc/init.d/spamassassin reload", because it will fail   ("/etc/cron.daily/spamassassin" runs this command, but fails)
```

Fixes an error reported by cron that the update failed (which is expected...)

## Type of change

## Checklist:


casperklein (approved):

target/scripts/startup/setup-stack.sh
```diff
@@ -940,7 +940,6 @@ function _setup_security_stack
 #! /bin/bash
 
 sa-update --gpgkey 24C063D8 --channel kam.sa-channels.mcgrail.com && \
```
polarathene, line 942: `&& \` still relevant?

```suggestion
sa-update --gpgkey 24C063D8 --channel kam.sa-channels.mcgrail.com
```

polarathene (changes requested):

casperklein (approved):

polarathene (approved):

ghnp5:
Ah... I was going to report this :)

```
/etc/cron.daily/aaa_spamassassin_KAM:
Update available for channel kam.sa-channels.mcgrail.com: 1646438287 -> 1646497491
http: (curl) GET http://sa-update-kam.snb.it/1646497491.tar.gz, success
http: (curl) GET http://sa-update-kam.snb.it/1646497491.tar.gz.sha512, success
Update was available, and was downloaded and installed successfully
run-parts: /etc/cron.daily/spamassassin_kam exited with return code 1
```

SpamAssassin runs under Amavis, so "reload" doesn't work.

`/etc/cron.daily/spamassassin` runs "reload", but the script itself doesn't seem to report as returning exit code 1.

PR #2466, opened 2022-03-06
Remove macOS specific code / support + shellcheck should avoid python, regardless of permissions

NorseGaud:
# Description

Supporting macOS is a nightmare. Users can run a headless linux VM, mount the docker-mailserver directory into it, and run make commands if they are on macOS.

## Type of change

## Checklist:

- [ https://github.com/docker-mailserver/docker-mailserver/issues/2023 ] New and existing unit tests pass locally with my changes


NorseGaud:
I'm sort of stuck running the tests and stuff locally... Per https://github.com/docker-mailserver/docker-mailserver/issues/2467, postfix won't install due to the builder container not having a full FQDN

NorseGaud:
I made two changes in this to: 1. Avoid leaving build container around (they would pile up while testing the build) + 2. Using grep to avoid python script since my virtual machine mount forces all of the target/bin/ files to be executable and fails shellcheck.

test/linting/lint.sh
```diff
@@ -104,10 +98,8 @@ function _shellcheck
     -not -path './test/test_helper/*' \
     -not -path './target/docker-configomat/*'
   )"
-  # macOS lacks parity for `-executable` but presently produces the same results: https://stackoverflow.com/a/4458361
-  [[ "$(uname)" == "Darwin" ]] && FIND_EXEC="-perm -711" || FIND_EXEC="-executable"
   # shellcheck disable=SC2248
-  F_BIN="$(find 'target/bin' ${FIND_EXEC} -type f)"
+  F_BIN="$(grep -l /bin/bash target/bin/*)"
```
casperklein, line 102: There are only bash scripts in `target/bin`, with the exception of `acme_extract`. We should make this clear by renaming `acme_extract` to `acme_extract.py`.

Then we can use this:

```suggestion
  F_BIN="$(find 'target/bin' -type f -not -name '*.py')"
```

casperklein: BTW: `print-environment` and `setup` are not executable at the moment and therefore not covered by shellcheck? This should also be fixed with either @NorseGaud or my proposal.

polarathene: > We should make this clear by renaming `acme_extract` to `acme_extract.py`.

Won't that require changing [how it's called then](https://github.com/docker-mailserver/docker-mailserver/blob/d101d0aa0a802eabf74404a8295a0148f98f9497/target/scripts/helpers/ssl.sh#L421)? At least `acme_extract.py` instead of `acme_extract`, or via `python /usr/local/bin/acme_extract.py`?

casperklein: Yes, this:

> acme_extract.py instead of acme_extract

Makefile
```diff
@@ -11,7 +11,7 @@ VCS_VER = $(shell git describe --tags --contains --always)
 all: lint build backup generate-accounts tests clean
 
 build:
-	docker build -t $(NAME) . --build-arg VCS_VER=$(VCS_VER) --build-arg VCS_REF=$(VCS_REF)
+	docker build --force-rm -t $(NAME) . --build-arg VCS_VER=$(VCS_VER) --build-arg VCS_REF=$(VCS_REF)
```
polarathene, line 14: ~~I'm not sure about this, doesn't this re-build the image from scratch each time? I'd rather re-use layers when I can, especially for small changes such as in scripts that take no time to build a new image from.~~

**EDIT:** `--rm` is default, which does the same behaviour for successful builds already. `--force-rm` ensures that applies to failed builds (_where the containers were retained to assist with debugging_).

So I guess I was mistaken, and this doesn't change caching of past image builds, it only ensures intermediate containers (an images layers from directives in the Dockerfile) are removed during failed builds too.. Was that your intention?

It sounds like you'd still get previous image builds piling up in storage, perhaps you wanted `--no-cache`? (_which I don't since I prefer build performance and a VM that I discard afterwards for cleanup_)

If you do want `--no-cache` perhaps have a separate `clean-build` command?

polarathene (dismissed):

Makefile
NorseGaud: Hmm, not sure. I just want failed builds to not pile up and force me to manually delete them. Let me think about it

NorseGaud: Reverting this to propose it in another PR

polarathene (approved):

polarathene (approved):

casperklein (approved):

georglauterbach (approved):
I'm currently on vacation in Madeira, therefore only had a quick look, but LGTM

PR #2468, opened 2022-03-06
Building on Ubuntu 21.10 failing to install postfix

NorseGaud:
# Description

Fixes https://github.com/docker-mailserver/docker-mailserver/issues/2467

This is a workaround until postfix gets this bug fixed.

## Type of change

## Checklist:


NorseGaud:
Original issue: https://github.com/docker-mailserver/docker-mailserver/issues/2023

Dockerfile
```diff
@@ -41,7 +41,12 @@ RUN \
   apt-get -qq update && \
   apt-get -qq install apt-utils 2>/dev/null && \
   apt-get -qq dist-upgrade && \
+  echo "applying patch for https://github.com/docker-mailserver/docker-mailserver/issues/2023" && \
```
polarathene, line 44: This lacks context when referencing my issue as it's not a patch I detail.

Modify your comment referencing this PR about your workaround approach, or at the very least link to your comment in my referenced issue.

Dockerfile
```diff
@@ -41,7 +41,12 @@ RUN \
...
+  mv /bin/hostname{,.bak} && \
+  echo "echo docker.mail.server" > /bin/hostname && \
```
polarathene, line 46: `.server` may not be a valid TLD presently, but it's not reserved either. Do not use this value. Use a [reserved TLD](https://en.wikipedia.org/wiki/Top-level_domain#Reserved_domains) that we know will not resolve to any domain in future (_even if that's not likely to be concern during build_).

I suggest `.invalid` as appropriate here. `dms.invalid` or `docker-mailserver.invalid` should be fine.

Dockerfile
```diff
@@ -41,7 +41,12 @@ RUN \
...
+  chmod +x /bin/hostname && \
   apt-get -qq install postfix && \
+  mv /bin/hostname{.bak,} && \
```
polarathene, line 49: Personal preference, I'd rather we split the bounds of this patch with Postfix install into a separate `RUN`.

That's not required for this PR, but would be better for maintenance imo when the scope of relevant lines is clearer, layers is not a concern if we adopt multi-stage build.

The whole `RUN` chaining is primarily to minimize layers fattening an image with redundant data, but that's no longer a concern to worry about with multi-stage builds.

polarathene (changes requested):
Please update the issue reference to a relevant comment and use a reserved TLD (`.invalid`) instead of an unregistered one.

Dockerfile
NorseGaud: > This lacks context when referencing my issue as it's not a patch I detail.

I used your code from https://github.com/docker-mailserver/docker-mailserver/issues/2023#issuecomment-855326403 -- what do you mean when you say it's not a patch I detail?

Dockerfile
NorseGaud: I agree, but do `update` and `apt-utils` need to install first? Or can postfix be installed above them all?

polarathene: I vaguely remember looking into this and I think it may have been necessary. I used a Debian base image at the time and shelled into it, then ran the commands to see what would work.

I also see that I detailed this same workaround as one of the alternatives we could use if fixing it within the Dockerfile :sweat_smile:

NorseGaud: No worries, you just had me confused for a second :)

NorseGaud: Wait a sec, was that a reply to this thread, or https://github.com/docker-mailserver/docker-mailserver/pull/2468#discussion_r820312198 ?

Dockerfile
polarathene: My apologies, I wrote that ages ago and hadn't read through all my collapsed parts. Just saw that I suggested it as a viable workaround.

My issue was more of, how quickly can a maintainer find the relevant context from that issue and my verbose details. It should be easier for a maintainer to lookup the relevant context for the patch, so perhaps?:

```
Applying Dockerfile workaround detailed in: https://github.com/docker-mailserver/docker-mailserver/issues/2023#issuecomment-855326403
```

Dockerfile
polarathene: It was a reply regarding `update` and `apt-utils`, I glanced through my comments on the reference issue to see if I detailed that to help answer you here, and noticed hidden away I had detailed this workaround (just with a different hostname value).

Dockerfile
```diff
@@ -41,7 +41,12 @@ RUN \
   apt-get -qq update && \
   apt-get -qq install apt-utils 2>/dev/null && \
   apt-get -qq dist-upgrade && \
+  echo "applying workaround for ubuntu/postfix bug described in https://github.com/docker-mailserver/docker-mailserver/issues/2023" && \
```
polarathene:
```suggestion
  echo "Applying Dockerfile workaround detailed in: https://github.com/docker-mailserver/docker-mailserver/issues/2023#issuecomment-855326403" && \
```

polarathene (changes requested):

Dockerfile
NorseGaud: I like my version better

NorseGaud:
To the team: Do we really want to have a workaround like this? I prefer ubuntu server, but I could switch to Debian I guess...

polarathene:
FWIW, it appears there's another option if using BuildKit with an ENV to provide a temporary hostname to the image builder, but comments near the end also mention it being less straight-forward to use than `docker build` and a possible cache issue (_I've not tried this approach myself_): https://github.com/moby/buildkit/pull/2373

polarathene:
> I prefer ubuntu server, but I could switch to Debian I guess...

It's a VM right? The docker host side change is fairly simple to add and for a throwaway VM not really an issue.

I don't think the workaround this PR uses is going to cause any issues, so long as something else isn't trying to use `hostname` differently. Regardless, since it's only for the build phase, the hostname is a temporary hexadecimal ID, so shouldn't make any real difference at container run-time?

Debian will presumably run into the same problem by it's next release, or they fix their postfix package. I wouldn't mind us using a newer version of Postfix either though instead of the package if someone wants to add that (_good if we adopt a multi-stage build?_).

NorseGaud:
> FWIW, it appears there's another option available now.
> 
> If using BuildKit with an ENV to provide a temporary hostname to the image builder.. however comments near the end also mention it being less straight-forward to use than `docker build` and a possible cache issue (_I've not tried this approach myself_): [moby/buildkit#2373](https://github.com/moby/buildkit/pull/2373)

Yep, I tried it. It doesn't work. Postfix still uses `localhost` for myhostname (the /etc/hosts has `127.0.0.1. localhost dms.invalid`).  :(

Dockerfile
polarathene: I don't think it's an `ubuntu` bug specifically. But as the issue link has been updated to reference a relevant comment, it's good enough for me :+1:

polarathene (approved):

NorseGaud:
> and for a throwaway VM not really an issue.

*gasp*, what did you call my special VM!? ;) I plan on keeping this VM around for any future changes I make.

casperklein:
> gasp, what did you call my special VM!? ;) I plan on keeping this VM around for any future changes

You mentioned in an other issue, that you mount DMS to your VM (which introduce permission problems). You could place DMS directly inside your VM. In case you are using VS Code, take a look at [remote-ssh](https://code.visualstudio.com/docs/remote/ssh-tutorial). With that, you can easily work with the files inside your VM.

casperklein:
> To the team: Do we really want to have a workaround like this? I prefer ubuntu server, but I could switch to Debian I guess...

For me it's a bit like the MacOS specific hacks. We have a supported base, in this case Debian. If it also works on other OS's fine, but I am unsure if we should take care for it (introducing workarounds).

polarathene:
> For me it's a bit like the MacOS specific hacks. We have a supported base, in this case Debian. If it also works on other OS's fine, but I am unsure if we should take care for it (introducing workarounds).

I'm happy to just go with my original proposal of documenting it. I imagine most aren't regularly building the Docker image, and if they need to they could make the host side workaround I've detailed, or build the image within a Debian guest VM (Ubuntu 20.04 should work too AFAIK since that's what our CI is using, but newer versions of Ubuntu won't).

The next Debian release isn't due until sometime in 2023 apparently, so it may be a while before it affects Debian and they fix the package patches on their end. The workaround in this PR isn't much to worry about either though AFAIK. We have other options too but those require more effort on our end (such as building Postfix ourselves).

georglauterbach:
Just chiming in here as I'm building on Ubuntu 21.10 too, without problems. My `/etc/resolv.conf`:

``` TXT
# This is /run/systemd/resolve/stub-resolv.conf managed by man:systemd-resolved(8).
# Do not edit.

nameserver 127.0.0.53
options edns0 trust-ad
search Speedport_W_724V_09011603_06_010
```

I'm not too sure on this change either, seems hacky, but I would not object either.

polarathene:
> Just chiming in here as I'm building on Ubuntu 21.10 too, without problems.

You would only have problems if you did not have that `search` line, as my solution in the [original issue](https://github.com/docker-mailserver/docker-mailserver/issues/2023#issuecomment-855347932) detailed for fixing on the host instead of within the Dockerfile.

georglauterbach:
> > Just chiming in here as I'm building on Ubuntu 21.10 too, without problems.
> 
> You would only have problems if you did not have that `search` line, as my solution in the [original issue](https://github.com/docker-mailserver/docker-mailserver/issues/2023#issuecomment-855347932) detailed for fixing on the host instead of within the Dockerfile.

I see. I think there are actually many contributors to this project that use Ubuntu. Therefore, I'd (although this being hacky) vote for merging this. We should also document it and add a note about removing this problem if there are upstream changes that fix this...

casperklein:
Host issues should be fixed on the host IMO and we should keep our code base free of workarounds. Therefore I vote against this addition. But feel free to outvote me - it's fine 😄

polarathene:
> Host issues should be fixed on the host IMO

It's actually an issue with the Debian post-install package making assumptions. If we compiled Postfix ourselves instead of using the package, or patched the package before installing it, we wouldn't have the issue either.

The only reason I am in more support of a workaround in the Dockerfile is because that's more of a viable solution than requiring affected users to modify their DNS resolution on host. It's probably unlikely when working with Docker, but in some environments the user may have the ability to run `docker-mailserver` image, but not modify the Docker host as much, like our CI for example.

casperklein:
Convinced 👍

PS: So if I understand correctly, we would have the some problem, when our CI upgrades to a newer Ubuntu version, right?

polarathene:
> when our CI upgrades to a newer Ubuntu version, right?

Possibly, it would depend if Github CI instances have the DNS configured with a `search` field like @georglauterbach showed above. If that is missing, then the image build will fail as the package post-install step fails to resolve a DNS query IIRC (_that's not actually of any real importance to us_).

NorseGaud:
@georglauterbach do I need to change the base branch since this is slotted for 11.0? I'm moving over the next week so my replies will be delayed. Feel free to merge it once approved!

casperklein:
Can anyone explain, why adding `search .` is a good idea? I couldn't find anything about this. I thought to find something, if this is common best-practice.

georglauterbach:
@NorseGaud feel free to just merge this into `master´ once approved :)

casperklein:
I saw this too, but I found nothing about best-practicing, adding `search .`.

polarathene:
> I couldn't find anything about this. I thought to find something, if this is common best-practice.

> _A DNS search domain is also different. When you query a name that is only a single label — a domain without any dots — a search domain gets appended to your query._
> 
> _For example, because I’m currently connected to my Red Hat VPN, I have a search domain configured for `redhat.com`. This means that if I make a query to a domain that is only a single label, `redhat.com` will be appended to the query._
> 
> _For example, I can query `bugzilla` and this will be treated as a query for `bugzilla.redhat.com`. This probably won’t work in your web browser, because web browsers like to convert single-label domains into web searches, but it does work at the DNS level._
> 
> [Source](https://blogs.gnome.org/mcatanzaro/2020/12/17/understanding-systemd-resolved-split-dns-and-vpn-configuration/)

> _Search domain means the domain that will be automatically appended when you only use the hostname for a particular host or computer. This is basically used in a local network._
> 
> _Lets say you have a domain name like `xyz.com` (it may be available globally or may be local only) and you have 100 computers in the LAN. Now you want this domain name to be automatically appended when you look for any computer by just hostname of the computer. _
> 
> _So, what search domain is doing in our case is that it is automatically appending a domain name to make it a FQDN when we are just using the hostname to look up a computer._
> 
> [Source](https://askubuntu.com/a/584059)

- [`resolved.conf` docs](https://www.freedesktop.org/software/systemd/man/resolved.conf.html).
- [`systemd-resolved` docs](https://www.freedesktop.org/software/systemd/man/systemd-resolved.service.html).

A single DNS label is not valid for DNS query, only mDNS.

---

Oh you were specifically asking about `search .`.. :sweat_smile: 

My linked issue about it references the [associated systemd update changelog](https://github.com/systemd/systemd/blob/33f2de7b6487c7fe7f45fd9d8fc73cd98e0a53ac/NEWS#L1341) which describes why they add `search .` by default under some conditions (_so the problem was more to do with that specific line, not the omission of `search` my bad_), and in an earlier message on that issue thread I also in a collapsed details section covered what was going on with the [Postfix post-install step](https://github.com/docker-mailserver/docker-mailserver/issues/2023#issuecomment-855326403) which results in an invalid FQDN value to configure postfix for.

`search .` is not best practice per se AFAIK, but systemd will add it to adjust the behaviour of `nss-dns` which otherwise implicitly uses the hostname as a search domain apparently?

georglauterbach:
Now I see the error on one of my devices as well, pretty annoying...

casperklein:
I thought, systemd would just add the search option to `/etc/resolve.conf` on boot if it does not exist. Therefore a container does have the same `/etc/resolve.conf` as the host.

But even if you use a custom `/etc/resolve.conf` without search option on the host (not generated by systemd on boot), systemd still adds `search .` to the containers `/etc/resolve.conf` 🤦‍♂️

```
root@ubuntu2110:/# grep search /etc/resolv.conf && \
                                  echo --- && \
                                  docker run -it debian:11-slim grep search /etc/resolv.conf
#search foo.bar
---
search .
```

systemd: Just mapping `/etc/resolv.conf` 1:1 to the container would be too easy.. Let's fuck it up, despite what the user actually want 🖕

polarathene:
> Therefore a container does have the same `/etc/resolve.conf` as the host.

If you want that, you can do so with host mode networking. systemd should not be doing anything to the containers config from the host, that's probably Docker meddling with it.

casperklein:
Sorry for asking again. But you said, that the addition of `search .` to the containers `/etc/resolve.conf` is a systemd (new version) thing? But then you said, it may be a docker thing?  I used the same latest Docker version on all tests systems.

Let me summarize it:

| OS                                                                       | Docker | systemd | Host resolve.conf      | Container resolve.conf    |
|--------------------------------------------------------------------------|--------|---------|------------------------|---------------------------|
| Ubuntu LTS                                                               | latest | old     | No `search .` is added | `search .` is not present |
| Ubuntu 2110                                                              | latest | new     | `search .` is added    | `search .` is present     |
| Ubuntu 2110 (custom resolve.conf without `search .`, not auto generated) | latest | new     | No `search .` is added | `search .` is present     |
| Debian 11                                                                | latest | old     | No `search .` is added | `search .` is not present |

I just try to finally understand, who's the bad guy and does weird things. From the table above it looks like, it's systemd?

polarathene:
> But then you said, it may be a docker thing?

I said systemd only modifies the system config, it will not interfere with the container. The container may "import" the system config, especially with host mode networking, otherwise Docker daemon is doing it's own modification separately, such as setting hostname to the container name/id.

The newer version of systemd as per the changelog adds `search .` if needed to prevent implicit `nss-dns` fallback behaviour. This should only be happening on the host though, so if it only happens in the container, it's probably something going on with Docker.

---

With your custom change on host, did you just modify the file, or configure it so systemd was re-generating `resolv.conf` without it? Perhaps without making systemd aware of it, Docker is sourcing the `resolv.conf` to use in the container differently?

If you use host networking, perhaps it will use the modified one on your host, I just recall systemd adding a comment about `resolv.conf` being generated by systemds own config, and that [my workaround](https://github.com/docker-mailserver/docker-mailserver/issues/2023#issuecomment-855326403) that changed the search domain (at `/etc/systemd/resolved.conf`) required restarting the relevant systemd service `systemd-resolved`.

If you apply a change to search domain like that, it probably carries of into the guest container. Perhaps there's a way to prevent injecting the `search .` line as well without providing an alternative search domain, I just didn't bother since `test` is reserved and won't go to a public DNS resolver anyway.

casperklein:
I didn't use host networking, just the default network mode docker provides.

> did you just modify the file, or configure it so systemd was re-generating [..]

I've already killed the test VM, so I cannot take a detailed look. But `/etc/resolv.conf` was symlinked to systemd's autogenerated file (which was in some systemd directory). I've deleted the symlink and created my own `resolv.conf`. I verified with `dig`, that the host uses `resolv.conf` created by me.

Maybe Docker don't cares for `/etc/resolv.conf` and uses systemd's autogenerated file directly?

Enough time invested 😆 Thanks for your help.

polarathene:
> > FWIW, it appears there's another option available now.
> > If using BuildKit with an ENV to provide a temporary hostname to the image builder.. however comments near the end also mention it being less straight-forward to use than `docker build` and a possible cache issue (_I've not tried this approach myself_): [moby/buildkit#2373](https://github.com/moby/buildkit/pull/2373)
> 
> Yep, I tried it. It doesn't work. Postfix still uses `localhost` for myhostname (the /etc/hosts has `127.0.0.1. localhost dms.invalid`). :(

[Recently looked into this](https://github.com/docker-mailserver/docker-mailserver/pull/2755#discussion_r959102147) and verified it works:

- [`BUILDKIT_SANDBOX_HOSTNAME`](https://docs.docker.com/engine/reference/builder/#buildkit-built-in-build-args) is a `--build-arg`, not ENV for the command :sweat_smile: 
- `# syntax=docker/dockerfile:1` must be added to the `Dockerfile` to enable the support. The feature is part of the `1.4` [Dockerfile syntax _directive_](https://docs.docker.com/engine/reference/builder/#syntax) that [requires BuildKit `0.10.0` (March 2022)](https://github.com/moby/buildkit/releases/tag/dockerfile%2F1.4.0) or newer.
- Usage: `docker buildx build --build-arg BUILDKIT_SANDBOX_HOSTNAME=docker.invalid --tag test-image .`
  - The `hostname` command will now output `docker.invalid` instead of `buildkitsandbox` (_default for BuildKit, or without BuildKit a hexadecimal ID_)
  - While `/etc/hostname` content is unaffected (_BuildKit defaults to `localhost`, without BuildKit a different hexadecimal ID_).

PR #2485, opened 2022-03-14
scripts: small refactorings

georglauterbach:
# Description

The commit descriptions are _somewhat_ detailed. It basically boils down to:

1. Cron scripts are now added with HEREDOCs
2. Minor style adjustments

## Type of change

- [x] Improvement (non-breaking change that does improve existing functionality)

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [x] If necessary I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #2486, opened 2022-03-15
scripts: update log

georglauterbach:
# Description

Cherry picked from commits in #2485, this PR holds all the log updates. The commit descriptions is somewhat detailed.

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/check-for-changes.sh
+++ b/target/scripts/check-for-changes.sh
@@ -6,37 +6,34 @@
 # shellcheck source=./helpers/index.sh
 source /usr/local/bin/helpers/index.sh
 
-function _log_date
-{
-  date +"%Y-%m-%d %H:%M:%S"
-}
-
 function _changedetector_notify
 {
-  local LOG_LEVEL_ARGUMENT="${1}"
+  local LEVEL="${1}"
   shift 1
 
-  _notify "${LOG_LEVEL_ARGUMENT}" '[ CHANGEDETECTOR ]' " $(_log_date) " "${*}"
+  _notify "${LEVEL}" '[ CHANGEDETECTOR ]' " $(date +"%Y-%m-%d %H:%M:%S") " "${*}"
 }
 
-LOG_DATE=$(_log_date)
-_changedetector_notify 'debug' "${LOG_DATE} Start check-for-changes script."
+_changedetector_notify 'debug' 'Starting changedetector'
 
 # ? --------------------------------------------- Checks
 
-cd /tmp/docker-mailserver || exit 1
+if ! cd /tmp/docker-mailserver &>/dev/null
+then
+  _changedetector_notify 'error' "Could not change into '/tmp/docker-mailserver' directory"
+  exit 1
+fi
 
-# check postfix-accounts.cf exist else break
 if [[ ! -f postfix-accounts.cf ]]
 then
-  _changedetector_notify 'info' "${LOG_DATE} postfix-accounts.cf is missing! 'check-for-changes.sh' will exit."
+  _changedetector_notify 'info' "postfix-accounts.cf is missing! 'check-for-changes.sh' will exit."
   exit 0
 fi
 
-# verify checksum file exists; must be prepared by start-mailserver.sh
+# checksum file must have been prepared by start-mailserver.sh
 if [[ ! -f ${CHKSUM_FILE} ]]
 then
-  _changedetector_notify 'info' "${LOG_DATE} ${CHKSUM_FILE} is missing! Start script failed? Exit!"
+  _changedetector_notify 'info' "${CHKSUM_FILE} is missing! 'check-for-changes.sh' will exit."
   exit 0
 fi
 
@@ -47,7 +44,7 @@ fi
 _obtain_hostname_and_domainname
 
 PM_ADDRESS="${POSTMASTER_ADDRESS:=postmaster@${DOMAINNAME}}"
-_changedetector_notify 'trace' "${LOG_DATE} Using postmaster address ${PM_ADDRESS}"
+_changedetector_notify 'trace' "Using postmaster address ${PM_ADDRESS}"
 
 REGEX_NEVER_MATCH="(?\!)"
 
@@ -143,7 +140,7 @@ do
       chown -R 5000:5000 /var/mail
     fi
 
-    _changedetector_notify 'debug' 'Restarting services due to detected changes..'
+    _changedetector_notify 'debug' 'Restarting services due to detected changes'
 
     supervisorctl restart postfix
 
--- a/target/scripts/check-for-changes.sh
+++ b/target/scripts/check-for-changes.sh
@@ -12,7 +12,7 @@ function _log_date
 }
 
 LOG_DATE=$(_log_date)
-_notify 'task' "${LOG_DATE} Start check-for-changes script."
+_notify 'debug' "${LOG_DATE} Start check-for-changes script."
 
 # ? --------------------------------------------- Checks
 
@@ -21,14 +21,14 @@ cd /tmp/docker-mailserver || exit 1
 # check postfix-accounts.cf exist else break
 if [[ ! -f postfix-accounts.cf ]]
 then
-  _notify 'inf' "${LOG_DATE} postfix-accounts.cf is missing! This should not run! Exit!"
+  _notify 'info' "${LOG_DATE} postfix-accounts.cf is missing! This should not run! Exit!"
   exit 0
 fi
 
 # verify checksum file exists; must be prepared by start-mailserver.sh
 if [[ ! -f ${CHKSUM_FILE} ]]
 then
-  _notify 'err' "${LOG_DATE} ${CHKSUM_FILE} is missing! Start script failed? Exit!"
+  _notify 'info' "${LOG_DATE} ${CHKSUM_FILE} is missing! Start script failed? Exit!"
   exit 0
 fi
 
@@ -39,14 +39,14 @@ fi
 _obtain_hostname_and_domainname
 
 PM_ADDRESS="${POSTMASTER_ADDRESS:=postmaster@${DOMAINNAME}}"
-_notify 'inf' "${LOG_DATE} Using postmaster address ${PM_ADDRESS}"
+_notify 'trace' "${LOG_DATE} Using postmaster address ${PM_ADDRESS}"
 
 REGEX_NEVER_MATCH="(?\!)"
 
 # Change detection delayed during startup to avoid conflicting writes
 sleep 10
 
-_notify 'inf' "$(_log_date) check-for-changes is ready"
+_notify 'trace' "$(_log_date) check-for-changes is ready"
 
 while true
 do
@@ -60,7 +60,7 @@ do
   # 2 – inaccessible or missing argument
   if [[ ${?} -eq 1 ]]
   then
-    _notify 'inf' "$(_log_date) Change detected"
+    _notify 'trace' "$(_log_date) Change detected"
     _create_lock # Shared config safety lock
     CHANGED=$(grep -Fxvf "${CHKSUM_FILE}" "${CHKSUM_FILE}.new" | sed 's/^[^ ]\+  //')
 
@@ -87,7 +87,7 @@ do
     # extracted for `docker-mailserver` services to adjust to.
     elif [[ ${CHANGED} =~ /etc/letsencrypt/acme.json ]]
     then
-      _notify 'inf' "'/etc/letsencrypt/acme.json' has changed, extracting certs.."
+      _notify 'trace' "'/etc/letsencrypt/acme.json' has changed, extracting certs.."
 
       # This breaks early as we only need the first successful extraction.
       # For more details see the `SSL_TYPE=letsencrypt` case handling in `setup-stack.sh`.
@@ -97,7 +97,7 @@ do
       FQDN_LIST=("${SSL_DOMAIN}" "${HOSTNAME}" "${DOMAINNAME}")
       for CERT_DOMAIN in "${FQDN_LIST[@]}"
       do
-        _notify 'inf' "Attempting to extract for '${CERT_DOMAIN}'"
+        _notify 'trace' "Attempting to extract for '${CERT_DOMAIN}'"
 
         if _extract_certs_from_acme "${CERT_DOMAIN}"
         then
@@ -135,7 +135,7 @@ do
       chown -R 5000:5000 /var/mail
     fi
 
-    _notify 'inf' "Restarting services due to detected changes.."
+    _notify 'debug' "Restarting services due to detected changes.."
 
     supervisorctl restart postfix
 
@@ -143,7 +143,7 @@ do
     [[ ${SMTP_ONLY} -ne 1 ]] && supervisorctl restart dovecot
 
     _remove_lock
-    _notify 'inf' "$(_log_date) Completed handling of detected change"
+    _notify 'trace' "$(_log_date) Completed handling of detected change"
   fi
 
   # mark changes as applied
--- a/target/scripts/check-for-changes.sh
+++ b/target/scripts/check-for-changes.sh
@@ -21,7 +21,7 @@ cd /tmp/docker-mailserver || exit 1
 # check postfix-accounts.cf exist else break
 if [[ ! -f postfix-accounts.cf ]]
 then
-  _notify 'info' "${LOG_DATE} postfix-accounts.cf is missing! This should not run! Exit!"
+  _notify 'info' "${LOG_DATE} postfix-accounts.cf is missing! 'check-for-changes.sh' will exit."
   exit 0
 fi
 
--- a/target/scripts/check-for-changes.sh
+++ b/target/scripts/check-for-changes.sh
@@ -42,17 +42,18 @@ then
   _changedetector_exit_error "'/tmp/docker-mailserver/${CHKSUM_FILE}' is missing"
 fi
 
-until grep -qE "export POSTMASTER_ADDRESS='.+'" /root/.bashrc
+# sleep some more during startup to avoid conflicting writes
+sleep 10
+
+until grep -qE "export POSTMASTER_ADDRESS=.+" /root/.bashrc
 do
   sleep 5
 done
 
-# sleep some more during startup to avoid conflicting writes
-sleep 10
 _get_dms_environment_variables
 
 _changedetector_notify 'trace' "Running with postmaster address '${POSTMASTER_ADDRESS}'"
-_changedetector_notify 'trace' "Changedetector is ready"
+_changedetector_notify 'trace' 'Changedetector is ready'
 
 while true
 do
@@ -82,7 +83,7 @@ do
       || [[ ${CHANGED} =~ ${SSL_ALT_CERT_PATH:-${REGEX_NEVER_MATCH}} ]] \
       || [[ ${CHANGED} =~ ${SSL_ALT_KEY_PATH:-${REGEX_NEVER_MATCH}} ]]
       then
-        _changedetector_notify 'debug' 'Manual certificates have changed'
+        _changedetector_notify 'debug' 'Manual certificates have changed - extracting certifcates now'
         # we need to run the SSL setup again, because the
         # certificates DMS is working with are copies of
         # the (now changed) files
@@ -93,7 +94,7 @@ do
     # extracted for `docker-mailserver` services to adjust to.
     elif [[ ${CHANGED} =~ /etc/letsencrypt/acme.json ]]
     then
-      _changedetector_notify 'debug' "'/etc/letsencrypt/acme.json' has changed"
+      _changedetector_notify 'debug' "'/etc/letsencrypt/acme.json' has changed - extracting certifcates now"
 
       # This breaks early as we only need the first successful extraction.
       # For more details see the `SSL_TYPE=letsencrypt` case handling in `setup-stack.sh`.
--- a/target/scripts/check-for-changes.sh
+++ b/target/scripts/check-for-changes.sh
@@ -5,8 +5,8 @@
 
 # shellcheck source=./helpers/index.sh
 source /usr/local/bin/helpers/index.sh
-
-REGEX_NEVER_MATCH="(?\!)"
+# shellcheck source=/dev/null
+source /etc/dms-settings
 
 function _changedetector_notify
 {
@@ -18,8 +18,8 @@ function _changedetector_notify
 
 function _changedetector_exit_error
 {
-  _changedetector_notify "${@}"
-  _changedetector_notify 'Aborting'
+  _changedetector_notify 'error' "${1}"
+  _changedetector_notify 'error' 'Aborting'
 
   exit 1
 }
@@ -45,12 +45,7 @@ fi
 # sleep some more during startup to avoid conflicting writes
 sleep 10
 
-until grep -qE "export POSTMASTER_ADDRESS=.+" /root/.bashrc
-do
-  sleep 5
-done
-
-_get_dms_environment_variables
+REGEX_NEVER_MATCH="(?\!)"
 
 _changedetector_notify 'trace' "Running with postmaster address '${POSTMASTER_ADDRESS}'"
 _changedetector_notify 'trace' 'Changedetector is ready'
@@ -129,7 +124,8 @@ do
 
     _rebuild_relayhost
 
-    # regenerate postix aliases
+    # regenerate postix aliases; this needs `POSTMASTER_ADDRESS` to be set
+    # and therefore requires to call to `source /etc/dms-settings`
     _create_aliases
 
     # regenerate /etc/postfix/vhost
--- a/target/scripts/check-for-changes.sh
+++ b/target/scripts/check-for-changes.sh
@@ -1,7 +1,16 @@
 #! /bin/bash
 
-# TODO: Adapt for compatibility with LDAP
-# Only the cert renewal change detection may be relevant for LDAP?
+# About this Script
+#
+# This script checks certain files for updates and restarts services accordingly.
+# Take care when altering this script. The Bash unit tests check the log of this
+# script, and if the log messages are altered in size, you may need to adjust the
+# tests.
+#
+# TODOs
+#
+# 1. Adapt for compatibility with LDAP
+#    Only the cert renewal change detection may be relevant for LDAP?
 
 # shellcheck source=./helpers/index.sh
 source /usr/local/bin/helpers/index.sh
@@ -52,7 +61,7 @@ _changedetector_notify 'trace' 'Changedetector is ready'
 
 while true
 do
-  # get chksum and check it, no need to lock config yet
+  # get checksum and check it, no need to lock config yet
   _monitored_files_checksums >"${CHKSUM_FILE}.new"
   cmp --silent -- "${CHKSUM_FILE}" "${CHKSUM_FILE}.new"
 
@@ -62,9 +71,12 @@ do
   # 2 – inaccessible or missing argument
   if [[ ${?} -eq 1 ]]
   then
+    CHANGED=$(grep -Fxvf "${CHKSUM_FILE}" "${CHKSUM_FILE}.new" | sed 's/^[^ ]\+  //')
+
     _changedetector_notify 'info' 'Change detected'
     _create_lock # Shared config safety lock
-    CHANGED=$(grep -Fxvf "${CHKSUM_FILE}" "${CHKSUM_FILE}.new" | sed 's/^[^ ]\+  //')
+
+    [[ -z ${CHANGED} ]] && _changedetector_notify 'debug' 'Found new files'
 
     # TODO Perform updates below conditionally too
     # Also note that changes are performed in place and are not atomic
--- a/target/scripts/check-for-changes.sh
+++ b/target/scripts/check-for-changes.sh
@@ -6,51 +6,52 @@
 # shellcheck source=./helpers/index.sh
 source /usr/local/bin/helpers/index.sh
 
+REGEX_NEVER_MATCH="(?\!)"
+
 function _changedetector_notify
 {
   local LEVEL="${1}"
   shift 1
 
-  _notify "${LEVEL}" '[ CHANGEDETECTOR ]' " $(date +"%Y-%m-%d %H:%M:%S") " "${*}"
+  _notify "${LEVEL}" "(CHANGEDETECTOR) $(date +"%Y-%m-%d %H:%M:%S") ${*}"
+}
+
+function _changedetector_exit_error
+{
+  _changedetector_notify "${@}"
+  _changedetector_notify 'Aborting'
+
+  exit 1
 }
 
 _changedetector_notify 'debug' 'Starting changedetector'
 
-# ? --------------------------------------------- Checks
-
 if ! cd /tmp/docker-mailserver &>/dev/null
 then
-  _changedetector_notify 'error' "Could not change into '/tmp/docker-mailserver' directory"
-  exit 1
+  _changedetector_exit_error "Could not change into '/tmp/docker-mailserver/' directory"
 fi
 
 if [[ ! -f postfix-accounts.cf ]]
 then
-  _changedetector_notify 'info' "postfix-accounts.cf is missing! 'check-for-changes.sh' will exit."
-  exit 0
+  _changedetector_exit_error "'/tmp/docker-mailserver/postfix-accounts.cf' is missing"
 fi
 
 # checksum file must have been prepared by start-mailserver.sh
 if [[ ! -f ${CHKSUM_FILE} ]]
 then
-  _changedetector_notify 'info' "${CHKSUM_FILE} is missing! 'check-for-changes.sh' will exit."
-  exit 0
+  _changedetector_exit_error "'/tmp/docker-mailserver/${CHKSUM_FILE}' is missing"
 fi
 
-# ? --------------------------------------------- Actual script begins
+until grep -qE "export POSTMASTER_ADDRESS='.+'" /root/.bashrc
+do
+  sleep 5
+done
 
-# determine postmaster address, duplicated from start-mailserver.sh
-# this script previously didn't work when POSTMASTER_ADDRESS was empty
-_obtain_hostname_and_domainname
-
-PM_ADDRESS="${POSTMASTER_ADDRESS:=postmaster@${DOMAINNAME}}"
-_changedetector_notify 'trace' "Using postmaster address ${PM_ADDRESS}"
-
-REGEX_NEVER_MATCH="(?\!)"
-
-# Change detection delayed during startup to avoid conflicting writes
+# sleep some more during startup to avoid conflicting writes
 sleep 10
+_get_dms_environment_variables
 
+_changedetector_notify 'trace' "Running with postmaster address '${POSTMASTER_ADDRESS}'"
 _changedetector_notify 'trace' "Changedetector is ready"
 
 while true
@@ -65,7 +66,7 @@ do
   # 2 – inaccessible or missing argument
   if [[ ${?} -eq 1 ]]
   then
-    _changedetector_notify 'info' "Change detected"
+    _changedetector_notify 'info' 'Change detected'
     _create_lock # Shared config safety lock
     CHANGED=$(grep -Fxvf "${CHKSUM_FILE}" "${CHKSUM_FILE}.new" | sed 's/^[^ ]\+  //')
 
@@ -81,7 +82,7 @@ do
       || [[ ${CHANGED} =~ ${SSL_ALT_CERT_PATH:-${REGEX_NEVER_MATCH}} ]] \
       || [[ ${CHANGED} =~ ${SSL_ALT_KEY_PATH:-${REGEX_NEVER_MATCH}} ]]
       then
-        _changedetector_notify 'debug' "Manual certificates have changed, extracting certs.."
+        _changedetector_notify 'debug' 'Manual certificates have changed'
         # we need to run the SSL setup again, because the
         # certificates DMS is working with are copies of
         # the (now changed) files
@@ -92,7 +93,7 @@ do
     # extracted for `docker-mailserver` services to adjust to.
     elif [[ ${CHANGED} =~ /etc/letsencrypt/acme.json ]]
     then
-      _changedetector_notify 'debug' "'/etc/letsencrypt/acme.json' has changed, extracting certs.."
+      _changedetector_notify 'debug' "'/etc/letsencrypt/acme.json' has changed"
 
       # This breaks early as we only need the first successful extraction.
       # For more details see the `SSL_TYPE=letsencrypt` case handling in `setup-stack.sh`.
@@ -148,7 +149,7 @@ do
     [[ ${SMTP_ONLY} -ne 1 ]] && supervisorctl restart dovecot
 
     _remove_lock
-    _changedetector_notify 'trace' "Completed handling of detected change"
+    _changedetector_notify 'trace' 'Completed handling of detected change'
   fi
 
   # mark changes as applied
--- a/target/scripts/check-for-changes.sh
+++ b/target/scripts/check-for-changes.sh
@@ -11,8 +11,16 @@ function _log_date
   date +"%Y-%m-%d %H:%M:%S"
 }
 
+function _changedetector_notify
+{
+  local LOG_LEVEL_ARGUMENT="${1}"
+  shift 1
+
+  _notify "${LOG_LEVEL_ARGUMENT}" '[ CHANGEDETECTOR ]' " $(_log_date) " "${*}"
+}
+
 LOG_DATE=$(_log_date)
-_notify 'debug' "${LOG_DATE} Start check-for-changes script."
+_changedetector_notify 'debug' "${LOG_DATE} Start check-for-changes script."
 
 # ? --------------------------------------------- Checks
 
@@ -21,14 +29,14 @@ cd /tmp/docker-mailserver || exit 1
 # check postfix-accounts.cf exist else break
 if [[ ! -f postfix-accounts.cf ]]
 then
-  _notify 'info' "${LOG_DATE} postfix-accounts.cf is missing! 'check-for-changes.sh' will exit."
+  _changedetector_notify 'info' "${LOG_DATE} postfix-accounts.cf is missing! 'check-for-changes.sh' will exit."
   exit 0
 fi
 
 # verify checksum file exists; must be prepared by start-mailserver.sh
 if [[ ! -f ${CHKSUM_FILE} ]]
 then
-  _notify 'info' "${LOG_DATE} ${CHKSUM_FILE} is missing! Start script failed? Exit!"
+  _changedetector_notify 'info' "${LOG_DATE} ${CHKSUM_FILE} is missing! Start script failed? Exit!"
   exit 0
 fi
 
@@ -39,14 +47,14 @@ fi
 _obtain_hostname_and_domainname
 
 PM_ADDRESS="${POSTMASTER_ADDRESS:=postmaster@${DOMAINNAME}}"
-_notify 'trace' "${LOG_DATE} Using postmaster address ${PM_ADDRESS}"
+_changedetector_notify 'trace' "${LOG_DATE} Using postmaster address ${PM_ADDRESS}"
 
 REGEX_NEVER_MATCH="(?\!)"
 
 # Change detection delayed during startup to avoid conflicting writes
 sleep 10
 
-_notify 'trace' "$(_log_date) check-for-changes is ready"
+_changedetector_notify 'trace' "Changedetector is ready"
 
 while true
 do
@@ -60,7 +68,7 @@ do
   # 2 – inaccessible or missing argument
   if [[ ${?} -eq 1 ]]
   then
-    _notify 'trace' "$(_log_date) Change detected"
+    _changedetector_notify 'info' "Change detected"
     _create_lock # Shared config safety lock
     CHANGED=$(grep -Fxvf "${CHKSUM_FILE}" "${CHKSUM_FILE}.new" | sed 's/^[^ ]\+  //')
 
@@ -76,7 +84,7 @@ do
       || [[ ${CHANGED} =~ ${SSL_ALT_CERT_PATH:-${REGEX_NEVER_MATCH}} ]] \
       || [[ ${CHANGED} =~ ${SSL_ALT_KEY_PATH:-${REGEX_NEVER_MATCH}} ]]
       then
-        _notify 'inf' "Manual certificates have changed, extracting certs.."
+        _changedetector_notify 'debug' "Manual certificates have changed, extracting certs.."
         # we need to run the SSL setup again, because the
         # certificates DMS is working with are copies of
         # the (now changed) files
@@ -87,7 +95,7 @@ do
     # extracted for `docker-mailserver` services to adjust to.
     elif [[ ${CHANGED} =~ /etc/letsencrypt/acme.json ]]
     then
-      _notify 'trace' "'/etc/letsencrypt/acme.json' has changed, extracting certs.."
+      _changedetector_notify 'debug' "'/etc/letsencrypt/acme.json' has changed, extracting certs.."
 
       # This breaks early as we only need the first successful extraction.
       # For more details see the `SSL_TYPE=letsencrypt` case handling in `setup-stack.sh`.
@@ -97,7 +105,7 @@ do
       FQDN_LIST=("${SSL_DOMAIN}" "${HOSTNAME}" "${DOMAINNAME}")
       for CERT_DOMAIN in "${FQDN_LIST[@]}"
       do
-        _notify 'trace' "Attempting to extract for '${CERT_DOMAIN}'"
+        _changedetector_notify 'trace' "Attempting to extract for '${CERT_DOMAIN}'"
 
         if _extract_certs_from_acme "${CERT_DOMAIN}"
         then
@@ -135,7 +143,7 @@ do
       chown -R 5000:5000 /var/mail
     fi
 
-    _notify 'debug' "Restarting services due to detected changes.."
+    _changedetector_notify 'debug' 'Restarting services due to detected changes..'
 
     supervisorctl restart postfix
 
@@ -143,7 +151,7 @@ do
     [[ ${SMTP_ONLY} -ne 1 ]] && supervisorctl restart dovecot
 
     _remove_lock
-    _notify 'trace' "$(_log_date) Completed handling of detected change"
+    _changedetector_notify 'trace' "Completed handling of detected change"
   fi
 
   # mark changes as applied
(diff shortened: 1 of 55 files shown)


target/scripts/startup/setup-stack.sh
```diff
@@ -1126,10 +1127,12 @@ function _setup_fail2ban
 
 function _setup_dnsbl_disable
 {
-  _notify 'task' 'Disabling postfix DNS block list (zen.spamhaus.org)'
-  sedfile -i '/^smtpd_recipient_restrictions = / s/, reject_rbl_client zen.spamhaus.org//' /etc/postfix/main.cf
+  _notify 'debug' 'Disabling postfix DNS block list (zen.spamhaus.org)'
+  sedfile -i                                                                      \
+    '/^smtpd_recipient_restrictions = / s/, reject_rbl_client zen.spamhaus.org//' \
+    /etc/postfix/main.cf
```
polarathene, line 1133:
```suggestion
  _notify 'debug' 'Disabling postfix DNS block list (zen.spamhaus.org)'
  sedfile -i \
    '/^smtpd_recipient_restrictions = / s/, reject_rbl_client zen.spamhaus.org//' \
    /etc/postfix/main.cf
```

target/scripts/helpers/error.sh
```diff
@@ -62,13 +62,11 @@ function dms_panic__no_file { dms_panic 'no-file' "${1}" "${2}"; }
...
 {
-  local FATAL_ERROR_MESSAGE=$1
-
-  _notify 'fatal' "${FATAL_ERROR_MESSAGE}"
-  _notify 'err' "Shutting down.."
-  kill 1
+  _notify 'error' "${1}"
+  _notify 'error' 'Shutting down...'
+  exit 1
 }
```
polarathene, line 72: Was dropping `kill 1` intentional here?

polarathene (changes requested):

target/scripts/helpers/error.sh
georglauterbach: No :D

test/mail_special_use_folders.bats
```diff
@@ -4,14 +4,13 @@ setup_file() {
     local PRIVATE_CONFIG
     PRIVATE_CONFIG="$(duplicate_config_for_container .)"
     docker run -d --name mail_special_use_folders \
-		-v "${PRIVATE_CONFIG}":/tmp/docker-mailserver \
-		-v "$(pwd)/test/test-files":/tmp/docker-mailserver-test:ro \
+		            -v "${PRIVATE_CONFIG}":/tmp/docker-mailserver \
+		            -v "$(pwd)/test/test-files":/tmp/docker-mailserver-test:ro \
```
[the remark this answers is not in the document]
georglauterbach: I will correct it here as well as VS Code won't really allow me to revert it back to the mixed indentation. git should be able to merge these changes just fine.

test/mail_ssl_manual.bats
```diff
@@ -101,7 +101,7 @@ function teardown_file() {
 @test "checking ssl: manual cert changes are picked up by check-for-changes" {
     printf 'someThingsChangedHere' \
       >>"$(pwd)/test/test-files/ssl/${DOMAIN_SSL_MANUAL}/with_ca/ecdsa/key.ecdsa.pem"
-    sleep 10
+    sleep 15
```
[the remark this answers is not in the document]
georglauterbach: Ah yes...

target/scripts/startup/setup-stack.sh
```diff
@@ -1075,7 +1076,7 @@ function _setup_logwatch
   echo "MailFrom = ${LOGWATCH_SENDER}" >> /etc/logwatch/conf/logwatch.conf
 
   case "${LOGWATCH_INTERVAL}" in
-    'daily' )
+    ( 'daily' )
```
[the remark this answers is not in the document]
georglauterbach: Missed this as well, this was indeed accidental and does not belong here.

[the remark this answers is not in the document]
georglauterbach: Done in [809b2b4](https://github.com/docker-mailserver/docker-mailserver/pull/2486/commits/809b2b42a9d4bc9340e91a5a29883b19b99d574e)

target/scripts/helpers/error.sh
georglauterbach: Done in [809b2b4](https://github.com/docker-mailserver/docker-mailserver/pull/2486/commits/809b2b42a9d4bc9340e91a5a29883b19b99d574e)

georglauterbach:
I will now see why CI does not end.

test/mail_ssl_manual.bats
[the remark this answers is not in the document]
georglauterbach: Done in [809b2b4](https://github.com/docker-mailserver/docker-mailserver/pull/2486/commits/809b2b42a9d4bc9340e91a5a29883b19b99d574e)

target/scripts/helpers/error.sh
```diff
@@ -62,13 +62,11 @@ function dms_panic__no_file { dms_panic 'no-file' "${1}" "${2}"; }
...
 function _shutdown
 {
-  local FATAL_ERROR_MESSAGE=$1
-
-  _notify 'fatal' "${FATAL_ERROR_MESSAGE}"
-  _notify 'err' "Shutting down.."
+  _notify 'error' "${1}"
+  _notify 'error' 'Shutting down...'
   kill 1
 }
```
polarathene, line 72: In the earlier PR before extracting to this one, you originally had added an `exit 1` alongside the `kill 1` (_that was a little more explicit in signal used_), do you no longer see `exit 1` as beneficial here? (_I assumed `kill` was enough, but perhaps `exit` is still useful in this scenario?_)

polarathene (dismissed):
LGTM :+1:

target/scripts/helpers/error.sh
georglauterbach: Done in [b022e20](https://github.com/docker-mailserver/docker-mailserver/pull/2486/commits/b022e20c4623bc12fbff24eb224715b560124c8e)

target/scripts/check-for-changes.sh
```diff
@@ -6,37 +6,34 @@
...
-
 function _changedetector_notify
 {
-  local LOG_LEVEL_ARGUMENT="${1}"
+  local LEVEL="${1}"
   shift 1
 
-  _notify "${LOG_LEVEL_ARGUMENT}" '[ CHANGEDETECTOR ]' " $(_log_date) " "${*}"
+  _notify "${LEVEL}" '[ CHANGEDETECTOR ]' " $(date +"%Y-%m-%d %H:%M:%S") " "${*}"
 }
```
polarathene, line 15: Is there a reason to use `shift 1` and `${*}` here instead of just `${2}`? It's expected to always be two args right? (log level and message)

It seems that you are passing `[ CHANGEDETECTOR ]` and the log date as separate values, instead of wrapping each as a single message string arg?

target/scripts/check-for-changes.sh
```diff
@@ -6,37 +6,34 @@
...
+_changedetector_notify 'debug' 'Starting changedetector'
 
 # ? --------------------------------------------- Checks
 
-cd /tmp/docker-mailserver || exit 1
+if ! cd /tmp/docker-mailserver &>/dev/null
+then
+  _changedetector_notify 'error' "Could not change into '/tmp/docker-mailserver' directory"
+  exit 1
+fi
```
polarathene, line 25: Just so I understand, we're still changing the directory here, but that command is part of a negative conditional now?

Would you say the failure here warrants the `exit 1` still while the other two checks below use `exit 0`? (_and explicitly state that the script will exit in the log_) Should they all be consistent or is there a difference? The other two also use an `info` level instead of `error` which is perhaps wrong but why they had `exit 0`?

target/scripts/check-for-changes.sh
```diff
@@ -6,37 +6,34 @@
...
+  _changedetector_notify 'info' "postfix-accounts.cf is missing! 'check-for-changes.sh' will exit."
   exit 0
 fi
 
-# verify checksum file exists; must be prepared by start-mailserver.sh
+# checksum file must have been prepared by start-mailserver.sh
 if [[ ! -f ${CHKSUM_FILE} ]]
 then
-  _changedetector_notify 'info' "${LOG_DATE} ${CHKSUM_FILE} is missing! Start script failed? Exit!"
+  _changedetector_notify 'info' "${CHKSUM_FILE} is missing! 'check-for-changes.sh' will exit."
```
polarathene, line 36: These are the only two logs in this script that use a `.` at the end. I know you love things being consistent, so just pointing that out :sweat_smile:

target/scripts/check-for-changes.sh
```diff
@@ -143,7 +140,7 @@ do
       chown -R 5000:5000 /var/mail
     fi
 
-    _changedetector_notify 'debug' 'Restarting services due to detected changes..'
+    _changedetector_notify 'debug' 'Restarting services due to detected changes'
```
polarathene, line 143: There are a couple other logs in this script that also end with `..`, did you want to drop those too? They were all probably by me as a way to convey intent of logging an action that was about to be performed (_but an ellipsis for such is apparently `...` which looks weird to me_).

polarathene (changes requested):
Nothing too important here, just some observations you might care about :)

target/scripts/check-for-changes.sh
georglauterbach: `_notify' itself is written in the same way: the first argument is the log level, all the others are messages. It is therefore not actually expected to be exactly two, but at least two. This is done in order to enable one to split longer lines into multiple arguments.

However, in this case, I could indeed write these arguments as one :D

target/scripts/check-for-changes.sh
georglauterbach: > Just so I understand, we're still changing the directory here, but that command is part of a negative conditional now?

Exactly, we're still changing the directory. I added this to make it more clear why the script exited with exit code 1.
 
> Would you say the failure here warrants the `exit 1` still while the other two checks below use `exit 0`? (_and explicitly state that the script will exit in the log_) Should they all be consistent or is there a difference? The other two also use an `info` level instead of `error` which is perhaps wrong but why they had `exit 0`?

I think so, because this really seems like an unrecoverable error to me. A `notify 'error'` should always be followed by an `exit` or `return` call with non-zero exit code (to properly follow the semantics of the `error` log level).

~~I did not touch the other two statements because I really am not sure whether they are errors or just, let's say LDAP related with `postfix-accounts.cf`. I think both could indeed be errors, and this would mean adjusting the log level to `error` and `exit`ing with `1`.~~

Okay, so according to

https://github.com/docker-mailserver/docker-mailserver/blob/b730942b96aec3ab52699b5d927577f42eef0a45/target/scripts/start-mailserver.sh#L197

this cannot be LDAP related. Consequently, these have to be errors.

target/scripts/check-for-changes.sh
georglauterbach: I adjusted it :D

target/scripts/check-for-changes.sh
georglauterbach: I have made a decision here: I'd avoid `..` or `...` at the end and indicate that

1. something is about to start by writing a gerund / -ing at the end of the verb (e.g. "Starting to extract certificates")
2. something has finished by writing (simple) past (e.g. "Extracted certificates")

I adjusted `check-for-changes.sh` now, and we can continuously adjust the other scripts.

target/scripts/check-for-changes.sh
polarathene: > I think both could indeed be errors, and this would mean adjusting the log level to error and exiting with 1.

Both result in the change detector exiting early instead of running like intended. I'd consider that an error?

> this cannot be LDAP related. Consequently, these have to be errors.

Yeah, but at some point we do want to support with LDAP enabled, at least for the cert renewal support :sweat_smile:

target/scripts/check-for-changes.sh
polarathene: > I adjusted it :D

In this PR? There's no new commits since this review.

target/scripts/check-for-changes.sh
georglauterbach: Done in cce33f80d545c23eaddb67a4afb2a2bd9596e0d9

target/scripts/check-for-changes.sh
georglauterbach: Done in cce33f80d545c23eaddb67a4afb2a2bd9596e0d9

target/scripts/check-for-changes.sh
georglauterbach: Done in ee3152f346241dd0b726571a38edbad15e8fe6e8

target/scripts/check-for-changes.sh
georglauterbach: Done in cce33f80d545c23eaddb67a4afb2a2bd9596e0d9

georglauterbach:
With cce33f80d545c23eaddb67a4afb2a2bd9596e0d9, I improved `check-for-changes.sh` a bit. I know this is not actually related to the log, but I figured the change was small enough to fit in here. I hope you don't mind it :)

georglauterbach:
@polarathene I need your help:

The test `mail_ssl_letsencrypt.bats` does not work. The last call of `_acme_wildcard` in the test `ssl(letsencrypt): Traefik 'acme.json' (*.example.test)` does not succeed, because the message `/etc/letsencrypt/acme.json' has changed - extracting certifcates now` is not logged. I spent more than an hour trying to debug this. To me, it seems as if I found a false positive by viewing too much log in the past.

``` INI
[   INF   ]  (CHANGEDETECTOR) 2022-03-17 13:47:28 Change detected
[  TRACE  ]  Creating lock /tmp/docker-mailserver/check-for-changes.sh.lock
[  TRACE  ]  Checking file line endings
[  TRACE  ]  Regenerating postfix user list
[  DEBUG  ]  Creating user 'user1' for domain 'localhost.localdomain'
[  DEBUG  ]  Creating user 'user2' for domain 'otherdomain.tld'
[  DEBUG  ]  Creating user 'user3' for domain 'localhost.localdomain' with attributes 'userdb_mail=mbox:~/mail:INBOX=~/inbox'
[  DEBUG  ]  Adding alias 'alias1@localhost.localdomain' for user 'user1@localhost.localdomain' to Dovecot's userdb
[  DEBUG  ]  Alias 'alias2@localhost.localdomain' is non-local (or mapped to a non-existing account) and will not be added to Dovecot's userdb
[  DEBUG  ]  Adding alias '@localdomain2.com' for user 'user1@localhost.localdomain' to Dovecot's userdb
[  TRACE  ]  Adding regexp alias file postfix-regexp.cf
virtual_alias_maps = texthash:/etc/postfix/virtual pcre:/etc/postfix/regexp
[  TRACE  ]  Configuring root alias
[  DEBUG  ]  (CHANGEDETECTOR) 2022-03-17 13:47:28 Restarting services due to detected changes
postfix: stopped
postfix: started
dovecot: stopped
dovecot: started
[  TRACE  ]  Removed lock /tmp/docker-mailserver/check-for-changes.sh.lock
[  TRACE  ]  (CHANGEDETECTOR) 2022-03-17 13:47:31 Completed handling of detected change
```

georglauterbach:
@polarathene indeed, the log in master of the changedetector for this test looks like this:

``` LOG
[[ TASKS ]]  2022-03-17 16:11:22 Start check-for-changes script.
[[  INF  ]]  2022-03-17 16:11:22 Using postmaster address postmaster@example.test
[[  INF  ]]  2022-03-17 16:11:32 check-for-changes is ready
[[  INF  ]]  2022-03-17 16:11:36 Change detected
[[  INF  ]]  '/etc/letsencrypt/acme.json' has changed, extracting certs..
[[  INF  ]]  Attempting to extract for '*.example.test'
[ WARNING ]  _extract_certs_from_acme | Unable to find key and/or cert for '*.example.test' in '/etc/letsencrypt/acme.json'
[[  INF  ]]  Attempting to extract for 'mail.example.test'
[[  INF  ]]  _extract_certs_from_acme | Certificate successfully extracted for 'mail.example.test'
[[  INF  ]]  Checking file line endings
[[  INF  ]]  Regenerating postfix user list
[[  INF  ]]  Creating user 'user1' for domain 'localhost.localdomain'
[[  INF  ]]  Creating user 'user2' for domain 'otherdomain.tld'
[[  INF  ]]  Creating user 'user3' for domain 'localhost.localdomain' with attributes 'userdb_mail=mbox:~/mail:INBOX=~/inbox'
[[  INF  ]]  Adding regexp alias file postfix-regexp.cf
virtual_alias_maps = texthash:/etc/postfix/virtual pcre:/etc/postfix/regexp
[[  INF  ]]  Configuring root alias
[[  INF  ]]  Restarting services due to detected changes..
postfix: stopped
postfix: started
dovecot: stopped
dovecot: started
[[  INF  ]]  Removed lock /tmp/docker-mailserver/check-for-changes.sh.lock.
[[  INF  ]]  2022-03-17 16:11:44 Completed handling of detected change
[[  INF  ]]  2022-03-17 16:11:48 Change detected
[[  INF  ]]  '/etc/letsencrypt/acme.json' has changed, extracting certs..
[[  INF  ]]  Attempting to extract for '*.example.test'
[[  INF  ]]  _extract_certs_from_acme | Certificate successfully extracted for 'example.test'
[[  INF  ]]  Checking file line endings
[[  INF  ]]  Regenerating postfix user list
[[  INF  ]]  Creating user 'user1' for domain 'localhost.localdomain'
[[  INF  ]]  Creating user 'user2' for domain 'otherdomain.tld'
[[  INF  ]]  Creating user 'user3' for domain 'localhost.localdomain' with attributes 'userdb_mail=mbox:~/mail:INBOX=~/inbox'
[[  INF  ]]  Adding regexp alias file postfix-regexp.cf
virtual_alias_maps = texthash:/etc/postfix/virtual pcre:/etc/postfix/regexp
[[  INF  ]]  Configuring root alias
[[  INF  ]]  Restarting services due to detected changes..
postfix: stopped
postfix: started
dovecot: stopped
dovecot: started
[[  INF  ]]  Removed lock /tmp/docker-mailserver/check-for-changes.sh.lock.
[[  INF  ]]  2022-03-17 16:11:53 Completed handling of detected change
```

and there is a third ` '/etc/letsencrypt/acme.json' has changed, extracting certs..` somehow missing. It could be that there are only 2, which would indicate that actually all is well but we're not looking at enough log. Are there overall only 2 changes with this test? Have I misunderstood it the whole time?

target/scripts/check-for-changes.sh
```diff
@@ -6,47 +6,54 @@
...
+  sleep 5
+done
 
-# Change detection delayed during startup to avoid conflicting writes
-sleep 10
+_get_dms_environment_variables
 
-_notify 'inf' "$(_log_date) check-for-changes is ready"
+_changedetector_notify 'trace' "Running with postmaster address '${POSTMASTER_ADDRESS}'"
+_changedetector_notify 'trace' 'Changedetector is ready'
```
polarathene, line 56: I read the [associated commit message](https://github.com/docker-mailserver/docker-mailserver/pull/2486/commits/cce33f80d545c23eaddb67a4afb2a2bd9596e0d9):

> The env variable `POSTMASTER_ADDRESS` was set in a way that is not very clear to the reader and indeed possibly wrong.
> The old way used the default while the new way uses the correct variable from `/root/.bashrc`.
> 
> Therefore, a new function `_get_dms_environment_variables` was introduced which handles exporting vaiables from `/root/.bashrc`. This could be used in the future as well.

In what case would the value be different? We're still presently in sync with `start-mailserver.sh`:

https://github.com/docker-mailserver/docker-mailserver/blob/321ae744fa873ca300af83983b4f916ac8dc9c35/target/scripts/start-mailserver.sh#L10-L19

Is there a change that is made to `/root/.bashrc` elsewhere? In what case is it wrong?

You also state that it's not very clear with the way it's being set, is that something that needs to be improved at `start-mailserver.sh` then too? How is this change more clear? With the previous content at least we had a comment stating that `check-for-changes.sh` would not work if `POSTMASTER_ADDRESS` was empty hence the helper script call to ensure `DOMAINNAME` was assigned for the fallback if no `POSTMASTER_ADDRESS` is set in container ENV, just like `start-mailserver.sh`?.

Now we're sleeping 5 or more seconds to depend on extracting what should be the same value?

---

Then you've added a new helper method `_get_dms_environment_variables` that interacts with `/root/.bashrc`.. but from what I understand casper introduced `/etc/dms-settings` to source for this specific purpose, it is used in `target/bin/` scripts at present. Aren't you duplicating logic with this method?

https://github.com/docker-mailserver/docker-mailserver/blob/321ae744fa873ca300af83983b4f916ac8dc9c35/target/scripts/startup/setup-stack.sh#L41-L57

That method is called early on in `start-mailserver.sh`:

https://github.com/docker-mailserver/docker-mailserver/blob/321ae744fa873ca300af83983b4f916ac8dc9c35/target/scripts/start-mailserver.sh#L102

and later after all the setup, we start the `check-for-changes.sh` service:

https://github.com/docker-mailserver/docker-mailserver/blob/321ae744fa873ca300af83983b4f916ac8dc9c35/target/scripts/start-mailserver.sh#L197

https://github.com/docker-mailserver/docker-mailserver/blob/321ae744fa873ca300af83983b4f916ac8dc9c35/target/scripts/startup/daemons-stack.sh#L150-L154

https://github.com/docker-mailserver/docker-mailserver/blob/321ae744fa873ca300af83983b4f916ac8dc9c35/target/scripts/start-mailserver.sh#L263-L276

So I'm not sure why we'd be waiting on such an ENV with `sleep` when by the time this script runs, the dependency should already be written. Neither location should be using remote storage/volume causing latency issues AFAIK, both would be using the internal container write layer?

---

If you're concerned about maintenance with potential for a desync, perhaps use a separate PR to introduce a common helper script. I assume `_get_dms_environment_variables` is not necessary, there is no mention of any additional ENV we're depending on from it.

If you rather source `/etc/dms-settings`, have a comment that explicitly states we're only doing so for `POSTMASTER_ADDRESS`, assuming that there would be no conflict with any other vars used in this script (I haven't checked).

target/scripts/helpers/utils.sh
```diff
@@ -11,3 +11,22 @@ function _is_comment
...
   grep -q -E "^\s*$|^\s*#" <<< "${1}"
 }
+
+# This function parses `/root/.bashrc` and "extracts" the environment
+# variables by making them available to the current shell. After the
+# call, you can work with the environment variables previously defined
+# in `/root/.bashrc` as if you defined them yourself. Note that the
+# file `/root/.bashrc` itself is not altered when you alter one of the
+# parsed environment variables.
+function _get_dms_environment_variables
```
polarathene, line 21: As per previous review comment, how is this different/better to `source /etc/dms-settings`?

target/scripts/helpers/error.sh
```diff
@@ -62,13 +63,13 @@ function dms_panic__no_file { dms_panic 'no-file' "${1}" "${2}"; }
 function dms_panic__misconfigured { dms_panic 'misconfigured' "${1}" "${2}"; }
 function dms_panic__invalid_value { dms_panic 'invalid-value' "${1}" "${2}"; }
 
-# Call this method when you want to panic (emit a 'FATAL' log level error, and exit uncleanly).
+# Call this method when you want to panic (i.e. emit an 'ERROR' log and exit DMS with a non-zero exit code).
```
polarathene, line 66: I'm not sure how the change at the end is an improvement.

It's not any `non-zero` exit code, it's specifically `1`, no variability there for other values. We're performing `kill 1`, which to my knowledge is not a graceful shutdown, it's an unclean shutdown/exit is it not? If you agree, I think "exit uncleanly" communicates that better, the reader of the comment can see that it uses `exit 1`, is that adjustment to mention "a non-zero exit code" providing much value/context?

polarathene (changes requested):
While I do appreciate nice improvements like the [`daemons-stack.sh`](https://github.com/docker-mailserver/docker-mailserver/pull/2486/commits/7f9108f468f5982bfd4424014bca4af4780246bc) addition. It like others probably should have been delayed as follow-up PRs.

Similar to how some of the broader sweeping changes of this PR could have been introduced in phases. You're not only adjusting log levels, but changing the existing ones, and making adjustments to content among other changes related to the logs.

It's easier for reviewers in this case to go into a PR knowing that you're only introducing a new and improved logger method, with existing log levels being remapped in the method, then a followup PR that just focuses on changes to drop that mapping, and adjust the log levels throughout the files, especially when you're adjusting based on context (_for example when `inf` became `info`, `debug`, `trace` or just a plain echo_).

There's more than just that going on with this PR though. I rather you don't attempt to split the PR again due to residue issues we had with the split to this PR.

I don't particularly want to wade through the diff of this size again to catch those hiccups. So unless another reviewer explicitly requests it to assist them reviewing, please don't take that initiative :sweat_smile:

---

One of the main concerns with changes introduced since is your new utility for importing ENV, as the relevant review comments detail, this seems to be from a lack of awareness of `/etc/dms-settings` or I'm mistaken and there's good justification for it.

I'm also wary of the alternative approach for sourcing `POSTMASTER_ADDRESS`, mostly due to the dependency on `sleep`. Already the log output size has affected tests, and varying layers of `sleep` is another common issue we run into, I prefer to minimize the need for it where possible.

polarathene:
> I spent more than an hour trying to debug this. To me, it seems as if I found a false positive (in the past, we were viewing too much log ???).

If we were viewing too much log, then the test should still fail, it explicitly looks for the FQDN arg the function is called with:

https://github.com/docker-mailserver/docker-mailserver/blob/321ae744fa873ca300af83983b4f916ac8dc9c35/test/mail_ssl_letsencrypt.bats#L245

And after that test, I also added another to count duplicate log output for expected number of restarts by this point. Perhaps not exactly foolproof but should be reasonably reliable:

https://github.com/docker-mailserver/docker-mailserver/blob/321ae744fa873ca300af83983b4f916ac8dc9c35/test/mail_ssl_letsencrypt.bats#L253-L260

---

> and there is a third ` '/etc/letsencrypt/acme.json' has changed, extracting certs..` somehow missing. It could be that there are only 2, which would indicate that actually all is well but we're not looking at enough log?? Are there overall only 2 changes with this test? Have I misunderstood it the whole time?

- `_acme_ecdsa()`: Only cares that certs were extracted at startup.
- `_acme_rsa()`: Certs have changed, should be detected and extracted to `mail.example.test` cert dir.
- `_acme_wildcard()`: Same as before, but this time the first check for wildcard is successful and has priority, this is not a `*.example.test` dir, but `example.test` dir (_which may seem a bit odd but was to avoid `*.` as part of the directory name_).

Priority is due to the `SSL_DOMAIN=*.example.test` ENV specified early in the test as additional container args/ENV beyond the defaults:

https://github.com/docker-mailserver/docker-mailserver/blob/321ae744fa873ca300af83983b4f916ac8dc9c35/test/mail_ssl_letsencrypt.bats#L110-L118

Priority logic in change detection for `acme.json`:

https://github.com/docker-mailserver/docker-mailserver/blob/321ae744fa873ca300af83983b4f916ac8dc9c35/target/scripts/check-for-changes.sh#L90-L114

That should be calling the extract certs helper:

https://github.com/docker-mailserver/docker-mailserver/blob/321ae744fa873ca300af83983b4f916ac8dc9c35/target/scripts/helpers/ssl.sh#L411-L442

which should log a success or failure.. but there is two `|| exit 1` lines at the end there which would not log and go undetected.. I did not introduce that logic for those two lines, it was pre-existing before me. I recall briefly discussing it with casper in a review that it probably should get addressed at some point I think..?

The directory is created in advance however, and we return early with warning logged if `KEY` or `CERT` is empty. I'm not sure why that would fail to trigger `|| exit`?

I assume there is nothing wrong with calling the python script since that is working at an earlier stage in the tests. There was a [recent change to use an explicit `.py` extension](https://github.com/docker-mailserver/docker-mailserver/commit/1ab8ea96b234fbd1de6df9da69f80fb49cfe1d22), otherwise [`acme_extract.py`](https://github.com/docker-mailserver/docker-mailserver/blob/master/target/bin/acme_extract.py) works like so:

1. Given a filepath to `acme.json` (_[rsa example](https://github.com/docker-mailserver/docker-mailserver/blob/master/test/test-files/ssl/example.test/with_ca/ecdsa/rsa.acme.json) and [ecdsa wildcard example](https://github.com/docker-mailserver/docker-mailserver/blob/master/test/test-files/ssl/example.test/with_ca/ecdsa/wildcard/ecdsa.acme.json)_) parse it and iterate through the top-level keys looking for the `Certificates` key:
    
    https://github.com/docker-mailserver/docker-mailserver/blob/321ae744fa873ca300af83983b4f916ac8dc9c35/target/bin/acme_extract.py#L22-L27

2. The `Certificates` field contains an array of objects, each a certificate. Check each one for a `domain` field, it has the subfields `main` and `sans`, these days `sans` should contain the FQDN, we check that it has the FQDN we requested via args:
    
    https://github.com/docker-mailserver/docker-mailserver/blob/321ae744fa873ca300af83983b4f916ac8dc9c35/target/bin/acme_extract.py#L28-L30

   https://github.com/docker-mailserver/docker-mailserver/blob/321ae744fa873ca300af83983b4f916ac8dc9c35/target/bin/acme_extract.py#L15-L18

3. In the event of a failure, we return `None` which will print as an empty string (_which will be caught/logged by the ssl helper function `_extract_certs_from_acme`_):
    
    https://github.com/docker-mailserver/docker-mailserver/blob/321ae744fa873ca300af83983b4f916ac8dc9c35/target/bin/acme_extract.py#L31-L39

---

So that whole flow seems sound AFAIK. The issue with `_acme_wildcard()` in your current PR error logs detects a change but doesn't log anything about attempting to extract or that `acme.json` changed.. Referencing `check-for-changes.sh` master:

https://github.com/docker-mailserver/docker-mailserver/blob/321ae744fa873ca300af83983b4f916ac8dc9c35/target/scripts/check-for-changes.sh#L63-L65

https://github.com/docker-mailserver/docker-mailserver/blob/321ae744fa873ca300af83983b4f916ac8dc9c35/target/scripts/check-for-changes.sh#L88-L90

As we don't reach that 2nd step, what may have occurred is in `_acme_rsa()` this logic ran which would be intentional, but somehow prevented change detection of `acme.json` on the following `_acme_wildcard()` test:

https://github.com/docker-mailserver/docker-mailserver/blob/321ae744fa873ca300af83983b4f916ac8dc9c35/target/scripts/check-for-changes.sh#L100-L111

Actually, it looks like the first `|` is being escaped here, but not the 2nd, I'm not sure if that's valid or not, I've probably mentioned it in the relevant PR or commit in that PR (maybe PR merge commit message):

https://github.com/docker-mailserver/docker-mailserver/blob/321ae744fa873ca300af83983b4f916ac8dc9c35/target/scripts/check-for-changes.sh#L108

That doesn't explain why `acme.json` was not part of the detected change however. Could be a timing issue with insufficient sleep? When are you retrieving the logs you shared? If that's not in sync with when the test attempts to match, you could be right that the PR tests are reading and matching old logs, change-detection for the `acme.json` file may occur after that.

https://github.com/docker-mailserver/docker-mailserver/blob/321ae744fa873ca300af83983b4f916ac8dc9c35/test/mail_ssl_letsencrypt.bats#L237-L246

---

If file locking logic has anything to do with it, I will reference [these findings](https://github.com/docker-mailserver/docker-mailserver/pull/2157#pullrequestreview-817286005) I made late last year. I'd like to highlight the collapsed details block there "Switch back to `flock`", which points out:

> The **intent of locking files** in the first place - `flock` locked the configs being interacted with, while the replacement applied locks to individual scripts instead, thus locks only blocked a specific script from being concurrently called, but not concurrent access to the actual config files being modified which was the original intent with `flock` being introduced.

It's possible that is part of the problem, along with sleeps and the changes to log output size affecting timing? I would look into the checksum file changes, as that should reveal more about what change is detected for `_acme_wildcard()` if it's not the `acme.json` update..

---

I would offer to investigate this myself, but it would presently be inconvenient :(

Feel free to workaround or disable the test, then open an issue about it and assign me to it and I'll try allocate some time later this month for it if needed.

test/mail_ssl_letsencrypt.bats
```diff
@@ -104,12 +104,12 @@ function teardown() {
...
 
-    # TODO: Provision wildcard certs via Traefik to inspect if `example.test` non-wildcard is also added to the cert.
-    # `DMS_DEBUG=1` required for catching logged `inf` output.
+    # TODO: Provision wildcard certs via Traefik to inspect
+    # if `example.test` non-wildcard is also added to the cert.
     # shellcheck disable=SC2034
     local TEST_DOCKER_ARGS=(
       --volume "${TEST_TMP_CONFIG}/letsencrypt/acme.json:/etc/letsencrypt/acme.json:ro"
-      --env DMS_DEBUG=1
+      --env LOG_LEVEL='trace' \
```
polarathene, line 112:
```suggestion
      --env LOG_LEVEL='trace'
```

polarathene (changes requested):

target/scripts/check-for-changes.sh
```diff
@@ -6,47 +6,54 @@
 # shellcheck source=./helpers/index.sh
 source /usr/local/bin/helpers/index.sh
 
-function _log_date
+REGEX_NEVER_MATCH="(?\!)"
```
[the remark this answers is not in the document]
georglauterbach: I will put it back to where it was originally located :)

target/scripts/check-for-changes.sh
```diff
@@ -6,47 +6,54 @@
...
 
-LOG_DATE=$(_log_date)
-_notify 'task' "${LOG_DATE} Start check-for-changes script."
+function _changedetector_exit_error
+{
+  _changedetector_notify "${@}"
+  _changedetector_notify 'Aborting'
+
+  exit 1
+}
```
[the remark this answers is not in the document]
georglauterbach: You're right, there is the `error` log level missing. I will correct this and change `"${@}"` with `"${1}"`.

target/scripts/check-for-changes.sh
georglauterbach: With

https://github.com/docker-mailserver/docker-mailserver/blob/321ae744fa873ca300af83983b4f916ac8dc9c35/target/scripts/start-mailserver.sh#L19

a new variable `PM_ADDRESS` is created, but at the same time, the variable `POSTMASTER_ADDRESS` is assigned its default value before this value is again assigned to `PM_ADDRESS`. This is useful in `start-mailserver.sh`, but no here, because the call to

https://github.com/docker-mailserver/docker-mailserver/blob/321ae744fa873ca300af83983b4f916ac8dc9c35/target/scripts/check-for-changes.sh#L126

expects `POSTMASTER_ADDRESS` to be set correctly (otherwise `/etc/aliases` has errors in it).

But you're correct, using `/etc/dms-settings` would have been (actually is) easier, I just forgot. I will update the parts of the code.

target/scripts/helpers/error.sh
georglauterbach: I will revert it back.

target/scripts/helpers/utils.sh
georglauterbach: See my comment (I will revert this, as it's not needed).

test/mail_ssl_letsencrypt.bats
georglauterbach: 👍🏼

target/scripts/check-for-changes.sh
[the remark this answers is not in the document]
georglauterbach: @polarathene anything wrong with my reply? Is this not what you wanted to happen; did I misunderstand anything? (because of the 👎🏼)

georglauterbach:
# Houston, we have a problem

## The Problem

So, after long and painful hours of debugging, I found some problems that are currently present with `check-for-changes.sh`. After I was unable to figure out why my log PR #2486 fails, I had enough and added the following lines to the script to look at the changes myself:

``` BASH
_changedetector_notify 'warn' 'DEBUG START'
echo -e "\nchanged: ${CHANGED}\n"
echo -e "\nnew: $(<${CHKSUM_FILE}.new)\n"
echo -e "\nold: $(<${CHKSUM_FILE})\n"
_changedetector_notify 'warn' 'DEBUG END'
```

For context: the current logs only show two changes being being detected in the `mail_ssl_letsencrypt.bats` test, while with #2486, three changes are detected... To my surprise, the last change looked like this:

``` TXT
changed:


new: d6825c549a553df95e3830b6a3ec3f161f320b2a600b4f1fd63540920ad33fd6b587dbdf8359ccb35dd1a17c9fd14c17f113954f73da8f19bcf83f1649fd757a  /tmp/docker-mailserver/postfix-accounts.cf
37d1a2f5cb4f99c81463b05b366f27f3877d10e83685fec1ec1f38aabf60e2d28bc32a7d4b810e66739b37912f9dd9d1424145b93c7e5969120d345c05c8f74c  /tmp/docker-mailserver/postfix-virtual.cf
cf83e1357eefb8bdf1542850d66d8007d620e4050b5715dc83f4a921d36ce9ce47d0d13c5d85f2b0ff8318d2877eec2f63b931bd47417a81a538327af927da3e  /tmp/docker-mailserver/postfix-aliases.cf
cf83e1357eefb8bdf1542850d66d8007d620e4050b5715dc83f4a921d36ce9ce47d0d13c5d85f2b0ff8318d2877eec2f63b931bd47417a81a538327af927da3e  /tmp/docker-mailserver/dovecot-quotas.cf
a0eafaf897cb7c793244fe71de32120d7b22897ffe20e5f9f295251e992104c48148e3103187a61c7d92ad773d1e00889d4790038ca4326b414439dfbd90045e  /etc/letsencrypt/acme.json
5f79d61f94f365052e7c079d1ea5d3ac8810e6e5c028d8b22b12d80f3ceeffc0c8182a4bace90cea265d357fc73bb3518214cb54247a7945152d6dbd94d23a31  /etc/letsencrypt/live/mail.example.test/fullchain.pem
4fd75dc009e4499cd72e12ace6f348bfd904ac4e816aa2a13200ec2749a0e9b86f362407bc35c945f7cd667b3da5449de771ed96d0ae6f7cb24c8f82fcd3d3b6  /etc/letsencrypt/live/mail.example.test/key.pem


old: d6825c549a553df95e3830b6a3ec3f161f320b2a600b4f1fd63540920ad33fd6b587dbdf8359ccb35dd1a17c9fd14c17f113954f73da8f19bcf83f1649fd757a  /tmp/docker-mailserver/postfix-accounts.cf
37d1a2f5cb4f99c81463b05b366f27f3877d10e83685fec1ec1f38aabf60e2d28bc32a7d4b810e66739b37912f9dd9d1424145b93c7e5969120d345c05c8f74c  /tmp/docker-mailserver/postfix-virtual.cf
cf83e1357eefb8bdf1542850d66d8007d620e4050b5715dc83f4a921d36ce9ce47d0d13c5d85f2b0ff8318d2877eec2f63b931bd47417a81a538327af927da3e  /tmp/docker-mailserver/postfix-aliases.cf
cf83e1357eefb8bdf1542850d66d8007d620e4050b5715dc83f4a921d36ce9ce47d0d13c5d85f2b0ff8318d2877eec2f63b931bd47417a81a538327af927da3e  /tmp/docker-mailserver/dovecot-quotas.cf
a0eafaf897cb7c793244fe71de32120d7b22897ffe20e5f9f295251e992104c48148e3103187a61c7d92ad773d1e00889d4790038ca4326b414439dfbd90045e  /etc/letsencrypt/acme.json
5f79d61f94f365052e7c079d1ea5d3ac8810e6e5c028d8b22b12d80f3ceeffc0c8182a4bace90cea265d357fc73bb3518214cb54247a7945152d6dbd94d23a31  /etc/letsencrypt/live/mail.example.test/fullchain.pem
4fd75dc009e4499cd72e12ace6f348bfd904ac4e816aa2a13200ec2749a0e9b86f362407bc35c945f7cd667b3da5449de771ed96d0ae6f7cb24c8f82fcd3d3b6  /etc/letsencrypt/live/mail.example.test/key.pem
f8911bc63794d532e2af7e24f71a92dea5fbef8a3f4da5249a13285fd715c958cdd265eea82721dbc5f9921613916d2b2cf21bf2012a1d6abae2a5a76c1e6faa  /etc/letsencrypt/live/example.test/fullchain.pem
760241712a2d96a6cbfe166c9fa063458c57ba869a8caf5bed1add6547f5445e1e890320f262ff04c60ecd757da15cbfc006547c5f0717af994a08edff20009b  /etc/letsencrypt/live/example.test/key.pem
```

After doing a manual `cmp`, the result was:

``` CONSOLE
$ cmp old new
cmp: EOF on old after byte 1207, line 7
$ echo $?
1
$ diff old new
7a8,9
> f8911bc63794d532e2af7e24f71a92dea5fbef8a3f4da5249a13285fd715c958cdd265eea82721dbc5f9921613916d2b2cf21bf2012a1d6abae2a5a76c1e6faa  /etc/letsencrypt/live/example.test/fullchain.pem
> 760241712a2d96a6cbfe166c9fa063458c57ba869a8caf5bed1add6547f5445e1e890320f262ff04c60ecd757da15cbfc006547c5f0717af994a08edff20009b  /etc/letsencrypt/live/example.test/key.pem
```

So, there are two **additional** lines in the diff now.

``` TXT
f8911bc63794d532e2af7e24f71a92dea5fbef8a3f4da5249a13285fd715c958cdd265eea82721dbc5f9921613916d2b2cf21bf2012a1d6abae2a5a76c1e6faa  /etc/letsencrypt/live/example.test/fullchain.pem
760241712a2d96a6cbfe166c9fa063458c57ba869a8caf5bed1add6547f5445e1e890320f262ff04c60ecd757da15cbfc006547c5f0717af994a08edff20009b  /etc/letsencrypt/live/example.test/key.pem
```

Adding an `ls -lhA /etc/letsencrypt/live/**` to the script yields

``` TXT
/etc/letsencrypt/live/example.test:
total 8.0K
-rw-r--r-- 1 root root 1.2K Mar 18 09:28 fullchain.pem
-rw-r--r-- 1 root root 1.7K Mar 18 09:28 key.pem

/etc/letsencrypt/live/mail.example.test:
total 8.0K
-rw-r--r-- 1 root root 1.3K Mar 18 09:27 fullchain.pem
-rw-r--r-- 1 root root 1.7K Mar 18 09:27 key.pem
```

Summing up: the additional files are picked by the changedetector-run and `cmp` (which actually is correct behaviour, if you ask me). I don't know why this would not be the case on the `master` branch... The wildcard certificate does not use the `mail.` prefix for `example.test`, therefore the new files in `example.test` without the `mail.` prefix exist. The old files not cleaned up? This causes an additional run for the changedetector.

## The Solution

I adjusted the tests to query an appropriate amount of log. This "solves" the issue, but in the future, a better solution could / should be found (the log should be queried in a more intelligent way). This is however not related to this PR.

georglauterbach:
@polarathene now I need your help again again:

https://github.com/docker-mailserver/docker-mailserver/blob/321ae744fa873ca300af83983b4f916ac8dc9c35/test/mail_ssl_letsencrypt.bats#L154-L173

Line 171 says "WARNING: This should fail...but requires resolving the above TODO. ". This test now "fails correctly", I guess, because #2404 (the above mentioned TODO) is merged? You wrote the test, therefore I'm not too sure whether the now failing test is actually supposed to fail?

If this is the case, all is well and we just need to adjust the test, but why was this not showing earlier? :D

polarathene:
> I don't know why this would not be the case on the `master` branch... The wildcard certificate does not use the `mail.` prefix for `example.test`, therefore the new files in `example.test` without the `mail.` prefix exist. The old files not cleaned up? This causes an additional run for the changedetector.

You don't think this is related to more earlier investigation notes above? This is in particular:

> As we don't reach that 2nd step, what may have occurred is in `_acme_rsa()` this logic ran which would be intentional, but somehow prevented change detection of `acme.json` on the following `_acme_wildcard()` test:
> 
> https://github.com/docker-mailserver/docker-mailserver/blob/321ae744fa873ca300af83983b4f916ac8dc9c35/target/scripts/check-for-changes.sh#L100-L111
> 
> Actually, it looks like the first `|` is being escaped here, but not the 2nd, I'm not sure if that's valid or not, I've probably mentioned it in the relevant PR or commit in that PR (maybe PR merge commit message):
> 
> https://github.com/docker-mailserver/docker-mailserver/blob/321ae744fa873ca300af83983b4f916ac8dc9c35/target/scripts/check-for-changes.sh#L108

I have not verified the `sed` concern, but if it's not correctly removing the lines to prevent an unnecessary change detection, then that explains why you're getting a change detection event regarding new cert files, instead of for `acme.json` which should be the relevant change to react to in the test - not changes that happened as a result of the cert extraction in `_acme_rsa()` prior, or after `example.test` wildcard (_if that change was detected and handled, but this logic resulted in an additional change after_).

I'd also have to give it some thought if that logic is testing with `*.example.test` or `example.test`, it looks like `CERT_DOMAIN` is originally `*.example.test` and then if the strip method works correctly, it should be adjusted to `example.test` for the remaining logic.

You say master has 2 change detections logged, and then share the third additional change log that this PR caught due to log verbosity changes.. does this mean that the first two are the correct `acme.json` changes as expected? And the third is an accidental event with the `example.test` wildcard cert files triggering detection? The failure in the letsencrypt test is now present because that 2nd change event for `acme.json` success with the wildcard cert got the log bumped out of the retrieved lines, with only the third change event logs that shouldn't have occurred?

polarathene:
> Line 171 says "WARNING: This should fail...but requires resolving the above TODO. ". **Or is he warning for the line below?**

This should be fine to enable now. It's referencing line 163 which would also be enabled along with it, and the TODO comment above it. They needed the change detection event to run the ssl startup script again to update the filepaths to the correct letsencrypt directory (changed from `mail.example.test` to `example.test`).

It's an edge case, most users won't be changing to/from a wildcard cert on a running instance I imagine, this test will ensure we catch it though if someone refactors `check-for-changes.sh` thinking it wasn't necessary.

polarathene (changes requested):
## Stay focused, refrain from extras that can wait

I want to see this PR merged, please stop adding additions not related to the feedback. This is already deviating again from the core focus that resulted in it being extracted to a separate PR.

If you feel comments could be refactored/added, unless specifically about the new log functionality being introduced, hold off on that. Create a branch and pile them up there for you to follow up with.

This PR is a good improvement, but becoming a bit frustrating to review when changes that could be deferred to another PR are thrown in (_and then reverted after review when I state not to unless requested by other reviewers, sometimes with residue errors_).

---

**Under no circumstances**, take the initiative to clean up this PR with a force-pushed interactive-rebase (_it's unlikely, but I want to communicate that clearly_).

If you feel a need to do such, wait until ready to merge, so that all feedback and discussion is resolved first, then diffing the prior state to the reworked commits won't be as much effort. 

Personally, since we squash commits on merge, and this PR history is already not pleasant to wade through for traceability to a future maintainer, I rather we not complicate it further and discourage altering PR history this way.

georglauterbach:
> ## Stay focused, refrain from extras that can wait
> 
> I want to see this PR merged, please stop adding additions not related to the feedback. This is already deviating again from the core focus that resulted in it being extracted to a separate PR.
> 
> If you feel comments could be refactored/added, unless specifically about the new log functionality being introduced, hold off on that. Create a branch and pile them up there for you to follow up with.
> 
> This PR is a good improvement, but becoming a bit frustrating to review when changes that could be deferred to another PR are thrown in (_and then reverted after review when I state not to unless requested by other reviewers, sometimes with residue errors_).
> 
> **Under no circumstances**, take the initiative to clean up this PR with a force-pushed interactive-rebase (_it's unlikely, but I want to communicate that clearly_).
> 
> If you feel a need to do such, wait until ready to merge, so that all feedback and discussion is resolved first, then diffing the prior state to the reworked commits won't be as much effort.
> 
> Personally, since we squash commits on merge, and this PR history is already not pleasant to wade through for traceability to a future maintainer, I rather we not complicate it further and discourage altering PR history this way.

I'm sorry, I tried to follow your advice and only added what I deemed necessary for tests run. I don't know how to proceed with this PR without annoying anyone even further - I have therefore come to the conclusion that I will close this and provide smaller updates from time to time, which should be easier to review and easier to test.

**I am very sorry to have wasted your time**, but I think we nevertheless got some insights - and if it's only about ditching `supervisor`.

test/mail_ssl_letsencrypt.bats
```diff
@@ -299,9 +310,15 @@ function _should_be_equal_in_content() {
 }
 
 function _get_service_logs() {
-  local SERVICE=${1:-'mailserver'}
-
-  local CMD_LOGS=(docker exec "${TEST_NAME}" "supervisorctl tail ${SERVICE}")
+  local SERVICE=${1?}
```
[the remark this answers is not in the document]
georglauterbach: `${1?}` is very similar to `${:?}`, only that with the former, there is no `null`ity check. With the latter, when the variable is unset or `null`, Bash will stop and show an error.

Implicit default values can be nice, but I had a very hard time finding out that a function call to another function call that calls this function uses the implicit value - to the point where I thought I had found a bug. I'd refrain from such default values unless it's obvious.

polarathene:
> I don't know how to proceed with this PR without annoying anyone even further - I have therefore come to the conclusion that I will close this and provide smaller updates from time to time, which should be easier to review and easier to test.

It's only really been me providing reviews each morning I woke up. It was getting a bit overwhelming with all the changes going on, I was checking the individual commits for diffs, reading the commit messages and referencing the `master` version and full diff for certain files like `check-for-changes.sh`.

I could have communicated a bit better, or delayed review until I got a morning coffee :sweat_smile: I know sometimes my communication can come off a bit blunt and harsh, while I don't often comment on all the good changes the PR brought and my appreciation for how responsive you have been to my feedback, there was plenty of that unsaid :)

> I am very sorry to have wasted your time, but I think we nevertheless got some insights - and if it's only about ditching supervisor.

I don't think you wasted my time. I chose to review regularly and invest the I did to provide as thorough review as I could because I value the changes you've put together here and I think you've invested a fair amount more of your time putting all this together.

I hope you do see it through, in whatever form that is, I'll continue to be happy to review it. Reviews work both ways, I can only be a better reviewer with your feedback when I overstep or get too pushy/fussy.

If you don't have the time to keep working on this, I can understand that too. I'm happy to piecemeal it through several PRs like I've done in the past with another maintainers large PR. It looks like a lot of thought went into it, it'd be a shame to lose it.

casperklein:
Just adding my 2 cents^^. PRs should be as simple as possible (just doing/introducing/fixing at best only one thing). In question, some should provide multiple PRs instead of one. This makes reviewing so much easier and faster.

There were so much discussion/changes/splitting in this PR, I was not able to keep track at some point. Then my "strategy" changed to wait until this has been calm down and then try to review it 😆

An approach for bigger PRs could also be, to open an issue and discuss the planed changes beforehand (if necessary/worth). I feel sometimes guilty, if someone took a lot of time to prepare something, for starting a general discussion about it. But just because it was a lot of work, doesn't automatically mean it's a good addition or that it can't be done better/simpler.

PR #2428, opened 2022-02-21
Adjust envrionment variables - more sensible defaults

georglauterbach:
# Description

The new setup will now set env variables in one place and in one place
only. The old setup used two separate places wich is not DRY and
confusing.

Some default values changed:

1. `PFLOGSUMM_TRIGGER`: logrotate => none
2. `REPORT_SENDER`: mailserver-report@HOSTNAME => mailserver-report@DOMAIN
3. `REPORT_RECIPIENT`: "0" => `POSTMASTER_ADDRESS`
4. `LOGROATE_INTERVAL`: `daily` => `weekly`

One env variable was renamed: `REPORT_INTERVAL` => `LOGROTATE_INTERVAL`

I believe these defaults to be more sensible, especially the REPORT_RECIPIENT
address. The PFLOGSUMM_TRIGGER value was changed to `none` because otherwise
people would start getting daily Postfix log summary reports automatically.
Now, this is opt-in, and reports are sent only when enabled properly.

Some of the variables changed were marked as deprecated. I removed the note,
as the variables now bear some (sane) defaults again for other variables
(i.e.) `REPORT_RECIPIENT` is now default for other recipient addresses.

Although some defaults changed, this is not a breaking change as the
mail server starts exactly as before. Even for REPORT_RECIPIENT="0"
explictly set, other reports are opt-in as well with their defaults,
and as the recipient default changed, this is fine as well.

I would leave it up to vote whether this PR qualifies for `10.5.0` or for `11.0.0`. I'm also open to other defaults, although I think they're okay as they are. Please watch out for inconsistencies in the documentation :D

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Improvement (non-breaking change that does improve existing functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [x] If necessary I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #2487, opened 2022-03-15
Use FQDN as REPORT_SENDER default value.

casperklein:
# Description

With the merge of #2428, the default sender address was changed.

Let's take our example config:
https://github.com/docker-mailserver/docker-mailserver/blob/b730942b96aec3ab52699b5d927577f42eef0a45/docker-compose.yml#L7-L8

Before: Reports sender address was mailserver-report@mail.example.com
After: Reports sender address is mailserver-report@example.com

This is wrong IMO.

The subject line of the reports are `Logwatch for mail.example.com (Linux)` and `Postfix summary for mail.example.com`. They correctly state the FQDN for which the reports are. The sender address should therefore also be `mail.example.com`.
`mail.example.com` is not necessary the same host as `example.com`, which is another reason for this change.

This PR uses the FQDN of the mailserver as default sender address.

Fixes #

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/start-mailserver.sh
+++ b/target/scripts/start-mailserver.sh
@@ -18,7 +18,7 @@ _obtain_hostname_and_domainname
 # These variables must be defined first; They are used as default values for other variables.
 VARS[POSTMASTER_ADDRESS]="${POSTMASTER_ADDRESS:=postmaster@${DOMAINNAME}}"
 VARS[REPORT_RECIPIENT]="${REPORT_RECIPIENT:=${POSTMASTER_ADDRESS}}"
-VARS[REPORT_SENDER]="${REPORT_SENDER:=mailserver-report@${DOMAINNAME}}"
+VARS[REPORT_SENDER]="${REPORT_SENDER:=mailserver-report@$(hostname -f)}"
 
 VARS[AMAVIS_LOGLEVEL]="${AMAVIS_LOGLEVEL:=0}"
 VARS[CLAMAV_MESSAGE_SIZE_LIMIT]="${CLAMAV_MESSAGE_SIZE_LIMIT:=25M}" # 25 MB
--- a/target/scripts/start-mailserver.sh
+++ b/target/scripts/start-mailserver.sh
@@ -18,7 +18,7 @@ _obtain_hostname_and_domainname
 # These variables must be defined first; They are used as default values for other variables.
 VARS[POSTMASTER_ADDRESS]="${POSTMASTER_ADDRESS:=postmaster@${DOMAINNAME}}"
 VARS[REPORT_RECIPIENT]="${REPORT_RECIPIENT:=${POSTMASTER_ADDRESS}}"
-VARS[REPORT_SENDER]="${REPORT_SENDER:=mailserver-report@$(hostname -f)}"
+VARS[REPORT_SENDER]="${REPORT_SENDER:=mailserver-report@${HOSTNAME}}"
 
 VARS[AMAVIS_LOGLEVEL]="${AMAVIS_LOGLEVEL:=0}"
 VARS[CLAMAV_MESSAGE_SIZE_LIMIT]="${CLAMAV_MESSAGE_SIZE_LIMIT:=25M}" # 25 MB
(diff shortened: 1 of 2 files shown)


target/scripts/start-mailserver.sh
```diff
@@ -18,7 +18,7 @@ _obtain_hostname_and_domainname
 # These variables must be defined first; They are used as default values for other variables.
 VARS[POSTMASTER_ADDRESS]="${POSTMASTER_ADDRESS:=postmaster@${DOMAINNAME}}"
 VARS[REPORT_RECIPIENT]="${REPORT_RECIPIENT:=${POSTMASTER_ADDRESS}}"
-VARS[REPORT_SENDER]="${REPORT_SENDER:=mailserver-report@${DOMAINNAME}}"
+VARS[REPORT_SENDER]="${REPORT_SENDER:=mailserver-report@$(hostname -f)}"
```
georglauterbach, line 21: I remember this was `HOSTNAME` before. I think there also is a test you'll need to adjust.

casperklein: You are right, one test failed. I must have done something wrong yesterday, as my local tests passed successful 😕

georglauterbach: I see, but here in `start-mailserver.sh`, we could also use the same env variable, right?

georglauterbach:
GitHub has some problems ATM, therefore tests are failing.

target/scripts/start-mailserver.sh
casperklein: I re-read your comment 😆 `$HOSTNAME` should also be fine.

PS; I still find it very confusing, that we modify system variables. `$HOSTNAME` is set by docker.
If I `bash` into DMS and `echo $HOSTNAME`, just `mail` is returned,  while in the `start-mailserver.sh` context `$HOSTNAME` is a FQDN:

```bash
root@mail:/# echo $HOSTNAME
mail
root@mail:/# source ./usr/local/bin/helpers/dns.sh
root@mail:/# _obtain_hostname_and_domainname
root@mail:/# echo $HOSTNAME
mail.chill.to
```

Maybe in a future PR we should change that. Instead of modifying $HOSTNAME, we should use $DMS_HOSTNAME or something else to make the difference clear and avoid confusion.

casperklein: I re-read your comment 😆 `$HOSTNAME` should also be fine.

PS; I find it very confusing, that we modify system variables. This is bad practice IMO. `$HOSTNAME` is set by docker.
If I `bash` into DMS and `echo $HOSTNAME`, just `mail` is returned,  while in the `start-mailserver.sh` context `$HOSTNAME` is a FQDN:

```bash
root@mail:/# echo $HOSTNAME
mail
root@mail:/# source ./usr/local/bin/helpers/dns.sh
root@mail:/# _obtain_hostname_and_domainname
root@mail:/# echo $HOSTNAME
mail.example.com
```

Maybe in a future PR we should change that. Instead of modifying $HOSTNAME, we should use $DMS_HOSTNAME or something else to make the difference clear and avoid confusion.

casperklein: > I re-read your comment 😆 $HOSTNAME should also be fine.

Just tested it - it's not. Because the containers `$HOSTNAME` (no FQDN) is different to `$HOSTNAME` (FQDN) used in `start-mailserver.sh`.

georglauterbach: I agree, this needs some reworking. I think @polarathene has this in his backlog somewhere :D

polarathene: > PS; I find it very confusing, that we modify system variables. This is bad practice IMO. $HOSTNAME is set by docker.

:+1: 

I've already expressed this concern in the past, and there was talk to change to `DMS_HOSTNAME` / `DMS_FQDN` IIRC, but required tests to be adapted for running through some different configs as I believe dropping `DOMAINNAME` was also being considered.

polarathene (approved):

georglauterbach (approved):

PR #2492, opened 2022-03-19
Improve test, get rid of sleep

casperklein:
# Description

This PR speeds up the test, by getting rid of `sleep 15`.

Before: The test waits 15 seconds and then checks for the string `Manual certificates have changed` in the changedetector log file.

After: The changedetector log file is constantly monitored until the string is found. If the string is not found within 15 seconds, the test fails.

The test time is reduced to [~6 seconds](https://github.com/docker-mailserver/docker-mailserver/runs/5608390029?check_suite_focus=true#step:8:103).

## Type of change

## Checklist:

### DIFF
--- a/test/mail_ssl_manual.bats
+++ b/test/mail_ssl_manual.bats
@@ -101,11 +101,9 @@ function teardown_file() {
 @test "checking ssl: manual cert changes are picked up by check-for-changes" {
     printf 'someThingsChangedHere' \
       >>"$(pwd)/test/test-files/ssl/${DOMAIN_SSL_MANUAL}/with_ca/ecdsa/key.ecdsa.pem"
-    sleep 15
 
-    run docker exec mail_manual_ssl /bin/bash -c "supervisorctl tail -3000 changedetector"
-    assert_output --partial 'Change detected'
-    assert_output --partial 'Manual certificates have changed'
+    run timeout 15 docker exec mail_manual_ssl bash -c "tail -F /var/log/supervisor/changedetector.log | sed '/Change detected/ q'"
+    assert_success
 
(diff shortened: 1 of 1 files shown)


test/mail_ssl_manual.bats
```diff
@@ -101,11 +101,9 @@ function teardown_file() {
 @test "checking ssl: manual cert changes are picked up by check-for-changes" {
     printf 'someThingsChangedHere' \
       >>"$(pwd)/test/test-files/ssl/${DOMAIN_SSL_MANUAL}/with_ca/ecdsa/key.ecdsa.pem"
-    sleep 15
 
-    run docker exec mail_manual_ssl /bin/bash -c "supervisorctl tail -3000 changedetector"
-    assert_output --partial 'Change detected'
-    assert_output --partial 'Manual certificates have changed'
+    run timeout 15 docker exec mail_manual_ssl bash -c "tail -F /var/log/supervisor/changedetector.log | sed '/Manual certificates have changed/ q'"
```
polarathene, line 105: Is this a way to get the full log? I think I had difficulty with tests I wrote that could properly retrieve that, so I relied on the supervisor tail command instead which is problematic..

---

Just to confirm I understand the command, it will tail the log file, piping the output to sed until the log is updated with the line sed is looking for, then sed quits with success, or the timeout results in failure?

polarathene (approved):

georglauterbach (approved):
LGTM 👍🏼

test/mail_ssl_manual.bats
casperklein: > Is this a way to get the full log?

In this case, yes. The log file does not exist when the test starts. That's why I used `-F`. It will start "tailing" as soon as the file exists.
If you go with the log files directly instead of `supervisorctl tail`, then you have the whole log available. I'll take a look at the other tests, that also uses `supervisorctl tail` and see, if those can also be improved. I wanted to discuss/merge this first, to see if that's a good way to go.

> Just to confirm ...

Correct.

PR #2498, opened 2022-03-21
scripts: refactored `check-for-changes.sh`

georglauterbach:
# Description

Slightly altered `check-for-changes.sh` and adjusted it to the new log. Removed `_notify` as it is now not used anywhere else. I will remove `DMS_DEBUG` with a separate PR.

## Type of change

## Checklist:


target/scripts/check-for-changes.sh
```diff
@@ -5,48 +5,40 @@
...
 fi
 
 # verify checksum file exists; must be prepared by start-mailserver.sh
 if [[ ! -f ${CHKSUM_FILE} ]]
 then
-  _notify 'err' "${LOG_DATE} ${CHKSUM_FILE} is missing! Start script failed? Exit!"
-  exit 0
+  _log 'error' "'/tmp/docker-mailserver/${CHKSUM_FILE}' is missing"
+  exit 1
 fi
```
polarathene, line 32: `_exit_with_error` is probably more suitable here? It'd retain the same behaviour but also include the "Aborting" message that you've dropped?

target/scripts/check-for-changes.sh
```diff
@@ -5,48 +5,40 @@
...
-_notify 'task' "${LOG_DATE} Start check-for-changes script."
-
-# ? --------------------------------------------- Checks
-
-cd /tmp/docker-mailserver || exit 1
+if ! cd /tmp/docker-mailserver &>/dev/null
+then
+  _log "Could not change into '/tmp/docker-mailserver/' directory"
+  exit 1
+fi
```
polarathene, line 19: Likewise, would `_exit_with_error` be a better choice here?

test/mail_ssl_letsencrypt.bats
```diff
@@ -241,7 +241,6 @@ function _should_extract_on_changes() {
 
   # Expected log lines from the changedetector service:
   run $(_get_service_logs 'changedetector')
-  assert_output --partial 'Change detected'
```
polarathene, line 244: Just curious about motivation for removal? Doesn't provide any value I assume?

polarathene (approved):

target/scripts/check-for-changes.sh
georglauterbach: > `_exit_with_error` is probably more suitable here? It'd retain the same behaviour but also include the "Aborting" message that you've dropped?

👍🏼

target/scripts/check-for-changes.sh
georglauterbach: 👍🏼

test/mail_ssl_letsencrypt.bats
georglauterbach: Commit message for 375d30e7e4bafa485787999a101e44cbc823cb07 says:

``` TXT
...
Moreover,

assert_output --partial 'Change detected'

was dropped because it appears first in the logs, and due to the now
more verbose log, we'd need to increase the lenght of the `tail`ed
output, but this an unreliable method. I therefore simply dropped this
message, and I think this is fine, because the other messages are still
fine for testing whether all is well with the script and its
functionality.
```

target/scripts/check-for-changes.sh
georglauterbach: Done in 5f71f074fca7f5b1eb44978206c9f589065f7799

target/scripts/check-for-changes.sh
```diff
@@ -87,7 +79,7 @@ do
     # extracted for `docker-mailserver` services to adjust to.
     elif [[ ${CHANGED} =~ /etc/letsencrypt/acme.json ]]
     then
-      _notify 'inf' "'/etc/letsencrypt/acme.json' has changed, extracting certs.."
+      _log 'debug' "'/etc/letsencrypt/acme.json' has changed - extracting certificates"
```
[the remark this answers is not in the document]
georglauterbach: Thanks for helping me here! :)

Done in 5f71f074fca7f5b1eb44978206c9f589065f7799

georglauterbach:
Feel free to have a look. I'm waiting on #2499 to get merged because this PR holds the `_log_with_date` function, which I want to use in `check-for-changes.sh` as well.

georglauterbach:
@polarathene I added a longer commit message for [`4e40ce1`](https://github.com/docker-mailserver/docker-mailserver/pull/2498/commits/4e40ce10b26a58b42259fa23234f07971d37d62a) which adjusts the LE test if you're curious why I did what I did.

polarathene:
From [commit message](https://github.com/docker-mailserver/docker-mailserver/pull/2498/commits/4e40ce10b26a58b42259fa23234f07971d37d62a):

> In the end, `_acme_wildcard` run, (FOR ANOTHER DOMAIN mind you). This results in another entry being created under `/etc/letsencrypt/live` which causes the changedetector to run twice, not once (which is correct).
> 
> Now, all is not well. Due to the (in total) third run, the log does not contain the `'/etc/letsencrypt/acme.json' has changed...` message anymore when `tail`ing normally (because the `acme.json` hasn't changed again).
> 
> 2. Added more `sleep` duration for two changedetector runs to complete without problems. This prevents timing issues in `_acme_wildcard`.
> 3. Obviously, the restart count needed to be adjusted as well

The third change event shouldn't be occurring? We identified this was triggered by the wildcard cert files, yet the rsa cert files did not trigger a change (as they shouldn't) despite different contents to the ecdsa cert files. The wildcard files should not be triggering a change event either... I prefer we resolve that bug tbh.

I detailed all of this with full flow and likely lines that contain the bug in the original PR discussion about this.

Getting restart counts from the full log is great, but adjusting the test to accommodate the bug is not when we should be able to fix it instead of ignore it.

georglauterbach:
@polarathene I found the regression, and [`31c91be`](https://github.com/docker-mailserver/docker-mailserver/pull/2498/commits/31c91be845dc096e5769610affbd26bc120e58b0) fixes the issue. I reverted the necessary test pieces back to their original state :)

polarathene:
> @polarathene I found the regression, and [`31c91be`](https://github.com/docker-mailserver/docker-mailserver/pull/2498/commits/31c91be845dc096e5769610affbd26bc120e58b0) fixes the issue.

Referencing the commit message:

> So, apparently
> 
> 1. `check-for-changes.sh` needs HOSTNAME and DOMAINNAME to to be set properly (this is now fixed again, I removed it earlier)
> 2. POSTMASTER_ADDRESS was not properly set to begin with, so this PR actually also fixes a bug :D

HOSTNAME and DOMAINNAME are needed for the `ssl.sh` helper right? Or other scripts? I wonder if there is a maintenance concern with the helpers we source if vars like those two internal ones can be expected in scripts but `check-for-changes.sh` doesn't keep in sync.. since I don't think we have much tests for such (_basically the whole test suite again after a change event which is a bit tricky_).

georglauterbach:
> > @polarathene I found the regression, and [`31c91be`](https://github.com/docker-mailserver/docker-mailserver/pull/2498/commits/31c91be845dc096e5769610affbd26bc120e58b0) fixes the issue.
> 
> Referencing the commit message:
> 
> > So, apparently
> > 
> > 1. `check-for-changes.sh` needs HOSTNAME and DOMAINNAME to to be set properly (this is now fixed again, I removed it earlier)
> > 2. POSTMASTER_ADDRESS was not properly set to begin with, so this PR actually also fixes a bug :D
> 
> HOSTNAME and DOMAINNAME are needed for the `ssl.sh` helper right? Or other scripts? I wonder if there is a maintenance concern with the helpers we source if vars like those two internal ones can be expected in scripts but `check-for-changes.sh` doesn't keep in sync.. since I don't think we have much tests for such (_basically the whole test suite again after a change event which is a bit tricky_).

I guess they are needed by `ssl.sh` yes, but other function also rely on other env variables. Because we're currently using HOSTNAME and not something like DMS_HOSTNAME, we're not adding this to `/etc/dms-settings` and we therefore cannot just rely on this file. But we already discussed that the whole DNS/HOSTNAME/DOMAINNAME/etc. thing will change in the future.

Keeping them in sync is a whole other task 😉 There is a maintenance concern, but there is currently no need to worry because we do not rely on this and we do not expect these variables to change during the lifetime of the container.

target/scripts/helpers/utils.sh
```diff
@@ -11,3 +11,8 @@ function _is_comment
 {
   grep -q -E "^\s*$|^\s*#" <<< "${1}"
 }
+
+function _log_date
+{
+  date +"%Y-%m-%d %H:%M:%S"
+}
```
casperklein, line 18: Outdated branch? This function should be in `logs.sh`

polarathene: https://github.com/docker-mailserver/docker-mailserver/pull/2498#issuecomment-1075040739

> I'm waiting on #2499 to get merged because this PR holds the `_log_with_date` function, which I want to use in `check-for-changes.sh` as well.

georglauterbach:
All important changes are included now. This PR is officially ready for review @casperklein @polarathene :)

polarathene (approved):

casperklein (approved):

PR #2499, opened 2022-03-21
scripts: refactoring & miscellaneous small changes

georglauterbach:
# Description

Please see the commit messages. This is a housekeeping PR which includes several small changes.

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/helpers/log.sh
+++ b/target/scripts/helpers/log.sh
@@ -91,6 +91,8 @@ function _log
   fi
 }
 
+function _log_with_date { _log "${1}" "$(date +"%Y-%m-%d %H:%M:%S")  ${2}" ; }
+
 # Still used by `check-for-changes.sh` for legacy / test purposes. Adjusting
 # `check-for-changes.sh` must be done with great care and requires some effort.
 # As a consequence, this function is kept to keep some of the original log for
--- a/target/scripts/helpers/log.sh
+++ b/target/scripts/helpers/log.sh
@@ -91,7 +91,7 @@ function _log
   fi
 }
 
-function _log_with_date { _log "${1}" "$(date +"%Y-%m-%d %H:%M:%S")  ${2}" ; }
+function _log_with_date { _log "${1}" "$(date '+%Y-%m-%d %H:%M:%S')  ${2}" ; }
 
 # Still used by `check-for-changes.sh` for legacy / test purposes. Adjusting
 # `check-for-changes.sh` must be done with great care and requires some effort.
--- a/target/scripts/startup/setup-stack.sh
+++ b/target/scripts/startup/setup-stack.sh
@@ -456,7 +456,7 @@ function _setup_postfix_sizelimits
 
 function _setup_clamav_sizelimit
 {
-  _log 'trace' "Configuring ClamAV message scan size limit to '${CLAMAV_MESSAGE_SIZE_LIMIT}'"
+  _log 'trace' "Setting ClamAV message scan size limit to '${CLAMAV_MESSAGE_SIZE_LIMIT}'"
   sedfile -i "s/^MaxFileSize.*/MaxFileSize ${CLAMAV_MESSAGE_SIZE_LIMIT}/" /etc/clamav/clamd.conf
 }
 
@@ -476,16 +476,16 @@ function _setup_spoof_protection
   if [[ ${ENABLE_LDAP} -eq 1 ]]
   then
     if [[ -z ${LDAP_QUERY_FILTER_SENDERS} ]]; then
-      postconf -e "smtpd_sender_login_maps = ldap:/etc/postfix/ldap-users.cf ldap:/etc/postfix/ldap-aliases.cf ldap:/etc/postfix/ldap-groups.cf"
+      postconf -e 'smtpd_sender_login_maps = ldap:/etc/postfix/ldap-users.cf ldap:/etc/postfix/ldap-aliases.cf ldap:/etc/postfix/ldap-groups.cf'
     else
-      postconf -e "smtpd_sender_login_maps = ldap:/etc/postfix/ldap-senders.cf"
+      postconf -e 'smtpd_sender_login_maps = ldap:/etc/postfix/ldap-senders.cf'
     fi
   else
     if [[ -f /etc/postfix/regexp ]]
     then
-      postconf -e "smtpd_sender_login_maps = unionmap:{ texthash:/etc/postfix/virtual, hash:/etc/aliases, pcre:/etc/postfix/maps/sender_login_maps.pcre, pcre:/etc/postfix/regexp }"
+      postconf -e 'smtpd_sender_login_maps = unionmap:{ texthash:/etc/postfix/virtual, hash:/etc/aliases, pcre:/etc/postfix/maps/sender_login_maps.pcre, pcre:/etc/postfix/regexp }'
     else
-      postconf -e "smtpd_sender_login_maps = texthash:/etc/postfix/virtual, hash:/etc/aliases, pcre:/etc/postfix/maps/sender_login_maps.pcre"
+      postconf -e 'smtpd_sender_login_maps = texthash:/etc/postfix/virtual, hash:/etc/aliases, pcre:/etc/postfix/maps/sender_login_maps.pcre'
     fi
   fi
 }
@@ -626,10 +626,10 @@ function _setup_SRS
 {
   _log 'debug' 'Setting up SRS'
 
-  postconf -e "sender_canonical_maps = tcp:localhost:10001"
+  postconf -e 'sender_canonical_maps = tcp:localhost:10001'
   postconf -e "sender_canonical_classes = ${SRS_SENDER_CLASSES}"
-  postconf -e "recipient_canonical_maps = tcp:localhost:10002"
-  postconf -e "recipient_canonical_classes = envelope_recipient,header_recipient"
+  postconf -e 'recipient_canonical_maps = tcp:localhost:10002'
+  postconf -e 'recipient_canonical_classes = envelope_recipient,header_recipient'
 }
 
 function _setup_dkim
@@ -1050,9 +1050,12 @@ function _setup_logrotate
 
 function _setup_mail_summary
 {
+  local ENABLED_MESSAGE
+  ENABLED_MESSAGE="Enabling Postfix log summary reports with recipient '${PFLOGSUMM_RECIPIENT}'"
+
   case "${PFLOGSUMM_TRIGGER}" in
     ( 'daily_cron' )
-      _log 'debug' "Enable postfix summary with recipient '${PFLOGSUMM_RECIPIENT}'"
+      _log 'debug' "${ENABLED_MESSAGE}"
       _log 'trace' 'Creating daily cron job for pflogsumm report'
 
       cat >/etc/cron.daily/postfix-summary << EOM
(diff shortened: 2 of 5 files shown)


target/scripts/update-check.sh
```diff
@@ -41,20 +38,21 @@ There is a docker-mailserver update available on your host: $(hostname -f)
 Current version: ${VERSION}
 Latest  version: ${LATEST}
 
-Changelog: ${CHANGELOG}
+Changelog: ${CHANGELOG_URL}
 EOM
       echo "${MAIL}" | mail -s "Mailserver update available! [ ${VERSION} --> ${LATEST} ]" "${POSTMASTER_ADDRESS}" && \
 
-      _log "Info: Update available [ ${VERSION} --> ${LATEST} ]" && \
+      _log 'info' "Update available [ ${VERSION} --> ${LATEST} ]"
```
casperklein, line 45: The removed `&& \` was on purpose, to only exit when notification was successful.

casperklein (changes requested):
You removed the logging of date/time. But that is essential.
I like to see in the log, when update-check runs for the last time.

georglauterbach:
> You removed the logging of date/time. But that is essential. I like to see in the log, when update-check runs for the last time.

I re-added the timestamps to the log messages with the format from `check-for-changes.sh` which I added to `utils.sh` as a common function.

casperklein:
> I re-added the timestamps to the log messages with the format from `check-for-changes.sh` which I added to `utils.sh` as a common function.

Instead of providing in each `_log` call `$(_print_date)`, something more generic would be more easier:

`_log 'debug' "$(_print_date) Remote version information fetched"` --> `_log_date 'debug' "Remote version information fetched"`

```
_log_date() {
  _log "$1" "$(_print_date)$2"
}
```

PS: As we take the output from the function and don't directly print it, `_get_date` sounds better to me.

georglauterbach:
> > I re-added the timestamps to the log messages with the format from `check-for-changes.sh` which I added to `utils.sh` as a common function.
> 
> Instead of providing in each `_log` call `$(_print_date)`, something more generic would be easier:
> 
> `_log 'debug' "$(_print_date) Remote version information fetched"` --> `_log_date 'debug' "Remote version information fetched"`
> 
> ```
> _log_date() {
>   _log "$1" "$(_print_date)$2"
> }
> ```
> 
> PS: As we take the output from the function and don't directly print it, `_get_date` sounds better to me.
> 
> Edit: Until otherwise needed, just include `_print_date` into `_log_date()`: `_log "$1" "$(date "%Y-%m-%d %H:%M:%S")$2"`
> 
> That way, only the log file needs to be sourced, instead of index.sh.

Will do 👍🏼

target/scripts/helpers/log.sh
```diff
@@ -91,6 +91,8 @@ function _log
   fi
 }
 
+function _log_with_date { _log "${1}" "$(date +"%Y-%m-%d %H:%M:%S")  ${2}" ; }
```
casperklein, line 94: >I'd like to use `''` where possible to indicate no variables are expanded here
```suggestion
function _log_with_date { _log "${1}" "$(date '+%Y-%m-%d %H:%M:%S')  ${2}" ; }
```

target/scripts/startup/setup-stack.sh
```diff
@@ -435,7 +435,7 @@ function _setup_postgrey
 
 function _setup_postfix_postscreen
 {
-  _log 'debug' "Configuring postscreen"
+  _log 'debug' "Configuring Postscreen"
```
casperklein, line 438:
```suggestion
  _log 'debug' 'Configuring Postscreen'
```

target/scripts/startup/setup-stack.sh
```diff
@@ -643,28 +643,28 @@ function _setup_dkim
...
+    _log 'trace' "DKIM keys added for: $(ls /etc/opendkim/keys/)"
+    _log 'trace' "Changing permissions on '/etc/opendkim'"
 
     chown -R opendkim:opendkim /etc/opendkim/
     chmod -R 0700 /etc/opendkim/keys/
   else
-    _log 'warn' 'No DKIM key provided. Check the documentation on how to get your keys.'
-    [[ ! -f "/etc/opendkim/KeyTable" ]] && touch "/etc/opendkim/KeyTable"
+    _log 'warn' 'No DKIM key(s) provided - check the documentation on how to get your keys'
+    [[ ! -f '/etc/opendkim/KeyTable' ]] && touch '/etc/opendkim/KeyTable'
```
casperklein, line 653:
```suggestion
    [[ ! -f /etc/opendkim/KeyTable ]] && touch /etc/opendkim/KeyTable
```

casperklein (changes requested):

target/scripts/update-check.sh
```diff
@@ -41,20 +38,19 @@ There is a docker-mailserver update available on your host: $(hostname -f)
...
       echo "${MAIL}" | mail -s "Mailserver update available! [ ${VERSION} --> ${LATEST} ]" "${POSTMASTER_ADDRESS}" && \
 
-      _log "Info: Update available [ ${VERSION} --> ${LATEST} ]" && \
-
       # only notify once
-      exit 0
+      _log_with_date 'info' "Update available [ ${VERSION} --> ${LATEST} ]" && exit 0
     else
-      _log "Info: No update available"
+      _log_with_date 'debug' 'No update available'
```
casperklein, line 48: This should also be `info`, so that once a day, there is a log line with either update available or not.

```suggestion
      _log_with_date 'info' 'No update available'
```

target/scripts/startup/setup-stack.sh
```diff
@@ -811,10 +811,10 @@ function _setup_postfix_override_configuration
     done < /tmp/docker-mailserver/postfix-master.cf
     _log 'trace' "Loaded '/tmp/docker-mailserver/postfix-master.cf'"
   else
-    _log 'trace' "No extra postfix settings loaded because optional '/tmp/docker-mailserver/postfix-master.cf' not provided."
+    _log 'trace' "No extra Üostfix settings loaded because optional '/tmp/docker-mailserver/postfix-master.cf' not provided"
```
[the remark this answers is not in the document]
polarathene:
```suggestion
    _log 'trace' "No extra Postfix settings loaded because optional '/tmp/docker-mailserver/postfix-master.cf' was not provided"
```

polarathene (approved):
Not much to add, most of the concerns I marked up for review earlier was already caught and addressed by you both :sweat_smile: 

None of these changes are all that important, feel free to dismiss them.

target/scripts/startup/setup-stack.sh
```diff
@@ -1066,14 +1065,15 @@ EOM
       ;;
 
     ( 'logrotate' )
+      _log 'debug' "Enable postfix summary with recipient '${PFLOGSUMM_RECIPIENT}'"
```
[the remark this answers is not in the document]
georglauterbach: Will do.

target/scripts/startup/setup-stack.sh
```diff
@@ -968,7 +968,7 @@ EOM
   # ClamAV
   if [[ ${ENABLE_CLAMAV} -eq 0 ]]
   then
-    _log 'warn' "ClamAV is disabled. You can enable it with 'ENABLE_CLAMAV=1'"
+    _log 'info' "ClamAV is disabled"
```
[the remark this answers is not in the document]
georglauterbach: :D

target/scripts/startup/setup-stack.sh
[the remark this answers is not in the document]
georglauterbach: Done in [e467607](https://github.com/docker-mailserver/docker-mailserver/pull/2499/commits/e4676074d02beb4ba422e842f05ff94fca76062f)

target/scripts/startup/setup-stack.sh
```diff
@@ -444,48 +444,48 @@ function _setup_postfix_postscreen
...
 {
-  _log 'trace' "Configuring Spoof Protection"
+  _log 'trace' 'Configuring spoof protection'
   sed -i \
     's|smtpd_sender_restrictions =|smtpd_sender_restrictions = reject_authenticated_sender_login_mismatch,|' \
     /etc/postfix/main.cf
 
   if [[ ${ENABLE_LDAP} -eq 1 ]]
   then
     if [[ -z ${LDAP_QUERY_FILTER_SENDERS} ]]; then
```
polarathene, line 478: Something you might be interested in :sweat_smile: 

```suggestion
    if [[ -z ${LDAP_QUERY_FILTER_SENDERS} ]]
    then
```

polarathene (approved):

target/scripts/startup/setup-stack.sh
georglauterbach: 😆

polarathene (dismissed):

target/scripts/update-check.sh
```diff
@@ -41,20 +38,19 @@ There is a docker-mailserver update available on your host: $(hostname -f)
...
-Changelog: ${CHANGELOG}
+Changelog: ${CHANGELOG_URL}
 EOM
       echo "${MAIL}" | mail -s "Mailserver update available! [ ${VERSION} --> ${LATEST} ]" "${POSTMASTER_ADDRESS}" && \
 
-      _log "Info: Update available [ ${VERSION} --> ${LATEST} ]" && \
-
       # only notify once
-      exit 0
+      _log_with_date 'info' "Update available [ ${VERSION} --> ${LATEST} ]" && exit 0
```
casperklein, line 46: I find the old structure more clear. "only notify once" explains `exit 0`

georglauterbach: With all due respect, I had difficulties seeing that an

``` BASH
... && \

# comment here
exit 1
```

all belongs together, because I had never seen a line break escape following a comment following the next command to execute if the command on the first line succeeded. Therefore, I disagree, I think the new structure is way more pleasant to read.

casperklein:
```bash
echo "${MAIL}" | mail -s "Mailserver update available! [ ${VERSION} --> ${LATEST} ]" "${POSTMASTER_ADDRESS}" && \
_log_with_date 'info' "Update available [ ${VERSION} --> ${LATEST} ]" && \
# only notify once
exit 0
```

If you still want the `exit` on the same line as `_log_with_date`, then at least the empty line between `echo` and `_log_with_date` should be removed, to make clear `_log_with_date` depends on the success of sending the mail.

casperklein: Or even better, remove `&& \` from the `echo` line. The info should be logged, regardless of the success of sending the mail.

casperklein: On the other, then there will be no further tries to send the email, because `exit 0` will stop the update-check.

Rare case: `update-check` tries to send the email, while `check-for-changes` kicks in and restarts postfix at the same time. Mail deliver fails, "update available" is logged and `update-check stops`. The admin won't be notified by mail in that case.
But this is the primary aim of the `update-checker`.

So I propose to either remove the empty line or take my proposal above with no empty line and exit on a separate line.

casperklein: Final final proposal 😆

I switched the `_log_with_date` and `echo` line. This means

1. Create a log entry every time
2. Stop `update-check` only, if mail was successful sent.

```bash
_log_with_date 'info' "Update available [ ${VERSION} --> ${LATEST} ]"
echo "${MAIL}" | mail -s "Mailserver update available! [ ${VERSION} --> ${LATEST} ]" "${POSTMASTER_ADDRESS}" && \
# only notify
exit 0
```

or

```bash
_log_with_date 'info' "Update available [ ${VERSION} --> ${LATEST} ]"
# only notify
echo "${MAIL}" | mail -s "Mailserver update available! [ ${VERSION} --> ${LATEST} ]" "${POSTMASTER_ADDRESS}" && exit 0
```

georglauterbach: Done in e2211394774b18cf2906c4e59e4cd7278562077e

polarathene (approved):

casperklein (approved):

PR #2500, opened 2022-03-21
scripts: refactored scripts located under `target/bin/`

georglauterbach:
# Description

The scripts under `target/bin/` now use the new log and I replaced some
`""` with `''` on the way. The functionality stays the same, this mostly
style and log.

## Type of change

## Checklist:

### DIFF
--- a/target/bin/addmailuser
+++ b/target/bin/addmailuser
@@ -39,8 +39,8 @@ FULL_EMAIL="${1}"
 shift
 PASSWD="${*}"
 
-[[ -z ${FULL_EMAIL} ]] && { __usage ; _errex 'No username specified' ; }
-[[ ${FULL_EMAIL} =~ .*\@.* ]] || { __usage ; _errex 'Username must include the domain' ; }
+[[ -z ${FULL_EMAIL} ]] && { __usage ; _exit_with_error 'No username specified' ; }
+[[ ${FULL_EMAIL} =~ .*\@.* ]] || { __usage ; _exit_with_error 'Username must include the domain' ; }
 
 touch "${DATABASE}"
 _create_lock # Protect config file with lock to avoid race conditions
@@ -54,7 +54,7 @@ if [[ -z ${PASSWD} ]]
 then
   read -r -s -p "Enter Password: " PASSWD
   echo
-  [[ -z ${PASSWD} ]] && _errex "Password must not be empty"
+  [[ -z ${PASSWD} ]] && _exit_with_error "Password must not be empty"
 fi
 
 HASH="$(doveadm pw -s SHA512-CRYPT -u "${FULL_EMAIL}" -p "${PASSWD}")"
--- a/target/bin/addmailuser
+++ b/target/bin/addmailuser
@@ -46,7 +46,7 @@ touch "${DATABASE}"
 _create_lock # Protect config file with lock to avoid race conditions
 if grep -qi "^$(_escape "${FULL_EMAIL}")|" "${DATABASE}"
 then
-  echo "User '${FULL_EMAIL}' already exists."
+  _log 'warn' "User '${FULL_EMAIL}' already exists"
   exit 1
 fi
 
@@ -64,11 +64,11 @@ USER="${FULL_EMAIL%@*}"
 DOMAIN="${FULL_EMAIL#*@}"
 
 # Tests fail if the creation of /var/mail/${DOMAIN}/${USER} doesn't happen fast enough after addmailuser executes (check-for-changes.sh race-condition)
-if [[ -e "/tmp/docker-mailserver-config-chksum" ]] # Prevent infinite loop in tests like "checking accounts: user3 should have been added to /tmp/docker-mailserver/postfix-accounts.cf even when that file does not exist"
+if [[ -e /tmp/docker-mailserver-config-chksum ]] # Prevent infinite loop in tests like "checking accounts: user3 should have been added to /tmp/docker-mailserver/postfix-accounts.cf even when that file does not exist"
 then
   while [[ ! -d "/var/mail/${DOMAIN}/${USER}" ]]
   do
-    echo "Waiting for dovecot to create /var/mail/${DOMAIN}/${USER}..."
+    _log 'info' "Waiting for dovecot to create '/var/mail/${DOMAIN}/${USER}/'"
     sleep 1
   done
 fi
--- a/target/bin/addmailuser
+++ b/target/bin/addmailuser
@@ -46,8 +46,7 @@ touch "${DATABASE}"
 _create_lock # Protect config file with lock to avoid race conditions
 if grep -qi "^$(_escape "${FULL_EMAIL}")|" "${DATABASE}"
 then
-  _log 'warn' "User '${FULL_EMAIL}' already exists"
-  exit 1
+  _exit_with_error "User '${FULL_EMAIL}' already exists"
 fi
 
 if [[ -z ${PASSWD} ]]
--- a/target/bin/delmailuser
+++ b/target/bin/delmailuser
@@ -74,7 +74,7 @@ do
 done
 shift $((OPTIND-1))
 
-[[ -z ${*} ]] && { __usage ; _exit_with_error "No user specifed" ; }
+[[ -z ${*} ]] && { __usage ; _exit_with_error 'No user specifed' ; }
 [[ -s ${DATABASE} ]] || exit 0
 
 if ! ${MAILDEL}
--- a/target/bin/delmailuser
+++ b/target/bin/delmailuser
@@ -74,12 +74,12 @@ do
 done
 shift $((OPTIND-1))
 
-[[ -z ${*} ]] && { __usage ; _exit_with_error 'No user specifed' ; }
+[[ -z ${*} ]] && { __usage ; _exit_with_error 'No user specified' ; }
 [[ -s ${DATABASE} ]] || exit 0
 
 if ! ${MAILDEL}
 then
-  read -r -p "Do you want to delete the mailbox as well (removing all mails) ? [Y/n] " MAILDEL_CHOSEN
+  read -r -p "Do you want to delete the mailbox as well (removing all mails)? [Y/n] " MAILDEL_CHOSEN
   if [[ ${MAILDEL_CHOSEN} =~ (y|Y|yes|Yes) ]] || [[ -z ${MAILDEL_CHOSEN} ]]
   then
     MAILDEL=true
--- a/target/bin/delmailuser
+++ b/target/bin/delmailuser
@@ -67,14 +67,14 @@ do
 
     * )
       __usage
-      _errex "The option '${OPT}' is unknown"
+      _exit_with_error "The option '${OPT}' is unknown."
       ;;
 
   esac
 done
 shift $((OPTIND-1))
 
-[[ -z ${*} ]] && { __usage ; _errex "No user specifed" ; }
+[[ -z ${*} ]] && { __usage ; _exit_with_error "No user specifed" ; }
 [[ -s ${DATABASE} ]] || exit 0
 
 if ! ${MAILDEL}
@@ -155,7 +155,7 @@ use 'sudo docker exec mailserver rm -R /var/mail/${DOMAIN}/${USER}'"
     ERROR=true
   fi
 
-  ${ERROR} && _errex 'See the error message(s) above'
+  ${ERROR} && _exit_with_error 'See the messages above.'
 done
 
 exit 0
--- a/target/bin/delmailuser
+++ b/target/bin/delmailuser
@@ -67,7 +67,7 @@ do
 
     * )
       __usage
-      _errex "The option ${OPT} is unknown."
+      _errex "The option '${OPT}' is unknown"
       ;;
 
   esac
@@ -103,7 +103,7 @@ do
   then
     if ! sedfile --strict -i "/^${EMAIL}|/d" "${DATABASE}"
     then
-      echo "${UNESCAPED_EMAIL} couldn't be deleted in ${DATABASE}." >&2
+      _log 'error' "'${UNESCAPED_EMAIL}' couldn't be deleted in '${DATABASE}'" >&2
       ERROR=true
     fi
   fi
@@ -116,9 +116,9 @@ do
       -e "/ ${EMAIL}$/d" -e "s/,${EMAIL}//g" -e "s/${EMAIL},//g" \
       "${ALIAS_DATABASE}"
     then
-      echo "${UNESCAPED_EMAIL} and potential aliases deleted."
+      _log 'info' "'${UNESCAPED_EMAIL}' and potential aliases deleted"
     else
-      echo "Aliases for ${UNESCAPED_EMAIL} couldn't be deleted in ${ALIAS_DATABASE}." >&2
+      _log 'error' "Aliases for '${UNESCAPED_EMAIL}' couldn't be deleted in '${ALIAS_DATABASE}'" >&2
       ERROR=true
     fi
   fi
@@ -128,7 +128,7 @@ do
   then
     if ! sedfile --strict -i -e "/^${EMAIL}:.*$/d" "${QUOTA_DATABASE}"
     then
-      echo "Quota for ${UNESCAPED_EMAIL} couldn't be deleted in ${QUOTA_DATABASE}." >&2
+      _log 'warn' "Quota for '${UNESCAPED_EMAIL}' couldn't be deleted in '${QUOTA_DATABASE}'" >&2
     fi
   fi
 
@@ -144,17 +144,18 @@ use 'sudo docker exec mailserver rm -R /var/mail/${DOMAIN}/${USER}'"
   then
     if rm -R "/var/mail/${DOMAIN}/${USER}"
     then
-      echo "Mailbox deleted."
+      _log 'info' 'Mailbox deleted'
     else
-      echo "Mailbox couldn't be deleted." >&2
+      _log 'error' 'Mailbox could not be deleted' >&2
       ERROR=true
     fi
-    rmdir "/var/mail/${DOMAIN}" 2>/dev/null || true
+    rm -rf "/var/mail/${DOMAIN}"
   else
-    echo "Mailbox directory '/var/mail/${DOMAIN}/${USER}' did not exist." >&2
+    log 'error' "Mailbox directory '/var/mail/${DOMAIN}/${USER}' did not exist" >&2
     ERROR=true
   fi
 
-  ${ERROR} && _errex 'See the messages above.'
+  ${ERROR} && _errex 'See the error message(s) above'
 done
+
 exit 0
--- a/target/bin/delmailuser
+++ b/target/bin/delmailuser
@@ -55,19 +55,19 @@ fi
 
 while getopts ":yYh" OPT
 do
-  case ${OPT} in
-    y | Y )
+  case "${OPT}" in
+    ( 'y' | 'Y' )
       MAILDEL=true
       ;;
 
-    h )
+    ( 'h' )
       __usage
       exit 0
       ;;
 
-    * )
+    ( * )
       __usage
-      _exit_with_error "The option '${OPT}' is unknown."
+      _exit_with_error "The option '${OPT}' is unknown"
       ;;
 
   esac
@@ -103,7 +103,7 @@ do
   then
     if ! sedfile --strict -i "/^${EMAIL}|/d" "${DATABASE}"
     then
-      _log 'error' "'${UNESCAPED_EMAIL}' couldn't be deleted in '${DATABASE}'" >&2
+      _log 'error' "'${UNESCAPED_EMAIL}' couldn't be deleted in '${DATABASE}'"
       ERROR=true
     fi
   fi
@@ -118,7 +118,7 @@ do
     then
       _log 'info' "'${UNESCAPED_EMAIL}' and potential aliases deleted"
     else
-      _log 'error' "Aliases for '${UNESCAPED_EMAIL}' couldn't be deleted in '${ALIAS_DATABASE}'" >&2
+      _log 'error' "Aliases for '${UNESCAPED_EMAIL}' couldn't be deleted in '${ALIAS_DATABASE}'"
       ERROR=true
     fi
   fi
@@ -128,7 +128,7 @@ do
   then
     if ! sedfile --strict -i -e "/^${EMAIL}:.*$/d" "${QUOTA_DATABASE}"
     then
-      _log 'warn' "Quota for '${UNESCAPED_EMAIL}' couldn't be deleted in '${QUOTA_DATABASE}'" >&2
+      _log 'warn' "Quota for '${UNESCAPED_EMAIL}' couldn't be deleted in '${QUOTA_DATABASE}'"
     fi
   fi
 
@@ -146,12 +146,12 @@ use 'sudo docker exec mailserver rm -R /var/mail/${DOMAIN}/${USER}'"
     then
       _log 'info' 'Mailbox deleted'
     else
-      _log 'error' 'Mailbox could not be deleted' >&2
+      _log 'error' 'Mailbox could not be deleted'
       ERROR=true
     fi
-    rm -rf "/var/mail/${DOMAIN}"
+    rmdir "/var/mail/${DOMAIN}" &>/dev/null
   else
-    log 'error' "Mailbox directory '/var/mail/${DOMAIN}/${USER}' did not exist" >&2
+    log 'error' "Mailbox directory '/var/mail/${DOMAIN}/${USER}' did not exist"
     ERROR=true
   fi
 
--- a/target/bin/fail2ban
+++ b/target/bin/fail2ban
@@ -68,7 +68,7 @@ else
         done
 
       else
-        log 'warn' "You need to specify an IP address. Run './setup.sh debug fail2ban' to get a list of banned IP addresses."
+        _log 'warn' "You need to specify an IP address: Run './setup.sh debug fail2ban' to get a list of banned IP addresses"
         exit 0
       fi
       ;;
--- a/target/bin/fail2ban
+++ b/target/bin/fail2ban
@@ -75,7 +75,7 @@ else
 
     ( * )
       usage
-      _errex "Unknown command: ${1}"
+      _exit_with_error "Unknown command '${1}'"
       ;;
 
   esac
--- a/target/bin/fail2ban
+++ b/target/bin/fail2ban
@@ -19,7 +19,7 @@ Possible causes for this error are
 "
 fi
 
-function usage { echo "Usage: ${0} [<unban> <ip-address>]" ; }
+function __usage { echo "Usage: ${0} [<unban> <ip-address>]" ; }
 
 unset JAILS
 declare -a JAILS
@@ -72,7 +72,7 @@ else
       ;;
 
     ( * )
-      usage
+      __usage
       _exit_with_error "Unknown command '${1}'"
       ;;
 
--- a/target/bin/fail2ban
+++ b/target/bin/fail2ban
@@ -39,7 +39,7 @@ then
 
     if [[ -n ${BANNED_IP} ]]
     then
-      echo -e "Banned in ${JAIL}: ${BANNED_IP//$'\n'/, }"
+      echo "Banned in ${JAIL}: ${BANNED_IP//$'\n'/, }"
       IP_COUNT=$(( IP_COUNT + 1 ))
     fi
   done
--- a/target/bin/fail2ban
+++ b/target/bin/fail2ban
@@ -16,7 +16,7 @@ Possible causes for this error are
 3. Host is configured incorrectly
 
 Aborting...
-"
+" >&2
 
   exit 1
 fi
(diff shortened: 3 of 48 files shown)


target/bin/sedfile
```diff
@@ -11,36 +11,34 @@
...
-OLD=$(${HASHTOOL} "${FILE}")
-sed "$@"
-NEW=$(${HASHTOOL} "${FILE}")
+OLD="$(${HASHTOOL} "${FILE}")"
+sed "${@}"
+NEW="$(${HASHTOOL} "${FILE}")"
 
 # fail if file was not modified
-if [[ ${OLD} == "${NEW}" ]] && [[ ${SKIP_ERROR} -eq 0 ]]
+if [[ ${OLD} == "${NEW}" ]] && ! ${SKIP_ERROR}
```
casperklein, line 39: This is bad practice. Take a look at this example why:

```
# foo=false
[ $foo ] && echo "Booyaka!"
Booyaka!
```

Integer comparing is fine here.

target/bin/sedfile
```diff
@@ -11,36 +11,34 @@
...
 
 # get last argument
 FILE=${*: -1}
 
-OLD=$(${HASHTOOL} "${FILE}")
-sed "$@"
-NEW=$(${HASHTOOL} "${FILE}")
+OLD="$(${HASHTOOL} "${FILE}")"
+sed "${@}"
+NEW="$(${HASHTOOL} "${FILE}")"
```
casperklein, line 36: When assigning a value to a variable and using `$()`, no quoting is required.
```suggestion
OLD=$(${HASHTOOL} "${FILE}")
sed "${@}"
NEW=$(${HASHTOOL} "${FILE}")
```

target/bin/sedfile
polarathene: I have seen usage of `ERROR` var with `true`/`false` values in this PR (but not the affected diff itself), probably applies to that too? That one used `${ERROR} &&` which would be true for both bool values unlike numbers?

target/bin/sedfile
```diff
@@ -11,36 +11,35 @@
...
 
 # fail if file was not modified
-if [[ ${OLD} == "${NEW}" ]] && [[ ${SKIP_ERROR} -eq 0 ]]
+if [[ ${OLD} == "${NEW}" ]] && ! ${SKIP_ERROR}
 then
-  echo "Error: sed $*"
+  echo "Error: sed ${*}" >&2
   exit 1
-fi >&2
+fi
```
polarathene, line 43: Any reason this echo isn't using log like others? I'm not quite sure what the decision process is for when to convert.

The if statements that end with `fi >&2` seem odd to me, so I like that you addressed that here, it's present elsewhere in files touched by this PR if that is of interest.

target/bin/setquota
```diff
@@ -15,38 +15,38 @@ USER="${1}"
...
+  [[ -z "${QUOTA}" ]] && _exit_with_error 'Quota must not be empty. Use 0 for unlimited quota'
 fi
 
 # check quota
 if [[ -n ${QUOTA} ]] && ! echo "${QUOTA}" | grep -qE "^([0-9]+(B|k|M|G|T)|0)\$"
 then
   usage
-  _exit_with_error "invalid quota format. e.g. 302M (B (byte), k (kilobyte), M (megabyte), G (gigabyte) or T (terabyte))"
+  _exit_with_error 'Invalid quota format. e.g. 302M (B (byte), k (kilobyte), M (megabyte), G (gigabyte) or T (terabyte))'
 fi
```
polarathene, line 50: It looks like this block of code is a 2nd instance. I'm not sure why there are two copies in close proximity of each other.

Probably more maintainable as a function with two calls if having both instances is intentional?

target/bin/restrict-access
```diff
@@ -19,20 +19,20 @@ case ${2} in
     ;;
   *)
     usage
-    _exit_with_error "missing parameters. Specify \"send\" or \"receive\""
+    _exit_with_error "Missing parameters. Specify 'send' or 'receive'"
     ;;
 esac
```
polarathene, line 24: The cases here were not converted to the style you've changed to further down in the file.

target/bin/delmailuser
```diff
@@ -144,17 +144,18 @@ use 'sudo docker exec mailserver rm -R /var/mail/${DOMAIN}/${USER}'"
...
     else
-      echo "Mailbox couldn't be deleted." >&2
+      _log 'error' 'Mailbox could not be deleted' >&2
       ERROR=true
     fi
-    rmdir "/var/mail/${DOMAIN}" 2>/dev/null || true
+    rm -rf "/var/mail/${DOMAIN}"
   else
-    echo "Mailbox directory '/var/mail/${DOMAIN}/${USER}' did not exist." >&2
+    log 'error' "Mailbox directory '/var/mail/${DOMAIN}/${USER}' did not exist" >&2
```
polarathene, line 154: This occurs multiple times in this file, the lines have been changed to log with `warn`/`error` status, but you retained the `>&2` at the end of the original echo command? These two log levels already do `>&2` internally right?

target/bin/delmailuser
```diff
@@ -67,7 +67,7 @@ do
 
     * )
       __usage
-      _exit_with_error "The option ${OPT} is unknown."
+      _exit_with_error "The option '${OPT}' is unknown."
```
polarathene, line 70: Not this line specifically, but right above it I noticed the case statement style is not what you've been adjusting to. Just leaving this as a note for you so you're aware of it due to this PR already making such a change elsewhere.

target/bin/addmailuser
```diff
@@ -46,7 +46,7 @@ touch "${DATABASE}"
 _create_lock # Protect config file with lock to avoid race conditions
 if grep -qi "^$(_escape "${FULL_EMAIL}")|" "${DATABASE}"
 then
-  echo "User '${FULL_EMAIL}' already exists."
+  _log 'warn' "User '${FULL_EMAIL}' already exists"
```
polarathene, line 49: If the script exits after the log, is that not an error?

polarathene (changes requested):

target/bin/sedfile
georglauterbach: I will revert back.

target/bin/sedfile
georglauterbach: 👍🏼

target/bin/delmailuser
```diff
@@ -144,17 +144,18 @@ use 'sudo docker exec mailserver rm -R /var/mail/${DOMAIN}/${USER}'"
...
     then
-      echo "Mailbox deleted."
+      _log 'info' 'Mailbox deleted'
     else
-      echo "Mailbox couldn't be deleted." >&2
+      _log 'error' 'Mailbox could not be deleted' >&2
       ERROR=true
     fi
-    rmdir "/var/mail/${DOMAIN}" 2>/dev/null || true
+    rm -rf "/var/mail/${DOMAIN}"
```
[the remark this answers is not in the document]
georglauterbach: Right, this would not be ideal. I will revert it back to `rmdir`.

The purpose of `|| true` is to mask the return value. `rmdir` will fail and exit with `1` if the directory was not empty, and therefore, the whole command would fail. This is however not needed here since we do not use `set -e`.

target/bin/fail2ban
```diff
@@ -16,7 +16,7 @@ Possible causes for this error are
 3. Host is configured incorrectly
 
 Aborting...
-"
+" >&2
```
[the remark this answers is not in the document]
georglauterbach: I used `_exit_with_error` and one big string. This should be fine :D

target/bin/fail2ban
```diff
@@ -41,21 +41,21 @@ then
 
     if [[ -n ${BANNED_IP} ]]
     then
-      echo "Banned in ${JAIL}: ${BANNED_IP//$'\n'/, }"
+      echo -e "Banned in ${JAIL}: ${BANNED_IP//$'\n'/, }"
```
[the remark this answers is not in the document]
georglauterbach: I was thinking about it, and I believe this is indeed not desirable here. Not sure though why I added `-e`... I'd like to not use `_log` in this case :)

target/bin/listmailuser
```diff
@@ -16,15 +16,16 @@ function dovecot_quota_to_hr()
   then
     echo $(( 1024 * ${1} )) | numfmt --to=iec
   else
-    notify 'err' "Supplied non-number argument '${1:-}' to 'dovecot_quota_to_hr()' in script 'listmailuser'"
+    _exit_with_error "Supplied non-number argument '${1:-}' to 'dovecot_quota_to_hr()' in script 'listmailuser'"
+    exit 1
```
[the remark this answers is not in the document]
georglauterbach: Because I forgot 😆

target/bin/open-dkim
```diff
@@ -52,47 +55,47 @@ function __usage
...
+  case "${1}" in
+    ( 'keysize' )
       if [[ -n ${2+'set'} ]]
       then
         KEYSIZE="${2}"
         shift
         shift
       else
-        echo "No keysize provided after 'keysize' argument. Aborting." >&2
+        _log 'error' "No keysize provided after 'keysize' argument"
```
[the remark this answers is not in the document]
georglauterbach: A good idea. Whether we exit with `2` or with `1` does not actually make a difference. I will adjust.

target/bin/restrict-access
```diff
@@ -47,16 +47,17 @@ case ${MODE} in
...
-  list)
+  ( 'list' )
     grep "REJECT" "${DATABASE}" 2>/dev/null || echo "Everyone is allowed to ${2} mails."
     ;;
-  *)
+
+  ( * )
     usage
-    _exit_with_error "missing mode. Specify \"add\", \"del\" or \"list\""
+    _exit_with_error "missing mode: specify \"add\", \"del\" or \"list\""
```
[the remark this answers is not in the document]
georglauterbach: Good catch!

target/bin/restrict-access
```diff
@@ -47,16 +47,17 @@ case ${MODE} in
     echo -e "${USER} \t\t REJECT" >>"${DATABASE}"
   ;;
 
-  del)
-    sed -ie "/^$(_escape "${USER}")/d" "${DATABASE}" 2>/dev/null || _exit_with_error "User \"${USER}\" not found."
-  ;;
+  ( 'del' )
+    sed -ie "/^$(_escape "${USER}")/d" "${DATABASE}" 2>/dev/null || _exit_with_error "User '${USER}' not found."
+    ;;
```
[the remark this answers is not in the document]
georglauterbach: I adjusted the other ones as well. This indentation is consistent with our code base, but the file it not :)

target/bin/sedfile
georglauterbach: This `if` statement only logs one message, so I thought it is fine to place the redirection right behind it. The other statement above logs more than one, all going to `stderr`, so I left it after the `fi`.

target/bin/setquota
georglauterbach: I will outsource this into a variable :)

target/bin/restrict-access
georglauterbach: 👍🏼

target/bin/delmailuser
georglauterbach: I overlooked this :) Thanks for noting it!

target/bin/delmailuser
georglauterbach: 👍🏼

target/bin/addmailuser
georglauterbach: Yes, yes it is :D

georglauterbach: Done in 08a05debd67e69c3e5010be547c0cd0ca65a28c7

target/bin/delmailuser
georglauterbach: Done in 08a05debd67e69c3e5010be547c0cd0ca65a28c7

target/bin/sedfile
georglauterbach: Done in 08a05debd67e69c3e5010be547c0cd0ca65a28c7

target/bin/sedfile
georglauterbach: Done in 08a05debd67e69c3e5010be547c0cd0ca65a28c7

target/bin/delmailuser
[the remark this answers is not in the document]
georglauterbach: Done in 08a05debd67e69c3e5010be547c0cd0ca65a28c7

target/bin/fail2ban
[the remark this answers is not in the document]
georglauterbach: Done in 08a05debd67e69c3e5010be547c0cd0ca65a28c7

target/bin/fail2ban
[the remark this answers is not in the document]
georglauterbach: Done in 08a05debd67e69c3e5010be547c0cd0ca65a28c7 (left the `-e` in there, just in case).

target/bin/listmailuser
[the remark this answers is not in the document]
georglauterbach: Done in 08a05debd67e69c3e5010be547c0cd0ca65a28c7

target/bin/open-dkim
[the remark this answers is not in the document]
georglauterbach: Done in 08a05debd67e69c3e5010be547c0cd0ca65a28c7

target/bin/restrict-access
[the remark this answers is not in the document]
georglauterbach: Done in 08a05debd67e69c3e5010be547c0cd0ca65a28c7

target/bin/restrict-access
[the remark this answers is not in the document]
georglauterbach: Done in 08a05debd67e69c3e5010be547c0cd0ca65a28c7

target/bin/setquota
georglauterbach: Done in 08a05debd67e69c3e5010be547c0cd0ca65a28c7

target/bin/restrict-access
georglauterbach: Done in 08a05debd67e69c3e5010be547c0cd0ca65a28c7

target/bin/delmailuser
georglauterbach: Done in 08a05debd67e69c3e5010be547c0cd0ca65a28c7

georglauterbach:
> Was it intentional to use `_exit_with_error` here and a multi-line message?
> 
> I assume so with the deletions, but you chose to keep the `>&2`?

Oh my, I again forgot to remove the redirection. It was intentional to use a multi-line message. This should be just fine I think. There is no need for the output to all be on one line, but I could make this a bit more clear. Next commit will correct this :)

georglauterbach:
> I'm still curious why the entire conditional statement is repeated, seems unnecessary?

It seems as if I misunderstood you. Which conditional statement are you referring to?

target/bin/setquota
```diff
@@ -10,43 +10,44 @@ source /usr/local/bin/helpers/index.sh
...
+  [[ -z "${QUOTA}" ]] && _exit_with_error 'Quota must not be empty (use 0 for unlimited quota)'
 fi
 
 # check quota
 if [[ -n ${QUOTA} ]] && ! echo "${QUOTA}" | grep -qE "^([0-9]+(B|k|M|G|T)|0)\$"
 then
   usage
-  _exit_with_error "invalid quota format. e.g. 302M (B (byte), k (kilobyte), M (megabyte), G (gigabyte) or T (terabyte))"
+  _exit_with_error "${MESSAGE_INVALID_FORMAT}"
 fi
```
polarathene, line 51: This doesn't address the concern fully ([_originally raised in this review comment_](https://github.com/docker-mailserver/docker-mailserver/pull/2500#discussion_r831638059)), but is an improvement at least.

I'm still curious why the entire conditional statement is repeated, seems unnecessary?

polarathene (dismissed):

target/bin/setquota
georglauterbach: It actually is :D Addressed in 9d3eda4879b1763f01dec55074ad4d72706eeb7d

polarathene (approved):

georglauterbach:
It seems I was 1 minute too late. I noticed that `usage` was used inconsistently @polarathene - sorry for getting on your nerves with reviewing this (again 😆)... I promise this was my final commit :D

polarathene (dismissed):
LGTM :+1:

target/bin/fail2ban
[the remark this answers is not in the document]
casperklein: > left the -e in there, just in case

Why?

target/bin/sedfile
```diff
@@ -11,19 +11,18 @@
...
-  echo
+  echo 'At least three parameters must be given'
+  echo -e "Usage: ${0} -i <replace/delete operation> <file>\n"
   exit 1
 fi >&2
 
-[[ -f /CONTAINER_START ]] && SKIP_ERROR=1 # Hide error, if container was restarted.
-if [[ "${1}" == "--strict" ]]         # Show error every time.
+[[ -f /CONTAINER_START ]] && SKIP_ERROR=1 # hide error if container was restarted
+if [[ ${1} == '--strict' ]]                  # show error every time
```
casperklein, line 25: This should align with the comment above

```suggestion
if [[ ${1} == '--strict' ]]               # show error every time
```

target/bin/sedfile
casperklein: We should use `_log` here instead of `echo`?

target/bin/sedfile
```diff
@@ -11,19 +11,18 @@
...
+HASHTOOL='sha1sum'
 SKIP_ERROR=0
 
 if [[ $# -lt 3 ]]
 then
-  echo "Error:  At least, three parameters must be given."
-  echo "Syntax: sedfile -i <replace/delete operation> <file>"
-  echo
+  echo 'At least three parameters must be given'
+  echo -e "Usage: ${0} -i <replace/delete operation> <file>\n"
```
casperklein, line 20: Personal gusto. I found the previous "3 echo solution" more readable, don't you?

target/bin/setquota
```diff
@@ -15,39 +15,33 @@ USER="${1}"
 shift
 QUOTA="${*}"
 
-function usage  { echo "Usage: setquota <user@domain> [<quota>]" ; }
+function __usage { echo 'Usage: setquota <user@domain> [<quota>]' ; }
 
-[[ -z ${USER} ]] && { usage ; _exit_with_error "no username specified" ; }
-[[ ${USER} =~ .*\@.* ]] || { usage ; _exit_with_error "username must include the domain" ; }
+[[ -z ${USER} ]] && { __usage ; _exit_with_error "No username specified" ; }
```
casperklein, line 20: Is it necessary to print `__usage` on every error? Should be removed IMO.

target/bin/updatemailuser
```diff
@@ -14,15 +14,15 @@ USER="${1}"
 shift
 PASSWD="${*}"
 
-function usage  {	echo "Usage: updatemailuser <user@domain.tld> [password]" ; }
+function __usage { echo 'Usage: updatemailuser <user@domain.tld> [password]' ; }
 
-[[ -z ${USER} ]] && { usage ; _exit_with_error "no username specified" ; }
+[[ -z ${USER} ]] && { __usage ; _exit_with_error 'No username specified' ; }
```
casperklein, line 19:
```suggestion
[[ -z ${USER} ]] && { _exit_with_error 'No username specified' ; }
```

casperklein (changes requested):

target/bin/sedfile
georglauterbach: If we're using `_log` as said above, the first message will become a log message and the `Usage: ...` will be outsourced in a function to be coherent with all the other scripts.

target/bin/setquota
georglauterbach: I think it is because we have no `--help` flag, so every wrong usage should print how you use it correctly. Otherwise, how do you justify where to put the `__usage` and where not?

target/bin/sedfile
georglauterbach: 👍🏼

target/bin/setquota
polarathene: FWIW, using python was pretty easy for little CLI tools with help generated automatically from your args. I know that you all prefer bash though, so nevermind  :stuck_out_tongue: 

https://github.com/docker-mailserver/docker-mailserver/blob/f22d75fa623675f5fa7dbb05316221dd1f02b07d/target/bin/acme_extract.py#L4-L13

georglauterbach: > FWIW, using python was pretty easy for little CLI tools with help generated automatically from your args. I know that you all prefer bash though, so nevermind 😛 

Nice, but I can do one better: [`clap-rs/clap`](https://github.com/clap-rs/clap) 😆 😉

target/bin/updatemailuser
georglauterbach: As I said in the comment above, I'd always log the usage in such cases.

target/bin/sedfile
georglauterbach: I just noticed, we should not use `log` or any other helper in `sedfile` because the script is copied earlier in Dockerfile and also used early, which means that the scripts under `/usr/local/bin/helpers/` are not yet there. I will adjust accordingly.

target/bin/sedfile
georglauterbach: Done in 4560650019f5e77321a6a46f825965bee51bd462

georglauterbach:
I have added another commit (4560650019f5e77321a6a46f825965bee51bd462) to adjust `sedfile` to be canonical with the other scripts under `target/bin/`. I think further changes are not required.

polarathene (dismissed):

target/bin/setquota
polarathene: Sure if other maintainers are happy to introduce rust, some may prefer scripts over compiled binaries though.

georglauterbach:
🤦🏼 🤦🏼 🤦🏼 Nice test failure (due to not matching substring). @polarathene sorry for updating the PR again. I feel like I'm pestering you, and I promise, I'm not doing it on purpose :)

target/bin/fail2ban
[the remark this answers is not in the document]
georglauterbach: Removed it with a later commit :)

target/bin/sedfile
georglauterbach: We can't / shouldn't (for the reasons described down below) :)

polarathene (approved):

target/bin/setquota
casperklein: You are absolutely right! I initially looked at

```bash
if ! grep -qE "^${USER}\|" "${USER_DATABASE}"
then
  __usage
  _exit_with_error "User '${USER}' does not exist"
fi
```
where `__usage` is not necessary, because the correct syntax was used?

I didn't look at the error message in the other case and just concluded, that `__usage` should removed there too 😞

target/bin/updatemailuser
casperklein: Wrong assumption form my side. I agree.

target/bin/setquota
```diff
@@ -15,39 +15,33 @@ USER="${1}"
...
+[[ ${USER} =~ .*\@.* ]] || { __usage ; _exit_with_error "Username must include the domain" ; }
 
 if ! grep -qE "^${USER}\|" "${USER_DATABASE}"
 then
-  usage; _exit_with_error "user ${USER} does not exist"
+  __usage
```
casperklein, line 25:
```suggestion

```

target/bin/setquota
georglauterbach: Done in 30ccbb7109ec582e66d00c563f51b0a19f0974cc

polarathene (approved):

target/bin/delmailuser
```diff
@@ -55,26 +55,26 @@ fi
...
-      _exit_with_error "The option ${OPT} is unknown."
+      _exit_with_error "The option '${OPT}' is unknown"
       ;;
 
   esac
 done
 shift $((OPTIND-1))
 
-[[ -z ${*} ]] && { __usage ; _exit_with_error "No user specifed" ; }
+[[ -z ${*} ]] && { __usage ; _exit_with_error 'No user specifed' ; }
```
casperklein, line 77:
```suggestion
[[ -z ${*} ]] && { __usage ; _exit_with_error 'No user specified' ; }
```

target/bin/sedfile
```diff
@@ -11,19 +11,20 @@
...
 SKIP_ERROR=0
 
-if [[ $# -lt 3 ]]
+if [[ ${#} -lt 3 ]]
 then
-  echo "Error:  At least, three parameters must be given."
-  echo "Syntax: sedfile -i <replace/delete operation> <file>"
-  echo
+  __usage
+  echo 'At least three parameters must be given' >&2
```
casperklein, line 22: Not sure, why you removed "Error:"?

```suggestion
  echo 'Error: At least three parameters must be given' >&2
```

target/bin/sedfile
```diff
@@ -33,14 +34,14 @@ fi
...
 OLD=$(${HASHTOOL} "${FILE}")
-sed "$@"
+sed "${@}"
 NEW=$(${HASHTOOL} "${FILE}")
 
 # fail if file was not modified
 if [[ ${OLD} == "${NEW}" ]] && [[ ${SKIP_ERROR} -eq 0 ]]
 then
-  echo "Error: sed $*"
+  echo "No difference after call to 'sed' in 'sedfile' (sed ${*})" >&2
```
casperklein, line 43: Same here. Why was `Error:` removed? If this line is printed, a change was expected, but did not take place. Therefore an error occured.

target/bin/sedfile
```diff
@@ -11,19 +11,20 @@
 
 set -ueo pipefail
 
-HASHTOOL="sha1sum"
+function __usage { echo "Usage: ${0} -i <replace/delete operation> <file>" ; }
```
casperklein, line 14: Why introducing a function, that is only called once?

target/bin/setquota
```diff
@@ -15,39 +15,29 @@ USER="${1}"
...
-if [[ -n ${QUOTA} ]] && ! echo "${QUOTA}" | grep -qE "^([0-9]+(B|k|M|G|T)|0)\$"
+if ! grep -qE "^([0-9]+(B|k|M|G|T)|0)\$" <<< "${QUOTA}"
 then
-  usage
-  _exit_with_error "invalid quota format. e.g. 302M (B (byte), k (kilobyte), M (megabyte), G (gigabyte) or T (terabyte))"
+  __usage
+  _exit_with_error 'Invalid quota format. e.g. 302M (B (byte), k (kilobyte), M (megabyte), G (gigabyte) or T (terabyte))'
 fi
 
+_create_lock # Protect config file with lock to avoid race conditions
```
casperklein, line 39: No `_remove_lock`?

target/bin/sedfile
casperklein: PS: The whole old if statement was just fine..

target/bin/setquota
polarathene: It doesn't look like a `_remove_lock` exists in the existing script.

I think this was probably because originally `flock` was used and this wasn't required. `flock` also worked properly by locking the config file being modified I believe, whereas the replacement approach only locks the script itself from being run, thus isn't as useful (eg the API project if it were to have multiple operations queried for different users or something using `delquota` for one and `setquota` for another perhaps a chance to cause problems that `flock` would avoid.

The motivation to move from `flock` was to support NFS, but when I investigated this it seemed to be a misconfiguration on the user end as `flock` supports NFS.

---

I wouldn't expect this PR to add a `_remove_lock` as a result. Correct fix is to revert to `flock`, I have plenty of notes on the subject and it's on my todo list :sweat_smile:

target/bin/sedfile
polarathene: Probably the intention was to just be consistent with other CLI utility scripts in this folder? That's not such a bad thing even if it might be redundant in this particular file :sweat_smile:

georglauterbach: All the changes are done for uniformity with the other scripts. In https://github.com/docker-mailserver/docker-mailserver/pull/2493#discussion_r830605532 you pledged for a "continuous look and feel" - this is it (looking at all the other scripts). I have become a bit tired of never-ending small nitpicks like this to be frank. Therefore, I will leave this in (because it really is a nice addition to script uniformity, but I will revert some of the other changes to `sedfile` - otherwise I fear, these changes might never end up in `master`...

target/bin/setquota
georglauterbach: The real reason `_remove_lock` is not needed is that `_create_lock` `trap`s the `EXIT` signal to call `_remove_lock` on script exit:

https://github.com/docker-mailserver/docker-mailserver/blob/f22d75fa623675f5fa7dbb05316221dd1f02b07d/target/scripts/helpers/lock.sh#L25

target/bin/sedfile
georglauterbach: Done in [`ed512d0`](https://github.com/docker-mailserver/docker-mailserver/pull/2500/commits/ed512d04830330fda13343f49f8d7621249ac424)

target/bin/sedfile
georglauterbach: Done in [`ed512d0`](https://github.com/docker-mailserver/docker-mailserver/pull/2500/commits/ed512d04830330fda13343f49f8d7621249ac424)

target/bin/setquota
polarathene: I actually thought I had mentioned that in my comment, I guess I completely forgot about it when I went on my revert to `flock` rant :sweat_smile:

polarathene (approved):

target/bin/sedfile
casperklein: > you pledged for a "continuous look and feel" - this is it

Fair point. And with that argumentation, I agree with keeping `_usage`. For the same reason I wonder, why you decline to introduce `_log` in this PR? Technically it's no problem to copy the needed files before the first usage of `sedfile`.
Keeping the echos is clearly against a "continuous look and feel".

casperklein (approved):

target/bin/sedfile
polarathene: If you want to add `_log` to `sedfile` do it as a follow-up PR, lets just get this stuff merged while it's in a good state and approved :sweat_smile:

georglauterbach: I agree with @polarathene :) @casperklein I will add a smaller follow-up PR that introduces `_log` to `sedfile` :D

PR #2506, opened 2022-03-26
log/scripts: introduce proper log level fallback and env getter function

georglauterbach:
# Description

This PR does two small things:

1. The log level, in case it is unset, will now be "calculated" from
   `/etc/dms-settings` and not always default to `info`. This way, we
   can ensure that more often than not, the log level the user chose
   when starting DMS is used everywhere.
2. I noticed that the way I obtained the log level could be used to
   obtain any env variable's log level. I therefore added a function to
   `utils.sh` in case we use it in the future.

The reason for not using the new function from `utils.sh` in `log.sh` is that I would like to preserve `log.sh`'s property of not being dependent of other helper scripts.

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/helpers/log.sh
+++ b/target/scripts/helpers/log.sh
@@ -39,10 +39,11 @@ function _log
     return 1
   fi
 
-  local MESSAGE LEVEL_AS_INT
+  local MESSAGE LEVEL_AS_INT FALLBACK_LOG_LEVEL
   MESSAGE="${LOG_RESET}["
+  FALLBACK_LOG_LEVEL=$(grep "LOG_LEVEL" /etc/dms-settings | cut -d '=' -f 2 | tr -d "'")
 
-  case "${LOG_LEVEL:-}" in
+  case "${LOG_LEVEL:-${FALLBACK_LOG_LEVEL}}" in
     ( 'trace'  ) LEVEL_AS_INT=5 ;;
     ( 'debug'  ) LEVEL_AS_INT=4 ;;
     ( 'warn'   ) LEVEL_AS_INT=2 ;;
--- a/target/scripts/helpers/utils.sh
+++ b/target/scripts/helpers/utils.sh
@@ -11,3 +11,10 @@ function _is_comment
 {
   grep -q -E "^\s*$|^\s*#" <<< "${1}"
 }
+
+# Provide the name of an environment variable to this function
+# and it will return its value stored in /etc/dms-settings
+function _get_dms_env_value
+{
+  grep "${1}" /etc/dms-settings | cut -d '=' -f 2 | tr -d "'"
+}


georglauterbach:
Not sure whether there actually is a use case for the new function in `utils.sh`. When I just thought about it, this would only come in handy when

1. `/etc/dms-settings` is present
2. but there is a reason we cannot just `source` it

Does this even happen?

polarathene:
> Does this even happen?

`/etc/dms-settings` is fairly new, besides the recent decision to source it in `check-for-changes.sh` only a few CLI utilities source it AFAIK? Often even when we're only interested in a few select vars.

In `check-for-changes.sh` however, those values are specifically needed in helper scripts themselves. Technically those scripts could run this utility method instead as a way to "import" their global/external ENV var dependency, and perhaps log an error/warning if empty? (unless there is a fallback/default given).

That might be a better way to safe-guard against maintenance changes shuffling/splitting stuff around across files or importing specific helpers (_although some of these may also depend on methods from other helpers... they were all originally introduced as importing `index.sh` to get everything including indirectly the one you want, but I've seen you've been sourcing specific helpers directly as of late.._).

polarathene (approved):

georglauterbach:
> That might be a better way to safe-guard against maintenance changes shuffling/splitting stuff around across files or importing specific helpers (_although some of these may also depend on methods from other helpers... they were all originally introduced as importing `index.sh` to get everything including indirectly the one you want, but I've seen you've been sourcing specific helpers directly as of late.._).

Yes, we need to take care what to import and where to import. We will need to prevent a "sourcing hell" where we're not sure what is already sourced and what is not. I think I have a solution. I will provide this solution and the appropriate adjustments in a separate PR.

casperklein (commented):
Disclaimer: I am a bit in a hurry and did not read all comments here (will do this later/tomorrow).

I am not sure/understand, what exactly is tried to solve here in this PR. Can someone maybe give an example?

georglauterbach:
> I am not sure/understand, what exactly is tried to solve here in this PR. Can someone maybe give an example?

I was worried in case `LOG_LEVEL` is not set in the environment we'd always just fall back to `info`. I thought this was not desired, and therefore I'd first check if we can parse `LOG_LEVEL` from `/etc/dms-settings`.

target/scripts/helpers/utils.sh
```diff
@@ -11,3 +11,10 @@ function _is_comment
 {
   grep -q -E "^\s*$|^\s*#" <<< "${1}"
 }
+
+# Provide the name of an environment variable to this function
+# and it will return its value stored in /etc/dms-settings
+function _get_dms_env_value
+{
+  grep "${1}" /etc/dms-settings | cut -d '=' -f 2 | tr -d "'"
```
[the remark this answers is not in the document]
georglauterbach: Oh, nice catch!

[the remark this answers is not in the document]
georglauterbach: Done in [f066834](https://github.com/docker-mailserver/docker-mailserver/pull/2506/commits/f0668344379b742839ccaf64fa4f1834815164fc)

polarathene (dismissed):

target/scripts/helpers/log.sh
```diff
@@ -39,15 +39,23 @@ function _log
     return 1
   fi
 
-  local MESSAGE LEVEL_AS_INT
+  local MESSAGE LEVEL_AS_INT FALLBACK_LOG_LEVEL
```
casperklein, line 42: We have `LOG_LEVEL`, that's why I would prefer `LOG_LEVEL_FALLBACK` here instead.

```suggestion
  local MESSAGE LEVEL_AS_INT LOG_LEVEL_FALLBACK
```

casperklein (commented):
One last minor thing I noticed. Otherwise LGTM.

target/scripts/helpers/log.sh
georglauterbach: I can do that :) But I will need to adjust more than just this line. I will need to manually push a commit as I cannot just apply this suggestion :)

Done in 3e5093b3b80d65509c8a672ea253d8f5d220eded

polarathene (approved):

target/scripts/helpers/log.sh
casperklein: > But I will need to adjust more than just this line

Of course. But I was confident that you know that 👍 So I did not suggest n changes.

casperklein (approved):

PR #2197, opened 2021-09-17
fix: Don't needlessly invalidate cache layers

polarathene:
# Description

Recent `sedfile` addition moved all scripts section earlier into the Dockerfile so that `sedfile` could be used within the Dockerfile.

However whenever a change is made to scripts which is most of the time for this project, building the Docker image for tests results in all layers after the scripts being invalidated, notably ClamAV, wasting storage of previous instances and increasing build time unnecessarily.

This isn't as noticeable of an issue via the CI as we don't leverage any caching at present there, but for iterating on a local branch and testing, it can be quite the drawback.

- `sedfile` is handled early in the Dockerfile still, while the scripts have been moved as far down as it made sense to.
- `chmod` was split out into it's own RUN command as again it's unnecessary for the rest of it's prior RUN command group to be invalidated.

## Type of change

- [x] Improvement (non-breaking change that does improve existing functionality)

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] New and existing unit tests pass locally with my changes

PR #2507, opened 2022-03-26
scripts: introduce `_log` to `sedfile`

georglauterbach:
# Description

Introduce `_log` to `sedfile`, see https://github.com/docker-mailserver/docker-mailserver/pull/2500#discussion_r834805222

## Type of change

## Checklist:


target/bin/sedfile
```diff
@@ -9,6 +9,17 @@
...
 
+if [[ -e /usr/local/bin/helpers/log.sh ]]
+then
+  # shellcheck source=../scripts/helpers/log.sh
+  source /usr/local/bin/helpers/log.sh
+else
+  # when running BATS (unit tests), this file is not located
+  # inside a container; as a consequence, we need to source
+  # from a different location
+  source target/scripts/helpers/log.sh
```
polarathene, line 20: Why is this exactly? It's built in the Dockerfile, are we calling it in a test somewhere outside of the container for some reason?

polarathene (approved):

target/bin/sedfile
georglauterbach: > are we calling it in a test somewhere outside of the container for some reason?

Yes, the `sedfile.bats` test executes the script directly, not in a DMS container. I think this makes sense though and it's fine - but it also requires this "workaround".

polarathene: > I think this makes sense though and it's fine - but it also requires this "workaround".

I don't see [anything in the test](https://github.com/docker-mailserver/docker-mailserver/blob/master/test/sedfile.bats) that requires it to run a copy in the host systems `/tmp`?

I would think it more appropriate to be testing the `sedfile` within the image we've built as the test suite already assumes a docker image for pretty much everything else?

---

I'm not going to push for that, it just seems a little out of place is all.

georglauterbach: I concur, maybe @casperklein can update this in the future if he agrees as well :)

georglauterbach:
Oops, sorry @polarathene you already had approved, GitHub has let me re-request an already approved review 🤦🏼‍♀️

Dockerfile
```diff
@@ -85,6 +85,7 @@ RUN \
   rm -rf /var/lib/apt/lists/* && \
   c_rehash 2>&1
 
+COPY ./target/scripts/helpers/log.sh /usr/local/bin/helpers/log.sh
```
casperklein, line 88: I think it does not really matter, if we copy all helpers at this stage. Also, this solution would not introduce another layer to the image.

```suggestion
COPY ./target/scripts/helpers /usr/local/bin/helpers
```

target/bin/sedfile
casperklein: We all agree, this solution is ugly. I did not run the `sedfile` tests in a container, just because at this point there was no reason to do + I am an all-time fan of avoiding unnecessary overhead 😎
I will adjust the bats test in a future PR to run inside a container.

Dockerfile
polarathene: I rather we didn't... that was the whole point of making [this PR](https://github.com/docker-mailserver/docker-mailserver/pull/2197). If you're working on helpers, you have to rebuild the bulk of the image needlessly.

If we get around to a multi-stage build at some point, that'd be less of an issue. Getting all packages installed and ClamAV were both big time sinks to build time I think, plus ClamAV layer requires higher RAM (around 1-2GB IIRC) to build successfully, which usually meant paying twice as much for a VPS or configuring swap on disk. Not that the cost matters much if it's throwaway, and I've just been paying for instances I don't use monthly anyway instead of local VMs which I intend to finally get around to soon :sweat_smile:

georglauterbach: @georglauterbach exactly, you explained why I only added `log.sh` here :D 👍

casperklein: Ack.

> If we get around to a multi-stage build at some point,

I proposed that one time (squashing the image), but it was dismissed. IIRC it was because then it's not possible to inspect each layer on the final image anymore. So I do it only for my personal image build.

casperklein (approved):

Dockerfile
georglauterbach: To be honest, I cannot remember :D Maybe having a multi-stage build with a final, squashed layer at the end is a nice thing to have. If we want to do it in the future, I'll be happy to review it :)

casperklein: I think it's somewhere in the maintainers discussions. But without a search function, it's pretty hard to find it 😆

polarathene: > IIRC it was because then it's not possible to inspect each layer on the final image anymore.

I think that's not true. We can compress directives/layers into groups via multi-stage build, you can copy each one to be a separate layer, not just a single build layer with a 2nd final image copying over the difference to single layer (_although that's a common approach_).

At least I'm pretty sure that is doable :)

PR #2511, opened 2022-03-27
tests: disabled "quota exceeded" test

georglauterbach:
# Description

The "quota exceeded" test is unreliable and failed too often lately for my taste. Therefore, I'd like to disable it because there is no use in having such a test.

## Type of change

## Checklist:

### DIFF
--- a/test/tests.bats
+++ b/test/tests.bats
@@ -860,7 +860,7 @@ EOF
 }
 
 @test "checking quota: warn message received when quota exceeded" {
-  skip 'disabled as it fails randomly: https://github.com/docker-mailserver/docker-mailserver/pull/2508'
+  skip 'disabled as it fails randomly: https://github.com/docker-mailserver/docker-mailserver/pull/2511'
 
   wait_for_changes_to_be_detected_in_container mail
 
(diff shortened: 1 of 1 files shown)


polarathene (approved):

georglauterbach:
> Isn't removing a test the wrong way to deal with an unreliable test? Why is it flaky? How can we make sure it will produce the correct output repeatedly?

Sure, this is not the most optimal way, but I cannot put more time into investigating the issue at the moment. Therefore, not runner the test (i.e. skipping it) is better than having an unreliable test IMHO...

polarathene:
- We have several tests disabled presently because they're not deterministic.
- Fixing them would be ideal, but we're all short for time to try debug these due to their difficulty to reproduce consistently.
- As they're not reliable for sustained periods, we don't consider them useful to have enabled, but retain by skipping so future maintenance can try enabling in future (may resolve itself, or maintainers have more time to investigate).
- With a bit of luck, a bug report would get raised that is a real issue which the test covered but is reproducible :sweat_smile:

wernerfred (approved):
Thx for explanation

PR #2524, opened 2022-04-02
tests: enable tests that should pass now for Let'sEncrypt

georglauterbach:
# Description

Enabled tests that should, according to the comments, pass now.

## Type of change

## Checklist:

### DIFF
--- a/test/mail_ssl_letsencrypt.bats
+++ b/test/mail_ssl_letsencrypt.bats
@@ -156,12 +156,10 @@ function teardown() {
     _should_extract_on_changes 'example.test' "${LOCAL_BASE_PATH}/wildcard/rsa.acme.json"
     _should_have_service_restart_count '2'
 
-    # note: https://github.com/docker-mailserver/docker-mailserver/pull/2404 solves this
-    # TODO: Make this pass.
     # As the FQDN has changed since startup, the configs need to be updated accordingly.
     # This requires the `changedetector` service event to invoke the same function for TLS configuration
     # that is used during container startup to work correctly. A follow up PR will refactor `setup-stack.sh` for supporting this.
-    # _should_have_valid_config 'example.test' 'key.pem' 'fullchain.pem'
+    _should_have_valid_config 'example.test' 'key.pem' 'fullchain.pem'
 
     local WILDCARD_KEY_PATH="${LOCAL_BASE_PATH}/wildcard/key.rsa.pem"
     local WILDCARD_CERT_PATH="${LOCAL_BASE_PATH}/wildcard/cert.rsa.pem"
@@ -169,8 +167,7 @@ function teardown() {
 
     # Verify this works for wildcard certs, it should use `*.example.test` for `mail.example.test` (NOT `example.test`):
     _should_succesfully_negotiate_tls 'mail.example.test'
-    # WARNING: This should fail...but requires resolving the above TODO.
-    # _should_not_have_fqdn_in_cert 'example.test'
+    _should_not_have_fqdn_in_cert 'example.test'
   }
 
   _prepare


georglauterbach:
@polarathene any idea why these tests do not pass? According to the TODO notes they should...

polarathene:
> @polarathene any idea why these tests do not pass? According to the TODO notes they should...

```
not ok 92 ssl(letsencrypt): Traefik 'acme.json' (*.example.test) in 40796ms
#  from function `_has_matching_line' in file test/mail_ssl_letsencrypt.bats, line 204,
#  from function `_should_have_valid_config' in file test/mail_ssl_letsencrypt.bats, line 195,
#  from function `_acme_wildcard' in file test/mail_ssl_letsencrypt.bats, line 162,
#  in test file test/mail_ssl_letsencrypt.bats, line 181)
#   `_acme_wildcard' failed
# [  DEBUG  ]  2022-04-02 18:51:56  Chagedetector is ready
# 
# -- output differs --
# expected : smtpd_tls_chain_files = /etc/letsencrypt/live/example.test/key.pem /etc/letsencrypt/live/example.test/fullchain.pem
# actual   : 
# --
# 
```

I'd take a wild guess it's a race condition? This test passes for other test cases, and for this one to run, the extraction must have been successful with the correct restart count detected.

I think the failure is due to you only adding the required `_setup_ssl` call for manual certs:

https://github.com/docker-mailserver/docker-mailserver/blob/b93d9708d03cb279c6dd411624165f38197a4f43/target/scripts/check-for-changes.sh#L78-L81

You would add it here too:

https://github.com/docker-mailserver/docker-mailserver/blob/b93d9708d03cb279c6dd411624165f38197a4f43/target/scripts/check-for-changes.sh#L100-L104

And then we can consider the following:

---

### Option A

Technically all we're interested in is updating the config files (_but the method called is private to `_setup_ssl`? Would have to be extracted out above it in `helpers/ssl.sh`? Note the local config vars referenced though_):

https://github.com/docker-mailserver/docker-mailserver/blob/b93d9708d03cb279c6dd411624165f38197a4f43/target/scripts/helpers/ssl.sh#L228-L231

Which might be better to do as `_setup_ssl` will otherwise perform extraction internally redundantly for `acme.json` here:

https://github.com/docker-mailserver/docker-mailserver/blob/b93d9708d03cb279c6dd411624165f38197a4f43/target/scripts/helpers/ssl.sh#L186

### Option B

Alternatively, this entire block can be refactored (_or the equivalent `_traefik_support` method for `_setup_ssl`_):

https://github.com/docker-mailserver/docker-mailserver/blob/35fb744ffb503e6f4e46d11d214d9100f6c17ca8/target/scripts/check-for-changes.sh#L86-L112

The logic is effectively the same as we have here in `_setup_ssl`:

https://github.com/docker-mailserver/docker-mailserver/blob/b93d9708d03cb279c6dd411624165f38197a4f43/target/scripts/helpers/ssl.sh#L99-L124

Except, that doesn't return/set the actual FQDN used as this logic is specific to `check-for-changes.sh`:

https://github.com/docker-mailserver/docker-mailserver/blob/35fb744ffb503e6f4e46d11d214d9100f6c17ca8/target/scripts/check-for-changes.sh#L102-L107

Technically if we extract these lines into two methods for getting the two LetsEncrypt local vars used, then `check-for-changes.sh` could call that too after `_setup_ssl` which will handle the extraction internally instead of `check-for-changes.sh` doing effectively the same?:

https://github.com/docker-mailserver/docker-mailserver/blob/b93d9708d03cb279c6dd411624165f38197a4f43/target/scripts/helpers/ssl.sh#L188-L219

---

I think I could open a PR for that with option B approach, once that's merged this PR should pass :)

polarathene:
[Option B implemented](https://github.com/docker-mailserver/docker-mailserver/pull/2525).

georglauterbach:
Changes implemented in #2525.

PR #2525, opened 2022-04-03
refactor: letsencrypt implicit location discovery

polarathene:
# Description

Small refactor to better support `check-for-changes.sh` functionality.

These changes should allow some additional test coverage to be enabled. This implements _Option B_ I described [in comment for a PR](https://github.com/docker-mailserver/docker-mailserver/pull/2524#issuecomment-1086750378) trying to enable that additional coverage.

- Changes to `ssl.sh` is minimal, effectively lifting the logic out of `_setup_ssl` to separate methods so that `check-for-changes.sh` can use one.
- `check-for-changes.sh` is primarily simplifying the logic by dropping the loop and inner conditional statement (_so some lines only have indentation changes_), calling `_setup_ssl` and `_find_letsencrypt_domain` instead. A separate commit improves relevant maintainer comments.
- Unnecessary to this PR is wrapping the change detection into a function that the loop calls instead of inlined. I did this to use `local` annotation.

<details>
<summary>All commit messages for easy reference</summary>

> **chore: Extract letsencrypt logic into methods**
> This allows other scripts to share the functionality to discover the correct letsencrypt folder from the 3 possible locations (where specific order is important).
> 
> As these methods should now return a string value, the `return 1` after a panic is now dropped.
> 
> ~~The `_find_letsencrypt_key` method can take a domain as an arg, otherwise it will retrieve the internal domain the same way as a fallback, which is slightly inefficient to repeat but should be a non-issue.~~ (No longer the case)

> **chore: Update comments**
> The todo is resolved with this PR; `_setup_ssl` will be called by both cert conditional statements, the purpose for each is better documented for maintainers.

> **refactor: Defer most logic to helper/ssl.sh**
> The loop is no longer required, extraction is delegated to `_setup_ssl` now.
> 
> For the change event prevention, we retrieve the relevant FQDN via the new helper method, beyond that it's just indentation diff.

> **chore: Appease the lint gods**
> The lint warning for the unused optional parameter could alternatively be resolved by prepending `#shellcheck disable=SC2120` to the function definition.
> As there are no other consumers, I changed to a required arg with a hard failure instead of providing a fallback.
> 
> `check-for-changes.sh` adjusted to allow locally scoped var declarations by wrapping a function. Presently no loop control flow is needed so this seems fine. Made it clear that `CHANGED` is local and `CHKSUM_FILE` is not.
> 
> The panic scope arg doesn't need `SSL_TYPE` for added context, it's clearly`letsencrypt`, removed.

</details>

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/check-for-changes.sh
+++ b/target/scripts/check-for-changes.sh
@@ -3,8 +3,10 @@
 # TODO: Adapt for compatibility with LDAP
 # Only the cert renewal change detection may be relevant for LDAP?
 
+# CHKSUM_FILE global is imported from this file:
 # shellcheck source=./helpers/index.sh
 source /usr/local/bin/helpers/index.sh
+
 # This script requires some environment variables to be properly set. This
 # includes POSTMASTER_ADDRESS (for alias (re-)generation), HOSTNAME and
 # DOMAINNAME (in ssl.sh).
@@ -46,8 +48,8 @@ sleep 10
 
 _log_with_date 'debug' "Chagedetector is ready"
 
-while true
-do
+function _check_for_changes
+{
   # get chksum and check it, no need to lock config yet
   _monitored_files_checksums >"${CHKSUM_FILE}.new"
   cmp --silent -- "${CHKSUM_FILE}" "${CHKSUM_FILE}.new"
@@ -60,6 +62,7 @@ do
   then
     _log_with_date 'info' 'Change detected'
     _create_lock # Shared config safety lock
+    local CHANGED
     CHANGED=$(grep -Fxvf "${CHKSUM_FILE}" "${CHKSUM_FILE}.new" | sed 's/^[^ ]\+  //')
 
     # TODO Perform updates below conditionally too
@@ -134,6 +137,11 @@ do
   mv "${CHKSUM_FILE}.new" "${CHKSUM_FILE}"
 
   sleep 2
+}
+
+while true
+do
+  _check_for_changes
 done
 
 exit 0
--- a/target/scripts/check-for-changes.sh
+++ b/target/scripts/check-for-changes.sh
@@ -87,29 +87,15 @@ do
     elif [[ ${CHANGED} =~ /etc/letsencrypt/acme.json ]]
     then
       _log_with_date 'debug' "'/etc/letsencrypt/acme.json' has changed - extracting certificates"
+      _setup_ssl
 
-      # This breaks early as we only need the first successful extraction.
-      # For more details see the `SSL_TYPE=letsencrypt` case handling in `setup-stack.sh`.
-      #
-      # NOTE: HOSTNAME is set via `helpers/dns.sh`, it is not the original system HOSTNAME ENV anymore.
-      # TODO: SSL_DOMAIN is Traefik specific, it no longer seems relevant and should be considered for removal.
-      FQDN_LIST=("${SSL_DOMAIN}" "${HOSTNAME}" "${DOMAINNAME}")
-      for CERT_DOMAIN in "${FQDN_LIST[@]}"
-      do
-        _log_with_date 'trace' "Attempting to extract for '${CERT_DOMAIN}'"
+      # Prevent an unnecessary change detection from the newly extracted cert files by updating their hashes in advance:
+      local CERT_DOMAIN
+      CERT_DOMAIN="$(_find_letsencrypt_domain)"
+      ACME_CERT_DIR="/etc/letsencrypt/live/${CERT_DOMAIN}"
 
-        if _extract_certs_from_acme "${CERT_DOMAIN}"
-        then
-          # Prevent an unnecessary change detection from the newly extracted cert files by updating their hashes in advance:
-          CERT_DOMAIN=$(_strip_wildcard_prefix "${CERT_DOMAIN}")
-          ACME_CERT_DIR="/etc/letsencrypt/live/${CERT_DOMAIN}"
-
-          sed -i "\|${ACME_CERT_DIR}|d" "${CHKSUM_FILE}.new"
-          sha512sum "${ACME_CERT_DIR}"/*.pem >> "${CHKSUM_FILE}.new"
-
-          break
-        fi
-      done
+      sed -i "\|${ACME_CERT_DIR}|d" "${CHKSUM_FILE}.new"
+      sha512sum "${ACME_CERT_DIR}"/*.pem >> "${CHKSUM_FILE}.new"
     fi
 
     # If monitored certificate files in /etc/letsencrypt/live have changed and no `acme.json` is in use,
--- a/target/scripts/check-for-changes.sh
+++ b/target/scripts/check-for-changes.sh
@@ -66,6 +66,10 @@ do
     # Also note that changes are performed in place and are not atomic
     # We should fix that and write to temporary files, stop, swap and start
 
+    # _setup_ssl is required for:
+    # manual - copy to internal DMS_TLS_PATH (/etc/dms/tls) that Postfix and Dovecot are configured to use.
+    # acme.json - presently uses /etc/letsencrypt/live/<FQDN> instead of DMS_TLS_PATH,
+    # path may change requiring Postfix/Dovecot config update.
     if [[ ${SSL_TYPE} == 'manual' ]]
     then
       # only run the SSL setup again if certificates have really changed.
@@ -75,9 +79,6 @@ do
       || [[ ${CHANGED} =~ ${SSL_ALT_KEY_PATH:-${REGEX_NEVER_MATCH}} ]]
       then
         _log_with_date 'debug' 'Manual certificates have changed - extracting certificates'
-        # we need to run the SSL setup again, because the
-        # certificates DMS is working with are copies of
-        # the (now changed) files
         _setup_ssl
       fi
     # `acme.json` is only relevant to Traefik, and is where it stores the certificates it manages.
@@ -113,7 +114,6 @@ do
 
     # If monitored certificate files in /etc/letsencrypt/live have changed and no `acme.json` is in use,
     # They presently have no special handling other than to trigger a change that will restart Postfix/Dovecot.
-    # TODO: That should be all that's required, unless the cert file paths have also changed (Postfix/Dovecot configs then need to be updated).
 
     # regenerate postfix accounts
     [[ ${SMTP_ONLY} -ne 1 ]] && _create_accounts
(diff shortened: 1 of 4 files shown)


target/scripts/helpers/ssl.sh
```diff
@@ -185,38 +185,10 @@ function _setup_ssl
...
-        LETSENCRYPT_KEY='key'
-      else
-        _log 'warn' "Cannot find key file ('privkey.pem' or 'key.pem') in '/etc/letsencrypt/live/${LETSENCRYPT_DOMAIN}/'"
-        dms_panic__misconfigured 'LETSENCRYPT_KEY' "${SCOPE_SSL_TYPE}"
-        return 1
-      fi
+      # checks folders in /etc/letsencrypt/live to identify which one to implicitly use:
+      local LETSENCRYPT_DOMAIN LETSENCRYPT_KEY
+      LETSENCRYPT_DOMAIN="$(_find_letsencrypt_domain)"
+      LETSENCRYPT_KEY="$(_find_letsencrypt_key)"
```
polarathene, line 191: This will redundantly run the same domain finding logic internally again.

Not likely to be a problem, but I can add the domain result explicitly as an arg here instead if desired?

target/scripts/check-for-changes.sh
```diff
@@ -148,6 +137,11 @@ do
   mv "${CHKSUM_FILE}.new" "${CHKSUM_FILE}"
 
   sleep 2
+}
+
+while true
+do
+  _check_for_changes
```
georglauterbach, line 144: I'd personally add the `sleep 2` here and not at the end of the function.

target/scripts/helpers/ssl.sh
```diff
@@ -408,6 +380,54 @@ function _setup_ssl
...
+  then
+    LETSENCRYPT_DOMAIN=$(_strip_wildcard_prefix "${SSL_DOMAIN}")
+  elif [[ -e /etc/letsencrypt/live/${HOSTNAME}/fullchain.pem ]]
+  then
+    LETSENCRYPT_DOMAIN=${HOSTNAME}
+  elif [[ -e /etc/letsencrypt/live/${DOMAINNAME}/fullchain.pem ]]
+  then
+    LETSENCRYPT_DOMAIN=${DOMAINNAME}
+  else
+    _log 'warn' "Cannot find a valid DOMAIN for '/etc/letsencrypt/live/<DOMAIN>/', tried: '${SSL_DOMAIN}', '${HOSTNAME}', '${DOMAINNAME}'"
```
georglauterbach, line 399: Because we panic afterwards, this should be `_log 'error'`.

``` SUGGESTION
    _log 'error' "Cannot find a valid DOMAIN for '/etc/letsencrypt/live/<DOMAIN>/', tried: '${SSL_DOMAIN}', '${HOSTNAME}', '${DOMAINNAME}'"
```

target/scripts/helpers/ssl.sh
```diff
@@ -408,6 +380,54 @@ function _setup_ssl
...
+    dms_panic__misconfigured 'LETSENCRYPT_DOMAIN' '_find_letsencrypt_domain'
+  fi
+
+  return "${LETSENCRYPT_DOMAIN}"
```
georglauterbach, line 403: I am not sure what is happening here 😆 I know exactly what you wanted to do though:

``` SUGGESTION
  printf '%s' "${LETSENCRYPT_DOMAIN}"
```

target/scripts/helpers/ssl.sh
```diff
@@ -408,6 +380,54 @@ function _setup_ssl
...
+    LETSENCRYPT_KEY='privkey'
+  elif [[ -e /etc/letsencrypt/live/${LETSENCRYPT_DOMAIN}/key.pem ]]
+  then
+    LETSENCRYPT_KEY='key'
+  else
+    _log 'warn' "Cannot find key file ('privkey.pem' or 'key.pem') in '/etc/letsencrypt/live/${LETSENCRYPT_DOMAIN}/'"
+    dms_panic__misconfigured 'LETSENCRYPT_KEY' '_find_letsencrypt_key'
+  fi
+
+  return "${LETSENCRYPT_KEY}"
```
georglauterbach, line 428: Same here:

``` SUGGESTION
  printf '%s' "${LETSENCRYPT_KEY}"
```

target/scripts/helpers/ssl.sh
```diff
@@ -408,6 +380,54 @@ function _setup_ssl
...
+  fi
+
+  if [[ -e /etc/letsencrypt/live/${LETSENCRYPT_DOMAIN}/privkey.pem ]]
+  then
+    LETSENCRYPT_KEY='privkey'
+  elif [[ -e /etc/letsencrypt/live/${LETSENCRYPT_DOMAIN}/key.pem ]]
+  then
+    LETSENCRYPT_KEY='key'
+  else
+    _log 'warn' "Cannot find key file ('privkey.pem' or 'key.pem') in '/etc/letsencrypt/live/${LETSENCRYPT_DOMAIN}/'"
```
georglauterbach, line 424: This should be an error too, because we panic afterwards (I just now noticed how much the Rust-slang is omnipresent with what I write 😆)

``` SUGGESTION
    _log 'error' "Cannot find key file ('privkey.pem' or 'key.pem') in '/etc/letsencrypt/live/${LETSENCRYPT_DOMAIN}/'"
```

target/scripts/check-for-changes.sh
```diff
@@ -148,6 +137,11 @@ do
   mv "${CHKSUM_FILE}.new" "${CHKSUM_FILE}"
 
   sleep 2
```
georglauterbach, line 139:
``` SUGGESTION
```

target/scripts/helpers/ssl.sh
casperklein: Yeah. `return` is only for exit codes 👍

Any reason why you suggest `printf '%s'` over `echo` here?

georglauterbach: I'm always afraid of `echo`'s newline at the end. This has caused me tremendous trouble in the past (with Kubernetes secrets that contained a newline when they shouldn't), so I generally prefer `printf`. I guess there is not much difference in this case, and we could as well use `echo -n` - I don't really care, as long as the newline can be avoided :)

casperklein: Get it, but I don't think there is a need to avoid the newline. Bash handles this just fine, afaik. This is done in a lot of places already in DMS.

georglauterbach: Yes, I'm just used to `printf`, that's why I proposed it here as well. If we want to use `echo` here, that's absolutely fine with me.

casperklein: I had a look at the remaining file and it uses `echo`'s too. Otherwise I would be fine with `printf` too.

georglauterbach: > I had a look at the remaining file and it uses `echo`'s too. Otherwise I would be fine with `printf` too.

Then I will adjust my suggestions.

polarathene: > I know exactly what you wanted to do though

Whoops, I haven't written shell scripts for a while :laughing:

polarathene:
Failure (**EDIT:** Resolved):

```
not ok 92 ssl(letsencrypt): Traefik 'acme.json' (*.example.test) in 41489ms
# (from function `assert_output' in file test/test_helper/bats-assert/src/assert.bash, line 223,
#  from function `_should_have_fqdn_in_cert' in file test/test_helper/tls.bash, line 95,
#  from function `_negotiate_tls' in file test/test_helper/tls.bash, line 55,
#  from function `_should_succesfully_negotiate_tls' in file test/test_helper/tls.bash, line 30,
#  from function `_acme_wildcard' in file test/mail_ssl_letsencrypt.bats, line 169,
#  in test file test/mail_ssl_letsencrypt.bats, line 182)
#   `_acme_wildcard' failed
# [  DEBUG  ]  2022-04-03 22:49:05  Chagedetector is ready
# 
# -- regular expression does not match output --
# regexp (1 lines):
#   Subject: CN = mail\.example\.test|DNS:mail\.example\.test
# output (2 lines):
#           Subject: CN = Smallstep Leaf
#                   DNS:*.example.test
# --
# 
```

<details>
<summary>Manual reproduction steps for inspection</summary>

```bash
git clone https://github.com/polarathene/docker-mailserver --branch refactor/letsencrypt-setup --recurse-submodules
cd docker-mailserver
make build
make generate-accounts

./test/bats/bin/bats --timing test/mail_ssl_letsencrypt.bats

# Manually reproduce:
export TEST_TMP_CONFIG="${PWD}/test/duplicate_configs/x_dmstest"
mkdir -p "${TEST_TMP_CONFIG}"
rsync -a test/config/ "${TEST_TMP_CONFIG}"

cp "${PWD}/test/test-files/ssl/example.test/with_ca/rsa/ecdsa.acme.json" "${TEST_TMP_CONFIG}/letsencrypt/acme.json"

docker run --rm -itd --name "dms-test" \
  --hostname 'mail.example.test' \
  --volume "${PWD}/test/test-files:/tmp/docker-mailserver-test:ro" \
  --volume "${TEST_TMP_CONFIG}:/tmp/docker-mailserver" \
  --volume "${TEST_TMP_CONFIG}/letsencrypt/acme.json:/etc/letsencrypt/acme.json:ro"  \
  --env SSL_TYPE='letsencrypt'  \
  --env SSL_DOMAIN='*.example.test' \
  --env LOG_LEVEL='trace' \
  --env PERMIT_DOCKER='container' \
  "mailserver-testing:ci"

cp "${PWD}/test/test-files/ssl/example.test/with_ca/rsa/wildcard/rsa.acme.json" "${TEST_TMP_CONFIG}/letsencrypt/acme.json"

docker exec dms-test sh -c "timeout 1 openssl s_client -connect localhost:587 -starttls smtp -CAfile /tmp/docker-mailserver-test/ssl/example.test/with_ca/rsa/ca-cert.rsa.pem -servername mail.example.test 2>/dev/null | openssl x509 -noout -text | grep -E 'Subject: CN = |DNS:'"
```

</details>

---

Occurred due to this PR actually updating the Postfix and Dovecot configs and missing functionality for considering a valid match for a explicit FQDN to a wildcard cert.

polarathene:
I have enabled the related tests in this PR but can revert that if preferred?

I did so due to renaming the helper method and reworking the comments.

georglauterbach:
> I have enabled the related tests in this PR but can revert that if preferred?
> 
> I did so due to renaming the helper method and reworking the comments.

That's not a problem. I will just close #2524.

georglauterbach (approved):
LGTM

casperklein (approved):

polarathene:
@georglauterbach as the v11 release PR has been open for testing, do you still want this merged for v11 or for me to wait until after that release?

georglauterbach:
#2537 is currently only WIP. I would very much like to see this PR go into `v11`. Therefore, I will update the branch and you can then merge it straight away :D

PR #2530, opened 2022-04-04
scripts: added `TZ` environment variable to set timezone

georglauterbach:
# Description

We currently set the time zone by mounting `/etc/localtime`. This is fine for Docker, and users will still be able to set their time like that. This PR introduces a new way of setting date and time, that one users can use alternatively. This is especially nice when mounting from the host is difficult. Moreover, I think this is nicer, more elegant and cleaner than mounting `/etc/localtime`. If you prefer mounting `/etc/localtime` though, you can still do that.

## Type of change

## Checklist:


target/scripts/startup/setup-stack.sh
```diff
@@ -1263,3 +1263,22 @@ EOF
...
+
+function _setup_timezone
+{
+  _log 'debug' "Setting timezone to '${TZ}'"
+
+  export DEBCONF_NONINTERACTIVE_SEEN=true
+  local AREA ZONE DPKG_RECONFIGURE_PRESEED
+
+  AREA=$(cut -d '/' -f 1 <<< "${TZ}")
+  ZONE=$(cut -d '/' -f 2 <<< "${TZ}")
```
wernerfred, line 1275:
```suggestion
  AREA=${TZ%%/*}
  ZONE=${TZ##*/}
```

maybe nicer to use pattern matching in parameter substitution then spawning subshell

georglauterbach: I like it :D

wernerfred (dismissed):
Not familiar with debconf tzdata command but the rest seems logical to me 👍🏻

georglauterbach:
> Not familiar with debconf tzdata command but the rest seems logical to me 👍🏻 

Trust me, I played around with this with the help of `user-patches.sh`. The `denconf` thingy is the most reliable way to set the system time :D

casperklein:
> The `debconf` thingy is the most reliable way to set the system time in Debian containers :D

Did you try something simple like this?

```bash
function _setup_timezone
{
  _log 'info' "Setting timezone to '${TZ}'"

  ln -fs "/usr/share/zoneinfo/${TZ}" /etc/localtime &&
  dpkg-reconfigure -f noninteractive tzdata &>/dev/null
}
```

If so, what problems did you encounter? That's similar to what I use and never had problems with

PS: Your test ran successful with this one too.

test/mail_time.bats
```diff
@@ -0,0 +1,29 @@
...
+teardown_file() {
+    docker rm -f mail_time
+}
+
+@test "checking time: setting the time with TZ works correctly" {
+  run docker exec mail_time cat /etc/timezone
+  assert_success
+  assert_output 'Asia/Jakarta'
+
+  run docker exec mail_time date +'%Z'
```
casperklein, line 26:
```suggestion
  run docker exec mail_time date '+%Z'
```

target/scripts/startup/setup-stack.sh
polarathene: What is this magic :mage_man: :monocle_face:

How you all remember all this syntax eludes me :joy:

georglauterbach: > What is this magic mage_man monocle_face

Parameter substitution is Bash's formidable way of adjusting the value than a variable contains. What @wernerfred proposed here is nice, because with the first line, we get everything _before_ the `/` character, and cut off everything behind it. The second line does the opposite, cutting off everything before the `/`, and keeping everything _after_ it. 

> How you all remember all this syntax eludes me joy

My go-to: <https://devhints.io/bash>

wernerfred (approved):

wernerfred:
> If so, what problems did you encounter? That's similar to what I use and never had problems with.\n\nPS: Your test ran successful with this one too.

Does something speaks against this implementation? Easier to follow along for me (who doesnt have a clue about tzdata). Just curious.

georglauterbach:
> > The `debconf` thingy is the most reliable way to set the system time in Debian containers :D
> 
> Did you try something simple like this?
> 
> ```shell
> function _setup_timezone
> {
>   _log 'info' "Setting timezone to '${TZ}'"
> 
>   ln -fs "/usr/share/zoneinfo/${TZ}" /etc/localtime &&
>   dpkg-reconfigure -f noninteractive tzdata &>/dev/null
> }
> ```

It's been a long time, but I think I tried a few ways.

> If so, what problems did you encounter? That's similar to what I use and never had problems with.
> 
> PS: Your test ran successful with this one too.

I have to admit, I do not remember (if there even were any).

> Does something speaks against this implementation? Easier to follow along for me (who doesnt have a clue about tzdata). Just curious.

Nothing, really. I got to be honest though, I'm not sure about this because I vividly remember that some setups caused issues. BUT since tests seem to pass though as @casperklein reported, I will change it because it indeed looks easier :D

georglauterbach:
Not sure what happened there with [88a1d5b](https://github.com/docker-mailserver/docker-mailserver/pull/2530/commits/88a1d5b21919f1a52809c270f6bac21e8e9d8230), but this is probably on me because I tried to push before I pulled (again...).

Dockerfile
```diff
@@ -21,6 +20,9 @@ LABEL org.opencontainers.image.url="https://github.com/docker-mailserver"
 LABEL org.opencontainers.image.documentation="https://github.com/docker-mailserver/docker-mailserver/blob/master/README.md"
 LABEL org.opencontainers.image.source="https://github.com/docker-mailserver/docker-mailserver"
 
+ENV DEBIAN_FRONTEND=noninteractive
```
wernerfred, line 23: why did you switch from ARG to ENV

georglauterbach: ARG is only visible during build time, while ENV will "export" the variable and its value in the running image as well. This is desired because we want to have the vars available during setup (and beyond).

wernerfred: Yeah i know the difference. Just wondering why your tz change now has the requirement of setting this as an ENV. Is it dpkg-reconfigure that needs that value? I only know this in combination with apt (which we only need during build time)

casperklein: > This is desired because we want to have the vars available during setup (and beyond).

I cannot think of a reason why we would want to have this var available in the image. BTW we wanted to get rid of almost all ENVs in the Dockerfile. I still guess most of them don't serve any purpose (anymore?). I've that on my to-do.

georglauterbach: Both of these variables ensure that while building the image there is no interaction required, and the same goes for `dpkg-reconfigure` during setup. I'd leave these two variables in the Dockerfile (as the only ones actually needed), otherwise we may run into problems. If we do want to remove them, I'd need to export the second variable again during debconf and also enable the CLI flags. I'd vote: if you two think it's better to remove them from the Dockerfile, I will do so, otherwise I'd leave it as is now.

casperklein: > Both of these variables ensure that while building the image there is no interaction required

Yes. We were just curious, why you changed from ARG to ENV. Having this var while building is fine (as ARG). But there is no general need to have it in the final image (ENV).
And if apt/dpkg-reconfigure are used, IMO it's best practice, to call these programs with explicit options like `dpkg-reconfigure -f noninteractive`.

georglauterbach: Alright, then I will take care of it.

Dockerfile
```diff
@@ -3,6 +3,7 @@ FROM docker.io/debian:11-slim
 ARG VCS_VER
 ARG VCS_REF
 ARG DEBIAN_FRONTEND=noninteractive
+ARG DEBCONF_NONINTERACTIVE_SEEN=true
```
wernerfred, line 6: So this ARG is not related to the TZ change (this PR) right? 
Otherwise it is not needed as ARG as TZ config is only done during runtime.

If it is useful for other tasks during build time we can keep it, otherwise i would remove it again

wernerfred (dismissed):

Dockerfile
georglauterbach: Found it [here](https://wiki.debian.org/Multistrap/Environment), not 100% sure though whether we absolutely need it. I guess for the current build, we don't...

casperklein: Not needed, don't keep it? 🤷

casperklein (approved):

Dockerfile
georglauterbach: Ahhh, yes... :D

georglauterbach: Done in 4d261e2ec2eb841125be99c0cf14ad30edbca22c

casperklein (approved):

wernerfred (approved):

PR #2538, opened 2022-04-07
Add ban feature to fail2ban script

casperklein:
# Description

This PR adds a new "ban" option to the [fail2ban](https://github.com/docker-mailserver/docker-mailserver/blob/master/target/bin/fail2ban) script. This allows to easily block custom IPs with fail2ban:

`./setup.sh failban ban <IP>`

The default bantime is 180 days, but can be adjusted easily like other fail2ban settings.

`setup.sh`: fail2ban was moved from the "debug" section to it's own fail2ban section:

`setup.sh debug fail2ban` --> `setup.sh fail2ban`

While adding the necessary changes to `config-examples/fail2ban-jail.cf`, I've also synced all values there, with our default settings.

A new fail2ban test was added and the existing ones were improved.

## Type of change

## Checklist:

### DIFF
--- a/config-examples/fail2ban-jail.cf
+++ b/config-examples/fail2ban-jail.cf
@@ -34,3 +34,4 @@ enabled = true
 [custom]
 enabled = true
 bantime = 180d
+port = smtp,pop3,pop3s,imap,imaps,submission,465,sieve
--- a/config-examples/fail2ban-jail.cf
+++ b/config-examples/fail2ban-jail.cf
@@ -1,21 +1,33 @@
 [DEFAULT]
 
 # "bantime" is the number of seconds that a host is banned.
-#bantime  = 10m
+bantime = 3h
 
 # A host is banned if it has generated "maxretry" during the last "findtime"
 # seconds.
-#findtime  = 10m
+findtime = 10m
 
 # "maxretry" is the number of failures before a host get banned.
-#maxretry = 5
+maxretry = 3
 
 # "ignoreip" can be a list of IP addresses, CIDR masks or DNS hosts. Fail2ban
 # will not ban a host which matches an address in this list. Several addresses
 # can be defined using space (and/or comma) separator.
-#ignoreip = 127.0.0.1/8
+ignoreip = 127.0.0.1/8
 
-# Default ban action
-# nftables-multiport:	block IP only on affected port
-# nftables-allports:	block IP on all ports
-#banaction = nftables-allports
+# default ban action
+# nftables-multiport: block IP only on affected port
+# nftables-allports:  block IP on all ports
+banaction = nftables-allports
+
+[dovecot]
+enabled = true
+
+[postfix]
+enabled = true
+
+[postfix-sasl]
+enabled = true
+
+[custom]
+enabled = true
--- a/config-examples/fail2ban-jail.cf
+++ b/config-examples/fail2ban-jail.cf
@@ -31,3 +31,4 @@ enabled = true
 
 [custom]
 enabled = true
+bantime = 180d
--- a/config-examples/fail2ban-jail.cf
+++ b/config-examples/fail2ban-jail.cf
@@ -34,4 +34,4 @@ enabled = true
 [custom]
 enabled = true
 bantime = 180d
-port = smtp,pop3,pop3s,imap,imaps,submission,465,sieve
+port = smtp,pop3,pop3s,imap,imaps,submission,submissions,sieve
--- a/config-examples/fail2ban-jail.cf
+++ b/config-examples/fail2ban-jail.cf
@@ -29,6 +29,8 @@ enabled = true
 [postfix-sasl]
 enabled = true
 
+# This jail is used for manual bans.
+# To ban an IP address use: setup.sh fail2ban ban <IP>
 [custom]
 enabled = true
 bantime = 180d
--- a/target/bin/fail2ban
+++ b/target/bin/fail2ban
@@ -3,7 +3,11 @@
 # shellcheck source=../scripts/helpers/index.sh
 source /usr/local/bin/helpers/index.sh
 
-function __usage { echo "Usage: ${0} [<unban> <ip-address|network|hostname>]" ; }
+function __usage {
+  echo "Usage: ${0}" ;
+  echo "       ${0} ban <IP>" ;
+  echo "       ${0} unban <IP>" ;
+}
 
 unset JAILS
 declare -a JAILS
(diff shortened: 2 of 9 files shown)


georglauterbach:
I assigned `v11` to this PR, but if you think this should rather go into `v11.1`, please change it :)

target/bin/fail2ban
```diff
@@ -3,7 +3,11 @@
 # shellcheck source=../scripts/helpers/index.sh
 source /usr/local/bin/helpers/index.sh
 
-function __usage { echo "Usage: ${0} [<unban> <ip-address>]" ; }
+function __usage {
+  echo "Usage: ${0}" ;
+  echo "       ${0} ban <IP>" ;
+  echo "       ${0} unban <IP>" ;
```
georglauterbach, line 9: Why the `;` at the end? Seems unnecessary to me.

target/bin/fail2ban
```diff
@@ -3,7 +3,11 @@
 # shellcheck source=../scripts/helpers/index.sh
 source /usr/local/bin/helpers/index.sh
 
-function __usage { echo "Usage: ${0} [<unban> <ip-address>]" ; }
+function __usage {
```
georglauterbach, line 6: This `__usage` definition does not really look like what, lets say `addmailuser` looks like. I think `addmailuser` looks nicer. Having an `echo` on each new line seems kind of awkward to me.

georglauterbach (commented):
Looks good all in all. I'd just like to discuss the `__usage´ section.

target/bin/fail2ban
casperklein: full ack. copy/paste error.

target/bin/fail2ban
casperklein: Personally I think, more echo's are looking much more better, than one echo/printf across multiple lines. But no need to discuss this further 😆 

I don't think we need a full blown `__usage` function like in addmailuser. However, you are right, to streamline this I would go for:

`function __usage { echo "Usage: ${0} [<ban|unban> <ip-address>]" ; }`

Beside the question of using a simple/full-blown `__usage` function, I just noticed this fact: `__usage` is never called 😉 If no argument is supplied, the script will list all banned IPs. In comparison, other scripts are calling `__usage`, if no parameter is supplied.

Some options to choose from:

1. Remove `__usage` function. But then this script is not 100% like the others.
2. Keep `__usage` function with dummy code: `function __usage() { :; }`
3. Change the script, so that listing banned IPs needs an explicit argument like `show` or `list` --> `./setup.sh fail2ban list`. Then usage could be displayed, if no parameter is given. This would be a breaking change.
4. `__usage` will only be called, when `help` parameter is given.

I tend to option 4 at the moment or maybe 3.

BTW: I think the options ban/unban are so self-explanatory, that I cannot imagine, someone will every try to see the usage explanation.

georglauterbach: > I tend to option 4 at the moment or maybe 3.

I'd go for option 4 as well.

> BTW: I think the options ban/unban are so self-explanatory, that I cannot imagine, someone will every need the usage explanation.

Me too 👍🏼

georglauterbach (approved):
LGTM

config-examples/fail2ban-jail.cf
```diff
@@ -1,21 +1,37 @@
...
+
+[postfix-sasl]
+enabled = true
+
+# This jail is used for manual bans.
+# To ban an IP address use: setup.sh fail2ban ban <IP>
+[custom]
+enabled = true
+bantime = 180d
+port = smtp,pop3,pop3s,imap,imaps,submission,465,sieve
```
polarathene, line 37:
```suggestion
port = smtp,pop3,pop3s,imap,imaps,submission,submissions,sieve
```

Might as well use the service mapping like the others?:

```console
$ grep submissions /etc/services
submissions     465/tcp         ssmtp smtps urd # Submission over TLS [RFC8314]
```

target/fail2ban/jail.local
```diff
@@ -28,3 +28,10 @@ enabled = true
 
 [postfix-sasl]
 enabled = true
+
+# This jail is used for manual bans.
+# To ban an IP address use: setup.sh fail2ban ban <IP>
+[custom]
+enabled = true
+bantime = 180d
+port = smtp,pop3,pop3s,imap,imaps,submission,465,sieve
```
polarathene, line 37:
```suggestion
port = smtp,pop3,pop3s,imap,imaps,submission,submissions,sieve
```

polarathene (changes requested):

config-examples/fail2ban-jail.cf
casperklein: This was on purpose. Let me explain why I decided to do it this way:

I have aggregated the port line from the ports used in the jails `postfix`, `postfix-sasl` and `dovecot`. For some reason, in fail2bans default config, all mentioned jails use the numeric value `465` instead of `submissions`:

```
[postfix]
# To use another modes set filter parameter "mode" in jail.local:
mode    = more
port    = smtp,465,submission
logpath = %(postfix_log)s
backend = %(postfix_backend)s
```

I thought this was maybe done for some reason (to prevent an issue?).
Instead of streamlining this, I went with sticking to the default config.

polarathene: > I have aggregated the port line from the ports used in the jails `postfix`, `postfix-sasl` and `dovecot`. For some reason, in fail2bans default config, all mentioned jails use the numeric value `465` instead of `submissions`

They may just be playing it safe if the config is distro-agnostic.

In a PR I was making in 2020 regarding fail2ban at the time, we had outdated configs for `imap2` or something iirc, which I think was related to earlier Ubuntu base image at the time maybe missing `imap`.. I think I detailed it in the PR.

AFAIK, these all reference `/etc/services` to lookup their port by name or alias.

You could just provide all ports instead of their service names too, personally I think we should try make it consistent. I don't see it being a problem unless we switch base images in future and for some reason the service isn't defined, we don't have a test for such, but we already use the outdated `smtps` service (now and alias) in Postfix `master.cf` (_and two doc examples with an `haproxy` line_):

https://github.com/docker-mailserver/docker-mailserver/blob/3cb5668b6f62258d8c21629c0f5e649085e4f0b6/target/postfix/master.cf#L30

As per [Postfix `master.cf` docs for the `inet` service type](http://www.postfix.org/master.5.html):

> The service name is specified as host:port, denoting the host and port on which new connections should be accepted. The host part (and colon) may be omitted.
> Either host or port may be given in symbolic form (see hosts(5) or services(5)) or in numeric form (IP address or port number).

---

> I thought this was maybe done for some reason (to prevent an issue?).

`465` was previously `smtps` which would be misleading now, we already know that those not working with mail servers as well as some outdated advice out there already can make that port a bit confusing for some.

Chances are they chose `465` just to be on the safe-side, perhaps some distros did not have `submissions` service in `/etc/services`, especially at the time that was contributed to fail2ban. Probably nobody has bothered or had reason to adjust it since (_I haven't gone looking through fail2ban history to verify_).

I do recall `/etc/services` being inconsistent across some distros when I originally wrote a PR in 2020 for `docker-mailserver` (didn't merge though).

polarathene (approved):

georglauterbach (approved):

PR #2541, opened 2022-04-10
Dockerfile: Remove not needed ENVs and add comment

casperklein:
# Description

This PR removes not needed ENVs and adds a comment to clarify, why the other ENVs are needed.

These ENVs are not needed, they are handled by `start-mailserver.sh` like all other ENVs in DMS:

- `ONE_DIR=1` 
- `ENABLE_POSTGREY`
- `SASLAUTHD_MECHANISMS`

The remaining ENVs are referenced in `target/supervisor/conf.d/saslauth.conf` and must be present when supervisord starts (tested).

If necessary, their values are adjusted by `start-mailserver.sh` on startup.

## Type of change

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ LABEL org.opencontainers.image.source="https://github.com/docker-mailserver/dock
 
 # These ENVs are referenced in target/supervisor/conf.d/saslauth.conf
 # and must be present when supervisord starts.
-# If necessary, the values are adjusted by target/scripts/start-mailserver.sh on startup.
+# If necessary, their values are adjusted by target/scripts/start-mailserver.sh on startup.
 ENV FETCHMAIL_POLL=300
 ENV POSTGREY_AUTO_WHITELIST_CLIENTS=5
 ENV POSTGREY_DELAY=300
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,14 +21,14 @@ LABEL org.opencontainers.image.url="https://github.com/docker-mailserver"
 LABEL org.opencontainers.image.documentation="https://github.com/docker-mailserver/docker-mailserver/blob/master/README.md"
 LABEL org.opencontainers.image.source="https://github.com/docker-mailserver/docker-mailserver"
 
-ENV ENABLE_POSTGREY=0
+# These ENVs are referenced in target/supervisor/conf.d/saslauth.conf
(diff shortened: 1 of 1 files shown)


georglauterbach (approved):
👍🏼

wernerfred (approved):
LGTM thx

Dockerfile
```diff
@@ -21,14 +21,14 @@ LABEL org.opencontainers.image.url="https://github.com/docker-mailserver"
...
+# These ENVs are referenced in target/supervisor/conf.d/saslauth.conf
+# and must be present when supervisord starts.
+# If necessary, their values are adjusted by target/scripts/start-mailserver.sh on startup.
 ENV FETCHMAIL_POLL=300
-ENV ONE_DIR=1
 ENV POSTGREY_AUTO_WHITELIST_CLIENTS=5
 ENV POSTGREY_DELAY=300
 ENV POSTGREY_MAX_AGE=35
 ENV POSTGREY_TEXT="Delayed by Postgrey"
-ENV SASLAUTHD_MECHANISMS=pam
```
polarathene, line 31: Just making a note that this ENV is not covered in `start-mailserver.sh` but in [`startup/setup-stack.sh`](https://github.com/docker-mailserver/docker-mailserver/blob/3cb5668b6f62258d8c21629c0f5e649085e4f0b6/target/scripts/startup/setup-stack.sh#L535), so it's not [persisted to our `/etc/dms-settings` file](https://github.com/docker-mailserver/docker-mailserver/blob/3cb5668b6f62258d8c21629c0f5e649085e4f0b6/target/scripts/startup/setup-stack.sh#L41) though I don't think that'd be an issue, if it were providing an explicit ENV to the container should resolve it?

polarathene (approved):

Dockerfile
casperklein: I noticed the same. There are other `SASLAUTHD_` ENVs that are also not covered by `start-mailserver.sh`:

https://github.com/docker-mailserver/docker-mailserver/blob/3cb5668b6f62258d8c21629c0f5e649085e4f0b6/target/scripts/startup/setup-stack.sh#L535-L541

`/etc/dms-settings` is only populated from ENVs defined in `start-mailserver.sh`. So removing this ENV from the Dockerfile is no regression.

PR #1469, opened 2020-04-24
Feature: quota per mailbox

youtous:
As mentioned in https://github.com/tomav/docker-mailserver/issues/1028, restricting quota per mailbox is an interesting feature. This PR implements a quota per mailbox.

Details:
- New commands: `setquota` and `delquota`
- New optional configuration file `dovecot-quotas.cf`
- Quota warning (email sent to the user) when quota reached 80 and 95%, and when quota is below 80% 
- Dovecot-quota and Postfix-quota are now synced (use `doveadm quota get -u user@domain` for getting quota informations)
- Supports `imap-quota` (IMAP clients requesting quota usage information)
- Quota per mailbox is DISABLED when using `SMTP_ONLY`
- Quota per mailbox is DISABLED when using `LDAP_ENABLE` _(I don't use LDAP, feel free to implement it if you want, it should not be difficult)_

TODO:
- ~Fix tests~
- ~When merged, update the wiki~

PR #2556, opened 2022-04-19
chore: Drop `setup.sh` DATABASE fallback ENV

polarathene:
# Description

The `DATABASE` ENV fallback was added in [this Dec 2016 PR](https://github.com/docker-mailserver/docker-mailserver/pull/441), no justification was provided.

Likewise for [`USER_DATABASE` in `delquota` and `addquota` files in Apr 2020](https://github.com/docker-mailserver/docker-mailserver/pull/1469) which was probably based on the earlier existing convention used for `DATABASE`.

---

Neither are in usage by our tests or detailed in documentation.

These are being dropped as a [recent PR](https://github.com/docker-mailserver/docker-mailserver/pull/2535) adapted existing scripts where `DATABASE` with an ENV fallback would be conflicting with usage elsewhere unless providing a temporary ENV when calling the command.

Until an actual use-case requires this, it's probably best to drop such support.

### DIFF
--- a/target/bin/addalias
+++ b/target/bin/addalias
@@ -3,7 +3,7 @@
 # shellcheck source=../scripts/helpers/index.sh
 source /usr/local/bin/helpers/index.sh
 
-DATABASE=${DATABASE:-/tmp/docker-mailserver/postfix-virtual.cf}
+DATABASE='/tmp/docker-mailserver/postfix-virtual.cf'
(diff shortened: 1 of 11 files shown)


georglauterbach (approved):
👍🏼

casperklein (approved):

PR #2564, opened 2022-04-27
Streamline 'printf' usage and consequently use a format string

casperklein:
# Description

This PR streamlines the `printf` usage and consequently uses a format string. This (best) practice also avoids possible problems with strings containing special characters like `%s`, `\n` etc.

## Type of change

## Checklist:

### DIFF
--- a/.github/workflows/default_on_push.yml
+++ b/.github/workflows/default_on_push.yml
@@ -24,7 +24,7 @@ jobs:
           submodules: recursive
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v1.6.0
+        uses: docker/setup-buildx-action@v1.7.0
         id: buildx
 
       - name: Cache Docker layers
@@ -68,7 +68,7 @@ jobs:
 
       - name: Prepare tags
         id: prep
-        uses: docker/metadata-action@v3.7.0
+        uses: docker/metadata-action@v3.8.0
         with:
           images: |
             ${{ secrets.DOCKER_REPOSITORY }}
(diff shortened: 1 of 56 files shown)


casperklein:
Converting this

`printf "\e[35mADDALIAS\e[31m(\e[93m8\e[31m)`

to

`printf '%s' "\e[35mADDALIAS\e[31m(\e[93m8\e[31m)`

does not work. `\e[35m` will not be evaluated as the color purple.

All scripts already source `helpers/index.sh`. I think it's the best, to centrally define all colors and then just use them as vars.
There is already something like this here:

https://github.com/docker-mailserver/docker-mailserver/blob/49a337ccfaf04ee58aefcd1b10e437817c228bd6/target/scripts/helpers/log.sh#L3-L8

But not all colors are covered there + variable names refer to log stuff.

I propose to:

1. add missing colors
2. change variable names to something more general like `$RED` or `$COLOR_RED` instead of `$LOG_RED`.
3. and use these variables in other scripts.

If there is consensus, I'll take care of this in a separate PR and adjust this PR afterwards.

georglauterbach:
> I propose to:
> 
> 1. add missing colors
> 2. change variable names to something more general like `$RED` or `$COLOR_RED` instead of `$LOG_RED`.
> 3. and use these variables in other scripts.
> 
> If there is consensus, I'll take care of this in a separate PR and adjust this PR afterwards.

Sounds very good to me - I’m all for it 👍🏼

georglauterbach (approved):

polarathene (approved):

PR #2565, opened 2022-04-28
Use color variables instead of escape codes

casperklein:
# Description

This PR streamlines the color usage of the scripts. Color variables are now globally defined in `log.sh`. I did not use a separate file like `colors.sh`, because some scripts only source `log.sh`

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/helpers/log.sh
+++ b/target/scripts/helpers/log.sh
@@ -1,11 +1,25 @@
 #! /bin/bash
 
-LOG_RESET='\e[0m'
-LOG_LGRAY='\e[37m'
-LOG_LBLUE='\e[94m'
-LOG_BLUE='\e[34m'
-LOG_LYELLOW='\e[93m'
-LOG_RED='\e[91m'
+# Color variables for global usage
+RED=$(echo -ne     '\e[0;31m')
+GREEN=$(echo -ne   '\e[0;32m')
+YELLOW=$(echo -ne  '\e[0;33m')
+BLUE=$(echo -ne    '\e[0;34m')
+PURPLE=$(echo -ne  '\e[0;35m')
+CYAN=$(echo -ne    '\e[0;36m')
+WHITE=$(echo -ne   '\e[0;37m')
+
+# Light/bold variants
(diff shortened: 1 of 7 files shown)


target/scripts/helpers/log.sh
```diff
@@ -1,11 +1,28 @@
...
+LRED=$(echo -ne    '\e[1;31m')
+LGREEN=$(echo -ne  '\e[1;32m')
+LYELLOW=$(echo -ne '\e[1;33m')
+LBLUE=$(echo -ne   '\e[1;34m')
+LPURPLE=$(echo -ne '\e[1;35m')
+LCYAN=$(echo -ne   '\e[1;36m')
+LWHITE=$(echo -ne  '\e[1;37m')
+
+ORANGE=$(echo -ne  '\e[38;5;214m')
+RESET=$(echo -ne   '\e[0m')
```
georglauterbach, line 25: I was just thinking: maybe we should prefix this with `COLOR_`?

casperklein: I thought about the same. But then decided to omit it:

- i think it's clear from the variable name what the usage is
- better readability in scripts with lots of colors (IMO), e.g. `${COLOR_LYELLOW}` is quiet long compared to `${YELLOW}`.

But I don't mind. We can change this. Lets wait what other maintainers say.

georglauterbach: I'm absolutely fine with it as is, just wanted to ask here before we merge this and think about it later :)

polarathene (approved):
> i think it's clear from the variable name what the usage is

I'm not sure if that applies to `${RESET}` though?

Depending on which PR gets merged first, might want to keep https://github.com/docker-mailserver/docker-mailserver/pull/2535 in mind since it's adding a bunch of new `setup` commands that would need to be converted.

casperklein:
> I'm not sure if that applies to ${RESET} though?

Standing alone, maybe not. But it's only used in context of color usage so I think it's fine.

georglauterbach (approved):
If we, in the future, realize that we may need the prefix we can always add it :)

PR #2572, opened 2022-05-02
Add version tag to avoid error "invalid config"

matheusbach:
# Description

Add version tag to avoid error "invalid config"

## Type of change

## Checklist:

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,3 +1,5 @@
+version: 
+
 services:
   mailserver:
     image: docker.io/mailserver/docker-mailserver:latest


casperklein:
Thank you for your contribution 👍

The version field is just informational and [deprecated](https://github.com/compose-spec/compose-spec/blob/master/spec.md#compose-file) for some time now.

>Top-level version property is defined by the specification for backward compatibility but is only informative.
>A Compose implementation SHOULD NOT use this version to select an exact schema to validate the Compose file, but >prefer the most recent schema at the time it has been designed.

https://github.com/compose-spec/compose-spec/blob/master/spec.md#version-top-level-element

Can you explain, in which case you encounter a problem?

PR #2591, opened 2022-05-13
scripts: fix error message caused by `grep` on first starts

georglauterbach:
# Description

On first start, the check for accounts would emit a warning, but also an
error:

`grep: /tmp/docker-mailserver/postfix-accounts.cf: No such file or directory`

since the `postfix-accounts.cf` file is not yet there. This commit fixes
the issue, so only the warning is logged correctly.

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/startup/setup-stack.sh
+++ b/target/scripts/startup/setup-stack.sh
@@ -300,7 +300,7 @@ function _setup_dovecot_local_user
   local SLEEP_PERIOD='10'
   for (( COUNTER = 11 ; COUNTER >= 0 ; COUNTER-- ))
   do
-    if [[ $(grep -cE '.+@.+\|' /tmp/docker-mailserver/postfix-accounts.cf 2>dev/null || printf '%s' '0') -ge 1 ]]
+    if [[ $(grep -cE '.+@.+\|' /tmp/docker-mailserver/postfix-accounts.cf 2>/dev/null || printf '%s' '0') -ge 1 ]]
     then
       return 0
     else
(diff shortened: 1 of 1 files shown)


polarathene (approved):

target/scripts/startup/setup-stack.sh
```diff
@@ -300,7 +300,7 @@ function _setup_dovecot_local_user
   local SLEEP_PERIOD='10'
   for (( COUNTER = 11 ; COUNTER >= 0 ; COUNTER-- ))
   do
-    if [[ $(grep -cE '.+@.+\|' /tmp/docker-mailserver/postfix-accounts.cf) -ge 1 ]]
+    if [[ $(grep -cE '.+@.+\|' /tmp/docker-mailserver/postfix-accounts.cf 2>dev/null || printf '0') -ge 1 ]]
```
casperklein, line 303: I would prefer a simple `echo` and avoid `printf` when not strictly needed:

```suggestion
    if [[ $(grep -cE '.+@.+\|' /tmp/docker-mailserver/postfix-accounts.cf 2>dev/null || echo '0') -ge 1 ]]
```

Or at least with a format string, to keep in sync with #2564:

```suggestion
    if [[ $(grep -cE '.+@.+\|' /tmp/docker-mailserver/postfix-accounts.cf 2>dev/null || printf '%s' '0') -ge 1 ]]
```

casperklein (changes requested):

target/scripts/startup/setup-stack.sh
georglauterbach: I‘ll go with the format string:)

polarathene (approved):

casperklein (approved):

PR #2592, opened 2022-05-14
scripts: add `chmod +x user-patches.sh`

pyy:
# Description

`chmod +x user-patches.sh`
in case of forgetting to set +x on the host.

## Type of change

## Checklist:


georglauterbach:
Why exactly do we need this? The script is executed by `/bin/bash` in the next line, and not via the shebang. That's the whole point of using `/bin/bash`, to avoid `chmod +x`, because this does not work in K8s setups.

Thank you this addition, but it actually does not add any value - in fact, it would cause an error in my logs as I'm using K8s :) We should be fine without the `chmod +x` :D

PR #2597, opened 2022-05-17
service(postfix): Better handling of the `compatibility_level` setting

polarathene:
# Description

I wanted to fix a typo specifically. But also looked into why the postfix compatibility level was set to `2` ([both from the same PR in Sep 2017](https://github.com/docker-mailserver/docker-mailserver/pull/708)):

> _The minimal compatibility level had to be set to 2 in order for pcre to work and not creating errors in the logging._

It seems leaving this at the [default (`0`)](http://www.postfix.org/postconf.5.html#compatibility_level) is passing the mentioned PCRE test just fine.

It was not made clear what logging errors were a concern. There was one regarding the `sender-cleanup` service that the PR added with implicit chroot default. The [default compatibility level `0` would set this to `y`](http://www.postfix.org/COMPATIBILITY_README.html#chroot), otherwise it'd be `n`. This PR makes the `n` explicit.

---

If choosing to leave Postfix `compatibility_level` at the default `0`, ~~when the config is changed (eg: via `postconf`), it will log the following each time~~ (_only seems to happen with `postfix reload` or `supervisorctl restart postfix`_):

```
May 17 07:46:10 mail postfix[3482]: Postfix is running with backwards-compatible default settings
May 17 07:46:10 mail postfix[3482]: See http://www.postfix.org/COMPATIBILITY_README.html for details
May 17 07:46:10 mail postfix[3482]: To disable backwards compatibility use "postconf compatibility_level=2" and "postfix reload"
```

This is noise, so reverting the `compatibility_level` back to being set to `2` would be understandable. Would we want this in `_setup_postfix_override_configuration` still, or in `target/postfix/main.cf`? Leaving it at `0` would catch any compatibility default settings being applied during tests though; each level is listed [here](http://www.postfix.org/COMPATIBILITY_README.html).

## Type of change


georglauterbach:
Haha :D I noticed the `compatibility_level` thing too :D I'm going to go ahead and remove it from my PR (#2598) as well :D

georglauterbach (dismissed):
LGTM

georglauterbach:
> This is noise, so reverting the compatibility_level back to being set to 2 would be understandable. Would we want this in _setup_postfix_override_configuration still, or in target/postfix/main.cf?

Sorry, did not read all in the beginning. I would actually add it to `/etc/postfix/main.cd` with a value of `2` as before. This way, we

1. let the user decide whether to override it with a custom value
2. remove clutter in a function that actually does other things

polarathene:
- I've applied `compatibility_level = 2` to `main.cf` as suggested.
- In our tests `postfix-main.cf` override, I have set the level back to the default `0`. This shouldn't affect tests negatively, but potentially trigger a test failure that checks logs when a backwards compatibility implicit default differs from the implicit default of newer releases. If such failures are caught in future, we would then know to provide an explicit default (older or newer default, depending on what makes most sense).

georglauterbach (approved):

target/postfix/main.cf
```diff
@@ -101,5 +101,7 @@ policyd-spf_time_limit = 3600
 # Header checks for content inspection on receiving
 header_checks = pcre:/etc/postfix/maps/header_checks.pcre
 
-# Remove unwanted headers that reveail our privacy
+# Remove unwanted headers that reveal our privacy
 smtp_header_checks = pcre:/etc/postfix/maps/sender_header_filter.pcre
+
+compatibility_level = 2
```
casperklein, line 107:
```suggestion
# Leaving Postfix compatibility_level at the default 0, when the config is changed (e.: via postconf), it will log the following each time:
#    Postfix is running with backwards-compatible default settings
#    See http://www.postfix.org/COMPATIBILITY_README.html for details
#    To disable backwards compatibility use "postconf compatibility_level=2" and "postfix reload"
compatibility_level = 2
```

casperklein (approved):
LGTM. You may add my comment suggestion if you want.

polarathene:
> You may add my comment suggestion if you want.

I will do that thanks!

I saw something related to the compatibility level change in postfix docs that I want to look into further before merging this. I want to make sure it does get caught if it's compatibility default would impact tests :sweat_smile:

polarathene:
Reverted the change to set `compatibility_level=0` for tests. Several settings didn't output any backwards-compatible log but did fallback to a conditional default shown by `postconf`. I believe when those features are interacted with, then the logs are emitted. It only affected the `chroot` one due to the service being started.

Other compatibility rules don't seem to be a concern with our current config. Since we can't properly detect each with the log detection and it may result in tests behaving differently than what we ship, I think it's best to ignore that idea :sweat_smile: 

---

When the next Debian release happens we should have Postfix 3.6 available and can bump the `compatibility_level` to `3.6`.

georglauterbach (approved):

casperklein (approved):

PR #2598, opened 2022-05-18
scripts: improve custom user-supplied Postfix configuration

### LINKED ISSUE #2594: Adding lines to postfix-main.cf parsing error
I am trying to add the following postfix rules (which other people are using) via postfix-main.cf:

```
#Postscreen for DNSBL 
postscreen_dnsbl_action = enforce
postscreen_dnsbl_sites = zen.spamhaus.org*3
        b.barracudacentral.org=127.0.0.[2..11]*2
        bl.spameatingmonkey.net*2
        bl.spamcop.net
        dnsbl.sorbs.net
       swl.spamhaus.org*-4,
       list.dnswl.org=127.[0..255].[0..255].0*-2,
       list.dnswl.org=127.[0..255].[0..255].1*-4,
       list.dnswl.org=127.[0..255].[0..255].[2..3]*-6
```

However, when I docker-compose up , I get the following errors .. I've doubled checked the format .. what might I be missing ?

```
mailserver    | postconf: fatal: missing '=' after attribute name: "bl.spameatingmonkey.net*2"
mailserver    | postconf: fatal: missing '=' after attribute name: "bl.spamcop.net"
mailserver    | postconf: fatal: missing '=' after attribute name: "dnsbl.sorbs.net"
```

georglauterbach:
# Description

The new way of providing your custom Postfix configuration is similar to
the old way, but now, the whole file is appended at once, and then
"filtered" so Postfix is okay with it.

Users can now override the default Postfix configuration and do not have
to worry about providing the config line-by-line.

The documentation was adjusted accordingly. It emphasises how the
configuration is applied to enable users to understand what they can do
and how they can achieve that.

Note: The new way of adjusting Postfix _should in theory_ be backwards
compatible. (Nothing is ever certain with Postfix...) I had to adjust for
a read-write-conflict with `postconf -nf` though, but this should be fine.

~~I also moved the default compatibility level into the configuration file instead of having it set in the function (where it cannot be properly overridden by the user...).~~ see #2597

---

This should be the last PR for `v11.1.0` :)

Fixes #2594

## Type of change

## Checklist:


georglauterbach:
This PR is expected to fail until #2597 is merged (because of `compatibility_level`).

reneploetz:
Is there a specific reason we cannot use the same idea for master.cf?
At first glance ```postconf -Mf``` seems to show the contents of master.cf (loosing comments and formatting too) so we could concat too, can't we?

georglauterbach:
I guess I was not looking at the manpage of `postconf` close enough, but that is a good idea @reneploetz! @polarathene what‘s your take?

polarathene:
> Is there a specific reason we cannot use the same idea for `master.cf`?

Yes. I looked into that too when I did my earlier attempt. We process that differently.. Rather than appending config, it's using a different syntax to update which is intended for single line updates:

https://github.com/docker-mailserver/docker-mailserver/blob/0d30b92a83118e7c1cea7cdc40007a3f41d4fe6c/test/config/postfix-master.cf#L1

This is from our tests. It updates an existing parameter for a service with that `service/type/parameter=value` syntax, rather than appending or overriding.

~~I have not looked if we can get the same append + merge behaviour.~~ (_**EDIT:** Does not work, appended an existing service line and updated one of the parameters (`-o` lines), `postconf -Mf` did not merge it. Same with updating chroot for a service line appended.. treated as separate service_)

~~If we could, that would be a breaking change though to any users that use the current format supported. It may be better to wait until a user requests support for extending rather than modifying `master.cf`, or at least queue separately for a major version release.~~ (_**EDIT:** Didn't read the full message, you already mentioned it'd be a breaking change due to format change_ :sweat_smile: )

---

This update format is described as:

> _With `-P`, edit the [`master.cf`](https://www.postfix.org/master.5.html) configuration file, and add or update one or more service parameter settings (`-o parameter=value` settings) with new values as specified with `service/type/parameter=value`_

That may not cover all desired changes/updates though (_there are other modes for `master.cf` management with `postconf` too_).

---

> @polarathene what‘s your take?

Better to leave as-is.

It doesn't seem like `postconf -Mf` will provide same merge of duplicates like `postconf -nf` does. If there is demand for appending/replacing `master.cf` (or `main.cf`) instead of modifying the one we ship in `/etc/postfix`, we could look into that later. For now users can use `user-patches.sh` if they need that.

polarathene (changes requested):
Nice improvements to the docs! :grinning: 

The config update logic LGTM! :+1:

georglauterbach:
> The other test failure I understand will be resolved by related PR dependency, but I didn't expect this one:
> 
> ```
> not ok 93 ssl(letsencrypt): Traefik 'acme.json' (*.example.test) in 42042ms
> # (from function `assert_output' in file test/test_helper/bats-assert/src/assert.bash, line 239,
> #  from function `_has_matching_line' in file test/mail_ssl_letsencrypt.bats, line 203,
> #  from function `_should_have_valid_config' in file test/mail_ssl_letsencrypt.bats, line 194,
> #  from function `_acme_wildcard' in file test/mail_ssl_letsencrypt.bats, line 157,
> #  in test file test/mail_ssl_letsencrypt.bats, line 180)
> #   `_acme_wildcard' failed
> # [   INF   ]  mail.example.test is up and running
> # [  DEBUG  ]  2022-05-18 10:41:27  Chagedetector is ready
> # 
> # -- output differs --
> # expected : smtpd_tls_chain_files = /etc/letsencrypt/live/example.test/key.pem /etc/letsencrypt/live/example.test/fullchain.pem
> # actual   : smtpd_tls_chain_files = /etc/letsencrypt/live/example.test/key.pem /etc/letsencrypt/live/example.test/fullchain.pem /etc/letsencrypt/live/mail.example.test/fullchain.pem
> # --
> # 
> # mail_ssl_letsencrypt
> ok 94 checking ssl: ENV vars provided are valid files in 387ms
> not ok 95 checking ssl: manual configuration is correct in 132ms
> # (from function `assert_output' in file test/test_helper/bats-assert/src/assert.bash, line 239,
> #  in test file test/mail_ssl_manual.bats, line 53)
> #   `assert_output "smtpd_tls_chain_files = ${PRIMARY_KEY} ${PRIMARY_CERT} ${FALLBACK_KEY} ${FALLBACK_CERT}"' failed
> # 
> # -- output differs --
> # expected : smtpd_tls_chain_files = /etc/dms/tls/key /etc/dms/tls/cert /etc/dms/tls/fallback_key /etc/dms/tls/fallback_cert
> # actual   : smtpd_tls_chain_files = /etc/dms/tls/key /etc/dms/tls/cert
> # --
> #
> ```
> 
> I will re-run the tests to double-check that the changes in this PR are actually causing that.

The Traefik test is still failing :=

polarathene (dismissed):

reneploetz (approved):

casperklein (approved):

polarathene:
> The Traefik test is still failing :O

I've looked into this and identified the problem :)

The tests are not matching the expected values, which are in fact there but `postconf -nf` is folding at a certain value length, so missing values have been split into a separate line(s).

This occurs for quite a few settings with long values, and may be at risk of scripts that use `grep`/`sed` or similar tailored with an expectation of the value on a single line, instead of delegating read/write via `postconf`.

Likewise, if any scripts rely on toggling a setting or similar logic that checks for `#` + `<setting name>`, this approach may cause breakage there.. I have not looked through scripts to see where we are doing these sort of operations though.

---

Fix for now would be to not fold values for storage. But this does not rule out risk with scripts built with `#` in mind.

polarathene:
I've noticed with this change that the logs get a bit of spam with `postconf -nf` line. This would only happen with the first occurrence and I could not repeat from manually running the commands. I could when I chained these two, so that `postconf` was run immediately afterwards:

```bash
cat /tmp/docker-mailserver/postfix-main.cf >>/etc/postfix/main.cf && \
  postconf -n >/tmp/postfix-main-new.cf
```

You'll get logs / output such as the following (blank lines added to show grouping):

```
[  TRACE  ]  Setting up Postfix Override configuration

postconf: warning: /etc/postfix/main.cf, line 103: overriding earlier entry: recipient_delimiter=+
postconf: warning: /etc/postfix/main.cf, line 107: overriding earlier entry: readme_directory=no

postconf: warning: /etc/postfix/main.cf, line 6: overriding earlier entry: readme_directory=/tmp
postconf: warning: /etc/postfix/main.cf, line 16: overriding earlier entry: recipient_delimiter=~

postconf: warning: /etc/postfix/main.cf, line 103: overriding earlier entry: recipient_delimiter=+
postconf: warning: /etc/postfix/main.cf, line 107: overriding earlier entry: readme_directory=no

postconf: warning: /etc/postfix/main.cf, line 6: overriding earlier entry: readme_directory=/tmp
postconf: warning: /etc/postfix/main.cf, line 16: overriding earlier entry: recipient_delimiter=~
postconf: warning: /etc/postfix/main.cf, line 103: overriding earlier entry: recipient_delimiter=+
postconf: warning: /etc/postfix/main.cf, line 107: overriding earlier entry: readme_directory=no

postconf: warning: /etc/postfix/main.cf, line 6: overriding earlier entry: readme_directory=/tmp
postconf: warning: /etc/postfix/main.cf, line 16: overriding earlier entry: recipient_delimiter=~
postconf: warning: /etc/postfix/main.cf, line 103: overriding earlier entry: recipient_delimiter=+
postconf: warning: /etc/postfix/main.cf, line 107: overriding earlier entry: readme_directory=no

postconf: warning: /etc/postfix/main.cf, line 6: overriding earlier entry: readme_directory=/tmp
postconf: warning: /etc/postfix/main.cf, line 16: overriding earlier entry: recipient_delimiter=~
postconf: warning: /etc/postfix/main.cf, line 103: overriding earlier entry: recipient_delimiter=+
postconf: warning: /etc/postfix/main.cf, line 107: overriding earlier entry: readme_directory=no

postconf: warning: /etc/postfix/main.cf, line 6: overriding earlier entry: readme_directory=/tmp
postconf: warning: /etc/postfix/main.cf, line 16: overriding earlier entry: recipient_delimiter=~
postconf: warning: /etc/postfix/main.cf, line 103: overriding earlier entry: recipient_delimiter=+
postconf: warning: /etc/postfix/main.cf, line 107: overriding earlier entry: readme_directory=no

postconf: warning: /etc/postfix/main.cf, line 6: overriding earlier entry: readme_directory=/tmp
postconf: warning: /etc/postfix/main.cf, line 16: overriding earlier entry: recipient_delimiter=~
postconf: warning: /etc/postfix/main.cf, line 103: overriding earlier entry: recipient_delimiter=+
postconf: warning: /etc/postfix/main.cf, line 107: overriding earlier entry: readme_directory=no

[  TRACE  ]  Loaded '/tmp/docker-mailserver/postfix-main.cf'
```

I'm not entirely familiar with what's going on, but I guess the `cat` append operation has not finished writing by the time the `postconf` command begins? I didn't expect that personally :|

This is running in a local VM with an XFS filesystem (virtual disk file on host internal SATA SSD) on a host OS also with XFS that is booted from a raw disk image (originally used for a VM) which is stored on XFS filesystem for an external USB SSD.

I don't know if something weird is going on I/O wise there with my setup if you're unable to reproduce. I used `time` on the reproduction code and it takes 1.5 seconds while logging the warnings (staggered). Individually timed with delay between both commands they take less than 50ms and only two warnings like expected are output:

```

**Comment by georglauterbach** [comment]
Very interesting observation @polarathene. I am not sure what the cause is, but I can add a small delay and silence the warnings, as they are indeed not in any way useful :)

File: target/scripts/startup/setup-stack.sh [hunk 1]
```diff
@@ -790,20 +790,16 @@ function _setup_postfix_virtual_transport
...
-      if [[ ${LINE} =~ ^[a-z] ]]
-      then
-        postconf -e "${LINE}"
-      fi
-    done < /tmp/docker-mailserver/postfix-main.cf
-    _log 'trace' "Loaded '/tmp/docker-mailserver/postfix-main.cf'"
+    cat /tmp/docker-mailserver/postfix-main.cf >>/etc/postfix/main.cf
+    sleep 1
+    # do not directly output to 'main.cf' as this causes a read-write-conflict
+    postconf -nf >/tmp/postfix-main-new.cf 2>/dev/null
```
polarathene, line 800:
```suggestion
    postconf -n >/tmp/postfix-main-new.cf 2>/dev/null
```

polarathene:
Whoops that commit was meant to be a `fix` not a `chore` :sweat_smile: 

Hope you don't mind, it should fix the failing tests. Which perhaps should be using `postconf` in future to interact with `main.cf`?

georglauterbach:
> Whoops that commit was meant to be a `fix` not a `chore` sweat_smile
> 
> Hope you don't mind, it should fix the failing tests.

> Which perhaps should be using `postconf` in future to interact with `main.cf`?

Yes, that may really be a good idea :)

polarathene:
> This is still expected, right?

Yep, that's due to depending on the other PR.

polarathene:
Creating a two line `main.cf` and just using `postconf` to read the setting override in this same way also reproduces the problem:

```shell
$ echo -e "max_idle = 300s\nmax_idle = 600s" >/etc/postfix/main.cf && postconf max_idle

postconf: warning: /etc/postfix/main.cf, line 2: overriding earlier entry: max_idle=300s
postconf: warning: /etc/postfix/main.cf, line 1: overriding earlier entry: max_idle=600s
postconf: warning: /etc/postfix/main.cf, line 2: overriding earlier entry: max_idle=300s
postconf: warning: /etc/postfix/main.cf, line 1: overriding earlier entry: max_idle=600s
postconf: warning: /etc/postfix/main.cf, line 2: overriding earlier entry: max_idle=300s
postconf: warning: /etc/postfix/main.cf, line 1: overriding earlier entry: max_idle=600s
postconf: warning: /etc/postfix/main.cf, line 2: overriding earlier entry: max_idle=300s
postconf: warning: /etc/postfix/main.cf, line 1: overriding earlier entry: max_idle=600s
postconf: warning: /etc/postfix/main.cf, line 2: overriding earlier entry: max_idle=300s
postconf: warning: /etc/postfix/main.cf, line 1: overriding earlier entry: max_idle=600s
postconf: warning: /etc/postfix/main.cf, line 2: overriding earlier entry: max_idle=300s
max_idle = 600s
```

Adding sleep will reduce the warning spam, but likely isn't contributing anything useful. I would need to test in a different environment to see if that occurs outside of this VM, it may just be disk I/O perf issue and `postconf` detecting something that causes it to restart multiple times (roughly every 250ms, for total 1.5 seconds).

target/scripts/startup/setup-stack.sh
```diff
@@ -790,20 +790,16 @@ function _setup_postfix_virtual_transport
...
-      # all valid postfix options start with a lower case letter
-      # http://www.postfix.org/postconf.5.html
-      if [[ ${LINE} =~ ^[a-z] ]]
-      then
-        postconf -e "${LINE}"
-      fi
-    done < /tmp/docker-mailserver/postfix-main.cf
-    _log 'trace' "Loaded '/tmp/docker-mailserver/postfix-main.cf'"
+    cat /tmp/docker-mailserver/postfix-main.cf >>/etc/postfix/main.cf
+    sleep 1
```
polarathene, line 798: Sleep doesn't seem helpful, lets revert that.

```suggestion
```

polarathene (approved):

casperklein (approved):

PR #2604, opened 2022-05-29
refactor: Revised `relay.sh` helper

polarathene:
# Description

This is a refactor of the current relay support offered via `docker-mailserver`. For the most part, this is preparation for a future PR that will make some breaking changes to revise the relay support offered. I want to simplify review for that by covering the changes here.

Reviewers that are not familiar with present relay support should have no trouble getting up to speed. Supporting comments can be dropped before merge or via follow-up commit if necessary. Changes have been staged out into separate commits (_with their own commit messages for context_) should that assist with review.

I also have a related change with `SASL_PASSWD` support that would have added noise to the review here. I've opted to cherry-pick it out of this PR and [into it's own PR](https://github.com/docker-mailserver/docker-mailserver/pull/2605) that will follow upon merging of this PR. (_**EDIT:** PR opened, will be rebased after merging this one_)

### Overview of changes (_ordered by appearance in `relay.sh`_)

- Inline docs heavily revised. Any maintainer touching relay support should easily be aware of associated content in the project to this feature that `helpers/relay.sh` manages.
- `_relayhost_default_port_fallback()` has been dropped. It no longer serves a purpose.
- `_env_relay_host()` has been added. It provides the default port fallback and ensures the configured relay-host is in sync between both `/etc/postfix/sasl_passwd` and `/etc/postfix/relayhost_map`, which is required for successful lookup.
- Config examples revised and shifted to the end of the document due to size. Additional supporting information for maintainers to be aware of included. Likely to be relocated in future to a "maintainers" section of our docs as historically some maintainers have removed contextual information (eg: the Dockerfile used to better document some layers).
- Some `main.cf` settings configured via `postconf` have been relocated from `_relayhost_configure_postfix()` to their relevant sections of code. There may be **potential risk** of breakage in a deployment that relied on those being set regardless when `RELAY_HOST` ENV was set, but did not leverage our feature support that specifically requires them (_and for which they were originally added for_). It is intended in future to manage separate transports beyond the present default `master.cf:smtp` one.
- `_populate_relayhost_map` should function the same. The changes are mostly to make it's purpose and concerns for changes in future clear (_normally I keep notes locally instead of inlining, but have chosen to make sure others can understand the feature quickly and what changes need to be done should future work lag behind on my part, someone else could pick it up_). Plan is to better implement the features intent without the excessive approach with accounts currently used.
- `_relayhost_configure_postfix()` is simplified. Some settings configured there were unnecessary, see associated commit for details. In future a separate SMTP client transport will be configured, which will better support the use-cases from past relay-host contributions and more easily allow for supporting multiple relay-hosts with different requirements (no auth, explicit TLS, implicit TLS, varying ports).

### Relay-host feature history:

I went through the history of the feature to understand why decisions for current support exist, use-cases and if any breakage may be an issue (_especially around `SASL_PASSWD` and why that path looked borked_).

Documenting that history here for others, which may be helpful to maintainers in future rather than trawling `git blame` :sweat_smile: 

- [Original relay support introduced in Feb 2016](https://github.com/docker-mailserver/docker-mailserver/pull/83) (`#83`):
  - Introduced `main.cf` customization via user config file (`postfix-main.cf`).
  - Introduced `SASL_PASSWD` ENV to automate creating a hash table (_the generated `.db` file received `chown`/`chmod` adjustments, then temporary input file removed_) as opposed to using _plaintext_ with the `static` table type in the `main.cf` setting.
  - Introduced `smtp_tls_security_level=none` to amavis transports. `smtp_tls_wrappermode=no` was probably not set as port 465 was not as relevant in 2016.
- [May 2016 expanded support for AWS SES](https://github.com/docker-mailserver/docker-mailserver/pull/194) (`#194`):
  - Simplified relay configuration via alternative ENV, but was intended with AWS SES in mind. It provided the required `main.cf` default settings for configuring an authenticated relay-host, instead of expecting the user to know what to add with a `postfix-main.cf` themselves. It was initially hard-coded to port 25, but future PRs resolved that.
  - `smtp_sasl_password_maps` was only enabled as part of this automated config. The prior `SASL_PASSWD` ENV support was very basic and did not include even that.
  - `smtp_tls_CAfile` was also introduced for relay configs via this PR. A [July 2016 PR introduced `smtp_tls_CApath` as a default](https://github.com/docker-mailserver/docker-mailserver/pull/243) (`#243`), making the relay specific one redundant.
- A [Nov 2016 PR changed the table types used from `hash` to `texthash`](https://github.com/docker-mailserver/docker-mailserver/pull/382) (`#382`).
- [April 2018 expands support for multiple relays and credentials](https://github.com/docker-mailserver/docker-mailserver/pull/926) (`#926`):
  - Use cases [requested here](https://github.com/docker-mailserver/docker-mailserver/issues/922)  (`#922`).
  - The bulk of relay tests were introduced in this PR.
  - This is also where relay code was duplicated over into `check-for-changes.sh`.
  - `RELAY_HOST` and related ENV were added, replacing earlier ENV for AWS SES (_despite the ENV name, was supporting other relays_).
  - Introduced the `setup.sh` relay commands.
  - Introduced `postfix-relayhost.cf` and `postfix-sasl-password.cf` files to support multiple per sender relay-hosts and/or credentials.
  - At the same time, it dropped configuring `main.cf:relayhost` in favor of adopting sender dependent mappings. This seemed unnecessary, as it would append all domains from `postfix-accounts.cf` by default, effectively mimicking `main.cf:relayhost` anyway.. Presumably this was to provide support to opt-out a domain via `postix-relayhost.cf`, but what they'd actually want really would be to opt-in such a domain to a separate outbound transport (_a future PR will introduce this and bring back `main.cf:relayhost`_).
- [Jan 2019 added `DEFAULT_RELAY_HOST` to support default relay without authentication](https://github.com/docker-mailserver/docker-mailserver/pull/1104) (`#1104`):
  - As existing relay support was enforcing that if using `RELAY_HOST`. This would also be addressed with the future relay-host refactor.
- [Aug 2020 refactored relay support a bit, and included virtual domains into the default `RELAY_HOST` per sender mapping support](https://github.com/docker-mailserver/docker-mailserver/pull/1596) (`#1596`):
  - Introduced the `_populate_relayhost_map` function to try sync some of the duplicated code that wasn't complete in the earlier copy added to the change detector script.
- An [Oct 2021 refactor addressed concern with duplicate relay code falling out of sync and extracted out to the current form of a `relay.sh` helper](https://github.com/docker-mailserver/docker-mailserver/pull/2260) (`#2260`):
  - This was by me. It should not have changed anything other than better organizing the feature support and merging back into a central location instead of two copies.

## Type of change

- [x] Breaking change (_Potentially with `postconf` settings relocation_)

## Checklist:


target/scripts/helpers/relay.sh
```diff
@@ -25,73 +74,112 @@ function _relayhost_sasl
...
-    # white-space separates value pairs (any length is valid)
-    echo "[${RELAY_HOST}]:${RELAY_PORT} ${RELAY_USER}:${RELAY_PASSWORD}" >> /etc/postfix/sasl_passwd
+    echo "$(_env_relay_host)    ${RELAY_USER}:${RELAY_PASSWORD}" >> /etc/postfix/sasl_passwd
   fi
 
   _sasl_set_passwd_permissions
+
+  # Technically if only a single relay host is configured, a `static` lookup table could be used instead?:
+  # postconf "smtp_sasl_password_maps = static:${RELAY_USER}:${RELAY_PASSWORD}"
+  postconf 'smtp_sasl_password_maps = texthash:/etc/postfix/sasl_passwd'
```
georglauterbach, line 100: I like the `texthash` method :)

target/scripts/helpers/relay.sh
```diff
@@ -25,73 +74,112 @@ function _relayhost_sasl
...
   then
     _log 'trace' "Adding relay mappings from postfix-relaymap.cf"
-    # keep lines which are not a comment *and* have a destination.
-    sed -n '/^\s*[^#[:space:]]\S*\s\+\S/p' /tmp/docker-mailserver/postfix-relaymap.cf >> /etc/postfix/relayhost_map
+
+    # Match two values with some white-space between them (eg: `@example.test [relay.service.test]:465`):
+    local MATCH_VALUE_PAIR='\S*\s+\S'
+
+    # Copy over lines which are not a comment *and* have a destination.
+    sed -n -r "/${MATCH_VALID}${MATCH_VALUE_PAIR}/p" /tmp/docker-mailserver/postfix-relaymap.cf >> /etc/postfix/relayhost_map
```
georglauterbach, line 136: Very tiny nitpick here:

``` SUGGESTION
    sed -n -r "/${MATCH_VALID}${MATCH_VALUE_PAIR}/p" /tmp/docker-mailserver/postfix-relaymap.cf >>/etc/postfix/relayhost_map
```

georglauterbach (approved):
LGTM 👍🏼 

Very nice with the comments you added to the code :D

target/scripts/helpers/relay.sh
polarathene: In future it may be better to use `hash` table type or provide a way to support both. This would only be relevant if the change detection was smarter on what changes were applied.

At least it should be better in situations where some databases are updated at a higher frequency, eg on a larger scale via API. For bulk updates it's probably not a big issue as long as the service disruption is minimal.

georglauterbach (approved):

casperklein (approved):
👍  for the extended comments.

PR #2605, opened 2022-05-29
refactor: Migrate `SASL_PASSWD` ENV support

polarathene:
# Description

I would like to merge this for `git blame` history, but otherwise see `SASL_PASSWD` being removed ~~shortly after~~ in next major.

See the [prior PR](https://github.com/docker-mailserver/docker-mailserver/pull/2604) this one builds off of for Relay support feature history if additional context is needed. These changes were originally part of the prior PR but seemed more suited for a separate review/discussion.

### Changes:

- `SASL_PASSWD` was an ENV introduced for the initial and very basic relay-host support offered back in 2016.
- It is a legacy var that I assume is not in usage anymore. Documentation on the ENV isn't helpful (_it's value should be a valid line for `/etc/postfix/sasl_passwd` config_).
- It has been moved into `helpers/relay.sh` which is where it belongs.
- Tests using the ENV aren't actually using it (_probably copy/paste from original test that didn't consider if it was necessary_).
- I believe it should be deprecated, or simply dropped in the next major release. I don't really see it serving any purpose that other relay ENV don't already cover, especially due to it's history and how relay-host support progressed.

<details>
<summary>Commit message for reference:</summary>

- `helpers/sasl.sh` was to support an earlier ENV for SASL auth support. When extracting logic into individual helpers, it was assumed this was separate from relay support, which it appears was not the case.
- The `SASL_PASSWD` ENV is specified in tests but no longer used. There is no `external-domain.com` relay configured or tested against anywhere in the project.
- The ENV was likely used in tests prior to improved relay support that allowed for adding more than a single set of relay credentials.
- It likewise has no real relevance anywhere else outside of `relay.sh` as it's the only portion of code to operate with it.
- It's only relevant for SASL auth as an SMTP client, not the SMTP server (`smtpd`) SASL support that is delegated to Dovecot. Functionality has been completely migrated into `relay.sh` as a result.
- Documentation is poor for this ENV, it is unlikely in wide use? Should consider for removal.
- The ENV has been dependent upon `RELAY_HOST` to actually enable postfix to use `/etc/postfix/sasl_passwd`, thus not likely relevant in existing setups?
- Migrate `/etc/postfix/sasl_passwd` check from `tests.bats` as it belongs to relay tests.

</details>

## Type of change

- [x] Breaking change (Potentially)

## Checklist:


georglauterbach (approved):
👍🏼

casperklein (approved):

PR #2606, opened 2022-05-29
tests(fix): Adjust for local testing conditions

polarathene:
# Description

When running some tests locally, there were several failures unrelated to the feature being tested. These were resolved and covered by this PR.

- Tests were run in a local VM where a few failed due to timeouts but would have been successful otherwise. Needed about 30 seconds (instead of current 20), but have increased to 60 seconds. The affected tests continue as soon as success, only failure would take longer due to increased timeout.
- `--hostname` + `--domainname` with both as multi-label was causing `helpers/dns.sh` to fail at container startup. `--domainname` isn't necessary for the affected container under test. Removed to avoid panic. Possible cause is CI running via Ubuntu 20.04 which may run an older Docker daemon, or some other environment condition causing `hostname -f` to behave this way. Reminder `--domainname` is for setting NIS domain, not DNS domain, so this should be a non-issue.
- `setup.sh` would fail in `test/tests.bats` when the `mail` container name was not targeted for the command, if an existing `docker-mailserver` container was already running separate from the tests (or lingering from a failed test that didn't clean it up).
- View diffs with white-space ignored, or individual commits as one commit normalizes white-space (_I got annoyed by the mixed tab/spaces indentation and finally addressed it_). I can split this out into a separate PR if preferred.

## Type of change

## Checklist:

### DIFF
--- a/test/default_relay_host.bats
+++ b/test/default_relay_host.bats
@@ -1,15 +1,17 @@
 load 'test_helper/common'
 
 function setup() {
-	local PRIVATE_CONFIG
- 	PRIVATE_CONFIG=$(duplicate_config_for_container relay-hosts)
-    docker run -d --name mail_with_default_relay \
-		-v "${PRIVATE_CONFIG}":/tmp/docker-mailserver \
-		-v "$(pwd)/test/test-files":/tmp/docker-mailserver-test:ro \
-		-e DEFAULT_RELAY_HOST=default.relay.host.invalid:25 \
-		--cap-add=SYS_PTRACE \
-		-e PERMIT_DOCKER=host \
-		-h mail.my-domain.com -t "${NAME}"
+  local PRIVATE_CONFIG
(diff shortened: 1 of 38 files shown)


georglauterbach (approved):
Big diff, a lot of formatting - thank your for doing this, very much appreciated!!!

I "quickly" had a look, if you want me to look into some sections in more detail, tell me where I can find them and I will look again, but LGTM 👍🏼

casperklein (approved):
LGTM.

For the whitespace changes: I only had a quick look and did not explicitly check each line.

polarathene:
> if you want me to look into some sections in more detail

Everything worth noting was laid out in the PR description :)

In this case it's easier to review the individual commits to avoid noise of white-space housekeeping. Not really anything needing special attention though AFAIK :)

For the white-space commit, clicking the cog on the diff view you can disable white-space diffs which will ignore displaying the indentation changes, only the blank lines and few minor edits (a comment, collapsing a multi-line, etc) would be visible then.

Issue #2601, opened 2022-05-20
relaying: mails are not relayed with implicit tls (465) / improve scripts & configuration


edge90:
### Miscellaneous first checks

- [X] I checked that all ports are open and not blocked by my ISP / hosting provider.
- [X] I know that SSL errors are likely the result of a wrong setup on the user side and not caused by DMS itself. I'm confident my setup is correct.

### Affected Component(s)

Mails are not relayed with implicit tls (465)

### What happened and when does this occur?

```Markdown
DEFAULT_RELAY_HOST=[relay.smtp.isp.tld]:465
RELAY_HOST=relay.smtp.isp.tld
RELAY_PORT=465
RELAY_USER=MYUSERNAME
RELAY_PASSWORD=MYCOMPLICATEDPASSWORD
```

### What did you expect to happen?

```Markdown
Mail should be forwarded through the relay.
```

### How do we replicate the issue?

```Markdown
1. mailserver.env:

DEFAULT_RELAY_HOST=[relay.smtp.isp.tld]:465
RELAY_HOST=relay.smtp.isp.tld
RELAY_PORT=465
RELAY_USER=MYUSERNAME
RELAY_PASSWORD=MYCOMPLICATEDPASSWORD
```
2. Will complain about `smtp_tls_wrappermode = yes` and `smtp_tls_security_level = encrypt`.
Added the following to `docker-data/dms/config/postfix-main.cf`
```
smtp_tls_wrappermode = yes
smtp_tls_security_level = encrypt
```
3. Sendmail from the cli works.
sendmail -F "Firstname surname" -f "firsname@domain.tld" -t testmail@gmail.com
Subject: Hello
World
4. Sending from a smtp client does not.
```
warning: smtp_tls_wrappermode requires "smtp_tls_security_level = encrypt" (or stronger)
to=<....>, relay=none, delay=0.33, delays=0.33/0/0/0, dsn=4.3.0, status=deferred (server unavailable or unable to receive mail)
```

### DMS version

v11.0.0

### What operating system is DMS running on?

Linux

### What instruction set architecture is DMS running on?

x86_64 / AMD64

### What container orchestration tool are you using?

Docker Compose

### docker-compose.yml

```yml
services:
  mailserver:
    image: docker.io/mailserver/docker-mailserver:latest
    container_name: mailserver
    # If the FQDN for your mail-server is only two labels (eg: example.com),
    # you can assign this entirely to `hostname` and remove `domainname`.
[...]

PR #2607, opened 2022-05-29
fix: Opt-out of implicit TLS for Amavis SMTP client transport

polarathene:
# Description

Explicitly set `smtp_tls_wrappermode=no` for `master.cf:smtp-amavis` so that it does not inherit this setting from `main.cf` when updated for relay-host support, which the configured Amavis transport does not support (_at least with `smtp_tls_security_level=none`_).

Additionally adds service names for better visibility of amavis specific `master.cf` client/server logs.

This change enables supporting relay-host connections over implicit TLS (port 465).

<details>
<summary>Commit message (verbose description)</summary>

Presently relay-host support modifies `main.cf` settings directly. This adjusts the default transport (`smtp`) which other transports in `master.cf` inherit from.

When configuring for implicit TLS to a `relay-host` this would set `main.cf:smtp_tls_wrappermode = yes` and affect the transport `master.cf:smtp-amavis` which does not set an override like it does for `smtp_tls_security_level`. This causes Amavis to fail working which the default transport relies on due to `main.cf:content_filter`.

Easy fix, by explicitly adding the override `smtp_tls_wrappermode=no`.`no` is default in `main.cf` so inheriting this setting hasn't been a problem in the past for other relay-hosts using StartTLS.

</details>

---

Fixes: #2601 

This is a partial fix for supporting implicit TLS with relay-hosts. The user still needs to configure `main.cf` to enable `smtp_tls_wrappermode = yes` with `postfix-main.cf` or `user-patches.sh`. We do not document this, and future improvements to relay support are planned to simplify support.

## Type of change

## Checklist:


georglauterbach (approved):
👍🏼

casperklein (approved):

PR #2608, opened 2022-05-29
fix: Only add state of enabled services for volume mount

### LINKED ISSUE #2602: [FR]: `misc-stack.sh` should append directories only when relevant services are enabled
# Feature Request

## Context

Despite some features like CLAM AV being disabled by default, we do copy over that data anyway during startup if it doesn't exist:

https://github.com/docker-mailserver/docker-mailserver/blob/0d30b92a83118e7c1cea7cdc40007a3f41d4fe6c/target/scripts/startup/misc-stack.sh#L20-L38

During a test setup with 3 instances of DMS, this had each instance use 220 MB of storage just for ClamAV which I had not enabled, so it was unexpected.

Yes, but isn't likely to be an issue for most users / deployments. It only matters if you're low on disk space and have multiple containers running for some reason and choose to persist some state with a `/var/mail-state/` volume bind mount.

Not persisting / copying over state that isn't relevant.

Maybe, otherwise anyone is welcome to take a shot before I get around to it :sweat_smile: 

I am raising the issue as a personal reminder and awareness for which could be a small improvement, instead of burying it in my own todo list :sweat_smile:

polarathene:
# Description

These services will no longer copy over state unless they are enabled.

The biggest offender here was ClamAV as it's database that is baked into the docker image is over 200MB and would copy over to every container instance with a volume mounted state directory, despite having the service disabled.

Fixes #2602

## Type of change

## Checklist:


georglauterbach (dismissed):
LGTM 👍🏼

target/scripts/startup/misc-stack.sh
```diff
@@ -24,14 +24,17 @@ function _misc_save_states
     FILES=(
       spool/postfix
       lib/postfix
-      lib/amavis
-      lib/clamav
-      lib/spamassassin
-      lib/fail2ban
-      lib/postgrey
       lib/dovecot
```
casperklein, line 27: Dovecot can also be conditional.

`$SMTP_ONLY -ne 1`

georglauterbach (approved):

casperklein (approved):

PR #2615, opened 2022-06-04
refactor: Revise `check-for-changes.sh`

polarathene:
# Description

This PR is mostly housekeeping for `check-for-changes.sh`, in preparation for further refactoring.

Changes have been staged via individual commits with messages for further context. It may be easier to follow the changes via commit diffs, than as a whole.

### Change Overview:

- Inline docs for `check-for-changes.sh` have been shuffled around and revised a bit.
- Change processing extracted from the main change detection loop method to their own methods:
   - `_get_changed_files()` - I extracted the `grep`+`sed` line to it's own method.
      - This was due to the verbosity of inline docs I added. I value being able to under to know quickly at a glance what I'm looking at without having to go refresh on the syntax :sweat_smile: - However if other maintainers feel this is a bad idea it can be reverted, no worries!
   - `_postfix_dovecot_changes()` - The bulk of change processing was moved to this method. I've added conditionals to only run relevant logic.
   - `_ssl_changes()` - Just shifted, no logic changed. `REGEX_NEVER_MATCH` and `ACME_CERT_DIR` vars scope set to `local`.

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/check-for-changes.sh
+++ b/target/scripts/check-for-changes.sh
@@ -125,6 +125,22 @@ function _check_for_changes
   mv "${CHKSUM_FILE}.new" "${CHKSUM_FILE}"
 }
 
+function _get_changed_files
+{
+
+}
+
+
+function _postfix_dovecot_changes
+{
+
+}
+
+function _ssl_changes
+{
+
+}
+
 while true
 do
   _check_for_changes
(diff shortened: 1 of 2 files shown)


polarathene:
The change to process some changes only when the relevant files changed needs some more work.

Looks like `/etc/postfix/vhost` is missing an entry from `postfix-virtual.cf`, since `tests.bats` adds an account, but didn't modify virtual aliases, so `/etc/postfix/vhost` rebuilt without those added..

Instead I'll resume work on this PR after some others are merged, so that I can extract the `vhost` building logic from `helpers/accounts.sh` + `helpers/aliases.sh` which `bin/open-dkim` appears to duplicate.

georglauterbach (approved):
LGTM 👍🏼

casperklein (approved):

PR #2616, opened 2022-06-04
fix: Amavis should reload config for `/etc/postfix/vhost` updates

polarathene:
# Description

When `/etc/postfix/vhost` is updated from change detection, Amavis did not refresh that source, and thus would not be aware of any new domains, ignoring them for spam/virus checking.

A test could probably be added, my time is focused on prepping PRs with the time I have available right now.

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/check-for-changes.sh
+++ b/target/scripts/check-for-changes.sh
@@ -59,6 +59,8 @@ function _check_for_changes
     # While some config changes may be properly applied by Postfix or Dovecot
     # via their 'reload' commands; some may require restarting?:
     _log_with_date 'debug' 'Restarting services due to detected changes'
+
+    _reload_amavis
     supervisorctl restart postfix
     [[ ${SMTP_ONLY} -ne 1 ]] && supervisorctl restart dovecot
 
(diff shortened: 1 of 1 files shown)


target/scripts/check-for-changes.sh
```diff
@@ -86,6 +88,16 @@ function _get_changed_files
   grep -Fxvf "${CHKSUM_CURRENT}" "${CHKSUM_NEW}" | sed -r 's/^\S+[[:space:]]+//'
 }
 
+function _reload_amavis
+{
+  if [[ ${CHANGED} =~ ${DMS_DIR}/postfix-accounts.cf ]] || [[ ${CHANGED} =~ ${DMS_DIR}/postfix-virtual.cf ]]
+  then
+    # /etc/postfix/vhost was updated, amavis must refresh it's config by
+    # reading this file again in case of new domains, otherwise they will be ignored.
+    amavisd-new reload
```
polarathene, line 97: Presumably `supervisorctl restart amavis` should be used.

I've [not received any feedback about better understanding why `postfix reload` was an issue in the past](https://github.com/docker-mailserver/docker-mailserver/pull/2624#issuecomment-1148043165) (_may have been due to the present `postfix-wrapper.sh` script?_).

georglauterbach: Truth be told, I don‘t know 🙈 @casperklein do you have an idea?:)

target/scripts/check-for-changes.sh
```diff
@@ -59,6 +59,8 @@ function _check_for_changes
     # While some config changes may be properly applied by Postfix or Dovecot
     # via their 'reload' commands; some may require restarting?:
     _log_with_date 'debug' 'Restarting services due to detected changes'
+
+    _reload_amavis
```
polarathene, line 63: Only do so if Amavis service is active:

```suggestion
    [[ ${ENABLE_AMAVIS} -eq 1 ]] && _reload_amavis
```

target/scripts/check-for-changes.sh
casperklein: The usage of `amavisd-new reload` should be fine. However, if for some reason a complete service restart is needed, `supervisorctl restart amavis` should be used.

casperklein (approved):

georglauterbach (approved):

PR #2617, opened 2022-06-04
fix(listmailuser): Don't parse comments when reading config

polarathene:
# Description

The quota support method `dovecot_quota_to_hr()` was failing when provided comment lines. Fixed by stripping them from the config used as input for the while loop.

---

~~Not sure how useful a test is for this? (_we do have some tests that detect if other commands work without parsing failures though_)~~ Probably don't need a test for this as failure impact is UX only? This provides basis of consistent/improved config parsing to avoid issue in general going forward.

<details>
<summary>Resolved concern</summary>

I haven't looked over the difference for the shell syntax for passing the input data to the while loop, but it seems the 2nd commit syntax looks more appropriate? I'm just referencing usage elsewhere in the project, so happy for @georglauterbach or @casperklein to weigh in on what is suitable :+1: 

```bash
done <<< "$(_config_without_comments "${DATABASE}")"
# vs (latest commit):
done < <(_config_without_comments "${DATABASE}")
```

`sed` usage variants (for `_config_without_comments()`):

```bash
sed -n -r '/^\s*[^#[:space:]]/p' "${1}"
# vs (latest commit):
sed '/^[[:space:]]*#/d' "${1}"

# another usage in project is (not sure what the '|| true' is for):
grep -v "^\s*$\|^\s*\#" || true
# and another:
grep -vE '^(\s*$|#)'
```

**EDIT:** Whoops, noticed that the patterns aren't equivalent matching (empty line, white-space only, comments) :sweat_smile: 

</details>

## Type of change

## Checklist:

- [ ] ~~If necessary I have added tests that prove my fix is effective or that my feature works~~


georglauterbach:
Alright, I'd definitely go with

``` BASH
done < <(_config_without_comments "${DATABASE}")
```

here for the other two, I don't really mind, but my preference is this:

``` BASH
sed '/^[[:space:]]*#/d' "${1}"
# and
grep -vE '^(\s*$|#)'
```

The `||` true at the end of the `grep` command is used in case: `grep` does not find a matching patter and returns with non-zero exit code but this is not desired so we run `true` to get a "zero" exit code.

georglauterbach (dismissed):

polarathene:
I have a follow-up PR that uses this line filter method, I'll add another commit here to move it to `helpers/utils.sh`.

> The `||` true at the end of the `grep` command is used in case: `grep` does not find a matching patter and returns with non-zero exit code but this is not desired so we run `true` to get a "zero" exit code.

Oh I see, `true` is a command, for some reason I thought it was like `'true'` (string) :sweat_smile: 

Is it intended for the future scripts where you wanted to have scripts bail on non-zero exit codes while running? It doesn't have anything to do with the actual loop logic right? Is it better to move that into the method?

https://github.com/docker-mailserver/docker-mailserver/blob/4e3e16189c1334ae91032713c90b08db119b0e17/target/scripts/helpers/aliases.sh#L26-L33

This is a case where it's fine to replace with a common filter function? I have it and others I found ready to push as a separate PR.

---

> here for the other two, I don't really mind, but my preference is this

In some naive testing, the `sed` option isn't going to output and exit code based on if any matches were found and operated on. So I guess `grep` is more appropriate (in this case it's pattern also matches all 3 conditions to filter out).

Instead of `-vE` I could make use of the long name options so they're self-documenting? (I won't remember `-v` is invert, verbosity comes to mind first if I haven't been using `grep` for a while):

```bash
grep --extended-regexp --invert-match "^\s*$|^\s*#" "${1}"
```

`helpers/utils.sh:_is_comment` is presently only used by `relay.sh`. We won't need it after this, `relay.sh` can switch to filtering the whole file input instead of checking each line :+1:

polarathene:
I need this PR and it's follow-up merged before I can continue https://github.com/docker-mailserver/docker-mailserver/pull/2615

I'm hoping another maintainer will have some time soon to spare, otherwise it's going to add friction to getting these PRs ready while I have time to spare :sweat_smile:

georglauterbach (approved):
👍🏼

target/scripts/helpers/utils.sh
```diff
@@ -5,8 +5,16 @@ function _escape
   echo "${1//./\\.}"
 }
 
-# Check if string input is an empty line, only whitespaces
-# or `#` as the first non-whitespace character.
+# Returns input after filtering out lines that are:
+# empty, white-space, comments (`#` as the first non-whitespace character)
+function _filter_to_valid_lines
```
casperklein, line 10: Maybe one of these fits better?

```suggestion
function _get_valid_lines
```

```suggestion
function _get_valid_lines_from_file
```

casperklein (approved):

target/scripts/helpers/utils.sh
polarathene: Sure, I'll make that change in the follow-up PR :+1:

PR #2619, opened 2022-06-05
chore(housekeeping): Normalize how config files filter out unwanted lines

polarathene:
# Description

We have used various ways to filter out unwanted lines (empty, white-space only, comments) from configs when reading them to iterate through. In some cases, we had not filtered for all kinds of unwanted lines, or did not filter at all (even though we probably should have). 

This PR has normalized that approach to call a helper method instead (added in  https://github.com/docker-mailserver/docker-mailserver/pull/2617), with a common approach to call the filter method applied:
- The changes are mostly converting config file paths into local vars to reference instead. These are staged out into commits for easy review if needed (no commit messages).
- Then normalizes the config file pre-processing step to use the recently added method `_get_valid_lines_from_file()` (renamed in this PR from earlier `_filter_to_valid_lines()` due to feedback in [#2617](https://github.com/docker-mailserver/docker-mailserver/pull/2617) ).
- `helpers/relay.sh` had a loop that only seems to strip out the unwanted lines for the target config. This doesn't seem necessary and could probably just be a `cp` instead? I collapsed that loop to call the filtering method and just write the output of that instead.
- `helpers/utils.sh:_is_comment()` has been removed, as only `relay.sh` was using it and it no longer seems to serve any purpose.

### Relevant history

Looks like initial filtering to avoid issues was [introduced via this PR](https://github.com/docker-mailserver/docker-mailserver/pull/1494) in May 2020.

## Type of change

## Checklist:

### DIFF
--- a/target/bin/listalias
+++ b/target/bin/listalias
@@ -8,5 +8,5 @@ DATABASE='/tmp/docker-mailserver/postfix-virtual.cf'
 [[ -f ${DATABASE} ]] || _exit_with_error "No 'postfix-virtual.cf' file"
 [[ -s ${DATABASE} ]] || _exit_with_error "Empty 'postfix-virtual.cf' - no aliases have been added"
 
-grep -v "^\s*$\|^\s*\#" "${DATABASE}"
+_get_valid_lines_from_file "${DATABASE}"
 exit 0
--- a/target/bin/listdovecotmasteruser
+++ b/target/bin/listdovecotmasteruser
@@ -16,6 +16,6 @@ while read -r LINE
 do
   USER=$(echo "${LINE}" | cut -d'|' -f1)
   echo "* ${USER}"
-done < "${DATABASE}"
+done < <(_get_valid_lines_from_file "${DATABASE}")
 
 exit 0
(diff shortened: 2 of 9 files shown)


polarathene:
Current test failure is unrelated to this PR. I can reproduce consistently on my local VM, it's not consistent in CI. It also affected https://github.com/docker-mailserver/docker-mailserver/pull/2611 (master branch merge commit).

Quite possibly related to what happens for some tests that have been disabled for being non-deterministic in the past, since this bug occurs when multiple DMS containers are running and the first container started gets used instead of the expected one.

```
not ok 141 checking relay hosts: default mapping is added from env vars for new virtual user entry in 11425ms
# (from function `run_until_success_or_timeout' in file test/test_helper/common.bash, line 67,
#  in test file test/mail_with_relays.bats, line 52)
#   `run_until_success_or_timeout 10 docker exec mail_with_relays grep -e domain2.tld /etc/postfix/relayhost_map' failed
# Timed out on command: docker exec mail_with_relays grep -e domain2.tld /etc/postfix/relayhost_map
```

Fixed via [this PR](https://github.com/docker-mailserver/docker-mailserver/pull/2622).

casperklein (approved):

georglauterbach (approved):

PR #2620, opened 2022-06-05
refactor: Share a common helper (vhost builder) for sourcing domains

polarathene:
# Description

`bin/open-dkim` and `helpers/accounts.sh` + `helpers/aliases.sh` have roughly duplicate code for collecting domains to process. This code is now in sync by updating methods in `helpers/postfix.sh`.

<details>
<summary>Commit messages (first to last) for quick reference</summary>

**chore: Split vhost helper method and use filepath vars**
- Helpers `accounts.sh` and `aliases.sh` can move their vhost code into this helper.
- They share duplicate code with `bin/open-dkim` which will also leverage this vhost helper going forward.

**chore: Sync vhost generation logic into helper**
- Chunky commit, but mostly copy/paste of logic into a common method.
- `bin/open-dkim` additionally wrapped relevant logic in a function call and revised inline docs.

**chore: Include LDAP vhost support**
- Revises notes for LDAP vhost support.
- This now ensures LDAP users get vhost rebuilt to match the startup script for when change detection support is enabled.
- `bin/open-dkim` will additionally be able to support the default `DOMAINNAME` var (set via `helpers/dns.sh`) for LDAP users instead of requiring them to provide one explicitly.

**chore(`bin/open-dkim`): Ensure `DOMAINNAME` is properly set**
- This will ensure LDAP users insert the same `DOMAINNAME` value as used during container startup.
- The container itself should panic at startup (during `helpers/dns.sh`) if this isn't configured correctly already, thus it should not introduce any breaking change to users of this utility?

</details>

---

- We use this functionality for populating `/etc/postfix/vhost` used by Postfix `main.cf:virtual_mailbox_domains` setting. `bin/open-dkim` leverages it to acquire domains relevant to Postfix.
- This PR will additionally benefit `check-for-changes.sh` to generate vhost config when `helpers/accounts.sh` or `helpers/aliases.sh` is not affected by the changed file. That support will be covered in https://github.com/docker-mailserver/docker-mailserver/pull/2615

### Relevant history - `bin/open-dkim` arg `DOMAINS`

Additional info regarding `bin/open-dkim` (most of this isn't useful in the context of this PR, but helps understand the background of the alternative support via `DOMAINS` arg):

- It's alternative [`DOMAINS` arg support was added in Jan 2021](https://github.com/docker-mailserver/docker-mailserver/pull/1753) to support the needs of LDAP users ([earlier issue from 2017 for LDAP users](https://github.com/docker-mailserver/docker-mailserver/issues/478#issuecomment-273426172)).
   - The contributor [shares in their original issue](https://github.com/docker-mailserver/docker-mailserver/issues/1735#issuecomment-753459953) a command for their setup to retrieve the domains that they'd supply as `DOMAINS`.
   - The `DOMAINS` arg replaced the need for a separate command that provided similar functionality ([originally for a user with DKIM sourced via SQL](https://github.com/docker-mailserver/docker-mailserver/issues/619), implemented Sep 2017 in https://github.com/docker-mailserver/docker-mailserver/pull/690 ).
     -  #619 notes the intention was to configure DKIM for equivalent domains available in `main.cf:virtual_mailbox_domains` (`/etc/postfix/vhost`), although LDAP setups also add a 2nd table value: `ldap:/etc/postfix/ldap-domains.cf` (_could possible source via `postconf` query? Unless `bin/open-dkim` is not meant to be dependent upon a running/configured Postfix instance_).
   - That PR introduced an unrelated change to the default DKIM keylength size which was bumped up from 2048-bit to 4096-bit ([with little justification](https://github.com/docker-mailserver/docker-mailserver/pull/1753#pullrequestreview-576855284)).
     - AFAIK 4096-bit RSA is excessive for DKIM, it's not for encryption and not damaging to past transactions? You want some strength to avoid forgery, but it's not as valuable as a root CA key.
     - Prior to that, [a March 2018 PR which added support to adjust keylength size](https://github.com/docker-mailserver/docker-mailserver/pull/868) references two issues about services lacking support for key lengths larger than 1024-bit.
       - In 2022, [OpenDKIM docs still document 1024-bit as the default size](http://www.opendkim.org/opendkim-genkey.8.html). 
       - Debian apparently [has defaulted to 2048-bit since 2017](https://github.com/docker-mailserver/docker-mailserver/issues/855#issuecomment-368994889).
       - The NameCheap provider [has since updated support to 2048-bit](https://www.namecheap.com/support/knowledgebase/article.aspx/317/2237/how-do-i-add-txtspfdkimdmarc-records-for-my-domain), unclear if other vendors improved support.

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/helpers/postfix.sh
+++ b/target/scripts/helpers/postfix.sh
@@ -51,7 +51,7 @@ function _vhost_collect_postfix_domains
   if [[ -f ${DATABASE_ACCOUNTS} ]]
   then
     # shellcheck disable=SC2034
-    while IFS=$'|' read -r LOGIN PASS
+    while IFS=$'|' read -r LOGIN _
     do
       DOMAIN=$(echo "${LOGIN}" | cut -d @ -f2)
       echo "${DOMAIN}" >>"${TMP_VHOST}"
@@ -61,9 +61,8 @@ function _vhost_collect_postfix_domains
   # getting domains FROM mail aliases
   if [[ -f ${DATABASE_VIRTUAL} ]]
   then
-    # the `to` is important, don't delete it
     # shellcheck disable=SC2034
-    while read -r FROM TO
+    while read -r FROM _
     do
       UNAME=$(echo "${FROM}" | cut -d @ -f1)
       DOMAIN=$(echo "${FROM}" | cut -d @ -f2)
--- a/target/scripts/helpers/postfix.sh
+++ b/target/scripts/helpers/postfix.sh
@@ -50,7 +50,6 @@ function _vhost_collect_postfix_domains
   # getting domains FROM mail accounts
   if [[ -f ${DATABASE_ACCOUNTS} ]]
   then
-    # shellcheck disable=SC2034
     while IFS=$'|' read -r LOGIN _
     do
       DOMAIN=$(echo "${LOGIN}" | cut -d @ -f2)
@@ -61,7 +60,6 @@ function _vhost_collect_postfix_domains
   # getting domains FROM mail aliases
   if [[ -f ${DATABASE_VIRTUAL} ]]
   then
-    # shellcheck disable=SC2034
     while read -r FROM _
     do
       UNAME=$(echo "${FROM}" | cut -d @ -f1)
--- a/target/scripts/helpers/postfix.sh
+++ b/target/scripts/helpers/postfix.sh
@@ -16,29 +16,63 @@
 # - `postmap` only seems relevant when the lookup type is one of these `file_type` values: http://www.postfix.org/postmap.1.html
 #   Should not be a concern for most types used by `docker-mailserver`: texthash, ldap, pcre, tcp, unionmap, unix.
 #   The only other type in use by `docker-mailserver` is the hash type for /etc/aliases, which `postalias` handles.
+
 function _create_postfix_vhost
 {
   # `main.cf` configures `virtual_mailbox_domains = /etc/postfix/vhost`
   # NOTE: Amavis also consumes this file.
   local DATABASE_VHOST='/etc/postfix/vhost'
-  local TMP_VHOST='/tmp/vhost.tmp'
+  local TMP_VHOST='/tmp/vhost.postfix.tmp'
 
+  _vhost_collect_postfix_domains
   _create_vhost
 }
 
-# Processes TMP_VHOST into final DATABASE_VHOST
+# Filter unique values into a proper DATABASE_VHOST config:
 function _create_vhost
 {
   : >"${DATABASE_VHOST}"
 
-  # Account and Alias generation will store values in `/tmp/vhost.tmp`.
-  # Filter unique values to the proper config.
   # NOTE: LDAP stores the domain value set by `docker-mailserver`,
   # and correctly removes it from `mydestination` in `main.cf` in `setup-stack.sh`.
   if [[ -f ${TMP_VHOST} ]]
   then
     sort < "${TMP_VHOST}" | uniq >>"${DATABASE_VHOST}"
     rm "${TMP_VHOST}"
+  fi
+}
+
+# Collects domains from configs (DATABASE_) into TMP_VHOST
+function _vhost_collect_postfix_domains
+{
+  local DATABASE_ACCOUNTS='/tmp/docker-mailserver/postfix-accounts.cf'
+  local DATABASE_VIRTUAL='/tmp/docker-mailserver/postfix-virtual.cf'
+  local DOMAIN UNAME
+
+  # getting domains FROM mail accounts
+  if [[ -f ${DATABASE_ACCOUNTS} ]]
+  then
+    # shellcheck disable=SC2034
+    while IFS=$'|' read -r LOGIN PASS
+    do
+      DOMAIN=$(echo "${LOGIN}" | cut -d @ -f2)
+      echo "${DOMAIN}" >>"${TMP_VHOST}"
+    done < <(_get_valid_lines_from_file "${DATABASE_ACCOUNTS}")
+  fi
+
+  # getting domains FROM mail aliases
+  if [[ -f ${DATABASE_VIRTUAL} ]]
+  then
+    # the `to` is important, don't delete it
+    # shellcheck disable=SC2034
+    while read -r FROM TO
+    do
+      UNAME=$(echo "${FROM}" | cut -d @ -f1)
+      DOMAIN=$(echo "${FROM}" | cut -d @ -f2)
+
+      # if they are equal it means the line looks like: "user1     other@domain.tld"
+      [[ ${UNAME} != "${DOMAIN}" ]] && echo "${DOMAIN}" >>"${TMP_VHOST}"
+    done < <(_get_valid_lines_from_file "${DATABASE_VIRTUAL}")
   fi
 }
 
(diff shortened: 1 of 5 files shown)


polarathene:
`helpers/aliases.sh` has a maintenance comment (_which I carried over into `helpers/postfix.sh`_):

```bash
# the `to` is important, don't delete it
```

Originally this `FROM TO` syntax was introduced in Oct 2015 PR](https://github.com/docker-mailserver/docker-mailserver/pull/26), but no explanation for `to` var usage. Later in April 2016, this was duplicated into opendkim support: [`bin/generate-dkim-config`](https://github.com/docker-mailserver/docker-mailserver/commit/4f611eec1db6dc2cf90161ec819ec1a14083d792#diff-c29b503958752c4c2507e499c4fd95ec5cff2f0af6bc71847b54cd06ce093c22R11), albeit no PR (direct commit by original project author).

The maintainer comment instructing the importance of `TO` was added by @georglauterbach in [Sep 2020 large refactor PR](https://github.com/docker-mailserver/docker-mailserver/pull/1605/files#diff-6d88c6c2abe074135b8efac1cf0dca97dd288a71720cdcd6ac48c2a6f85dcf08R976), no additional context in the PR discussion or commit messages.

Can we better clarify why it's important here?

polarathene:
Updated to include `/etc/postfix/vhost` handling for LDAP in the helper as well. `bin/open-dkim` can leverage this too when not using the explicit `DOMAINS` arg, it may be an improvement/fix for LDAP users.

It does require using `helpers/dns.sh` to ensure the `DOMAINNAME` var is set. This shouldn't be an issue as any potential panic would have already occurred during container startup AFAIK? So not a breaking change?

georglauterbach:
> `helpers/aliases.sh` has a maintenance comment (_which I carried over into `helpers/postfix.sh`_):
> 
> ```shell
> # the `to` is important, don't delete it
> ```
> 
> Originally this `FROM TO` syntax was introduced in an [Oct 2015 PR](https://github.com/docker-mailserver/docker-mailserver/pull/26), but no explanation for `to` var usage. Later in April 2016, this was duplicated into opendkim support: [`bin/generate-dkim-config`](https://github.com/docker-mailserver/docker-mailserver/commit/4f611eec1db6dc2cf90161ec819ec1a14083d792#diff-c29b503958752c4c2507e499c4fd95ec5cff2f0af6bc71847b54cd06ce093c22R11), albeit no PR (direct commit by original project author).
> 
> The maintainer comment instructing the importance of `TO` was added by @georglauterbach in [Sep 2020 large refactor PR](https://github.com/docker-mailserver/docker-mailserver/pull/1605/files#diff-6d88c6c2abe074135b8efac1cf0dca97dd288a71720cdcd6ac48c2a6f85dcf08R976), no additional context in the PR discussion or commit messages.
> 
> Can we better clarify why it's important here?

Yes, we can :) So, when I first removed it, it was due to `shellcheck` telling me that the variable is not needed because it is not used in the loop. This is correct, but the `read` command uses it to put "the rest of the current line" after the first whitespace into the `TO` variable. Now, when I think about it, maybe we can use `_` instead of `TO`, but I'd need to test this.

georglauterbach:
> Now, when I think about it, maybe we can use _ instead of TO, but I'd need to test this.

This looks promising. Since this PR is not merging a branch of this repository, I cannot commit the changes directly - I have appended a `.patch` file further down below that you can apply with `git apply`, you may just need to rename it and delete the `.txt` suffix.

[FROM-TO.patch.txt](https://github.com/docker-mailserver/docker-mailserver/files/8852392/FROM-TO.patch.txt)

target/scripts/helpers/postfix.sh
```diff
@@ -16,23 +16,76 @@
...
+{
+  local DATABASE_ACCOUNTS='/tmp/docker-mailserver/postfix-accounts.cf'
+  local DATABASE_VIRTUAL='/tmp/docker-mailserver/postfix-virtual.cf'
+  local DOMAIN UNAME
+
+  # getting domains FROM mail accounts
+  if [[ -f ${DATABASE_ACCOUNTS} ]]
+  then
+    # shellcheck disable=SC2034
+    while IFS=$'|' read -r LOGIN PASS
```
polarathene, line 54:
```suggestion
    while IFS=$'|' read -r LOGIN _
```

target/scripts/helpers/postfix.sh
```diff
@@ -16,23 +16,76 @@
...
+      echo "${DOMAIN}" >>"${TMP_VHOST}"
+    done < <(_get_valid_lines_from_file "${DATABASE_ACCOUNTS}")
+  fi
+
+  # getting domains FROM mail aliases
+  if [[ -f ${DATABASE_VIRTUAL} ]]
+  then
+    # the `to` is important, don't delete it
+    # shellcheck disable=SC2034
+    while read -r FROM TO
```
polarathene, line 66:
```suggestion
    while read -r FROM _
```

polarathene, line 66:
```suggestion
    while read -r FROM _
```

target/scripts/helpers/postfix.sh
```diff
@@ -16,23 +16,76 @@
...
+    do
+      DOMAIN=$(echo "${LOGIN}" | cut -d @ -f2)
+      echo "${DOMAIN}" >>"${TMP_VHOST}"
+    done < <(_get_valid_lines_from_file "${DATABASE_ACCOUNTS}")
+  fi
+
+  # getting domains FROM mail aliases
+  if [[ -f ${DATABASE_VIRTUAL} ]]
+  then
+    # the `to` is important, don't delete it
```
polarathene, line 64:
```suggestion
```

target/scripts/helpers/postfix.sh
```diff
@@ -16,23 +16,75 @@
...
+function _vhost_collect_postfix_domains
+{
+  local DATABASE_ACCOUNTS='/tmp/docker-mailserver/postfix-accounts.cf'
+  local DATABASE_VIRTUAL='/tmp/docker-mailserver/postfix-virtual.cf'
+  local DOMAIN UNAME
+
+  # getting domains FROM mail accounts
+  if [[ -f ${DATABASE_ACCOUNTS} ]]
+  then
+    # shellcheck disable=SC2034
```
georglauterbach, line 53: I think this line can be removed as well as the old `PASS` is no longer there.

target/scripts/helpers/postfix.sh
```diff
@@ -61,9 +61,8 @@ function _vhost_collect_postfix_domains
   # getting domains FROM mail aliases
   if [[ -f ${DATABASE_VIRTUAL} ]]
   then
-    # the `to` is important, don't delete it
     # shellcheck disable=SC2034
```
georglauterbach, line 64: This line could be removed as well if I'm not mistaken.

target/scripts/helpers/postfix.sh
polarathene, line 53:
```suggestion
```

target/scripts/helpers/postfix.sh
```diff
@@ -16,23 +16,75 @@
...
+    do
+      DOMAIN=$(echo "${LOGIN}" | cut -d @ -f2)
+      echo "${DOMAIN}" >>"${TMP_VHOST}"
+    done < <(_get_valid_lines_from_file "${DATABASE_ACCOUNTS}")
+  fi
+
+  # getting domains FROM mail aliases
+  if [[ -f ${DATABASE_VIRTUAL} ]]
+  then
+    # shellcheck disable=SC2034
```
polarathene, line 64:
```suggestion
```

polarathene (commented):
Good catch with those two lint lines! :grinning:

georglauterbach (approved):

casperklein (approved):

PR #2622, opened 2022-06-06
fix(setup.sh): Do not override an existing `CONTAINER_NAME` value

polarathene:
# Description

This bug was causing `setup.sh -c` to target the wrong container when an earlier `docker-mailserver` container was already running.

I believe this is to try lookup an implicit default, but an explicit value for `-c` sets `CONTAINER_NAME` prior to this and should be used. It was causing some recent test failures in `test/mail_with_relays.bats`, both on CI and my local VM.

## Type of change

## Checklist:


georglauterbach (approved):
Nice observation!

casperklein (approved):

PR #1065, opened 2018-10-28
Configuration support for /etc/aliases, see issue #988

peter-hartmann:
This is work in progress (sorry perhaps I should create this when done, still learning a bit here)
[x] add postfix-aliases.cf to ./config (/tmp/docker-mailserver)
[x] in check-for-changes.sh monitor postfix-aliases.cf in and apply changes to /etc/aliases
[x] in start-mailserver.sh apply postfix-aliases.cf to /etc/aliases
This solution addresses issue #988 .

PR #2623, opened 2022-06-06
chore(`check-for-changes.sh`): Drop redundant guards

polarathene:
# Description

Housekeeping PR mostly, added a few extra configs that should be monitored for changes.

I looked into some conditional guards and why they were needed. Turns out they're not, most of it was introduced from early support of Change Detection service and no longer relevant.

Origin of conditional guards:

- `check-for-changes.sh` originally used relative paths to supported config files to monitor. This and checksum file creation relied on changing the working directory to `/tmp/docker-mailserver`. No longer necessary as we use absolute paths.
- This carried over to associated scripts to only run logic if the directory existed (_such as [here in a Aug 2019 commit](https://github.com/docker-mailserver/docker-mailserver/commit/7f3e5a22e18e091e2eeabaa47a18a854788e7db8)_). We support monitoring changes elsewhere now (certs), the guard is legacy and can be dropped.
   - The [full PR that commit belongs to](https://github.com/docker-mailserver/docker-mailserver/pull/1201) is rather large, but useful discussion regarding transition to single checksum file and file locking.
- The `postfix-accounts.cf` guard was introduced as part of the [original `check-for-changes.sh` feature PR in Oct 2017](https://github.com/docker-mailserver/docker-mailserver/pull/738), shortly followed up with a [fix to the guard to `exit` if the config file didn't exist](https://github.com/docker-mailserver/docker-mailserver/pull/748) to prevent log spam.
- `postfix-aliases.cf` was always generated, for what seemed to be a compatibility workaround for `check-for-changes.sh` support. That logic was originally introduced in a [Nov 2018 PR](https://github.com/docker-mailserver/docker-mailserver/pull/1065). That PR did not introduce the original alias support requiring `POSTMASTER_ADDRESS`, but did carry over the dependency for it into `check-for-changes.sh`. As `aliases.sh` has since de-duplicated the alias code and `check-for-changes.sh` is in much better shape, this workaround is no longer necessary.

---

I've additionally added 3 user configs that make sense for change detection to support as well since we already run recreation logic involving them: `postfix-sasl-password.cf`, `postfix-relaymap.cf`, `postfix-regexp.cf`.

All these `/tmp/docker-mailserver` config files are wrapped into the directory check before adding them to `STAGING_FILES`, although with `shopt -s nullglob` and the filtering to `CHANGED_FILES` array, it probably makes no difference and the conditional could be dropped?

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/check-for-changes.sh
+++ b/target/scripts/check-for-changes.sh
@@ -27,12 +27,6 @@ then
   _exit_with_error "Could not change into '/tmp/docker-mailserver/' directory" 0
 fi
 
-# check postfix-accounts.cf exist else break
-if [[ ! -f postfix-accounts.cf ]]
-then
-  _exit_with_error "'/tmp/docker-mailserver/postfix-accounts.cf' is missing" 0
-fi
-
 # verify checksum file exists; must be prepared by start-mailserver.sh
 if [[ ! -f ${CHKSUM_FILE} ]]
 then
--- a/target/scripts/check-for-changes.sh
+++ b/target/scripts/check-for-changes.sh
@@ -22,11 +22,6 @@ _log_with_date 'debug' 'Starting changedetector'
 #            to be properly set.
 _obtain_hostname_and_domainname
 
-if ! cd /tmp/docker-mailserver &>/dev/null
-then
-  _exit_with_error "Could not change into '/tmp/docker-mailserver/' directory" 0
-fi
-
 # verify checksum file exists; must be prepared by start-mailserver.sh
 if [[ ! -f ${CHKSUM_FILE} ]]
 then
--- a/target/scripts/helpers/aliases.sh
+++ b/target/scripts/helpers/aliases.sh
@@ -61,13 +61,8 @@ function _handle_postfix_aliases_config
 
   echo "root: ${POSTMASTER_ADDRESS}" >/etc/aliases
 
-  if [[ -f /tmp/docker-mailserver/postfix-aliases.cf ]]
-  then
-    cat /tmp/docker-mailserver/postfix-aliases.cf >>/etc/aliases
-  else
-    _log 'trace' "'/tmp/docker-mailserver/postfix-aliases.cf' is not provided, it will be auto created."
-    : >/tmp/docker-mailserver/postfix-aliases.cf
-  fi
+  local DATABASE_ALIASES='/tmp/docker-mailserver/postfix-aliases.cf'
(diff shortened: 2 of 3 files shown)


georglauterbach (dismissed):
👍🏼

casperklein (dismissed):

georglauterbach:
> `test/tests.bats - "checking restart of process: postfix"`:
> 
> ```shell
> @test "checking restart of process: postfix" {
>   run docker exec mail /bin/bash -c "pkill master && sleep 10 && ps aux --forest | grep -v grep | grep '/usr/lib/postfix/sbin/master'"
>   assert_success
> }
> ```

The test command looks awfully off... `grep -v grep` is used to "filter out" the grep command itself, which is a very interesting decision.

> Not quite sure why that's happening with these changes tbh. I cannot seem to reliably reproduce this failure. I'll re-run the tests.

I do not see a reason for the failure as well ATM...

target/scripts/check-for-changes.sh
```diff
@@ -22,17 +22,6 @@ _log_with_date 'debug' 'Starting changedetector'
 #            to be properly set.
 _obtain_hostname_and_domainname
 
-if ! cd /tmp/docker-mailserver &>/dev/null
-then
-  _exit_with_error "Could not change into '/tmp/docker-mailserver/' directory" 0
-fi
```
georglauterbach, line 28: Where is the `cd` done instead? I was at first not really thinking about it, but now, I cannot see where the directory change is happening instead :D

georglauterbach: I don't know either, but maybe there is some logic that we don't currently know of that depends on being in the correct directory...

polarathene: It's not performed as it shouldn't need to be anymore due to everything using absolute paths?

I oddly cannot reproduce the failure on my local VM instance for the branch either. It might be something with the environment? But it's not clear why the changes would be breaking that particular test?

I could revert changes in the PR until the CI tests pass I guess to find out what it's unhappy about?

polarathene:
Prior to the `supervisord` kill tests, just a little bit earlier is `setup.sh` relay command tests that manipulate `postfix-relaymap.cf`:

https://github.com/docker-mailserver/docker-mailserver/blob/62fdcb05f58c7252a163dafe153b16a9904b7ff4/test/tests.bats#L1131-L1166

As this is running in `test/tests.bats` each test is sharing the same container and running sequentially, so I suspect one of these earlier tests is impacting it.

georglauterbach:
Nice detective work :D

georglauterbach (approved):

Issue #1057, opened 2018-10-16
Why SYS_PTRACE capabilty ?


adlp:
I don't want to give too much capability ! :)

## Context
While reading the docker-compose.yml, I see the need of some capability, NET_ADMIN and SYS_PTRACE.
Ok for NET_ADMIN (needed for fail2ban) But why SYS_PTRACE ?

## Expected Behavior
A container with not too much capability :)

## Possible Fix
Only Documentation, and maybe an explanation of each capability ?

Thanks !

johansmitsnl:
@adlp it was added in this commit: https://github.com/tomav/docker-mailserver/commit/420e7741a16b1de617611f95aca73a1088214e59

jirislav:
> @adlp it was added in this commit: [420e774](https://github.com/tomav/docker-mailserver/commit/420e7741a16b1de617611f95aca73a1088214e59)

I've browsed through the commit and I've found **no reason** to allow the docker to have **PTRACE** capability. It was probably added because of testing.

The vulnerability exploited by `SYS_PTRACE` is the ability to skip the seccomp check as noted in [this paper](https://www.nccgroup.trust/globalassets/our-research/us/whitepapers/2016/june/container_whitepaper.pdf) (See the *ptrace(2) hole* section):

> The seccomp check will not be run again after the tracer is notified.  (This means that seccomp-based sandboxes MUST NOT allow use of ptrace, even of other sandboxed processes, without extreme care; ptracers can use this mechanism to escape.)

### IMO this is great security issue and the SYS_PTRACE should be disabled.

adlp:
@jirislav, yes it's exactly my opinion, but I haven't take the time to come back with a full test.
In one week I will be able to test it not before :/
Thanks a lot, I feel less crazy now :)

jirislav:
> @jirislav, yes it's exactly my opinion, but I haven't take the time to come back with a full test.
> In one week I will be able to test it not before :/

I'm already testing this on my production mailserver and I see no drawbacks yet.

Btw I wonder if you wanted to take a look at more hardened `docker-compose.yml` I've come up with and successfully tested on production environment. (Inspired by [CIS Docker Benchmark](https://blogs.tunelko.com/wp-content/uploads/2017/08/CIS_Docker_1.13.0_Benchmark_v1.0.0.pdf))
```yml
version: '2.1'
services:
  mail:
    image: tvial/docker-mailserver:stable
[...]

PR #676, opened 2017-07-30
Issue 631 run services with supervisor

johansmitsnl:
Use the supervisor as the main process. The start-mailserver is started
from the supervisord and then this process triggers others.

Defined some default variable in the Dockerfile. In order for
supervisored to build the command lines the ENV variable need to be set.
Therefore the defaults are defined.

Some processes are not single processes like postfix and fail2ban and
they have a wrapper. The wrapper takes care of proper shutdown and checking
if the process is running or not. Supervisored will restart the wrapping
script if the process is gone.

Increased some delays between tests because sometimes they where to short
for all containers to be running.

PR #2624, opened 2022-06-06
chore: Remove redundant capability `SYS_PTRACE`

polarathene:
# Description

This capability was added as part of the [original `check-for-changes.sh` feature PR in Oct 2017](https://github.com/docker-mailserver/docker-mailserver/pull/738):
- The original explanation for needing this was detailed in an [associated issue comment](https://github.com/docker-mailserver/docker-mailserver/issues/552#issuecomment-335243949), but it is unclear if they were referring to the earlier `postfix reload` commit, or the `supervisorctl restart postfix` line that replaced it.
- A review comment expressed a concern with using `postfix reload` that may possibly lead to a process (_which one?_) losing a PID (_for Postfix?_).

---

In June 2018, the [`SYS_PTRACE` requirement was clarified](https://github.com/docker-mailserver/docker-mailserver/issues/1057#issuecomment-503985435):

> _Without `SYS_PTRACE` one of the unit tests fails with several of these in syslog:_
> 
> `Jun 20 08:51:22 mail postfix/postfix-script[2964]: fatal: the Postfix mail system is already running`
> 
> _I think the postfix script is trying to connect to the master process with the saved pid and fails without `SYS_PTRACE`._
> _I haven't seen this on my server so far, but perhaps the tests are doing something that doesn't happen very often._

and:

> _This can happen when the system is running as well. The postfix management script needs `SYS_PTRACE`._

That is still a bit vague. None of the current unit tests seem to fail anymore. I assume the referenced "postfix management script" is [`wrappers/postfix-wrapper.sh`](https://github.com/docker-mailserver/docker-mailserver/blob/0010786d181a3a6c70856d4c34e46b60ad015b7e/target/scripts/wrapper/postfix-wrapper.sh)?

`postfix-wrapper.sh` was [introduced in Aug 2017](https://github.com/docker-mailserver/docker-mailserver/pull/676) (_as part of adding `supervisord` into the image_), so a little bit earlier than the change detection service. The wrapper purpose is described:

> _Some processes are not single processes like `postfix` and `fail2ban` and they have a wrapper._
> _The wrapper takes care of proper shutdown and checking if the process is running or not. Supervisored will restart the wrapping script if the process is gone._

---

I'm not sure what has changed since. Perhaps @casperklein or @georglauterbach know of a reason to keep `SYS_PTRACE` around?

I'd also like to understand the intent to require restarting Postfix and Dovecot, as opposed to `postfix reload` /`dovecot reload`. This was used prior to adopting `supervisord` (_and the wrapper scripts that it seemed to require?_). All I have to go by is the concern that maintainer expressed with a PID possibly being lost.

## Type of change

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,7 +259,6 @@ services:
       - ONE_DIR=1
     cap_add:
       - NET_ADMIN
-      - SYS_PTRACE
     restart: always
 ```
 
@@ -316,6 +315,5 @@ services:
       - POSTFIX_MESSAGE_SIZE_LIMIT=100000000
     cap_add:
       - NET_ADMIN
-      - SYS_PTRACE
     restart: always
 ```
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -26,5 +26,3 @@ services:
     stop_grace_period: 1m
     cap_add:
       - NET_ADMIN
-      - SYS_PTRACE
-
--- a/docs/content/config/advanced/full-text-search.md
+++ b/docs/content/config/advanced/full-text-search.md
@@ -83,7 +83,6 @@ While indexing is memory intensive, you can configure the plugin to limit the am
           stop_grace_period: 1m
           cap_add:
             - NET_ADMIN
-            - SYS_PTRACE
     ```
 
 3. Recreate containers:
--- a/docs/content/config/advanced/kubernetes.md
+++ b/docs/content/config/advanced/kubernetes.md
@@ -199,7 +199,6 @@ spec:
                 - NET_BIND_SERVICE
                 # miscellaneous  capabilities
                 - SYS_CHROOT
-                - SYS_PTRACE
                 - KILL
               drop: [ALL]
             seccompProfile:
--- a/docs/content/examples/tutorials/basic-installation.md
+++ b/docs/content/examples/tutorials/basic-installation.md
@@ -61,7 +61,6 @@ In this setup `docker-mailserver` is not intended to receive email externally, s
               - SPOOF_PROTECTION=0
             cap_add:
               - NET_ADMIN # For Fail2Ban to work
-              - SYS_PTRACE
         ```
 
     - The docs have a detailed page on [Environment Variables][docs-environment] for reference.
--- a/test/default_relay_host.bats
+++ b/test/default_relay_host.bats
@@ -8,7 +8,6 @@ function setup() {
     -v "${PRIVATE_CONFIG}":/tmp/docker-mailserver \
     -v "$(pwd)/test/test-files":/tmp/docker-mailserver-test:ro \
     -e DEFAULT_RELAY_HOST=default.relay.host.invalid:25 \
-    --cap-add=SYS_PTRACE \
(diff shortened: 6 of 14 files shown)


casperklein:
Older Postfix versions `< 3.4.7-2` shipped with the following line in the init script, to determine whether the process is running or not:

`dir=$(ls -l /proc/$pid/exe 2>/dev/null | sed 's/.* -> //; s/\/[^\/]*$//')`

That required the Docker container `SYS_PTRACE` capability.

Reference: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=941293

casperklein (approved):

georglauterbach (approved):

polarathene:
@casperklein Oh awesome thank you so much for sharing that! :grin: 

Good to know this is safe to go ahead with then :)

polarathene:
@casperklein would this also mean that the wrapper scripts aren't likely required anymore? At least for Postfix?

Or do you know if the concerns about losing PID remain valid for using `postfix reload` in `check-for-changes.sh` instead of `supervisorctl restart postfix`?

casperklein:
The wrapper script is (still) needed, because:

https://github.com/docker-mailserver/docker-mailserver/blob/851ec8cbcd35fe7b01fdbe8c9737289ea3d23c40/target/scripts/wrapper/postfix-wrapper.sh#L3-L8

However, there should be no problem when issuing `service postfix reload` etc.

Not sure if I understood your question correctly, let me know if not 😉

georglauterbach:
> The wrapper script is (still) needed, because:
> 
> https://github.com/docker-mailserver/docker-mailserver/blob/851ec8cbcd35fe7b01fdbe8c9737289ea3d23c40/target/scripts/wrapper/postfix-wrapper.sh#L3-L8

What about <https://serverfault.com/a/916072>?

casperklein:
I didn't know that. That might be worth investigating/testing. Not sure, if the things `sekretts` mentions in his comment could cause problems in our setup.

polarathene:
> Not sure, if the things `sekretts` mentions in his comment could cause problems in our setup.

Someone running Podman Rootless had some odd permission issues occur, not sure why though:

https://github.com/docker-mailserver/docker-mailserver/issues/2630#issuecomment-1151728893

It affected `/var/spool/postfix/private` rather than `/var/spool/postfix/etc`, although this was with our current image not the suggested alternative way to start Postfix.

casperklein:
> I'm not sure what benefit there is from starting Postfix in foreground mode

No benefit, but a requirement for supervisord. Because postfix did not support foreground mode in the past, the wrapper script was needed.

1. supervisord expects the command to be running in the foreground, so that signals like SIGTERM can be passed through. `service postfix start` for example is a one-shot (it immediately exits after execution, so no signals can be passed).
2. More important, if the command exists, supervisord "thinks" the process is gone and trys to restart, since we set `autorestart=true` in `target/supervisor/conf.d/supervisor-app.conf`. This leads to an endless restart loop (postfix is running, but supervisord falsely think its gone and trys to restart it)

PR #2628, opened 2022-06-07
chore: Extract out `/var/mail` ownership workaround

polarathene:
# Description

Container startup and the Change Detector have their own copy of the `chown -R 5000:5000 /var/mail` logic. I replaced the duplicate code with a call to a shared method in `helpers/utils.sh`.

I've added a `TODO` maintenance comment for the method. I've already looked into that concern and will detail it with a follow-up PR (for a breaking change) where we can discuss if removal is appropriate. For now, at least lets keep the shared logic in sync :+1: 

This PR is part of investigating https://github.com/docker-mailserver/docker-mailserver/pull/2256 so that it can be resolved.

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/check-for-changes.sh
+++ b/target/scripts/check-for-changes.sh
@@ -117,10 +117,9 @@ function _check_for_changes
     # be sure to mimic `setup-stack.sh:_setup_ldap` which appends to `/tmp/vhost.tmp`.
     _create_postfix_vhost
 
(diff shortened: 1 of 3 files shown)


georglauterbach (approved):
Good first action in finally resolving the old PR you mentioned - much appreciated!

casperklein (approved):

PR #2662, opened 2022-06-26
Fail2ban can use legacy iptables

### LINKED ISSUE #2661: [FR] Allow fail2ban on legacy iptables
# Feature Request

## Context

Docker running in some appliance (e.g., QNAP NAS') still uses legacy iptables.
This prevents fail2ban shipped with docker-mailserver from working properly, as it uses nftables.
 
Yes, the problem is "fail2ban in docker-mailserver won't work on systems still using legacy iptables (vs nftables)"

Add a new environment variable to switch to legacy iptables.

Yes, because I know the probability of someone else doing it is low, and I can learn from it.

PR is coming soon

I tried to load nftable kernel module on my NAS. That won't be a general solution, but still, I learned that QNAP doesn't ship with a Linux kernel with nftable.

Anyone using docker-mailserver on an appliance running docker, and not supporting modern firewall for Linux (QNAP NAS is an example, but there are more)

PE is coming. I am not sure if I am respecting conventions used for this project. Happy to change my approach if needed.

dguerri:
On some appliance (e.g., QNAP NAS') nftables on docker is not
available.
This commit adds FAIL2BAN_LEGACY_IPTABLES environment variable to allow configuring fail2ban to use iptables-legacy.

Fixes #2661

# Description

Fixes #

## Type of change

## Checklist:

- [N] If necessary I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -55,7 +55,7 @@ RUN \
   dovecot-ldap dovecot-lmtpd dovecot-managesieved dovecot-pop3d \
   dovecot-sieve dovecot-solr dumb-init \
   # E - O
-  ed fetchmail file gamin gnupg gzip iproute2 \
+  ed fetchmail file gamin gnupg gzip iproute2 iptables \
   locales logwatch lhasa libdate-manip-perl libldap-common liblz4-tool \
   libmail-spf-perl libnet-dns-perl libsasl2-modules lrzip lzop \
   netcat-openbsd nftables nomarch opendkim opendkim-tools opendmarc \
@@ -78,6 +78,9 @@ RUN \
     echo "ERROR: Wrong GPG fingerprint!" >&2; exit 1; fi && \
   dpkg -i fail2ban.deb 2>&1 && \
   rm fail2ban.deb fail2ban.deb.asc && \
+  # allow legacy iptables to be configured via FAIL2BAN_LEGACY_IPTABLES
+  update-alternatives --set iptables /usr/sbin/iptables-legacy && \
+  update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy && \
   # cleanup
   apt-get -qq autoremove && \
   apt-get -qq autoclean && \
--- a/docs/content/config/environment.md
+++ b/docs/content/config/environment.md
@@ -112,6 +112,14 @@ Otherwise, `nftables` won't be able to ban IPs.
 - reject => reject packet (send ICMP unreachable)
 FAIL2BAN_BLOCKTYPE=drop
 
+##### FAIL2BAN_LEGACY_IPTABLES
+
+Enable legacy iptables firewall. Useful when docker is running on some type of appliance (e.g. QNAP NAS)
+
+- **0**   => use nftables
+- 1 => use iptables-legacy
+FAIL2BAN_LEGACY_IPTABLES=0
+
 ##### SMTP_ONLY
 
 - **empty** => all daemons start
--- a/mailserver.env
+++ b/mailserver.env
@@ -126,6 +126,11 @@ ENABLE_FAIL2BAN=0
 # reject => reject packet (send ICMP unreachable)
 FAIL2BAN_BLOCKTYPE=drop
 
+# Fail2Ban use legacy iptables
+# **0**   => use nftables
+# 1 => use iptables-legacy
+FAIL2BAN_LEGACY_IPTABLES=0
+
 # 1 => Enables Managesieve on port 4190
 # empty => disables Managesieve
 ENABLE_MANAGESIEVE=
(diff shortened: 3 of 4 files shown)


polarathene (commented):
**TL;DR:**
- Defer to docs only contribution to share with other users. If more users express a need for this support, we can consider it then.
- Problem is likely due to NAS products (even those released last year) shipping kernels that are known to be 5 years old. Customers should engage with the product vendors to resolve that.

---

Looking at the PR that swapped `iptables` to `nftables` for v11, the [`fail2ban` debug `setup.sh` command for debugging had this `iptables` specific logic](https://github.com/docker-mailserver/docker-mailserver/pull/2505/files#diff-786fc2d200a8acdeaeae366a62e1f5f75d4bb0159083c82a5fc146ee6f7a4852L6-L21), which may be helpful context if your setup would run into the issue?

Otherwise this mostly looks like it covers the relevant changes from that PR, our tests don't however, and there was quite a bit of difference there in syntax.

I'm not sure if other maintainers would want to carry this legacy support even if equivalent test support was added. As the PR shows, you could extend the image to add the relevant `Dockerfile` lines, and use `user-patches.sh` script to apply the single `sed` edit to `/etc/fail2ban/jail.local`?

While your time spent looking into the support and creating a PR is appreciated, the contribution may be better served as docs only contribution for how to solve this problem. We're trying to avoid adding more niche-case (and especially legacy support related ones) features when possible, and since the required changes to support this are small, the other maintainers likely share this opinion to defer to docs only.

I do understand that's probably frustrating, since needing to extend the image is more bothersome than carrying around `user-patches.sh` alone and maintaining compatibility with upstream over time. I believe we're more open to accepting this support when there's enough demand expressed for it, otherwise the preference is to defer to documenting it for others until then.

---

I will note that the PR which switched us to `nftables` [had a discussion stating a `iptables` host + `nftables` container worked fine](https://github.com/docker-mailserver/docker-mailserver/pull/2505#issuecomment-1076767508), and presumably the other way around as we've had no complaints since.

So I'm not sure why that differs with your QNAP host? [Another docker project has this issue too](https://github.com/DyonR/docker-qbittorrentvpn/issues/25#issuecomment-891185904), the users affected are QNAP and DSM (both NAS products), where at the end it's said that at least DSM 7 (released June 2021) was found to be using an extremely old kernel (4.4, Jan 2016)... that is not encouraging, and probably applies to your QNAP system? You should probably raise a complaint as a customer if you're on a kernel that old as well..

dguerri:
@polarathene everything you said totally makes sense.
I am happy to have this issue and its solution mentioned in the documentation.

For completeness, QNAP does use a recent (longterm supported) kernel. They seem to have intentionally opted for the legacy iptables, probably because of some of the software running on these appliances.

```
[/] # cat /etc/os-release
NAME="QTS"
VERSION="5.0.0 (20220531)"
ID=qts
PRETTY_NAME="QTS 5.0.0 (20220531)"
VERSION_ID="5.0.0"
[/] # cat /proc/version
Linux version 5.10.60-qnap (root@U16BuildServer172) (x86_64-QNAP-linux-gnu-gcc (toolchain config: [gcc-4.9.2 binutils-2.25 glibc-2.21]) 4.9.2, GNU ld (GNU Binutils) 2.25) #1 SMP Tue May 31 01:57:44 CST 2022
[/] # lsmod | grep table
iptable_nat            16384  2
nf_nat                 36864  5 ip_vs_ftp,xt_nat,xt_MASQUERADE,xt_REDIRECT,iptable_nat
[/] # ls /lib/modules/5.10.60-qnap/*table*
/lib/modules/5.10.60-qnap/ip6table_filter.ko  /lib/modules/5.10.60-qnap/ip6table_mangle.ko  /lib/modules/5.10.60-qnap/ip6_tables.ko  /lib/modules/5.10.60-qnap/iptable_nat.ko
[/] #
```

dguerri:
@polarathene I commented on PR #2505. The author of that patch was probably using the new iptables drop-in replacement (which actually uses nftables)

georglauterbach:
@dguerri was indeed right, my host was running the `iptables` version that under the hood already used `nftables`. Id I'm not mistaken, this had already begun shipping for some `4.x.x` Linux kernels (Linux `3.13` introduced `nftables`).

Nevertheless, I agree with @polarathene in this case, i.e. this PR should become a documentation-only PR. I have a bit of experience with these NAS systems, and I know that they are not that nice to handle, but I think this is exactly the reason we should not provide extra support in our code base - it makes support and maintainability more difficult than it already is.

I'd be happy to see this PR being merged with new documentation (only) though! @dguerri could you update this PR, we will gladly help you with our docs if some questions come up :)

dguerri:
@georglauterbach SGTM. Will do asap

georglauterbach:
ping @dguerri

dguerri:
Sorry guys, something came up and I don't have much time right now to reflector the PR.
Shall we close it and I will submit a documentation one as soon as I can?

PR #2729, opened 2022-08-19
chore: `addmailuser` - Remove delaying completion until `/var/mail` is ready

polarathene:
# Description

## Quick Summary

Resolves a `TODO` task with `addmailuser`.

## Overview

The main change is adding three new methods in `common.bash`, which replace the completion delay in `addmailuser` / `setup email add` command.

Other than that:

- I swapped `sh -c 'addmailuser ...'` to `setup email add ...`.
- Improved three tests in `setup-cli.bats` for `setup email add|update|del` (_logic remains effectively the same still_).
- Rewrote the `TODO` comment for `setup-cli.bats` test on `setup email del` to better clarify the concern, but the test itself was no longer affected due to changes prior to this PR, so I enabled the commented out assertion.
- Removed unnecessary waits. The two `skip` tests in `test/tests.bats` could be enabled again after this PR.

Commit messages provide additional details if helpful.

## Details

Presently when adding a user to `postfix-accounts.cf` with a running container, the `addmailuser` (_aka `setup email add`_) command will delay completion until the account has a directory in `/var/mail` (_which is created by the `changedetector` service during creation of the Dovecot user_).

AFAIK the delay was added primarily for our test suite and serves no other purpose. This PR migrates that delay logic into `common.bash` test helper methods, resolving the `TODO`.

### Future work

A future improvement might instead skip the `changedetector` service and add the Dovecot account directly.

Presently that is not an option as [that functionality](https://github.com/docker-mailserver/docker-mailserver/blob/0d5f550bdf22ab21bb7a852861221ae0b82264e1/target/scripts/helpers/accounts.sh#L12-L43) is wrapped in a loop and assumes a clean slate.

The upcoming Dovecot `mailcrypt` plugin feature does require `addmailuser` delaying until `/var/mail` account folder is available however. This PR will not help users of that feature until Dovecot accounts can be created directly without relying on `changedetector`.

---

### TODO addressed by this PR

Removed comment inline-docs `TODO` for reference (_and having Github link the PR to those issues if traceability is needed in future_)

> `TODO:` Remove this method or at least it's usage in `addmailuser`. If tests are failing, correct the tests.
> 
> This method was added to delay command completion until a change detection event had processed the newly added user,
> confirmed once maildir was created.
> It was a workaround to accommodate the test suite apparently, but otherwise
> prevents batch adding users (_each one would have to go through their own change detection event_).
> 
> Originally introduced in [PR 1980](https://github.com/docker-mailserver/docker-mailserver/pull/1980) (_afterwards two futher PRs deleted, and then reverted that deletion_):
> Not much details/discussion were in the PR, these are the specific commits:
> - [Initial commit](https://github.com/docker-mailserver/docker-mailserver/pull/1980/commits/2ed402a12cedd412abcf577e8079137ea05204fe#diff-92d2047e4a9a7965f6ef2f029dd781e09265b0ce171b5322a76e35b66ab4cbf4R67)
> - [Follow-up commit](https://github.com/docker-mailserver/docker-mailserver/pull/1980/commits/27542867b20c617b63bbec6fdcba421b65a44fbb#diff-92d2047e4a9a7965f6ef2f029dd781e09265b0ce171b5322a76e35b66ab4cbf4R67)
> 
> Original reasoning for this method (_sounds like a network storage I/O issue_):
> Tests fail if the creation of `/var/mail/${DOMAIN}/${USER}` doesn't happen fast enough after `addmailuser` executes (_`check-for-changes.sh` race-condition_)
> Prevent infinite loop in tests like: _"checking accounts: user3 should have been added to /tmp/docker-mailserver/postfix-accounts.cf even when that file does not exist"_

---

Fixes: https://github.com/docker-mailserver/docker-mailserver/issues/2096#issuecomment-885205611

## Type of change

## Checklist:

### DIFF
--- a/test/tests.bats
+++ b/test/tests.bats
@@ -102,6 +102,16 @@ teardown_file() {
 }
 
 #
+# healthcheck
+#
+
+@test "checking container healthcheck" {
+  run bash -c "docker inspect mail | jq -r '.[].State.Health.Status'"
+  assert_output "healthy"
+  assert_success
+}
+
+#
 # processes
 #
 
@@ -958,16 +968,6 @@ EOF
   # check sender is not the default one.
   run docker exec mail grep "From: mailserver-report@mail.my-domain.com" /var/mail/localhost.localdomain/user1/new/ -R
   assert_failure
-}
-
-#
-# healthcheck
-#
-
-@test "checking container healthcheck" {
-  run bash -c "docker inspect mail | jq -r '.[].State.Health.Status'"
-  assert_output "healthy"
-  assert_success
 }
 
 #
--- a/test/tests.bats
+++ b/test/tests.bats
@@ -10,6 +10,7 @@ setup_file() {
   PRIVATE_CONFIG=$(duplicate_config_for_container . mail)
   mv "${PRIVATE_CONFIG}/user-patches/user-patches.sh" "${PRIVATE_CONFIG}/user-patches.sh"
 
+  # `LOG_LEVEL=debug` required for using `wait_until_change_detection_event_completes()`
   docker run --rm -d --name mail \
     -v "${PRIVATE_CONFIG}":/tmp/docker-mailserver \
     -v "$(pwd)/test/test-files":/tmp/docker-mailserver-test:ro \
@@ -22,6 +23,7 @@ setup_file() {
     -e ENABLE_SPAMASSASSIN=1 \
     -e ENABLE_SRS=1 \
     -e ENABLE_UPDATE_CHECK=0 \
+    -e LOG_LEVEL='debug' \
     -e PERMIT_DOCKER=container \
     -e PERMIT_DOCKER=host \
     -e PFLOGSUMM_TRIGGER=logrotate \
--- a/test/tests.bats
+++ b/test/tests.bats
@@ -641,7 +641,7 @@ EOF
 }
 
 @test "checking user updating password for user in /tmp/docker-mailserver/postfix-accounts.cf" {
-  docker exec mail /bin/sh -c "addmailuser user4@domain.tld mypassword"
+  add_mail_account_then_wait_until_ready mail 'user4@domain.tld'
 
   initialpass=$(docker exec mail /bin/sh -c "grep '^user4@domain\.tld' -i /tmp/docker-mailserver/postfix-accounts.cf")
   sleep 2
@@ -691,8 +691,7 @@ EOF
 
 
 @test "checking quota: setquota user must be existing" {
-  run docker exec mail /bin/sh -c "addmailuser quota_user@domain.tld mypassword"
-  assert_success
+  add_mail_account_then_wait_until_ready mail 'quota_user@domain.tld'
 
   run docker exec mail /bin/sh -c "setquota quota_user 50M"
   assert_failure
@@ -707,8 +706,7 @@ EOF
 }
 
 @test "checking quota: setquota <quota> must be well formatted" {
-  run docker exec mail /bin/sh -c "addmailuser quota_user@domain.tld mypassword"
-  assert_success
+  add_mail_account_then_wait_until_ready mail 'quota_user@domain.tld'
 
   run docker exec mail /bin/sh -c "setquota quota_user@domain.tld 26GIGOTS"
   assert_failure
@@ -737,8 +735,7 @@ EOF
 }
 
 @test "checking quota: delquota user must be existing" {
-  run docker exec mail /bin/sh -c "addmailuser quota_user@domain.tld mypassword"
-  assert_success
+  add_mail_account_then_wait_until_ready mail 'quota_user@domain.tld'
 
   run docker exec mail /bin/sh -c "delquota uota_user@domain.tld"
   assert_failure
@@ -759,8 +756,7 @@ EOF
 }
 
 @test "checking quota: delquota allow when no quota for existing user" {
-  run docker exec mail /bin/sh -c "addmailuser quota_user@domain.tld mypassword"
-  assert_success
+  add_mail_account_then_wait_until_ready mail 'quota_user@domain.tld'
 
   run docker exec mail /bin/sh -c "grep -i 'quota_user@domain.tld' /tmp/docker-mailserver/dovecot-quotas.cf"
   assert_failure
@@ -814,8 +810,7 @@ EOF
 }
 
 @test "checking quota: quota directive is removed when mailbox is removed" {
-  run docker exec mail /bin/sh -c "addmailuser quserremoved@domain.tld mypassword"
-  assert_success
+  add_mail_account_then_wait_until_ready mail 'quserremoved@domain.tld'
 
   run docker exec mail /bin/sh -c "setquota quserremoved@domain.tld 12M"
   assert_success
@@ -853,7 +848,8 @@ EOF
   skip 'disabled as it fails randomly: https://github.com/docker-mailserver/docker-mailserver/pull/2511'
 
   # create user
-  run docker exec mail /bin/sh -c "addmailuser quotauser@otherdomain.tld mypassword && setquota quotauser@otherdomain.tld 10k"
+  add_mail_account_then_wait_until_ready mail 'quotauser@otherdomain.tld'
+  run docker exec mail /bin/sh -c 'setquota quotauser@otherdomain.tld 10k'
   assert_success
 
   # wait until quota has been updated
--- a/test/tests.bats
+++ b/test/tests.bats
@@ -51,16 +51,18 @@ setup_file() {
   # setup sieve
   docker cp "${PRIVATE_CONFIG}/sieve/dovecot.sieve" mail:/var/mail/localhost.localdomain/user1/.dovecot.sieve
 
-  # this relies on the checksum file beeing updated after all changes have been applied
+  # this relies on the checksum file being updated after all changes have been applied
   wait_until_change_detection_event_completes mail
-
-  wait_for_smtp_port_in_container mail
 
   # wait for ClamAV to be fully setup or we will get errors on the log
   repeat_in_container_until_success_or_timeout 60 mail test -e /var/run/clamav/clamd.ctl
 
-  # sending test mails
-  docker exec mail /bin/sh -c "nc 0.0.0.0 25 < /tmp/docker-mailserver-test/email-templates/amavis-spam.txt"
+  wait_for_smtp_port_in_container mail
+
+  # The first mail sent leverages an assert for better error output if a failure occurs:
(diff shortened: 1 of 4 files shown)


polarathene:
Tests were restarted, this time [the original failure](https://github.com/docker-mailserver/docker-mailserver/runs/7910857994?check_suite_focus=true#step:8:224) did not occur, but a test that an earlier test (_[that originally passed](https://github.com/docker-mailserver/docker-mailserver/runs/7910857994?check_suite_focus=true#step:8:193)_) is now failing:

```
not ok 186 checking setup.sh: setup.sh email del in 391ms
#  in test file test/setup-cli.bats, line 76)
#   `assert_success' failed
#
# -- command failed --
# status : 1
# output (11 lines):
#   [  ERROR  ]  Mailbox directory '/var/mail/impsum.org/lorem' does not exist
#   [  ERROR  ]  Aborting
#
#   --- UNCHECKED ERROR
#     - script    = ./setup.sh
#     - function  = _main
#     - command   = ${CRI} exec "${USE_TTY}" "${CONTAINER_NAME}" setup "${@}"
#     - line      = 229
#     - exit code = 1
#
#   This should not have happened. Please file a bug report.
# --
#
```

Since that was not caught earlier or locally for me, it's likely a timing issue. I've improved `test/setup-cli.bats` so that this shouldn't happen. Although `test/tests.bats` may be a bit more tricky :sweat_smile:

georglauterbach:
Will check this next week :)

polarathene:
> Will check this next week :)

The change isn't as big as it looks :sweat_smile: 

The main change is adding three new methods in `common.bash`, which replace the completion delay in `addmailuser` / `setup email add` command.

Other than that:

- I swapped `sh -c 'addmailuser ...'` to `setup email add ...`.
- Improved three tests in `setup-cli.bats` for `setup email add|update|del` (_logic remains effectively the same still_).
- Rewrote the `TODO` comment for `setup-cli.bats` test on `setup email del` to better clarify the concern, but the test itself was no longer affected due to changes prior to this PR, so I enabled the commented out assertion.
- Removed unnecessary waits. The two `skip` tests in `test/tests.bats` could be enabled again after this PR.

If you need extra information, viewing the full file from the web UI context menu should let you lookup `git blame` commit message, or just ask :)

All good though, no rush :+1:

georglauterbach (approved):
Had a look again, got a nice overview with your comments, and this looks good - LGTM 👍🏼

casperklein (approved):

polarathene:
Transient failure again:

```
not ok 217 setup_file failed
# (from function `setup_file' in test file test/tests.bats, line 63)
#   `docker exec mail /bin/sh -c "nc 0.0.0.0 25 < /tmp/docker-mailserver-test/email-templates/amavis-spam.txt"' failed
# 0a153c81aa5dd7fa5bfecd19e42362b4e52964d3ff267479472f72288deef212
```

Did you both look over the new `wait_until_change_detection_event_completes()` method in `common.bash`? It is not a replacement yet for `wait_for_changes_to_be_detected_in_container()`, but is similar in functionality.

Previously, `addmailuser` would wait for `changedetector` to be processing until the `/var/mail/` account folder was created, then not long after that, the test would monitor checksum file for a change (_which presumably happens before the active event completes_). Either that or timeout failure in `setup_file()` with no assertion would hide the failure?

This is the previous helper used:

```bash
function wait_for_changes_to_be_detected_in_container() {
  local CONTAINER_NAME="${1}"
  local TIMEOUT=${TEST_TIMEOUT_IN_SECONDS}
  # shellcheck disable=SC2016
  repeat_in_container_until_success_or_timeout "${TIMEOUT}" "${CONTAINER_NAME}" bash -c 'source /usr/local/bin/helpers/index.sh; _obtain_hostname_and_domainname; cmp --silent -- <(_monitored_files_checksums) "${CHKSUM_FILE}" >/dev/null'
}
```

That does not detect if a change event is starting/ending, just that a transition from event start to end, or new event starting has occurred.

I think what has happened is Postfix is still restarting at this point since my new change detection is based on when the event has finished processing. The older approach detection happened a little bit later with the `mv` (_and perhaps bit more delay in processing_):

https://github.com/docker-mailserver/docker-mailserver/blob/26053c22bd9c8b41735d6628caa6810316f27bc1/target/scripts/check-for-changes.sh#L68-L72

I am not sure how reliable `wait_for_service "${TEST_NAME}" postfix` will be, if Postfix is `RUNNING` but has a slight startup delay? If it's still problematic, perhaps changing from _restarting_ Postfix to _reloading_ will minimize the downtime.

test/tests.bats
```diff
@@ -50,7 +52,7 @@ setup_file() {
   docker cp "${PRIVATE_CONFIG}/sieve/dovecot.sieve" mail:/var/mail/localhost.localdomain/user1/.dovecot.sieve
 
   # this relies on the checksum file beeing updated after all changes have been applied
-  wait_for_changes_to_be_detected_in_container mail
+  wait_until_change_detection_event_completes mail
```
polarathene, line 55: Attempt to fix transient failure with `nc` mail lines. Postfix is presumably unreachable..?

```suggestion
  wait_until_change_detection_event_completes mail
  wait_for_service mail postfix
```

This seems unlikely though if `wait_for_smtp_port_in_container mail` is working.. It could have something to do with the clamav wait that follows?

polarathene:
Well it didn't fail at the same spot, this time it was the healthcheck test:

```
not ok 312 checking container healthcheck in 87ms
#  in test file test/tests.bats, line 969)
#
# -- output differs --
# expected : healthy
# actual   : unhealthy
# --
#
```

https://github.com/docker-mailserver/docker-mailserver/blob/26d241381f9c683d1cbcb7300534fa63f1127615/test/tests.bats#L974-L978

https://github.com/docker-mailserver/docker-mailserver/blob/26d241381f9c683d1cbcb7300534fa63f1127615/test/tests.bats#L40

That's fairly far down in the test, not sure how it ended up unhealthy?

- https://github.com/docker-mailserver/docker-mailserver/pull/2625#issuecomment-1147329092
- https://docs.docker.com/engine/reference/builder/#healthcheck

Defaults:

```
--interval=DURATION (default: 30s)
--timeout=DURATION (default: 30s)
--start-period=DURATION (default: 0s)
--retries=N (default: 3)
```

I think `retries` has no delay between attempts, and all `retries` are part of the same check in the `timeout` bounds? So likely a false-positive here due to the check being affected at a time postfix was being restarted due to change detection event.. I'll shift this test up to the start then.

georglauterbach:
I hope I got this right: as the new `addmailuser` does not wait for the `changedetector` to pick up the changes and create the Dovecot account, there is a slight delay after the `addmailuser` command succeeds, right? Our test suite has always been fragile when it comes to timing in certain areas, your logic and code seem hardened to me however (nice work BTW). I do not see why tests would fail at the moment. I had experienced the healthcheck failing in the past too, so moving it up seems like a good idea, but I would add a comment if there really is a inter-test-dependency.

polarathene:
> I hope I got this right: as the new `addmailuser` does not wait for the `changedetector` to pick up the changes and create the Dovecot account, there is a slight delay after the `addmailuser` command succeeds, right?

Before:
- `addmailuser` make an account to `postfix-accounts.cf`, don't exit until `/var/mail` has created an associated folder `/var/mail/<domain>/<user>`.
- `changedetector` notice update in `postfix-accounts.cf`:
  - Add account to UserDB and PassDB for Dovecot.
  - Create the folder (**`addmailuser` condition met**).
  - Apply check for user/group ownership (`mkdir` called by root in script, so condition triggers `chown`) - potential race condition (_encountered in PR for Dovecot encrypted storage, I have a fix for that to wait on ownership instead of only checking `/var/mail` account directory exists_).
  - Change detection restarts Postfix and Dovecot.
  - Change detection complete.
  - Update checksum file.
  - Dovecot and Postfix may still be coming back up for a moment.

Now:
- `addmailuser` make an account to `postfix-accounts.cf`, exit.
- Tests can use `wait_until_account_maildir_exists()`:
  - Wait until directory exists, just like `addmailuser` did before.
- Tests can use `wait_until_change_detection_event_completes()`:
  - Wait on "Change detection complete" (after first waiting on a "Change detection start" equivalent event is present in logs).
  - Only confirms a change detection has occurred (_may be sufficient, account now exist in Dovecot UserDB and PassDB, `doveadm` utilities can work without running Dovecot instance I think_)
  - Otherwise use additional wait on service needed, or worse sleep for some time. No different to handling `addmailuser` before.

So mostly, `addmailuser` will finish much more quickly now, and if you actually need the delay to wait until maildir creation, you use `wait_until_account_maildir_exists()`, it should be have the same. If you want to instead wait a little bit longer, but avoid some potential race conditions, `wait_until_change_detection_event_completes()` is more useful.

Neither guarantees Dovecot or Postfix is up, as `changedetector` only reports when it's done processing, not on async activity of services restarting which can still occur after `changedetector` is finished it's part. `wait_on_service()` is available, but it just queries `supervisord` if the service in `RUNNING` state, which should be the case when the service is not yet ready during it's restart.

v12 can change from `restart` to `reload` of the services which should help minimize that issue assuming it causes no regressions.

---

New transient failure (this PR is just the gift that keeps on giving with random failing tests..):

```
not ok 186 checking setup.sh: setup.sh email del in 429ms
#  in test file test/setup-cli.bats, line 116)
#   `assert_failure' failed
#
# output :
# --
#
```

This time it's a test this PR actually touches, and is enabling a check that was previously disabled:

```bash
  # Mail storage for account was actually removed by `-y`:
  run docker exec "${TEST_NAME}" ls /var/mail/example.com/user
  assert_failure
```

Potentially a race condition with the previous `setup email update` test. Change detection is likely running, and perhaps starts to create all accounts from scratch while `setup email del` deletes the directory, and that is recreated by the `changedetector`? Doesn't look like it'd have much of a window to do that within :\

This is perhaps one of those scenarios where proper locking with `flock()` would avoid the issue as the user management commands and the change detector would be actually aware of each other (_present file locking was changed around about 10.2, maybe earlier but mistakenly places a lock on the active script running, not the relevant config_).

For now I'll ensure the change detector is done processing..

georglauterbach (approved):

casperklein (approved):

test/tests.bats
georglauterbach: This seems to be the change that caused the transient `nc` failures show up. I hope #2811 can do something about it.

polarathene: It could, but I think it's more likely that it's helped surfaced awareness of an issue we've had for some time (_possibly related to the cause that required some tests to be disabled_). I am inclined to believe ClamAV is still the actual problem.

Issue #2722, opened 2022-08-15
[BUG] `ENABLE_SRS=1` causing high CPU usage with `postsrsd`


polarathene:
### Miscellaneous first checks

- [X] I checked that all ports are open and not blocked by my ISP / hosting provider.
- [X] I know that SSL errors are likely the result of a wrong setup on the user side and not caused by DMS itself. I'm confident my setup is correct.

### Affected Component(s)

postsrsd

### What happened and when does this occur?

Another user [reported to upstream `postsrsd`](https://github.com/roehling/postsrsd/issues/122) experiencing this same issue earlier in the year (_specifically with DMS_).

I noticed this while working on a PR and debugging local test failures. Why the tests do not fail in our CI is unclear (_might be due to older environment/packages in CI?_):

```
 ✗ checking configuration: hostname/domainname override: check headers of received mail [235]
   (from function `assert_success' in file test/test_helper/bats-assert/src/assert.bash, line 114,
    in test file test/mail_hostname.bats, line 104)
     `assert_success' failed
   
   -- command failed --
   status : 1
   output : ls: cannot access '/var/mail/localhost.localdomain/user1/new': No such file or directory
   --
```

```
 ✗ checking configuration: non-subdomain: check headers of received mail [308]
   (from function `assert_success' in file test/test_helper/bats-assert/src/assert.bash, line 114,
    in test file test/mail_hostname.bats, line 163)
     `assert_success' failed
   
   -- command failed --
   status : 1
   output : ls: cannot access '/var/mail/localhost.localdomain/user1/new': No such file or directory
   --
```

I had a process monitor active in the background and noticed the four containers spun up in that test each used 25% CPU for the containers `postsrsd` process.

When starting a single container in my VM (2vCPU) `postsrsd` was consistently showing 50% CPU usage, despite no mail to process.

The ports `10001` and `10002` are not bound to localhost:

```console
$ ss -tulpn | grep LISTEN

tcp   LISTEN 0      4096       127.0.0.1:10024      0.0.0.0:*                                     
tcp   LISTEN 0      100        127.0.0.1:10025      0.0.0.0:*                                     
tcp   LISTEN 0      100          0.0.0.0:587        0.0.0.0:*                                     
tcp   LISTEN 0      100          0.0.0.0:143        0.0.0.0:*    users:(("dovecot",pid=642,fd=32))
tcp   LISTEN 0      100          0.0.0.0:465        0.0.0.0:*                                     
[...]

PR #2730, opened 2022-08-19
tests: Ensure excessive FD limits are avoided

polarathene:
# Description

Local tests can fail depending on environment if Docker is configured with an excessive FD limit.

This was affecting tests using `ENABLE_SRS=1` and `ENABLE_FAIL2BAN=1` due to a common step for daemonization taking considerably longer to complete (8 minutes on Fedora 36), causing the tests to fail due to timeout or unreachable service.

I have chosen kernel defaults (1024 soft, 4096 hard), as opposed to systemd (1024, 512K) which should be fine for tests. The limit is for how many files/streams a process can have open at one time. A process can request to raise the soft limit so long as it's below the hard limit.

I avoided documentation as I don't expect this to be a common issue users would face in real deployments. If an issue is raised about high CPU usage or similar odd activity to troubleshoot, it may be due to this, and if so warrant documenting for users. Otherwise I only expect it to affect contributors running tests locally.

### Additional Details

Processes that run as daemons (`postsrsd` and `fail2ban-server`) initialize by closing all FDs (File Descriptors).

This behaviour queries that maximum limit and iterates through the entire range even if only a few FDs are open. In some environments (_Docker, limit configured varies by distro_) this can be a range exceeding 1 billion (_from kernel default of 1024 soft, 4096 hard_), causing an 8 minute delay with heavy CPU activity.

`postsrsd` has since been updated to use `close_range()` syscall, and `fail2ban` will now iterate through `/proc/self/fd` (open FDs) which should resolve the performance hit. Until those updates reach our Docker image, we need to workaround it with `--ulimit` option.

NOTE: The CI does not seem affected, but it can affect local development when running tests causing failures. If `docker.service` on a distro sets `LimitNOFILE=` to approx 1 million or lower, it should not be an issue. On distros such as Fedora 36, it is `LimitNOFILE=infinity` (approx 1 billion) that causes excessive delays.

When `close_range()` syscall is available, at least in Python, this requires kernel 5.9+ and `glibc` >= 2.34, the performance hit is avoided. On Debian 11 and Alpine Linux 3.16, neither container meets the `glibc` requirement (_Debian 11 package is too old, Alpine uses `musl` and has no equivalent support_), whilst Fedora 36 container has `glibc` 2.35 (_[performance improvement demonstrated](https://github.com/docker-mailserver/docker-mailserver/issues/2722#issuecomment-1219104993)_). Thus this PR can likely be reverted once the next Debian release in 2023 occurs.

---

Fixes: #2722

## Type of change

## Checklist:

### DIFF
--- a/test/mail_fail2ban.bats
+++ b/test/mail_fail2ban.bats
@@ -9,7 +9,10 @@ function setup_file() {
     -e ENABLE_FAIL2BAN=1 \
     -e POSTSCREEN_ACTION=ignore \
     --cap-add=NET_ADMIN \
-    -h mail.my-domain.com -t "${NAME}"
+    --hostname mail.my-domain.com \
+    --tty \
+    --ulimit 'nofile=1024:4096' \
+    "${NAME}"
 
   # Create a container which will send wrong authentications and should get banned
   docker run --name fail-auth-mailer \
--- a/test/mail_fail2ban.bats
+++ b/test/mail_fail2ban.bats
@@ -11,7 +11,7 @@ function setup_file() {
     --cap-add=NET_ADMIN \
     --hostname mail.my-domain.com \
     --tty \
-    --ulimit 'nofile=1024:4096' \
+    --ulimit "nofile=$(ulimit -Sn):$(ulimit -Hn)" \
     "${NAME}"
 
   # Create a container which will send wrong authentications and should get banned
--- a/test/mail_hostname.bats
+++ b/test/mail_hostname.bats
@@ -11,8 +11,10 @@ function setup_file() {
     -e PERMIT_DOCKER=network \
     -e ENABLE_SRS=1 \
     -e OVERRIDE_HOSTNAME=mail.my-domain.com \
-    -h unknown.domain.tld \
-    -t "${NAME}"
+    --hostname unknown.domain.tld \
+    --tty \
+    --ulimit 'nofile=1024:4096' \
+    "${NAME}"
 
   PRIVATE_CONFIG_TWO=$(duplicate_config_for_container . mail_non_subdomain_hostname)
   docker run --rm -d --name mail_non_subdomain_hostname \
@@ -21,7 +23,9 @@ function setup_file() {
     -e PERMIT_DOCKER=network \
     -e ENABLE_SRS=1 \
     --hostname domain.com \
-    -t "${NAME}"
+    --tty \
+    --ulimit 'nofile=1024:4096' \
+    "${NAME}"
 
   PRIVATE_CONFIG_THREE=$(duplicate_config_for_container . mail_srs_domainname)
   docker run --rm -d --name mail_srs_domainname \
@@ -32,7 +36,9 @@ function setup_file() {
     -e SRS_DOMAINNAME='srs.my-domain.com' \
     --domainname 'my-domain.com' \
     --hostname 'mail' \
-    -t "${NAME}"
+    --tty \
+    --ulimit 'nofile=1024:4096' \
+    "${NAME}"
 
   PRIVATE_CONFIG_FOUR=$(duplicate_config_for_container . mail_domainname)
   docker run --rm -d --name mail_domainname \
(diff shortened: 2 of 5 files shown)


polarathene:
The current test failure doesn't make much sense:

```
not ok 217 setup_file failed
# (from function `repeat_until_success_or_timeout' in file test/test_helper/common.bash, line 43,
#  from function `wait_for_tcp_port_in_container' in file test/test_helper/common.bash, line 90,
#  from function `wait_for_smtp_port_in_container' in file test/test_helper/common.bash, line 95,
#  from function `setup_file' in test file test/tests.bats, line 56)
#   `wait_for_smtp_port_in_container mail' failed
# 83ed3565c20ff1214d1cabbd4bc469272614fdb2aae3913d8efe3fb4ae8f8f05
# [   INF   ]  Waiting for dovecot to create '/var/mail/localhost.localdomain/pass'
# [   INF   ]  Waiting for dovecot to create '/var/mail/localhost.localdomain/pass'
# [   INF   ]  Waiting for dovecot to create '/var/mail/localhost.localdomain/pass'
# [   INF   ]  Waiting for dovecot to create '/var/mail/localhost.localdomain/pass'
# [   INF   ]  Waiting for dovecot to create '/var/mail/localhost.localdomain/pass'
# [   INF   ]  Waiting for dovecot to create '/var/mail/localhost.localdomain/pass'
# [   INF   ]  Waiting for dovecot to create '/var/mail/localhost.localdomain/pass'
# [   INF   ]  Waiting for dovecot to create '/var/mail/localhost.localdomain/pass'
# [   INF   ]  Waiting for dovecot to create '/var/mail/localhost.localdomain/pass'
# Timed out on command: docker exec mail /bin/sh -c nc -z 0.0.0.0 25
# mail
# bats warning: Executed 217 instead of expected 318 tests
```

This PR shouldn't be affecting that test in `test/tests.bats` like that?:

https://github.com/docker-mailserver/docker-mailserver/blob/0d5f550bdf22ab21bb7a852861221ae0b82264e1/test/tests.bats#L43-L56

The only relevant change is `--ulimit 'nofile=1024:4096'`, but looking online it seems that may be too low for the CI (_tests passed fine on my Fedora 36 VM_). I'll change to inherit the Docker host limits (typically set by systemd `DefaultLimitNOFILE=`).

georglauterbach (approved):
LGTM

casperklein (approved):
Two things:

1. It might be worth adding a little comment (`#2730`) or so to the `--ulimit lines`, to prevent research for someone wondering about these lines.
2. Reminder: Every time a "sleep" is used to fix something, god kills a kitten 😆

polarathene:
> 1\. It might be worth adding a little comment (`#2730`) or so to the `--ulimit lines`, to prevent research for someone wondering about these lines.

With them being only in the tests, hopefully `git blame` is easy enough to look up this PR. All core maintainers are currently aware of the purpose now. The tests are rarely modified, I'm pretty much the only one who does extensive refactoring there :sweat_smile: 

I'm not too concerned as by mid 2023 (?) the next Debian release will be ready, and that hopefully has the new `postsrsd` release packaged. While `fail2ban` I know you've opted to install from external source, so that'll probably get the update sooner?

I can add the reference comments if you still think they'd be useful. I would like to look at moving the container setup for tests into `docker-compose.yml` files at some point, `git blame` would be a bit more work by then, so more context inline would be helpful at that point.

polarathene:
Apologies, I pushed a fix to the wrong PR :grimacing: 

Dropped the commit and force-pushed. You can see that the commit is the same as the last merge master branch commit `2c7237`. so nothing has changed since approval :sweat_smile:

georglauterbach (approved):

casperklein (approved):

polarathene:
> Even with `git blame` I find it sometimes very hard to track code changes. There are so many refactorings or code is splitted/moved to other files, that it can be quite challenging to find the the origin commit that introduced something.

Absolutely, that's why I try to provide some rather verbose history in some refactoring PRs to save other maintainers from trawling through it :laughing: 

The most annoying ones are around 2020-2021 due to how large several were at moving everything around, so my special trick for that is to take something like the Dockerfile and jump back in time prior to that, then browse the repo from that point in time and find the file to git blame :sweat_smile: 

I do agree that it's not pleasant, and definitely prefer adding contextual doc comments for stuff that is harder to lookup / grok :+1:

PR #2731, opened 2022-08-21
fix: `_create_accounts()` should run after waiting

polarathene:
# Description

When waiting on an account to be added to `postfix-accounts.cf`, Dovecot account creation during the startup process had already run.

Startup continued without properly creating the mail account for Dovecot. Methods like `setup email list` (_with `ENABLE_QUOTAS=1`_) would fail. `changedetector` service was required to be triggered to re-create Dovecot users.

- Wrapped the logic for wait + shutdown into a function call.
- Moved `_create_accounts()` to bottom of the setup function.

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/startup/setup-stack.sh
+++ b/target/scripts/startup/setup-stack.sh
@@ -276,27 +276,34 @@ function _setup_dovecot_local_user
 {
   _log 'debug' 'Setting up Dovecot Local User'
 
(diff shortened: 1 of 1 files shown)


georglauterbach (approved):
👍

casperklein (approved):

PR #2737, opened 2022-08-21
scripts: revised linting script

georglauterbach:
# Description

The new version uses our `log.sh` helper to simplify logging
significantly. Moreover, the script was adjusted to the current style
and the GitHub workflow was streamlined. The workflow is ot providing
the version anymore (which was useless anyway), and has been compacted.

## Type of change

## Checklist:

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -2,7 +2,6 @@ name: Lint
 
 on:
   pull_request:
-    branches: [ '**' ]
   push:
     branches: [ master ]
 
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -1,11 +1,10 @@
-name: "Lint"
+name: Lint
 
 on:
   pull_request:
-    branches: [ "*" ]
+    branches: [ '**' ]
   push:
-    branches:
-      - master
+    branches: [ master ]
 
 permissions:
   contents: read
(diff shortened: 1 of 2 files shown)


.github/workflows/linting.yml
```diff
@@ -1,11 +1,10 @@
-name: "Lint"
+name: Lint
 
 on:
   pull_request:
-    branches: [ "*" ]
+    branches: [ '**' ]
```
polarathene, line 5: Why the double `*`? Does it do something differently?

polarathene (approved):
LGTM, Great work! :grinning:

.github/workflows/linting.yml
georglauterbach: According to the [GH docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet):

- `*`: Matches zero or more characters, but does not match the `/` character. For example, `Octo*` matches `Octocat`.
- `**`: Matches zero or more of any character.

polarathene: My PRs often have the format `<type>/<description>`, so I don't think it would be due to that as linting works on those PRs fine every time :sweat_smile:

georglauterbach: Indeed, mine too, but according to the docs it should only have worked with `**`. I will remove the (somewhat useless) restriction entirely as we want the linting to always run on every PR.

georglauterbach: Done in 9aba16fb20c1dbe5af21c1569faa323062755c71

polarathene (approved):

casperklein (approved):

PR #1256, opened 2019-09-09
Provide version information through labels

fbartels:
Can be retrieved by calling e.g. `docker inspect tvial/docker-mailserver:testing`

Sample output:

```
$ docker inspect -f '{{ range $k, $v := .ContainerConfig.Labels -}}
> {{ $k }}={{ $v }}
> {{ end -}}' tvial/docker-mailserver:testing
maintainer=Thomas VIAL
org.label-schema.description=A fullstack but simple mailserver (smtp, imap, antispam, antivirus, ssl...)
org.label-schema.name=docker-mailserver
org.label-schema.schema-version=1.0
org.label-schema.url=https://github.com/tomav/docker-mailserver
org.label-schema.vcs-ref=7580aec
org.label-schema.vcs-url=https://github.com/tomav/docker-mailserver
org.label-schema.version=v6.1.0-111-g7580aec
```

PR #2742, opened 2022-08-23
ci: Better build caching for CI

polarathene:
# Description

Fixing flaws to make the tests workflow and Dockerfile more cache friendly.

## Change Overview

In Sep 2019, a PR introduced the VCS `ARG` directives with [acknowledgement that it invalidates the build cache](https://github.com/docker-mailserver/docker-mailserver/pull/1256#issuecomment-529406724).

Every time the commit SHA changes, due to this placement all layers afterwards are invalidated and must be rebuilt, even though only `LABEL` directive values were being updated (_even without those, `RUN` will implicitly be invalidated due to `ARG` changing as it is cached as part of the build ENV_).

I've shifted all directives related to image config that aren't necessary until the end to a 2nd stage. The additional `FROM` introducing a 2nd stage is not necessary, but provides a basis for later refactoring the Dockerfile to split out ClamAV and Fail2Ban layers into their own stages to further improve build caching.

The original intent of this PR was to address the caching with a more stable cache key (_instead of the commit SHA_), such that if the build context (_files involved in the image build process_) has not changed since the last build, that we can re-use the build cache for that.

---

**Presently the existing cache key would:**
- Only be valid when running tests for the same commit again, but that is rarely helpful if tests failed as it prevents uploading the cache (_cache upload happens at end of workflow job if successful only_).
- Would needlessly upload cache (_new key per push if job successful, even if the image didn't really change_) if a PR only contributes tests wasting cache storage (_10GB storage limit, less frequently used cache is evicted first_)

**The new cache key:**
- Is valid when `Dockerfile`, `VERSION` and `target/**` has not changed between commits.
- Uploads cache even when tests fail (_due to cache eviction policy and open-source plan this is fine provided cache upload is not excessive_), but only when there was no direct cache `key` hit (_`restore-key` fallback to an older build cache, reducing build time and upon job success uploading for the new derived cache key_).

**Complimentary changes:**
- The `type=local` cache mode needs a [workaround](https://github.com/docker/build-push-action/blob/965c6a410d446a30e95d35052c67d6eded60dad6/docs/advanced/cache.md?plain=1#L193-L199) to avoid accidentally accumulating cache wastefully which adds to upload/download time and more evictions (_330-810MB + 7-20 sec depending on supported platforms, vs **2-3GB I've seen presently cached** with download times of approx 1 minute_).
  - A new `type=gha` cache mode has since been made available (still deemed experimental) earlier this year that uses the `actions/cache` API (_reverse engineered IIRC, not officially supported_), I've chosen to avoid that for now and have not had time to try and compare it.
- Fixing the `Dockerfile` issue with `ARG` usage is an important one, as while the derived cache key can be effective at caching, when the commit SHA changed, the cache was completely ignored due to the builder invalidating cache... which caused a problem as that cache hit would not upload the new build cache (_build args passed in aren't part of the derived cache key_). Although the built image will still be slightly different due to these `ARG`, we are only sharing the build layer cache between jobs, so there is minimal work to recreate the image from cache and append the `ARG` changes (_avoiding wasteful cache upload for `ARG` changes_).
- Workflow is better documented. Removed unnecessary inputs like default `Dockerfile` and buildx `builder` instance.

**Impact of platforms on cache storage + build time:**
- I also checked the build cache for each platform, approx 330MB per platform, but the two ARM platforms seem to be capable of sharing a bit as when they're both included with AMD64, the total size is 810MB (_655MB with AMD64 and ARM64 only_).
- The ARM builds turned out to **introduce 13 mins to build time** and work in parallel (_no worthwhile speed improvement with only ARM64_), thus deprecating ARMv7 won't provide CI improvements (_except reducing cache storage_).
- AMD64 alone seems warranted as it can be **built within 2 minutes without cache** (_30 seconds with cache_). It might make sense to run ARM64 as a separate job in parallel (_or if available a native ARM64 runner_), or as a dependent job after the more important ARM64 build and test.

---

## Additional details

In the first commit, I also experimented with `upload-artifact` and `download-artifact` to transfer across jobs but this was much worse than just using `actions/cache`.

Future improvements could leverage buildx `COPY --link` feature that is available since March 2022. `type=gha` for the image cache might be interesting too if it caches on step success instead of job success, allowing for tests to be in the same job. Similar workflows could also potentially be more DRY with Github Actions support for re-usable workflows.

For now though, this PR should help, especially if a follow-up moves out the ARM64 build.

## Type of change

## Checklist:

### DIFF
--- a/.github/workflows/test_merge_requests.yml
+++ b/.github/workflows/test_merge_requests.yml
@@ -81,7 +81,9 @@ jobs:
       - name: 'Set up Docker Buildx'
         uses: docker/setup-buildx-action@v2.0.0
 
-      # NOTE: ARM platforms removed as exporting the tar image output does not support multi-platform:
+      # NOTE: AMD64 can build within 2 minutes, ARM adds 13 minutes. 330MB each
+      # ARMv7 can build in parallel, adding no extra time (but does add 150MB cache size).
+      # Moving ARM build to a separate job would cut down time to start running tests.
       - name: 'Build images'
         uses: docker/build-push-action@v3.1.1
         with:
@@ -90,16 +92,16 @@ jobs:
             VCS_REF=${{ github.sha }}
             VCS_VER=${{ github.ref }}
           tags: mailserver-testing:ci
-          # Build for AMD64 (runs against test suite) and ARM64 (we only test a successful build):
-          platforms: linux/amd64
+          # Build for AMD64 (runs against test suite) and ARM64 (only to verify building works):
+          platforms: linux/amd64,linux/arm64
           # Paired with steps `actions/cache` and `Replace cache` (replace src with dest):
           # NOTE: `mode=max` is only for `cache-to`, it configures exporting all image layers.
           # https://github.com/docker/buildx/blob/master/docs/reference/buildx_build.md#cache-from
           cache-from: type=local,src=/tmp/.buildx-cache
           cache-to: type=local,dest=/tmp/.buildx-cache-new,mode=max
-          # Export AMD64 image for upload-artifact step, allowing other jobs to use it:
-          # https://github.com/docker/buildx/blob/master/docs/reference/buildx_build.md#docker
-          outputs: type=docker,dest=/tmp/dms-docker-img.tar
+          # This job just builds the image and stores to cache, no other exporting required:
+          # https://github.com/docker/build-push-action/issues/546#issuecomment-1122631106
+          outputs: type=cacheonly
 
       # WORKAROUND: The `cache-to: type=local` input for `build-push-action` persists old-unused cache.
       # The workaround is to write the new build cache to a different location that replaces the
@@ -110,15 +112,6 @@ jobs:
         run: |
           rm -rf /tmp/.buildx-cache
           mv /tmp/.buildx-cache-new /tmp/.buildx-cache
-
-     - name: 'Upload DMS Docker Image for job transfer'
-       uses: actions/upload-artifact@v3
-       with:
-         name: dms-build-${{ steps.derive-image-cache-key.outputs.digest }}
-         path: /tmp/dms-docker-img.tar
-          # Artifact limits are unclear for open-source projects,
-          # Each image is roughly 500MB+ (330MB on CI), if necessary reduce retention period:
-          #retention-days: 1
 
   job-run-tests:
     name: 'Run Test Suite'
@@ -131,15 +124,31 @@ jobs:
           # Required to retrieve bats (core + extras):
           submodules: recursive
 
-      - name: 'Retrieve DMS test image'
-        uses: actions/download-artifact@v3
+      # Get the cached build layers from the build job:
+      # This should always be a cache-hit, no new uploads should happen when this job finishes:
+      - name: 'Retrieve image build from build cache'
+        uses: actions/cache@v3
         with:
-          name: dms-build-${{ needs.job-build-image.outputs.image-build-key }}
-          path: /tmp
+          path: /tmp/.buildx-cache
+          key: cache-buildx-${{ needs.job-build-image.outputs.image-build-key }}
+          restore-keys: |
+            cache-buildx-
 
-      - name: 'Import retrieved image'
-        run: |
-          docker load --input /tmp/dms-docker-img.tar
+      # Importing from the cache should create the image within approx 30 seconds:
+      # buildx not needed as no exporting and only single AMD64 platform is loaded:
+      - name: 'Build AMD64 image from cache'
+        uses: docker/build-push-action@v3.1.1
+        with:
+          context: .
+          build-args: |
+            VCS_REF=${{ github.sha }}
+            VCS_VER=${{ github.ref }}
+          tags: mailserver-testing:ci
+          # Export the built image for the Docker host to use:
+          load: true
+          # Rebuilds the AMD64 image from the cache:
+          platforms: linux/amd64
+          cache-from: type=local,src=/tmp/.buildx-cache
 
       - name: 'Run tests'
         env:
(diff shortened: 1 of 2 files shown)


georglauterbach:
Very nice and welcome changes / additions!

Would just adding another job (dependent on whether the AMD64 build was successful maybe) that builds the ARM64 image in parallel to the BATS test suite cut down on the build time (the 13 minutes)? Is there something I am missing, that would prevent this solution?

polarathene:
> Is there something I am missing, that would prevent this solution?

You are not wrong, I am just really stretched for time now. I am enrolled for a course until december with 40-80 hour weeks demanded from students, and need to be prepared before it starts in a few days.

You're welcome to adjust the PR or add a follow-up change :+1: 

I mostly didn't want to copy/paste more of the same step as we have similar jobs in other workflows. Github docs mention a thing called "re-usable workflows" that might be appropriate way of keeping that DRY. Alternatively, we could also explore a `matrix` job which would spin up two separate runners in parallel to perform the same job but allow us to adjust the platform used. That would also work while keeping code minimal, but need some condition so the test job only runs afterwards for AMD64 build is done.

polarathene:
Oh and the test failure is probably related to the recent PR, tripping up more timing sensitive tests :(

georglauterbach (approved):
LGTM 👍🏼

casperklein:
I didn't have time yet to look at this content wise. Just some stylistic questions so far:

1. In [this](https://github.com/docker-mailserver/docker-mailserver/pull/2737/commits/0daea44aa2220958bbec1284cf0021230c77cabf) previous PR, quotes were removed (line 1). This PR add quotes (line 27).
2. What is the rational behind naming stages "stage" `--> stage-base` and jobs "job" `--> job-build-image`? Following that, the steps should be named "step" `--> step-Checkout`. On the other hand, we don't name our functions `function_` etc. Unless there is a good reason to so, I would stick with not including keywords in namings.

polarathene:
> 1. Using quotes seems a good practice.

Probably just different style preference between me and @georglauterbach , happy to be consistent with whatever preference :+1: 

I don't think it makes any difference in this case (_at least for `name:` value_), but quoted value stands out better by triggering different syntax highlighting.

---

> What is the rational behind naming stages
> Unless there is a good reason to so, I would stick with not including keywords in namings.

Fair. In both cases I chose to prefix the name with the scope/type of value for added clarity, as they're a fair bit apart.

I'm not sure how familiar other maintainers are with multi-job workflows or a multi-stage `Dockerfile`, so it was to assist with review :sweat_smile: 

Happy to remove it if you like, `ctrl + f` / `grep` works just as well to get the context :+1:

casperklein:
> needs: job-build-image (makes it clear if it wasn't that needs is referencing a job dependency, whereas build-image might seem more ambiguous)

Ack 👍 

> FROM stage-base AS stage-final (more evident that it's referencing stages, not images. FROM base AS final may seem a tad ambiguous)

Fair. It's just that I've mostly see `FROM base AS final` in the wild (and was fine with it).

>Happy to remove it if you like, ctrl + f / grep works just as well to get the context 👍

If it's only me, keep it 😄

casperklein (approved):
Good work! I haven't tested the CI part, but I think you know what you do 😉

polarathene:
> Fair. It's just that I've mostly see `FROM base AS final` in the wild (and was fine with it).

Yeah I'm fine with that too and don't mind doing so on my own Dockerfiles :+1: 

I've only been a bit more verbose since it's harmless and may help maintainers / contributors that aren't as familiar :)

I'll merge as-is. If any maintainer wants to remove the `job-` or `stage-` prefixes as redundant, I'm happy to add approval to a PR for that.

casperklein:
Another question for clarity: The tests have been moved to a new job (job-run-tests). There the image is build again (from cache).

In `job-build-image` the cache is replaced. Can this lead to problems/race conditions, when more than one `job-build-image` is running? For example, when there are two or more PR opened at the same time?

polarathene:
> Can this lead to problems/race conditions, when more than one `job-build-image` is running?

That's an interesting question :)

The **TL;DR** version is not that I know of.

This is just a build cache, changes to Dockerfile or content that `docker build` would invalidate locally happen here too. We're just providing the previous building blocks to finish the image, not a cache of final result and skipping builds (_The `stage-final` portion is likely to often be invalidated due to `ARG` reflecting the git commit SHA_).

- I have the build layer cache assigned a cache key derived from the file content in `target/` + `Dockerfile` + `VERSION`, so if nothing changes, they should use the same cache contents, otherwise they retrieve different caches.
- It should be unlikely that we'd experience a cache eviction (_AMD64 + ARM64 = `~660MB` per cache stored, least used should be evicted from the total 10GB repo cache storage limit_), but if it did, then we'd just re-build from scratch at this step.
- What also happens when there is no direct cache hit, we will fallback to the last used cache key (_anything with the key name that excludes the content hash suffix_) instead (_aka `restore-keys` used_). That may provide cache up to a certain point for the build, and then the build will re-create any layers necessary and upload that new cache afterwards (_assuming build was successful_).

---

When I wrote this improvement I did encounter one failure case. This was due to the `ARG` usage previously being too early.

The image would build successfully, and then upload the cache. Then any future build where the `ARG` changed (basically any new commit), the cache hit was successful, but all of it was invalidated from the start due to the `ARG`...

So it rebuilt from scratch each time, and at the end of the job refused to update the cache for that key because it was a successful cache hit (_only applies to exact match for cache key, `restore-keys` as fallbacks would update_).

polarathene:
> For example, when there are two or more PR opened at the same time?

While I was careful to handle that here as I also had that same concern. I did notice I messed up with the doc previews workflow.

It uploads an artifact to a common name, I can't recall if it's scoped to the workflow or gets overwritten by other PRs building doc previews in parallel. This provides the docs build result, and PR metadata, I don't think I've seen it yet get mixed up, but might be worth investigating :sweat_smile: 

Thankfully it's just for doc previews and shouldn't cause any risks.

PR #2747, opened 2022-08-28
Streamline COPY statements in Dockerfile

casperklein:
# Description

Nothing big.. Just streamlining the `COPY` usage In `Dockerfile`. Some statements used `./`, while some did not.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Improvement (non-breaking change that does improve existing functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [ ] If necessary I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #2752, opened 2022-08-29
ci: add cache for ClamAV and use dedicated script for package installation

georglauterbach:
# Description

This PR does two things:

1. Add build caching for ClamAV's bytecode
2. Use a separate script for installing packages

The separate script provides a better overview of the installation process. While I was at it, I removed non-essential packages from the packages list. Therefore, this PR is a breaking change.

Moreover, users that want to build DMS themselves are now required to use a recent version of Docker because we need BuildKit (one needs Docker version `>= 20.10.14`).

## Type of change

## Checklist:


georglauterbach:
I noticed that this PR supersedes #2747. #2747 may be closed.

target/scripts/build/packages.sh
```diff
@@ -0,0 +1,101 @@
+#! /bin/bash
+
+set -eE -u -o pipefail
```
polarathene, line 3: I'm not too good with memory, but I'm probably the odd one out here with active maintainers currently. Is there any value including a brief comment that describes what is accomplished? (Optional)

target/scripts/build/packages.sh
```diff
@@ -0,0 +1,101 @@
...
+
+# shellcheck source=../helpers/log.sh
+source /usr/local/bin/helpers/log.sh
+
+export DEBIAN_FRONTEND=noninteractive
+QUIET='-qq' ; _log_level_is_trace && QUIET='-y'
```
polarathene, line 9: Is this the only place that `DEBIAN_FRONTEND` and `LOG_LEVEL=trace` matters?

target/scripts/build/packages.sh
```diff
@@ -0,0 +1,101 @@
...
+  apt-get "${QUIET}" dist-upgrade
+
+  _log 'trace' 'Updating CA certificates'
+  update-ca-certificates
+}
+
+function _install_postfix() {
+  _log 'debug' 'Installing Postfix'
+
+  _log 'warn' "Applying workaround for Postfix bug (see https://github.com/docker-mailserver/docker-mailserver/issues/2023#issuecomment-855326403)"
```
polarathene, line 33:
```suggestion
  _log 'warn' 'Applying workaround for Postfix bug (see https://github.com/docker-mailserver/docker-mailserver/issues/2023#issuecomment-855326403)'
```

target/scripts/build/packages.sh
```diff
@@ -0,0 +1,101 @@
...
+
+  mv /bin/hostname /bin/hostname.bak
+  echo "echo 'docker-mailserver.invalid'" >/bin/hostname
+  chmod +x /bin/hostname
```
polarathene, line 37: Just as a note, I think `docker build` offers a `--hostname` arg now which might allow to remove the workaround. Don't worry about that for this PR though.

target/scripts/build/packages.sh
```diff
@@ -0,0 +1,101 @@
...
+  local RSPAMD_GPG_KEY_LOCATION='/etc/apt/trusted.gpg.d/rspamd.gpg'
+  curl -sSfL https://rspamd.com/apt-stable/gpg.key \
+    | gpg --dearmor >"${RSPAMD_GPG_KEY_LOCATION}"
+  echo \
+    "deb [arch=amd64 signed-by=${RSPAMD_GPG_KEY_LOCATION}] http://rspamd.com/apt-stable/ bullseye main" \
+    >/etc/apt/sources.list.d/rspamd.list
+
+  _log 'trace' 'Updating package signatures again'
+  apt-get "${QUIET}" update
+}
```
polarathene, line 55: This is new and looks like it could be deferred to a separate PR (_better for `git blame` / context_).

My preference (_unless you know of an issue otherwise_) would be to have `rspamd` managed via it's own stage, like `clamav` and `fail2ban` should be. For now though, I think this is fine, but should be introduced via a separate PR.

target/scripts/build/packages.sh
```diff
@@ -0,0 +1,101 @@
...
+    amavisd-new binutils cron dbconfig-no-thanks \
+    clamav clamav-daemon \
+    "${DOVECOT_PACKAGES[@]}" dumb-init \
+    fetchmail fail2ban gzip htop iproute2 \
+    locales logwatch libldap-common libsasl2-modules netcat-openbsd nftables \
+    opendkim opendkim-tools opendmarc \
+    pflogsumm postgrey p7zip-full \
+    postfix-ldap postfix-pcre postfix-policyd-spf-python \
+    postsrsd pyzor razor redis-server rspamd rsyslog sasl2-bin \
+    spamassassin supervisor uuid xz-utils
```
polarathene, line 83: Likewise, it would be good to separate packages by features. It's not always clear why we have some packages (_and with time gets harder to track via `git blame`_).

- I know that `uuid` was added for the current file locking scripts.
- `clamav` + `clamav-daemon` could be by itself or for the time being mixed with `amavisd-new` + `spamassassin`. I think `p7zip-full` is for `amavisd-new` support.
- `libldap-common`, `postfix-ldap`, `libsasl2-modules`, `sasl2-bin` for LDAP (_and Cyrus based SASL? One might not be_ :man_shrugging: )
- `redis-server` stood out. This DB package is for `rspamd` I assume? (_highlights why I'd like the separation / scoping to features/dependencies_)
- You've dropped quite a few packages...why? (**EDIT:** [Mentioned in this commit](https://github.com/docker-mailserver/docker-mailserver/pull/2752/commits/f9961baf2861ad53e5a421b1ad0e0621c27a79ec)) These provide support to Amavis AFAIK, it lists all the supported codecs found during the startup logs, ClamAV maybe also uses it to check for viruses? Is this something you can confidently drop? (_with proper stage separation, we can halve the weight of the image if we publish an alternative antivirus tag variant, or the other way around with a lightweight `core` base_)

Some of the packages dropped aren't related to the archive support that the commit messages justifies the removal with. **I am not comfortable this being part of the PR.**

Please maintain the existing package list as-is, and defer the `rspamd` ones to a separate PR for introducing `rspamd` support. A separate PR may follow this one to introduce the breaking change of removing packages, and that is all that PR should cover, preferably the scoping/grouping would already be laid out here.

target/scripts/helpers/log.sh
```diff
@@ -42,32 +42,22 @@ RESET=$(echo -ne   '\e[0m')
...
   if [[ -z ${2+set} ]]
   then
-    echo "Call to '_log' is missing a message to log" >&2
+    _log 'error' "Call to '_log' is missing a message to log"
     return 1
   fi
 
-  local MESSAGE LEVEL_AS_INT LOG_LEVEL_FALLBACK
-  MESSAGE="${RESET}["
+  local LEVEL_AS_INT MESSAGE="${RESET}["
```
polarathene, line 58: No need to squeeze this into one line. I don't mind it when it's declaration only, but not a fan with assignment involved:

```suggestion
  local LEVEL_AS_INT
  local MESSAGE="${RESET}["
```

target/scripts/helpers/log.sh
```diff
@@ -116,4 +106,24 @@ function _log
...
+
+function _get_log_level_or_default() {
+  if [[ -n ${LOG_LEVEL+set} ]]
+  then
+    printf '%s' "${LOG_LEVEL}"
+  elif [[ -e /etc/dms-settings ]]
+  then
+    local LOG_LEVEL
+    LOG_LEVEL=$(grep "^LOG_LEVEL=" /etc/dms-settings | cut -d '=' -f 2)
+    printf '%s' "${LOG_LEVEL:1:-1}"
```
polarathene, line 121: I tried to follow what `LOG_LEVEL:1:-1` is intended for since you compare to strings... I looked at the original PR [commit that introduced this](https://github.com/docker-mailserver/docker-mailserver/pull/2506/commits/f0668344379b742839ccaf64fa4f1834815164fc), but didn't see anything to explain that either.

I assume it's for triggering the `*` case?

target/scripts/helpers/log.sh
```diff
@@ -42,32 +42,22 @@ RESET=$(echo -ne   '\e[0m')
 # If the first argument is not set or invalid, an error
 # message is logged. Likewise when the second argument
 # is missing. Both failures will return with exit code '1'.
-function _log
-{
+function _log() {
```
polarathene, line 45: I saw your [associated commit message](https://github.com/docker-mailserver/docker-mailserver/pull/2752/commits/d5e33f5f9ac36325a2f43d66f635ab8da0486adf) for this, but I'm not sure what you're referencing?

We generally keep to that style now don't we? Has there been complaints somewhere? I think the only case that doesn't happen is with `.bats`, or in my case if I forget to style it that way and it slips through review (_in languages I'm used to, it's common to take this approach with `{` on the same line_).

Personally, unless there's a good reason to, we should already be pretty consistent with the function style used, I don't think we need to slowly lose that consistency (_or do a big PR to change it_). If another maintainer wants that, I'm fine with it, but it is a gripe I have when trawling history via `git blame` for more information on something to wade through these kind of broad sweeping adjustments.

polarathene (changes requested):

.github/workflows/default_on_push.yml
```diff
@@ -46,8 +46,8 @@ jobs:
           context: .
           file: ./Dockerfile
           build-args: |
-            VCS_REF=${{ github.sha }}
-            VCS_VER=${{ github.ref }}
+            VCS_REVISION=${{ github.sha }}
+            VCS_VERSION=${{ github.ref }}
```
polarathene, line 50: In reference to [associated commit message](https://github.com/docker-mailserver/docker-mailserver/pull/2752/commits/05c533e8dd0a065367c1b995522be2d4fd7a128a):

I cannot track down the PR (perhaps it was a direct commit) that introduced this for more information.

Only where it was introduced before these workflow files with just [`Makefile` usage](https://github.com/docker-mailserver/docker-mailserver/pull/1256#issuecomment-529406724). It looks like whomever added this to the workflows misunderstood it?

I don't see why `VCS_VERSION` is assigned `github.ref`, but maybe my memory is bad and that will use a tag when it's a tag event. Otherwise AFAIK it would just be the branch ref it's running from.

It's intention was to just bake in the info into image metadata so it could be queried and look like:

```
org.label-schema.vcs-ref=7580aec
org.label-schema.version=v6.1.0-111-g7580aec
```

---

`VCS_REF` probably referred to [`git ref`](https://git-scm.com/book/en/v2/Git-Internals-Git-References). That's the commit SHA for the inverse of `github.ref` that `VCS_VER` has been using here (_while the `Makefile` instead uses [`git describe`](https://git-scm.com/docs/git-describe), so isn't consistent_).

polarathene:
Sorry for lack of overview / summary, browser was acting up and I thought I almost lost the review before submitting so rushed it :sweat_smile:

I don't have enough time to provide one right now :weary: I also may not be as available until later in the week/weekend.

> 2\. Use a separate script for installing packages

I am not sure if this is something I've seen in `Dockerfile` following best practices. I am not sure if that causes different behaviour for `RUN` caching? The directive itself won't change, but I think when the file itself does it's ok, so it should be fine?

Dockerfile
```diff
@@ -1,87 +1,49 @@
...
-ARG FAIL2BAN_GPG_FINGERPRINT="8738 559E 26F6 71DF 9E2C  6D9E 683B F1BE BD0A 882C"
+#
+# Base stage provides all packages, config, and adds scripts
+#
 
+FROM ${BASE_IMAGE} AS stage-base
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
+ENV LOG_LEVEL=trace
+ENV DEBIAN_FRONTEND=noninteractive
```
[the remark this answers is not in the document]
casperklein: I was about to suggest the same, use `ARG` here 👍

Dockerfile
```diff
@@ -1,87 +1,49 @@
...
+RUN /build/packages.sh
+
+COPY target/bin/sedfile /usr/local/bin/sedfile
 
 # -----------------------------------------------
 # --- ClamAV & FeshClam -------------------------
 # -----------------------------------------------
 
+# hadolint ignore=DL3021
+COPY --link --from=stage-clamav /var/lib/clamav /var/lib/clamav
```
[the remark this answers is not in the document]
casperklein: I think, copying the database from the official image is a good idea, instead of running `freshclam` on our own. Running freshclam is rate limited. I didn't check it, if that was already the case in the past, but theoretically we can end with an image without a current DB due to rate limits by clamav. At least while testing/building locally, I run very quickly in that rate limit.

target/scripts/build/packages.sh
casperklein: `rspamd` is not scope of that PR and should be done separately.

casperklein:
It was said somewhere else already and I second this again: To make reviews easier, keep them as small as possible and avoid changes not directly needed for the goal a PR achieves (such as beautifying things or introducing other little additions like rspamd).
Like the unix philosophy: do one thing well 👍

georglauterbach:
I see, my PR became a bit bit and not UNIX-oid.. I will provide smaller ones that incorporate the review feedback - thank's @polarathene :)

Dockerfile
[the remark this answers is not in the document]
georglauterbach: 👍🏼

Dockerfile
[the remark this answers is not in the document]
georglauterbach: Nice observation. I will see what I can do.

Dockerfile
```diff
@@ -310,7 +273,10 @@ LABEL org.opencontainers.image.description="A fullstack but simple mail server (
 LABEL org.opencontainers.image.url="https://github.com/docker-mailserver"
 LABEL org.opencontainers.image.documentation="https://github.com/docker-mailserver/docker-mailserver/blob/master/README.md"
 LABEL org.opencontainers.image.source="https://github.com/docker-mailserver/docker-mailserver"
+
 # ARG invalidates cache when it is used by a layer (implicitly affects RUN)
 # Thus to maximize cache, keep these lines last:
-LABEL org.opencontainers.image.revision=${VCS_REF}
-LABEL org.opencontainers.image.version=${VCS_VER}
+ARG VCS_REVISION
+ARG VCS_VERSION
```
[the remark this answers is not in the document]
georglauterbach: Just style, can be removed. I will not move them in upcoming PRs.

Makefile
```diff
@@ -11,7 +11,9 @@ VCS_VER = $(shell git describe --tags --contains --always)
 all: lint build backup generate-accounts tests clean
 
 build:
-	docker build -t $(NAME) . --build-arg VCS_VER=$(VCS_VER) --build-arg VCS_REF=$(VCS_REF)
+	DOCKER_BUILDKIT=1 docker build -t $(NAME) . \
+		--build-arg VCS_REVISION=$(VCS_REVISION) \
+		--build-arg VCS_VERSION=$(VCS_VERSION)
```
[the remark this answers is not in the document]
georglauterbach: 👍🏼

VERSION
```diff
@@ -1 +1 @@
-11.1.0
+12.0.0-alpha.1
```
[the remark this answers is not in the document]
georglauterbach: This is indeed an accident and should not have happened :D

target/scripts/build/packages.sh
georglauterbach: I can provide one :)

target/scripts/build/packages.sh
georglauterbach: Yes, so we can probably remove them entirely from the Dockerfile of we use a dedicated script.

target/scripts/build/packages.sh
georglauterbach: 👍🏼

target/scripts/build/packages.sh
georglauterbach: Agreed!

target/scripts/build/packages.sh
georglauterbach: I see. I will not touch the packages :)

PS: You're right: `redis-server` is for `rspamd`

target/scripts/helpers/log.sh
georglauterbach: Agreed. I always thought people hate me for doing this. I am used to writing it like this in C and Rust.. and I am happy to uphold this style :D 😆

casperklein: > I always thought people hate me for doing this.

I do 😆 It's just that I don't like wasting a line for no benefits. The some goes for if statements, I prefer for simplicity:

```bash
if true; then
    foo
fi
```

So if there are plans to change that, I am all for it 👍 However I understand that I am not the only maintainer, so my personal style wishes may not always meet those of the others. Most important for me is consistency across our code base, even if I don't like the decision that was chosen 😆

target/scripts/helpers/log.sh
casperklein: `grep "^LOG_LEVEL=" /etc/dms-settings | cut -d "'" -f 2` should also work

Dockerfile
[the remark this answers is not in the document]
casperklein: It should be as simple as this:

`COPY --link --from=clamav/clamav:latest /var/lib/clamav /var/lib/clamav`

target/scripts/helpers/log.sh
georglauterbach: @casperklein I have no problem with a PR that re-styled _the whole_ code base at once 😆 But I guess this also has not benefit. I would just sick to the style I imposed upon all of you 😆 😆

polarathene: > Agreed. I always thought people hate me for doing this.

Oh, totally not a fan, but since you've already converted the code-base I just go with it. Not a big deal.

> Most important for me is consistency across our code base, even if I don't like the decision that was chosen :laughing: 

This.

---

If someone wants to do that mundane task of switching a style through all the files, go for it... Just wait until after a release. I especially don't want to have to rebase my WIP PR against it :sweat_smile:

target/scripts/helpers/log.sh
georglauterbach: > `grep "^LOG_LEVEL=" /etc/dms-settings | cut -d "'" -f 2` should also work

I implemented this in #2754 :)

PR #2754, opened 2022-08-30
scripts: improve `helpers/log.sh`

georglauterbach:
# Description

This PR prepares for other PRs that use the newly introduced helper
functions. The `_log` function itself was adjusted a bit to be shorter
and more concise.

Follow up of #2752 taking care of `log.sh`.

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/helpers/log.sh
+++ b/target/scripts/helpers/log.sh
@@ -56,7 +56,8 @@ function _log
     return 1
   fi
 
-  local LEVEL_AS_INT MESSAGE="${RESET}["
+  local LEVEL_AS_INT
+  local MESSAGE="${RESET}["
 
   case "$(_get_log_level_or_default)" in
     ( 'trace' ) LEVEL_AS_INT=5 ;;
--- a/target/scripts/helpers/log.sh
+++ b/target/scripts/helpers/log.sh
@@ -125,9 +125,7 @@ function _get_log_level_or_default
     printf '%s' "${LOG_LEVEL}"
   elif [[ -e /etc/dms-settings ]]
   then
-    local LOG_LEVEL
-    LOG_LEVEL=$(grep "^LOG_LEVEL=" /etc/dms-settings | cut -d '=' -f 2)
-    printf '%s' "${LOG_LEVEL:1:-1}"
+    grep "^LOG_LEVEL=" /etc/dms-settings | cut -d "'" -f 2
   else
     printf 'info'
   fi
--- a/target/scripts/helpers/log.sh
+++ b/target/scripts/helpers/log.sh
@@ -46,28 +46,19 @@ function _log
 {
   if [[ -z ${1+set} ]]
   then
-    echo "Call to '_log' is missing a valid log level" >&2
+    _log 'error' "Call to '_log' is missing a valid log level"
     return 1
   fi
 
   if [[ -z ${2+set} ]]
   then
-    echo "Call to '_log' is missing a message to log" >&2
+    _log 'error' "Call to '_log' is missing a message to log"
     return 1
   fi
 
-  local MESSAGE LEVEL_AS_INT LOG_LEVEL_FALLBACK
-  MESSAGE="${RESET}["
+  local LEVEL_AS_INT MESSAGE="${RESET}["
 
-  if [[ -e /etc/dms-settings ]]
-  then
-    LOG_LEVEL_FALLBACK=$(grep "^LOG_LEVEL=" /etc/dms-settings | cut -d '=' -f 2)
-    LOG_LEVEL_FALLBACK="${LOG_LEVEL_FALLBACK:1:-1}"
-  else
-    LOG_LEVEL_FALLBACK='info'
-  fi
-
-  case "${LOG_LEVEL:-${LOG_LEVEL_FALLBACK}}" in
+  case "$(_get_log_level_or_default)" in
     ( 'trace' ) LEVEL_AS_INT=5 ;;
     ( 'debug' ) LEVEL_AS_INT=4 ;;
     ( 'warn'  ) LEVEL_AS_INT=2 ;;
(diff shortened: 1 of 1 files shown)


target/scripts/helpers/log.sh
```diff
@@ -116,4 +108,34 @@ function _log
...
+function _get_log_level_or_default
+{
+  if [[ -n ${LOG_LEVEL+set} ]]
+  then
+    printf '%s' "${LOG_LEVEL}"
+  elif [[ -e /etc/dms-settings ]]
+  then
+    local LOG_LEVEL
+    LOG_LEVEL=$(grep "^LOG_LEVEL=" /etc/dms-settings | cut -d '=' -f 2)
+    printf '%s' "${LOG_LEVEL:1:-1}"
```
polarathene, line 130: Just to clarify, neither of these are setting a fallback value of `1`/`-1` right? This is bash syntax for truncating left/right to remove wrapped quotes of the value stored in `/etc/dms-settings`?

Is there a situation where this ever makes sense to support?

Your last `else` condition is to use the same default `info` as the startup scripts assign via `variables.sh` for a fallback?

The original addition of `/etc/dms-settings` was for a Dovecot quota test that needed to assume to workaround an ENV issue IIRC. In that situation, there isn't a fallback and the ENV may not be set but be assigned a default during startup that could be checked with `/etc/dms-settings`.

target/scripts/helpers/log.sh
```diff
@@ -116,4 +108,34 @@ function _log
...
+
+# Get the value of the environment variable LOG_LEVEL if
+# it is set. Otherwise, try to query the common environment
+# variables file. If this does not yield a value either,
+# use the default log level.
+function _get_log_level_or_default
+{
+  if [[ -n ${LOG_LEVEL+set} ]]
+  then
+    printf '%s' "${LOG_LEVEL}"
```
polarathene, line 125: <details>
<summary>Resolved</summary>

Is it desired for this to work with invalid `LOG_LEVEL` when set? (eg: `LOG_LEVEL=whatever`)

Previously we had:

https://github.com/docker-mailserver/docker-mailserver/blob/e6f481b59f7f67a7f605f9d97fa6a5ee1de68840/target/scripts/helpers/log.sh#L62-L78

Where `( *       ) LEVEL_AS_INT=3 ;;` is effectively only for `info`, as any other value would then hit:

https://github.com/docker-mailserver/docker-mailserver/blob/e6f481b59f7f67a7f605f9d97fa6a5ee1de68840/target/scripts/helpers/log.sh#L103-L107

Which now is an `error` type log.

</details>

---

**EDIT:** Ok, I grok the flow now :sweat_smile: 

- I think `/etc/dms-settings` will still only ever be hit when `LOG_LEVEL` isn't set, thus almost always is `info` (_unless overriding the ENV for commands with `docker exec` or modifying `/etc/dms-settings` for some reason_).
- We don't handle bad `LOG_LEVEL` early (_other than making mistakes more apparent via the `error` log fallback in this PR_), which should always be due to `LOG_LEVEL` being set wrong.
- Due to those two niche scenarios, `log.sh` handles this internally, instead of `variables.sh` at startup.

polarathene (dismissed):

target/scripts/helpers/log.sh
georglauterbach: > Just to clarify, neither of these are setting a fallback value of `1`/`-1` right? This is bash syntax for truncating left/right to remove wrapped quotes of the value stored in `/etc/dms-settings`?

This syntax look very much like the "assign a default value" syntax, I agree, but you are correct, this is _not_ setting any default :) It is just for truncating quotes. @casperklein mentioned that `| cut -f "'" -f 2` would also work. I am fine with both, but maybe you prefer the alternative @polarathene?

> Is there a situation where this conditional for `/etc/dms-settings` ever makes sense to support?
> Resolved (answer: potentially...?)

I guess when you import the `log.sh` helper in a custom script without sourcing `etc/dms-settings`. I was unable to identify a situation where it would do harm though :)

target/scripts/helpers/log.sh
georglauterbach: I guess you provided the solution to your own question (?) :D

polarathene: Yeah, I eventually figured it out :sweat_smile: :+1:

target/scripts/helpers/log.sh
```diff
@@ -116,4 +108,32 @@ function _log
...
+  elif [[ -e /etc/dms-settings ]]
+  then
+    grep "^LOG_LEVEL=" /etc/dms-settings | cut -d "'" -f 2
```
polarathene, line 128: Just a reminder that you added similar functionality in `helper/utils.sh`. Might be good to apply the improvement there too:

https://github.com/docker-mailserver/docker-mailserver/blob/e6f481b59f7f67a7f605f9d97fa6a5ee1de68840/target/scripts/helpers/utils.sh#L15-L22

polarathene (approved):

target/scripts/helpers/log.sh
georglauterbach: Nice observation. I will provide a follow-up PR :)

target/scripts/helpers/log.sh
```diff
@@ -116,4 +108,32 @@ function _log
...
+  else
+    printf 'info'
```
casperklein, line 130: 🤷 

```suggestion
    echo 'info'
```

target/scripts/helpers/log.sh
```diff
@@ -116,4 +108,32 @@ function _log
...
+
+# Get the value of the environment variable LOG_LEVEL if
+# it is set. Otherwise, try to query the common environment
+# variables file. If this does not yield a value either,
+# use the default log level.
+function _get_log_level_or_default
+{
+  if [[ -n ${LOG_LEVEL+set} ]]
+  then
+    printf '%s' "${LOG_LEVEL}"
```
casperklein, line 125: Change it or leave it 😉 

```suggestion
    echo "${LOG_LEVEL}"
```

casperklein (approved):

target/scripts/helpers/log.sh
georglauterbach: Will do in a follow-up PR, so I can merge this first :)

target/scripts/helpers/log.sh
georglauterbach: Will do in a follow-up PR, so I can merge this first :)

PR #2777, opened 2022-09-18
tests: Wait at least 30 seconds before checking the health state of the container

casperklein:
# Description

This PR fixes failed health-checks like [this](https://github.com/docker-mailserver/docker-mailserver/actions/runs/3076842772/jobs/4971288987#step:5:227), when the CI is faster than 30 seconds when running the health check (container isn't healthy yet at this point --> first health check is run after 30 seconds):

```
not ok 219 checking container healthcheck in 113ms
#  in test file test/tests.bats, line 114)
#
# -- output differs --
# expected : healthy
# actual   : starting
```

## Type of change

## Checklist:

### DIFF
--- a/test/tests.bats
+++ b/test/tests.bats
@@ -4,6 +4,7 @@ export IMAGE_NAME
 IMAGE_NAME="${NAME}"
 
 setup_file() {
+  export START_TIME
   local PRIVATE_CONFIG
   PRIVATE_CONFIG=$(duplicate_config_for_container . mail)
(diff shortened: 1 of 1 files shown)


polarathene (approved):
An alternative might be to shift the test to the end of the file (_but should keep the note about 30 sec requirement_), or moving to a separate test file.

casperklein:
> or moving to a separate test file.

I thought about this also, but that would at 30 seconds to the overall test run time. The solution I provided added max 1-2 seconds in my tests and ensures, that the test succeed.

I prefer this solution, as it is bullet proof. Moving the test more to the bottom could help, but might fail in the future again.

georglauterbach (approved):
👍🏼

Issue #217, opened 2016-06-26
Catch-All Address


QuadStingray:
Hi,

when I configure an chatch-all-address all mails for an existing account will delivered to the catch-all adress and not to the account. What I`m doing wrong?

postfix-accounts.cf

`account@example.com|{SHA512-CRYPT}$6$Pi8yxp1ztD/lSkyG$kYd1NizLTUSz7Pc8y3v0b9vxcpaBROTUf0Fn3JVPrxBZftWPJfy2h3n.MhYSVAk6BzDeWfNB1cSJYftWLii.W0`

`account2@example.com|{SHA512-CRYPT}$6$Pi8yxp1ztD/lSkyG$kYd1NizLTUSz7Pc8y3v0b9vxcpaBROTUf0Fn3JVPrxBZftWPJfy2h3n.MhYSVAk6BzDeWfNB1cSJYftWLii.W0`

postfix-virtual.cf
`
@example.com account2@example.com
`

And one other question.
Can I configure that mails will send to an account and forwarded to an other email address?

approached:
Catchall work fine with:
```
@example.com account2@example.com 
```

But you can always use like this:
```
@example.com example@gmail.com 
```

daitangio:
So to configure a catch all rule to my account "account1" I could issue the following setup command?
`./setup.sh alias add @example.com account1@example.com`

(Thank you for this wonderful code by the way!)

PR #2786, opened 2022-09-21
style: minor adjustments

georglauterbach:
# Description

These adjustments are just follow-ups for the latest PRs I merged. **If you have more adjustments, feel free to add them to this PR.**

## Type of change

## Checklist:


polarathene (approved):

wernerfred (approved):

target/scripts/build/packages.sh
```diff
@@ -130,7 +130,7 @@ function _install_fail2ban
   curl -Lkso fail2ban.deb "${FAIL2BAN_DEB_URL}"
   curl -Lkso fail2ban.deb.asc "${FAIL2BAN_DEB_ASC_URL}"
 
-  FINGERPRINT=$(LANG=C gpg --verify fail2ban.deb.asc fail2ban.deb 2>&1 | sed -n 's#Primary key fingerprint: \(.*\)#\1#p')
+  FINGERPRINT=$(LANG=C gpg --verify fail2ban.deb.asc fail2ban.deb |& sed -n 's#Primary key fingerprint: \(.*\)#\1#p')
```
polarathene, line 133: I didn't expect `|&` to have much luck with search engines, but [I stand corrected](https://stackoverflow.com/questions/35384999/what-does-mean-in-bash) :sweat_smile: 

I suppose we have a lot of bash specific syntax throughout the project that a shell like `ash` probably isn't compatible?

polarathene (approved):

target/scripts/build/packages.sh
casperklein: I don't know exactly, how compatible ash/dash etc. are to bash. But as we wrote our scripts for bash, I am very fine with using bash specific syntax to unleash the full power 😆

casperklein: @docker-mailserver/maintainers While we at it: what about replacing the sed command with:

`grep -oP '(?<=Primary key fingerprint: ).+'`

Regex explanation: The `?<=` within the group means --> ignore this group and match only `.+` that comes after.

This seems simpler and easier read (for me at least).

But no need for a lengthy discussion IMO. Just give a 👎 to keep it as it is, or 👍 to go with `grep`.

polarathene: I don't think it's worth it personally. Very minor improvement, and not something I'm going to remember, or like looking up if I saw it months later and forgot (`git blame` aside).

The `sed` command is familiar enough, I've seen the usage here elsewhere in the project, so it's easier to maintain.

polarathene:
@casperklein currently [tests failed with the healthcheck](https://github.com/docker-mailserver/docker-mailserver/actions/runs/3162276455/jobs/5148743130#step:5:227), which I think you tried to address with a recent PR:

```
not ok 219 checking container healthcheck in 115ms
#  in test file test/tests.bats, line 123)
#
# -- output differs --
# expected : healthy
# actual   : starting
# --
#
```

target/scripts/build/packages.sh
georglauterbach: I do not see an inherent need for this change as well - doesn't really matter I think. I'd just like to merge this sometime soon so we can get a feature freeze in place for v11.2.0 :)

casperklein:
I've provided a possible fix for the health check in #2810.

Afaik, that was the first time, that the health check failed since I addressed it recently? More annoying is however the test #217 which pretty often fails. This forces me sometimes to restart the tests three or more times.

casperklein (approved):

georglauterbach:
We will definitely need to do something about the issue where only 217 tests run. The `setup_file` function for `tests.bats` fails, which leads to all the tests in this big file not being run. I will have a look.

polarathene:
When you both experienced the test 217 failures, I assume it was on PRs that would not result in invalidating the ClamAV cache, and that cache may be old enough to have ClamAV update when the container runs?

If so I have noticed on my system how problematic that can be for getting `tests.bats` to be successful.

georglauterbach:
> When you both experienced the test 217 failures, I assume it was on PRs that would not result in invalidating the ClamAV cache, and that cache may be old enough to have ClamAV update when the container runs?

Yes, this is absolutely possible.


> If so I have noticed on my system how problematic that can be for getting `tests.bats` to be successful.

Interesting - I will have a look into this too.

Issue #2791, opened 2022-09-22
[BUG] Dockerfile is not longer compatible with 


ap-wtioit:
### Miscellaneous first checks

- [X] I checked that all ports are open and not blocked by my ISP / hosting provider.
- [X] I know that SSL errors are likely the result of a wrong setup on the user side and not caused by DMS itself. I'm confident my setup is correct.

### Affected Component(s)

not buildable

### What happened and when does this occur?

```Markdown
Error response from daemon: Dockerfile parse error line 43: Unknown flag: link
```

### What did you expect to happen?

I expected to be able to build this image on Ubuntu 18.04 LTS with docker:
```
Client: Docker Engine - Community
 Version:           20.10.18
 API version:       1.40
 Go version:        go1.18.6
 Git commit:        b40c2f6
 Built:             Thu Sep  8 23:11:34 2022
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          19.03.14
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       5eb3275d40
  Built:            Tue Dec  1 19:18:45 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.8
  GitCommit:        9cd3357b7fd7218e4aec3eae239db1f68a5a6ec6
 runc:
  Version:          1.1.4
  GitCommit:        v1.1.4-0-g5fd4c4d
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

### How do we replicate the issue?

```Markdown
1. Use Ubuntu 18.04
2. Use Docker
3. try to build image
4. Get the following error:

Error response from daemon: Dockerfile parse error line 43: Unknown flag: link

```

### DMS version

dockerfile from master

### What operating system is DMS running on?

Linux

### What instruction set architecture is DMS running on?

x86_64 / AMD64

### What container orchestration tool are you using?

Docker

### docker-compose.yml

```yml
--
```

### Relevant log output

```bash
Error response from daemon: Dockerfile parse error line 43: Unknown flag: link
```

### What level of experience do you have with Docker and mail servers?

- [ ] I am inexperienced with docker
- [ ] I am inexperienced with mail servers
- [ ] I am uncomfortable with the CLI

### Code of conduct

[...]

PR #2799, opened 2022-09-26
docs: add note about Docker version when building

georglauterbach:
# Description

See #2791. Adds a note that we require a recent version of Docker when building.

## Type of change

## Checklist:

### DIFF
--- a/docs/content/examples/tutorials/docker-build.md
+++ b/docs/content/examples/tutorials/docker-build.md
@@ -25,6 +25,22 @@ Retrieving the git submodules will fix the error:
 COPY failed: file not found in build context or excluded by .dockerignore: stat target/docker-configomat/configomat.sh: file does not exist
 ```
 
-### Docker Version
+### About Docker
+
+#### Version
 
 We make use of build-features that require a recent version of Docker. Depending on your distribution, please have a look at [the official installation documentation for Docker](https://docs.docker.com/engine/install/) to get the latest version. Otherwise, you may encounter issues, for example with the `--link` flag for a `#!dockerfile COPY` command.
+
+#### Environment
+
+If you are not using `make` to build the image, note that you will need to provide `DOCKER_BUILDKIT=1` to the `docker build` command for the build to succeed.
+
+#### Build Arguments
+
+The `Dockerfile` takes additional, so-called build arguments. These are
+
+1. `VCS_VERSION`: the image version (default = edge)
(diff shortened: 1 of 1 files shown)


casperklein:
The mentioned issue was not due the Docker version, but the non usage of `DOCKER_BUILDKIT=1` while building.

georglauterbach:
> The mentioned issue was not due the Docker version, but the non usage of `DOCKER_BUILDKIT=1` while building.

The initial issue suggested otherwise. Regardless, I think this addition is well worth it :) We could add the note about `DOCKER_BUILDKIT=1` as well!

docs/content/examples/tutorials/docker-build.md
```diff
@@ -12,11 +14,33 @@ docker build -t mailserver/docker-mailserver .
...
+
+```txt
 COPY failed: file not found in build context or excluded by .dockerignore: stat target/docker-configomat/configomat.sh: file does not exist
 ```
+
+### About Docker
+
+#### Version
+
+We make use of build-features that require a recent version of Docker. Depending on your distribution, please have a look at [the official installation documentation for Docker](https://docs.docker.com/engine/install/) to get the latest version. Otherwise, you may encounter issues, for example with the `--link` flag for a `#!dockerfile COPY` command.
```
casperklein, line 32: `#!dockerfile COPY`

Is `#!dockerfile` a typo?

```suggestion
We make use of build-features that require a recent version of Docker. Depending on your distribution, please have a look at [the official installation documentation for Docker](https://docs.docker.com/engine/install/) to get the latest version. Otherwise, you may encounter issues, for example with the `--link` flag for a Dockerfile [COPY](https://docs.docker.com/engine/reference/builder/#copy) command.
```

georglauterbach: No, not a type: <https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#highlighting-inline-code-blocks>.

But we can also provide the link you mentioned. I think this is a good all-in-one solution.

georglauterbach: resolved in [b78deb8](https://github.com/docker-mailserver/docker-mailserver/pull/2799/commits/b78deb8ccd0b3cbe5e9ee283c3475ee2642112a2)

casperklein (approved):

wernerfred (approved):

PR #2811, opened 2022-10-02
ci: attempt to fix recent transient test failures

georglauterbach:
# Description

This adds more waiting barriers to allow a bit more time to pass in order to prevent [this](https://github.com/docker-mailserver/docker-mailserver/actions/runs/3166230210/jobs/5155809372).

Moreover, I added a barrier for the test of the `setup email del` command, which failed from time to time as well.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [x] If necessary I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #2815, opened 2022-10-05
ci: misc test enhancements

georglauterbach:
# Description

Fresh start of #2811

## Type of change

## Checklist:


.gitignore
```diff
@@ -52,3 +52,4 @@ config.bak
 testconfig.bak
 
 docs/site
+tmp.bats
```
polarathene, line 55: I assume this addition wasn't meant to be part of the PR?

I'm going to revert it due to checksuite being stuck on lint job and no apparent action actually running to restart it.

georglauterbach: Yes, this was not intended. GH actions experience degraded performance yesterday so I had to wait :)

georglauterbach:
@polarathene  do you have any idea, btw, why the test job does not use the cache from the build job?

casperklein:
>  `sleep`s is not a permanent solution, but for the time being, better than an always-failing CI.

I haven't looked in detail in this current issue. But another pragmatic approach could be: find out, when the failing tests started and then revert the last changes made to the tests? That should fix it or am I wrong.

Anything was introduced that caused that issue. So instead of fixing the symptoms, I would roll back the change that introduced the issue in the first place.

georglauterbach:
> >  `sleep`s is not a permanent solution, but for the time being, better than an always-failing CI.
> 
> I haven't looked in detail in this current issue. But another pragmatic approach could be: find out, when the failing tests started and then revert the last changes made to the tests? That should fix it or am I wrong.

IIRC #2729 showed that CI is fragile. While I do not think #2729 introduced a bug, it unveiled the fragility. That's why it is hard to pinpoint: there is no obvious that serves as the cause.

> Anything was introduced that caused that issue. So instead of fixing the symptoms, I would roll back the change that introduced the issue in the first place.

As I said above, this works in theory, but it is more or less difficult in this case for the reasons mentioned above.

polarathene:
> @polarathene do you have any idea, btw, why the test job does not use the cache from the build job?

The build job did not create a new cache. It had a primary key cache hit. The file contents that derive the cache key has been seen by the CI before.

When this happens, if something has changed that invalidates that cache, unfortunately the cache action does not upload a new cache, this only happens when the expected cache key for the contents (the primary key that we used) didn't have a match and a fallback cache key was used.

So we restored a previous cached build, that for some reason uploaded the build cache that for the same key still got it invalidated. I only remember this happening when I worked on the feature and `ARG` was invalidating the build, or there was something causing incorrect cache key generation.

---

I am wondering [if we implemented the re-usable workflows correctly](https://github.com/docker-mailserver/docker-mailserver/blob/ff969509f8b9c49e399327887c9a29f6fd664b80/.github/workflows/test_merge_requests.yml), or if they're incompatible here. I'd need to check the cache key generated is for this PR contents or if it was generating that from `master`?

Cache key CI used: `cache-buildx-a0ceadd91401151c4991b14c3f1e2d38e77032937df0f7d9dcf9a12582b50b89`

I just noticed this is quite a problem with [commits on master often failing to pass/complete the Build, Test & Deploy workflows](https://github.com/docker-mailserver/docker-mailserver/commits/master).

polarathene:
~~[As suspected](https://github.com/docker-mailserver/docker-mailserver/actions/runs/3168322443/jobs/5159615148#step:6:14)...~~ (_**EDIT:** Tracked down where the key was from and why it hasn't changed, totally fine_)

> `Cache restored from key: cache-buildx-a0ceadd91401151c4991b14c3f1e2d38e77032937df0f7d9dcf9a12582b50b89`

That is the same cache key used by this PR CI build/test, as was used on Master for merging of [this recent PR](https://github.com/docker-mailserver/docker-mailserver/pull/2786). That was also the [cache key used in that PRs last commits](https://github.com/docker-mailserver/docker-mailserver/actions/runs/3162276455/jobs/5148727919#step:15:5) (_merges from master, that was the specific job that generated the cache key quoted above_) which isn't too surprising. We can also see that non-merge commits in that PR was [using different cache keys as they should](https://github.com/docker-mailserver/docker-mailserver/actions/runs/3154743646/jobs/5133307380#step:4:16) (_this ones build step restored a fallback key, then generated a new one, that was used to restore cache in future commits on the PR and generate the same cache key for the image build being used here_).

Oh ok, just noticed that this PR doesn't contribute anything that should cause a new image cache key to be created, so **it appears to be behaving correctly**.

---

Looking back at the related PR merge commit on master and the CI build job there, it does not have the known build cache key from the PR due to the security scoping stuff, it needs to do a fresh build/cache upload of that. And we see that it selects and restores the `cache-buildx` cache key, which isn't valid but was created as a bug AFAIK, if that hasn't been evicted after 7 days of whenever it was created, Github might renew the expiry if it's accessed within the past 7 days? 

Anyway, that cache from whenever that was, [is invalidated and we get a fresh build](https://github.com/docker-mailserver/docker-mailserver/actions/runs/3168322443/jobs/5159441037#step:4:14). At the end of the build job, since we did not restore a primary cache key for the given contents, it [uploads to master branch cache the same key](https://github.com/docker-mailserver/docker-mailserver/actions/runs/3168322443/jobs/5159441037#step:15:5) this PR is using. Now any branch/PR that derives this cache key should have access to that, and it should be the most recent fallback cache key used too presently.

I am not sure what exactly is causing this invalidation, but I can see that your previous version of this PR was able to build using the cache](https://github.com/docker-mailserver/docker-mailserver/actions/runs/3170398519/jobs/5162954169#step:7:160) for this same cache key:

```
#9 [internal] load metadata for docker.io/clamav/clamav:latest
#9 DONE 0.2s

#8 [internal] load metadata for docker.io/library/debian:11-slim
#8 DONE 0.4s

#10 importing cache manifest from local:3943879211245529940
#10 DONE 0.0s

#11 FROM docker.io/clamav/clamav:latest@sha256:0353dfffc8e8d6e035208e380cdc0f7f64299b849acef4b20ca8d87d874ba409
#11 resolve docker.io/clamav/clamav:latest@sha256:0353dfffc8e8d6e035208e380cdc0f7f64299b849acef4b20ca8d87d874ba409 done
#11 DONE 0.0s

#12 [stage-base  1/47] FROM docker.io/library/debian:11-slim@sha256:5cf1d98cd080595[148](https://github.com/docker-mailserver/docker-mailserver/actions/runs/3170398519/jobs/5162954169#step:7:151)4f33b34c1ab25aac7007bb41c8b9901d97e4be3cf3ab04
#12 resolve docker.io/library/debian:11-slim@sha256:5cf1d98cd0805951484f33b34c1ab25aac7007bb41c8b9901d97e4be3cf3ab04 0.0s done
#12 DONE 0.0s

#13 [internal] load build context
#13 transferring context: 290.82kB 0.0s done
#13 DONE 0.1s

#14 [stage-base 21/47] RUN <<EOF (chmod 755 /etc/init.d/postgrey...)
#14 CACHED
```

Your last commit on that PR that built with cache was: [`properly provide ENV`](https://github.com/docker-mailserver/docker-mailserver/pull/2811/commits/5a2753cf6cfdfc91a0a8f55ffe46cc3fa12a8395), afterwards it started invalidating cache. **Nothing stands out as the cause** though.

polarathene:
> I'm not for introducing `sleep` as well, but unless someone else finds the cause (which may take time), I prefer it over restarting CI 3 or 4 times.

It's fine for now and I doubt we're going to benefit much time wise by reverting. It's a minor hit considering how long tests normally take, and we know that the `sleep` part aside, we're not implicitly hiding the timing problems that were surfaced by recent additions to `tests.bats`.

**However**, we can do better:

https://github.com/docker-mailserver/docker-mailserver/blob/ff969509f8b9c49e399327887c9a29f6fd664b80/test/test_helper/common.bash#L120-L122

https://github.com/docker-mailserver/docker-mailserver/blob/ff969509f8b9c49e399327887c9a29f6fd664b80/test/test_helper/common.bash#L204-L216

https://github.com/docker-mailserver/docker-mailserver/blob/ff969509f8b9c49e399327887c9a29f6fd664b80/test/test_helper/common.bash#L243-L249

https://github.com/docker-mailserver/docker-mailserver/blob/ff969509f8b9c49e399327887c9a29f6fd664b80/test/test_helper/common.bash#L89-L95

These are the methods you're adding sleep around, and apart from the helper I wrote with a fixed `60` seconds (_and advice to wait on postfix/dovecot after separately if necessary_), we can see a common way to avoid the sleeps so that they're only optional when needed:

https://github.com/docker-mailserver/docker-mailserver/blob/ff969509f8b9c49e399327887c9a29f6fd664b80/test/test_helper/common.bash#L8-L9

If 15 secs of sleep is really all that's required, surely we could just bump up the default a bit more. Otherwise I'd expect failures to still occur and increasing `sleep` each time isn't going to be any different? You could bump that up from 2 minutes to 3, more than enough padding.

If we still get failures, it's not a fix that extra time is likely to resolve, and more an indication that something else is going wrong (_[as has been suggested](https://github.com/docker-mailserver/docker-mailserver/pull/2811#issuecomment-1267701466) in the earlier version of this PR for what we could try_).

Additionally as per [my comments here](https://github.com/docker-mailserver/docker-mailserver/pull/2729#issuecomment-1221689781), we can switch `check-for-changes.sh` to have Postfix and Dovecot use `reload` instead of `restart`, as that should minimize any race conditions from the downtime of `restart`. It's quite possible since we are handling change detection in the middle of the setup.

I also suggested considering `wait_for_service "${TEST_NAME}" postfix` directly after `wait_until_change_detection_event_completes` is called if we're to wait on ensuring Postfix is ready.

casperklein (dismissed):

polarathene:
The following two tests are failing in `test/tests.bats`. Possibly due to changes in this PR. I'll re-run the tests just in-case they're not :sweat_smile: 

---

[256](https://github.com/docker-mailserver/docker-mailserver/actions/runs/3204783080/jobs/5240594967#step:5:264)
```
not ok 256 checking smtp: rejects spam in 153ms
#  in test file test/tests.bats, line 329)
#   `assert_output 1' failed
#
# -- output differs --
# expected : 1
# actual   : 2
# --
#
```

https://github.com/docker-mailserver/docker-mailserver/blob/dd6f967094e5d5856fa7b00c9a27598d68986342/test/tests.bats#L325-L330

---

[278](https://github.com/docker-mailserver/docker-mailserver/actions/runs/3204783080/jobs/5240594967#step:5:295):

```
not ok 278 checking system: /var/log/mail/mail.log is error free in 671ms
#  in test file test/tests.bats, line 511)
#   `assert_failure' failed
#
# output : Oct  7 18:14:55 mail amavis[2604]: (02604-01) (!)connect to [127.0.0.1]:10025 failed, attempt #1: Can't connect to socket [127.0.0.1]:10025 using module IO::Socket::IP: Connection refused
# --
#
```

https://github.com/docker-mailserver/docker-mailserver/blob/dd6f967094e5d5856fa7b00c9a27598d68986342/test/tests.bats#L510-L511

https://github.com/docker-mailserver/docker-mailserver/blob/dd6f967094e5d5856fa7b00c9a27598d68986342/test/tests.bats#L499-L516

polarathene:
`test/tests.bats` [failed early during setup again](https://github.com/docker-mailserver/docker-mailserver/actions/runs/3204783080/jobs/5243152333#step:5:231) at a common place:

```
not ok 223 setup_file failed
# (from function `setup_file' in test file test/tests.bats, line 77)
#   `docker exec mail /bin/sh -c "nc 0.0.0.0 25 < /tmp/docker-mailserver-test/email-templates/non-existing-user.txt"' failed
# 5512b8d095f19644a77e931d6acc10041cdf7751b88e13bdc51f7664ccc0438a
```

https://github.com/docker-mailserver/docker-mailserver/blob/dd6f967094e5d5856fa7b00c9a27598d68986342/test/tests.bats#L77

Worth noting is this is a failure because of the return codde/exit status of the `docker exec` command being in the test setup, not due to any test assertions failing. Perhaps we could simply move these to a test case directly after setup?

There is nothing in the output that indicates the failure, and if previouse `nc` commands were successful, I'm not sure why this would failure other than perhaps being run without pause. We also have helpers to run commands like this until they're successful.

test/tests.bats
```diff
@@ -52,10 +53,7 @@ setup_file() {
 
   # this relies on the checksum file being updated after all changes have been applied
   wait_until_change_detection_event_completes mail
-
-  # wait for ClamAV to be fully setup or we will get errors on the log
-  repeat_in_container_until_success_or_timeout 60 mail test -e /var/run/clamav/clamd.ctl
-
+  wait_for_service mail postfix
```
polarathene, line 56: There is two accounts added a few lines up. I am not sure if they are the cause of the problem here, if they introduce a change that gets detected and handled while another is in process.

Probably need to check how many changes are being detected / processed.

---

Technically we can use the `repeat_in_container_until_success_or_timeout` as  a workaround for each `nc` command below. Then if they fail due, they've got ample time to try again until everything should be back up to be successful.

polarathene: > Probably need to check how many changes are being detected / processed.

Confirmed. 2 changes detected.

<details>
<summary>Investigation</summary>

Interestingly, this is due to the `docker run` of a separate instance of the DMS image that uses an alternative approach to append the account to the `postifix-accounts.cf`. And it's this specific approach that is consistently producing two changes (_I've added sufficient sleep before and after, checking the logs in-between to confirm this_).

Without either of the account lines, no changes are present for detection. With the `setup email add` with `docker exec` approach only, we get 1 change detected, even when using it to add two accounts (though it's possible change detection could occur slightly inbetween those two updates, sometimes resulting in two change events.

---

Here is the logs with a simple to reproduce test:

```bash
wait_for_finished_setup_in_container mail
sleep 60

# The problem line that creates two change events:
docker run --rm -t "${NAME}" /bin/sh -c 'echo "# comment"' >> "${PRIVATE_CONFIG}/postfix-accounts.cf"

wait_until_change_detection_event_completes mail
sleep 60
```

```
docker exec mail sh -c "cat /var/log/supervisor/changedetector.log"

[  DEBUG  ]  2022-10-09 04:43:50  Starting changedetector
[  DEBUG  ]  2022-10-09 04:44:00  Changedetector is ready
---------------------------------------------------------------------------------------------------------------------------------------------------------
[   INF   ]  2022-10-09 04:44:00  Change detected
---------------------------------------------------------------------------------------------------------------------------------------------------------
user1@localhost.localdomain|{SHA512-CRYPT}$6$PPeLajmedrqhKiTx$z.XP2qjPS196IhAZDsQ/JI.pm2HKCBaDZZG/sQ9qbSlpEbZoslwaN7WvBbXiCB662yTWsBjiZZTg613aZgVdj1
user2@otherdomain.tld|{SHA512-CRYPT}$6$yhZeuyPYOpdLiNUi$3zotjjHCLqa2.s4eXdrJPDyF1Oh6D8G49wRWbMfGq6HaI.IMd48KcCgJ6cg2koD2AFjs9sYWlDvDyDrR8LGz00
user3@localhost.localdomain|{SHA512-CRYPT}$6$xVgbbov7CMgZbSg2$KQyxDFh36Sh1ha8ZyoU9CCr/DwjcJnIaWA.Sooya7gCmg9Yo4/UeHS5tPnBaKXOKyed3jkECnIatqknUr3zyX/|userdb_mail=mbox:~/mail:INBOX=~/inbox
# this is a test comment, please don't delete me :'(
           # this is also a test comment, :O
# comment
---------------------------------------------------------------------------------------------------------------------------------------------------------
[  DEBUG  ]  Creating master user 'masterusername'
[  DEBUG  ]  Creating user 'user1' for domain 'localhost.localdomain'
[  DEBUG  ]  Creating user 'user2' for domain 'otherdomain.tld'
[  DEBUG  ]  Creating user 'user3' for domain 'localhost.localdomain' with attributes 'userdb_mail=mbox:~/mail:INBOX=~/inbox'
[  DEBUG  ]  Adding alias 'alias1@localhost.localdomain' for user 'user1@localhost.localdomain' to Dovecot's userdb
[  DEBUG  ]  Alias 'alias2@localhost.localdomain' is non-local (or mapped to a non-existing account) and will not be added to Dovecot's userdb
[  DEBUG  ]  Adding alias '@localdomain2.com' for user 'user1@localhost.localdomain' to Dovecot's userdb
[  DEBUG  ]  2022-10-09 04:44:01  Restarting services due to detected changes
postfix: stopped
postfix: started
dovecot: stopped
dovecot: started
[  DEBUG  ]  2022-10-09 04:44:06  Completed handling of detected change
---------------------------------------------------------------------------------------------------------------------------------------------------------
[   INF   ]  2022-10-09 04:44:08  Change detected
---------------------------------------------------------------------------------------------------------------------------------------------------------
user1@localhost.localdomain|{SHA512-CRYPT}$6$PPeLajmedrqhKiTx$z.XP2qjPS196IhAZDsQ/JI.pm2HKCBaDZZG/sQ9qbSlpEbZoslwaN7WvBbXiCB662yTWsBjiZZTg613aZgVdj1
user2@otherdomain.tld|{SHA512-CRYPT}$6$yhZeuyPYOpdLiNUi$3zotjjHCLqa2.s4eXdrJPDyF1Oh6D8G49wRWbMfGq6HaI.IMd48KcCgJ6cg2koD2AFjs9sYWlDvDyDrR8LGz00
user3@localhost.localdomain|{SHA512-CRYPT}$6$xVgbbov7CMgZbSg2$KQyxDFh36Sh1ha8ZyoU9CCr/DwjcJnIaWA.Sooya7gCmg9Yo4/UeHS5tPnBaKXOKyed3jkECnIatqknUr3zyX/|userdb_mail=mbox:~/mail:INBOX=~/inbox
# this is a test comment, please don't delete me :'(
           # this is also a test comment, :O
# comment
---------------------------------------------------------------------------------------------------------------------------------------------------------
[  DEBUG  ]  Creating master user 'masterusername'
[  DEBUG  ]  Creating user 'user1' for domain 'localhost.localdomain'
[  DEBUG  ]  Creating user 'user2' for domain 'otherdomain.tld'
[  DEBUG  ]  Creating user 'user3' for domain 'localhost.localdomain' with attributes 'userdb_mail=mbox:~/mail:INBOX=~/inbox'
[  DEBUG  ]  Adding alias 'alias1@localhost.localdomain' for user 'user1@localhost.localdomain' to Dovecot's userdb
[  DEBUG  ]  Alias 'alias2@localhost.localdomain' is non-local (or mapped to a non-existing account) and will not be added to Dovecot's userdb
[  DEBUG  ]  Adding alias '@localdomain2.com' for user 'user1@localhost.localdomain' to Dovecot's userdb
[  DEBUG  ]  2022-10-09 04:44:08  Restarting services due to detected changes
postfix: stopped
postfix: started
dovecot: stopped
dovecot: started
[  DEBUG  ]  2022-10-09 04:44:14  Completed handling of detected change
```

I added the `---` to better section the log, should make it easier to spot the change detection and the additional `postfix-accounts.cf` content that I added to `check-for-changes.sh` to see if the content was different at all.

The detected checksums are different (_as they should be to detect the event_), so I'm not sure what is being changed. Possibly due to the 2 second loop window, the checksums are generated while the file is still writing the update, so the 2nd time around it would be a different checksum?... I ruled that out by increasing the `check-for-changes.sh` loop time to 30 seconds, and we still get 2 change events, just with a larger time apart. So the file is likely being manipulated during the change event processing.

Meanwhile, neither of these reproduce the issue, both only cause 1 change event when the only line:

```bash
docker exec -it mail /bin/sh -c 'echo "# comment" >> "/tmp/docker-mailserver/postfix-accounts.cf"'
echo "# comment" >> "${PRIVATE_CONFIG}/postfix-accounts.cf"
```

Used `cmp -b file-before file-after` to detect diff by bytes at the start and end of a change event. It revealed that `^M` (CRLF: `\r\n`) and `^J` (LF: `\n`) are the difference... line endings :grimacing: well that was annoying :laughing: 

</details>

---

## Resolved :tada: 

https://github.com/docker-mailserver/docker-mailserver/blob/ff969509f8b9c49e399327887c9a29f6fd664b80/test/tests.bats#L47

This particular way of adding an account causes two change events as it inserts a CRLF instead of a LF, and when `postfix-accounts.cf` is a change event trigger it will run the `helpers/accounts.sh` script that has this line converting CRLF into LF:

https://github.com/docker-mailserver/docker-mailserver/blob/ff969509f8b9c49e399327887c9a29f6fd664b80/target/scripts/helpers/accounts.sh#L25

There is similar line duplicated down at the bottom of the file for the master accounts feature, and likewise `helpers/aliases.sh` has a similar method that can modify a config file.

I don't think we need this processing, if we do we could probably it to a temporary file rather than modify the existing config.

polarathene: [I chose just to fix only `tests.bats`](https://github.com/docker-mailserver/docker-mailserver/pull/2820) for now instead.

I don't have time to look into changing the helper scripts, and rather address that by adopting something like rust to do the processing instead :sweat_smile:

georglauterbach: Marking this as resolved :)

georglauterbach:
Very nice investigative work @polarathene!

I think this PR has some nice additions too that should make it into v12 later :) Feel free to review :D

polarathene (changes requested):
Few questions and a request to remove the `TEST_TIMEOUT_IN_SECONDS` addition at the start of `tests.bats`.

The rest of the feedback is just `clamav.bats` suggestions as the "checking" bit seemed redundant and I'd rather the space suffix be moved out of the ENV - although the suggestions reduce the usefulness of the ENV a bit :sweat_smile:

Makefile
```diff
@@ -42,12 +43,11 @@ generate-accounts:
...
-tests:
-	@ NAME=$(NAME) ./test/bats/bin/bats --timing test/*.bats
+tests: ALWAYS_RUN
+	@ ./test/bats/bin/bats --timing test/*.bats
 
-.PHONY: ALWAYS_RUN
-test/%.bats: ALWAYS_RUN
-	@ ./test/bats/bin/bats $@
+test/%: ALWAYS_RUN
+	@ ./test/bats/bin/bats --timing $@.bats
```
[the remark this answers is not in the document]
georglauterbach: As `make` is a weird build system, `ALWAYS_RUN` is a `.PHONY` target, so it will make the recipe for which it is a requirement run everytime. Otherwise, `make` _could_ somehow not run the recipe? Maybe this is not actually needed? I will have a look.

test/clamav.bats
```diff
@@ -0,0 +1,58 @@
+load 'test_helper/common'
+
+TEST_NAME_PREFIX='checking ClamAV: '
```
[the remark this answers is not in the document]
georglauterbach: Haha, I was thinkting about that, but wanted to be consistent with other tests. I prefer only havng `clamAV` though very much :D

test/tests.bats
```diff
@@ -332,11 +329,12 @@ EOF
...
-  run docker exec mail /bin/sh -c "grep 'Blocked INFECTED' /var/log/mail/mail.log | grep external.tld=virus@my-domain.com | wc -l"
-  assert_success
-  assert_output 1
-}
+# TODO move this into `clamav.bats`
+# @test "checking smtp: rejects virus" {
+#   run docker exec mail /bin/sh -c "grep 'Blocked INFECTED' /var/log/mail/mail.log | grep external.tld=virus@my-domain.com | wc -l"
+#   assert_success
+#   assert_output 1
+# }
```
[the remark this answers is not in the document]
georglauterbach: I will have a look into it.

test/tests.bats
```diff
@@ -481,9 +464,9 @@ EOF
 # system
 #
 
-@test "checking system: freshclam cron is enabled" {
+@test "checking system: freshclam cron is disabled" {
```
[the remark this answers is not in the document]
georglauterbach: I thought I had copied this into the new `clamav.bats` file...

Makefile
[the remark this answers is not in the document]
georglauterbach: Turns out: not needed. I will remove it.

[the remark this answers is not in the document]
casperklein: I spotted that too before we came up with the real test failure solution and queued it on my personal todo as a follow up, because it wasn't related to "fix transient test failures".

The PR title/description should be changed .

test/tests.bats
```diff
@@ -1,7 +1,8 @@
-load 'test_helper/common'
+# this test needs more startup time, and we need
+# to increase the timeout
+export TEST_TIMEOUT_IN_SECONDS=180
```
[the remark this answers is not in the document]
georglauterbach: Done in fbe020003da793d4fa8dbd2e0433f95971b2af56

test/tests.bats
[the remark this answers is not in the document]
georglauterbach: Done in fbe020003da793d4fa8dbd2e0433f95971b2af56

test/tests.bats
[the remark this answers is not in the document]
georglauterbach: Done in fbe020003da793d4fa8dbd2e0433f95971b2af56

Makefile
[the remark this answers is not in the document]
georglauterbach: Done in fbe020003da793d4fa8dbd2e0433f95971b2af56

georglauterbach:
I have changed the title as @casperklein proposed (it was indeed not fitting anymore).

test/clamav.bats
```diff
@@ -0,0 +1,79 @@
...
+    --hostname mail.my-domain.com \
+    --volume "${PRIVATE_CONFIG}:/tmp/docker-mailserver" \
+    --volume "${PWD}/test/test-files:/tmp/docker-mailserver-test:ro" \
+    --env ENABLE_AMAVIS=1 \
+    --env AMAVIS_LOGLEVEL=2 \
+    --env ENABLE_CLAMAV=1 \
+    --env ENABLE_UPDATE_CHECK=0 \
+    --env ENABLE_SPAMASSASSIN=0 \
+    --env ENABLE_FAIL2BAN=0 \
+    --env PERMIT_DOCKER=host \
```
polarathene, line 22: You recently added the `PERMIT_DOCKER=host` line here, what was it needed for?

test/clamav.bats
```diff
@@ -0,0 +1,79 @@
...
+
+  wait_for_finished_setup_in_container "${CONTAINER_NAME}"
+
+  # wait for ClamAV to be fully setup or we will get errors on the log
+  repeat_in_container_until_success_or_timeout 60 "${CONTAINER_NAME}" test -e /var/run/clamav/clamd.ctl
+
+  wait_for_service "${CONTAINER_NAME}" postfix
+  wait_for_smtp_port_in_container "${CONTAINER_NAME}"
+
+  "${RUN_COMMAND[@]}" bash -c "nc 0.0.0.0 25 < /tmp/docker-mailserver-test/email-templates/amavis-virus.txt"
```
polarathene, line 35: This was not dropped from `tests.bats`, is it still serving a purpose there?

polarathene (approved):
LGTM :+1: 

Two small questions remain :)

test/clamav.bats
georglauterbach: The container was refusing the mails sent via `nc` with `DATA without valid RCPT from ...`, so I tried and tried, and found this toggle to solve the problem. `tests.bats` uses it the same way, presumably because `nc` does not respect proper protocol? There was definitely (improper) command pipelining involved, which seemed to cause issues too.. or at least warnings. I am not 100% sure why this solved the issue though..

test/clamav.bats
georglauterbach: I missed that. I will need to adjust the `listmailuser` check too. I am planning to revise some tests in the future, so I hope it's okay to leave it here for now. I have some plans for the test suite that will revise this.

georglauterbach: Nevermind, I added it to this PR - why not. Its more fitting for this PR anyways.

test/clamav.bats
polarathene: Ah ok, I am familiar with that issue, but assumed since it was run in the container that it shouldn't matter :sweat_smile: 

I assume it's because that without a `PERMIT_DOCKER` value, we get `none` which clears out the Postfix `main.cf:mynetworks` setting:

https://github.com/docker-mailserver/docker-mailserver/blob/70ad765a30cd5ce8385994f592f51e05b1e0a744/target/scripts/startup/setup-stack.sh#L663-L667

All good then, thanks for clarifying :grinning:

georglauterbach: You're welcome :)

polarathene (approved):
LGTM

Good work! :sunglasses: :tada:

casperklein (approved):

PR #2817, opened 2022-10-09
 Run user-patches.sh right before starting daemons

casperklein:
# Description

From the [docs](https://docker-mailserver.github.io/docker-mailserver/edge/config/advanced/override-defaults/user-patches/):

>  The user patches script runs right before starting daemons.

That is not the case at the moment.

For example, changes to the file `/root/.bashrc` done by `user-patches.sh` will be discarded by the setup function `_environment_variables_export`, which is called after.

This PR solves that, by not registering `_setup_user_patches` via `_register_setup_function`.
Instead `_setup_user_patches` is called directly right before `_start_daemons`.

## Type of change

## Checklist:


polarathene:
[Current failure in `test/tests.bats` early in setup](https://github.com/docker-mailserver/docker-mailserver/actions/runs/3212309795/jobs/5251083290#step:5:226), common issue with `nc` commands:

```
not ok 218 setup_file failed
# (from function `setup_file' in test file test/tests.bats, line 73)
#   `docker exec mail /bin/sh -c "nc 0.0.0.0 25 < /tmp/docker-mailserver-test/email-templates/existing-user-and-cc-local-alias.txt"' failed
# c76247e583fa6165ad1805638f705bcfb901d5d8c00491726cb523a2be5c59d3
# 220 mail.my-domain.com ESMTP
```

Any clue why the failures happen at random times here? Could it be due to no delay between the commands? I know that this approach doesn't respect proper SMTP exchange protocol as it doesn't wait for server responses. Perhaps we can resolve this by an alternative way to deliver these.

casperklein:
> I know that this approach doesn't respect proper SMTP exchange protocol as it doesn't wait for server responses

I don't think that's an issue. If I do the `nc` commands from a remote host to DMS I get:

`503 5.5.0 <myremote-most]>: Client host rejected: Improper use of SMTP command pipelining`.

If I do the same on the DMS host, I don't get an error. I guess it's because PERMIT_DOCKER=host is in place and some restrictions do not apply?

Maybe [`nc -v`](https://github.com/casperklein/archive_docker-mailserver/actions/runs/3212527188/jobs/5251485995) gives more insights.

casperklein:
[`# nc: connect to 0.0.0.0 port 25 (tcp) failed: Connection refused`](https://github.com/casperklein/archive_docker-mailserver/actions/runs/3212527188/jobs/5251485995#step:5:232)

Postfix seems not to be running when `nc` is executed.

polarathene:
> Postfix seems not to be running when `nc` is executed.

I'm not sure how we can debug that. Immediately before this, we verify that port 25 is reachable. The same failures are happening in the [PR trying to resolve these issues](https://github.com/docker-mailserver/docker-mailserver/pull/2815#issuecomment-1272382061) that additionally waits on Postfix to be ready (_which seems possibly redundant with the existing port 25 wait method that follows?_)

Presumably something is causing Postfix to restart during these `nc` calls, hence the timing being unreliable where it fails during the setup? It could be the account generation after starting the container is the cause.

A change event may already sometimes be in process and waited on correctly, but by the time it's handled, the change detector notices that a change has happened again and brings the services down to handle those added accounts?

polarathene:
The cause of change events being unreliably handled [has been found](https://github.com/docker-mailserver/docker-mailserver/pull/2815#discussion_r990740402) :)

I'll do a quick PR for it and we can update these PRs after and there should be no more problematic `tests.bats` setup :+1:

polarathene (approved):

georglauterbach (approved):

casperklein:
If possible, we should exclude `CHANGELOG.md` or markdown in general from the tests. As long as we have no markdown specific tests, there is no need to run the tests.

georglauterbach:
> If possible, we should exclude `CHANGELOG.md` or markdown in general from the tests. As long as we have no markdown specific tests, there is no need to run the tests.

A good idea 👍🏼

Also, #2815 should fix the issue with the tests were seeing currently.

georglauterbach (approved):

georglauterbach:
@casperklein which tastes exactly are you referring to?:)

casperklein:
I noticed with my last commit, that the CI tests were run again. The only change in that commit however was in `CHANGELOG.md`.

But I think, that is how it works.

Markddown files per se are not considered for running the tests:

https://github.com/docker-mailserver/docker-mailserver/blob/488b256e954b48c8f1b70a27cdc66fbcf6680f26/.github/workflows/default_on_push.yml#L8-L13

I guess, on a commit, the whole PR is taking into account and not just the files touched with the current commit.

georglauterbach:
> I guess, on a commit, the whole PR is taking into account and not just the files altered with the current commit.

Yes :)

polarathene (approved):

Issue #159, opened 2016-04-24
/etc/dovecot/userdb not generated correctly


stonemaster:
I am using the **v2** of the container.

I think that the `/etc/dovecot/userdb` isn't generated correctly. I have a `postfix-accounts.cf` with two users (just posting one as example).

`
test@test.com|{CRAM-MD5}0..[65 bytes total]..f
`

During startup of the container I see the message

`
mail_1     | user 'test' for domain 'test.com' with password '********'
`

.. which seems to indicate that everything is fine and the users are created. But the file `/etc/dovecot/userdb` contains an incorrect user line:

`
:5000:5000::/var/mail/test.com/test::2...[MD5 bytes]f
`

Note the `:` at the beginning where username and password should be set. At the end of a line is some MD5 string which is a **part** of the MD5 password hash (last 52 bytes).

I looked at the startup code and everything looks great. I am totally clueless.
Do you have an idea what is going wrong?

Niklv:
Have same issue.

Niklv:
@stonemaster found that file config/postfix-accounts.cf has CLRF line endings.
I recreate this file with LF endings and it fork fine for me.

You can check endings:

```
root@mail:/# file /tmp/docker-mailserver/postfix-accounts.cf
/tmp/docker-mailserver/postfix-accounts.cf: ASCII text
```

tomav:
`CLRF` is windows format, and we're on a linux docker image :-)
That said, it's possible to add a check on that using:

    tr -d '\r' < input.file > output.file

or

    sed 's/$'"/`echo \\\r`/" input.txt > output.txt

(should probably be adapted)

By the way, how did you generate your `postfix-accounts.cf`? Using:
- command line with password and `>> config/postfix-accounts.cf`
- command line without password pasted manually in the file (I suppose this one)

Thanks.

tomav:
I added the tip regarding line endings in the FAQ.

stonemaster:
I am unsure how I actually added the accounts. But that's it. It's just about the wrong line endings. I would suggest adding some check in the script because this is an evil one if you don't know what happens and the file actually **looks** good :-)

Niklv:
@tomav @stonemaster I generate file on my linux VPS using this command from README.md:
````
docker run --rm \
  -e MAIL_USER=user1@domain.tld \
  -e MAIL_PASS=mypassword \
  -ti tvial/docker-mailserver:v2 \
[...]

Issue #13, opened 2015-08-13
User-Creation


ParadoxDelta:
I have some issues to make my mailserver work and i think it's due to my  wrong user creation. I took your template and pasted in a new file accounts.cf under ./postfix and filled my data in there. somehow like this:

# imap
username:                  <reply-to@mydomain.com>
password:                  <mypassword>
server:                    <123.132.123.123>
imap port:                 993
imap path prefix:        INBOX
auth method:            md5 challenge-response

# smtp
smtp port:                587
username:                  <reply-to@mydomain.com>
password:                  <mypassword>
auth method:            md5 challenge-response

or is it supposed to be:

# imap
username:                  reply-to@mydomain.com
password:                  mypassword
server:                    123.132.123.123
imap port:                 993
imap path prefix:        INBOX
auth method:            md5 challenge-response

# smtp
smtp port:                587
username:                  reply-to@mydomain.com
password:                  mypassword
auth method:            md5 challenge-response

either way, i cant connect a client to the server and when i send an e-mail to it, nothing is saved under /var/mail (also did postfix reload)
i could access the ports with telnet so i'm pretty shure something is wrong with the server-setup

ParadoxDelta:
ah, i found the example-file with user@domain.tld|password
but my problems remain the same. client wont connect and emails wont get saved even though ports are reachable. any idea what the problem might be?

gegere:
Hello @ParadoxDelta this is a great docker instance. I thought I would share [another repo](https://github.com/htmlgraphic/Mail-Server) I have been working on maybe you can piece a few things together and get your build working.

tomav:
@ParadoxDelta, I can help you, but I need you configurations to reproduce the problem. Do not hesitate.

@gegere, advertisement spotted! :-) Please...

gegere:
Pretty Please? Happy to help your build anytime. Open source projects are epic!

ParadoxDelta:
Alright but i didn't change much. Just my URL as ENV when run and one user in the account.cf
Container looks like this
bbc381af9435        tvial/docker-mailserver   "/bin/sh -c /usr/loc   15 hours ago        Up 15 hours         0.0.0.0:25->25/tcp, 0.0.0.0:143->143/tcp, 0.0.0.0:587->587/tcp, 0.0.0.0:993->993/tcp   mailserver

main.cf
# See /usr/share/postfix/main.cf.dist for a commented, more complete version

[...]

PR #2820, opened 2022-10-09
tests(fix): Avoid introducing CRLF into `postfix-accounts.cf` during setup

polarathene:
# Description

`test/tests.bats` has been extremely unreliable during the setup stage. This should finally resolve that. Big thanks to @casperklein and @georglauterbach who have helped rule out other causes to narrow it down to this.

Recent changes I introduced to `tests.bats` better surfaced change detection problems more consistently, to the point failure frequency was extremely probable. I investigated and identified the cause :tada: 

Currently when a change detection would be triggered and during processing, a `CRLF` is found and converted to a `LF`, which modifies the `postfix-accounts.cf` file, triggering another change event :grimacing: 

Marked for `v11.2.0` as it's only a change to our tests and should make existing PRs much happier :)

---

## Solution: Avoid introducing `CRLF` into config by adding accounts only via `setup email add`

There is no need for the first approach used in `tests.bats` to add an account, and it is [the culprit for causing the CRLF to appear](https://github.com/docker-mailserver/docker-mailserver/pull/2815#discussion_r990740402).

It is unclear why a `CRLF` is added to the end of the file, only when using `docker run` with the DMS docker image to output a `postfix-accounts.cf` config line via `echo` and using that stdout result to append (`>>`) to the `postfix-accounts.cf`.

It did not seem to occur via `docker exec` and performing a similar command within the container to update the file internally, nor without docker by updating `postfix-accounts.cf` in the same manner directly to the file at the volume mount location.

### Potential run-time modification of configs intended for fixing to keep in mind

In future, the following could be considered for removal, which may require better handling of the scripts (_or ideally, porting to a more capable language where this type of problem would not cause bugs when processing configs_):

https://github.com/docker-mailserver/docker-mailserver/blob/ff969509f8b9c49e399327887c9a29f6fd664b80/target/scripts/helpers/accounts.sh#L25

https://github.com/docker-mailserver/docker-mailserver/blob/ff969509f8b9c49e399327887c9a29f6fd664b80/target/scripts/helpers/accounts.sh#L32

https://github.com/docker-mailserver/docker-mailserver/blob/ff969509f8b9c49e399327887c9a29f6fd664b80/target/scripts/helpers/accounts.sh#L181

https://github.com/docker-mailserver/docker-mailserver/blob/ff969509f8b9c49e399327887c9a29f6fd664b80/target/scripts/helpers/accounts.sh#L187

https://github.com/docker-mailserver/docker-mailserver/blob/ff969509f8b9c49e399327887c9a29f6fd664b80/target/scripts/helpers/aliases.sh#L20

I do not have the time to ensure removing those modifications would not risk causing problems with our scripts that read them. We also have another issue open regarding making a copy of config files to read to avoid issues with something else modifying the file while we process it.

### History of these particular lines

- `sed -i -e '$a\' /tmp/postfix/accounts.cf` introduced in [Aug 2015 by original DMS project author](https://github.com/docker-mailserver/docker-mailserver/commit/63a7be0e975b39b72fe666a61517e32d32b98f55). No PR, but does [reference an issue the commit is meant to resolve](https://github.com/docker-mailserver/docker-mailserver/issues/13) - however I don't see anything in the discussion that looks related to it being added. I assume it's mostly specific to compatibility with shell scripts and unix utilities that expect it?
- `sed -i 's/\r//g' /tmp/docker-mailserver/postfix-accounts.cf` introduced [April 2016](https://github.com/docker-mailserver/docker-mailserver/pull/160), which [references an issue where a user experienced a bug](https://github.com/docker-mailserver/docker-mailserver/issues/159) from their `postfix-accounts.cf` having a `CRLF` present, scripts did not generate accounts in Dovecot `/etc/dovecot/userdb` properly.
- `sed -i -e "s|, |,|g" -e "s|,$||g" "${DATABASE_VIRTUAL}"` introduced in [March 2018](https://github.com/docker-mailserver/docker-mailserver/pull/897) to better support alias management (_notably `setup alias del`_) so that `,` would be normalized to always having a space suffix `, `.

Technically, if config files are only managed by our `setup` command, then there should be no need for any of these I think? Users may manually edit the files directly or have setup some other way to modify / create the configs, like the legacy `make generate-accounts` approach and what we see here in `tests.bats`.

Thus these are more a precaution / safe-guard. If we don't care about external config updates being made while the container is running, and not having change events "repair" / fix those potential issues, then we could move these into startup or `setup` fixes, or into the docs.

Otherwise they're not likely to cause much trouble, they were just silently messing with this particular test file for some time. While `make generate-accounts` probably introduces CRLF as well, those were handled during startup before a checksum was made for the change detector, thus already fixed ahead of time.

## Type of change

## Checklist:

### DIFF
--- a/test/tests.bats
+++ b/test/tests.bats
@@ -44,8 +44,8 @@ setup_file() {
   wait_for_finished_setup_in_container mail
 
   # generate accounts after container has been started
-  docker run --rm -e MAIL_USER=added@localhost.localdomain -e MAIL_PASS=mypassword -t "${NAME}" /bin/sh -c 'echo "${MAIL_USER}|$(doveadm pw -s SHA512-CRYPT -u ${MAIL_USER} -p ${MAIL_PASS})"' >> "${PRIVATE_CONFIG}/postfix-accounts.cf"
-  docker exec mail addmailuser pass@localhost.localdomain 'may be \a `p^a.*ssword'
+  docker exec mail setup email add 'added@localhost.localdomain' 'mypassword'
+  docker exec mail setup email add 'pass@localhost.localdomain' 'may be \a `p^a.*ssword'
 
   # setup sieve
   docker cp "${PRIVATE_CONFIG}/sieve/dovecot.sieve" mail:/var/mail/localhost.localdomain/user1/.dovecot.sieve


casperklein (approved):
> Currently when a change detection would be triggered and during processing, a CRLF is found and converted to a LF, which modifies the postfix-accounts.cf file, triggering another change event 😬

Very nice investigation 👍

casperklein:
> It is unclear why a CRLF is added to the end of the file, 

I found out, that the `-t` switch here is the cause for that:

`docker run --rm -e MAIL_USER=added@localhost.localdomain -e MAIL_PASS=mypassword -t "${NAME}" /bin/sh -c 'echo "${MAIL_USER}|$(doveadm pw -s SHA512-CRYPT -u ${MAIL_USER} -p ${MAIL_PASS})"' >> "${PRIVATE_CONFIG}/postfix-accounts.cf"`

Proof:

```
docker run --rm debian bash -c "echo foo" > bar; hd bar
00000000  66 6f 6f 0a                                       |foo.|
00000004
```

Same, but with `-t`

```
docker run --rm -t debian bash -c "echo foo" > bar; hd bar
00000000  66 6f 6f 0d 0a                                    |foo..|
```

polarathene:
> I found out, that the `-t` switch here is the cause for that

Very interesting, thanks for sharing that finding with `--tty` introducing it :exploding_head:

casperklein:
In my test branch, [tests did immediately run successfully](https://github.com/casperklein/archive_docker-mailserver/actions/runs/3213777827/jobs/5253705061), after removing `-t` 🎉

The solution in this PR much clearer, than just removing `-t` 👍

I don't understand, why in the first place this was used:

```bash
docker run ... > file-on-host
docker exec ... # modify file directly in the container
```

Why were there two methods used for adding accounts to a running container 🤷

georglauterbach (approved):

georglauterbach:
Hooray :D

PS: Notices that CI does not use cache here as well. GH had some recent downtime, maybe there now is something wrong with our cache? I will have a look. The changes themselves should definitely not have anything to do with the current caching issue. The issue itself is minor as well, as we only loose a minute, but annoying nevertheless.

polarathene:
> maybe there now is something wrong with our cache? I will have a look.

It is using the same cache-key as I only made a change to a test which doesn't warrant caching a new image.

I assume something didn't cache well with that key, but once we have a new one we'll have a better idea about cache working properly or not.

polarathene:
> Why were there two methods used for adding accounts to a running container

Since it was for `tests.bats` my guess is it's legacy from before when it was all in a Makefile, especially since we have similar with `make generate-accounts`. Then when setup commands arrived to handle account management better, they added the `addmailuser` command to test that was working too, when maybe we didn't have much tests for that support until later?

PR #2822, opened 2022-10-10
ci: make tests run in parallel

georglauterbach:
# Description

This is a major step forward for our CI. This PR enables the first tests to run in parallel. The `Makefile` and the GH workflow file `generic_test.yml` were adjusted. Files under `test/` were renamed. 6 files were adjusted to provide a "template" for future PRs that port even more file from running in serial to running in parallel.

The new system works in the following way: files under `test/` that begin with "parallel" can be run in parallel. The number that follows is the "execution unit" that they are executed in, i.e. with which other tests they actually run in parallel. For the GH merge request job, this is equivalent to telling which job certain tests run together in. Tests that begin with "serial" run one-after-another. This is either because they actually require running serialzed (`serial.setup.sh.bats`) or because they were not yet adjusted.

I also added a template (`parallel.2.template.bats`) which serves as a template for all other tests, especially parallel tests. I will provide a commit that adds documentation about this change, and the template file will be referenced.

## Size of This PR

I know, it's a big PR. I know you don't like that. But this is mostly because of the few test files I adjusted and the documentation additions.. the important changes are in `Makefile` and the GH Action workflow files. I think having these changes in one PR and not in separate ones (this only applied to the first PR of slowly adopting the test suite to run in parallel) is the right way.

## Disclaimer

One could consider this a breaking change for tests. This change is not backwards compatible (by default!), because the `--jobs x` where `x > 0` requires GNU parallel to be installed. On most Linux distributions, this is installed by default, but I cannot guarantee it. One can "revert" the parallelization by using `PARALLEL_JOBS=1 make ...` though (the default is `2`).

## Effect

We have an "outer" and "inner" parallelization with GH's CI. The "outer" by using [`matrix`](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs), which _should, when tests are evenly distributed_ in the future, speed up running the test suite **by a factor of 4**. The inner parallelization comes from using BATS with the `--jobs` flag to run test files themselves in parallel. **With `--jobs 2`, the speedup can be as large as 30%** (on my system, maybe larger on more potent systems), with `3`, which is possible on GH runners I think, the speedup should be even larger.

This should yield a theoretical speedup of 82.5% (with even distribution of tests into their respective execution units).

## Type of change

## Checklist:


test/serial.setup.sh.bats
```diff
@@ -0,0 +1,150 @@
+load 'test_helper/common'
+
+# ! NEEDS SERIAL EXECTION - DO NOT RUN IN PARALLEL
```
georglauterbach, line 3: The diff for this file is misleading. Nothing changed except for this line.. but because of the renaming, git thinks this is a completely new file. Makes the diff appear large :D

docs/content/contributing/issues-and-pull-requests.md
```diff
@@ -37,6 +37,28 @@ The development workflow is the following:
...
+    commit
+    checkout master
+    merge another-branch
+    checkout your-branch
+    merge master
+    commit
+    checkout master
+    merge your-branch
+```
+
```
georglauterbach, line 61: Oh, may need to remove this again. Does not actually belong to changes in this PR. Reminder to future-self.

georglauterbach: Done.

test/serial.setup.sh.bats
polarathene: > but because of the renaming, git thinks this is a completely new file

This is going to be annoying, but I'd like to request you avoid this issue by separating out commits of renaming and then a commit for file change, or if it's easier at this point the other way around.

You can probably handle this without making it a pain to address. Try this:

1. Use a rename tool or script to remove the `serial.` / `parallel.` (or similar with the optional numbers) to revert the filename.
2. Commit just these file renames.
3. Commit a revert commit that does the opposite.
4. Try rebase that revert commit (original test names to new test names) so that it's the first commit in this PR. If this works you're done.
5. If it's not going to go smoothly, revert the existing commits (_in GitKraken I can just right-click a commit and create a revert commit, I assume there is a direct git command for this_). Then you can revert the revert :laughing: and a rebase that should be able to drop original commits and some of the reverts that follow... that should give a clean commit history now. This is fairly easy to do with GitKraken UI for interactive rebasing a range of commits, if you've got the commits pushed here I can handle that part if you like :+1: 

I remember for when we imported the Github Wiki git history to merge into our git repo history here (_via merge commit, which visually interleaved the commit history, but didn't rewrite the main project commit history_), we made sure to separate the file renames from content edits to ensure git could follow a file to it's original and diff properly :)

---

I won't have time to look at this until about 8 hours or so, but I'm curious about the current naming convention choice vs using folders for `serial` and `parallel`? Any benefit in prefixing filenames with it?

Perhaps you're using that from one of the BATS ENV or scripts on what to run? (_although from our end grouping by folder should be just as easy?_) I know that I'd like to eventually have some folders to group tests by topics/relevance.

I was planning for CI to have several runners/jobs that would run a subset of the tests (as a different form of parallelism), a good candidate would be the SSL tests, where one of those takes a huge amount of time to iterate through permutations of RSA + ECDSA and modern / intermediate (legacy) TLS levels.

georglauterbach: I wanted to create directories, but this caused the ´load` directive in BATS in conjunction with `test/test_helper/common.sh`. Maybe a future PR can take care of this, for now, we should be fine with a simple naming scheme that is somewhat similar to DNS.

---

I will take care  of the commit history.

georglauterbach: Re-wrote history and I hope it is now easier to review.

polarathene: > Re-wrote history and I hope it is now easier to review.

Hey @georglauterbach sorry for the late response. Looking over the commit history, I can see a few files in the commits were not just renamed but added changes in that same commit, which breaks the rename detection for git blame history.

It happened in a few commits, if I have your permission I don't mind correct this history and doing a force-push?

polarathene: Also to preserve this rename history, this PR will use a rebase merge, not a squash merge.

polarathene:
**NOTE:** To preserve this rename history, this PR will use a rebase merge, **not a squash merge**.

test/parallel.2.test_helper.bats
```diff
@@ -35,56 +37,60 @@ load 'test_helper/common'
...
+@test "${TEST_NAME_PREFIX} wait_for_smtp_port_in_container aborts wait after timeout" {
   local CONTAINER_NAME
   CONTAINER_NAME=$(docker run --rm -d alpine sleep 100)
   SECONDS=0
   TEST_TIMEOUT_IN_SECONDS=2 run wait_for_smtp_port_in_container "${CONTAINER_NAME}"
   [[ ${SECONDS} -ge 2 ]]
   assert_failure
   assert_output --partial "Timed out on command"
+
+  docker rm -f "${CONTAINER_NAME}"container_has_service_running/wait_for_service
```
polarathene, line 89: What is the need for this line here?

The container should remove itself once done due to the `docker run --rm`?

test/parallel.2.test_helper.bats
```diff
@@ -35,56 +37,60 @@ load 'test_helper/common'
...
+@test "${TEST_NAME_PREFIX} repeat_in_container_until_success_or_timeout run command in container" {
   local CONTAINER_NAME
   CONTAINER_NAME=$(docker run --rm -d alpine sleep 100)
   SECONDS=0
   ! repeat_in_container_until_success_or_timeout 10 "${CONTAINER_NAME}" sh -c "echo '${CONTAINER_NAME}' > /tmp/marker"
   [[ ${SECONDS} -le 1 ]]
   run docker exec "${CONTAINER_NAME}" cat /tmp/marker
   assert_output "${CONTAINER_NAME}"
+
+  docker rm -f "${CONTAINER_NAME}"
```
polarathene, line 69: Same here, the container should already remove itself after completion of the `sleep`?

Are you just wanting to bring it down sooner for cleanup? There is no explicit name assigned, so it should not cause any problems, the container will just run for 100 seconds and be seen in `docker ps` until then.

Makefile
```diff
@@ -33,17 +35,22 @@ clean:
...
-	@ docker run --rm -e MAIL_USER=user3@localhost.localdomain -e MAIL_PASS=mypassword -t $(NAME) /bin/sh -c 'echo "$$MAIL_USER|$$(doveadm pw -s SHA512-CRYPT -u $$MAIL_USER -p $$MAIL_PASS)|userdb_mail=mbox:~/mail:INBOX=~/inbox"' >> test/config/postfix-accounts.cf
+	@ docker run --rm -e MAIL_USER=user1@localhost.localdomain -e MAIL_PASS=mypassword -t $(IMAGE_NAME) /bin/sh -c 'echo "$$MAIL_USER|$$(doveadm pw -s SHA512-CRYPT -u $$MAIL_USER -p $$MAIL_PASS)"' > test/config/postfix-accounts.cf
+	@ docker run --rm -e MAIL_USER=user2@otherdomain.tld -e MAIL_PASS=mypassword -t $(IMAGE_NAME) /bin/sh -c 'echo "$$MAIL_USER|$$(doveadm pw -s SHA512-CRYPT -u $$MAIL_USER -p $$MAIL_PASS)"' >> test/config/postfix-accounts.cf
+	@ docker run --rm -e MAIL_USER=user3@localhost.localdomain -e MAIL_PASS=mypassword -t $(IMAGE_NAME) /bin/sh -c 'echo "$$MAIL_USER|$$(doveadm pw -s SHA512-CRYPT -u $$MAIL_USER -p $$MAIL_PASS)|userdb_mail=mbox:~/mail:INBOX=~/inbox"' >> test/config/postfix-accounts.cf
```
polarathene, line 40: Not exactly related to this PR, but I suppose we could update this to be similar to the recent change in `tests.bats`? (_we can defer this to another PR if preferable_)

If we're going to spin up a DMS image, we might as well add an account through `setup` and a volume mount or `cat /tmp/docker-mailserver/postfix-accounts.cf` output?

Alternatively, we just add another `postfix-accounts.cf` file and `cp` that reference copy to `test/config/postfix-accounts.cf` :man_shrugging:

Makefile
```diff
@@ -33,17 +35,22 @@ clean:
...
 	@ echo "# this is a test comment, please don't delete me :'(" >> test/config/postfix-accounts.cf
 	@ echo "           # this is also a test comment, :O" >> test/config/postfix-accounts.cf
 
 # Dovecot master accounts
-	@ docker run --rm -e MASTER_USER=masterusername -e MASTER_PASS=masterpassword -t $(NAME) /bin/sh -c 'echo "$$MASTER_USER|$$(doveadm pw -s SHA512-CRYPT -u $$MASTER_USER -p $$MASTER_PASS)"' > test/config/dovecot-masters.cf
+	@ docker run --rm -e MASTER_USER=masterusername -e MASTER_PASS=masterpassword -t $(IMAGE_NAME) /bin/sh -c 'echo "$$MASTER_USER|$$(doveadm pw -s SHA512-CRYPT -u $$MASTER_USER -p $$MASTER_PASS)"' > test/config/dovecot-masters.cf
 
-tests:
-	@ NAME=$(NAME) ./test/bats/bin/bats --timing test/*.bats
+tests: test/part/0 test/part/1 test/part/2 test/part/3
```
polarathene, line 47: What's going on here now? The previous test command to run was simple enough to follow.

Is this calling the `Makefile` command below it, `test/part/%`?

Makefile
polarathene: Likewise for the master account creation directly after this.

Makefile
```diff
@@ -33,17 +35,22 @@ clean:
...
+
+test/part/%:
+# part/0 => tests run in a serialized manner
+# part/x where (x > 0) => tests are run in parallel
+	@ if [[ $* -eq 0 ]]; then ./test/bats/bin/bats --timing test/serial.*.bats; else \
+		./test/bats/bin/bats --timing --jobs $(PARALLEL_JOBS) test/parallel.$*.*.bats; fi
```
polarathene, line 53: Not a huge fan of the change here.

You've got a conditional to handle serial vs parallel testing, and supporting commentary to further clarify that.

Let's change this to `test/part/0:` and after that `test/part/%`, I assume it's smart enough to only match and handle numbers after `0`, if not, we can just have clear commands `tests-serial` + `tests-parallel`, both called by `make tests`. I think that would be clearer?

Can you clarify if I understand this correctly? You are running `parallel` test files in parallel, but not their individual test cases within a file correct? Default parallel jobs is two, so for local users the resources required may be twice what was needed before? I think that was mostly only an issue for ClamAV which is now it's own serial test, thus probably nothing to worry about.

If a user does want to run tests all in serial, then setting `PARALLEL_JOBS=1` will do the trick?

.github/workflows/generic_test.yml
```diff
@@ -14,6 +14,9 @@ jobs:
   run-tests:
     name: 'Test'
     runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        part: [0, 1, 2, 3]
```
polarathene, line 19: Is this only to support the `make generate-accounts` for tests below?

Is that an issue for running tests locally in this manner too? I'd need to review the Github Actions docs. It's not clear to me if you're configuring the CI here to run the matrix on the same runner, or separate runners per `part`. I assume it's the latter.

Thus we're splitting the testing out to 4 runners / machines instead of the present 1 runner on a single machine? And then 1 of those will run all serial tests, while the other 3 run two parallel tests at a time?

Our CI has introduced some concurrency here in addition to parallelism? (_I think the technically correct term is the CI is providing parallelism, and BATS jobs is providing concurrency, but whatever_ :man_shrugging: :stuck_out_tongue: )

polarathene (changes requested):
The `Makefile` and a few others presently have some conflicts to resolve (_due to other PRs being merged, such as introducing ClamAV test file, while this didn't have that to rename into your parallel version at the time_).

When you've addressed the feedback, and if I've been granted permission to interactively rebase with a force push, I'll rewrite commit history to accommodate the file history the way I'd like it. Depending on my availability, the rebase may be delayed up until next weekend at the latest (hopefully).

polarathene:
> I know, it's a big PR. I know you don't like that.
> I think having these changes in one PR and not in separate ones (_this only applied to the first PR of slowly adopting the test suite to run in parallel_) is the right way.

It's not likely to be a problem when the history is properly fixed.

Limiting the amount of tests converted for running in parallel in this PR is the correct choice and keeps this concern with diff size under control :+1: 

---

> This change is not backwards compatible (by default!), because the `--jobs x` where `x > 0` requires GNU parallel to be installed.

Technically we could probably check if `parallel` is available, otherwise set the ENV to 1. That can be deferred until a contributor expresses any frustration regarding it. I'm sure we're all good with just using `PARALLEL_JOBS=1 make tests` if it were an issue anyway :+1:

.github/workflows/generic_test.yml
```diff
@@ -45,7 +48,7 @@ jobs:
           cache-from: type=local,src=/tmp/.buildx-cache
 
       - name: 'Run tests'
-        run: make generate-accounts tests
+        run: make generate-accounts test/part/${{ matrix.part }}
```
casperklein, line 51: Am I right, that `make generate-accounts` is run 4 times? For each `matrix.part`? If so, lets change it to:

```suggestion
	name: Prepare tests
        run: make generate-accounts
        name: Run tests
        run: make test/part/${{ matrix.part }}
```

Makefile
casperklein: With that, running `make tests` is equal to:

```bash
make test/part/0
make test/part/1
make test/part/2
make test/part/3
```

test/serial.setup.sh.bats
georglauterbach: > > Re-wrote history and I hope it is now easier to review.
> It happened in a few commits, if I have your permission I don't mind correct this history and doing a force-push?

Oh, I see. [`gitui`](https://github.com/extrawurst/gitui) showed them as renaming to me (maybe I did not pay enough attention). Please, go ahead :)

---

I will make sure resolve merge conflicts with `master` later.

georglauterbach:
> The `Makefile` and a few others presently have some conflicts to resolve (_due to other PRs being merged, such as introducing ClamAV test file, while this didn't have that to rename into your parallel version at the time_).

If you interactive rebase does not take care of all the conflicts I can after spare some time to resolve them too, so you don't have too much work with this :)

> When you've addressed the feedback, and if I've been granted permission to interactively rebase with a force push, I'll rewrite commit history to accommodate the file history the way I'd like it. Depending on my availability, the rebase may be delayed up until next weekend at the latest (hopefully).

Go ahead and do it, sound very good to me :) And thanks a lot :D

test/parallel.2.test_helper.bats
georglauterbach: Ah yes, this was due to me testing repeatedly, and I did not want to have that many containers running on my small machine. Can be removed :)

test/parallel.2.test_helper.bats
georglauterbach: See https://github.com/docker-mailserver/docker-mailserver/pull/2822#discussion_r996285441 :)

Makefile
georglauterbach: Good observation. This should go into another PR I think :)

Makefile
georglauterbach: Exactly. This is just running all the "dependencies" as they are called in `make` I think. I think this is pretty easy to understand, and it's not some elusive `make` syntax either :)

Makefile
georglauterbach: This part could get some rework. Depending on your rebase I will tackle it afterwards.

The docs preview explains the new strategy better: <https://pullrequest-2822--dms-doc-previews.netlify.app/contributing/general/#understanding-the-test-suite>. Locally, individual test files are run in parallel, but only those that have a `parallel` in their name.

---

> If a user does want to run tests all in serial, then setting `PARALLEL_JOBS=1` will do the trick?

Exactly.

.github/workflows/generic_test.yml
georglauterbach: > Thus we're splitting the testing out to 4 runners / machines instead of the present 1 runner on a single machine? And then 1 of those will run all serial tests, while the other 3 run two parallel tests at a time?

Yes

> Our CI has introduced some concurrency here in addition to parallelism?

Indeed :D I think this is not a problem though, as there are no write-after-x conflicts, because we do not write to a cache. Or de we? In the `generic-test`, there is no `cache-to: ..`, so all is well?

---

> Think the technically correct term is the CI is providing parallelism, and BATS jobs is providing concurrency

This is absolutely correct.

.github/workflows/generic_test.yml
georglauterbach: I will wait with this until after @polarathene has taken care of the history :)

polarathene:
> Go ahead and do it, sound very good to me :) And thanks a lot :D

I'll try get this done today or tomorrow if possible, but definitely by the weekend :sweat_smile:

georglauterbach:
@polarathene if you have the time figuring out how to put the tests into subdirectories, please go ahead :) The naming scheme that I introduced is just a "workaround" for the import errors in BATS that I did not yet have time to resolve. Possibly, constructing a complete path would solve the problems.. I may be able to have a look today in the evening. I will let you know.

.github/workflows/generic_test.yml
polarathene: Done, sorry for the delay!

It did not help that much with the review UI on Github (_due to a merge commit diff I think_), but the history looks like it should be preserved fine with a rebase merge now (_Github blame won't likely honour it, but other `git blame` tooling should_).

georglauterbach:
Closing in favour of #2850.

PR #2833, opened 2022-10-15
build: cleaned up Makefile

georglauterbach:
# Description

Got some inspiration from comments in #2822 and decided to clean up the whole `Makefile`.

## Type of change

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,10 +11,11 @@ export NAME       ?= $(IMAGE_NAME)
 all: lint build backup generate-accounts tests clean
 
 build:
-	@ DOCKER_BUILDKIT=1 docker build . \
+	@ DOCKER_BUILDKIT=1 docker build \
 		--tag $(IMAGE_NAME) \
 		--build-arg VCS_VERSION=$(shell git rev-parse --short HEAD) \
-		--build-arg VCS_REVISION=$(shell cat VERSION)
+		--build-arg VCS_REVISION=$(shell cat VERSION) \
+		.
 
 generate-accounts:
 	@ cp test/config/templates/postfix-accounts.cf test/config/postfix-accounts.cf
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,8 @@ export NAME       ?= $(IMAGE_NAME)
 all: lint build backup generate-accounts tests clean
 
 build:
-	@ DOCKER_BUILDKIT=1 docker build . --tag $(NAME) \
+	@ DOCKER_BUILDKIT=1 docker build . \
+		--tag $(IMAGE_NAME) \
 		--build-arg VCS_VERSION=$(shell git rev-parse --short HEAD) \
 		--build-arg VCS_REVISION=$(shell cat VERSION)
 
@@ -27,8 +28,8 @@ backup:
 clean:
 # remove test containers and restore test/config directory
 	-@ [[ -d testconfig.bak ]] && { sudo rm -rf test/config ; mv testconfig.bak test/config ; } || :
-	-@ for CONTAINER in $$(docker ps -a --filter name='^dms-test-.*|^hadolint$$|^eclint$$|^shellcheck$$' | sed 1d | cut -f 1-1 -d ' '); do docker rm -f $${CONTAINER}; done
-	-@ while read -r LINE; do [[ $${LINE} =~ test/.+ ]] && rm -rf $${LINE}; done < .gitignore
+	-@ for CONTAINER in $$(docker ps -a --filter name='^dms-test-.*|^mail_.*|^hadolint$$|^eclint$$|^shellcheck$$' | sed 1d | cut -f 1-1 -d ' '); do docker rm -f $${CONTAINER}; done
+	-@ while read -r LINE; do [[ $${LINE} =~ test/.+ ]] && sudo rm -rf $${LINE}; done < .gitignore
 
 # -----------------------------------------------
 # --- Tests  & Lints ----------------------------
--- a/Makefile
+++ b/Makefile
@@ -1,19 +1,23 @@
-SHELL = /bin/bash
+SHELL             := /bin/bash
+.SHELLFLAGS       += -e -u -o pipefail
 
-export NAME       ?= mailserver-testing:ci
-export IMAGE_NAME := $(NAME)
+export IMAGE_NAME := mailserver-testing:ci
+export NAME       ?= $(IMAGE_NAME)
 
 # -----------------------------------------------
-# --- Generic Build Targets ---------------------
+# --- Generic Targets ---------------------------
 # -----------------------------------------------
 
 all: lint build backup generate-accounts tests clean
 
 build:
-	@ DOCKER_BUILDKIT=1 docker build --tag $(NAME) \
+	@ DOCKER_BUILDKIT=1 docker build . --tag $(NAME) \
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -1,19 +1,23 @@
-SHELL = /bin/bash
+SHELL             := /bin/bash
+.SHELLFLAGS       += -e -u -o pipefail
 
-export NAME       ?= mailserver-testing:ci
-export IMAGE_NAME := $(NAME)
+export IMAGE_NAME := mailserver-testing:ci
+export NAME       ?= $(IMAGE_NAME)
```
casperklein, line 5: What's the difference between NAME/IMAGE?

Would be either one of them not enough?

georglauterbach: On paper, yes; in the future, yes; currently, no :) I want to migrate from `NAME` to `IMAGE_NAME`, but as of now, not all tests use `IMAGE_NAME`. I will resolve this later. We could add a comment though, if requested.

Makefile
```diff
@@ -1,19 +1,23 @@
...
 all: lint build backup generate-accounts tests clean
 
 build:
-	@ DOCKER_BUILDKIT=1 docker build --tag $(NAME) \
+	@ DOCKER_BUILDKIT=1 docker build . --tag $(NAME) \
 		--build-arg VCS_VERSION=$(shell git rev-parse --short HEAD) \
-		--build-arg VCS_REVISION=$(shell cat VERSION) \
-		.
+		--build-arg VCS_REVISION=$(shell cat VERSION)
+
```
polarathene, line 17: Why was the `.` shifted into before args? I had a quick look and didn't see any better ways to [set the context in the `docker build` docs](https://docs.docker.com/engine/reference/commandline/build/) other than the `.` :\

If this is what another maintainer prefers, I won't argue the change. I usually expect it at the end after options personally, and would have just preferred bumping `--tag` to a newline instead:

```Makefile
build:
	@ DOCKER_BUILDKIT=1 docker build \
	    --tag $(NAME) \
		--build-arg VCS_VERSION=$(shell git rev-parse --short HEAD) \
		--build-arg VCS_REVISION=$(shell cat VERSION) \
		.
```

Makefile
```diff
@@ -1,19 +1,23 @@
...
+generate-accounts:
+	@ cp test/config/templates/postfix-accounts.cf test/config/postfix-accounts.cf
+	@ cp test/config/templates/dovecot-masters.cf test/config/dovecot-masters.cf
```
polarathene, line 20: :heart:

Makefile
```diff
@@ -23,24 +27,13 @@ backup:
...
-	-@ sudo rm -rf test/onedir test/alias test/quota test/relay test/config/dovecot-lmtp/userdb test/config/key* test/config/opendkim/keys/domain.tld/ test/config/opendkim/keys/example.com/ test/config/opendkim/keys/localdomain2.com/ test/config/postfix-aliases.cf test/config/postfix-receive-access.cf test/config/postfix-receive-access.cfe test/config/dovecot-quotas.cf test/config/postfix-send-access.cf test/config/postfix-send-access.cfe test/config/relay-hosts/chksum test/config/relay-hosts/postfix-aliases.cf test/config/dhparams.pem test/config/dovecot-lmtp/dh.pem test/config/relay-hosts/dovecot-quotas.cf test/config/user-patches.sh test/alias/config/postfix-virtual.cf test/quota/config/dovecot-quotas.cf test/quota/config/postfix-accounts.cf test/relay/config/postfix-relaymap.cf test/relay/config/postfix-sasl-password.cf test/duplicate_configs/
+	-@ for CONTAINER in $$(docker ps -a --filter name='^dms-test-.*|^hadolint$$|^eclint$$|^shellcheck$$' | sed 1d | cut -f 1-1 -d ' '); do docker rm -f $${CONTAINER}; done
```
polarathene, line 30: Just to clarify, this `clean` line is to handle removing any hanging containers if something went wrong with tests that didn't properly cleanup their containers?

The `dms-test-*` is going to be a new prefix for our container names instead of `mail_` that was common on many?

Makefile
```diff
@@ -23,24 +27,13 @@ backup:
...
+	-@ while read -r LINE; do [[ $${LINE} =~ test/.+ ]] && rm -rf $${LINE}; done < .gitignore
```
polarathene, line 31: I like this, smart :sunglasses: 

Is this going to be ok with `test/config/key*`? The `.gitignore` also has some mixed usage of a trailing `/` for directories from the looks of it. I assume none of that is an issue though?

There may be some irrelevant lines we're cleaning up on now, I think the `letsencrypt` one (_in the `.gitignore` may no longer be used in tests_), but we can do some house keeping on that at another time :+1: 

`sudo` might be needed still? If these are files being created from a container mount, then it may be writing to the disk as the `root` user / group on the host. If it's just stuff tests are generating without that issue though, not needing `sudo` would be great :grinning:

polarathene (dismissed):
Just some questions, but nothing block, LGTM :+1:

Makefile
georglauterbach: I did this to avoid the single dot on the new line, which looks weird to me. I think this might be a good compromise:

```make
build:
	@ DOCKER_BUILDKIT=1 docker build . \
	        --tag $(NAME) \
		--build-arg VCS_VERSION=$(shell git rev-parse --short HEAD) \
		--build-arg VCS_REVISION=$(shell cat VERSION)
```

Makefile
georglauterbach: Indeed. In the future, all the test containers should get the `dms-test-` prefix. I will re-add the `mail_` prefix again though, right :)

Makefile
georglauterbach: > I like this, smart sunglasses

:D

> Is this going to be ok with `test/config/key*`? The `.gitignore` also has some mixed usage of a trailing `/` for directories from the looks of it. I assume none of that is an issue though?

I double-checked it, we should be fine. This is using something that `shellcheck` would actually complain about: Bash "globbing" the star at the end :) This is what we want and need. I also made sure only directories have trailing slashes in `.gitignore`.
 
> There may be some irrelevant lines we're cleaning up on now, I think the `letsencrypt` one (_in the `.gitignore` may no longer be used in tests_), but we can do some house keeping on that at another time +1

Removed the LE line, kept others there. Another housekeeping PR can solve this, exactly :)

> `sudo` might be needed still? If these are files being created from a container mount, then it may be writing to the disk as the `root` user / group on the host. If it's just stuff tests are generating without that issue though, not needing `sudo` would be great grinning

Indeed, I added `sudo` to the `rm` call.

georglauterbach:
Applied PR feedback and merged master to be up-to-date. Also enabled auto-merge :)

Makefile
casperklein: > In the future, all the test containers should get the dms-test- prefix.

Good addition.

Makefile
casperklein: The man page synopsis mentions the build context as last argument. Not sure, how future proof the above will be. Also IMO we should avoid unusual syntax and revert them were we can instead of introducing new ones ;-)
Looking at you weird space in the shebang lines 😆

georglauterbach: Convinced 🙃

casperklein (approved):

polarathene (approved):

PR #2854, opened 2022-10-23
scripts: Improve error handling, when parameters are missing

casperklein:
# Description

When parameters are missing (alias destination in that case), this error is shown:

```bash
./setup.sh alias add postmaster@office
/usr/local/bin/setup: line 119: 2: unbound variable

--- UNCHECKED ERROR
  - script    = ./setup.sh
  - function  = _main
  - command   = ${CRI} exec "${USE_TTY}" "${CONTAINER_NAME}" setup "${@}"
  - line      = 229
  - exit code = 1

This should not have happened. Please file a bug report.
```

This PR solves that, by showing the usage instead.

## Type of change

## Checklist:

### DIFF
--- a/target/bin/addalias
+++ b/target/bin/addalias
@@ -5,7 +5,7 @@ source /usr/local/bin/helpers/index.sh
 
 function _main
 {
-  [[ ${1:-} == 'help' ]] && { __usage ; exit 0 ; }
+  [[ ${#} -lt 2 ]] && { __usage ; exit 0 ; }
 
   local MAIL_ALIAS="${1}"
   local RECIPIENT="${2}"
--- a/target/bin/addalias
+++ b/target/bin/addalias
@@ -5,7 +5,7 @@ source /usr/local/bin/helpers/index.sh
 
 function _main
 {
-  [[ ${#} -lt 2 ]] && { __usage ; exit 0 ; }
+  _require_n_parameters_or_print_usage 2 "${@}"
 
   local MAIL_ALIAS="${1}"
   local RECIPIENT="${2}"
--- a/target/scripts/helpers/utils.sh
+++ b/target/scripts/helpers/utils.sh
@@ -50,7 +50,7 @@ function _require_n_parameters_or_print_usage
     fi
   else
     # n parameters
-    [[ ${#} -lt $COUNT ]] && FAIL=1
+    [[ ${#} -lt ${COUNT} ]] && FAIL=1
   fi
 
   if [[ ${FAIL} -eq 1 ]]
--- a/target/scripts/helpers/utils.sh
+++ b/target/scripts/helpers/utils.sh
@@ -37,25 +37,10 @@ function _chown_var_mail_if_necessary
 
 function _require_n_parameters_or_print_usage
 {
-  local COUNT FAIL=0
+  local COUNT
   COUNT=${1}
   shift
 
-  # One or more parameters?
-  if [[ ${COUNT} -eq 1 ]]; then
-    # one parameter
-    if [[ ${1:-} == 'help' ]] || [[ -z ${1:-} ]]
-    then
-      FAIL=1
-    fi
-  else
-    # n parameters
-    [[ ${#} -lt ${COUNT} ]] && FAIL=1
-  fi
-
-  if [[ ${FAIL} -eq 1 ]]
-  then
-    __usage
-    exit 0
-  fi
+  [[ ${1:-} == 'help' ]]  && { __usage ; exit 0 ; }
+  [[ ${#} -lt ${COUNT} ]] && { __usage ; exit 1 ; }
 }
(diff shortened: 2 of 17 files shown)


target/bin/addalias
```diff
@@ -5,7 +5,7 @@ source /usr/local/bin/helpers/index.sh
 
 function _main
 {
-  [[ ${1:-} == 'help' ]] && { __usage ; exit 0 ; }
+  [[ ${#} -lt 2 ]] && { __usage ; exit 0 ; }
```
polarathene, line 8: Would it make sense to be DRY here and just have these all use a helper?

`addalias` and `adddovecotmasteruser` both had the same content originally, but you've diverged it due to number of args? If we had a method that just set the `-lt` value to compare to, would that work for all cases being changed here?

polarathene (approved):

target/bin/addalias
casperklein: Hm, I had the same idea. But then I dismissed the thought.

For the previous `  [[ ${1:-} == 'help' ]] && { __usage ; exit 0 ; }` that would have been the same. We haven't used a function here also.

I think this is so small and flat - not worth it. And who knows, maybe there are some cases in the future, where some files needs more flexibility than a standard function can provide?

After all, I am fine to change that to use a function - if the majority want's that 😎

georglauterbach (approved):

georglauterbach (approved):

target/bin/addalias
polarathene: > We haven't used a function here also.

That is because before I did a refactor pass, it was a mess and I wanted to minimize the already large diff for that PR I think :sweat_smile: 

> I think this is so small and flat - not worth it. And who knows, maybe there are some cases in the future, where some files needs more flexibility than a standard function can provide?

I believe the `delmailuser` already has something in place to check for `-y`.

Up to you, I just noticed you're introducing two variants now, and one of those becomes multiple lines, while the other uses new syntax (_I roughly could guess what it was, but still looked it up to verify_).

Personally for maintaining things like this, I prefer the DRY approach suggested, even though it's fairly small, I've seen contributions in the past for these files make small changes to a subset of files (or just one file) with the shared duplicate code, and then we get this divergence that's caused some bugs. It's not likely to happen with current maintainers, but it has been an issue before.

> After all, I am fine to change that to use a function - if the majority want's that

I'm fine either way. Ideally if I had the time, I'd rather port this all to something like rust :sweat_smile:

polarathene (approved):

target/bin/addalias
casperklein: Check this commit: [`4f2c92c`](https://github.com/docker-mailserver/docker-mailserver/pull/2854/commits/4f2c92c0925df46781e0795b1009eb58bbe5cccf)

target/scripts/helpers/utils.sh
```diff
@@ -34,3 +34,28 @@ function _chown_var_mail_if_necessary
...
+  else
+    # n parameters
+    [[ ${#} -lt ${COUNT} ]] && FAIL=1
+  fi
+
+  if [[ ${FAIL} -eq 1 ]]
+  then
+    __usage
+    exit 0
+  fi
```
polarathene, line 60: Thanks for doing this :)

Would this be simpler? Or does it not cover something well?

```suggestion
  if [[ ${1:-} == 'help' ]] || [[ -z ${1:-} ]] || [[ ${#} -lt ${COUNT} ]]
    then
    __usage
    exit 0
  fi
```

polarathene (approved):

target/scripts/helpers/utils.sh
casperklein: I think so 😆. I also had another idea that I want to distinguish between 'help' is given and everything else: `exit 0` / `exit 1`.

polarathene: Does this work? :sweat_smile:

```bash
  if [[ ${1:-} == 'help' ]] || [[ -z ${1:-} ]] || [[ ${#} -lt ${COUNT} ]]
    then
    __usage

    [[ ${1:-} == 'help' ]] && exit 1
    exit 0
  fi
```

I'm not sure if the `-z` condition is doing anything for us though? Doesn't the `-lt` also cover that?

casperklein: Love this 🚀

georglauterbach (approved):

polarathene (dismissed):
Awesome :tada: 

---

I will point out that while in the review page I see a few other files that could throw in:

```bash
_require_n_parameters_or_print_usage 1 "${@}"
```

Like `excluderelaydomain` and `setquota`, both still have the existing `help` handling and could be consistent with the other files changed here, but not too fussed either way :sweat_smile:

casperklein:
> and could be consistent with the other files changed here, but not too fussed either way 😅

Consistency FTW 🚀 

You know, that I am not a fan of "get this merged" and providing a follow up in such cases 😆 Therefore: https://github.com/docker-mailserver/docker-mailserver/pull/2854/commits/1935bb077c39f4982d0b9b150020e3fb378bc2b2

georglauterbach (approved):

polarathene (approved):

PR #2857, opened 2022-10-23
ci: run tests in parallel (part 1)

georglauterbach:
# Description

See notable changes below. Commit messages should help too.

Thanks to @polarathene for helping me with the history. I tried to apply the same pattern with the new commits.

---

## Notable Changes

- tests can now be arbitrarily grouped into subdirectories
- some tests (test files to be more precise) are now run in parallel
- CI will now spawn 4 jobs to run the whole test suite in parallel

## Type of change

## Checklist:


georglauterbach:
[16a28e9](https://github.com/docker-mailserver/docker-mailserver/pull/2857/commits/16a28e968ef21654f7926ba583925fa480a25887) and [93881a9](https://github.com/docker-mailserver/docker-mailserver/pull/2857/commits/93881a9b709368b3fe3a346e3505ba1aae319f85) are the interesting commits :)

polarathene:
Sorry about the delay, I may need to wait until the weekend again :sweat_smile: 

It looks like glancing over commit history that you continued off your previous branch rather than extending from the one I revised?

If you're ok with it, I'll carry over any changes from here onto the branch I revised, rewrite that history again so the rename is only happening once, and I can push it to this PR instead of my own? (_I had enabled the checkbox for reviewers to have rights to update my branch but apparently that wasn't working for some reason_ :thinking: )

georglauterbach:
> Sorry about the delay, I may need to wait until the weekend again 😅

No time pressure here :)

> It looks like glancing over commit history that you continued off your previous branch rather than extending from the one I revised?

Uuuufffff - I thought I  had used your commit history, but you're right, I did not. Silly mistake from my side - but:
 
> If you're ok with it, I'll carry over any changes from here onto the branch I revised, rewrite that history again so the rename is only happening once, and I can push it to this PR instead of my own? (_I had enabled the checkbox for reviewers to have rights to update my branch but apparently that wasn't working for some reason_ 🤔 )

I am very much fine with this 👍🏼 Go ahead when you have time :D
I think the checkboxes for "let maintainers edit the PR" is about changes in the webinterface..

georglauterbach:
@polarathene I marked this WIP. Wehn you revised the history, please mark it as ready for review again :D

polarathene:
I've dropped the commit cited below for the following reasons:
- The linked PR comment concern was about running `generate-accounts` multiple times. This doesn't change that, the `matrix` feature will run all `steps` on separate CI instances, thus will happen regardless and is necessary.
- `CI: true` I assume from your commit message you maybe were wanting pretty output instead? Github has limited support here, it's able to support colour, but not other features/syntax used to manipulate terminal output AFAIK. I had looked into this back in August, and raised an [issue at `bats-core`](https://github.com/bats-core/bats-core/issues/639#event-7246625622) which explains that `--pretty` will not work.

If I'm mistaken and you intended something else, make this change in a separate PR as it's unrelated to this one :+1: 

---

## Dropped commit

**Commit message (original commit title differed):**
```
tests(CI): Drop `CI: true` + separate step for `generate-accounts`
===
See https://github.com/docker-mailserver/docker-mailserver/pull/2822#discussion_r996283731

I also removed `CI: true` because I think GH Runners can now properly handle colored TTY output.
```

**Commit Diff:**
```diff
+      - name: 'Prepare tests'
+        run: make generate-accounts
+
      - name: 'Run tests'
-        run: make generate-accounts tests/${{ matrix.part }}
-        env:
-          CI: true
+        run: make tests/${{ matrix.part }}
```

polarathene:
> [16a28e9](https://github.com/docker-mailserver/docker-mailserver/pull/2857/commits/16a28e968ef21654f7926ba583925fa480a25887) and [93881a9](https://github.com/docker-mailserver/docker-mailserver/pull/2857/commits/93881a9b709368b3fe3a346e3505ba1aae319f85) are the interesting commits :)

These are now https://github.com/docker-mailserver/docker-mailserver/pull/2857/commits/369e4b47f5b447bb67c19a2dbf3a638961c36fe0 and https://github.com/docker-mailserver/docker-mailserver/pull/2857/commits/5789915165765abc2f93df333766cac46e456a2e respectively.

polarathene:
Commit history has been revised, here's a breakdown:

1. https://github.com/docker-mailserver/docker-mailserver/pull/2857/commits/3fc8945fa5ff827ccd92b294219fa8af54be852b - Move all existing tests into their `serial` / `parallel` subdirectories.
2. https://github.com/docker-mailserver/docker-mailserver/pull/2857/commits/638eedbb5ffd409269e01548b10b20f65f3d4b8d - @georglauterbach modified some tests here, presumably as an example of changes / improvements for a future PR to make to all other tests as a convention. I have written a more detailed commit message highlighting the changes covered in the diffs, noting the common change across files and some file specific ones that stood out.
3. https://github.com/docker-mailserver/docker-mailserver/pull/2857/commits/01a498d12fbfad12c4de7ba970a62b268fc4250d - Has a bit more going on. I'm not quite following the comment explanation for using explicit / hard-coded container names instead of the ENV var that had been working, apparently that breaks with the parallel feature? (_couldn't we just assign different var names though?_)
4. https://github.com/docker-mailserver/docker-mailserver/pull/2857/commits/7b009077e65f3f8fc20397a70b725c10fd84dbfa - Adds what looks like a reference / example (`template.bats`) to copy / paste for new tests, or for updating existing tests with to support parallel testing? The additional serial test I haven't looked over yet, and doesn't seem relevant to this PR, but a separate test in itself that could have been added via it's own PR?
5. https://github.com/docker-mailserver/docker-mailserver/pull/2857/commits/369e4b47f5b447bb67c19a2dbf3a638961c36fe0 - Besides the file renames, the bulk of files changed is just the top line to use an absolute path via `REPOSITORY_ROOT` (_defined in the `Makefile`_), this allows for the nested test folders to properly load test helpers.
6. https://github.com/docker-mailserver/docker-mailserver/pull/2857/commits/5789915165765abc2f93df333766cac46e456a2e - Finally, all changes to `Makefile` and the CI testing workflow is bundled into this commit. The commit message is revised based on the content of multiple original commit messages for these file changes from @georglauterbach .

test/tests/parallel/set2/template.bats
```diff
@@ -0,0 +1,55 @@
...
+  docker run --rm --detach --tty \
+    --name "${CONTAINER_NAME}" \
+    --hostname mail.my-domain.com \
+    --volume "${PRIVATE_CONFIG}:/tmp/docker-mailserver" \
+    --env ENABLE_AMAVIS=0 \
+    --env ENABLE_CLAMAV=0 \
+    --env ENABLE_UPDATE_CHECK=0 \
+    --env ENABLE_SPAMASSASSIN=0 \
+    --env ENABLE_FAIL2BAN=0 \
+    --env LOG_LEVEL=debug \
```
polarathene, line 35: I noticed you added most of these ENV to some other tests you've updated (_two other files I think_).

Most of those are default `0`, but if you'd like to encourage a common base of ENV here, using the helper I had setup a few tests to use is a good way to do that.

They could probably be moved into that helpers base ENV, and this test adapted to use the helper. That does not need to happen for this PR though, which already has a lot going on.

I do like that it's more clear here, but I'm less fond of the copy/paste that I assume would occur from this template being applied to a wide number of test files.

I have also been taking interest with the "new" `docker compose` v2 command and leveraging `compose.yml` with the features it offers (_it may have been called profiles, or overrides_), which allows for establishing a base config, and providing other configs that can override or extend that.

It may be nicer to manage (we can still create some ENV var like `CONTAINER_NAME` and `PRIVATE_CONFIG` for it to use AFAIK, but avoid copy/paste, and the awkward helper approach I initially implemented.

If I have time in dec/jan I may take a shot at it, but anyone is welcome to investigate that if they've got the time :muscle:

test/tests/parallel/set2/test_helper.bats
```diff
@@ -97,29 +101,35 @@ load 'test_helper/common'
   assert_success
 }
 
-@test "wait_for_finished_setup_in_container" {
+@test "${TEST_NAME_PREFIX} wait_for_finished_setup_in_container" {
+  # the container name is used here explicitly, because it cannot be a local
+  # variable as the `teardown` function would not have the correct scope
+  # and does not see the variable
```
polarathene, line 107: Only the `teardown()` needs the explicit value then? You could still use a variable for everywhere else that it is used?

test/tests/parallel/set2/test_helper.bats
```diff
@@ -97,29 +101,35 @@ load 'test_helper/common'
...
+    -t "${IMAGE_NAME}"
 
-  teardown() { docker rm -f "${CONTAINER_NAME}"; }
+  teardown() { docker rm -f dms-test-wait_for_finished_setup; }
 
   # the setup should not be finished immediately after starting
-  ! TEST_TIMEOUT_IN_SECONDS=0 wait_for_finished_setup_in_container "${CONTAINER_NAME}"
+  ! TEST_TIMEOUT_IN_SECONDS=0 wait_for_finished_setup_in_container dms-test-wait_for_finished_setup
+
+  docker exec dms-test-wait_for_finished_setup setup email add 'example_user@domain.com' 'password'
```
polarathene, line 123: Is this needed due to the changed `PRIVATE_CONFIG`?

test/tests/parallel/set2/test_helper.bats
```diff
@@ -130,43 +140,55 @@ load 'test_helper/common'
   assert_failure
 }
 
-@test "container_has_service_running/wait_for_service" {
+@test "${TEST_NAME_PREFIX} container_has_service_running/wait_for_service" {
+  # the container name is used here explicitly, because it cannot be a local
+  # variable as the `teardown` function would not have the correct scope
+  # and does not see the variable
```
polarathene, line 146: Same feedback here regarding `teardown()` shouldn't require usage elsewhere to avoid referencing a var.

test/tests/parallel/set2/test_helper.bats
```diff
@@ -130,43 +140,55 @@ load 'test_helper/common'
...
   # variable not local to make visible to teardown
-  CONTAINER_NAME=$(docker run -d --rm \
+  docker run -d --rm --name dms-test-services_running \
     -v "${PRIVATE_CONFIG}":/tmp/docker-mailserver \
     -h mail.my-domain.com \
-    -t "${NAME}")
+    -e ENABLE_CLAMAV=0 \
+    -e ENABLE_AMAVIS=0 \
+    -e ENABLE_FAIL2BAN=0 \
+    -e ENABLE_SPAMASSASIN=0 \
```
polarathene, line 158: These ENV were added for what reason?

test/tests/parallel/set2/test_helper.bats
```diff
@@ -130,43 +140,55 @@ load 'test_helper/common'
...
+    -t "${IMAGE_NAME}"
 
-  teardown() { docker rm -f "${CONTAINER_NAME}"; }
+  teardown() { docker rm -f dms-test-services_running; }
+
+  docker exec dms-test-services_running setup email add 'example_user@domain.com' 'password'
+  wait_for_finished_setup_in_container dms-test-services_running
+  wait_for_changes_to_be_detected_in_container dms-test-services_running
```
polarathene, line 165: Likewise, is this related to using a different `PRIVATE_CONFIG` for running in parallel?

Your additions here are a bit more extensive than the previous test (_although your addition there had `wait_for_finished_setup_in_container` both before and after it which seemed a bit odd_)

test/tests/serial/setup.sh.bats
```diff
@@ -0,0 +1,150 @@
+load "${REPOSITORY_ROOT}/test/test_helper/common"
+
+# ! NEEDS SERIAL EXECTION - DO NOT RUN IN PARALLEL
+
+IMAGE_NAME=${NAME:-mailserver-testing:ci}
+TEST_NAME_PREFIX='checking setup.sh: '
```
polarathene, line 6: I have just realized this was not a new test, but a changed version of `serial/no_container.bats`?

Perhaps this was an error on my part, and I missed that it was renamed :grimacing:

polarathene (commented):
I've now reviewed almost everything (_I still need to go over `setup.sh.bats` (previously `no_containers.bats`_).

Majority of feedback is questions for clarification on changes, or additions and possible improvements. No change requests atm :+1:

Makefile
```diff
@@ -1,8 +1,10 @@
...
-.SHELLFLAGS       += -e -u -o pipefail
+SHELL       := /bin/bash
+.SHELLFLAGS += -e -u -o pipefail
 
-export IMAGE_NAME := mailserver-testing:ci
-export NAME       ?= $(IMAGE_NAME)
+PARALLEL_JOBS          ?= 2
+export REPOSITORY_ROOT := $(CURDIR)
+export IMAGE_NAME      ?= mailserver-testing:ci
+export NAME            ?= $(IMAGE_NAME)
```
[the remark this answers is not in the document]
georglauterbach: Exactly

Makefile
```diff
@@ -33,14 +35,23 @@ clean:
...
 
-tests:
-	@ ./test/bats/bin/bats --timing test/*.bats
+tests: tests/serial tests/parallel/set1 tests/parallel/set2 tests/parallel/set3
+
+tests/serial:
+	@ ./test/bats/bin/bats --timing --jobs 1 test/$@/**.bats
+
+tests/parallel/set%:
+	@ ./test/bats/bin/bats --timing --jobs $(PARALLEL_JOBS) test/$@/**.bats
```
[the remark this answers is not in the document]
georglauterbach: > `test/$@/**.bats` expands to the target like `tests/parallel/set%` (_which then expands to the relevant `%` value_) for the `$@` portion, and after that it grabs whatever `.bats` files are found recursively nested in that location?

Yes :)

> So `tests/serial` and `tests/parallel/set%` are just bulk / batch versions of `test/%` right? (_ignoring the `--jobs` option_)

Indeed

Makefile
```diff
@@ -33,14 +35,23 @@ clean:
...
 
 test/%:
-	@ ./test/bats/bin/bats --timing $@.bats
+	@ shopt -s globstar nullglob ; ./test/bats/bin/bats --timing test/tests/**/{$*,}.bats
```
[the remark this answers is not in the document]
georglauterbach: > I am familiar with this `shopt -s globstar nullglob` line from the change detection scripts. It's just excluding any paths that aren't valid?

Yes

> What's going on with the `**/{$*,}.bats` portion? This is like the `/**.bats` from earlier but if I understand your commit message and example for this (thanks!), it's adding the support to provide multiple files to test instead of just one?

Yes :)

> The example in the commit message was `make test/clamav,template`. So it is `make test/%`, and then we specify the test names (_without serial / parallel folder location, nor `.bats` extension_), and separate with a `,` delimiter if we want to test multiple files?

Exactly; you could actually specify path inside the the `test/tests/` directory as well, that would work too, without the `.bats` extension of course.

> It would be good for the test docs to mention that, especially that the directory path from `test/tests/` to the `.bats` file is not relevant.

Will do

> I assume you can provide additional specificity there, in the event that we had two tests with the same filename at different locations? (_Otherwise it'd run both of those right?_)

Yes

test/tests/parallel/set1/default_relay_host.bats
```diff
@@ -0,0 +1,30 @@
+load "${REPOSITORY_ROOT}/test/test_helper/common"
+
+TEST_NAME_PREFIX='default relay host:'
+CONTAINER_NAME='dms-test-default_relay_host'
+RUN_COMMAND=('run' 'docker' 'exec' "${CONTAINER_NAME}")
```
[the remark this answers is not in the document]
georglauterbach: We can probably remove it in this  case, yes, but this should be the default when more the one `run docker exec...` command is executed.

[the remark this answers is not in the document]
georglauterbach: > This is a bit redundant due to the single usage below, but I take it you'd like to implement this for all tests in future?

We _could_ probably remove it in this case, yes, but this should definitely be the default when more the one `run docker exec ...` command is executed.

> How would you approach that with tests that have multiple containers like the `dovecot_inet_protocols.bats` file you also updated?

I'm not sure about that to be honest. Probably not use such an array though..

test/tests/parallel/set2/test_helper.bats
```diff
@@ -35,56 +35,60 @@ load 'test_helper/common'
...
+@test "${TEST_NAME_PREFIX} wait_for_smtp_port_in_container aborts wait after timeout" {
   local CONTAINER_NAME
   CONTAINER_NAME=$(docker run --rm -d alpine sleep 100)
   SECONDS=0
   TEST_TIMEOUT_IN_SECONDS=2 run wait_for_smtp_port_in_container "${CONTAINER_NAME}"
   [[ ${SECONDS} -ge 2 ]]
   assert_failure
   assert_output --partial "Timed out on command"
+
+  docker rm -f "${CONTAINER_NAME}"container_has_service_running/wait_for_service
```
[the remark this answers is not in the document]
georglauterbach: Definitely not 🙈 I will remove that :) Thanks for spotting!

test/tests/parallel/set2/test_helper.bats
georglauterbach: I think I actually did not write this comment, but it was in the file before. Maybe I can have a look later.

test/tests/serial/setup.sh.bats
georglauterbach: > I have just realized this was not a new test, but a changed version of `serial/no_container.bats`?

Indeed

> Perhaps this was an error on my part, and I missed that it was renamed :grimacing: 

No problem I guess :)

test/tests/parallel/set1/default_relay_host.bats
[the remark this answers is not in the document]
casperklein: Is there a need for an array? Less typing and same result:

```
RUN_COMMAND="run docker exec '${CONTAINER_NAME}'"

${RUN_COMMAND} ..
```

Also, this variable might be renamed to `$RUN_IN_CONTAINER`.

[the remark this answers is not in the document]
georglauterbach: > Is there a need for an array? Less typing and same result:
> 

I'd say yes, because

> ```
> RUN_COMMAND="run docker exec '${CONTAINER_NAME}'"
> 
> ${RUN_COMMAND} ..
> ```

exploits the fact that the string is split here _because_ the variable is _not_ properly quoted. ShellCheck would show this as well.

> Also, this variable might be renamed to `$RUN_IN_CONTAINER`.

👍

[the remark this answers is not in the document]
casperklein: It's not exploiting, it's working as expected 😆

> is not properly quoted

A variable don't have to be quoted. There is nothing wrong in not quoting them, for the exact reason in my example. It's just, that most of the time you don't want word splitting, so best practice is to quote them. But when you want it, omit the quotes.

That's also the reason, why shellcheck will not complain about it, if used as "command". Only when used as "parameter", a warning is given:

```
# good
${RUN_COMMAND} ..

# maybe problematic (as parameter unquoted)
my-command ${RUN_COMMAND}
>> Double quote to prevent globbing and word splitting. shellcheck(SC2086)
```

shellcheck gives just an info, to state how things work. And if it works like some expects, the warning can be ignore.

Leave it or change, I just wanted to correct the wrong statements made. ✌️

test/tests/parallel/set2/template.bats
georglauterbach: You're right, doing this in a common helper or with Compose v2 is better. I will revert these changes for default-values.

test/tests/parallel/set2/test_helper.bats
georglauterbach: Yes, it is :)

test/tests/parallel/set1/default_relay_host.bats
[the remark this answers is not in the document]
georglauterbach: I found a better approach - using a function :)

test/tests/parallel/set2/test_helper.bats
georglauterbach: Reverted those for which the default is already 0

test/tests/parallel/set2/test_helper.bats
georglauterbach: Yes, this is related to the new `PRIVATE_CONFIG`. The additions here are those used to check whether the container has started properly - the other test for which I added these lines were checking the startup itself, hence less changes from my side with these tests.

test/tests/parallel/set3/helper-functions.bats
```diff
@@ -0,0 +1,41 @@
...
+    --hostname mail.my-domain.com \
+    --volume "${PRIVATE_CONFIG}":/tmp/docker-mailserver \
+    --volume "$(pwd)/test/test-files":/tmp/docker-mailserver-test:ro \
+    --cap-add=NET_ADMIN \
+    --env ENABLE_AMAVIS=0 \
+    --env ENABLE_CLAMAV=0 \
+    --env ENABLE_UPDATE_CHECK=0 \
+    --env ENABLE_SPAMASSASSIN=0 \
+    --env ENABLE_FAIL2BAN=0 \
+    --env ENABLE_FETCHMAIL=1 \
```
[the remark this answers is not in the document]
georglauterbach: Same as above.

test/tests/parallel/set2/test_helper.bats
georglauterbach: I did not put these in place, but cleaned up some of them as the notice is confusing I think.

test/tests/parallel/set2/test_helper.bats
georglauterbach: Reverted the explicit naming, and therefore this change as well :)

test/tests/parallel/set2/test_helper.bats
georglauterbach: Reverted this change.

test/tests/parallel/set2/test_helper.bats
georglauterbach: This is now reverted.

georglauterbach:
Alright, I thought I solved the issue because I was able to reproduce it locally, but now the issue is another one: <https://github.com/docker-mailserver/docker-mailserver/actions/runs/3384677405/jobs/5621936134#step:5:330>.

I do not see _any_ reason why this test would _reliably pass_ locally on my machine while it fails during CI on a GH Actions runner.. Any idea why the container seems to not exist (` Error: No such container: ccca8332c747d62303f82ca5c1cd76619022928a51a1116c89ff20bcd396226a`)?

georglauterbach:
So I revised the commit history and did some things in a slightly different way. Had to force-push the branch - hope you don't mind.

test/helper/common.bash
```diff
@@ -0,0 +1,259 @@
...
+
+function _run_in_container() {
+  run docker exec "${CONTAINER_NAME}" "${@}"
+}
+
+function _default_teardown() {
+  docker rm -f "${CONTAINER_NAME}"
+}
+
+# -------------------------------------------------------------------
```
georglauterbach, line 20: From hereon, its simply a copy of `test_helper/common.bash`.

test/tests/serial/setup.sh.bats
georglauterbach: I will remove this file with one of the next (and hopefully final) commits.

Makefile
```diff
@@ -33,14 +37,27 @@ clean:
...
+# --- Tests  ------------------------------------
 # -----------------------------------------------
 
-tests:
-	@ ./test/bats/bin/bats --timing test/*.bats
+tests: ALWAYS_RUN
+	@ $(MAKE) generate-accounts tests/serial
+	@ $(MAKE) generate-accounts tests/parallel/set1
+	@ $(MAKE) generate-accounts tests/parallel/set2
+	@ $(MAKE) generate-accounts tests/parallel/set3
```
georglauterbach, line 47: I switched to this syntax as I encountered issues before when `generate-accounts` was not run before certain parallel tests were run because of prior modifications. Just FYI.

casperklein: BTW, do I miss something, where is the ALWAYS_RUN target? I only see references to it.

georglauterbach: At the top of the file; it is **not** defined as a target like `ALWAYS_RUN:`, but just as `.PHONY: ALWAYS_RUN` :)

Makefile
```diff
@@ -1,8 +1,12 @@
...
+.SHELLFLAGS += -e -u -o pipefail
 
-export IMAGE_NAME := mailserver-testing:ci
-export NAME       ?= $(IMAGE_NAME)
+PARALLEL_JOBS          ?= 2
+export REPOSITORY_ROOT := $(CURDIR)
+export IMAGE_NAME      ?= mailserver-testing:ci
+export NAME            ?= $(IMAGE_NAME)
+
+.PHONY: ALWAYS_RUN
```
georglauterbach, line 9: @casperklein here is what you were looking for :)

casperklein: I am not sure if I am wrong (not an Makefile expert at all😆 ) or you. What is the purpose of ALWAYS_RUN at the moment? 
You cannot run "make ALWAYS_RUN".

For example, I have this:

```
.PHONY: pull start

pull:
        @docker-compose pull

start: pull
		@docker-compose up
```

`pull/start` is phony, because there is no file `pull/start` to build. In the `start` target, `pull` is defined as dependency.

With this I can use `make pull` or `make start` (which also executes `docker-compose pull` before).

That said, I've no clue whats going on with ALWAYS_RUN 😄

georglauterbach: > What is the purpose of ALWAYS_RUN at the moment? You cannot run "make ALWAYS_RUN".

At the moment, `ALWAYS_RUN` can be provided as a dependency to a target, which will make the target itself behave as if was declared `.PHONY`. `ALWAYS_RUN` itself is not actually a real target and you cannot run it.

> For example, I have this:
> 
> ```
> .PHONY: pull start
> 
> pull:
>         @docker-compose pull
> 
> start: pull
> 		@docker-compose up
> ```
> 
> `pull/start` is phony, because there is no file `pull/start` to build.

A target is not automatically `.PHONY` only because there is not file(s) tracking it, AFAIK. I have a lot of experience with `make`, but I have not read that targets become `.PHONY` implicitly.

> That said, I've no clue whats going on with ALWAYS_RUN

`ALWAYS_RUN` is just a convenience to not declare every target `.PHONY` individually, but have one `.PHONY` "target" (not actually, but `make` does not really care) that you can add as a dependency to make other targets behave as if they were `.PHONY`.

> > This suffices to tell make to always run the target.
> 
> Which target? I am on the go, I'll play/test later a bit.

Every target that has `ALWAYS_RUN` as a dependency. I said _target;; in `make´, this is actually a _recipe_ 😄

georglauterbach: Here is an example I came up with:

```make
.PHONY: ALWAYS_RUN

main.o: main.c
	@ echo "not phony"
	@ gcc -c $^ -o $@

.PHONY: main1.o
main1.o: main1.c
	@ echo "explicit phony"
	@ gcc -c $^ -o $@

main2.o: main2.c ALWAYS_RUN
	@ echo "implicit phony"
	@ gcc -c main2.c -o $@

clean: ALWAYS_RUN
	@ rm -f main*.o
```

and we have a simple `main.c`, `main1.c` and `main2.c` with `int main() { return 0; }` as content. Now here is me using Make:

```console
georg@swift /tmp  ➜ make
not phony
georg@swift /tmp  ➜ make
make: 'main.o' is up to date.
georg@swift /tmp  ➜ make main1.o
explicit phony
georg@swift /tmp  ➜ make main1.o
explicit phony
georg@swift /tmp  ➜ make main2.o
implicit phony
georg@swift /tmp  ➜ make main2.o
implicit phony
```

and you can see the advantages and disadvantages of `ALWAYS_RUN`. You have it as a dependency, and you can therefore not really use `$^` or `$+` anymore.

georglauterbach: If you ever want to see or use an over-engineered `Makefile` for compiling C/C++ files with Assembly mixed in:

<details>

```make
SHELL        := /bin/bash -eEu -o pipefail -c

SRCDIR       := src
OUTDIR       := build
DEPDIR       := $(OUTDIR)/dependencies
BINARY       := $(OUTDIR)/binary

CPP          ?= cpp
CC           ?= gcc
CXX          ?= g++
AS           ?= as

SRCFILES     := $(shell find $(SRCDIR) -regextype egrep -regex ".*\.(c|cc|S|asm)")
OBJFILES     := $(addsuffix .o, $(patsubst $(SRCDIR)%, $(OUTDIR)%, $(basename $(SRCFILES))))
DEPFILES     := $(patsubst $(OUTDIR)%.o,$(DEPDIR)%.d,$(OBJFILES))

__CFLAGS     := -Wall -Wextra -pedantic -I $(SRCDIR) -g
CFLAGS       := $(CFLAGS)
CXXFLAGS     := $(CFLAGS)
DEPFLAGS      = -MT $@ -MMD -MP -MF $(DEPDIR)/$*.d

# COMMON TARGETS
# ---------------------------------------------------------

.PHONY: force clean

default: serial_build
all: force
force: clean default

clean:
	@ $(RM) -r $(OUTDIR)

serial_build: $(BINARY)

parallel_build:
	@ $(MAKE) -j $$(nproc) $(BINARY)

$(BINARY): $(OBJFILES)
	@ printf 'LD\t$@\n'
	$(CXX) $(CXXFLAGS) -o $@ $^

# CREATE OBJECT FILES
# ---------------------------------------------------------

# build all .o files from source .c files
$(OUTDIR)/%.o: $(SRCDIR)/%.c | $(OUTDIR) $(DEPDIR)
	@ printf 'CC\t%-40s -> $@\n' '$<'
	@ [[ -d $(@D) ]] || mkdir -p $(@D)
	@ [[ -d $(DEPDIR)/$(*D) ]] || mkdir -p $(DEPDIR)/$(*D)
	@ $(CC) $(DEPFLAGS) $(CFLAGS) -c -o $@ $<

# build all .o files from source .cc files
$(OUTDIR)/%.o: $(SRCDIR)/%.cc | $(OUTDIR) $(DEPDIR)
	@ printf 'CXX\t%-40s -> $@\n' '$<'
	@ [[ -d $(@D) ]] || mkdir -p $(@D)
	@ [[ -d $(DEPDIR)/$(*D) ]] || mkdir -p $(DEPDIR)/$(*D)
	@ $(CXX) $(DEPFLAGS) $(CXXFLAGS) -c -o $@ $<

# build all .o files from source .S files
$(OUTDIR)/%.o: $(SRCDIR)/%.S | $(OUTDIR)
	@ printf 'AS\t%-40s -> $@\n' '$<'
	@ [[ -d $(@D) ]] || mkdir -p $(@D)
	@ $(CPP) -o $(OUTDIR)/$*.clean.S $<
	@ $(AS) -o $@ $(OUTDIR)/$*.clean.S

# build all .o files from source .asm files
$(OUTDIR)/%.o: $(SRCDIR)/%.asm | $(OUTDIR)
	@ printf 'NASM\t%-40s -> $@\n' '$<'
	@ [[ -d $(@D) ]] || mkdir -p $(@D)
	@ nasm -w-zeroing -f elf64 -o $@ $<

# MISCELLANEOUS
# ---------------------------------------------------------

$(DEPDIR) $(OUTDIR) $(QEMU_DIR) $(EFI_BOOT_DIR):
	@- mkdir -p $@

$(DEPFILES):
-include $(wildcard $(DEPFILES))
```
</details>

casperklein: > ALWAYS_RUN is just a convenience to not declare every target .PHONY individually

Ack, now I understand. I always did it as one-liner to achieve the same, e.g.:

`.PHONY: all build generate-accounts backup clean tests lint hadolint shellcheck eclint`

polarathene:
> Had to force-push the branch - hope you don't mind.

A little bit as I'll need to look over the branch commit history again, and if we are to rebase merge the commits it'll probably need a rewrite again at some point, so hopefully any necessary changes have been done :sweat_smile: 

I had something else require my time during the weekend, so I'll need to defer until the next one to review this, thanks for the patience :sweat_smile:

georglauterbach:
> > Had to force-push the branch - hope you don't mind.
> 
> 
> 
> A little bit as I'll need to look over the branch commit history again, and if we are to rebase merge the commits it'll probably need a rewrite again at some point, so hopefully any necessary changes have been done :sweat_smile: 

Oh, I think a small misunderstanding here :) I pushed over _my own commits_ but completely kept the history you rewrote :D We could probably squash my latest commits into two or three commits when we rebase merge, but there should be no need for rewriting history to a bit extend.

> I had something else require my time during the weekend, so I'll need to defer until the next one to review this, thanks for the patience 😅 

We got time :)

> BTW, this weekend I did get to look into `docker compose` features a bit with `extends`, fragments, overrides, etc. Ideally tests will move from `docker ...` to `docker compose ...` (`compose` v2) and leverage those features rather than stitching configs with bash (_at least it may be better than my bash helper approach_).

👍

polarathene:
In commit https://github.com/docker-mailserver/docker-mailserver/pull/2857/commits/22d358f6eace9c15ed49653995fc82fcbf4f79b1 you reverted some changes you introduced.

The commit message explains why, but I am curious about some of the reverted content. Noting here in case you'd like to add via a follow-up PR at some point:

---

Is the `docker rm -f "${CONTAINER_NAME}"` still useful to have at the end of these two tests?:

- `@test "${TEST_NAME_PREFIX} repeat_in_container_until_success_or_timeout run command in container"`
- `@test "${TEST_NAME_PREFIX} wait_for_smtp_port_in_container aborts wait after timeout"`

---

`docker exec dms-test-wait_for_finished_setup setup email add 'example_user@domain.com' 'password'` was added before `#but it will finish eventually` in `@test "${TEST_NAME_PREFIX} wait_for_finished_setup_in_container"`. Is it no longer a useful addition?

---

`@test "${TEST_NAME_PREFIX} container_has_service_running/wait_for_service"` likewise had added these three lines before `# pick a service that was not started`:

```bash
docker exec dms-test-services_running setup email add 'example_user@domain.com' 'password'
wait_for_finished_setup_in_container dms-test-services_running
wait_for_changes_to_be_detected_in_container dms-test-services_running
```

And altered the `docker run` ENV to include the following:

```bash
    -e ENABLE_CLAMAV=0 \
    -e ENABLE_AMAVIS=0 \
    -e ENABLE_FAIL2BAN=0 \
    -e ENABLE_SPAMASSASIN=0 \
```

Which I assume was decided as more suited for default test config with overrides where relevant?

test/tests/parallel/set2/spam_bounced.bats
```diff
@@ -0,0 +1,31 @@
+TEST_NAME_PREFIX='spam (Amavis):'
+CONTAINER_NAME='dms-test-spam_bounced'
+
+load "${REPOSITORY_ROOT}/test/helper/setup"
+load "${REPOSITORY_ROOT}/test/helper/common"
+
```
polarathene, line 6:
```bash
# Can run tests in parallel?: No
# Shared static container name: TEST_NAME

# Test case
# ---------
# When SPAMASSASSIN_SPAM_TO_INBOX=0, spam messages must be bounced (rejected).
# SPAMASSASSIN_SPAM_TO_INBOX=1 is covered in `mail_spam_junk_folder.bats`.
# Original test PR: https://github.com/docker-mailserver/docker-mailserver/pull/1485
```

Is the information here not useful anymore?

- Looks like this test file had more than one test at some point, so the first few lines aren't useful.
- It does mention the other variation of the ENV being covered in a different test file, and presumably tracking down the original test contribution warranted adding the PR link for reference :sweat_smile: 

I am fine with dropping it if no other maintainer sees value in it. I think it was intended to provide some summary of tests at the top and perhaps some additional context to ease maintenance / review, especially on larger or more complicated test files.

test/tests/parallel/set2/template.bats
```diff
@@ -0,0 +1,39 @@
...
+
+# ? test setup
+
+function setup_file() {
+  # ? optional setup before container is started
+
+  # ? initialize the test helpers
+  init_with_defaults
+
+  # ?  add custom arguments supplied to `docker run` here
```
polarathene, line 19:
```suggestion
  # ? add custom arguments supplied to `docker run` here
```

test/tests/parallel/set2/template.bats
```diff
@@ -0,0 +1,39 @@
...
+  local CUSTOM_SETUP_ARGUMENTS=(
+    --env LOG_LEVEL=trace
+  )
+
+  # ?   use a helper to correctly setup the container
```
polarathene, line 24:
```suggestion
  # ? use a helper to correctly setup the container
```

test/tests/parallel/set3/dovecot_inet_protocol.bats
```diff
@@ -0,0 +1,47 @@
+TEST_NAME_PREFIX='Dovecot protocols:'
+CONTAINER_NAME='dummy'
+CONTAINER_NAME_ALL='dms-test-dovecot_protocols_all'
+CONTAINER_NAME_IPV4='dms-test-dovecot_protocols_ipv4'
+CONTAINER_NAME_IPV6='dms-test-dovecot_protocols_ipv6'
```
polarathene, line 5: Are the individual `@test` tests running concurrently? Or sequentially?

- The `dummy` usage could probably be avoided.
- `CONTAINER_NAME_*` can use `export` keyword to be available throughout execution of the file AFAIK? If so you could move these into `setup_file()` co-located with the relevant block just before `CONTAINER_NAME`.

test/tests/parallel/set3/dovecot_inet_protocol.bats
```diff
@@ -0,0 +1,47 @@
...
+  CONTAINER_NAME=${CONTAINER_NAME_IPV4}
+  init_with_defaults
+  local CUSTOM_SETUP_ARGUMENTS=(--env DOVECOT_INET_PROTOCOLS=ipv4)
+  common_container_setup 'CUSTOM_SETUP_ARGUMENTS'
+
+  CONTAINER_NAME=${CONTAINER_NAME_IPV6}
+  init_with_defaults
+  local CUSTOM_SETUP_ARGUMENTS=(--env DOVECOT_INET_PROTOCOLS=ipv6)
+  common_container_setup 'CUSTOM_SETUP_ARGUMENTS'
+}
```
polarathene, line 25: Not required for the PR, but `setup_file()` preps setup for one or more tests, but as there is only 1 test for each, technically we could have each `@test` inline the relevant setup block from here, reducing resource requirements and time to first test running.

Alternatively in future it could be split into 3 sub-test files too I suppose that can all be run at once if desired :man_shrugging:

test/helper/setup.bash
```diff
@@ -0,0 +1,96 @@
...
+    "${IMAGE_NAME}"
+
+  assert_success
+}
+
+function common_container_start() {
+  run docker start "${CONTAINER_NAME}"
+  assert_success
+
+  wait_for_finished_setup_in_container "${CONTAINER_NAME}"
```
polarathene, line 95: Just a note that `setup.bash` relies on implicit import externally for `duplicate_config_for_container` and this method to be called from `common.bash`?

Would there value in adding a note about these two external methods `common.bash` dependency at the top of this file?

polarathene (commented):
Finally got around to tackling the review :sweat_smile: 

I am expecting I won't have much time for feedback next weekend, thanks for the patience :pray: I know this PR has a slow feedback cycle from me, but I really appreciate the time and effort you've put into tackling this :muscle: 

---

So far it's looking good :+1: 

Just making notes about some changes I noticed, few minor change suggestions, and some questions about new stuff you've done.

I reviewed iterating through the commit history, the commit messages you added were appreciated :)

georglauterbach:
> In commit [22d358f](https://github.com/docker-mailserver/docker-mailserver/commit/22d358f6eace9c15ed49653995fc82fcbf4f79b1) you reverted some changes you introduced.
> 
> The commit message explains why, but I am curious about some of the reverted content. Noting here in case you'd like to add via a follow-up PR at some point:

The file in question probably needs a dedicated follow-up PR indeed. I completely reverted the changes because I was not sure on how much of an impact my initial changes were and _why_ some of the changes lead to test failures. In short: The questions you stated should indeed be picked up later; I cannot give you a comprehensive answer as I currently lack the answers and a proper explanation myself.

test/helper/setup.bash
```diff
@@ -0,0 +1,96 @@
...
+  __check_if_set 'REPOSITORY_ROOT'
+  __check_if_set 'IMAGE_NAME'
+  __check_if_set 'CONTAINER_NAME'
+
+  TEST_TIMEOUT_IN_SECONDS=${TEST_TIMEOUT_IN_SECONDS:-120}
+  NUMBER_OF_LOG_LINES=${NUMBER_OF_LOG_LINES:-10}
+  SETUP_FILE_MARKER="${BATS_TMPDIR:?}/$(basename "${BATS_TEST_FILENAME:?}").setup_file"
+}
+
+__initialize_variables
```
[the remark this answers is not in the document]
georglauterbach: Please note: this is a completely new directory (`test/helper/` and not `test/test_helper`). I want to keep the old helper until all tests are migrated.

I want a file that just takes care of container initialization. I do not really want this to be mixed up with other helpers as it makes maintenance easy.

test/helper/common.bash
```diff
@@ -0,0 +1,259 @@
...
+
+  if [[ ${STATUS} -eq 1 ]]; then
+    echo "Last ${NUMBER_OF_LOG_LINES} lines of container \`${1}\`'s log"
+    docker logs "${1}" | tail -n "${NUMBER_OF_LOG_LINES}"
+  fi
+
+  return ${STATUS}
+}
+
+SETUP_FILE_MARKER="${BATS_TMPDIR}/$(basename "${BATS_TEST_FILENAME}").setup_file"
```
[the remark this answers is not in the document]
georglauterbach: No :)

test/helper/setup.bash
```diff
@@ -0,0 +1,96 @@
...
+
+  # Common complimentary test files, read-only safe to share across containers:
+  export TEST_FILES_CONTAINER_PATH='/tmp/docker-mailserver-test'
+  export TEST_FILES_VOLUME="${REPOSITORY_ROOT}/test/test-files:${TEST_FILES_CONTAINER_PATH}:ro"
+
+  # The config volume cannot be read-only as some data needs to be written at container startup
+  # - two sed failures (unknown lines)
+  # - dovecot-quotas.cf (setup-stack.sh:_setup_dovecot_quotas)
+  # - postfix-aliases.cf (setup-stack.sh:_setup_postfix_aliases)
+  export TEST_CONFIG_VOLUME="${TEST_TMP_CONFIG}:/tmp/docker-mailserver"
```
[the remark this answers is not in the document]
georglauterbach: Will revert

test/helper/setup.bash
```diff
@@ -0,0 +1,96 @@
...
+  SETUP_FILE_MARKER="${BATS_TMPDIR:?}/$(basename "${BATS_TEST_FILENAME:?}").setup_file"
+}
+
+__initialize_variables
+
+# -------------------------------------------------------------------
+
+# Common defaults appropriate for most tests, override vars in each test when necessary.
+# For all tests override in `setup_file()` via an `export` var.
+# For individual test override the var via `local` var instead.
```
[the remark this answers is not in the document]
georglauterbach: Ah, there was a misunderstanding from my side here. I will re-add the comment.

test/helper/setup.bash
```diff
@@ -0,0 +1,96 @@
...
+function init_with_defaults() {
+  export TEST_TMP_CONFIG
+
+  # In `setup_file()` the default name to use for the currently tested docker container
+  # is `${CONTAINER_NAME}` global defined here. It derives the name from the test filename:
+  # `basename` to ignore absolute dir path and file extension, only extract filename.
+  # In `setup_file()` creates a single copy of the test config folder to use for an entire test file:
+  TEST_TMP_CONFIG=$(duplicate_config_for_container . "${CONTAINER_NAME}")
```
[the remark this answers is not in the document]
georglauterbach: `CONTAINER_NAME` is, in the new tests, a required variable. If it is not set, the test wont continue. Hence, there needn't be a default. Your `EDIT:` is correct.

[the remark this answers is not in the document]
georglauterbach: > **EDIT:** Seems to be for `__initialize_variables` check?

Indeed

test/tests/parallel/set2/template.bats
georglauterbach: 👍🏼

test/tests/parallel/set2/template.bats
georglauterbach: 👍🏼

test/tests/parallel/set3/dovecot_inet_protocol.bats
georglauterbach: Inside a test file, test run sequentially, IIRC. The `dummy` can be avoided, right. I would like to avoid using `export` in `setup_file` as I want to have all variables that need to be set for a file to be set in the very beginning.

test/tests/parallel/set3/dovecot_inet_protocol.bats
georglauterbach: Agreed, should be in another PR. Not sure what to do now though: resolve or leave unresolved?

test/helper/setup.bash
georglauterbach: I will see what a good solution is. Maybe even put these methods into `setup.bash`.

test/helper/setup.bash
```diff
@@ -0,0 +1,96 @@
...
+  export TEST_FILES_VOLUME="${REPOSITORY_ROOT}/test/test-files:${TEST_FILES_CONTAINER_PATH}:ro"
+
+  # The config volume cannot be read-only as some data needs to be written at container startup
+  # - two sed failures (unknown lines)
+  # - dovecot-quotas.cf (setup-stack.sh:_setup_dovecot_quotas)
+  # - postfix-aliases.cf (setup-stack.sh:_setup_postfix_aliases)
+  export TEST_CONFIG_VOLUME="${TEST_TMP_CONFIG}:/tmp/docker-mailserver"
+
+  # The common default FQDN assigned to the container `--hostname` option:
+  export TEST_FQDN="${TEST_FQDN:-mail.my-domain.com}"
```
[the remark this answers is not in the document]
georglauterbach: I see your point. I will have a look.

test/helper/setup.bash
```diff
@@ -0,0 +1,96 @@
...
+#
+# `X_EXTRA_ARGS` - Optional: Pass an array by it's variable name as a string, it will
+# be used as a reference for appending extra config into the `docker create` below:
+#
+# NOTE: Using array reference for a single input parameter, as this method is still
+# under development while adapting tests to it and requirements it must serve (eg: support base config matrix in CI)
+function common_container_create() {
+  [[ -n ${1} ]] && local -n X_EXTRA_ARGS=${1}
+
+  run docker create --tty \
```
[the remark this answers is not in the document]
georglauterbach: No inherent reason. I can move it to the next line too.

test/helper/setup.bash
```diff
@@ -0,0 +1,96 @@
...
+    --name "${CONTAINER_NAME}" \
+    --hostname "${TEST_FQDN}" \
+    --volume "${TEST_FILES_VOLUME}" \
+    --volume "${TEST_CONFIG_VOLUME}" \
+    --env ENABLE_AMAVIS=0 \
+    --env ENABLE_CLAMAV=0 \
+    --env ENABLE_UPDATE_CHECK=0 \
+    --env ENABLE_SPAMASSASSIN=0 \
+    --env ENABLE_FAIL2BAN=0 \
+    --env LOG_LEVEL=debug \
```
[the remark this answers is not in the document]
georglauterbach: I set `ENABLE_AMAVIS=0` because I think it is just faster not to enable Amavis for tests where it is not required, i.e. the startup / receiving of mails is quicker.

docs/content/contributing/general.md
```diff
@@ -24,6 +24,10 @@ To run the test suite, you will need to
 
     We do not support running linting, tests, etc on macOS at this time. Please use a linux VM.
 
+??? tip "Running a Specific Test"
+
+    If you want to run a specific test, you can use `make build generate-accounts test/<TEST NAME>`, where `<TEST NAME>` can be just the name of the test or the path from `test/test/<PATH>` without the `.bats` suffix. Say you want to run the `template.bats` tests: you can use `make test/template` or `make test/parallel/set2/template`.
```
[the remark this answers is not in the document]
georglauterbach: 👍🏼

test/helper/setup.bash
```diff
@@ -0,0 +1,96 @@
...
+    if [[ ${!1+set} != 'set' ]]
+    then
+      echo "ERROR: (helper/setup.sh) '${1:?No variable name given to __check_if_set}' is not set" >&2
+      exit 1
+    fi
+  }
+
+  __check_if_set 'REPOSITORY_ROOT'
+  __check_if_set 'IMAGE_NAME'
+  __check_if_set 'CONTAINER_NAME'
```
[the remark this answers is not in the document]
georglauterbach: Yes, these variables are required to be set for new tests to run. This enables tests to be sure that these are set, and as a consequence, tests can use them without worrying that they are not set. There is not much effort, since Make setes two of the three values. You will only need to provide `CONTAINER_NAME`, which I believe to be completely fine.

You're right about the dummy value earlier.

test/tests/parallel/set1/clamav.bats
```diff
@@ -0,0 +1,69 @@
...
+load "${REPOSITORY_ROOT}/test/helper/setup"
+load "${REPOSITORY_ROOT}/test/helper/common"
+
+function setup_file() {
+  init_with_defaults
+
+  local CUSTOM_SETUP_ARGUMENTS=(
+    --env ENABLE_CLAMAV=1
+    --env ENABLE_AMAVIS=1
+    --env PERMIT_DOCKER=host
```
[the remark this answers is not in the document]
georglauterbach: I think there was a comment about this earlier: https://github.com/docker-mailserver/docker-mailserver/pull/2815/files#r991087509

test/tests/parallel/set2/spam_bounced.bats
georglauterbach: I can re-add this comment :)

georglauterbach:
@polarathene for some reason I cannot directly answer https://github.com/docker-mailserver/docker-mailserver/pull/2857#discussion_r1020831427, so here it is:

`generate-accounts` is run 4 times, because I observed some test files alter the accounts files which leads to other tests failing for no apparent reason. There is some sort of dependency here on the order test files that are run. This is not optimal, but currently solved by running the `generate-accounts` target before every new test set invocation.

test/tests/parallel/set3/dovecot_inet_protocol.bats
polarathene: > I would like to avoid using `export` in `setup_file` as I want to have all variables that need to be set for a file to be set in the very beginning.

I prefer co-location myself, especially since it avoids the `dummy` value in this case which would make any additional tests pass the check for `CONTAINER_NAME` in `setup.bash` if the `dummy` usage became common place for scenarios like this one.

`setup_file()` is fairly close to setting at the beginning :sweat_smile: 

Not going to block on this though :+1:

test/tests/parallel/set3/dovecot_inet_protocol.bats
polarathene: Can leave as-is, and this feedback can remain unresolved so it's easier to notice this needs to be resolved in a follow-up PR? (_although @casperklein isn't fond of those AFAIK, so if he'd prefer we tackle it now that's fine_)

polarathene:
> In short: The questions you stated should indeed be picked up later; I cannot give you a comprehensive answer as I currently lack the answers and a proper explanation myself.

No worries, that can be looked into after this is merged then :+1:

test/helper/setup.bash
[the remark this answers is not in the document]
polarathene: Yes, I'm aware of the intention and support that. As well as the transitional stage where the original `test_helper` is to remain (_ideally not for an extended period of time to avoid maintenance concerns_).

With the new changes you've made this only leaves `TEST_TIMEOUT_IN_SECONDS` and `NUMBER_OF_LOG_LINES` as vars covered in `common.bash`, not as big of a concern for me though, so this is fine now :+1:

polarathene (dismissed):
LGTM, might be nice to add some contextual maintenance comments, but otherwise I'm happy with current progress and we can revise the commit history once another approval comes in.

- ~~Regarding https://github.com/docker-mailserver/docker-mailserver/pull/2857#discussion_r1020832478 do you think it would be worth adding a maintainer comment referencing https://github.com/docker-mailserver/docker-mailserver/pull/2815/files#r991087509 into the test file for context?~~ (**DONE**)
- ~~Likewise for the `Makefile` query with 4 `generate-accounts`, it may be worthwhile adding a maintainer comment linking back to [this explanation](https://github.com/docker-mailserver/docker-mailserver/pull/2857#issuecomment-1312724303) as it seems like something that warrants investigation and a proper fix if someone finds the time.~~ (**DONE**)
- Unless another reviewer disagrees, ~~https://github.com/docker-mailserver/docker-mailserver/pull/2857#discussion_r1020952150~~ (**DONE**) and https://github.com/docker-mailserver/docker-mailserver/pull/2857#issuecomment-1312716198 (_reverted commit_) can be deferred to a follow-up PR.

test/tests/parallel/set3/dovecot_inet_protocol.bats
georglauterbach: I agree, setting it in `setup_file` would be okay as well, but sourcing `setup.bash` alone is enough for the check, and since we run `setup_file` after we source `setup.bash`, I think there is little room to put the ENV other than at the beginning of the file.

test/helper/setup.bash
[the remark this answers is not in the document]
georglauterbach: I must have missed them. I can move them to `setup.bash` too :)

test/tests/parallel/set3/dovecot_inet_protocol.bats
georglauterbach: You're right, @casperklein does indeed not like it. I decided to put it into this PR and inlined container creation.

test/helper/setup.bash
```diff
@@ -0,0 +1,133 @@
...
+    then
+      echo "ERROR: (helper/setup.sh) '${1:?No variable name given to __check_if_set}' is not set" >&2
+      exit 1
+    fi
+  }
+
+  local REQUIRED_VARIABLES_FOR_TESTS=(
+    'REPOSITORY_ROOT'
+    'IMAGE_NAME'
+    'CONTAINER_NAME'
```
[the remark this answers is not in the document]
georglauterbach: I like that! I revised the setup procedure a bit and will push a final commit that incorporates some of the proposals from your latest comments as well.

georglauterbach:
@casperklein @polarathene I pushed the commits addressing the latest PR feedback. Hope all as well now - happy final reviewing :D

polarathene (approved):
Awesome work thanks! :partying_face:

casperklein (dismissed):

polarathene (changes requested):
I'll handle the interactive rebase before we merge this when I have some time spare :+1:

polarathene:
Apologies for the delay, finally got around to tackling this! :grinning: 

- Refactored commit history from 24 commits down to 6.
- Commit messages rewritten accordingly.
- There is no difference in output to review.

Upon approval, merge via merge commit or rebase, **not squash**.

---

<details>
<summary>Commit messages:</summary>

### tests(chore): Rename test files to serial and parallel types
- `test_helper.bats` needs more work than this PR provides to be compatible with parallel tests, so must remain as a serial test for now.
- `spam_bounced.bats` had failures as a serial test, but works well converted to a parallel test in a future commit.


### tests(CI): Adjust Makefile & GHA workflow to support new test layout
These updates support running tests that have been relocated into `serial` and `parallel/set*` directories.

- `make tests` now calls the two make targets beneath it. The only difference is that `serial` continues the "1 test at a time" approach used prior to this PR, while the `parallel` target increases the `--jobs` arg to run multiple tests concurrently (_configured by `PARALLEL_JOBS`_).
- The `test/%` target leverages Bash syntax magic to ease running single tests without providing the exact path.
- This syntax also supports providing multiple test names (eg: `make test/clamav,template`) to run.
- `**` (globstar) allows for future improvements that can group multiple test files into sub-directories by their scope (eg: anti-spam, ssl, etc).

---

chore: Add `shopt -s globstar` to other targets
I realized that other targets should have this as well in case it is not set.
It is better to be more explicit here than to have weird errors due to `**` not expanding properly.

---

fix(Makefile):  Add back `.PHONY` targets

I encountered `make` telling me the target was already up-to-date, which of course is nonsense.
I therefore added back the `.PHONY` targets to ensure tests are always run. 

---

docs: Added instructions for running  a single test
See https://github.com/docker-mailserver/docker-mailserver/pull/2857/files#r1008582760


### tests(chore): Use `REPOSITORY_ROOT` export var from Makefile

Allows for using `load` with an absolute path instead of a relative one, which makes it possible to group tests into different directories.

Parallel tests differ slightly, loading the newer `helper/common.bash` and `helper/setup.bash` files instead of the older `test_helper/common.bash` which serial tests continue to use.

### tests(refactor): `common.bash` helper split into two files
The current `test/test_helper/common.bash` was getting large. Setup logic has been extracted out into a new file.

`common.bash` resides in a directory  named `test_helper/`, the `test_` prefix is redundant. 

As an interim solution this provides a new approach for the updated tests, while the "old" tests can use the "old" `common.bash`. Eventually all tests should migrate to the new approach in `helper/` instead of the older `test_helper/`.

The new helper files are located under `test/helper/` (_which drops the `test_` prefix_). The new and updated helpers apply the new naming convention for ENV variables (_such as `CONTAINER_NAME` or `IMAGE_NAME`_).

---

Some refactoring occurred, including new methods like `_run_in_container()` and `_default_teardown()`.

---

I encountered a situation before in which the updated tests would fail because there were collisions of ENV names in the tests (_for example with `CONTAINER_NAME`_).

### tests(refactor): Conversion to parallel tests and use revised helpers
- Introduced `CONTAINER_NAME` and `TEST_NAME_PREFIX` as new vars for better managing test consistency (DRY).
- `CONTAINER_NAME` replaces any repeated container name with the variable. The value will differ slightly as the prior prefix (`mail_`) has been changed to `dms-test-`.
- `TEST_NAME_PREFIX` provides a prefix value for each `@test` description string.

---

chore: Add a reference template for tests

</details>

polarathene (approved):
:partying_face:

casperklein (approved):

PR #2881, opened 2022-11-01
Bugfix: './setup.sh email list' does not display aliases correctly

### LINKED ISSUE #2877: [BUG] './setup.sh email list' does not display aliases correctly
For reference: https://github.com/docker-mailserver/docker-mailserver/discussions/2875

### Affected Component(s)

./setup.sh email list

On a fresh setup:
```bash
./setup.sh email add media@example.com
./setup.sh email add socialmedia@example.com

./setup.sh alias add facebook@example.com socialmedia@example.com

./setup.sh email list
* media@example.com
    [ aliases -> facebook@example.com ]

* socialmedia@example.com
    [ aliases -> facebook@example.com ]
```

```Markdown
facebook@example.com should only be listed for the socialmedia@example.com mail account.
```

### DMS version

v11.2.0

Linux

Debian 11

x86_64 / AMD64

Docker

### Code of conduct

casperklein:
# Description

Details see [here](https://github.com/docker-mailserver/docker-mailserver/issues/2877).

Before:

```bash
./setup.sh email list
* media@example.com
    [ aliases -> facebook@example.com ]

* socialmedia@example.com
    [ aliases -> facebook@example.com ]
```

After:

```bash
./setup.sh email list
* media@example.com

* socialmedia@example.com
    [ aliases -> facebook@example.com ]
```

Fixes #2877

## Type of change

## Checklist:

### DIFF
--- a/target/bin/listmailuser
+++ b/target/bin/listmailuser
@@ -101,7 +101,7 @@ function _alias_list_for_account
     --ignore-case
     --extended-regexp
     -e "\s${MAIL_ACCOUNT}($|,)" # match first and second sample line
-    -e ",${MAIL_ACCOUNT}($|,)"  # match third and fourth sample line
+    -e  ",${MAIL_ACCOUNT}($|,)" # match third and fourth sample line
     "${DATABASE_VIRTUAL}"
   )
 
--- a/target/bin/listmailuser
+++ b/target/bin/listmailuser
@@ -87,6 +87,7 @@ function _bytes_to_human_readable_size
 # Returns a comma delimited list of aliases associated to a recipient (ideally the recipient is a mail account):
 function _alias_list_for_account
 {
+  local GREP_OPTIONS
   local MAIL_ACCOUNT=${1}
 
 # postfix-virtual.cf sample lines:
--- a/target/bin/listmailuser
+++ b/target/bin/listmailuser
@@ -90,15 +90,9 @@ function _alias_list_for_account
   local MAIL_ACCOUNT=${1}
 
   # `__db_list_already_contains_value` would be a more reliable check:
-  function _account_has_an_alias
-  {
-    local ANY_ALIAS='\S\+\s'
-    grep -qis "^${ANY_ALIAS}.*${MAIL_ACCOUNT}" "${DATABASE_VIRTUAL}"
-  }
-
-  if _account_has_an_alias
+  if grep -qis "\s${MAIL_ACCOUNT}$" "${DATABASE_VIRTUAL}"
   then
-    grep "${MAIL_ACCOUNT}" "${DATABASE_VIRTUAL}" | awk '{print $1;}' | sed ':a;N;$!ba;s/\n/, /g'
+    grep "\s${MAIL_ACCOUNT}$" "${DATABASE_VIRTUAL}" | awk '{print $1;}' | sed ':a;N;$!ba;s/\n/, /g'
   fi
 }
 
--- a/target/bin/listmailuser
+++ b/target/bin/listmailuser
@@ -100,7 +100,7 @@ function _alias_list_for_account
   GREP_OPTIONS=(
     --ignore-case
     --extended-regexp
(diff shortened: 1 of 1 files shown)


georglauterbach (dismissed):
LGTM 👍

target/bin/listmailuser
```diff
@@ -90,15 +90,9 @@ function _alias_list_for_account
   local MAIL_ACCOUNT=${1}
 
   # `__db_list_already_contains_value` would be a more reliable check:
-  function _account_has_an_alias
-  {
-    local ANY_ALIAS='\S\+\s'
-    grep -qis "^${ANY_ALIAS}.*${MAIL_ACCOUNT}" "${DATABASE_VIRTUAL}"
-  }
```
polarathene, line 97: This check you've removed ensured that the alias is defined at the start of the line, and that the target mail account comes after some white-space (`ANY_ALIAS`), as there may be more than one account expanded from the alias.

Your change insists that the mail account is at the end of the line. This is not always the case when the alias has more than one account (_or other valid values such as another alias_) to expand to. Your fix introduces a new bug :(

The proper fix would be to additionally ensure `MAIL_ACCOUNT` has white-space surrounding it as well (unless it's the last value on the line.

The comment right above this method (_that you retained, despite removing the method_) points out a more thorough check, in the event this kind of problem happened :sweat_smile: 

https://github.com/docker-mailserver/docker-mailserver/blob/5f671a4c0b0e41e82cc6432d011a342b9e881094/target/scripts/helpers/database/db.sh#L130-L140

Which notes it's an internal method, so will either need refactoring or the easier way is to maintain similar code instead of extracting out a common utility that can be shared elsewhere?

https://github.com/docker-mailserver/docker-mailserver/blob/5f671a4c0b0e41e82cc6432d011a342b9e881094/target/scripts/helpers/database/db.sh#L38-L66

---

That said, this logic was still fine AFAIK, it only checked if the alias existed on a line for a given account in our alias DB / config file.

So perhaps revert the change and keep the condition? Just fix the issue with the actual filtering logic below?

target/bin/listmailuser
```diff
@@ -90,15 +90,9 @@ function _alias_list_for_account
...
-
-  if _account_has_an_alias
+  if grep -qis "\s${MAIL_ACCOUNT}$" "${DATABASE_VIRTUAL}"
   then
-    grep "${MAIL_ACCOUNT}" "${DATABASE_VIRTUAL}" | awk '{print $1;}' | sed ':a;N;$!ba;s/\n/, /g'
+    grep "\s${MAIL_ACCOUNT}$" "${DATABASE_VIRTUAL}" | awk '{print $1;}' | sed ':a;N;$!ba;s/\n/, /g'
```
polarathene, line 95: This should be all you need to add really, the `if` condition can stay as it was.

This still has that earlier problem of not working for an alias that is used for multiple accounts (_eg: `marketing@example.com` / `sales@example.com` aliased to several staff at a company_).

~~For the most part however, `\s${MAIL_ACCOUNT}` should be reliable, and then it should either end with `$` or `\s`~~ (**EDIT:** My bad, the config uses `,` to delimit multiple values for an alias).

---

`grep` pattern should match the account with these bounds: (`\s` || `,`) + `${MAIL_ACCOUNT}` + (`,` || `$`)

polarathene (changes requested):
Exchanges one bug for another.

- Bring back conditional check.
- Improve the pattern for `grep` on `${MAIL_ACCOUNT}` that matches the supported bounds (_prefix / suffix_).

target/bin/listmailuser
casperklein: Should be fixed with https://github.com/docker-mailserver/docker-mailserver/pull/2881/commits/2f4ed0b083a153de5275049c2d4d1e6136983ad1

georglauterbach (approved):

polarathene (approved):
Fantastic improvement thanks! :partying_face:

PR #2890, opened 2022-11-04
fix: adjust DNSBL return code interpretation

### LINKED ISSUE #2889: [BUG] DNSBL error response handling is incorrect

### Affected Component(s)

Mail can no longer be received

```Markdown
Spamhaus recently changed how they deal with AWS and its DNS servers (see https://www.spamhaus.com/resource-center/if-you-query-spamhaus-projects-dnsbls-via-amazon-web-services-dns-move-to-the-free-data-query-service/).

In effect, every DNSBL request made by DMS on our AWS EC2 instance (and I suppose every existing EC2 instance) got the following response:

Oct 30 14:13:29 mail postfix/smtpd[14922]: NOQUEUE: reject: RCPT from mail-lj1-f182.google.com[209.85.208.182]: 554 5.7.1 Service unavailable; Client host [209.85.208.182] blocked using zen.spamhaus.org; Error: 
open resolver; https://www.spamhaus.org/returnc/pub/obfuscated; from=<obfuscated@gmail.com> to=<obfuscated@obfuscated> proto=ESMTP helo=<mail-lj1-f182.google.com>

The response from DNSBL is 127.255.255.254 (see their blog post) with a resulting 554 error code. 
Unfortunately, the result is that every single incoming mail is blocked, which of course was only noticed after a couple of days of missed emails.

According to Spamhaus, DNSBL should be configured in such a way that it deals correctly with error codes (see also https://www.spamhaus.org/news/article/788/spamhaus-dnsbl-return-codes-technical-update). In their words:

"Some users continue to query Spamhaus blocklists via public resolvers, unaware that this means that our data does not actually protect their mail stream. We have introduced an error code for these users to provide a clear signal that there is an issue, and that the mailserver configuration needs to be updated."
```

```Markdown
DMS should handle Spamhaus' "new" error codes correctly and only block mail if the sender is actually on a list, and never block mails because of a configuration error.
```

```Markdown
1. Spin up a AWS EC2
2. Set ENABLE_DNSBL to 1
3. Try to receive any mail
...
```

### DMS version

v11.2.0

Linux

Debian 10 (buster)

x86_64 / AMD64

Docker Compose

### docker-compose.yml

```yml
mail:
    hostname: mail
    domainname: obfuscated.com
    container_name: mail
    ports:
    environment:
      LETSENCRYPT_HOST: mail.obfuscated.com
      LETSENCRYPT_RESTART_CONTAINER: "true"
      VIRTUAL_HOST: mail.obfuscated.com
      SSL_TYPE: manual
      SSL_CERT_PATH: /etc/letsencrypt/mail.obfuscated.com.crt
      SSL_KEY_PATH: /etc/letsencrypt/mail.obfuscated.com.key
    volumes:
      - maildata:/var/mail
      - mailstate:/var/mail-state
      - maillogs:/var/log/mail
      - mail-config:/tmp/docker-mailserver/
      - ssl:/etc/letsencrypt:ro
    restart: unless-stopped
    stop_grace_period: 1m
    env_file:
      - ./mail/.env
      - ./mail/mailserver.env
    cap_add:
      - NET_ADMIN
      - SYS_PTRACE
    logging:
      options:
        max-size: "250m"
        max-file: "3"
```

### Relevant log output

```bash
Oct 30 14:13:29 mail postfix/smtpd[14922]: NOQUEUE: reject: RCPT from mail-lj1-f182.google.com[209.85.208.182]: 554 5.7.1 Service unavailable; Client host [209.85.208.182] blocked using zen.spamhaus.org; Error: 
open resolver; https://www.spamhaus.org/returnc/pub/obfuscated; from=<obfuscated@gmail.com> to=<obfuscated@obfuscated> proto=ESMTP helo=<mail-lj1-f182.google.com>
```

_No response_

### Code of conduct

```Markdown
"What level of experience do you have with Docker and mail servers?" offers only answers in the negative form. I don't mind indicating that I have experience and am comfortable with the CLI, but I do not have that option. You can only guess why I didn't click the negative options if you do not give me the positive options.
```

georglauterbach:
# Description

Adjusts the DNSBL return codes interpretation. Now, only "valid" return codes, i.e. the ones that indicate spam, are taken into account.

Fixes #2889

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/startup/setup-stack.sh
+++ b/target/scripts/startup/setup-stack.sh
@@ -399,7 +399,7 @@ function _setup_postgrey
   _log 'debug' 'Configuring Postgrey'
 
   sed -i -E \
-    's|, reject_rbl_client zen.spamhaus.org$|, reject_rbl_client zen.spamhaus.org, check_policy_service inet:127.0.0.1:10023|' \
+    's|, reject_rbl_client zen.spamhaus.org=127.0.0.[2..11]$|, reject_rbl_client zen.spamhaus.org=127.0.0.[2..11], check_policy_service inet:127.0.0.1:10023|' \
     /etc/postfix/main.cf
 
   sed -i -e \
--- a/target/scripts/startup/setup-stack.sh
+++ b/target/scripts/startup/setup-stack.sh
@@ -1078,7 +1078,7 @@ function _setup_dnsbl_disable
   _log 'debug' 'Disabling postfix DNS block list (zen.spamhaus.org)'
 
   sedfile -i \
-    '/^smtpd_recipient_restrictions = / s/, reject_rbl_client zen.spamhaus.org//' \
+    '/^smtpd_recipient_restrictions = / s/, reject_rbl_client zen.spamhaus.org=127.0.0.\[2..11\]//' \
     /etc/postfix/main.cf
 
   _log 'debug' 'Disabling postscreen DNS block lists'
--- a/target/scripts/startup/setup-stack.sh
+++ b/target/scripts/startup/setup-stack.sh
@@ -399,7 +399,7 @@ function _setup_postgrey
   _log 'debug' 'Configuring Postgrey'
 
   sedfile -i -E \
-    's|(smtpd_recipient_restrictions.*)|\1, check_policy_service inet:127.0.0.1:10023|' \
+    's|(^smtpd_recipient_restrictions =.*)|\1, check_policy_service inet:127.0.0.1:10023|' \
     /etc/postfix/main.cf
 
   sed -i -e \
--- a/target/scripts/startup/setup-stack.sh
+++ b/target/scripts/startup/setup-stack.sh
@@ -398,8 +398,8 @@ function _setup_postgrey
 {
   _log 'debug' 'Configuring Postgrey'
 
-  sed -i -E \
-    's|, reject_rbl_client zen.spamhaus.org=127.0.0.[2..11]$|, reject_rbl_client zen.spamhaus.org=127.0.0.[2..11], check_policy_service inet:127.0.0.1:10023|' \
+  sedfile -i -E \
+    's|(smtpd_recipient_restrictions.*)|\1, check_policy_service inet:127.0.0.1:10023|' \
     /etc/postfix/main.cf
 
   sed -i -e \
--- a/test/mail_with_postgrey.bats
+++ b/test/mail_with_postgrey.bats
@@ -25,7 +25,7 @@ function teardown_file() {
 }
 
 @test "checking postgrey: /etc/postfix/main.cf correctly edited" {
-  run docker exec mail_with_postgrey /bin/bash -c "grep 'zen.spamhaus.org=127.0.0.[2..11], check_policy_service inet:127.0.0.1:10023' /etc/postfix/main.cf | wc -l"
+  run docker exec mail_with_postgrey /bin/bash -c "grep -F 'zen.spamhaus.org=127.0.0.[2..11], check_policy_service inet:127.0.0.1:10023' /etc/postfix/main.cf | wc -l"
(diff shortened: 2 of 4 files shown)


polarathene (dismissed):

target/scripts/startup/setup-stack.sh
```diff
@@ -399,7 +399,7 @@ function _setup_postgrey
   _log 'debug' 'Configuring Postgrey'
 
   sed -i -E \
-    's|, reject_rbl_client zen.spamhaus.org$|, reject_rbl_client zen.spamhaus.org, check_policy_service inet:127.0.0.1:10023|' \
+    's|, reject_rbl_client zen.spamhaus.org=127.0.0.[2..11]$|, reject_rbl_client zen.spamhaus.org=127.0.0.[2..11], check_policy_service inet:127.0.0.1:10023|' \
```
casperklein, line 402: Isn't `[2..11]` interpreted as character class and should be escaped?

test/mail_with_postgrey.bats
```diff
@@ -25,7 +25,7 @@ function teardown_file() {
 }
 
 @test "checking postgrey: /etc/postfix/main.cf correctly edited" {
-  run docker exec mail_with_postgrey /bin/bash -c "grep 'zen.spamhaus.org, check_policy_service inet:127.0.0.1:10023' /etc/postfix/main.cf | wc -l"
+  run docker exec mail_with_postgrey /bin/bash -c "grep 'zen.spamhaus.org=127.0.0.[2..11], check_policy_service inet:127.0.0.1:10023' /etc/postfix/main.cf | wc -l"
```
casperklein, line 28: `grep -F` or escape `[2..11]`.

target/scripts/startup/setup-stack.sh
```diff
@@ -1078,7 +1078,7 @@ function _setup_dnsbl_disable
   _log 'debug' 'Disabling postfix DNS block list (zen.spamhaus.org)'
 
   sedfile -i \
-    '/^smtpd_recipient_restrictions = / s/, reject_rbl_client zen.spamhaus.org//' \
+    '/^smtpd_recipient_restrictions = / s/, reject_rbl_client zen.spamhaus.org=127.0.0.\[2..11\]//' \
```
casperklein, line 1081: Same here.

casperklein (changes requested):

target/scripts/startup/setup-stack.sh
georglauterbach: Probably, yes - thanks :)

target/scripts/startup/setup-stack.sh
georglauterbach: The current diff already shows it escaped (`\[2..11\]`). Is there more to it?

casperklein: My fault.

target/scripts/startup/setup-stack.sh
```diff
@@ -398,8 +398,8 @@ function _setup_postgrey
 {
   _log 'debug' 'Configuring Postgrey'
 
-  sed -i -E \
-    's|, reject_rbl_client zen.spamhaus.org$|, reject_rbl_client zen.spamhaus.org, check_policy_service inet:127.0.0.1:10023|' \
+  sedfile -i -E \
+    's|(smtpd_recipient_restrictions.*)|\1, check_policy_service inet:127.0.0.1:10023|' \
```
casperklein, line 402: I like this. Even more future safe:

```suggestion
    's|(^smtpd_recipient_restrictions =.*)|\1, check_policy_service inet:127.0.0.1:10023|' \
```

casperklein (approved):

polarathene (approved):

PR #2902, opened 2022-11-16
feature: provide initial Rspamd support

georglauterbach:
# Description

This will install the Rspamd (and Redis) package and provide minimal infrastructure to start working with Rspamd. Notably, this is really just bare-minimum infrastructure, but it allows users to use Rspamd as of now. Users will need to adjust Postfix's configuration themselves - a PR that provides ease of use will follow. This is completely backwards compatible.

## Type of change

## Checklist:


docs/content/config/environment.md
```diff
@@ -67,6 +67,19 @@ Note: you probably want to [set `POSTFIX_INET_PROTOCOLS=ipv4`](#postfix_inet_pro
 
 Set the timezone. If this variable is unset, the container runtime will try to detect the time using `/etc/localtime`, which you can alternatively mount into the container. The value of this variable must follow the pattern `AREA/ZONE`, i.e. of you want to use Germany's time zone, use `Europe/Berlin`. You can lookup all available timezones [here](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones#List).
 
+##### ENABLE_RSPAMD
+
+!!! warning "Current State"
+
+    Rspamd-support is under active development. Be aware that breaking changes can happen at any time. Moreover, you will _currently_ need to adjust Postfix's configuration _yourself_ if you want to use Rspamd; you may use [`user-patches.sh`][docs-userpatches].
```
polarathene, line 74: A bit more context with the mention of what needs to be manually done here would be useful. Otherwise I'm not sure how useful the suggestion is? What sort of changes do users wanting to try this feature need to make to Postfix?

target/rspamd/local.d/redis.conf
```diff
@@ -0,0 +1,6 @@
+# documentation: https://rspamd.com/doc/configuration/redis.html
+
+servers = "127.0.0.1:7788";
```
polarathene, line 3: This is the redis port? It differs from supervisor config?

target/scripts/startup/misc-stack.sh
```diff
@@ -33,6 +33,8 @@ function _misc_save_states
     [[ ${ENABLE_FAIL2BAN} -eq 1 ]] && FILES+=('lib/fail2ban')
     [[ ${ENABLE_FETCHMAIL} -eq 1 ]] && FILES+=('lib/fetchmail')
     [[ ${ENABLE_POSTGREY} -eq 1 ]] && FILES+=('lib/postgrey')
+    [[ ${ENABLE_RSPAMD} -ne 1 ]] && FILES+=('lib/rspamd')
+    # [[ ${ENABLE_RSPAMD} -ne 1 ]] && FILES+=('lib/redis')
```
polarathene, line 37: Intentional to leave commented out?

target/scripts/startup/setup-stack.sh
```diff
@@ -89,6 +89,25 @@ function _setup_amavis
...
+  _log 'warn' 'Rspamd support is under active development, expect breaking changes at any time'
+
+  if [[ ${ENABLE_AMAVIS} -eq 1 ]] || [[ ${ENABLE_SPAMASSASSIN} -eq 1 ]]
+  then
+    _shutdown 'You cannot run Amavis/SpamAssassin and Rspamd at the same time'
+  fi
+
+  if [[ ${ENABLE_CLAMAV} -eq 0 ]]
+  then
+    _log 'debug' 'Rspamd will not use ClamAV (which is disabled)'
```
polarathene, line 103:
```suggestion
    _log 'debug' 'Rspamd will not use ClamAV (which has not been enabled)'
```

I think the convention we have with other features (some dovecot ones?), is to have a different extension that is renamed with `mv`. In this case since ClamAV is disabled by default, it should probably be the other way around and if `-eq 1` then log about configuring Rspamd for ClamAV followed by the rename to enable the config?

target/supervisor/conf.d/supervisor-app.conf
```diff
@@ -96,6 +96,24 @@ stdout_logfile=/var/log/supervisor/%(program_name)s.log
...
+command=/usr/bin/rspamd --no-fork --user=rspamd --group=rspamd
+
+[program:redis]
+startsecs=0
+stopwaitsecs=55
+autostart=false
+autorestart=true
+stdout_logfile=/var/log/supervisor/%(program_name)s.log
+stderr_logfile=/var/log/supervisor/%(program_name)s.log
+command=redis-server --daemonize no --bind 127.0.0.1 --port 6379 --loglevel warning
```
polarathene, line 115: I don't use redis, so just noticed that this port differed from the `redis.conf`, is that correct?

polarathene (changes requested):

docs/content/config/environment.md
georglauterbach: I will provide a better explanation.

target/rspamd/local.d/redis.conf
georglauterbach: Ooooops, wrong port :D

target/scripts/startup/setup-stack.sh
georglauterbach: 👍🏼

target/supervisor/conf.d/supervisor-app.conf
georglauterbach: The port number is correct here :) It is the default here.

docs/content/config/environment.md
```diff
@@ -67,6 +67,25 @@ Note: you probably want to [set `POSTFIX_INET_PROTOCOLS=ipv4`](#postfix_inet_pro
...
+
+    You will need to add Rspamd to the `smtpd_milters` in Postfix's `main.cf`. This can easily be done with `sed`: `sed -i -E 's|^(smtpd_milters = .*)|\1,inet:localhost:11332|g' /etc/postfix/main.cf`. Moreover, have a look at the [integration of Rspamd into Postfx](https://rspamd.com/doc/integration.html). You will need to provide additional configuration files at the moment (to `/etc/rspamd/local.d/`) to make Rspamd run in milter-mode.
```
polarathene, line 76: This is better thanks :+1: 

Below is just an alternative revision suggested. Optional to apply.

```suggestion
    Required adjustments ([Integration of Rspamd for Postfix](https://rspamd.com/doc/integration.html)):

    - Configure Postfix to include Rspamd as an smtpd milter (`main.cf:smtpd_milters`):
       `sed`: `sed -i -E 's|^(smtpd_milters = .*)|\1,inet:localhost:11332|g' /etc/postfix/main.cf`
    - You will need to provide additional configuration files at the moment (to `/etc/rspamd/local.d/`) to make Rspamd run in milter-mode.
```

The 2nd step still seems vague for a user to follow. I'm not going to block on that however as this is temporary docs for a feature not officially supported.

polarathene (approved):
LGTM :+1:

target/scripts/startup/misc-stack.sh
polarathene: Fair enough. I'm happy to leave as-is then for the time being :+1:

georglauterbach:
This is now ready to be merged (when @casperklein approves as well of course) ;)

target/scripts/startup/misc-stack.sh
casperklein: Redis is an in-memory data store. I think there is no reason to persist its data.

casperklein (approved):

PR #2903, opened 2022-11-16
update: update to latest version of Fail2Ban

georglauterbach:
# Description

Tests pass locally; we just need to add a "fix". The new Fail2Ban configuration complains about `allowipv6` not being set in the default configuration. Hence, I set it in a custom configuration. There is an additional description of this issue in the file itself. Other than that, all looks good.

This should actually not be a breaking change, hence marked for v11.3.

## Type of change

## Checklist:


wernerfred (approved):

polarathene (approved):

casperklein:
> This should actually not be a breaking change, hence marked for v11.3.

F2B is bumped from `0.11.2` to `1.0.2`. That a major version upgrade. Also the [changelog](https://github.com/fail2ban/fail2ban/blob/1.0.2/ChangeLog) is quiet heavy. I haven't had time to read it completely, did you? Our tests are nice, but I doubt, they cover everything. So we should really sure, it has no drawbacks before shipping it.

PR itself: LGTM.

georglauterbach:
> > This should actually not be a breaking change, hence marked for v11.3.
> 
> F2B is bumped from `0.11.2` to `1.0.2`. That's a major version upgrade. Also the [changelog](https://github.com/fail2ban/fail2ban/blob/1.0.2/ChangeLog) is quiet heavy. I haven't had time to read it completely, did you?

Read through most of it :) Studied the  "Compatibility" section in detail, looked good.

> Our tests are nice, but I doubt, they cover everything. So we should be really sure, it has no drawbacks before shipping it.

We can wait with this and ship it with v12.

PR #2904, opened 2022-11-16
improvement: run `source` instead of `/bin/bash` for `user-patches.sh`

georglauterbach:
# Description

Running `source` will "execute" the file as well, but it allows tinkering with existing code. I have a use case where I want to override a specific function, which I can do with `source`, but I cannot right now, because it is run in a "subshell", where changes are not propagates.

This change changes the current behavior slightly: if you use `set -e`, then this option is set for the whole rest of the script setup of DMS. Tell me whether you think this is a breaking change :)

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/startup/setup-stack.sh
+++ b/target/scripts/startup/setup-stack.sh
@@ -1055,7 +1055,7 @@ function _setup_user_patches
   if [[ -f ${USER_PATCHES} ]]
   then
     _log 'debug' 'Applying user patches'
-    /bin/bash "${USER_PATCHES}"
+    source "${USER_PATCHES}"
   else
     _log 'trace' "No optional '${USER_PATCHES}' provided"
   fi


polarathene (approved):

casperklein:
I am unsure, if I like this change. Some thoughts:

The original intend was to provide a standalone script, that runs in a clean and unpolluted environment (e.g. no predefined variables, functions, options already set, traps etc).

Out of interest: Am I right, that you want to modify anything in `daemons-stack.sh`? That's the only code that follows after the execution of `user-patches.sh`.

---

Pro: ability to modify daemon-stack functions

Con: ability to modify daemon-stack functions 😆 Should we give the average user the ability to mess with our functions (on purpose or by accident?).

If something like the log functions are needed, they can manually be sourced within `user-patches.sh`.

`start-mailserver.sh` sources a lot of files. Any change in any of that files in the future could possibly break a users `user-patches.sh`.

Changing this behaviour (sourcing instead of executing) should only be done, when a major benefit for most users is achieved. I am not sure, if that's the case here.

---

I do also some changes in my DMS setup. That's why I am building DMS myself, giving me the freedom, to modify anything beforehand. Isn't that an option for you?

georglauterbach:
> I am unsure, if I like this change. Some thoughts:
> 
> The original intend was to provide a standalone script, that runs in a clean and unpolluted environment (e.g. no predefined variables, functions, options already set, traps etc).

I understand.

> Out of interest: Am I right, that you want to modify anything in `daemons-stack.sh`? That's the only code that follows after the execution of `user-patches.sh`.

Haha, this is exactly what I am trying to do :D
 
> Pro: ability to modify daemon-stack functions
> 
> Con: ability to modify daemon-stack functions laughing Should we give the average user the ability to mess with our functions (on purpose or by accident?).

Absolutely right. When they use `set -e`, this would now possibly screw with out scripts.
 
> If something like the log functions are needed, they can manually be sourced within `user-patches.sh`.

Indeed, I am currently doing that.

> `start-mailserver.sh` sources a lot of files. Any change in any of that files in the future could possibly break a users `user-patches.sh`.
> 
> Changing this behaviour (sourcing instead of executing) should only be done, when a major benefit for most users is achieved. I am not sure, if that's the case here.

I see. I will provide another PR then that solves my issue in a more idiomatic way.

> I do also some changes in my DMS setup. That's why I am building DMS myself, giving me the freedom, to modify anything beforehand. Isn't that an option for you?

This is not really an option for me :)

PR #2933, opened 2022-12-13
Change OpenDMARC policy default to reject

k3it:
fixes #2932

# Description

Fixes #2932 

## Type of change

- [x] Improvement (non-breaking change that does improve existing functionality)
## Checklist:

- [x] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [ ] If necessary I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #2936, opened 2022-12-18
Revert "Change OpenDMARC policy default to reject"

polarathene:
Reverts docker-mailserver/docker-mailserver#2933

I accidentally merged this before 11.3 release, sorry about that :(

Unless it's considered acceptable for a minor release inclusion, which it might as it seems to be a fix?


casperklein (approved):
I am fine with having this fix already in `v11.3`.

georglauterbach:
I am too; I could have waited though too :D Let's just release as is.

PR #2965, opened 2022-12-27
Fix condition for relay hosts when using only a provided file

### LINKED ISSUE #2964: [BUG] SASL Relay is broken when using only postfix-sasl-password.cf

### Affected Component(s)

Relaying emails

Commit fe21fe78e297785fd5ced455ceed1924eb382db3 might have broken the relay host support when using only ```/tmp/docker-mailserver/postfix-sasl-password.cf```.

The condition in line 65 of ```target/scripts/helpers/relay.sh``` before was:
```
  if [[ ! -f /tmp/docker-mailserver/postfix-sasl-password.cf ]] \
    && [[ -z ${RELAY_USER} || -z ${RELAY_PASSWORD} ]] \
    && [[ -z ${SASL_PASSWD} ]]
```

Now the condition is:
```
  if [[ ! -f /tmp/docker-mailserver/postfix-sasl-password.cf ]] \
    && [[ -z ${RELAY_USER} ]] || [[ -z ${RELAY_PASSWORD} ]]
```

The problem is that this condition seems to be only non-satisfiable if ```RELAY_PASSWORD``` set to a nonempty value - regardless of whether the user only wants to use the data from ```postfix-sasl-password.cf```

It appears that this is a bug as the related tickets do not specify this kind of change.

I would suggest to fix the condition to preserve the backwards compatibility, e.g.:
```
  if [[ ! -f /tmp/docker-mailserver/postfix-sasl-password.cf ]] \
    && [[ -z ${RELAY_USER} || -z ${RELAY_PASSWORD} ]]
```

1. Startup the container with ```RELAY_HOST=any-value``` and ```postfix-sasl-password.cf``` existing
2. "Missing relay-host mapped credentials provided via ENV, or from postfix-sasl-password.cf" is being printed on stdout.

### DMS version

master

Linux

Debian 11

AMD64 / x86_64

Docker

### docker-compose.yml

_No response_

### Relevant log output

_No response_

_No response_

### Code of conduct

_No response_

reneploetz:
Signed-off-by: René Plötz <reneploetz@users.noreply.github.com>

# Description
Fixes #2964

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/helpers/relay.sh
+++ b/target/scripts/helpers/relay.sh
@@ -63,7 +63,7 @@ function _env_relay_host
 function _relayhost_sasl
 {
   if [[ ! -f /tmp/docker-mailserver/postfix-sasl-password.cf ]] \
-    && [[ -z ${RELAY_USER} ]] || [[ -z ${RELAY_PASSWORD} ]]
+    && [[ -z ${RELAY_USER} || -z ${RELAY_PASSWORD} ]]
   then
     _log 'warn' "Missing relay-host mapped credentials provided via ENV, or from postfix-sasl-password.cf"
     return 1


georglauterbach:
Very interesting; now I see. I had this changed in the original PR, but now I see that there is precedence involved. Thanks for the fix!

georglauterbach (approved):

polarathene (approved):

PR #2988, opened 2023-01-08
update BATS & helper

georglauterbach:
# Description

BATS core:
  previous: 1.7.0
  now:      1.8.2

BATS assert:
  previous: ffe84ea5dd43b568851549b3e241db150c12929c
  now:      2.1.0

## Type of change

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,6 @@ export NAME            ?= $(IMAGE_NAME)
 MAKEFLAGS              += --no-print-directory
 BATS_FLAGS             ?= --timing
 BATS_PARALLEL_JOBS     ?= 2
-BATS_FLAGS_PARALLEL    ?= $(BATS_FLAGS) --no-parallelize-within-files --jobs $(BATS_PARALLEL_JOBS)
 
 .PHONY: ALWAYS_RUN
 
@@ -53,7 +52,8 @@ tests/serial: ALWAYS_RUN
 	@ shopt -s globstar ; ./test/bats/bin/bats $(BATS_FLAGS) test/$@/*.bats
 
 tests/parallel/set%: ALWAYS_RUN
-	@ shopt -s globstar ; ./test/bats/bin/bats $(BATS_FLAGS_PARALLEL) test/$@/**/*.bats
+	@ shopt -s globstar ; ./test/bats/bin/bats $(BATS_FLAGS) \
+		--no-parallelize-within-files --jobs $(BATS_PARALLEL_JOBS) test/$@/**/*.bats
 
 test/%: ALWAYS_RUN
 	@ shopt -s globstar nullglob ; ./test/bats/bin/bats $(BATS_FLAGS) test/tests/**/{$*,}.bats


casperklein (approved):

Makefile
```diff
@@ -53,7 +52,8 @@ tests/serial: ALWAYS_RUN
 	@ shopt -s globstar ; ./test/bats/bin/bats $(BATS_FLAGS) test/$@/*.bats
 
 tests/parallel/set%: ALWAYS_RUN
-	@ shopt -s globstar ; ./test/bats/bin/bats $(BATS_FLAGS_PARALLEL) test/$@/**/*.bats
+	@ shopt -s globstar ; ./test/bats/bin/bats $(BATS_FLAGS) \
+		--no-parallelize-within-files --jobs $(BATS_PARALLEL_JOBS) test/$@/**/*.bats
```
polarathene, line 56: Prefer vertically listing the explicit options, easier to grok and track changes :sweat_smile: 

```suggestion
	@ shopt -s globstar ; ./test/bats/bin/bats $(BATS_FLAGS) \
		--no-parallelize-within-files \
		--jobs $(BATS_PARALLEL_JOBS) \
		test/$@/**/*.bats
```

---

A reminder that `--no-parallelize-within-files` prevents `--jobs 1` for running a parallel set sequentially.

Although I wouldn't worry about this much as I'm doubtful the default of `--jobs 2` will cause anyone running these tests a problem, since we can still run a few individual tests with `make test/file1,file2,file3` easily enough.

---

If of interest to you:
- There is also a [tagging feature](https://bats-core.readthedocs.io/en/v1.8.2/writing-tests.html#tagging-tests) to be selective on what test cases to run in a file(s).
- `--no-parallelize-within-files` could be considered for removal if we can confidently evaluate which test files need to explicitly opt-out via [`export BATS_NO_PARALLELIZE_WITHIN_FILE=true` in `setup_file()`](https://bats-core.readthedocs.io/en/v1.8.2/usage.html#parallel-execution).

polarathene (approved):
Oh I somehow missed the assert lib finally getting a tagged release (_I had been subscribed to the issue closed in Oct 2022_) we could use instead of a master branch commit, thanks for sorting this out :tada:

test/tests/parallel/set1/spam_virus/amavis.bats
```diff
@@ -18,8 +18,10 @@ function setup_file() {
...
 function teardown_file() { _default_teardown ; }
 
-@test "${TEST_NAME_PREFIX} Amavis integration should be active" {
-  _run_in_container grep 'ANTI-SPAM-SA' /var/log/mail/mail.log
+@test "SpamAssassin integration should be active" {
+  # give Amavis just a bit of time to print out its full debug log
+  run repeat_in_container_until_success_or_timeout 5 "${CONTAINER_NAME}" grep 'ANTI-SPAM-SA' /var/log/mail/mail.log
+  assert_success
   assert_output --partial 'loaded'
   refute_output --partial 'NOT loaded'
```
polarathene, line 26: This is works too :+1: (_I noticed this locally and was going to use an existing helper method to wait for the amavis service to be ready_)

casperklein: Alternatively, instead of "grep-ing until success", grep until a string appears:

```bash
tail -F /var/log/mail/mail.log | grep --max-count 1 'ANTI-SPAM-SA'

timeout 5 tail -F /var/log/mail/mail.log | grep --max-count 1 'ANTI-SPAM-SA'
```

Makefile
georglauterbach: I have merged this to make CI pass again, but we can tackle this in another PR - I don't really mind since it's only a stylistic change :)

test/tests/parallel/set1/spam_virus/amavis.bats
polarathene: > grep until a string appears

I have been refactoring a test where a similar requirement with counting was necessary and being a hassle for me to shell-fu my way through, I may want to switch to your tip here now :joy:

PR #2994, opened 2023-01-11
tests(refactor): Adjust `mail_tls_dhparams.bats`

polarathene:
# Description

- Refactored this test to the new test structure. Logic remains the same, just easier to grok.
- `ONE_DIR` handling was [removed for this feature in Sep 2021](https://github.com/docker-mailserver/docker-mailserver/pull/2192), dropping those redundant test cases.

## Checklist:

### DIFF
--- a/test/tests/serial/mail_tls_dhparams.bats
+++ b/test/tests/serial/mail_tls_dhparams.bats
@@ -70,7 +70,7 @@ function _should_match_service_copies() {
 }
 
 function _should_emit_warning() {
-  run grep '[ WARNING ]' <<< $(docker logs "${CONTAINER_NAME}")
+  run docker logs "${CONTAINER_NAME}"
   assert_success
-  assert_output --partial 'Using self-generated dhparams is considered insecure - unless you know what you are doing, please remove'
+  assert_output --partial '[ WARNING ]  Using self-generated dhparams is considered insecure - unless you know what you are doing, please remove'
 }
--- a/test/tests/serial/mail_tls_dhparams.bats
+++ b/test/tests/serial/mail_tls_dhparams.bats
@@ -14,8 +14,8 @@ CONTAINER2_NAME='dms-test_tls-dh-params_custom'
 
 function teardown() { _default_teardown ; }
 
-# Verify that the file `ffdhe4096.pem` has not been modified (checksum verification against trusted third-party copy).
 # Verify Postfix and Dovecot are using the default `ffdhe4096.pem` from Dockerfile build.
+# Verify that the file `ffdhe4096.pem` has not been modified (checksum verification against trusted third-party copy).
 @test "Default" {
   export CONTAINER_NAME=${CONTAINER1_NAME}
   local DH_PARAMS_DEFAULT='target/shared/ffdhe4096.pem'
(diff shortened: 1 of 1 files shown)


test/tests/serial/mail_tls_dhparams.bats
```diff
@@ -1,127 +1,76 @@
...
+  __should_have_expected_checksum '/etc/dovecot/dh.pem'
+  __should_have_expected_checksum '/etc/postfix/dhparams.pem'
 }
 
-function should_emit_warning() {
-  run sh -c "docker logs mail_dhparams | grep 'Using self-generated dhparams is considered insecure.'"
+function _should_emit_warning() {
+  run grep '[ WARNING ]' <<< $(docker logs "${CONTAINER_NAME}")
   assert_success
+  assert_output --partial 'Using self-generated dhparams is considered insecure - unless you know what you are doing, please remove'
```
casperklein, line 75: Not tested, but doesn't this work?

```suggestion
  run docker logs "${CONTAINER_NAME}"
  assert_success
  assert_output --partial '[ WARNING ]  Using self-generated dhparams is considered insecure - unless you know what you are doing, please remove'
```

polarathene: Yep, works well thanks for catching that! :sweat_smile:

test/tests/serial/mail_tls_dhparams.bats
```diff
@@ -1,127 +1,76 @@
...
+  __should_have_expected_checksum '/etc/postfix/dhparams.pem'
 }
 
-function should_emit_warning() {
-  run sh -c "docker logs mail_dhparams | grep 'Using self-generated dhparams is considered insecure.'"
+function _should_emit_warning() {
+  run docker logs "${CONTAINER_NAME}"
   assert_success
+  assert_output --partial '[ WARNING ]  Using self-generated dhparams is considered insecure - unless you know what you are doing, please remove'
 }
```
georglauterbach, line 76: Useless function? This is called just once. Is it intended to be used in the future as well? If not, please move the inside of this function into the test.

georglauterbach (changes requested):
Just the one comment; otherwise LGTM 👍🏼

test/tests/serial/mail_tls_dhparams.bats
polarathene: I tend to do this for tests sometimes so the test case itself is simpler to grok :sweat_smile: 

I'll make this change, but then I'm wondering if I should do the same for the mozilla copy method? Personally less noise in the test case is nicer. I don't see any harm with the approach, but happy to adjust :+1: (_I realize this one at least is pretty self-explanatory that it's a tad silly being extracted out to a function_ :sweat_smile: )

georglauterbach: I'd definitely go for it with `_should_emit_warning`, but probably also for `_should_match_mozilla_copy`. I'd leave `_should_match_service_copies` as is though :)

georglauterbach (approved):
LGTM 👍🏼

casperklein (approved):

PR #2997, opened 2023-01-11
tests(refactor): Adjust `mail_changedetector` + change detection helpers

polarathene:
# Description

- Refactored `mail_changedetector` to the new test conventions. Logic remains roughly the same, just easier to grok.
- `tls_letsencrypt` test had some improvements as a result of this.
- `setup-cli` test used the original setup helper methods, no change to logic there, just fully adapted to `helper/setup.bash` and adjusted the test case prefix.
- Change detection helper was refactored. A generic log watching method is used, with support for waiting on an expected count.
   - This was necessary as the `mail_changedetector` test as two concurrent change events happening, where one may have already processed before you can wait on it. Another test case cannot wait on completion, but instead needs to wait on a change detection starting.
- Change detection helpers now live in a separate file. Not sure how useful that is. All tests that use it are migrated to the `helper/common.bash` methods. This also allowed for dropping the helpers from `test_helper/common.bash`, as well as the related methods in that file that now exist in `helper/setup.bash`.
- We no longer use the [previous change detection helper method](https://github.com/docker-mailserver/docker-mailserver/commit/c6c6fb708cc33c93c99c130f76ca066b70429069) (Oct 2020 for `tests.bats`). Thus I've removed that method and the related `test_helper.bats` cases for it. Existing methods should have fairly decent coverage in `mail_changedetector` using it against the actual `changedetector` service, and not just watching for a checksum change separately (_which is not that useful for us_).

---

This refactor removes the need to depend on `sleep`, shaving off approx 2 mins of time (_potentially a bit less until some additional PRs are merged_).

I've staged out changes across commits (with associated commit messages) that should make the review of diffs a bit more easier if it helps :sweat_smile: 

The [original `mail_changedetector` test was implemented in Sep 2021](https://github.com/docker-mailserver/docker-mailserver/pull/2183). It was added with a focus on testing the locking script with multiple containers sharing a network volume for config (`/tmp/docker-mailserver`).

---

I have taken extra care with the changes done here (_over several days_). I ran repeat test runs, along with some other changes I'm pushing as separate PRs to ensure that the change detection tests remain consistent / reliable.

I was quite thorough with the refactor (_easier to follow through the staged commit diffs_). **I am confident** in it and the revised change detection helper.

## Type of change

## Checklist:


test/helper/change-detection.bash
```diff
@@ -0,0 +1,29 @@
...
+
+function _get_logs_since_last_change_detection() {
+  local CONTAINER_NAME=$1
+  local MATCH_IN_FILE='/var/log/supervisor/changedetector.log'
+  local MATCH_STRING='Change detected'
+
+  # Read file in reverse, collect lines until match with sed is found,
+  # then stop and return these lines back in original order (flipped again through tac):
+  docker exec "${CONTAINER_NAME}" bash -c "tac ${MATCH_IN_FILE} | sed '/${MATCH_STRING}/q' | tac"
+}
```
polarathene, line 29: I had tried a few attempts before this, this works and seemed ok to go with?

---

I had come across the following which seemed to work, but if there was no match it returned the entire input:

```bash
docker exec "${CONTAINER_NAME}" bash -c "cat ${MATCH_IN_FILE} | sed -n 'H; /${MATCH_STRING}/h; \${g;p;}'"
```

Likewise, this seemed to work, but was starting from the first match found (instead of the last one) which wasn't desirable:

```bash
docker exec "${CONTAINER_NAME}" bash -c "cat ${MATCH_IN_FILE} | sed -n '/${MATCH_STRING}/,\$p'"
```

test/helper/common.bash
```diff
@@ -172,40 +172,38 @@ function wait_for_service() {
...
+  # Default behaviour is to wait until one new match is found (eg: incremented),
+  # unless explicitly set (useful for waiting on a min count to be reached):
+  if [[ -z $EXPECTED_COUNT ]]
+  then
+    # +1 of starting count:
+    EXPECTED_COUNT=$(bc <<< "$(__get_count) + 1")
+  fi
 
-  repeat_until_success_or_timeout 60 __is_changedetector_finished
+  repeat_until_success_or_timeout 20 __has_expected_count
```
polarathene, line 206: I haven't compared, but [casper had suggested a different approach than polling](https://github.com/docker-mailserver/docker-mailserver/pull/2988#discussion_r1064246820) like I'm doing here.

```suggestion
  timeout 60 docker exec "${CONTAINER_NAME}" bash -c "tail -F '${MATCH_IN_LOG}' | grep --max-count 1 '${MATCH_CONTENT}'"
```

casperklein: I cannot reproduce the hanging. I tried this:

```bash
# exit as soon as the keyword "spawned" appears
docker logs -n0 -f mail | grep -q spawned
```

Then in another session:

`docker exec mail supervisorctl restart update-check`

polarathene: I will give it a look again, perhaps there was something else affecting it. When I tried it was with an Amavis test unrelated to change detection, so I assumed it'd be the same issue if used elsewhere. I'll see if I can reproduce again :+1:

test/helper/change-detection.bash
georglauterbach: I like the way it is at the moment best :)

test/helper/change-detection.bash
```diff
@@ -0,0 +1,29 @@
...
+# Reliably track the completion of a change event by counting events:
+function wait_until_change_detection_event_completes() {
+  local MATCH_CONTENT='Completed handling of detected change'
+  local MATCH_IN_LOG='/var/log/supervisor/changedetector.log'
+  
+  _wait_until_expected_count_is_matched "${@}"
+}
+
+function _get_logs_since_last_change_detection() {
+  local CONTAINER_NAME=$1
```
georglauterbach, line 22:
```suggestion
  local CONTAINER_NAME=${1}
```

test/helper/common.bash
```diff
@@ -172,40 +172,38 @@ function wait_for_service() {
...
-  repeat_in_container_until_success_or_timeout "${TIMEOUT}" "${CONTAINER_NAME}" bash -c 'source /usr/local/bin/helpers/index.sh; _obtain_hostname_and_domainname; cmp --silent -- <(_monitored_files_checksums) "${CHKSUM_FILE}" >/dev/null'
-}
+  # WARNING: Keep in mind it is a '>=' comparison.
+  # If you provide an explict count to match, ensure it is not too low to cause a false-positive.
+  function __has_expected_count() {
+    [[ $(__get_count) -ge "${EXPECTED_COUNT}" ]]
+  }
+
+  local CONTAINER_NAME=$1
+  local EXPECTED_COUNT=$2
```
georglauterbach, line 188:
```suggestion
  local CONTAINER_NAME=${1}
  local EXPECTED_COUNT=${2}
```

test/helper/setup.bash
```diff
@@ -66,13 +66,11 @@ function wait_for_finished_setup_in_container() {
 # in the file, then use `local TEST_TMP_CONFIG=$(duplicate_config_for_container . "${UNIQUE_ID_HERE}")`
 function init_with_defaults() {
   __initialize_variables
+  # REQUIRED: `CONTAINER_NAME` must be set before this method is called.
+  # It only affects the `TEST_TMP_CONFIG` directory created,
+  # but will be used in `common_container_create()` and implicitly in other helper methods.
```
georglauterbach, line 71: Should this be located above the function signature?

test/tests/serial/mail_changedetector.bats
```diff
@@ -1,87 +1,139 @@
...
-  run docker exec mail_changedetector_two /bin/bash -c 'supervisorctl tail -3000 changedetector'
-  assert_output --partial 'Change detected'
+  # Should have removed the stale lock file, then handle the change event:
+  run _get_logs_since_last_change_detection "${CONTAINER1_NAME}"
+  assert_output --partial 'Lock file older than 1 minute - removing stale lock file'
+  _assert_has_standard_change_event_logs
+}
+
+function _should_perform_standard_change_event() {
+  local CONTAINER_NAME=$1
```
georglauterbach, line 78:
```suggestion
  local CONTAINER_NAME=${1}
```

test/tests/serial/mail_changedetector.bats
```diff
@@ -1,87 +1,139 @@
...
+
+  # Container should have created it's own lock file,
+  # and later removed it when finished processing:
+  run _get_logs_since_last_change_detection "${CONTAINER_NAME}"
+  _assert_has_standard_change_event_logs
+}
+
+function _should_block_change_event_from_processing() {
+  local CONTAINER_NAME=$1
+  local EXPECTED_COUNT=$2
```
georglauterbach, line 92:
```suggestion
  local CONTAINER_NAME=${1}
  local EXPECTED_COUNT=${2}
```

test/tests/serial/mail_changedetector.bats
```diff
@@ -1,87 +1,139 @@
...
+  # Restore Container2 changedetector service:
+  # NOTE: The last known checksum is retained in Container2,
+  #       It will be compared to and start a change event.
+  docker exec "${CONTAINER2_NAME}" bash -c 'supervisorctl start changedetector'
+}
 
-  run docker exec mail_changedetector_one /bin/bash -c "supervisorctl tail -3000 changedetector"
-  assert_output --partial "removing stale lock file"
+function _create_change_event() {
+  echo '' >> "${TEST_TMP_CONFIG}/postfix-accounts.cf"
```
georglauterbach, line 138:
```suggestion
  echo '' >>"${TEST_TMP_CONFIG}/postfix-accounts.cf"
```

georglauterbach (changes requested):
Looks **really** good to me! Just some very minor stylistic changes :)

test/helper/common.bash
polarathene: I was able to reproduce on current master the same issue. I tested with replacing this:

https://github.com/docker-mailserver/docker-mailserver/blob/1650cdf76f0b1067192d0d12f92508b561a87597/test/tests/parallel/set1/spam_virus/spam_junk_folder.bats#L33-L35

```bash
# Passes before timing out:
timeout 20 docker exec "${CONTAINER_NAME}" bash -c "tail -F /var/log/mail/mail.log | grep --max-count 1 'Passed SPAM'"
# Same result here:
docker exec "${CONTAINER_NAME}" bash -c "timeout 20 tail -F /var/log/mail/mail.log | grep --max-count 1 'Passed SPAM'"
# Times out with failure, and due to -q has no output:
timeout 20 bash -c "docker logs --follow ${CONTAINER_NAME} | grep -q 'Passed SPAM'"
# Times out with failure, but correctly shows output of requested max lines (even if there were more present)
timeout 20 bash -c "docker logs --follow ${CONTAINER_NAME} | grep --max-count 1 'Passed SPAM'"
# Times out but **passes** (docker logs was stopped by timeout, grep was successful)
timeout 20 docker logs --follow "${CONTAINER_NAME}" | grep --max-count 1 'Passed SPAM'
```

```bash
function _passed_spam() {
  # Any method from above here
}

# Replace as 2nd assert in either test case of `tests/parallel/set1/spam_virus/spam_junk_folder.bats:
run _passed_spam
assert_success
```

---

Perhaps it's something different with our environments or how `docker logs --follow`  works on our machines? For me, given the above observations, it doesn't seem like the stream is closed?

I had also verified by running the same command in another terminal session directly while the test was running. It would find the match and output it, but not exit until the timeout was reached.

I've had a similar experience with `-q` (not as nice as `--max-count`), where either option was working at least for `decoder` as the value to match when the container was starting up (_I could not reproduce with `Passed SPAM` for some reason.._). I could repeat the command in the terminal several times and it'll exit quickly with either option successful. However a few seconds after that and it would fail, even though the logs command would clearly show the lines :shrug: 

Clearly something is iffy there, at least on my end :\

polarathene: I swapped `grep` for [`rg` (ripgrep)](https://github.com/BurntSushi/ripgrep), and now it works as you'd expect...

So something wrong with my `grep` command I guess? :man_shrugging: 

---

I'd rather avoid the `-q` or `--max-count` approach, at least with a local grep call, as it seems that's where the problem is. The containers grep was working fine with `tail`, which would be ok as it avoids false positives when I run tests locally.

We could add `ripgrep` as a dependency for testing like we do with `jq` I suppose? (_but then CI may need to grab/install it each time, unless it already provides that like it does `jq`_)

casperklein: > So something wrong with my grep command I guess? 🤷‍♂️

yes 😄 

`docker logs ${CONTAINER_NAME} | grep 'Passed SPAM {RelayedTaggedInbound,Quarantined}'"`

When you don't pass `-q` or `--max-count 1`, grep runs infinitly. Why should it stop?

If you only need to now, if a keyword was found, go with `-q`. If you need the output to parse further or make assertions, go with `--max-count 1`.

polarathene: > When you don't pass `-q` or `--max-count 1`, grep runs infinitly. Why should it stop?

You're referring to the current version in master I referenced? That is not using `--follow`, so it does not run infinitely. The attempts I tried can be found in the collapsed "> Reproduction" section.

I collapsed it after responding with an update about my local `grep` being the problem. I understand the reason for `-q` or `--max-count 1`, but only if we're using `grep` within the container (_doesn't make sense for `docker logs --follow` due to that_), as for whatever reason my local `grep` is misbehaving which would be problematic for me running tests locally :sweat_smile: (_something I've been doing quite frequently lately_)

If we're to use either option with `docker logs --follow`, then I'd need to know why my `grep` is broken and how to fix it, or add `ripgrep` as a dependency for tests.

georglauterbach (approved):

georglauterbach (approved):

georglauterbach:
Got an upcoming PR about changes to `test/helper/*.sh` which should be merged after this and #3004. The PR will adjust `mail_privacy.bats` and refactor `test/helper/*.sh`, i.e. apply our style guides and documentation. Just FYI so we avoid large merge conflicts.

polarathene:
> UPDATE: branch pushed, name = [`tests/more-reqrites-1`](https://github.com/docker-mailserver/docker-mailserver/tree/tests/more-rewrites-1)

I see you've got a nice clean separation of commits so that's fine if you want to bundle `mail_privacy.bats` into that :+1:

If you're planning to rework more tests into that branch, maybe use a separate branch for the broader change with renaming; then after that's merged cherry-pick the extra commits into a new branch / PR? :sweat_smile:

I'm not too bothered either way though, but it is preferable to separate out broad diff noise like the method renaming from the "review changes" view when possible :)

---

Small heads-up, there will be some conflicts for either you or me to deal with as I'm presently putting together a test file that collects all the process checking and restarts into a single file. That affects some of the tests your function rename touches as it deletes the test cases from those parallel set tests.

It affects the following tests presently:

- `parallel/set1/spam_virus/clamav.bats`
- `parallel/set1/spam_virus/fail2ban.bats`
- `parallel/set1/spam_virus/disabled_clamav_spamassassin.bats`
- `parallel/set1/spam_virus/postgrey_enabled.bats`
- `serial/mail_fetchmail.bats`
- `serial/mail_smtponly.bats`
- `serial/mail_with_ldap.bats`
- `serial/mail_with_postgrey_disabled_by_default.bats`
- `serial/tests.bats`

It may also remove the need for `check_if_process_is_running()` helper, or I'll be modifying that method.

test/helper/common.bash
casperklein: > The attempts I tried can be found in the collapsed "> Reproduction" section.

Thanks, I overlooked that.

BTW: Are you running the tests on debian/ubuntu?

polarathene: > Are you running the tests on debian/ubuntu?

EndeavourOS (ArchLinux based)

georglauterbach:
I'll be resolving the conflicts, don't worry! I'll be using a train ride tomorrow, so I hope we can merge this until tomorrow evening - that'd be nice :D

test/helper/common.bash
```diff
@@ -180,40 +180,38 @@ function wait_for_service() {
...
-
-  # Count by completions of this debug log line from `check-for-changes.sh`:
-  local CHANGE_EVENT_END='Completed handling of detected change'
-  local NUM_CHANGE_EVENTS_BEFORE=$(__change_event_count)
+  # Default behaviour is to wait until one new match is found (eg: incremented),
+  # unless explicitly set (useful for waiting on a min count to be reached):
+  if [[ -z $EXPECTED_COUNT ]]
+  then
+    # +1 of starting count:
+    EXPECTED_COUNT=$(bc <<< "$(__get_count) + 1")
```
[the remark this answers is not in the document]
polarathene: Oh weird, I remember having trouble getting that to work at the time and had ended up with the `bc <<<` version :sweat_smile: 

That works well, thanks! :grin:

test/helper/common.bash
casperklein: Depending on how much more time you want to invest, you may try the "grep" binary from debian:

```
docker create --name grep debian
docker cp grep:/bin/grep .
docker rm grep
```

I am wondering, that such a basic tool behaves differently between linux distros.

PS: If we are ever in doubt, we should stick with debian based distros supported for testing.

casperklein (approved):

polarathene:
There was a failure related to the change detection wait method in parallel set3 `smtp-delivery`. Not sure if it's related to the last change suggestion. I re-ran the jobs and it seems to have failed that test again.

polarathene:
Confirmed. The `mail_changedetector` test didn't error as that conditional path in the helper method isn't covered by it. When no explicit expected count is set, which is the case for usage elsewhere, it should set the +1 of current count.

The change @casperklein proposed caused instant failure during `setup_file()` when called. Reverted to how it was before which works correctly.

casperklein (approved):

test/helper/common.bash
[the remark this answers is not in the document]
casperklein: No glue why this doesn't work and the tests fail 🤷

georglauterbach (approved):

georglauterbach:
Can be merged. I'll be placing a new PR then for the refactored helper ;)

georglauterbach (dismissed):

casperklein:
> There was a failure related to the change detection wait method in parallel set3 `smtp-delivery`.

How to run only the set3 tests? This doesn't work:

```
make tests/parallel/set3

parallel: invalid option -- '-'
parallel [OPTIONS] command -- arguments
        for each argument, run command with argument, in parallel
parallel [OPTIONS] -- commands
        run specified commands in parallel
   bats warning: Executed 0 instead of expected 31 tests

31 tests, 0 failures, 31 not run in 1 seconds

make: *** [Makefile:50: tests/parallel/set3] Error 1
```

georglauterbach:
Did you run `make generate-accounts` before? Running the command `make clean generate-accounts tests/parallel/setX` works for me.

test/helper/common.bash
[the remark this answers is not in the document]
casperklein: Alright, I found the root cause.

Because, `set -e` is in use, the $(()) operation fails, because the called function __get_count returns exit code 1, when there are 0 matches.

As 0 matches is a valid return value (for us), this should be fixed by appending `|| true`:

`docker exec "${CONTAINER_NAME}" grep --count "${MATCH_CONTENT}" "${MATCH_IN_LOG}" || true`.

After that, `EXPECTED_COUNT=$(( $(__get_count) + 1 ))` works fine.

casperklein:
I did run just "make" to create all necessary stuff. Then tests failed and I wanted to re-run only set3.

georglauterbach:
Tested it again; runs smoothly on my system. Packages are up-to-date.

casperklein:
> EDIT: Does `make clean generate-accounts test/smtp-delivery` work?

yes

polarathene:
Ok, missed my opportunity to merge with both approvals it seems :stuck_out_tongue: 

I've added the solution @casperklein requested, hope this is good for merging now :+1:

casperklein (approved):

georglauterbach (approved):

georglauterbach (approved):

Issue #2613, opened 2022-06-01
Postfix and dovecot killed by supervisor on each mail creation


abstract-entity:
# Postfix and dovecot killed by supervisor on each mail creation

I would like some feedback concerning a use case

## Description

Hello,

Each time i create an user, i can see in logs postfix and dovecot restarting ! 

``` BASH
addmailuser -y toto@toto.com password
```

Containers logs:
``` BASH
Jun  1 08:40:31 feature-mail postfix/master[29611]: terminating on signal 15
Jun  1 08:40:33 feature-mail dovecot: master: Warning: Killed with signal 15 (by pid=7 uid=0 code=kill)
Jun  1 08:40:34 feature-mail dovecot: master: Dovecot v2.3.13 (89f716dc2) starting up for imap, lmtp
Jun  1 08:40:37 feature-mail postfix/master[2385]: daemon started -- version 3.5.6, configuration /etc/postfix
```

changedetector logs: 
``` BASH
[   INF   ]  2022-06-01 08:40:21  Change detected
postfix: stopped
postfix: started
dovecot: stopped
dovecot: started
```

Is this normal? is there an impact on the reception of emails?

polarathene:
Do you plan on adding accounts frequently? (eg: more than once every minute?)

Otherwise there shouldn't be anything to worry about. Change detection support watches config and certificate files for changes in content. If it notices a change, it reprocesses them and restarts Postfix and Dovecot to apply the changes.

In some cases they could probably just use their reload commands. I am working on improving the change detection support, but it might be a while before it's much smarter.

---

> Is this normal?

Yes it's normal. Adding an account updates the config file `postfix-accounts.cf` which is where each account is stored. During startup of the container or when processing changes, that file is read and accounts regenerated (the mail storage is left as-is).

> is there an impact on the reception of emails?

Potentially an issue I think if receiving or sending an email that's being worked on to arrive in the users mail account or send out, sometime there's a bit of a delay from completing those actions (eg sitting in a queue). In a spam test we have I noticed it took around 30 seconds to store after receiving it (it was queued and Amavis didn't process it until about 30 seconds later).

Outside of that window, when Postfix is down and starting back up, an e-mail could arrive and not be delivered, but the service sending the mail is likely to try again not too long afterwards, so it should arrive again. Postfix also is aware of mail it has not successfully sent AFAIK, and will attempt again.

[...]

PR #2998, opened 2023-01-11
fix: Workaround `postconf` write settling logic

### LINKED ISSUE #2985: [TODO] Remove unnecessary 2 second delays from Postfix utilities during tests
## Description

This issue is mostly to document an observation that impacts testing time from some Postfix utilities introducing 2 second delays when used if `/etc/postfix/main.cf` has been modified within that time frame.

---

### Background

In May 2022 a change was made for how Postfix config overrides were applied. An observation was noticed [here](https://github.com/docker-mailserver/docker-mailserver/pull/2598#issuecomment-1141561786). The [original approach](https://github.com/docker-mailserver/docker-mailserver/pull/2598/files#diff-141cf3167e0d3e34aa4f5242772a4e7519a0abb8a7cd2ba4493e8ea0ed607a61L738-L747) does not incur the observed delay.

Recently I investigated this some more to reduce a few seconds from every container spun up for CI tests. I had noticed that if the configs for that override logic weren't present, 2 seconds was shaved off. And another part of the startup scripts was adding 2 seconds as well with `postalias /etc/aliases`. It turns out these Postfix utilities are reading `/etc/postfix/main.cf` during execution.

I found `pausing to let /etc/postfix/main.cf cool down` would be output with `-vv` verbosity set when running commands. This prompted to look into the source for Postfix utilities, I found the [following polling / delay logic](https://github.com/vdukhovni/postfix/blob/1f2df46e5211065a16f78739308131122e185c7c/postfix/src/util/load_file.c#L76-L78) (_`doze()` is another utility method in the source that takes microseconds_). It compares the `mtime` of the file to load to be less than 1 second passed or a value in the future, and waits 300ms until checking again.

Thus anytime startup modifies `/etc/postfix/main.cf` and tries to run one of the postfix utilities shortly after, there is a delay measured of about 1.5-2 seconds. 

`postconf 'max_idle = 600s'` can be run with no penalty. But `postconf max_idle` to perform a lookup will incur that delay.

---

## To consider

This is probably only a concern with tests where the start-up time can be compounded for common start-up logic.

- Only one test actually hit the config override path, and that has been [addressed in a recent PR](https://github.com/docker-mailserver/docker-mailserver/pull/2980/files#diff-1420e606ddac9d72de297d6a257e6c98fbf68c4dd7a5ee6f985d27f7ec09edda).
- All tests presently add a default `root` postmaster alias. This hit can be worked around by modifying the `mtime` (_`stat -c %Y /path/to/file` for checking this value as unix epoch timestamp_) via `touch -d '2 seconds ago' /path/to/file`. I think that should be fine? (_This also reduces the output spam from the original config override observation PR, avoids the repetition_)
- Other tests or logic may need to be evaluated for similar delays incurred by this stalling behaviour (_although AFAIU it's intended to ensure a file is actually ready to be read by no longer being written to_). An alternative approach may be to replace the `sed` + `postconf` calls and instead queue up the modifications to be applied.

polarathene:
# Description

After updating `main.cf`, to avoid an enforced delay from reading the config by postfix tools, we can ensure the modified time is at least 2 seconds in the past as a workaround. This should be ok with our usage AFAIK.

Shaves off 2+ seconds roughly off each container startup, reduces roughly 2+ minutes off tests.

Fixes #2985

## Type of change

## Checklist:

### DIFF
--- a/target/bin/restrict-access
+++ b/target/bin/restrict-access
@@ -44,8 +44,7 @@ case "${MODE}" in
           sed -i 's|smtpd_sender_restrictions =|smtpd_sender_restrictions = check_sender_access texthash:/tmp/docker-mailserver/postfix-send-access.cf,|' /etc/postfix/main.cf \
         || sed -i 's|smtpd_recipient_restrictions =|smtpd_recipient_restrictions = check_recipient_access texthash:/tmp/docker-mailserver/postfix-receive-access.cf,|' /etc/postfix/main.cf
 
-      _adjust_mtime_for_postfix_maincf
-      postfix reload
+      _reload_postfix
     fi
 
     echo -e "${USER} \t\t REJECT" >>"${DATABASE}"
--- a/target/bin/restrict-access
+++ b/target/bin/restrict-access
@@ -44,7 +44,8 @@ case "${MODE}" in
           sed -i 's|smtpd_sender_restrictions =|smtpd_sender_restrictions = check_sender_access texthash:/tmp/docker-mailserver/postfix-send-access.cf,|' /etc/postfix/main.cf \
         || sed -i 's|smtpd_recipient_restrictions =|smtpd_recipient_restrictions = check_recipient_access texthash:/tmp/docker-mailserver/postfix-receive-access.cf,|' /etc/postfix/main.cf
 
-      service postfix reload >/dev/null
+      _adjust_mtime_for_postfix_maincf
+      postfix reload
     fi
 
     echo -e "${USER} \t\t REJECT" >>"${DATABASE}"
--- a/target/scripts/check-for-changes.sh
+++ b/target/scripts/check-for-changes.sh
@@ -96,12 +96,6 @@ function _reload_amavis
   fi
 }
 
-function _reload_postfix
-{
-  _adjust_mtime_for_postfix_maincf
-  postfix reload
-}
-
 # Also note that changes are performed in place and are not atomic
 # We should fix that and write to temporary files, stop, swap and start
 function _postfix_dovecot_changes
--- a/target/scripts/check-for-changes.sh
+++ b/target/scripts/check-for-changes.sh
@@ -59,7 +59,7 @@ function _check_for_changes
     _log_with_date 'debug' 'Reloading services due to detected changes'
 
     [[ ${ENABLE_AMAVIS} -eq 1 ]] && _reload_amavis
-    postfix reload
+    _reload_postfix
     [[ ${SMTP_ONLY} -ne 1 ]] && dovecot reload
 
     _remove_lock
@@ -94,6 +94,12 @@ function _reload_amavis
     # reading this file again in case of new domains, otherwise they will be ignored.
     amavisd-new reload
   fi
+}
+
+function _reload_postfix
+{
+  _adjust_mtime_for_postfix_maincf
+  postfix reload
 }
 
 # Also note that changes are performed in place and are not atomic
--- a/target/scripts/helpers/aliases.sh
+++ b/target/scripts/helpers/aliases.sh
@@ -54,6 +54,7 @@ function _handle_postfix_aliases_config
   local DATABASE_ALIASES='/tmp/docker-mailserver/postfix-aliases.cf'
   [[ -f ${DATABASE_ALIASES} ]] && cat "${DATABASE_ALIASES}" >>/etc/aliases
 
+  _adjust_mtime_for_postfix_maincf
   postalias /etc/aliases
 }
 
--- a/target/scripts/helpers/utils.sh
+++ b/target/scripts/helpers/utils.sh
@@ -57,3 +57,9 @@ function _adjust_mtime_for_postfix_maincf
     touch -d '2 seconds ago' /etc/postfix/main.cf
   fi
 }
+
+function _reload_postfix
+{
+  _adjust_mtime_for_postfix_maincf
+  postfix reload
+}
--- a/target/scripts/helpers/utils.sh
+++ b/target/scripts/helpers/utils.sh
@@ -44,3 +44,13 @@ function _require_n_parameters_or_print_usage
   [[ ${1:-} == 'help' ]]  && { __usage ; exit 0 ; }
   [[ ${#} -lt ${COUNT} ]] && { __usage ; exit 1 ; }
 }
+
+# NOTE: Postfix commands that read `main.cf` will stall execution,
+# until the config file has not be written to for at least 2 seconds.
+# After we modify the config explicitly, we can safely assume (reasonably)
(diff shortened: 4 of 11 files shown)


georglauterbach (dismissed):
LGTM 👍🏼 Nice work!

georglauterbach:
The pipeline `Test Merge Requests / Test AMD64 Image / Test (serial) (pull_request)` is now `Successful in 13m`. We're already pretty close to 12min with the tests. When we improve on this further, and we definitely can (and will :D), we will see times lower than 12mins, so the next logical step when it comes to improving the CI is to speed up the ARM64 build. @polarathene @casperklein @wernerfred do you have any ideas how to do that?

polarathene:
> When we improve on this further, and we definitely can (and will :D), we will see times lower than 12mins

Yes definitely possible. There's still the supervisorctl process restarting checks that can all be moved into a single test file I think. Presently I am just focusing on adjusting existing serial tests to the new format, and then they'll be distributed into the parallel sets which will bring the time down nicely :+1: 

> do you have any ideas how to do that?

It would require leveraging the image cache well. It's a bit tricky with Github CI cache keys IIRC. On a new PR, it can use a cache key from master branch, but master cannot use one from a PR, nor can other PRs use  cache keys created by adjacent PRs.

I'd have to look over it some time to refresh what can be done. I do recall there is an experimental cache mode for Github CI, our buildx action has a `gha` mode that's meant to be better at caching IIRC. I won't be investigating that until I've tackled converting all these tests. Even then I've still got other tasks to prioritize :sweat_smile:

polarathene:
> I will be doing that as well to help you in this regard.

I don't mind raising the PRs, I was slowed down on the `mail_changedetector` one to remove the sleeps as it required some extra work on the change detection helper and understanding the test better, extra careful not to mess it up :sweat_smile: 

If you are planning to update other tests do gives a heads-up on which ones you're going to tackle. I would also encourage just converting them without the rename, so we can just do a PR afterwards that handles that for preserving history :+1: 

I plan to spend some time on `mail_hostname` (_if I can find my old local branch that had a revision of it sitting around from a while back :grimacing: )

georglauterbach:
> If you are planning to update other tests do gives a heads-up on which ones you're going to tackle. I would also encourage just converting them without the rename, so we can just do a PR afterwards that handles that for preserving history 👍

I will be tackling

- `mail_time`
- `sedfile`
- `no_container`

next :D

> I plan to spend some time on `mail_hostname` (_if I can find my old local branch that had a revision of it sitting around from a while back 😬 )

There are plans to revise the DNS name stuff (I already added the [new ENV variables to the scripts](https://github.com/docker-mailserver/docker-mailserver/blob/master/target/scripts/helpers/variables.sh#L104-L107)). Probably worthwhile investigating this first and adjusting the tests according to these changes?

casperklein (approved):

polarathene:
> I'm thinking perhaps to add a check on the `mtime` being `< 2 seconds ago` into this util method. Do you think it'd be useful?

- I've now added the conditional check for this, and added it into `check-for-changes.sh` which may benefit from it.
- Also identified a few other locations where it should be applied. Since it was needed across a few tests that manually modify `main.cf`, I've added a helper method into `common.bash`.

polarathene:
[Failure](https://github.com/docker-mailserver/docker-mailserver/actions/runs/3897558295/jobs/6655376329#step:5:64):

```
not ok 56 checking smtp_only: mail send should work in 1327ms
# (from function `_reload_postfix' in file test/test_helper/common.bash, line 17,
#  in test file test/tests/serial/mail_smtponly.bats, line 49)
#   `_reload_postfix mail_smtponly' failed
```

Reproduced locally, looking into it :sweat_smile:

polarathene:
Identified that:

- `docker exec "${CONTAINER_NAME}" sh -c "/etc/init.d/postfix reload"` works.
- `docker exec "${CONTAINER_NAME}" sh -c "service postfix reload"` works.
- `docker exec "${CONTAINER_NAME}" sh -c "postfix reload"` **fails**.

Yet it only fails in this test :thinking: 

---

I've got a separate branch where I was looking into removing the need for the `postfix-wrapper.sh`, and when I apply the changes for that, then `sh -c 'postfix reload'` does not fail. So it's presumably something to do with `postfix-wrapper.sh`?

Maybe this PR should be deferred until I have that related change (remove needing `postfix-wrapper.sh`) considered for merging?

georglauterbach (dismissed):

target/bin/restrict-access
```diff
@@ -44,7 +44,8 @@ case "${MODE}" in
           sed -i 's|smtpd_sender_restrictions =|smtpd_sender_restrictions = check_sender_access texthash:/tmp/docker-mailserver/postfix-send-access.cf,|' /etc/postfix/main.cf \
         || sed -i 's|smtpd_recipient_restrictions =|smtpd_recipient_restrictions = check_recipient_access texthash:/tmp/docker-mailserver/postfix-receive-access.cf,|' /etc/postfix/main.cf
 
-      service postfix reload >/dev/null
+      _adjust_mtime_for_postfix_maincf
+      postfix reload
```
casperklein, line 48:
```suggestion
      _reload_postfix
```

if `_reload_postfix` function is moved from `check-for-changes.sh` to `utils.sh`, the function can be used here.

casperklein:
> * I've now added the conditional check for this

Just to be save or do you expect any problems?

polarathene:
> Just to be save or do you expect any problems?

Not from what I understand. In the [associated issue](https://github.com/docker-mailserver/docker-mailserver/issues/2985) I linked to the source code upstream in Postfix that is responsible for the delay.

It's a safety guard for reading `main.cf` by checking `mtime` isn't too recent so that it can be assumed safe to read (_not in a partially updated state, all writes completed / settled_). It's a valid condition to wait on to ensure correct config is read.

I believe it's safe to read after make changes, but perhaps depending on hardware or third-party software something may not play well (or try to abuse/exploit it?). Postfix will tend to ignore anything that it doesn't recognize as valid config keys, which could result in a default setting used instead, or the key is valid but the value is not the intended value? I think exploiting sensitive timing like that is a difficult attack and if that were an option the system is likely already vulnerable to much worse?

---

It's mostly tests that will benefit from this, or in actual deployments minimal downtime during change events (_[seems to be an actual concern](https://github.com/docker-mailserver/docker-mailserver/issues/2613), I think there are some similar issues reported_). Additionally reduces container start-up time, but I doubt that matters much to most users.

I am inclined to say it should be fine to workaround. It's usually more of an issue when wanting to write to a file that other processes might also attempt, or reading while a file is actively being written to (what this protects against).

georglauterbach (approved):

casperklein (approved):

PR #3001, opened 2023-01-12
scripts/config: update DNS configuration ENV

georglauterbach:
# Description

This PR updated the handling of DNS related ENV variables. This was on my todo list, and @polarathene has mentioned he wanted to do something about this too. Hence I figured I'd go over the scripts and do a fancy search-and-replace and add the new variables. @polarathene, you can then with a later PR change the functionality in case you need/want to improve on this further. This PR just solves the task of deprecating `OVERRIDE_HOSTNAME` and adding the new ENVs `DMS_FQDN`, `DMS_DOMAINNAME` and `DMS_HOSTNAME`.

This is technically not a breaking change, as I made sure these changes are backwards compatible (i.e. when using `OVERRIDE_HOSTNAME`). Users will just see a deprecation warning and a link to help them.

Reviewing commit-by-commit should be the most pleasant reviewing experience :D

## Type of change

## Checklist:

### DIFF
--- a/target/bin/open-dkim
+++ b/target/bin/open-dkim
@@ -2,6 +2,8 @@
 
 # shellcheck source=../scripts/helpers/index.sh
 source /usr/local/bin/helpers/index.sh
+# shellcheck source=/dev/null
+source /etc/dms-settings
 
 KEYSIZE=4096
 SELECTOR=mail
@@ -108,7 +110,6 @@ function _generate_domains_config
   # unless CLI arg DOMAINS provided an alternative list to use instead:
   if [[ -z ${DOMAINS} ]]
   then
-    _obtain_hostname_and_domainname
     # uses TMP_VHOST:
     _vhost_collect_postfix_domains
   else
--- a/target/bin/open-dkim
+++ b/target/bin/open-dkim
@@ -2,8 +2,14 @@
 
 # shellcheck source=../scripts/helpers/index.sh
 source /usr/local/bin/helpers/index.sh
-# shellcheck source=/dev/null
-source /etc/dms-settings
+
+if [[ -f /etc/dms-settings ]]
+then
+  # shellcheck source=/dev/null
+  source /etc/dms-settings
+else
+  _handle_dns_names
+fi
 
 KEYSIZE=4096
 SELECTOR=mail
--- a/target/scripts/helpers/dns.sh
+++ b/target/scripts/helpers/dns.sh
@@ -7,59 +7,74 @@ function _get_label_count
   awk -F '.' '{ print NF }' <<< "${1}"
 }
 
-# Sets HOSTNAME and DOMAINNAME globals used throughout the scripts,
-# and any subprocesses called that intereact with it.
-function _obtain_hostname_and_domainname
+# This function is called very early during the setup process, directly after setting up Supervisor.
+# It will check whether DNS-related variables (https://docker-mailserver.github.io/docker-mailserver/edge/config/environment/#dns-names)
+# are set. If not, it will try to derive the values from the environment.
+function _obtain_dns_names
 {
-  # Normally this value would match the output of `hostname` which mirrors `/proc/sys/kernel/hostname`,
-  # However for legacy reasons, the system ENV `HOSTNAME` was replaced here with `hostname -f` instead.
-  #
-  # TODO: Consider changing to `DMS_FQDN`; a more accurate name, and removing the `export`, assuming no
-  # subprocess like postconf would be called that would need access to the same value via `$HOSTNAME` ENV.
-  #
-  # ! There is already a stub in variables.sh which contains DMS_FQDN. One will just need to uncomment the
-  # ! correct lines in variables.sh.
-  #
-  # TODO: `OVERRIDE_HOSTNAME` was introduced for non-Docker runtimes that could not configure an explicit hostname.
-  # Kubernetes was the particular runtime in 2017. This does not update `/etc/hosts` or other locations, thus risking
-  # inconsistency with expected behaviour. Investigate if it's safe to remove support. (--net=host also uses this as a workaround)
-  export HOSTNAME="${OVERRIDE_HOSTNAME:-$(hostname -f)}"
-
-  # If the container is misconfigured.. `hostname -f` (which derives it's return value from `/etc/hosts` or DNS query),
-  # will result in an error that returns an empty value. This warrants a panic.
-  if [[ -z ${HOSTNAME} ]]
+  # TODO remove when OVERRIDE_HOSTNAME is dropped in v13.0.0.
+  # We return early when OVERRIDE_HOSTNAME is set because it will be used to set the values
+  # of the other DNS related variables.
+  if [[ -n ${OVERRIDE_HOSTNAME+set} ]]
   then
-    dms_panic__misconfigured 'obtain_hostname' '/etc/hosts'
+    return 0
   fi
 
-  # If the `HOSTNAME` is more than 2 labels long (eg: mail.example.com),
-  # We take the FQDN from it, minus the 1st label (aka _short hostname_, `hostname -s`).
-  #
-  # TODO: For some reason we're explicitly separating out a domain name from our FQDN,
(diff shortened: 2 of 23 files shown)


casperklein:
We use `example.com` in our docs which is recommended. Why did you introduce `example.test` now?

georglauterbach:
> We use `example.com` in our docs which is recommended. Why did you introduce `example.test` now?
> 
> PS: `example.test` was only used in our tests so far. IMO this should also be changed to example.com.

I thought we wanted to slowly convert to using `example.test` since `.test` is a known TLD to be used for tests. This helps resolvers, i.e. they know they do not need to issue a query. I think this is a very nice addition, so I changed it on the fly.

target/bin/open-dkim
```diff
@@ -2,6 +2,8 @@
 
 # shellcheck source=../scripts/helpers/index.sh
 source /usr/local/bin/helpers/index.sh
+# shellcheck source=/dev/null
+source /etc/dms-settings
```
polarathene, line 6: Does that work without DMS running? If you use `setup.sh` to setup DKIM it skips the startup scripts doesn't it?

target/scripts/helpers/dns.sh
```diff
@@ -7,59 +7,106 @@ function _get_label_count
...
+  then
+    echo "
+
+# The following lines are desirable for IPv6 capable hosts
+::1     ip6-localhost ip6-loopback
+fe00::0 ip6-localnet
+ff00::0 ip6-mcastprefix
+ff02::1 ip6-allnodes
+ff02::2 ip6-allrouters" >>/etc/hosts 2>/dev/null
+  fi
```
polarathene, line 111: I'm pretty sure I'm not going to approve of this.

Docker does a fair bit here depending on Docker daemon config, the host itself and how the container is run or any connected networks. We should not mess with it.

polarathene (changes requested):
To clarify the `example.com` vs `example.test` discussion:
- [`.test`](https://en.wikipedia.org/wiki/.test) is a special TLD for use in testing. DNS resolvers should ignore it. Presently we still have some email domains in tests that if they fail, attempt to send a bounce mail to the external domain, or similar network activity to try reach an external server / DNS lookup (`my-domain.com`).
- [`example.com`](https://en.wikipedia.org/wiki/.example) is similar, but intended for usage in documentation as a placeholder of the readers actual domain.

So we keep both. `.test` for usage in tests. `example.com` for documentation.

---

This is not going to be something I have time to allocate for a proper review any time soon.

It looks like you want to push a bunch of new changes for v12 here, and I don't feel that I can approve that. This is a feature that needs careful consideration (_we've had a variety of bugs from changes in the past for different reasons I've documented in the past_).

Tests will be reworked first. Then conversion to `compose.yaml` with a DNS container and replacing `nc` mail delivery testing with proper container to container with FQDNs instead of `0.0.0.0` or similar IP addresses will be put into place to properly test for this change (_ensures nothing is missed as current test approach bypasses a fair amount_). The DNS container will ensure FQDNs with appropriate DNS records is supported. I've already got a basic setup for that to adapt once we get to the next phase of formatting tests.

At a glance through the changes, this is not quite what I remember proposing to go forward with. I have to refer to my notes and old PR discussions related to the topic, and I'll do that closer to the time as it requires a fair amount of investment in time on my part :sweat_smile: 

---

All that aside, this is great and I really appreciate you taking a pass at this! :grinning: I will just need to revisit it later :+1: 

I think we have enough other changes being introduced for v12 that I'm putting forward where I'd rather not mix in a change like this with them should we get weird bug reports to debug and try to track down the cause. This is definitely something better suited for v13.

target/scripts/helpers/dns.sh
casperklein: My current `/etc/hosts` looks like this:

```
127.0.0.1       localhost
::1     localhost ip6-localhost ip6-loopback
fe00::0 ip6-localnet
ff00::0 ip6-mcastprefix
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
172.19.0.2      mail.example.com mail
```

This looks already fine to me. I am not sure, what exactly is achieved by manually creating the hosts file.

target/bin/open-dkim
georglauterbach: It does indeed _not_ work haha

target/scripts/helpers/dns.sh
georglauterbach: The rationale here is that one should not need to provide `hostname:` or something similar with Compose or other environments. The ENV should be enough. I can make the changes less intrusive though, and just append something instead of recreating it.

georglauterbach:
I think I should have made my intention more clear. This PR is not about a change in functionality but rather about providing an updated base to adjust DNS name handling in the future.

> Tests will be reworked first.

This sounds rather final. From what I read, this looks as if it is not going to be merged anytime soon. I do not actually understand why this is the case. I tried to design it in a way that we can build _on top_ of it, not rebuild it later. Could you please clarify?

polarathene (changes requested):
> This PR is not about a change in functionality but rather about providing an updated base to adjust DNS name handling in the future.

It is absolutely introducing changes in functionality under the hood. Changes that I am extremely uncomfortable with introducing.

> This sounds rather final. From what I read, this looks as if it is not going to be merged anytime soon.

It is not something I am comfortable to approve until our testing situation can instil more confidence in the changes.

I am aware of numerous bugs in the past for this project where maintainers changed this logic with assumptions. Significant time was spent investigating that in the past, and reviewing/collaborating on getting to the scripts we presently have with confidence.

I am rarely this firm on blocking changes, but this is one I feel I must be, and hope that other maintainers agree with. I want this change, but I want to ensure it's done right.

You don't seem to mind overwriting `/etc/hosts` into a single entry? I am fairly certain that will break functionality under some configurations (_we've had problems with this in the past_). There are other changes, it's not just renaming some variables, but moving away from `--hostname` to ENV.

I would love to review and collaborate on this change with you, but please understand my reluctance until we've got better testing in place to be more confident about the impact of such changes. Do note that `docker run` vs `docker-compose up` have had disparity in this area in the past, as have `docker-compose` vs `docker compose` (v2) commands. I documented a fair amount of this in a previous PR (_some of it was lost to a crash before posting unfortunately_).

> I do not actually understand why this is the case. I tried to design it in a way that we can build on top of it, not rebuild it later.
> Could you please clarify?

Hopefully the above or review comments help clarify.

This is a breaking change proposed. I don't see the point in being eager to change this logic to a half-way point before we are in a better position to be confident about those changes.

It's the tests that are the time consuming part. The actual logic changes that we want isn't difficult to make, but having confidence in them presently is. I really want to avoid a repeat of past problems.

I especially do not want such changes mixed into v12 release. There are still a few other changes unrelated to tests that I want to push and if we were to have any issues reported, it would easier to troubleshoot when a network change like this isn't muddled into other breaking changes.

---

Is that all ok? I don't mean to be harsh / pushy. I did not plan to invest so much time into DMS lately, but I've switched it to be my priority so that these long standing issues can be resolved.

I ask for a bit more patience, I have been spending 6-12 hours daily on DMS related tasks.

polarathene:
The [current test failure](https://github.com/docker-mailserver/docker-mailserver/actions/runs/3911595738/jobs/6685189167#step:5:196) rings a bell, and I've probably documented it somewhere in an issue or PR comment :sweat_smile: 

```
not ok 188 checking system: sets the server fqdn in 141ms
#  in test file test/tests/serial/tests.bats, line 299)
#   `assert_output "mail.example.test"' failed
#
# -- output differs --
# expected : mail.example.test
# actual   : 8d62c9acaaf6
# --
```

I believe you try to fix that by setting `/etc/hostname` explicitly, but this doesn't work. If you find the comment about it from the past issue/PR, I'm pretty sure I document how to make that work, but I think it required adding a capability, and may have only worked once then error if attempted again.

georglauterbach:
Read all the feedback; looks as if this should really be merged at a later point in time. I am too in favour of converting all the tests now! I will help with this. Thanks for taking the time and detailed feedback!

---

I will close this in the meantime and we can come back to it when the time has come :)

PR #3010, opened 2023-01-17
tests(refactor): `mail_fetchmail.bats` + co-locate test cases for processes

polarathene:
# Description

All common test cases that checked if a process was running or that it restarts have been migrated into a new test file that co-locates them. Seems more appropriate to centralize that into one place. Brings serial test time down by about 2 minutes (_540 sec run-time was my best result_).

Commits have been staged out with additional context in their commit messages to explain certain changes/choices made.

Additionally:
- The fetchmail test files were merged into one and updated to the new format. A postgrey test that only checked the service was disabled due to ENV also removed as redundant.
- `postsrsd` in supervisord config had it's [`autorestart` policy](http://supervisord.org/configuration.html#program-x-section-values) as `unexpected` instead of `true` like other services. I looked up the [relevant PR](https://github.com/docker-mailserver/docker-mailserver/pull/814) and there is a commit that changes to `unexpected` but it provides no context as to why, nor any discussion in the PR thread. Tests aren't showing any issues in switching to `true` but would fail to restart `postsrsd` with `unexpected`.
- SASLAuthd config for supervisord has most commands use `-O` with an ENV, which if unset fails to run the process as the command itself is invalid. Quote wrapping the ENV in the config however treats the arg with an empty string value and runs without configuring the ENV explicitly. I decided to quote wrapped each ENV referenced in supervisord config.

---

## Details

The original test cases were not great and the revised versions were a bit too eager to be reliable. A common test function is called by iterating an array of processes to test against.
- It'll wait until the process has been running for a little bit of time.
- Then kill the process and wait until that process is no longer running (by polling the process age until no result), as some processes take a couple seconds to be killed.
- Finally, the new process should be running, but we poll again as this isn't always the case. Postfix and Fail2Ban are managed via wrapper scripts which have `sleep 5` that delay the restart. I've documented local observations of the timings in the test itself.

Earlier commits did have separate test cases calling the same function instead of running through a loop. The loop seems fine and correctly provides errors mentioning the involved process in the event of a failure. A similar loop is used for the disabled ENV test case.

### Optional Optimization (Remove up to 30 seconds test time)

I was able to bring the enabled test case time down to approx 10 seconds (_it can be 20-30 currently, with another 10 if keeping the separate ClamAV test case_). It would be a bit more verbose and add extra complexity for maintenance though, so I opted against it.

Realistically it seems to run in 10 seconds vs 20 seconds of the current approach (_with clamd not in a separate test case, and wrapper scripts removed_). If useful here is that earlier attempt:

<details>
<summary>Alternative process checking test case</summary>

```bash
@test "enabled - should restart processes when killed" {
  local MIN_PROCESS_AGE=4

  # Average time: 4 seconds (7 seconds with chroot + wrapper)
  run repeat_until_success_or_timeout 10 _should_have_all_processes_running "${MIN_PROCESS_AGE}"
  assert_success

  # Average time: 2 seconds
  _should_kill_all_processes_successfully

  # Average time: 3 seconds
  local PROCESS_LIST=("${ENABLED_PROCESS_LIST[@]}")
  run repeat_until_success_or_timeout 10 _should_have_restarted_all_processes "${MIN_PROCESS_AGE}"
  assert_success

  # Average time: 1 second (4-6 seconds with chroot + wrapper)
  run repeat_until_success_or_timeout 10 _should_have_all_processes_running
  assert_success
}

function _is_process_running() {
  local PROCESS=${1}
  local MIN_SECS_RUNNING
  [[ -n ${2} ]] && MIN_SECS_RUNNING="--older ${2}"

  docker exec "${CONTAINER_NAME}" pgrep --list-full ${MIN_SECS_RUNNING} "${PROCESS}"
}

function _should_have_all_processes_running() {
  for PROCESS in "${ENABLED_PROCESS_LIST[@]}"
  do
    if [[ ! $(_is_process_running "${PROCESS}" "${1}") =~ "${PROCESS}" ]]
    then
      echo "(fail) '${PROCESS}' is not running"
      return 1
    fi
  done
}

# Kill each process to trigger a restart:
function _should_kill_all_processes_successfully() {
  for PROCESS in "${ENABLED_PROCESS_LIST[@]}"
  do
    _run_in_container pkill --echo "${PROCESS}"
    assert_output --partial "${PROCESS}"
  done
}

# NOTE: Processes that have been restarted are removed from additional iterations,
# so that they don't fail the check while waiting on other processes to restart.
# NOTE: `--older` being too low (eg: 2) will match restarted process age, causing failure.
function _should_have_restarted_all_processes() {
  local STILL_RUNNING=()

  for i in "${!PROCESS_LIST[@]}"
  do
    local PROCESS="${PROCESS_LIST[$i]}"

    if [[ "$(_is_process_running "${PROCESS}" "${1}")" =~ "${PROCESS}" ]]
    then
      # Output helpful for debugging failures:
      echo "(fail) '${PROCESS}' is still running"
      STILL_RUNNING+="${PROCESS}"
    else
      # Output helpful for debugging failures:
      echo "(pass) Excluding '${PROCESS}' from next iteration"
      unset PROCESS_LIST[$i]
    fi
  done

  # Fail if any older processes were still found running:
  [[ ! ${#STILL_RUNNING[@]} -gt 0 ]]
}
```

</details>

### Maintenance notes

Some frustrations in helpers to be aware of if someone maintains the test in future:
- `run_until_success_or_timeout ...` - Cannot assert output if encountering a timeout failure. Only the `timedout` command message is emitted.
- `run repeat_until_success_or_timeout ...` - Captures the output of each polled function call. Fine for success status, but if you try to **refute** output against the success, the previous failures get considered as well.
- `repeat_until_success_or_timeout run ...`   - The first call is always successful even if the function called failed, since `run` doesn't bubble that error up. Causes false positives. Cannot pair with `_run_in_container()` as a result, caution needs to be taken there (_or anything else with an implicit `run` call inside the helper_).

## Type of change

## Checklist:


test/tests/parallel/set3/process-check-restart.bats
```diff
@@ -0,0 +1,188 @@
...
+  assert_failure
+
+  # Should be running:
+  # (poll as some processes a slower to restart, such as those run by wrapper scripts adding delay via sleep)
+  run_until_success_or_timeout 30 _check_if_process_is_running "${PROCESS}"
+  assert_success
+  assert_output --partial "${PROCESS}"
+}
+
+# NOTE: CONTAINER_NAME is implicit; it should have be set prior to calling.
```
polarathene, line 171:
```suggestion
# NOTE: CONTAINER_NAME is implicit; it should have been set prior to calling.
```

georglauterbach (approved):
LGTM 👍🏼 

---

I will add documentation to the functions for process checking with my next PR. I will put out my PR ([for this branch](https://github.com/docker-mailserver/docker-mailserver/tree/tests/more-rewrites-1)) when this PR is merged (resolving the conflicts). Please wait before merging other PRs after this one until my new PR is merged in order to avoid more conflicts :)

polarathene:
> I will add documentation to the functions for process checking with my next PR.

What do you mean? In our actual docs?

Or did you mean regarding the gotchas I found with polling / timeout helper methods?

> Please wait before merging other PRs after this one until my new PR is merged in order to avoid more conflicts :)

No worries I'll hold off from tests for the time being :+1: 

---

> I'd opt for moving the already adjusted test under `test/tests/serial/` into the parallel sets with another PR so we can see which tests are really still left to be revised.

After this PR is merged, the following can be moved:

- `mail_changedetector.bats`
- `mail_fetchmail.bats`
- `mail_lmtp_ip.bats`
- `mail_tls_dhparams.bats`
- `setup-cli.bats` (_Converted but I will still refactor it a bit, it can be moved though_)

`tests.bats` has been partially converted. My plan is to continue splitting out smaller test files from that. A bunch of other tests can quickly be converted rather without trouble, they could be bundled into another PR if you'd like to wait on that before moving into parallel sets, or deferred to another PR to move them if you'd rather get the existing list above relocated.

> tell me whether you'd like to do this or whether I should take care of this.
> We should try to evenly distribute tests across the sets to save as much time as possible.

You're welcome to move the above list (_either after this merge, or a separate PR that defers `mail_fetchmail.bats` to a 2nd round with others_).

I don't mind doing it myself, I've just been delaying until I've converted the bulk of tests (_often with some refactoring first from looking into the test cases history or test times_).

test/tests/parallel/set3/process-check-restart.bats
polarathene: @georglauterbach did you want this change applied? (_I'm a little hesitant to end up in approval ping pong again_ :laughing: )

casperklein (approved):

casperklein:
I haven't look in detail for what exactly that is used for, just a general thought/improvement for the future:

```
test/tests/parallel/set3/process-check-restart.bats:

# (this allows us to more confidently check the process was restarted)`
# Wait until process has been running for at least MIN_PROCESS_AGE:
```

Instead of checking the proccess age, the PID before/after restart could be compared.

polarathene:
> the PID before/after restart could be compared.

That gets a bit complicated though, as the commented section on each process notes; there is several that have multiple processes returned.

I think they're all children to a single parent process (one of the returned results), so if that can be better filtered that could work too perhaps?

I did like matching by process with `--list-all` for the returned output to contain information about the actual process instead of just a PID(s) though. At least while writing the tests it was helpful for debugging (_`master` at one point was matching the Postfix master process, and an Amavis one that had "(master)" matched_).

georglauterbach:
> > I will add documentation to the functions for process checking with my next PR.
>
> What do you mean? In our actual docs?

Just what the function does and its parameters.

> Or did you mean regarding the gotchas I found with polling / timeout helper methods?

Probably a good idea to add them as well.

> [...] A bunch of other tests can quickly be converted rather without trouble, they could be bundled into another PR if you'd like to wait on that before moving into parallel sets, or deferred to another PR to move them if you'd rather get the existing list above relocated.

I'll have a look at that later today. Will probably do an intermediate PR to move already revised tests.

> You're welcome to move the above list (_either after this merge, or a separate PR that defers `mail_fetchmail.bats` to a 2nd round with others_).

Alright, I will tend to it.

PR #3012, opened 2023-01-18
tests(refactor): Improve consistency and documentation for test helpers

georglauterbach:
# Description

Refactors test helper. Naming was incoherent and documentation was missing.

## Type of change

- [x] Improvement (non-breaking change that does improve existing functionality)

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [x] If necessary I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #3013, opened 2023-01-21
chore(Makefile): Ensure targets are always run

polarathene:
# Description

This commit is pulled out of https://github.com/docker-mailserver/docker-mailserver/pull/3012 and is [explained in this PR comment](https://github.com/docker-mailserver/docker-mailserver/pull/2857#discussion_r1013767738):

> it is **not** defined as a target like `ALWAYS_RUN:`, but just as `.PHONY: ALWAYS_RUN` :) This suffices to tell `make` to always run the target.

`.PHONY: ALWAYS_RUN` already exists in the Makefile. This PR just applies it more broadly as decided by @georglauterbach 

---

**NOTE:** From what I have seen in [Makefile docs for `.PHONY`](https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html) this seems a bit of a different approach?

The targets seem to get appended to the `.PHONY:` target, rather than this usage of prefixing target values with `ALWAYS_RUN`? (_a target dependency that is `.PHONY`_) It seems like it may behave the same way, so perhaps it doesn't matter?

## Type of change

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ BATS_PARALLEL_JOBS     ?= 2
 
 all: lint build generate-accounts tests clean
 
-build:
+build: ALWAYS_RUN
 	@ DOCKER_BUILDKIT=1 docker build \
 		--tag $(IMAGE_NAME) \
 		--build-arg VCS_VERSION=$(shell git rev-parse --short HEAD) \
@@ -30,7 +30,7 @@ generate-accounts: ALWAYS_RUN
 
 # `docker ps`:  Remove any lingering test containers
 # `.gitignore`: Remove `test/duplicate_configs` and files copied via `make generate-accounts`
-clean:
+clean: ALWAYS_RUN
 	-@ for CONTAINER in $$(docker ps -a --filter name='^dms-test_.*|^mail_.*|^hadolint$$|^eclint$$|^shellcheck$$' | sed 1d | cut -f 1-1 -d ' '); do docker rm -f $${CONTAINER}; done
 	-@ while read -r LINE; do [[ $${LINE} =~ test/.+ ]] && sudo rm -rf $${LINE}; done < .gitignore
 
@@ -47,7 +47,7 @@ tests/serial: ALWAYS_RUN
 	@ shopt -s globstar ; ./test/bats/bin/bats $(BATS_FLAGS) test/$@/*.bats
 
 tests/parallel/set%: ALWAYS_RUN
-	@ shopt -s globstar ; ./test/bats/bin/bats $(BATS_FLAGS) \
+	@ shopt -s globstar ; $(REPOSITORY_ROOT)/test/bats/bin/bats $(BATS_FLAGS) \
 		--no-parallelize-within-files \
 		--jobs $(BATS_PARALLEL_JOBS) \
 		test/$@/**/*.bats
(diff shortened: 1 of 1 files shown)


georglauterbach (approved):

georglauterbach:
> **NOTE:** From what I have seen in [Makefile docs for `.PHONY`](https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html) this seems a bit of a different approach?
> 
> The targets seem to get appended to the `.PHONY:` target, rather than this usage of prefixing target values with `ALWAYS_RUN`? (_a target dependency that is `.PHONY`_) It seems like it may behave the same way, so perhaps it doesn't matter?

I guess both work, I have seen both too.

casperklein (approved):
BTW: Even when `.PHONY: ALWAYS_RUN` and all other "ALWAYS_RUN" are removed, the function keeps the same. All targets are always run.

georglauterbach:
> BTW: Even when `.PHONY: ALWAYS_RUN` and all other "ALWAYS_RUN" are removed, the function keeps the same. All targets are always run.
> 
> At least I couldn't identify any difference.

I thought exactly the same, and had already removed it at some point in time. But then, `make` told me the recipe was already up to date - very confusing! I don't reall know why, but this way, we're on the safe side.

As an alternative to `make`, I'd go with [`just`](https://github.com/casey/just), a proper command runner (remember: `make` is a build system!). But since everyone has `make`, I guess the transition would result in a bad development experience..

polarathene:
> As an alternative to `make`, I'd go with [`just`](https://github.com/casey/just), a proper command runner (remember: `make` is a build system!). But since everyone has `make`, I guess the transition would result in a bad development experience..

I'm not opposed to it changing from `make` in a future release. Most development using it is by us regular maintainers, and I'm not sure if it offers much value for anyone else not contributing?

[Earthly](https://earthly.dev/) has been on my radar, but I think that'd be more invasive of a change, and I'm not sure how beneficial it'd be as I think it may not support some features we're using with `buildx` (_plus that's what our CI is setup with, switching would need to be worthwhile_ :sweat_smile: ). `just` looks pretty nice, plus it's in rust which is a win for me :stuck_out_tongue:

georglauterbach:
> I'd be open towards changing from `make` in a future release too. Most development using it is by us regular maintainers, and I'm not sure if it offers much value for anyone else not contributing?

You're probably right. @casperklein what's your take on this?

> [Earthly](https://earthly.dev/) has been on my radar, but I think that'd be more invasive of a change, and I'm not sure how beneficial it'd be as I think it may not support some features we're using with `buildx` (_plus that's what our CI is setup with, switching would need to be worthwhile_ :sweat_smile: ). `just` looks pretty nice, plus it's in rust which is a win for me :stuck_out_tongue: 

I've never used earthly, I will have a look at it :) Just is something I use on a daily basis, and it works very well. I discovered it because I live Rust too 🙈

casperklein:
Pro `make`: well known and wide spread. And sufficient for our usage?
Pro something else: Could make things easier/more efficient.

Switching to something else should bring useful benefits. Otherwise I don't think it's worth to convert to something else. Also it's not that often, that our Makefile is modified. It's there and it does it's job.
But personally I don't care, if I have to run `make`, `just` or something else in the future 👍

polarathene:
> Switching to something else should bring useful benefits. Otherwise I don't think it's worth to convert to something else.

Personally the Makefile has not been a fun experience for me with it's syntax and gotchas. So I think the benefit is improved maintenance with `just` (_supposedly, I haven't looked into it much myself_). That tends to be a good thing for new maintainers too if it would be cryptic to them :sweat_smile: (_some parts aren't particularly google friendly, so a lot of sifting through `git blame` to grok what is going on_)

> Also it's not that often, that our Makefile is modified. It's there and it does it's job.

Fair. It's been stripped down quite a bit since it's first usage, but the syntax + usage perhaps more complicated to grok.

It's at least easy to use and does work well for what we need. Only a small file that shouldn't need much more maintenance going forward, so I agree there's no major reason to switch away, but I'm open to it :+1:

georglauterbach:
I started a side-by-side comparison of the `Makefile` & the corresponding `Justfile` from `just`. Although `just` is _way_ nicer, it occured to me that we would need this in our CI as well. This is inconvenient! I'd rather invest work in keeping the `Makefile` clean and short, even though I love just (I use it on a daily basis).

PR #3011, opened 2023-01-18
fix: Ensure state persisted to `/var/mail-state` retains correct group

polarathene:
# Description

This PR covers changes mostly related to `misc-stack.sh`:
- Some minor refactoring of the conditional handling in the symlink loop (see commit for details).
- Ensures `chown -R` is more explicit about not only setting the user, but also the correct group. Postfix also needed a little adjustment there.
- Moves the related functionality for Amavis to be handled with the rest instead of a separate function call.
- Fixes `misc-stack.sh` rspamd ENV condition to add to `FILES` array when enabled, not disabled. This typo was missed [during initial review](https://github.com/docker-mailserver/docker-mailserver/pull/2902).

Likewise we've had a [similar issue with our `Dockerfile` copying over ClamAV database files](https://github.com/docker-mailserver/docker-mailserver/issues/2942#issuecomment-1383512079) from their official image. This prompted reviewing `misc-stack.sh` which should fix the issue for the user.

Fixes #2942 

## References

- [I previously wrote a similar PR adjusting `misc-stack.sh`](https://github.com/docker-mailserver/docker-mailserver/pull/2273), particularly for Postfix. It may be useful reference for any future maintainer stumbling on these kind of changes.
- Contribution of the [original Amavis fix was in Feb 2017](https://github.com/docker-mailserver/docker-mailserver/pull/510). Another contributor [describes the bug in the associated issue](https://github.com/docker-mailserver/docker-mailserver/issues/507#issuecomment-277800277), which is the same bug we've recently had reported with ClamAV I think. Hence being more thorough with the `chown -R` usage in `misc-stack.sh` (_and making it more clear about the purpose via comment_).
- `COPY --link` was [introduced into the `Dockerfile` in the `v11.2` release](https://github.com/docker-mailserver/docker-mailserver/releases/tag/v11.2.0) in [this Sep 2022 PR](https://github.com/docker-mailserver/docker-mailserver/pull/2755).

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Improvement (non-breaking change that does improve existing functionality)

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] New and existing unit tests pass locally with my changes

PR #3016, opened 2023-01-21
feature: provide better rspamd suppport

georglauterbach:
# Description

This PR brings advanced rspamd support to DMS. It features:

- updated configuration with sane defaults
- proper documentation
- initial tests (which can / should be extended in the future)
- the ability to disable OpenDKIM & OpenDMARC

## Type of change

## Checklist:


docs/content/config/environment.md
```diff
@@ -114,16 +112,30 @@ Note: Emails will be rejected, if they don't pass the block list checks!
 - **0** => DNS block lists are disabled
 - 1     => DNS block lists are enabled
 
-##### ENABLE_CLAMAV
+##### ENABLE_OPENDKIM=1
```
georglauterbach, line 115: typo spotted (will be resolved when resolving PR feedback)

docs/content/config/environment.md
```diff
@@ -114,16 +112,30 @@ Note: Emails will be rejected, if they don't pass the block list checks!
...
 
-- **0** => ClamAV is disabled
-- 1 => ClamAV is enabled
+Enables the OpenDKIM service.
+
+- **1** => Enabled
+- 0 => Disabled
+
+##### ENABLE_OPENDMARC=1
```
georglauterbach, line 122: typo spotted (will be resolved when resolving PR feedback)

target/scripts/startup/setup-stack.sh
```diff
@@ -626,6 +652,8 @@ function _setup_dkim_dmarc
   # shellcheck disable=SC2016
   sed -i -E 's|^(non_smtpd_milters =.*)|\1 \$dkim_milter|g' /etc/postfix/main.cf
 
+  mkdir -p /etc/opendkim/keys/ && touch /etc/opendkim/SigningTable
```
georglauterbach, line 654: this actually belongs in the first commit; but I am too lazy to rewrite history now.. :D

target/scripts/startup/setup-stack.sh
```diff
@@ -102,19 +102,45 @@ function _setup_rspamd
...
   then
-    _shutdown 'You cannot run Amavis/SpamAssassin and Rspamd at the same time'
+    _log 'warn' 'Running Amavis/SpamAssassin & rspamd at the same time is discouraged'
   fi
 
   if [[ ${ENABLE_CLAMAV} -eq 1 ]]
   then
     _log 'debug' 'Rspamd will use ClamAV'
-    mv /etc/rspamd/local.d/disabled/antivirus.conf /etc/rspamd/local.d/antivirus.conf
+    usermod -a -G clamav _rspamd
```
georglauterbach, line 111: this is necessary as Rspamd reads from a UNIX socket, not from a TCP socket

test/tests/parallel/set1/spam_virus/rspamd.bats
```diff
@@ -0,0 +1,98 @@
...
+  # https://github.com/docker-mailserver/docker-mailserver/pull/2815/files#r991087509
+  local CUSTOM_SETUP_ARGUMENTS=(
+    --env ENABLE_CLAMAV=1
+    --env ENABLE_RSPAMD=1
+    --env CLAMAV_MESSAGE_SIZE_LIMIT=30M
+    --env ENABLE_OPENDKIM=0
+    --env ENABLE_OPENDMARC=0
+    --env PERMIT_DOCKER=host
+    --env LOG_LEVEL=trace
+    -p 11334:11334
```
georglauterbach, line 20: oops; this does not hurt but should be removed..

georglauterbach:
I marked this as high priority because I am currently in the process of actually switching from Amavis/SA to Rspamd. I discovered some misconfigurations with my first PR so I need to fix / adjust DMS (fast). Just FYI.

docs/content/config/security/rspamd.md
```diff
@@ -0,0 +1,33 @@
...
+1. place files at `/etc/rspamd/override.d/` which will override Rspamd settings and DMS settings
+2. (re-)place files at `/etc/rspamd/local.d/` to override DMS settings and merge them with Rspamd settings
+
+You can find a list of all Rspamd modules [on their website][modules].
+
+### DMS' Defaults
+
+You can choose to enable ClamAV, and Rspamd will then use it to check for viruses. Just set the environment variable `ENABLE_CLAMAV=1`.
+
+DMS disables certain modules (clickhouse, dkim_signing, elastic, greylist, rbl, reputation, spamassassin, url_redirector, metric_exporter) by default. We believe these do are not required in a standard-setup and needlessly use resources. You can re-activate them by replacing `/etc/rspamd/local.d/<MODULE>.conf` or overriding DMS' default with `/etc/rspamd/override.d/<MODULE>.conf`.
```
polarathene, line 28:
```suggestion
DMS disables certain modules (clickhouse, dkim_signing, elastic, greylist, rbl, reputation, spamassassin, url_redirector, metric_exporter) by default. We believe these are not required in a standard setup, and needlessly use resources. You can re-activate them by replacing `/etc/rspamd/local.d/<MODULE>.conf` or overriding DMS' default with `/etc/rspamd/override.d/<MODULE>.conf`.
```

target/rspamd/local.d/worker-controller.inc
```diff
@@ -0,0 +1,3 @@
+# documentation: https://rspamd.com/doc/workers/controller.html
+
+bind_socket = "0.0.0.0:11334";
```
polarathene, line 3: Is this one meant to be `0.0.0.0` or `127.0.0.1` like the others?

target/scripts/startup/setup-stack.sh
```diff
@@ -95,19 +102,44 @@ function _setup_rspamd
 
   if [[ ${ENABLE_AMAVIS} -eq 1 ]] || [[ ${ENABLE_SPAMASSASSIN} -eq 1 ]]
   then
-    _shutdown 'You cannot run Amavis/SpamAssassin and Rspamd at the same time'
+    _log 'warn' 'Running Amavis/SpamAssassin & rspamd at the same time is discouraged'
```
polarathene, line 105:
```suggestion
    _log 'warn' 'Running rspamd at the same time as Amavis or SpamAssassin is discouraged'
```

target/scripts/startup/setup-stack.sh
```diff
@@ -95,19 +102,44 @@ function _setup_rspamd
...
+  )
+
+  for MODULE in "${DISABLE_MODULES[@]}"
+  do
+    cat >"/etc/rspamd/local.d/${MODULE}.conf" << EOM
+#documentation: https://rspamd.com/doc/modules/${MODULE}.html
+
+enabled = false;
+
+EOM
```
polarathene, line 138: Is `EOM` different to the common `EOF` used for heredoc syntax elsewhere in the project? Otherwise perhaps it should be `EOF` here too?

target/scripts/startup/setup-stack.sh
```diff
@@ -599,18 +631,36 @@ function _setup_SRS
...
+  fi
+
+  [[ ${ENABLE_OPENDKIM} -eq 1 ]] || return 0
+
   _log 'debug' 'Setting up DKIM'
 
-  mkdir -p /etc/opendkim && touch /etc/opendkim/SigningTable
+  mkdir -p /etc/opendkim/keys/ && touch /etc/opendkim/SigningTable
+
+  _log 'trace' "Adding OpenDKIM tp Postfix's milters"
```
polarathene, line 651:
```suggestion
  _log 'trace' "Adding OpenDKIM to Postfix's milters"
```

test/tests/parallel/set1/spam_virus/rspamd.bats
```diff
@@ -0,0 +1,98 @@
...
+
+function setup_file() {
+  _init_with_defaults
+
+  # Comment for maintainers about `PERMIT_DOCKER=host`:
+  # https://github.com/docker-mailserver/docker-mailserver/pull/2815/files#r991087509
+  local CUSTOM_SETUP_ARGUMENTS=(
+    --env ENABLE_CLAMAV=1
+    --env ENABLE_RSPAMD=1
+    --env CLAMAV_MESSAGE_SIZE_LIMIT=30M
```
polarathene, line 15: Is this ENV relevant to the test?

polarathene (changes requested):

georglauterbach:
Thanks @polarathene for the quick review ❤️

target/rspamd/local.d/worker-controller.inc
georglauterbach: It should `0.0.0.0` here and `127.0.0.1` for the others. The controller exposes a web interface, and I want to make sure I can expose this, for example via `Traefik`. So I think we need `0.0.0.0` for the controller. Please correct me if I'm wrong :)

docs/content/config/security/rspamd.md
```diff
@@ -0,0 +1,38 @@
...
+
+!!! warning "Implementation of Rspamd into DMS is WIP!"
+
+## About
+
+Rspamd is a "fast, free and open-source spam filtering system". It offers high performance as it is written in C. Visit [their homepage][homepage] for more details.
+
+## Integration & Configuration
+
+We provide a very simple but easy to maintain setup of RSpamd. The proxy worker works in [self-scan mode][proxy-self-scan-mode]. This simplifies the setup as we do not require a normal worker. You can easily change this though by [overriding our configuration](#providing-overriding-settings).
```
polarathene, line 13:
```suggestion
We provide a very simple but easy to maintain setup of RSpamd. The proxy worker operates in [self-scan mode][proxy-self-scan-mode]. This simplifies the setup as we do not require a normal worker. You can easily change this though by [overriding the configuration by DMS](#providing-overriding-settings).
```

target/scripts/startup/setup-stack.sh
georglauterbach: For the heredoc, I doesn't make any difference. But I will streamline it to `EOF` :D

test/tests/parallel/set1/spam_virus/rspamd.bats
georglauterbach: Not at all :D

georglauterbach: Done in [d604127](https://github.com/docker-mailserver/docker-mailserver/pull/3016/commits/d604127f5ebf09ccf97281c104f0d552bbcfd52b)

target/scripts/startup/setup-stack.sh
georglauterbach: Done in 2b40f7848c53b0772262dcda04582d6a0f2162f7

target/rspamd/local.d/worker-controller.inc
polarathene: To be reachable outside of the container (_that includes another container performing as a reverse-proxy_), you have to bind to a common network AFAIK, so that should be fine.

That wouldn't require binding the port on the Docker host public networks either. I just wanted to verify the difference :+1:

polarathene:
> Thanks @polarathene for the quick review

Welcome :)

I will sleep now, but I can review again + approve in roughly 10 hours time.

polarathene (dismissed):
LGTM :+1:

target/scripts/startup/setup-stack.sh
```diff
@@ -599,18 +632,36 @@ function _setup_SRS
...
 
-function _setup_dkim
+function _setup_dkim_dmarc
 {
+  if [[ ${ENABLE_OPENDMARC} -eq 1 ]]
+  then
+    _log 'trace' "Adding OpenDMARC to Postfix's milters"
+
+    # shellcheck disable=SC2016
+    sed -i -E 's|^(smtpd_milters =.*)|\1 \$dmarc_milter|g' /etc/postfix/main.cf
```
casperklein, line 642: `smtpd_milters = ` expects a comma separated list. This `sed` statement may not be a problem, but the ones below are.

target/scripts/startup/setup-stack.sh
```diff
@@ -95,19 +102,45 @@ function _setup_rspamd
...
+    _log 'warn' 'Running rspamd at the same time as Amavis or SpamAssassin is discouraged'
   fi
 
   if [[ ${ENABLE_CLAMAV} -eq 1 ]]
   then
     _log 'debug' 'Rspamd will use ClamAV'
-    mv /etc/rspamd/local.d/disabled/antivirus.conf /etc/rspamd/local.d/antivirus.conf
+    sedfile -i -E 's|^(enabled).*|\1 = true;|g' /etc/rspamd/local.d/antivirus.conf
+    # RSpamd uses ClamAV's UNIX socket, and to be able to read it, it must be in the same group
+    usermod -a -G clamav _rspamd
```
casperklein, line 113: Just wondering, if that is the correct user? What up with the underscore?

polarathene: > What up with the underscore?

:man_shrugging: it's a user created by the installed package. It is correct and can be found in `/etc/passwd`.

casperklein: Ack, I didn't use rspamd yet and couldn't find that the user was created manually in one of the setup scripts.

target/scripts/startup/setup-stack.sh
polarathene: Good catch, so our tests are lacking an integration case that should have failed with Postfix being tested for DKIM and DMARC support working all together?

target/scripts/startup/setup-stack.sh
polarathene: For reference in [this comment](https://github.com/docker-mailserver/docker-mailserver/pull/3011#issuecomment-1399120252) on another PR, I have a collapsed section for a `setup-users.sh` script that details `/etc/passwd`, `/etc/group` and `/etc/shadow` with upstream links to each packages specific source lines for users and groups added. That includes the `_rspamd`.

target/scripts/startup/setup-stack.sh
georglauterbach: Where'd you get this? Postfix docs: "A list of Milter (mail filter) applications for new mail that arrives via the Postfix smtpd(8) server. Specify space or comma as separator. See the MILTER_README document for details."

target/scripts/startup/setup-stack.sh
georglauterbach: This is the correct user, as @polarathene pointed put :)

target/scripts/startup/setup-stack.sh
casperklein: Probably yes. Not sure if a space separated list is somehow valid, however all other options use commas and we should keep it uniformly.

georglauterbach: Doing it this way is really easy because the `sed` command is easy; using command, you have to handle way more logic about which one is last, which is first, and where to place the comma. So; a space-separated list is easier to maintains. Postfix does not care either way..

casperklein: This should handle this:

```bash
function _postfix_maincf_extend_option() {
	local option value

	# get current option and value(s)
	option=$(postconf "${1}")
	# extract current value(s)
	value=$(cut -d' ' -f3- <<< "$option")

	# extend value(s)
	if [ -z "${value}" ]; then
		postconf "${option}${2}"
	else
		postconf "${option}, ${2}"
	fi
}
```

polarathene: Minor nitpick if adding that function, `append` is probably better than `extend` :sweat_smile: 

Doesn't that make removal more fussy too though? I can't recall exactly without looking at the alias utilities, but I know this support was problematic for one format there, possibly it was without `,`.

As long as it's working as-is, I'm fine with the current `sed` approach. I'd sooner rather we delegate more config juggling to something not relying on shell scripts / commands since it's way easier to deal with array like data in a proper language :sweat_smile:

casperklein: > Moreover, a quick look in our Postfix `main.cf` shows that we in fact do not use commas everywhere. I see no issue here.

You are right! I just had a quick look at main.cf and missed that. I was confused because comma was used earlier [here](https://github.com/docker-mailserver/docker-mailserver/blob/cb8e336d256136b79cf58d6becd596cdb56fdb47/target/postfix/main.cf#L96).

casperklein:
```
smtpd_milters (default: empty)

    A list of Milter (mail filter) applications for
    new mail that arrives via the Postfix smtpd server.
    Specify space or comma as separator.
```

🎉

casperklein: Something was pretty weird here. I just refreshed the page and read this thread again. When I was answering, I did not see any of Georgs replys. That's why my answers seem to be unsuited 😖 Sorry for that. Also my last post.. just saw that Georg already posted that 🤦

casperklein (approved):

polarathene (approved):

target/scripts/startup/setup-stack.sh
georglauterbach: No worries :) I actually have no objections to making it uniform, just not in this PR :D

GitHub status is actually also showing degraded performance for Action, Codespaces & API requests, so something is going on..

I will resolve this thread though, because we resolved the matter.

PR #3018, opened 2023-01-22
tests: refactor 4 more tests

georglauterbach:
# Description

More rewrites. Note: `mail_smtponly.bats` seems to be really broken. I added TODOs.

## Type of change

## Checklist:

### DIFF
--- a/test/test-files/email-templates/smtp-only.txt
+++ b/test/test-files/email-templates/smtp-only.txt
@@ -1,4 +1,4 @@
-HELO mail.localhost
+HELO mail.example.test
 MAIL FROM: test@localhost
 RCPT TO: user2@external.tld
 DATA
--- a/test/tests/serial/mail_quotas_disabled.bats
+++ b/test/tests/serial/mail_quotas_disabled.bats
@@ -2,7 +2,7 @@ load "${REPOSITORY_ROOT}/test/helper/common"
 load "${REPOSITORY_ROOT}/test/helper/setup"
 
 BATS_TEST_NAME_PREFIX='[Quotas Disabled] '
-CONTAINER_NAME='dms-test_timezone'
+CONTAINER_NAME='dms-test_quotas-disabled'
 
 function setup_file() {
   _init_with_defaults
@@ -13,18 +13,16 @@ function setup_file() {
 function teardown_file() { _default_teardown ; }
 
 @test "(Dovecot) quota plugin is disabled" {
-  _run_in_container cat /etc/dovecot/conf.d/10-mail.conf
-  assert_success
-  refute_output --partial '$mail_plugins quota'
+  _run_in_container_bash_and_filter_output 'cat /etc/dovecot/conf.d/10-mail.conf'
+  refute_output --partial 'quota'
 
-  _run_in_container cat /etc/dovecot/conf.d/20-imap.conf
-  assert_success
-  refute_output --partial '$mail_plugins imap_quota'
+  _run_in_container_bash_and_filter_output 'cat /etc/dovecot/conf.d/20-imap.conf'
+  refute_output --partial 'imap_quota'
 
-  _run_in_container test -e /etc/dovecot/conf.d/90-quota.conf
+  _run_in_container_bash "[[ -f /etc/dovecot/conf.d/90-quota.conf ]]"
   assert_failure
 
-  _run_in_container test -e /etc/dovecot/conf.d/90-quota.conf.disab
+  _run_in_container_bash "[[ -f /etc/dovecot/conf.d/90-quota.conf.disab ]]"
   assert_success
 }
 
--- a/test/tests/serial/mail_quotas_disabled.bats
+++ b/test/tests/serial/mail_quotas_disabled.bats
@@ -1,43 +1,35 @@
-load "${REPOSITORY_ROOT}/test/test_helper/common"
+load "${REPOSITORY_ROOT}/test/helper/common"
+load "${REPOSITORY_ROOT}/test/helper/setup"
 
-# Test case
-# ---------
-# When ENABLE_QUOTAS is explicitly disabled (ENABLE_QUOTAS=0), dovecot quota must not be enabled.
-
+BATS_TEST_NAME_PREFIX='[Quotas Disabled] '
+CONTAINER_NAME='dms-test_timezone'
 
 function setup_file() {
-  local PRIVATE_CONFIG
-  PRIVATE_CONFIG=$(duplicate_config_for_container .)
-
-  docker run -d --name mail_no_quotas \
-    -v "${PRIVATE_CONFIG}":/tmp/docker-mailserver \
-    -v "$(pwd)/test/test-files":/tmp/docker-mailserver-test:ro \
-    -e ENABLE_QUOTAS=0 \
-    -h mail.my-domain.com -t "${NAME}"
-
-  wait_for_finished_setup_in_container mail_no_quotas
+  _init_with_defaults
+  local CUSTOM_SETUP_ARGUMENTS=(--env ENABLE_QUOTAS=0)
+  _common_container_setup 'CUSTOM_SETUP_ARGUMENTS'
 }
 
-function teardown_file() {
-  docker rm -f mail_no_quotas
-}
+function teardown_file() { _default_teardown ; }
 
-@test "checking dovecot: (ENABLE_QUOTAS=0) quota plugin is disabled" {
-  run docker exec mail_no_quotas /bin/sh -c "grep '\$mail_plugins quota' /etc/dovecot/conf.d/10-mail.conf"
+@test "(Dovecot) quota plugin is disabled" {
+  _run_in_container cat /etc/dovecot/conf.d/10-mail.conf
+  assert_success
+  refute_output --partial '$mail_plugins quota'
+
+  _run_in_container cat /etc/dovecot/conf.d/20-imap.conf
+  assert_success
+  refute_output --partial '$mail_plugins imap_quota'
+
+  _run_in_container test -e /etc/dovecot/conf.d/90-quota.conf
   assert_failure
 
-  run docker exec mail_no_quotas /bin/sh -c "grep '\$mail_plugins imap_quota' /etc/dovecot/conf.d/20-imap.conf"
-  assert_failure
-
-  run docker exec mail_no_quotas ls /etc/dovecot/conf.d/90-quota.conf
-  assert_failure
-
-  run docker exec mail_no_quotas ls /etc/dovecot/conf.d/90-quota.conf.disab
+  _run_in_container test -e /etc/dovecot/conf.d/90-quota.conf.disab
   assert_success
 }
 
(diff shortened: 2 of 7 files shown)


test/tests/serial/mail_smtponly.bats
```diff
@@ -1,64 +1,50 @@
...
-  _reload_postfix mail_smtponly
+# TODO: needs complete rework when proper DNS container is running for tests
+@test "sending mail should work" {
+  skip 'TODO: This test is absolutely broken and needs reworking!'
 
-  wait_for_smtp_port_in_container mail_smtponly
-  run docker exec mail_smtponly /bin/sh -c "nc 0.0.0.0 25 < /tmp/docker-mailserver-test/email-templates/smtp-only.txt"
+  # the value `no` is not even valid for `smtp_host_lookup`
+  # `smtp_host_lookup` seems to be deprecated too
+  _run_in_container postconf smtp_host_lookup=no
```
polarathene, line 35: [`smtp_host_lookup`](https://www.postfix.org/postconf.5.html#smtp_host_lookup) is not deprecated, but [`disable_dns_lookups`](https://www.postfix.org/postconf.5.html#disable_dns_lookups) is in favor of [`smtp_dns_support_level`](https://www.postfix.org/postconf.5.html#smtp_dns_support_level).

`smtp_host_lookup` is to configure how that lookup should be performed. If we want to opt-out we can disable with `smtp_dns_support_level=disabled`. That disables MX record lookups, but still defers regular hostname lookups to "native" as the Postfix docs describe it.

Chances are this isn't relevant at all to the test. I'll look through the history to see if there's any context, but it's doubtful.

---

<details>
<summary>References for revisiting this test in future</summary>

Based on the test name. The general approach in tests is to just send mail to the same DMS instance (_or sometimes a 2nd instance_), wait for the mail queue to empty and then check the mail directory it should have delivered to for a local account.

This test is only interested in if a mail was sent. `smtp-delivery.bats` has a bunch of test cases doing that, they look like this:

```bash
_run_in_container grep 'to=<user1~test@localhost.localdomain>, orig_to=<alias1~test@localhost.localdomain>' /var/log/mail/mail.log
assert_success
assert_output --partial 'status=sent'
_should_output_number_of_lines 1
```

And here's another from `mail_lmtp_ip.bats`:

```bash
@test "delivers mail to existing account" {
  _wait_for_smtp_port_in_container

  # Send a test mail:
  _run_in_container_bash "nc 0.0.0.0 25 < /tmp/docker-mailserver-test/email-templates/existing-user1.txt"
  assert_success

  # Verify delivery was successful, log line should look similar to:
  # postfix/lmtp[1274]: 0EA424ABE7D9: to=<user1@localhost.localdomain>, relay=127.0.0.1[127.0.0.1]:24, delay=0.13, delays=0.07/0.01/0.01/0.05, dsn=2.0.0, status=sent (250 2.0.0 <user1@localhost.localdomain> ixPpB+Zvv2P7BAAAUi6ngw Saved)
  local MATCH_LOG_LINE='postfix/lmtp.* status=sent .* Saved)'
  run timeout 60 docker exec "${CONTAINER_NAME}" bash -c "tail -F /var/log/mail/mail.log | grep --max-count 1 '${MATCH_LOG_LINE}'"
  assert_success
}
```

---

Since Dovecot is not available and no other mail storage is intended in this mode, a 2nd container to receive the mail probably makes more sense :sweat_smile: 

The `fail2ban.bats` test seems to handle this for SMTP auth by getting the container IP to use with `nc`, which would probably work?:

```bash
CONTAINER1_IP=$(_get_container_ip "${CONTAINER1_NAME}")
_run_in_container_explicit "${CONTAINER2_NAME}" bash -c "nc ${CONTAINER1_IP} 25 < /tmp/docker-mailserver-test/auth/smtp-auth-login-wrong.txt"
```

</details>

I'm fine with the `skip` you've added for now, and addressing this afterwards with a follow-up PR.

For now these lines at least don't seem to provide any value as you've documented (_thanks for pointing that out btw_ :grin: )

```suggestion
```

test/tests/serial/mail_smtponly.bats
```diff
@@ -1,64 +1,50 @@
...
   assert_success
-  run docker exec mail_smtponly /bin/sh -c 'grep -cE "to=<user2\@external.tld>.*status\=sent" /var/log/mail/mail.log'
-  [[ ${status} -ge 0 ]]
-}
 
-#
-# PERMIT_DOCKER=network
-#
+  _reload_postfix
+  _wait_for_smtp_port_in_container
```
polarathene, line 39: If dropping `smtp_host_lookup` in this PR, the reload isn't necessary. The SMTP port should already be available as it's waited on in `setup_file()`.

```suggestion
```

test/tests/serial/mail_smtponly.bats
```diff
@@ -1,64 +1,50 @@
...
 
-@test "checking PERMIT_DOCKER=network: opendmarc/opendkim config" {
-  run docker exec mail_smtponly /bin/sh -c "cat /etc/opendmarc/ignore.hosts | grep '172.16.0.0/12'"
+  # it looks as if someone tries to send mail to another domain outside of DMS
+  _run_in_container_bash "nc 0.0.0.0 25 < /tmp/docker-mailserver-test/email-templates/smtp-only.txt"
```
polarathene, line 42: So test this isn't meant to deliver mail to the DMS container, but relay it externally?

I'm not too familiar with the feature, if accounts are tied to Dovecot for authentication, then it's also trying to submit mail without auth (_which I think it is permitted to bypass by being sent internally like this_).

Not much we can do about fixing that any time soon then :sweat_smile:

test/tests/serial/mail_special_use_folders.bats
```diff
@@ -1,40 +1,38 @@
-load "${REPOSITORY_ROOT}/test/test_helper/common"
+load "${REPOSITORY_ROOT}/test/helper/common"
+load "${REPOSITORY_ROOT}/test/helper/setup"
 
-setup_file() {
-  local PRIVATE_CONFIG
-  PRIVATE_CONFIG=$(duplicate_config_for_container .)
+BATS_TEST_NAME_PREFIX='[Special Use Folders] '
+CONTAINER_NAME='dms-test_special_use_folders'
```
polarathene, line 5:
```suggestion
CONTAINER_NAME='dms-test_special-use-folders'
```

test/tests/serial/mail_special_use_folders.bats
```diff
@@ -1,40 +1,38 @@
...
   assert_success
 
-  # shellcheck disable=SC2016
-  repeat_until_success_or_timeout 30 docker exec mail_special_use_folders /bin/sh -c '[ $(ls /var/mail/localhost.localdomain/user1/new | wc -l) -eq 1 ]'
+  _count_files_in_directory_in_container /var/mail/localhost.localdomain/user1/new 1
 }
 
-@test "checking special-use folders not yet created" {
-  run docker exec mail_special_use_folders /bin/bash -c "ls -A /var/mail/localhost.localdomain/user1 | grep -E '.Drafts|.Sent|.Trash' | wc -l"
+@test "(IMAP) special-use folders are not yet created" {
```
polarathene, line 23:
```suggestion
@test "(IMAP) special-use folders should not exist yet" {
```

test/tests/serial/mail_special_use_folders.bats
```diff
@@ -1,40 +1,38 @@
...
   assert_success
-  assert_output 0
+  refute_output --partial '.Drafts'
+  refute_output --partial '.Sent'
+  refute_output --partial '.Trash'
 }
 
-@test "checking special-use folders available in IMAP" {
-  run docker exec mail_special_use_folders /bin/sh -c "nc -w 8 0.0.0.0 143 < /tmp/docker-mailserver-test/nc_templates/imap_special_use_folders.txt | grep -E 'Drafts|Junk|Trash|Sent' | wc -l"
+@test "(IMAP) special-use folders are available" {
```
polarathene, line 31:
```suggestion
@test "(IMAP) special-use folders should be created when necessary" {
```

polarathene (changes requested):
Nice work! :grinning: 

I almost started working on `mail_smtponly.bats` myself, so glad I checked GitHub first :sweat_smile:

test/test-files/email-templates/smtp-only.txt
```diff
@@ -1,4 +1,4 @@
-HELO mail.localhost
+HELO mail.example.test
```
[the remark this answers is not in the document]
georglauterbach: 👍🏼

test/tests/serial/mail_quotas_disabled.bats
```diff
@@ -1,43 +1,35 @@
...
+@test "(Dovecot) quota plugin is disabled" {
+  _run_in_container cat /etc/dovecot/conf.d/10-mail.conf
+  assert_success
+  refute_output --partial '$mail_plugins quota'
 
-  run docker exec mail_no_quotas /bin/sh -c "grep '\$mail_plugins imap_quota' /etc/dovecot/conf.d/20-imap.conf"
-  assert_failure
+  _run_in_container cat /etc/dovecot/conf.d/20-imap.conf
+  assert_success
+  refute_output --partial '$mail_plugins imap_quota'
```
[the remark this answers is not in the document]
georglauterbach: I will have a look :)

test/tests/serial/mail_smtponly.bats
georglauterbach: I don't really know why this is the case, but the RCPT is for `@external.tld` .. I actually don't know if this makes sense either. Maybe it is useless and we can just deliver mail with Postfix to a local account and test the functionality this way too?

test/tests/serial/mail_quotas_disabled.bats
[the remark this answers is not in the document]
georglauterbach: Took some time and thinking but I think I came up with something quite nice: I made the test more agressive asserting partial output only for "quota" / "imap_quota", but I stripped the comments and blank lines so they don't interfere. I then thought this functionality (running a command and filtering the output) is actually one that we need all the time, I added a helper function.

Corresponding commit:

test/tests/serial/mail_smtponly.bats
polarathene: Can DMS receive mail and store it without Dovecot when using `SMTP_ONLY=1`?

test/tests/serial/mail_smtponly.bats
polarathene: Keeping this open for reference

polarathene (approved):

test/tests/serial/mail_smtponly.bats
georglauterbach: Probably not, but I have not tried it TBH :D I'g go about this way: when all tests are refactored so we can run them in parallel, I'd then (after or before the DNS updates I don't know) tackle the left-over TODOs that we added when refactoring tests. I mean, this test was not properly working all the time, and we have enough to do when it comes to reworking the tests for now :)

casperklein (approved):

PR #3020, opened 2023-01-22
quality-of-life: improve the `clean` recipe (don't require `sudo` anymore)

georglauterbach:
# Description

This bugged me for a long time: we need `sudo` when running `make clean`, which is undesirable as you can image. I implemented a proper solution:

We run a small container (Alpine) and mount the files we need to remove into the container. This has the advantage that users that run Docker with an unprivilged user now don't have to type their sudo-password anymore and users that run Docker by default with password will need to type in the passward, but they need to do this anyways because of the command before.

---

To explain what I did: I introduced an array where I save files in. The `.gitignore` is still read, but the files read in the loop are now saved in the array. I adjust the path, prefixing `/mnt` and removing the `test` prefix from the file path before so the path is correct inside the container. I can then just mount the whole test directory at `/mnt` inside the container and run an `rm -rf` inside the container on the `FILES` array. Done. Enjoy :D

## Type of change

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,9 +32,9 @@ generate-accounts: ALWAYS_RUN
 # `.gitignore`: Remove `test/duplicate_configs` and files copied via `make generate-accounts`
 clean: ALWAYS_RUN
 	-@ while read -r LINE; do CONTAINERS+=("$${LINE}"); done < <(docker ps -qaf name='^(dms-test|mail)_.*') ; \
-	for CONTAINER in "$${CONTAINERS[@]}"; do docker rm -f "$${CONTAINER}"; done
+		for CONTAINER in "$${CONTAINERS[@]}"; do docker rm -f "$${CONTAINER}"; done
 	-@ while read -r LINE; do [[ $${LINE} =~ test/.+ ]] && FILES+=("/mnt$${LINE#test}"); done < .gitignore ; \
-	docker run --rm -v "$(REPOSITORY_ROOT)/test/:/mnt" docker.io/alpine:3.17.1 ash -c "rm -rf $${FILES[@]}"
+		docker run --rm -v "$(REPOSITORY_ROOT)/test/:/mnt" alpine ash -c "rm -rf $${FILES[@]}"
 
 # -----------------------------------------------
 # --- Tests  ------------------------------------
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ generate-accounts: ALWAYS_RUN
 # `.gitignore`: Remove `test/duplicate_configs` and files copied via `make generate-accounts`
 clean: ALWAYS_RUN
 	-@ for CONTAINER in $$(docker ps -a --filter name='^dms-test_.*|^mail_.*|^hadolint$$|^eclint$$|^shellcheck$$' | sed 1d | cut -f 1-1 -d ' '); do docker rm -f $${CONTAINER}; done
-	-@ declare -a FILES ; while read -r LINE; do [[ $${LINE} =~ test/.+ ]] && FILES+=("/mnt$${LINE#test}"); done < .gitignore ; docker run --rm --volume "$(REPOSITORY_ROOT)/test/:/mnt" docker.io/alpine:3.17.1 rm -rf "$${FILES[@]}"
+	-@ declare -a FILES ; while read -r LINE; do [[ $${LINE} =~ test/.+ ]] && FILES+=("/mnt$${LINE#test}"); done < .gitignore ; docker run --rm --volume "$(REPOSITORY_ROOT)/test/:/mnt" docker.io/alpine:3.17.1 ash -c "rm -rf '$${FILES[@]}'"
 
 # -----------------------------------------------
 # --- Tests  ------------------------------------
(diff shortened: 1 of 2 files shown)


polarathene (approved):
Not entirely fond of the approach (albeit clever! :stuck_out_tongue: )

It's definitely been something that has annoyed me too. My plan was to wait until tests are all migrated to the common container method, and then just leverage `docker cp` (_or a `Dockerfile` that extends `mailserver-testing:ci` with `COPY`_), or an anonymous volume..?

Although my approach is still a bit more work to sort out. We're reasonably close to being able to mount as `:ro` in some cases too, but need to handle a few startup scripts that currently perform some sanitation pre-processing IIRC.

georglauterbach:
> It's definitely been something that has annoyed me too. My plan was to wait until tests are all migrated to the common container method, and then just leverage `docker cp` (_or a `Dockerfile` that extends `mailserver-testing:ci` with `COPY`_), or an anonymous volume..?

Sounds good! Until we have your solution in-place, I think this PR is well worth it :)

polarathene (approved):

casperklein:
Two thoughts:

1. Add some line breaks to make it more readable?
2. `declare -a FILES` is not needed.

georglauterbach:
> 1. Add some line breaks to make it more readable?

Yeah, I can do that :)

> 2. `declare -a FILES` is not needed.

I will remove it.

---

Just FYI: #3016 is important for me, so if you think about which PR to review next, it'd be nice of you picked this one ❤️

casperklein:
PS: you only need `declare`, when you want to use an associative array (`declare -A`).

> Just FYI: #3016 is important for me, so if you think about which PR to review next, it'd be nice of you picked this one ❤️

Ack 👍

Makefile
```diff
@@ -31,8 +31,10 @@ generate-accounts: ALWAYS_RUN
 # `docker ps`:  Remove any lingering test containers
 # `.gitignore`: Remove `test/duplicate_configs` and files copied via `make generate-accounts`
 clean: ALWAYS_RUN
-	-@ for CONTAINER in $$(docker ps -a --filter name='^dms-test_.*|^mail_.*|^hadolint$$|^eclint$$|^shellcheck$$' | sed 1d | cut -f 1-1 -d ' '); do docker rm -f $${CONTAINER}; done
-	-@ while read -r LINE; do [[ $${LINE} =~ test/.+ ]] && sudo rm -rf $${LINE}; done < .gitignore
+	-@ while read -r LINE; do CONTAINERS+=("$${LINE}"); done < <(docker ps -qaf name='^(dms-test|mail)_.*') ; \
+	for CONTAINER in "$${CONTAINERS[@]}"; do docker rm -f "$${CONTAINER}"; done
+	-@ while read -r LINE; do [[ $${LINE} =~ test/.+ ]] && FILES+=("/mnt$${LINE#test}"); done < .gitignore ; \
+	docker run --rm -v "$(REPOSITORY_ROOT)/test/:/mnt" docker.io/alpine:3.17.1 ash -c "rm -rf $${FILES[@]}"
```
casperklein, line 37:
```suggestion
	-@ while read -r LINE; do \
		CONTAINERS+=("$${LINE}"); \
	done < <(docker ps -qaf name='^(dms-test|mail)_.*') ; \
	for CONTAINER in "$${CONTAINERS[@]}"; do \
		docker rm -f "$${CONTAINER}"; \
	done
	-@ while read -r LINE; do \
		[[ $${LINE} =~ test/.+ ]] && \
		FILES+=("/mnt$${LINE#test}"); \
	done < .gitignore ; \
	docker run --rm -v "$(REPOSITORY_ROOT)/test/:/mnt" docker.io/alpine:3.17.1 ash -c "rm -rf $${FILES[@]}"
```

casperklein (approved):

Makefile
polarathene: At this point I'm wondering if it should just be a small script that gets called outside of the Makefile :sweat_smile: 

I don't know about pinning the alpine version, the minor changes every few months, we're not likely to keep that maintained, especially with the point releases. Perhaps just omit the tag for implicit `:latest`? Shouldn't cause any issues for this usage.

polarathene (approved):

Makefile
casperklein: Good point about the version pinning. I don't think the behaviour of `rm` will significantly change with newer versions 😆 Sticking to `:latest` should be fine.

georglauterbach: Definitely keep it in the Makefile, no outside script. And the version pinning is just so Docker does not pull a new version ever time the image is updated.

polarathene: > Docker does not pull a new version ever time the image is updated.

It shouldn't for `:latest`, that's often a drawback due to newer Docker users assuming it always gets the `:latest` rather than the latest digest at the time it was originally pulled, then uses the local `:latest` tagged image from then on (_unless explicitly pulling again_).

polarathene: I'm not overly fussed either way due to usage, but the version pinning here isn't serving much purpose.

I know in our tests we have a very outdated version pinning for OpenLDAP container that I'll sort out when I get around to working on the LDAP test.

Our CI may want to bump from Ubuntu 20.04 LTS to 22.04 LTS at some point, but no rush there as it shouldn't offer any advantage to us I think. Github keeps the pre-installed software we care about updated (eg Docker Engine).

georglauterbach: There you go: [8f006ba](https://github.com/docker-mailserver/docker-mailserver/pull/3020/commits/8f006bab40371fb7e200b9934d5854c82a2b5377)

Can we please get over with this?

polarathene (approved):
That commit message :joy: I know the feeling! :stuck_out_tongue:

casperklein (approved):

Issue #401, opened 2016-12-17
/etc/default/locale: No such file or directory


tomav:
```
Dec 17 18:00:01 mail CRON[4709]: pam_env(cron:session): Unable to open env file: /etc/default/locale: No such file or directory
Dec 17 18:00:01 mail CRON[4709]: pam_unix(cron:session): session opened for user root by (uid=0)
Dec 17 18:02:20 mail CRON[4709]: pam_unix(cron:session): session closed for user root
Dec 17 18:17:01 mail CRON[4742]: pam_env(cron:session): Unable to open env file: /etc/default/locale: No such file or directory
Dec 17 18:17:01 mail CRON[4742]: pam_unix(cron:session): session opened for user root by (uid=0)
Dec 17 18:17:01 mail CRON[4742]: pam_unix(cron:session): session closed for user root
Dec 17 18:18:01 mail CRON[4745]: pam_env(cron:session): Unable to open env file: /etc/default/locale: No such file or directory
Dec 17 18:18:01 mail CRON[4745]: pam_unix(cron:session): session opened for user amavis by (uid=0)
Dec 17 18:18:04 mail CRON[4745]: pam_unix(cron:session): session closed for user amavis
Dec 17 19:17:01 mail CRON[4833]: pam_env(cron:session): Unable to open env file: /etc/default/locale: No such file or directory
Dec 17 19:17:01 mail CRON[4833]: pam_unix(cron:session): session opened for user root by (uid=0)
Dec 17 19:17:01 mail CRON[4833]: pam_unix(cron:session): session closed for user root
Dec 17 20:17:01 mail CRON[4943]: pam_env(cron:session): Unable to open env file: /etc/default/locale: No such file or directory
Dec 17 20:17:01 mail CRON[4943]: pam_unix(cron:session): session opened for user root by (uid=0)
Dec 17 20:17:01 mail CRON[4943]: pam_unix(cron:session): session closed for user root
```

PR #3032, opened 2023-01-27
chore: Remove wrapper script for fail2ban service

polarathene:
# Description

Main motivation for this was avoiding the `sleep` approach in the wrapper script if we don't actually need to rely on it.

I've additionally removed the `/var/log/auth.log` references in the project (_excluding a test case in `tests.bats` that's probably not that useful and will be removed later_), and added two optional packages for Fail2Ban.

Commit messages do provide context, but more verbose details have been provided below.

### Supervisord starts `fail2ban-server` directly now

Fail2Ban docs for the [CLI (only lists for v0.8)](http://www.fail2ban.org/wiki/index.php/MANUAL_0_8#Server) or [Github master branch](https://github.com/fail2ban/fail2ban/blob/master/man/fail2ban-server.1). This removes the need to manually manage removal of the socket at startup (`-x`) and run in foreground (`-f`), letting us drop some scripts.

I have compared to the `fail2ban-client start` command that runs the `fail2ban-server` process with slightly different options (such as `--async`), but according to the docs that is for internal usage only and not advised to run the server with directly, so I've avoided matching those additional args.

### `/var/log/auth.log`

- April 2016: In the [start-up scripts for Fail2Ban](https://github.com/docker-mailserver/docker-mailserver/commit/d6f9ffe9a6eefdfbb8a8d78bd3fb6f99dbda9575) (_direct commit to master, no associated PR, does reference [v2 discussion](https://github.com/docker-mailserver/docker-mailserver/issues/109), but there is no mention about the change there_).
- Dec 2016: Added into the [Dockerfile to avoid logging an error](https://github.com/docker-mailserver/docker-mailserver/pull/402) (_references [an issue regarding CRON + PAM unix error log](https://github.com/docker-mailserver/docker-mailserver/issues/401) for the `root` user_) "Unable to open env file: /etc/default/locale".
- On Fail2Ban Github, I did notice a [Dec 2022 discussion mentioning the file being monitored](https://github.com/fail2ban/fail2ban/discussions/3426). Presently I've only seen the file created and populated with content from SASLAUTHD (PAM) initializing. I'm not too familiar with what else would write to this `auth.log` file or when in DMS, but assume they would create the file if it doesn't exist?
- I am aware that it's typically [associated to `rsyslogd` / `syslog` to manage it](https://stackoverflow.com/questions/22526016/docker-container-sshd-logs), and I've most commonly seen it referenced for SSHD. But also that the `rsyslog` package has some differences on Debian from the Ubuntu one that we had installed (_at least with default user/group, possibly configs?_).

### Added Fail2Ban optional dependencies

`pyinotify` and `dnspython` are listed on the [F2B Github README](https://github.com/fail2ban/fail2ban#installation), both are minimal overhead to image weight.

While we presently have the `gamin` package installed, and Fail2Ban offers it as one of the notification methods, it can only do so with a Python 2 package no longer available on Debian as Python 2 is EOL now. Besides [Fail2Ban devs have noted it's performance was very bad](https://github.com/fail2ban/fail2ban/issues/1109#issuecomment-121586557).

I'm not entirely sure about the `pyinotify` package. It may be more beneficial than constant polling of log files, but [the project has not been maintained since June 2015](https://github.com/seb-m/pyinotify/issues/202#issue-1420791542). I have looked at the [`python3-pyinotify` debian package](https://packages.debian.org/bullseye/python3-pyinotify) where the [Gitlab source](https://salsa.debian.org/python-team/packages/pyinotify) appears to maintain a fork of the upstream project.

The [`dnspython`](https://www.dnspython.org/) package may also be of some value? (_It is [actively maintained](https://github.com/rthalley/dnspython)_) I do recall there was a PR a while back to disable DNS lookups due to NFTables/IPTables (possibly with F2B?) not resolving DNS requests properly? Having this package may have resolved that :man_shrugging:

If neither seem worthwhile, let me know and I can drop them. I only looked into this when viewing Fail2Ban logs warning about failure to initialize `gamin` and `pyinotify` backends, so it falls back to polling strategy. The default config is `backend=auto` AFAIK, hence no change beyond installing `python3-pyinotify` is required. I have noticed the logs for that backend is not as nice as the polling ones. We could drop it since there has been no complaints with the polling backend currently used?

## Background history for the wrapper script

A [PR from July 2017](https://github.com/docker-mailserver/docker-mailserver/pull/676) introduced `supervisord` to manage processes. It added wrapper scripts for Postfix and Fail2Ban citing that they could not be run in foreground mode, and these wrapper scripts were necessary to support these services due to their usage of multiple processes.

That doesn't seem to align to the current state of the image though? It'd seem we'd want more wrapper scripts then:

<details>
<summary>Click to view process tree</summary>

```
/usr/bin/dumb-init -- supervisord -c /etc/supervisor/supervisord.conf
/usr/bin/python3 /usr/bin/supervisord -c /etc/supervisor/supervisord.conf
 \_ /bin/bash /usr/local/bin/start-mailserver.sh
 |   \_ tail -Fn 0 /var/log/mail/mail.log
 \_ /usr/sbin/cron -f
 \_ /usr/sbin/rsyslogd -n
 \_ /usr/sbin/dovecot -F -c /etc/dovecot/dovecot.conf
 |   \_ dovecot/anvil
 |   \_ dovecot/log
 |   \_ dovecot/config
 \_ /usr/sbin/opendkim -f
 |   \_ /usr/sbin/opendkim -f
 \_ /usr/sbin/opendmarc -f -p inet:8893@localhost -P /var/run/opendmarc/opendmarc.pid
 \_ postgrey --inet=127.0.0.1:10023 --syslog-facility=mail --delay=300 --max-age=35 --auto-whitelist-clients=5 --greylist-text=Delayed by Postgrey
 \_ /usr/sbin/saslauthd -d -a pam -O
 |   \_ /usr/sbin/saslauthd -d -a pam -O
 |   \_ /usr/sbin/saslauthd -d -a pam -O
 |   \_ /usr/sbin/saslauthd -d -a pam -O
 |   \_ /usr/sbin/saslauthd -d -a pam -O
 \_ /usr/bin/python3 /usr/bin/fail2ban-server -xf start
 \_ /usr/bin/fetchmail -f /etc/fetchmailrc.d/fetchmail-1.rc -v --nodetach --daemon 300 -i /var/lib/fetchmail/.fetchmail-UIDL-cache --pidfile /var/run/fetchmail/fetchmail-1.pid
 \_ /usr/bin/fetchmail -f /etc/fetchmailrc.d/fetchmail-2.rc -v --nodetach --daemon 300 -i /var/lib/fetchmail/.fetchmail-UIDL-cache --pidfile /var/run/fetchmail/fetchmail-2.pid
 \_ /usr/sbin/clamd -c /etc/clamav/clamd.conf
 \_ /bin/bash /usr/local/bin/check-for-changes.sh
 |   \_ sleep 2
 \_ /usr/sbin/amavisd-new (master)
 |   \_ /usr/sbin/amavisd-new (virgin child)
 |   \_ /usr/sbin/amavisd-new (virgin child)
 \_ /bin/bash /usr/local/bin/postfix-wrapper.sh
     \_ sleep 5
/usr/sbin/postsrsd -f 10001 -r 10002 -d example.test -s /etc/postsrsd.secret -a = -n 4 -N 4 -u postsrsd -p /var/run/postsrsd.pid -c /var/lib/postsrsd -l 127.0.0.1 -D -X
/usr/lib/postfix/sbin/master
 \_ pickup -l -t fifo -u -c -o content_filter= -o receive_override_options=no_header_body_checks
 \_ qmgr -l -t unix -u
```

</details>

**I don't think the wrapper script is providing any actual value?** If it is, could we get a proper test for that, and confirmation that none of the other processes above are at risk? (_the list is output from this PRs changes applied_)

The referenced PR did originally run fail2ban in foreground mode, but [in a later commit](https://github.com/docker-mailserver/docker-mailserver/pull/676/commits/c3da995b99b34e6d6346725676cee25f6278c89e#diff-364dacb93654cfce6e7db25d1265023f720ca8719a601c86ec24f959ee158eae) of that PR changed it to the wrapper script approach instead, with the context of:

> You cannot start fail2ban in some foreground mode, and it's more or less important that docker doesn't kill fail2ban and its children if you stop the container.

Clearly you could start it in foreground mode? We also had a [stop grace period added many years later](https://github.com/docker-mailserver/docker-mailserver/pull/1945) after that PR. So I'm not sure that concern is valid anymore. **It seems safe to remove the wrapper script?**

**UPDATE:** I've since realized that PR was originally based off [this PR](https://github.com/docker-mailserver/docker-mailserver/pull/631) (_which the merged PR referenced as an "issue" it resolved, rather than the [actual issue / FR](https://github.com/docker-mailserver/docker-mailserver/issues/434)_) which had the `fail2ban-server -f` usage without the wrapper script. The associated issue from Dec 2016 [references a serverfault thread](https://serverfault.com/questions/608069/managing-daemons-with-supervisor-no-foreground-mode-available) which is where the wrapper script appears to have been adapted from in the [July 2017 PR that was merged](https://github.com/docker-mailserver/docker-mailserver/pull/676).

## Type of change

## Checklist:


georglauterbach (approved):
I actuall wanted to tackle this for a long time! Outstanding work, very nice!

casperklein (approved):

PR #3038, opened 2023-01-29
ci: move tests than can be run in parallel now

georglauterbach:
# Description

Move & renaming of tests that can now run in parallel. I tried to group tests a bit, and I tried to evenly balance the test across the sets. All sets _should_ now take _more or less_ the same amount of time to complete. 

## Type of change

## Checklist:


georglauterbach:
[`tls_ciperlists.bats`](https://github.com/docker-mailserver/docker-mailserver/blob/ci/move-tests/test/tests/parallel/set2/tls_cipherlists.bats) still takes 6 minutes alone to complete 🤣

polarathene:
> [`tls_ciperlists.bats`](https://github.com/docker-mailserver/docker-mailserver/blob/ci/move-tests/test/tests/parallel/set2/tls_cipherlists.bats) still takes 6 minutes alone to complete

It should average around 5 minutes (300 sec) in my experience, at least on my 4 core/thread i5-6500 :sweat_smile: 

I have tried a variety of things to speed it up, and noticed that single tests could be sped up by 50%, but once running in parallel (_`testssl.sh` mode, not parallel bats which I think I observed before as slowing it down even further_), the difference is much less.

[Changing to a different base image for testssl](https://github.com/drwetter/testssl.sh/issues/2299) showed about 40 seconds or more improvement for me, although I've also been observing a smaller improvement once the base image upgrades to a newer version of Alpine, I'm not 100% sure but it appears to be due to `musl` vs `glibc` for the most part. 

We can potentially look at splitting the test case up into multiple files grouped in a folder instead (_with the methods extracted out into a helper_) for bats to run in parallel (_each test case runs `testssl.sh` in parallel on 7 ports though_). Alternatively splitting the `testssl.sh` parallel tests from 7 at once into to two separate runs of 3 and 4 might be another option. I've observed 10-15 seconds for a port to be tested locally, vs around 40-ish for all 7 in parallel, so it's hard to say they'd be much improvement :sweat_smile: 

For CI at least, we can always leverage more runners to spread out those test cases :thinking: 

Another option is to only run those tests only when relevant files would affect it for PRs, while master branch could run on each merge and raise an issue if the test failed.

polarathene (approved):
Are we making snake_case for file names a convention? I noticed we were using kebab-case before.

Thanks for tackling this :)

georglauterbach:
> Are we making snake_case for file names a convention? I noticed we were using kebab-case before.
>
> Thanks for tackling this :)

I'd go for snake case, indeed :)

---

About the current state about the single 6min test: I actually don't mind because it's working fine. Just noting that there is room for future optimisations. I'd go for splitting the single file into multiple too :)

casperklein (approved):

PR #3040, opened 2023-01-29
postfix header filter: correct the casing for Mime vs. MIME

georglauterbach:
# Description

This occured to me when looking at my Rspamd logs. Messages I sent were given a penalty (0.5) for not using the correct case (uppercase).

## Type of change

## Checklist:

### DIFF
--- a/target/postfix/sender_header_filter.pcre
+++ b/target/postfix/sender_header_filter.pcre
@@ -2,7 +2,7 @@
 /^\s*Received:.*amavisd-new/        IGNORE
 /^\s*X-Originating-IP:/             IGNORE
 /^\s*X-Mailer:/                     IGNORE
-/^\s*Mime-Version: 1.0.*/           REPLACE Mime-Version: 1.0
+/^\s*Mime-Version: 1.0.*/           REPLACE MIME-Version: 1.0
 /^\s*User-Agent/                    IGNORE
 /^\s*X-Enigmail/                    IGNORE
 /^\s*X-Mailer/                      IGNORE


polarathene (approved):

casperklein:
I am wondering, what the use case for this line is? Any idea?

casperklein (approved):

polarathene:
> If it doesn't serve any purpose (nowadays?), we could remove it.

`git blame` shows it belonging to a [privacy related feature PR](https://github.com/docker-mailserver/docker-mailserver/pull/708), although they have not created that as a separate feature that is applied during startup (_which tends to be better for maintaining even if it's not toggled by an ENV_), and instead opted to just have it always applied. We have a single test case only for it, but it does not test anything else beyond the user-agent :man_shrugging:

georglauterbach:
I will do a thorough analysis of the whole file. For now, I will merge this and correct the casing though. Expect an answer this week, maybe even today :)

georglauterbach:
How come? Can you give an example?

georglauterbach:
I understand you problem, but these defaults were and are working rather nicely for a majority of users. Overwriting the config is actually the "elegant" solution here. May I suggest you use `user-patches.sh`; it seems appropriate for what you want to do.

mpldr:
On Thu 15 Feb 2024 10:46:49, Georg Lauterbach wrote:
> I understand you problem, but these defaults were and are working rather nicely for a majority of users. Overwriting the config is actually the "elegant" solution here. May I suggest you use `user-patches.sh`; it seems appropriate for what you want to do.

Thanks, will do. Though my user-patches.sh is
git pull --commit --autostash

(worked quite effortless so far)

-- 
Moritz Poldrack
https://moritz.sh

> Sanitized for your protection.

polarathene:
> The mail that was sent

Was this sent by you through DMS on port 587/465?

It seems this privacy feature is configured in a way that it may try modify at multiple points. Unclear to me if one of those is prior to DKIM signing.

It would be good to know if this is the case, I agree it shouldn't be messing with signed content like that breaking the `bh` (_while `b` in the signature should be valid AFAIK_).

---

> this is too broad and touches the headers sent for MIME parts as well

It seems that when you set `header_checks`, it will implicitly be the default for a separate setting [`mime_header_checks`](https://www.postfix.org/postconf.5.html#mime_header_checks). Although from reading the [header_checks config docs](https://www.postfix.org/header_checks.5.html), it doesn't seem to suggest it'll affect the body content in anyway as you mentioned later.

I don't have time to investigate but wonder if this is what is actually affecting you? Perhaps try disabling that, it'll need to be done via `postfix-master.cf` override (_although `user-patches.sh` could also handle it_).

```
sender-cleanup/unix/header_checks=
```

That might work, otherwise to properly remove the entry from config instead of setting it to an empty value, you could try `user-patches.sh` with [`postconf`](https://www.postfix.org/postconf.1.html):

```bash
postconf -PX sender-cleanup/unix/header_checks
```

Inspect `/etc/postfix/master.cf` and you should see that this `-o` line has been removed:

https://github.com/docker-mailserver/docker-mailserver/blob/79a9656f4876e50264060eafbc066287a036a045/target/postfix/master.cf#L53

If that doesn't resolve it, then I assume it's the related `header_checks` config for outbound traffic [`smtp_header_checks`](https://www.postfix.org/postconf.5.html#smtp_header_checks), which is where I assume the DKIM signing would occur?:

https://github.com/docker-mailserver/docker-mailserver/blob/79a9656f4876e50264060eafbc066287a036a045/target/postfix/main.cf#L114-L118

Showing that `header_checks` is also configured in `main.cf` for some reason. Not sure why as that'd apply to any inbound mail from third-parties too? 🤷‍♂️ (_while it is possible to submit mail to be sent outbound by sending to port 25 if `PERMIT_DOCKER` raises trust, it's not the correct approach, I know in the past we wrongfully supported SMTP auth on port 25 which would also have permitted it I think_)

The [original issue motivating this privacy focused addition](https://github.com/docker-mailserver/docker-mailserver/issues/681) doesn't discuss this much, nor the PR. I'm more inclined that the contribution has flaws.

DKIM milter is configured:

https://github.com/docker-mailserver/docker-mailserver/blob/79a9656f4876e50264060eafbc066287a036a045/target/scripts/startup/setup.d/dmarc_dkim_spf.sh#L16-L21

https://github.com/docker-mailserver/docker-mailserver/blob/79a9656f4876e50264060eafbc066287a036a045/target/scripts/startup/setup.d/security/rspamd.sh#L148-L156

I'm not familiar with DKIM and milters well enough to comment on that, but OpenDKIM using [`non_smtpd_milters` is apparently to run it for internal mail if something like `sendmail` is used](https://www.postfix.org/MILTER_README.html) (_which behaves differently than `swaks` by dropping straight into the mail queue IIRC_). While `smtpd_milters` is meant for signing?

Would be good to identify what part of that original PR config is causing this for you. It may be a fault of either of the `main.cf` referenced configs (_`user-patches.sh` can remove those with `postconf -X setting_name`, no `-P` which is for `master.cf`_).

---

> The latter of these being the culprit for the broken signature. I understand that the inclusion of the header with every MIME part is not strictly necessary, but the modifying of body parts is something I can not accept

Looking at [Postfix docs for milters](https://www.postfix.org/MILTER_README.html), they appear to run during the `smtpd` stage while `header_checks` applies in the later `cleanup` stage.

References (for me):
- https://blog.tyk.nu/blog/postfix-and-privacy/
- https://github.com/docker-mailserver/docker-mailserver/pull/3043#issuecomment-1409296675
- https://github.com/docker-mailserver/docker-mailserver/pull/708

If the cause can be better isolated and resolved, happy to accept a fix. Although we'd probably want a test-case which is going to add some friction AFAIK.

polarathene:
> I understand you problem, but these defaults were and are working rather nicely for a majority of users. Overwriting the config is actually the "elegant" solution here.

I don't know if it's working rather nicely, or if the users just didn't notice/care about DMS sending mail out with invalid DKIM signatures (_provided the body had MIME encoding headers_).

I am not familiar with how that sort of content is produced, but I do recognize `quoted-printable` as related to [this bug report](https://github.com/docker-mailserver/docker-mailserver/issues/3881).

If we can have `swaks` submit a mail that would have content to implicitly add those headers to the message body, we could probably reproduce a test-case for the test suite. DKIM only needs DNS for verification IIRC?

---

> Thanks, will do.
> Though my `user-patches.sh` is `git pull --commit --autostash`

@mpldr I think you pasted the wrong snippet there?

mpldr:
On Thu 15 Feb 2024 13:56:18, Brennan Kinney wrote:
> > The mail that was sent
>
> Was this sent by you through DMS on port 587/465?

Yes, though the copy acquired is from the Sent folder.

>
> ---
>
> > this is too broad and touches the headers sent for MIME parts as well
>
> It seems that when you set `header_checks`, it will implicitly be the default for a separate setting [`mime_header_checks`](https://www.postfix.org/postconf.5.html#mime_header_checks). Although from reading the [header_checks config docs](https://www.postfix.org/header_checks.5.html), it doesn't seem to suggest it'll affect the body content in anyway as you mentioned later.
>
> I don't have time to investigate but wonder if this is what is actually affecting you? Perhaps try disabling that, it'll need to be done via `postfix-master.cf` override (_although `user-patches.sh` could also handle it_).

Since removing the line solves the issue, my money is on that it is
actually the culprit

-- 
Moritz Poldrack
https://moritz.sh

> Misuse may cause suffocation.

polarathene:
> Since removing the line solves the issue, my money is on that it is actually the culprit

Could you be more specific please? Was it `mime_header_checks`? Or were you referring to `header_checks`?

There is several considerations, but since you reply by email I don't know if you see everything like the referenced code snippets or my edits after first posting a response.

Try identify if you can if it's due to any of these settings by removing them. I would assume either `mime_header_checks` but first might try just removing `smtp_header_checks`, technically I don't think `header_checks` in `main.cf` should be there either.

```bash
# main.cf:
postconf -X header_checks
# Quite possibly this one:
postconf -X smtp_header_checks
# Set to empty instead of default that uses header_checks:
postconf 'mime_header_checks='

# This one should be the last one to try, shouldn't be necessary to fix?
# master.cf on submission(s) ports (587/465):
postconf -PX sender-cleanup/unix/header_checks
```

mpldr:
On Sat 17 Feb 2024 15:55:12, Brennan Kinney wrote:
> > Since removing the line solves the issue, my money is on that it is actually the culprit
>
> Could you be more specific please? Was it `mime_header_checks`? Or were you referring to `header_checks`?

Sorry, I was referring to the line in sender_header_filter.pcre

> There is several considerations, but since you reply by email I don't know if you see everything like the referenced code snippets or my edits after first posting a response.

I do not get the updates through mail, no.

> Try identify if you can if it's due to any of these settings by removing them. I would assume either `mime_header_checks` but first might try just removing `smtp_header_checks`, technically I don't think `header_checks` in `main.cf` should be there either.

Will do and let you know the results.

> ```bash
> # main.cf:
> postconf -X header_checks
> # Quite possibly this one:
> postconf -X smtp_header_checks
> # Set to empty instead of default that uses header_checks:
> postconf 'mime_header_checks='
>
> # This one should be the last one to try, shouldn't be necessary to fix?
> # master.cf on submission(s) ports (587/465):
> postconf -PX sender-cleanup/unix/header_checks
> ```

It does sound reasonable that the last line overwrites `mime_header_checks=`

-- 
Moritz Poldrack
https://moritz.sh

> Swim at your own risk.

PR #3043, opened 2023-01-30
postfix: adjust header checks for outgoing emails

georglauterbach:
# Description

I consulted the Arch wiki and found the first two options in the new adjusted file (`Received` & `User-Agent`) (here
https://wiki.archlinux.org/title/postfix). This immediately made sense to me, so I added them.

I then looked through various emails I sent (with different mail clients as well). I could find none of the headers in the file except for the MIME version, which contained user agent information, so I kept the replacement there as well.

I removed the other entries; they seemed either useless or pretty niche to me.

---

References:

1. https://github.com/docker-mailserver/docker-mailserver/pull/3040#issuecomment-1407877464
2. https://github.com/docker-mailserver/docker-mailserver/pull/3040#issuecomment-1408106883

## Type of change

## Checklist:

### DIFF
--- a/target/postfix/sender_header_filter.pcre
+++ b/target/postfix/sender_header_filter.pcre
@@ -1,11 +1,3 @@
-/^\s*Received:.*with ESMTPSA/       IGNORE
-/^\s*Received:.*amavisd-new/        IGNORE
-/^\s*X-Originating-IP:/             IGNORE
-/^\s*X-Mailer:/                     IGNORE
-/^\s*Mime-Version: 1.0.*/           REPLACE MIME-Version: 1.0
-/^\s*User-Agent/                    IGNORE
-/^\s*X-Enigmail/                    IGNORE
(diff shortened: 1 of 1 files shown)


casperklein:
`X-Mailer` is used for example by scripts. In case of PHP for example, this contains the script path on the server which shouldn't be leaked.

`X-Originating-IP` may be used by web fronends like rainloop etc.

`X-Enigmail` is used by a thunderbird extension google told me.

---

`/^Received: .*/ IGNORE`

This would remove all received headers, not only those added by DMS. The current solution might be better. For example, when DMS is in relay/forward mode?

georglauterbach:
I didn't know about that, thank's for clarifying this :)

About the `/^Received: .*/ IGNORE`, I actually just wanted to say this because I noticed it right now, but you beat me to it :D I think this can be closed again, as the current configuration seems to be fine after all.

Issue #2692, opened 2022-07-24
[BUG] LDAP bind password shows up in log as cleartext


spirkaa:
### Miscellaneous first checks

- [X] I checked that all ports are open and not blocked by my ISP / hosting provider.
- [X] I know that SSL errors are likely the result of a wrong setup on the user side and not caused by DMS itself. I'm confident my setup is correct.

### Affected Component(s)

cleartext password in log

### What happened and when does this occur?

```Markdown
At container startup LDAP bind password shows up in log as cleartext at LOG_LEVEL=info.
```

### What did you expect to happen?

```Markdown
No LDAP bind password shows up in log as cleartext at LOG_LEVEL=info.
```

### How do we replicate the issue?

```Markdown
1. Configure LDAP_* env vars
2. Start up container
3. Look at logs
```

### DMS version

v11.1.0

### What operating system is DMS running on?

Linux

### What instruction set architecture is DMS running on?

x86_64 / AMD64

### What container orchestration tool are you using?

Docker

### Relevant log output

```bash
>> /etc/postfix/ldap-users.cf: saslauthd_ldap_password = myLdapBindPass
>> /etc/postfix/ldap-users.cf: bind_pw = myLdapBindPass
>> /etc/postfix/ldap-groups.cf: saslauthd_ldap_password = myLdapBindPass
>> /etc/postfix/ldap-groups.cf: bind_pw = myLdapBindPass
>> /etc/postfix/ldap-aliases.cf: saslauthd_ldap_password = myLdapBindPass
>> /etc/postfix/ldap-aliases.cf: bind_pw = myLdapBindPass
>> /etc/postfix/ldap-domains.cf: saslauthd_ldap_password = myLdapBindPass
>> /etc/postfix/ldap-domains.cf: bind_pw = myLdapBindPass
>> /etc/postfix/ldap-senders.cf: saslauthd_ldap_password = myLdapBindPass
>> /etc/postfix/ldap-senders.cf: bind_pw = myLdapBindPass
>> /etc/dovecot/dovecot-ldap.conf.ext: dnpass = myLdapBindPass
```

### Other relevant information

```Markdown
I'm using read-only service account created exclusively for docker-mailserver, but someone less experienced may use personal or even domain admin account for ldap bind and leak his password through the log.
```

### What level of experience do you have with Docker and mail servers?

- [ ] I am inexperienced with docker
- [ ] I am inexperienced with mail servers
- [ ] I am uncomfortable with the CLI

[...]

PR #3045, opened 2023-01-30
removal: configomat (submodule)

georglauterbach:
# Description

I wanted to do this for a long time. The submodule has only one script in it, and this script does not adhere to our high standards when it comes to Bash. I rewrote it as a proper helper function. This way, we can get rid of the whole submodule. This makes the LDAP code a bit easier to maintain too; although I got to admit that I didn't do it specifically for this reason; my main concern was code quality and usage of an unnecessary submodule.

## Type of change

## Checklist:


georglauterbach:
Noting here for future inspections: test fails randomly (and this test is not associated with the changes in any way)

```txt
not ok 32 checking dovecot mailbox format: mdbox file created in 31119ms
# (from function `repeat_until_success_or_timeout' in file test/test_helper/common.bash, line 51,
#  in test file test/tests/serial/mail_with_mdbox.bats, line 28)
#   `repeat_until_success_or_timeout 30 docker exec mail_with_mdbox_format /bin/sh -c '[ $(ls /var/mail/localhost.localdomain/user1/storage/m.1 | wc -l) -eq 1 ]'' failed
```

There is _parsing of `ls` involved_ .. This needs to change :D

polarathene:
> Noting here for future inspections: test fails randomly (and this test is not associated with the changes in any way)

They're due to the Postfix wrapper removal I think, that's where I first saw them pop up, but forgot to PR revised versions of those tests in response :sweat_smile: 

Presumably (_I haven't looked at the tests yet_) they fail now because Postfix loads faster which likewise broke the amavis spam tests (that did get fixed).

target/scripts/helpers/utils.sh
```diff
@@ -63,3 +63,48 @@ function _reload_postfix
...
+# If you want to set a new value for `readme_directory` in Postfix's `main.cf`,
+# you can set the environment variable `POSTFIX_README_DIRECTORY='/new/dir/'`
+# (`POSTFIX_` is an arbitrary prefix, you can choose the one you like),
+# and then call this function:
+# `_replace_by_env_in_file 'POSTFIX_' 'PATH TO POSTFIX's main.cf>`
+function _replace_by_env_in_file
+{
+  if [[ -z ${1+set} ]]
+  then
+    _log 'warn' "(ldap.sh:_replace_in_file) missing first argument"
```
polarathene, line 87: Note that these logs you have several of reference `ldap.sh:_replace_in_file`, I assume that needs to be updated to new file/method name?

Is a warning correct, or should it be failing hard since this is at start-up? There's `dms_panic` method to do a similar failure with context of where the failure is happening (_I've used it in `ssl.sh` helper_).

polarathene (changes requested):
I would like to see the method use the same test coverage that `configomat` has, I detailed those in an [old issue where I proposed a rust port](https://github.com/docker-mailserver/docker-mailserver/issues/2692#issuecomment-1198105049).

If you're open to introducing rust into the project, it might also be an alternative option. The rust version is a bit more thorough and verbose, but I'd argue easier to maintain for most than all the shell voodoo going on for this utility method :sweat_smile: (_unless you're @casperklein and @georglauterbach of course who have crazy good shell-fu skills_)

I would like similar rust code to handle other areas of the project, especially with the user management, but need other maintainers to be comfortable with rust in the project as I am likely to have reduced availability in the future for DMS.

georglauterbach:
> > Noting here for future inspections: test fails randomly (and this test is not associated with the changes in any way)
> 
> They're due to the Postfix wrapper removal I think, that's where I first saw them pop up, but forgot to PR revised versions of those tests in response 😅
> 
> Presumably (_I haven't looked at the tests yet_) they fail now because Postfix loads faster which likewise broke the amavis spam tests (that did get fixed).

IIRC I have seen them even before your change to Postfix, so I think they might actually be unrleated. They are a timing issie though, I agree, since our CI is running faster and faster.

georglauterbach:
> I would like to see the method use the same test coverage that `configomat` has, I detailed those in an [old issue where I proposed a rust port](https://github.com/docker-mailserver/docker-mailserver/issues/2692#issuecomment-1198105049).

Agreed; I can add such tests.

> If you're open to introducing rust into the project, it might also be an alternative option. The rust version is a bit more thorough and verbose, but I'd argue easier to maintain for most than all the shell voodoo going on for this utility method 😅 (_unless you're @casperklein and @georglauterbach of course who have crazy good shell-fu skills_)

To start things off: I love Rust. I _think_ it is by far the best programming language out there. Im using RUst in private and at work, almost every single day. So from a maintenance point of view, I can definitely handle it.

But I guess this is not enough, not even close. Before we get going with Rust, and I am not against that really, we need to make up our mind what this would look like. In the meantime, I'd go forward with Bash.

We would need to answer a few important questions:

1. Can people maintain the Rust code? (I actually think that the Bash code we exploy right now is is pretty high-class, but sometimes also hard to understand as a consequence. Therefore, maintaining Rust would probably solely depend on the amount of people that can actually program in Rust?)
2. What would the integration look like? Would there be Rust code in this project? WOuld you package it, or build it when building the container?

> I would like similar rust code to handle other areas of the project, especially with the user management, but need other maintainers to be comfortable with rust in the project as I am likely to have reduced availability in the future for DMS.

I think DMS could profit from Rust as well. But we should have a dedicated maintainers discussion or an issue for this topic IMO.

georglauterbach:
Noting another spurious test failure here:

```txt
not ok 42 checking dovecot mailbox format: sdbox file created in 30905ms
# (from function `repeat_until_success_or_timeout' in file test/test_helper/common.bash, line 51,
#  in test file test/tests/serial/mail_with_sdbox.bats, line 28)
#   `repeat_until_success_or_timeout 30 docker exec mail_with_sdbox_format /bin/sh -c '[ $(ls /var/mail/localhost.localdomain/user1/mailboxes/INBOX/dbox-Mails/u.1 | wc -l) -eq 1 ]'' failed
# ls: cannot access '/var/mail/localhost.localdomain/user1/mailboxes/INBOX/dbox-Mails/u.1': No such file or directory
# ls: cannot access '/var/mail/localhost.localdomain/user1/mailboxes/INBOX/dbox-Mails/u.1': No such file or directory
# ls: cannot access '/var/mail/localhost.localdomain/user1/mailboxes/INBOX/dbox-Mails/u.1': No such file or directory
# ls: cannot access '/var/mail/localhost.localdomain/user1/mailboxes/INBOX/dbox-Mails/u.1': No such file or directory
# ls: cannot access '/var/mail/localhost.localdomain/user1/mailboxes/INBOX/dbox-Mails/u.1': No such file or directory
# ls: cannot access '/var/mail/localhost.localdomain/user1/mailboxes/INBOX/dbox-Mails/u.1': No such file or directory
# ls: cannot access '/var/mail/localhost.localdomain/user1/mailboxes/INBOX/dbox-Mails/u.1': No such file or directory
# ls: cannot access '/var/mail/localhost.localdomain/user1/mailboxes/INBOX/dbox-Mails/u.1': No such file or directory
# ls: cannot access '/var/mail/localhost.localdomain/user1/mailboxes/INBOX/dbox-Mails/u.1': No such file or directory
# ls: cannot access '/var/mail/localhost.localdomain/user1/mailboxes/INBOX/dbox-Mails/u.1': No such file or directory
# ls: cannot access '/var/mail/localhost.localdomain/user1/mailboxes/INBOX/dbox-Mails/u.1': No such file or directory
# ls: cannot access '/var/mail/localhost.localdomain/user1/mailboxes/INBOX/dbox-Mails/u.1': No such file or directory
# ls: cannot access '/var/mail/localhost.localdomain/user1/mailboxes/INBOX/dbox-Mails/u.1': No such file or directory
# ls: cannot access '/var/mail/localhost.localdomain/user1/mailboxes/INBOX/dbox-Mails/u.1': No such file or directory
# ls: cannot access '/var/mail/localhost.localdomain/user1/mailboxes/INBOX/dbox-Mails/u.1': No such file or directory
# ls: cannot access '/var/mail/localhost.localdomain/user1/mailboxes/INBOX/dbox-Mails/u.1': No such file or directory
# ls: cannot access '/var/mail/localhost.localdomain/user1/mailboxes/INBOX/dbox-Mails/u.1': No such file or directory
# ls: cannot access '/var/mail/localhost.localdomain/user1/mailboxes/INBOX/dbox-Mails/u.1': No such file or directory
# ls: cannot access '/var/mail/localhost.localdomain/user1/mailboxes/INBOX/dbox-Mails/u.1': No such file or directory
# ls: cannot access '/var/mail/localhost.localdomain/user1/mailboxes/INBOX/dbox-Mails/u.1': No such file or directory
# ls: cannot access '/var/mail/localhost.localdomain/user1/mailboxes/INBOX/dbox-Mails/u.1': No such file or directory
# ls: cannot access '/var/mail/localhost.localdomain/user1/mailboxes/INBOX/dbox-Mails/u.1': No such file or directory
# ls: cannot access '/var/mail/localhost.localdomain/user1/mailboxes/INBOX/dbox-Mails/u.1': No such file or directory
# ls: cannot access '/var/mail/localhost.localdomain/user1/mailboxes/INBOX/dbox-Mails/u.1': No such file or directory
# ls: cannot access '/var/mail/localhost.localdomain/user1/mailboxes/INBOX/dbox-Mails/u.1': No such file or directory
# ls: cannot access '/var/mail/localhost.localdomain/user1/mailboxes/INBOX/dbox-Mails/u.1': No such file or directory
# ls: cannot access '/var/mail/localhost.localdomain/user1/mailboxes/INBOX/dbox-Mails/u.1': No such file or directory
# ls: cannot access '/var/mail/localhost.localdomain/user1/mailboxes/INBOX/dbox-Mails/u.1': No such file or directory
# Timed out on command: docker exec mail_with_sdbox_format /bin/sh -c [ $(ls /var/mail/localhost.localdomain/user1/mailboxes/INBOX/dbox-Mails/u.1 | wc -l) -eq 1 ]
```

polarathene:
> Before we get going with Rust, we need to make up our mind what this would look like. In the meantime, I'd go forward with Bash.
> We would need to answer a few important questions:

For quite some time now, the active maintainers have effectively been 3 of us. I've learned more about bash / shell scripts through you two, but if I'm not actively using those skills I forget what the syntax / features are and will need to look up again how to interpret what I'm seeing.

As you're already comfortable with Rust, this mostly flips the DX issue from me to casper. But as I am likely to become less active as a maintainer here, that could be a concern.

On the plus side, I can't say that bash has attracted many contributions, and definitely not more maintainers. There's been multiple times where I've had to lose time into figuring out how to do something with bash / commands that would be much simpler with a language like rust (or even python :grimacing: ), especially if that's nested arrays and parsing / editing files.

As long as there's already a solid foundation of rust introduced, that upfront knowledge is taken care of and isn't likely to cause much maintenance burden. I think there would be less need to reach out to a search engine / docs as the syntax of what is going on is often more obvious. Plus if something goes wrong the compiler is pretty good at telling you where and often how to fix it.

Rust experience from contributors probably won't be any more common than decent bash, so it may not help attract more maintainers. I do think it is easier to maintain with proper modules vs the "globals" sourcing approach we tend to take with shell scripts. There is also less portability issues (_macOS / Windows contributors, and even linux like my local `grep` misbehaving vs `rg`_).

> What would the integration look like? Would there be Rust code in this project? Would you package it, or build it when building the container?

That's a good question. For CLI like programs, those could be separate repos, but once you start porting the bulk of shell scripts, then it would need to be part of this project.

- When it's CLI programs, you could just have a CI build and publish artifacts that can be pulled into a main Dockerfile.
- Once replacing core scripts, then it's being updated more frequently and tied more heavily to what goes on in this repo, thus needs PRs for traceability to not fall out of sync like we used to have with features and the old Github wiki. We'd probably be building in the container at that point?

---

> I think DMS could profit from Rust as well. But we should have a dedicated maintainers discussion or an issue for this topic IMO.

Yeah all good. My focus right now is to get the project where I'd like it to be first, so that's with bash.

If I then find time to propose introducing some rust I'll raise a proper maintainer discussion about it :+1: 

---

> Noting another spurious test failure here:

I'll tackle those tests today if you've not already started :sweat_smile:

target/scripts/startup/setup-stack.sh
```diff
@@ -379,7 +379,7 @@ function _setup_ldap
     [[ ${FILE} =~ ldap-aliases ]] && export LDAP_QUERY_FILTER="${LDAP_QUERY_FILTER_ALIAS}"
     [[ ${FILE} =~ ldap-domains ]] && export LDAP_QUERY_FILTER="${LDAP_QUERY_FILTER_DOMAIN}"
     [[ ${FILE} =~ ldap-senders ]] && export LDAP_QUERY_FILTER="${LDAP_QUERY_FILTER_SENDERS}"
-    _replace_by_env_in_file 'LDAP_' "${FILE}"
+    [[ -f ${FILE} ]] && _replace_by_env_in_file 'LDAP_' "${FILE}"
```
polarathene, line 382: If you add your guard here, then the failure log/panic would be not be raised? Makes more sense to communicate the failure?

```suggestion
    _replace_by_env_in_file 'LDAP_' "${FILE}"
```

test/test-files/replace_by_env_in_file.conf
```diff
@@ -0,0 +1,6 @@
+key_1 = value
```
polarathene, line 1: I would probably move this into `test/config/` sub-folder instead. Look at the config overrides test that does similar. It's better for scoping a file that belongs only to the test I think? Especially when they're meant to be config files.

polarathene (changes requested):
While I find the bash version daunting to digest, excellent work! :grinning: 

I compared the PR test to the [configomat one](https://github.com/alinmear/docker-configomat/blob/415550485438d7e6ca6136016cca4bbab99ccf6f/test/tests.bats), looks good + nice improvements! :+1:  (_and fixes to pass the previously disabled test cases_)

test/test-files/replace_by_env_in_file.conf
polarathene: Technically if you retain the separate key/value lines per test like the original, then you could set all ENV and run a single test in the container. Each `--env` arg can be pre-faced with a comment for the test-case description.

Then again individual test cases are nice too :+1:

target/scripts/startup/setup-stack.sh
georglauterbach: Done in [20c16eb](https://github.com/docker-mailserver/docker-mailserver/pull/3045/commits/20c16eb0526548a4e343b678eb97f6dde44e1216)

test/test-files/replace_by_env_in_file.conf
georglauterbach: Something similar done in 1849c7182a8b4b0697e47dc46b7ab5bc690fefcf

target/scripts/helpers/utils.sh
```diff
@@ -84,27 +84,26 @@ function _replace_by_env_in_file
...
+    dms_panic__invalid_value 'first argument unset' 'utils.sh:_replace_by_env_in_file'
   elif [[ -z ${2+set} ]]
   then
-    _log 'warn' "(ldap.sh:_replace_in_file) missing second argument"
-    return 1
+    dms_panic__invalid_value 'second argument unset' 'utils.sh:_replace_by_env_in_file'
   elif [[ ! -f ${2} ]]
   then
-    _log 'warn' "(ldap.sh:_replace_in_file) file '${2}' could not be found"
+    _log 'warn' "(utils.sh:_replace_by_env_in_file) file '${2}' does not exist"
```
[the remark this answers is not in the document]
polarathene: Do we need a flag to toggle between logging a warning or panic failing for this? When would you want to prefer a logged warning instead of a hard fail when your usage references a file (or no file at all) that doesn't exist?

test/test-files/replace_by_env_in_file.conf
polarathene: Not exactly what I meant :sweat_smile: 

I was thinking about a single container with `--env` for each test case, the input config, and then comparing the transformed config in the container to some "expected" version of the config.

Without the container like that, the separate test cases seemed fine, and were a bit more descriptive about each test, now they're more vague and I'd need to reference the config a bit more to make sense of what the lines are actually testing that differentiates them from the others in the same test case group.

polarathene (dismissed):

target/scripts/helpers/utils.sh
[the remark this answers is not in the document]
georglauterbach: Okay, okay, go convinced me; I will make it use panic-only and then we check for the file to exists before we call the method.

polarathene (approved):

target/scripts/helpers/utils.sh
```diff
@@ -63,3 +63,54 @@ function _reload_postfix
...
+    dms_panic__invalid_value 'second argument unset' 'utils.sh:_replace_by_env_in_file'
+  elif [[ ! -f ${2} ]]
+  then
+    dms_panic__invalid_value "file '${2}' does not exist" 'utils.sh:_replace_by_env_in_file'
+  fi
+
+  local ENV_PREFIX=${1} CONFIG_FILE=${2}
+  local ESCAPED_VALUE ESCAPED_KEY
+
+  while IFS='=' read -r KEY VALUE
```
casperklein, line 107: This might not work in all cases, see [here](https://github.com/alinmear/docker-configomat/blob/415550485438d7e6ca6136016cca4bbab99ccf6f/configomat.sh#L26-L27).

target/scripts/helpers/utils.sh
```diff
@@ -63,3 +63,54 @@ function _reload_postfix
...
+  do
+    KEY=${KEY#"${ENV_PREFIX}"} # strip prefix
+    ESCAPED_KEY=$(sed -E 's#([\=\&\|\$\.\*\/\[\\^]|\])#\\\1#g' <<< "${KEY,,}")
+    ESCAPED_VALUE=$(sed -E 's#([\=\&\|\$\.\*\/\[\\^]|\])#\\\1#g' <<< "${VALUE}")
+    [[ -n ${ESCAPED_VALUE} ]] && ESCAPED_VALUE=" ${ESCAPED_VALUE}"
```
casperklein, line 112: Not sure, if I understand that correctly. When can `${ESCAPED_VALUE}` be empty?

casperklein (approved):
Nice work!

PR #3046, opened 2023-01-30
chore(Amavis): only add configuration to Postfix when enabled

georglauterbach:
# Description

Since I am running Rspamd nowadays, I noticed there still are ports open that belong to Amavis. This is because the Amavis configuration is a fixed part of Postfix's `master.cf`. I changed that. Now, the Amavis section is added when Amavis really is enabled.

I took the chance and added proper indentation to `master.cf`; hence the diff is a bit fuzzy. **But**, only the Amavis part was adjusted, the rest is just styling.

## Type of change

## Checklist:


target/postfix/master.cf
```diff
@@ -40,74 +40,36 @@ smtps     inet  n       -       n       -       -       smtpd
...
-retry     unix  -       -       n       -       -       error
-discard   unix  -       -       n       -       -       discard
-local     unix  -       n       n       -       -       local
-virtual   unix  -       n       n       -       -       virtual
-lmtp      unix  -       -       n       -       -       lmtp
-anvil     unix  -       -       n       -       1       anvil
-scache    unix  -       -       n       -       1       scache
-
-sender-cleanup unix n   -       n       -       0       cleanup
+sender-cleanup unix n        -       n       -       0       cleanup
```
polarathene, line 47: This appears to be slightly misaligned:

```suggestion
sender-cleanup unix  n       -       n       -       0       cleanup
```

polarathene (approved):
Not really sure about storing the config in `/etc/amavis` within the container, but otherwise no other concerns :+1: 

Might be worth doing similar for the other 3 additions we have, but these are all default and not configurable I think. `pickup` in particular is modifying the [original `pickup` config](https://github.com/vdukhovni/postfix/blob/6876e420275bb44e98827c76306f7ff5481eb192/postfix/conf/master.cf#L1-L81).

If I get around to refactoring the relay feature support, it'd be taking a similar approach to this Amavis appending. So I'm a bit more swayed towards a common location in the container (`/etc/dms/postfix/master.d`?)

target/postfix/master.cf
georglauterbach: Done

polarathene (approved):

polarathene:
@georglauterbach I noticed this:

https://github.com/docker-mailserver/docker-mailserver/blob/9e9f8026d9260db89da37b555b16975fb8cb2ae7/target/postfix/main.cf#L88-L89

It's removed if `ENABLE_AMAVIS=0`, but probably should be added instead if `ENABLE_AMAVIS=1`?:

https://github.com/docker-mailserver/docker-mailserver/blob/9e9f8026d9260db89da37b555b16975fb8cb2ae7/target/scripts/startup/setup-stack.sh#L73-L82

PR #3050, opened 2023-02-01
tests: Reduce time taken by a third for testing cipher suites

polarathene:
# Description

This improvement is from [investigating the performance of `testssl.sh` containers](https://github.com/drwetter/testssl.sh/issues/2299#issuecomment-1409545691). The test goes from ~300 sec locally to ~200 sec.

Using `--openssl` uses the native `openssl` package within the image instead of the older `1.0.2` bundled from `testssl.sh`.

The test is only testing cipher suite compatibility is what we expect it to be, thus we do not need to run `testssl.sh` with a broader range of ciphers.

## Type of change

## Checklist:


georglauterbach (approved):
👍🏼

casperklein (approved):

PR #3051, opened 2023-02-01
tests: Migrate Dovecot DBox tests (sdbox + mdbox)

polarathene:
# Description

Two more tests migrated. These two have been failing more regularly, but should work more reliably now :+1: 

They test the [Dovecot DBox mailbox format](https://doc.dovecot.org/admin_manual/mailbox_formats/dbox/), which was [contributed in Nov 2019 as mostly an ENV to make 1-2 changes](https://github.com/docker-mailserver/docker-mailserver/pull/1314).

## Type of change

## Checklist:


test/tests/parallel/set1/dovecot/mailbox_format_dbox.bats
```diff
@@ -0,0 +1,59 @@
...
+  local CUSTOM_SETUP_ARGUMENTS=(
+    --env DOVECOT_MAILBOX_FORMAT=mdbox
+    --env PERMIT_DOCKER=container
+  )
+
+  _init_with_defaults
+  _common_container_setup 'CUSTOM_SETUP_ARGUMENTS'
+  _wait_for_smtp_port_in_container
+
+  _send_email 'existing-user1'=
```
polarathene, line 51:
```suggestion
  _send_email 'existing-user1'
```

test/tests/parallel/set1/dovecot/mailbox_format_dbox.bats
```diff
@@ -0,0 +1,59 @@
...
+  local CUSTOM_SETUP_ARGUMENTS=(
+    --env DOVECOT_MAILBOX_FORMAT=mdbox
+    --env PERMIT_DOCKER=container
+  )
+
+  _init_with_defaults
+  _common_container_setup 'CUSTOM_SETUP_ARGUMENTS'
+  _wait_for_smtp_port_in_container
+
+  _send_email 'existing-user1'
```
georglauterbach, line 51: Why the `=` at the end? If this is not needed, why does `_send_mail` still work? 😂

test/helper/common.bash
```diff
@@ -410,8 +410,9 @@ function _count_files_in_directory_in_container()
   _run_in_container_bash "[[ -d ${DIRECTORY} ]]"
   assert_success
 
-  _run_in_container_bash "find ${DIRECTORY} -maxdepth 1 -type f -printf 'x\n'"
+  _run_in_container_bash "find ${DIRECTORY} -maxdepth 1 -type f"
```
georglauterbach, line 413: I know that the `-printf 'x\n'` is not strictly required, but I read on SO that this way, one can avoid trouble with special charaters in file names. BATS probably already does that, but it really does not hurt either, does it?

georglauterbach (changes requested):
One question, one remark - rest looks good!

test/tests/parallel/set1/dovecot/mailbox_format_dbox.bats
polarathene: It looks like I removed it 2 hours roughly before your review, I'm not sure how it got there actually. I probably had my tests passing, then made some minor changes to tidy things up and had a typo there that got saved and amended into my commit by accident.

test/helper/common.bash
polarathene: I was originally getting a failure from the test as I was only guessing one line should have been returned. To get some feedback I added an `assert_output` into the test method, but the results were all `x` lines... Then I saw this addition and removed it so I could see what the test was vs what I expected.

We could revert this, but then the asserts I do afterwards on what lines (files) are expected in the results wouldn't be usable.

georglauterbach: Ah, I see, we really do check the output. Then this is completely fine!

georglauterbach (approved):

casperklein (approved):

Issue #3022, opened 2023-01-23
[FR] Add the ability to redirect a domain catch-all to some user mailbox


Jule-:
# Feature Request

## Context

As stated in the [documentation](https://docker-mailserver.github.io/docker-mailserver/edge/faq/#how-can-i-configure-a-catch-all), right now, if you want to add a domain catch-all I have to add a new alias in the `docker-data/dms/config/postfix-virtual.cf`.

### Is your Feature Request related to a Problem?

The problem with this approach is that you will have to setup dummy aliases in that file for all users to correctly deliver their emails to their inboxes because if you don't, the catch-all will be selected in the alias resolution phase for all regular users. Pretty cumbersome if you are adding/deleting users regularly.

### Describe the Solution you'd like

A better approach for this need is to have a catch-all directive in the `vmailbox` file, that is to say, in the Postfix [virtual_mailbox_maps](http://www.postfix.org/postconf.5.html#virtual_mailbox_maps).
Extract from the Postfix [documentation](http://www.postfix.org/VIRTUAL_README.html#:~:text=Here%20is%20an%20example%20of,18%20%20%20%20%20postmaster%40example.com%20postmaster):
```
Here is an example of a virtual mailbox domain "example.com":

 1 /etc/postfix/main.cf:
 2     virtual_mailbox_domains = example.com ...more domains...
 3     virtual_mailbox_base = /var/mail/vhosts
 4     virtual_mailbox_maps = hash:/etc/postfix/vmailbox
 5     virtual_minimum_uid = 100
 6     virtual_uid_maps = static:5000
 7     virtual_gid_maps = static:5000
 8     virtual_alias_maps = hash:/etc/postfix/virtual
 9 
10 /etc/postfix/vmailbox:
11     info@example.com    example.com/info
12     sales@example.com   example.com/sales/
13     # Comment out the entry below to implement a catch-all.
14     # @example.com      example.com/catchall
15     ...virtual mailboxes for more domains...
16 
17 /etc/postfix/virtual:
18     postmaster@example.com postmaster
```

But `vmailbox` is auto-generated by the `postfix-accounts.cf` file, and it seems that it could potentially not be stable to modify it with `user-patches.sh` file regarding potential override by the change-detection code.

So it seems this use case needs some code modification around the `vmailbox` creation and potentially the `setup.sh` script.

[...]

PR #3055, opened 2023-02-03
tests(refactor): Extract mail account management tests from `tests.bats`

polarathene:
# Description

Test cases should remain mostly the same, except:
- In the `user4` test, there was an account removal at the end that was probably a typo and intended as `user4` as a clean-up step ([original PR](https://github.com/docker-mailserver/docker-mailserver/pull/441)). I've dropped it as it doesn't seem necessary.
- The account password update check was part of a [PR introducing the feature](https://github.com/docker-mailserver/docker-mailserver/pull/413), but the `sleep 2` lines don't seem necessary so I've removed those.
- Two tests without `postfix-accounts.cf` via an alternative volume and `docker run` were refactored to not need that a all.
- `tests.bats` had two groups of "accounts" tests, I've separated them with some comment headers. The very first test didn't seem as thorough, and improving on that seems to have raised a bug? (_not entirely sure, just that it's related to the Dovecot dummy alias workaround, potentially related feature request https://github.com/docker-mailserver/docker-mailserver/issues/3022 _)

To assist with review, if helpful I've staged changes out into multiple commits improved diffs:
- First commit handles cut / paste between files.
- Next commit revises test case descriptions.
- Followed by refactoring of test cases.

Some test cases could probably be collapsed or changed further, but this is good enough as an initial pass :+1: 

## Type of change

## Checklist:


test/tests/parallel/set3/mta/account_management.bats
```diff
@@ -0,0 +1,161 @@
...
+  # Waits until change event has created directory but not completed:
+  _wait_until_account_maildir_exists 'user3@domain.tld'
+  # Should trigger a new change event:
+  _exec_in_container setup email del -y 'user3@domain.tld'
+
+  # NOTE: This is only checking `postfix-accounts.cf`, account may still persist
+  # elsewhere momentarily such as the Dovecot UserDB until change event kicks in.
+  __check_mail_account_exists 'user3@domain.tld'
+  assert_failure
+  assert_output ''
```
georglauterbach, line 97: The command fails, and we check for empty output. Is that intended?

georglauterbach (changes requested):
We should take care of the parsing of `ls` and the directory checks; the rest looks very good!

test/tests/parallel/set3/mta/account_management.bats
polarathene: I mostly ported over whatever was in the original test cases. This used to be something like a `[[ -z ${output} ]]` I think. I haven't looked into if it was meant to ensure there wasn't a failure with output.

I can remove the check if it doesn't make sense?

I do recall that if the sed + grep commands didn't handle preventing the substring or `.` wildcard matches, you'd get the wrong result in either of these `__check_mail_account_exists()` calls with output. Although with `assert_failure` I guess this 2nd assert for output doesn't contribute anything? :thinking: 

```suggestion
  assert_failure
```

test/tests/parallel/set3/mta/account_management.bats
```diff
@@ -0,0 +1,161 @@
...
+  _should_output_number_of_lines 7
+
+  # Relevant log output from scripts/helpers/accounts.sh:_create_dovecot_alias_dummy_accounts():
+  # [  DEBUG  ]  Adding alias 'alias1@localhost.localdomain' for user 'user1@localhost.localdomain' to Dovecot's userdb
+  # [  DEBUG  ]  Alias 'alias2@localhost.localdomain' is non-local (or mapped to a non-existing account) and will not be added to Dovecot's userdb
+  # [  DEBUG  ]  Adding alias '@localdomain2.com' for user 'user1@localhost.localdomain' to Dovecot's userdb
+}
+
+@test "should have created maildir for 'user1@localhost.localdomain'" {
+  _run_in_container ls -d '/var/mail/localhost.localdomain/user1'
```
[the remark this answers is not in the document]
polarathene: It does look like that was there intent. Applied your suggestions :+1:

georglauterbach (approved):

casperklein (approved):

Issue #3028, opened 2023-01-26
How to disable SA_SPAM_SUBJECT?


ferdinandyb:
# Subject

Other

## Description

I tried setting `SA_SPAM_SUBJECT` to empty, and also to `" "`, but in both cases I got the default `***SPAM***` prepended to the subject. Since I'm moving spam to the Junk folder, I don't want anything prepended. How can I achieve this?

ferdinandyb:
I also think it's pretty dangerous to kill spam by default (`SA_TAG2` and `SA_KILL` being on the same level). I think most people would expect spam to arrive to a spam folder by default.

georglauterbach:
Try setting `SA_SPAM_SUBJECT=undef`

georglauterbach:
> I also think it's pretty dangerous to kill spam by default (`SA_TAG2` and `SA_KILL` being on the same level). I think most people would expect spam to arrive to a spam folder by default.

This could probably be adjusted, but I am not an SA expert.

georglauterbach:
IIRC I had some mails markes as Spam in my inbox. Make sure you have set `SPAMASSASSIN_SPAM_TO_INBOX=1` & `MOVE_SPAM_TO_JUNK=1`.

I have moved to Rspamd though (for which support has recently improved (due to #3016)) and support will continue to improve. Maybe this is interesting for you :)

polarathene:
We have a test that covers ENV config to send spam to Inbox or a Junk folder. I can agree with the latter being a smarter default and we could switch to that:

https://github.com/docker-mailserver/docker-mailserver/blob/05db27f817aa0a512f710c10eafe3ab71903f99a/test/tests/parallel/set1/spam_virus/spam_junk_folder.bats#L57-L59

> but having `SA_TAG2 = SA_KILL` means nothing ever goes to the spam folder, everything that is regarded as spam is silently deleted.

[This explains the config fairly well](http://gogs.info/books/debian-mail/chunked/antispam.amavis.html) (_**NOTE:** Lookup the actual config key like `sa_kill_level_deflt`, not our DMS specific ENV `SA_KILL`_). `SA_KILL` is a bit misleading in that sense as it's intended to be a value to trigger an action (`final_spam_destiny`) at a certain threshold value.

[...]

Issue #2816, opened 2022-10-06
Weird SPAM behaviour


mindrunner:
I see a weird behaviour on my mailserver which I cannot explain myself. My spam setup is pretty much the one described in:

```
ENABLE_SPAMASSASSIN=1
ENABLE_SPAMASSASSIN_KAM=1
SPAMASSASSIN_SPAM_TO_INBOX=1
MOVE_SPAM_TO_JUNK=0
SA_TAG=-100000.0
SA_TAG2=3.75
SA_KILL=100000.0
SA_SPAM_SUBJECT=undef
```

Furthermore I set up `sa-learn` --> https://docker-mailserver.github.io/docker-mailserver/edge/config/environment/#how-can-i-make-spamassassin-better-recognize-spam

as well as the better control thing -> https://docker-mailserver.github.io/docker-mailserver/edge/config/environment/#how-do-i-have-more-control-about-what-spamassasin-is-filtering

For filtering I am using a sieve rule like this:

```
   require ["comparator-i;ascii-numeric","relational","fileinto"];                 
                                                                                   
if address :domain "To" ["seconddomain.com", "mx0.seconddomain.com"] {                        
   fileinto "anotherfolder";                                                          
} elsif header :contains "X-Spam-Flag" "YES" {                                  
   fileinto "Junk";                                                              
} elsif allof (                                                                 
   not header :matches "x-spam-score" "-*",                                     
   header :value "ge" :comparator "i;ascii-numeric" "x-spam-score" "3.75" ) {   
   fileinto "Junk";                                                              
}                                                                              
```

My expectation for this setup would be that whatever goes to `seconddomain.com` will always land in `anotherfolder` no matter what the Spam Flags and Scores are. However I see randomly mails matching that pattern moved to the Junk folder. Furthermore, I find mails in Junk which are not for `seconddomain.com` but have `x-spam-flag` set to `NO` and a score lower than `3.75`.

![image](https://user-images.githubusercontent.com/1413542/194185521-ab5ebe34-76a1-44af-86cd-c6bb078b528d.png)

Any hints highly appreciated

PR #2708, opened 2022-07-31
docs: Improve cron tasks and fix link in documentation

JiLleON:
# Description
Everything that was mentioned in the issue and fixing the link to the actual

Fixes #2702 

## Type of change

- [x] Improvement (non-breaking change that does improve existing functionality)
- [x] This change requires a documentation update

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [ ] If necessary I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

Issue #2491, opened 2022-03-17
SA_KILL does not superseeds SPAMASSASSIN_SPAM_TO_INBOX


frugan-it:
# Subject

I would like to configure a not well documented mail server use case

## Description

In the [documentation](https://docker-mailserver.github.io/docker-mailserver/edge/faq/#how-do-i-have-more-control-about-what-spamassasin-is-filtering) it says:

> SA_KILL needs to be very high, to make sure nothing is bounced at all (SA_KILL superseeds SPAMASSASSIN_SPAM_TO_INBOX)

but, if `SPAMASSASSIN_SPAM_TO_INBOX=1`, it does not appear to me that *SA_KILL superseeds SPAMASSASSIN_SPAM_TO_INBOX*.

SPAM emails that exceed the `SA_KILL` level are still sent to the inbox, because (always when `SPAMASSASSIN_SPAM_TO_INBOX=1`) these 2 variables are set as `D_PASS`:

**/etc/amavis/conf.d/49-docker-mailserver**:
```
$final_spam_destiny       = D_PASS;    
$final_bad_header_destiny = D_PASS;
```
by `setup-stack.sh` script:

https://github.com/docker-mailserver/docker-mailserver/blob/321ae744fa873ca300af83983b4f916ac8dc9c35/target/scripts/startup/setup-stack.sh#L930-L931

Mind you, this is exactly how I want to work, I don't want SPAM emails to be bounced.
I just wanted to point out that IMO what is indicated in the documentation is not correct..

In a nutshell: it is not necessary to set a high value of `SA_KILL` when `SPAMASSASSIN_SPAM_TO_INBOX=1`.
Correct me if I'm wrong..

georglauterbach:
Thank you for sharing this. Please provide a PR that the maintainers can review if you want the documentation to change :)

PR #3058, opened 2023-02-03
Better default value for SA_KILL variable

casperklein:
# Description

Fixes https://github.com/docker-mailserver/docker-mailserver/issues/3028#issuecomment-1405206243

## Type of change

## Checklist:

### DIFF
--- a/mailserver.env
+++ b/mailserver.env
@@ -323,7 +323,7 @@ SA_TAG=2.0
 SA_TAG2=6.31
 
 # triggers spam evasive actions
-SA_KILL=10000.0
+SA_KILL=10.0
 
 # add tag to subject if spam detected
 SA_SPAM_SUBJECT=***SPAM*****
--- a/mailserver.env
+++ b/mailserver.env
@@ -323,7 +323,7 @@ SA_TAG=2.0
 SA_TAG2=6.31
 
 # triggers spam evasive actions
-SA_KILL=6.31
+SA_KILL=10000.0
 
 # add tag to subject if spam detected
 SA_SPAM_SUBJECT=***SPAM*****


georglauterbach (approved):

polarathene:
**EDIT:** This should already provide the desired behaviour by default... what are we actually fixing with this then?

https://github.com/docker-mailserver/docker-mailserver/blob/05db27f817aa0a512f710c10eafe3ab71903f99a/target/scripts/helpers/variables.sh#L69

https://github.com/docker-mailserver/docker-mailserver/blob/05db27f817aa0a512f710c10eafe3ab71903f99a/target/scripts/helpers/variables.sh#L79

---

This is a no from me :-1: 

I have [documented why in the associated issue](https://github.com/docker-mailserver/docker-mailserver/issues/3028#issuecomment-1416531516). Basically `SA_KILL` is a threshold to trigger the Amavis action (_we default to `D_BOUNCE`_), and simply bumping this value to an excessively high default to avoid mail being bounced seems redundant?

We have other ENV to support the alternative behaviour this PR is attempting to resolve, it would be better to adjust the defaults instead.

---

<details>
<summary>Debian Bullseye image with Amavis package defaults (no DMS involved): `/etc/amavis/conf.d/20-debian_defaults`:</summary>

```
$sa_tag_level_deflt  = 2.0;  # add spam info headers if at, or above that level
$sa_tag2_level_deflt = 6.31; # add 'spam detected' headers at that level
$sa_kill_level_deflt = 6.31; # triggers spam evasive actions
$final_spam_destiny       = D_PASS;    
$final_bad_header_destiny = D_PASS;     # False-positive prone (for spam)
```

We can see that we're currently just leaving these ENV to their respective defaults upstream and expecting users to change them if they're not happy with that. That probably isn't an issue, except that we have `D_BOUNCE` instead of the default `D_PASS`.

</details>

---

<details>
<summary>Resources for references:</summary>

[Here is an article](https://marekbosman.com/site/tweak-spamassassin-thresholds/) that looks into tweaking the values a bit for what works better for them. I don't know about those values or relevance now, but it explains the config nicely with some feedback on their experience (_some mail still not being detected as spam, some getting treated as spam_):

> There is a certain minimum score for obvious spam messages. This value can be used for **`$sa_tag2_level_deflt`**. (Note that this has no effect on the number of messages being blocked. It just indicates that anything below this limit is considered non-spam.)
> Anything below 0.5 is considered clean. Most major players (Gmail, Hotmail, Mailchimp etc.) have excellent scores (far far below zero) and therefore these will always be delivered. Some spam messages have scores just above or even below zero and will also pass the filter. This is something I take for granted.

> The problem area is the twilight zone between `0.5` and `2.0`.
> For some reason too many legit messages look kind of spammy (no subject line, only images or videos, all caps, misconfigured server, …) and there is no foolproof way for a non-human to distinguish them from the spam messages that succeed in hiding their spammy intentions.
> By narrowing down this range you will receive less spam, but potentially miss certain legit messages. Using trial and error will give result in values that are acceptable for you.

[More technical reference](http://gogs.info/books/debian-mail/chunked/antispam.amavis.html) on the config, that also seemed helpful:

> When SpamAssassin processes message, it gives them a score identifying what is the probability of message being a spam.
> **`$sa_tag_level_deflt`** tells amavis that if the score is greater or equal to this level that amavis should append `X-Spam` headers to the message.
> The score of `-999` means that we want to apply `X-Spam` header to all of our messages, so it is easier for us to see why is, or why is not, a message considered spam by examining the headers of the message.

> **`$sa_tag2_level_deflt = 5`** is the spam level the message needs to reach for amavis to apply the header `X-Spam-Flag: YES` so we (our mail client) knows that this message is considered to be spam. At this level the message subject is prefixed with the value of the variable **`$sa_spam_subject_tag`**.

> **`$sa_kill_level_deflt`** holds the value of the spam level that message needs to reach to do something with the message. What we are going to do with messages that are spam depends of the value of **`$final_spam_destiny`** variable. 
> In this case we are discarding this messages.
> This means that the message is not going to reach to recipients mailbox at all. But that does not mean that the message is lost. Spam messages that are discarded can still be fetched from the system. Default configuration is to store them in the `/var/lib/amavis/virusmails` folder.

</details>

georglauterbach:
Oh, I misunderstood this too then. Thanks for the clarification! I'd not merge this as is then too, but maybe this PR can bring an enhanced documentation?

casperklein:
- `MOVE_SPAM_TO_JUNK` - This is only for IMAP setups, not for POP3.
- `SPAM_TO_INBOX` disables bouncing of SPAM messages completely

Some might set this both to `0`.

`SA_TAG2` defines when to add spam headers, e.g. changing the subject to `*** SPAM ***` and `SA_KILL` defines at what level to bounce the mail.

>We can see that we're currently just leaving these ENV to their respective defaults upstream and expecting users to change them if they're not happy with that. That probably isn't an issue, except that we have D_BOUNCE instead of the default D_PASS.

Because we changed it to `D_BOUNCE`, it's a an issue, or at least not sane:

Setting `SA_TAG2 == SA_KILL` leads to never receiving an email with `*** SPAM ***` subject?

`SA_KILL=10000.0` might be the wrong approach. As you pointed out, this will basically prevent any mail from bouncing.

But `SA_KILL` should be higher than `SA_TAG2`. What do you think?

polarathene:
> * `MOVE_SPAM_TO_JUNK` - This is only for IMAP setups, not for POP3.

Is it? I thought Dovecot sieve scripts were for handling any mail arriving into the inbox, protocol agnostic?

Mail that is delivered to DMS goes through SA and Amavis, which would add the spam header to trigger the sieve script would it not?:

https://github.com/docker-mailserver/docker-mailserver/blob/9df71c27a0e8824c6ea14f5ce6d828eed7a9b8dd/target/dovecot/sieve/before/60-spam.sieve#L1-L4

Then IMAP and POP3 are for mail clients to retrieve mail from Dovecot after the fact? Why would POP3 be excluded from the feature?

> Setting `SA_TAG2 == SA_KILL` leads to never receiving an email with `*** SPAM ***` subject?

That's what I understand. I am not familiar with spam scores as I don't actually run DMS (_all this time and I still just contribute to a project I don't use_ :joy: ), so I don't know what a good value would be for `SA_KILL`. Based on the blog I referenced, they had `SA_KILL` at `2.0`, but it wasn't clear what score is 99% likely to be actual spam :man_shrugging: 

> But `SA_KILL` should be higher than `SA_TAG2`. What do you think?

I am not sure. In what scenario do you want to mark mail as spam, but not handle it differently? (_I'm not sure if the action `D_PASS` actually treats the mail differently, maybe metadata or logs?_ :man_shrugging: )

If there was a reasonable "definitely spam" score that `SA_KILL` could use for `D_BOUNCE` perhaps that'd make sense if `6.31` is still prone to a fair amount of false-positives.

> `SA_KILL=10000.0` might be the wrong approach. As you pointed out, this will basically prevent any mail from bouncing.

Honestly considering what this would do, it just sounds like the smarter choice is to default to `D_PASS` instead of `D_BOUNCE`. (**EDIT: _which I forgot we already do, I just forgot to update the part of my message that you quoted_ :sweat_smile: )

https://github.com/docker-mailserver/docker-mailserver/blob/05db27f817aa0a512f710c10eafe3ab71903f99a/target/scripts/startup/setup-stack.sh#L923-L934

Ok so if bouncing mail is the problem when that ENV is flipped, we could also use `D_DISCARD` (_I think that's the correct one_), which my other reference link said would not deliver the mail, but still keep it stored locally in quarantine.

Alternatively, we better document that disabling `SPAMASSASSIN_SPAM_TO_INBOX` will bounce any mail marked as spam? And that if that is too aggressive, they can raise the `SA_KILL` value to increase the threshold of when spam is bounced? (_allowing some mail marked as spam through still if `SA_TAG2` is lower_)

casperklein:
> MOVE_SPAM_TO_JUNK - This is only for IMAP setups, not for POP3.
> Is it? I thought Dovecot sieve scripts were for handling any mail arriving into the inbox, protocol agnostic?

That's my assumption, or at least part of it, because POP3 does not now the concept of folders. So moving SPAM to junk folder has no effect. In POP3 this have to be configured client side in the MUA as a rule (not in scope of DMS).

You are right, that Sieve processing also apply to POP3, but moving a mail to junk folder in particular will have no effect (from POP3 view).

---

`SPAM_TO_INBOX` disables rejecting emails completely. The use case I thought of was:

```
SPAM_TO_INBOX=0
SA_TAG2=6
SA_KILL=10
```

This allows to tag mails as SPAM beginning with a score of 6 and rejecting mails with a score of >=10.

`SA_KILL=10000.0` is bad I agree, because it disables rejecting mails. But `SA_KILL` should be higher than `SA_TAG2` IMO.

georglauterbach (approved):

polarathene:
This seems acceptable :+1:  Although I prefer there to be an actual demand to warrant the change :thinking: 

Since it only affects the users explicitly opt-ing in to bouncing mail marked as spam, this change may result in receiving unwanted spam in their inbox. Improving docs probably makes more sense.

---

> That's my assumption, or at least part of it, because POP3 does not now the concept of folders. So moving SPAM to junk folder has no effect.
> In POP3 this have to be configured client side in the MUA as a rule (not in scope of DMS).

I thought that this is about what happens when mail arrives to Postfix, it gets filtered through Amavis + SA, then sent to Dovecot for mail storage. The MUA then can access that mailbox via IMAP or POP3.

It should still be in Dovecot's junk folder for that mailbox, and marked as spam in the mail headers?

> You are right, that Sieve processing also apply to POP3, but moving a mail to junk folder in particular will have no effect (from POP3 view).

My understanding is that sieve is on Dovecot's end with how it manages mail storage. Agnostic from MUA access regardless of protocol.

---

> This allows to tag mails as SPAM beginning with a score of 6 and rejecting mails with a score of >=10.
> 
> `SA_KILL=10000.0` is bad I agree, because it disables rejecting mails. But `SA_KILL` should be higher than `SA_TAG2` IMO.

<details>
<summary>`SA_KILL=10` seems okay-ish from a quick look at non-standard values used elsewhere:</summary>

- [Earlier reference](https://marekbosman.com/site/tweak-spamassassin-thresholds/) found `SA_TAG=0`, `SA_TAG2=0.5`, `SA_KILL=2` to work reasonably well for them.
- [Other earlier reference](http://gogs.info/books/debian-mail/chunked/antispam.amavis.html) used `SA_TAG=-999` (always add spam info, for testing?), `SA_TAG2=5`, `SA_KILL=12`. They use `D_DISCARD`, so mail is not delivered to inbox, but not bounced either, but quarantined in local storage.
- [ArchWiki config example](https://wiki.archlinux.org/title/Amavis#SpamAssassin_support) uses `SA_TAG=1`, `SA_TAG2=1` , `SA_KILL=5`.
- [Some spam scores for test spam mails shared](http://verchick.com/mecham/public_html/spam/image_spam.html). `SA_KILL=10` seems alright based on that.
- [2017 DMS user report with `SA_TAG2=5, `SA_KILL=1000`](https://github.com/docker-mailserver/docker-mailserver/issues/473#issuecomment-272019228): 25 detected as spam, 40 undetected, 10 legitimate emails.
- [2017 DMS user reports with defaults](https://github.com/docker-mailserver/docker-mailserver/issues/768) legitimate mail receiving a score of `8`.
- [2018 DMS user report with `SA_KILL=4`](https://github.com/docker-mailserver/docker-mailserver/issues/1045) based on their observations of spam often being marked between 4 and 6.
- [2019 DMS user report](https://github.com/docker-mailserver/docker-mailserver/issues/1214#issuecomment-521946518) only mentions legitimate mail marked as spam as sender only sent a URL. That's enough to score the current `6.31` I guess.
- [2020 DMS user report](https://github.com/docker-mailserver/docker-mailserver/issues/1396#issuecomment-597210154) default `D_BOUNCE` at the time prevented receiving a registration email. UX expressed that some spam came through, **they did not expect only some spam to be bounced**.

Ok, so for the defaults, this makes no difference right? Due to `D_PASS`, the mail still gets tagged as spam (`SA_TAG2`) at the same score, and thus the sieve will move to junk.

`SA_KILL` only has relevance for our alternative action `D_BOUNCE`, those users will potentially get more spam delivered instead of rejected (_and we don't have any major complaints AFAIK of users opting in to this feature about legitimate mail being rejected_).

I suppose with release notes, if there is more spam being delivered when they'd rather it bounced, they can manually change `SA_KILL` or raise an issue about it :man_shrugging: Likewise, no reason to lower `SA_TAG2`... so sounds good :+1: 

<details>
<summary>Amavis config observations (not too important)</summary>

**EDIT:** I have noticed in the [relevant Amavis docs section](https://www.ijs.si/software/amavisd/amavisd-new-docs.html#tagkill):

> kill level
> if spam score is at or above kill level, mail is blocked; and sender receives a nondelivery notification unless spam score exceeds dsn cutoff level.

`/etc/amavis/conf.d/20-debian_defaults`:
```
$sa_dsn_cutoff_level = 10;   # spam level beyond which a DSN is not sent
```

I thought I had seen something like a percentage somewhere, and was going to raise this as a concern with increasing the `SA_KILL` value, but I'm mistaken :sweat_smile: (_I believe it was when I saw `sa_quarantine_cutoff_level = 25` in ArchWiki config snippet_)

That config file has some other info that might be worth noting:

```
# You should:
#   Use D_DISCARD to discard data (viruses)
#   Use D_BOUNCE to generate local bounces by amavisd-new
#   Use D_REJECT to generate local or remote bounces by the calling MTA
#   Use D_PASS to deliver the message
#
# Whatever you do, *NEVER* use D_REJECT if you have other MTAs *forwarding*
# mail to your account.  Use D_BOUNCE instead, otherwise you are delegating
# the bounce work to your friendly forwarders, which might not like it at all.
#
# On dual-MTA setups, one can often D_REJECT, as this just makes your own
# MTA generate the bounce message.  Test it first.
#
# Bouncing viruses is stupid, always discard them after you are sure the AV
# is working correctly.  Bouncing real SPAM is also useless, if you cannot
# D_REJECT it (and don't D_REJECT mail coming from your forwarders!).

$final_virus_destiny      = D_DISCARD;  # (data not lost, see virus quarantine)
$final_banned_destiny     = D_DISCARD;  
$final_spam_destiny       = D_PASS;    
$final_bad_header_destiny = D_PASS;     # False-positive prone (for spam)
```

</details>

</details>

---

## Related issue history

<details>
<summary>For better traceability via referencing existing issues / PRs</summary>

- March 2020: https://github.com/docker-mailserver/docker-mailserver/pull/1431
  - Introduced the `SPAMAASSASSIN_SPAM_TO_INBOX` ENV.
- May 2020: https://github.com/docker-mailserver/docker-mailserver/pull/1485
  - Introduced the `MOVE_SPAM_TO_JUNK` ENV (_and documents `SPAMASSASSIN_SPAM_TO_INBOX=1` but leaving it [disabled by default due to maintainer decision](https://github.com/docker-mailserver/docker-mailserver/issues/1396#issuecomment-622996857)_).
- Jan 2022: https://github.com/docker-mailserver/docker-mailserver/issues/2366
  - Sieve error due to lack of `+w` permission for sieve script. Unresolved, potentially user misconfiguration or environment issue.
  - References [Feb 2022 PR that changes `SPAMASSASSIN_SPAM_TO_INBOX` ENV default to 1](https://github.com/docker-mailserver/docker-mailserver/pull/2361), released in [March 2022 in DMS `v10.5`](https://github.com/docker-mailserver/docker-mailserver/releases/tag/v10.5.0).
- Mar 2022: https://github.com/docker-mailserver/docker-mailserver/issues/2491
  - [Reports FAQ entry in docs](https://docker-mailserver.github.io/docker-mailserver/v11.3/faq/#how-do-i-have-more-control-about-what-spamassasin-is-filtering), suggestion to create a PR was the only activity, no progress made:

    > By default, SPAM and INFECTED emails are put to a quarantine which is not very straight forward to access.

    Apparently `D_BOUNCE` and some other actions will still quarantine. [Our ENV docs for `SA_KILL`](https://docker-mailserver.github.io/docker-mailserver/edge/config/environment/#sa_kill) also detail the quarantine location (_while a FAQ entry describes how to send to a dedicated mailbox_). [Amavis docs on quarantine](https://www.ijs.si/software/amavisd/amavisd-new-docs.html#quarantine).

    > `SA_KILL` needs to be very high, to make sure nothing is bounced at all (`SA_KILL` superseeds `SPAMASSASSIN_SPAM_TO_INBOX`)

    Likewise, inaccurate. As the FAQ entry continues to detail, it seems these were not the defaults, and have not been updated to clarify that this is default behaviour now, and when `SA_KILL` is relevant.

    These docs were [contributed to the original wiki in May 2020](https://github.com/docker-mailserver/docker-mailserver/commit/57837ccd92cfb6538fc2439589d17fa6dc5871a8) (_[associated issue](https://github.com/docker-mailserver/docker-mailserver/issues/1471#issuecomment-619793407)_). No real content changes to update them since.
- Oct 2022: https://github.com/docker-mailserver/docker-mailserver/issues/2816
  - Received no interaction. References [another SA FAQ entry](https://docker-mailserver.github.io/docker-mailserver/v11.3/faq/#how-can-i-make-spamassassin-better-recognize-spam) (_no new updates since migration from Wiki, which had [multiple edits in 2019](https://github.com/docker-mailserver/docker-mailserver/blame/463bc967d229a55de6522f4b08d3a6427a3f8f67/docs/content/faq.md)_) (_[associated issue](https://github.com/docker-mailserver/docker-mailserver/issues/473)_). There was [some minor improvements contributed in Aug 2022](https://github.com/docker-mailserver/docker-mailserver/pull/2708), but it was mostly focused on adding a section on SA cron for `fts_xapian` config.
  - They also reference the sieve script from the other mentioned FAQ entry. Their report was about non-deterministic delivery handling, potentially they had that suggested sieve script conflicting with the default one we ship? So first in first serve :man_shrugging: 

</details>

polarathene (approved):

polarathene:
I noticed that this only changed a default in the optional `.env` file. Docs and `variables.sh` were not updated :sweat_smile: 

Due to #3112 PR wanting to move `variables.sh`, may be best to wait until after that is sorted out though.

PR #3059, opened 2023-02-04
rspamd: add feature for adjusting options with a file parsed by DMS

georglauterbach:
# Description

Users should be able to do some easy adjustments with our help. With this addition, they can mount a file `rspamd-modules.conf` and our setup will evaluate it. You can read the documentation for this when looking at the docs preview for this PR :)

## Type of change

## Checklist:

### DIFF
--- a/docs/content/config/security/rspamd.md
+++ b/docs/content/config/security/rspamd.md
@@ -78,7 +78,8 @@ where `COMMAND` can be:
 3. `set-option-for-module`: sets the value for option `ARGUMENT2` to `ARGUMENT3` inside module `ARGUMENT1`
 4. `set-option-for-controller`: set the value of option `ARGUMENT1` to `ARGUMENT2` for the controller worker
 5. `set-option-for-proxy`: set the value of option `ARGUMENT1` to `ARGUMENT2` for the proxy worker
-6. `add-line`: this will add the complete line after `ARGUMENT1` (with all characters) to the file `/etc/rspamd/override.d/<ARGUMENT1>`
+6. `set-common-option`: set the option `ARGUMENT1` that [defines basic Rspamd behaviour][basic-options] to value `ARGUMENT2`
+7. `add-line`: this will add the complete line after `ARGUMENT1` (with all characters) to the file `/etc/rspamd/override.d/<ARGUMENT1>`
 
 !!! note "File Names & Extensions"
 
@@ -96,3 +97,4 @@ You can also have comments (the line starts with `#`) and blank lines in `rspamd
 [dms-default-configuration]: https://github.com/docker-mailserver/docker-mailserver/tree/master/target/rspamd
 [rbl-vs-dnsbl]: https://forum.eset.com/topic/25277-dnsbl-vs-rbl-mail-security/?do=findComment&comment=119818
 [dkim-signing-module]: https://rspamd.com/doc/modules/dkim_signing.html
+[basic-options]: https://rspamd.com/doc/configuration/options.html
--- a/docs/content/config/security/rspamd.md
+++ b/docs/content/config/security/rspamd.md
@@ -16,6 +16,12 @@ We provide a very simple but easy to maintain setup of Rspamd. If you want to ha
 
 If you want to adjust Rspamd's configuration, have a look at the ["Providing Custom Settings & Overriding Settings" section](#providing-custom-settings-overriding-settings) down below.
 
+!!! note "AMD64 vs ARM64"
+
+    We are currently doing a best-effort installation of Rspamd for ARM64 (from the Debian backports repository for Debian 11). The current version difference is two minor versions (AMD64 is at version 3.4, ARM64 at 3.2 \[13th Feb 2023\]).
+
+    Maintainers noticed only few differences, some of them with a big impact though. For those running Rspamd on ARM64, we recommend [disabling](#with-the-help-of-a-custom-file) the [DKIM signing module][dkim-signing-module] if you don't use it.
+
 ## The Default Configuration
 
 ### Mode of Operation
@@ -42,9 +48,9 @@ The [RBL module](https://rspamd.com/doc/modules/rbl.html) is enabled by default.
 
 !!! danger "Rspamd and DNS Block Lists"
 
-    When using Rspamd, the [RBL module](https://rspamd.com/doc/modules/rbl.html) is enabled by default. As a consequence, Rspamd will do a variety of DNS requests. Amongst other things, Rspamd will query DNS block lists (DNSBLs). There are a variety of issues involved when using DNSBLs. Rspamd will try to mitigate some of them by properly evaluating all return codes. This evaluation is a best effort, and adjustments made by
+    When the RBL module is enabled, Rspamd will do a variety of DNS requests to (amongst other things) DNSBLs. There are a variety of issues involved when using DNSBLs. Rspamd will try to mitigate some of them by properly evaluating all return codes. This evaluation is a best effort though, so if the DNSBL operators change or add return codes, it may take a while for Rspamd to adjust as well.
 
-    If you want to use RBLs, **try to use your own DNS resolver** and make sure it is set up correctly, i.e. it should be a non-public & **recursive** resolver. Otherwise, you might not be able ([see this Spamhaus post](https://www.spamhaus.org/faq/section/DNSBL%20Usage#365)) to make use of the block lists.
+    If you want to use DNSBLs, **try to use your own DNS resolver** and make sure it is set up correctly, i.e. it should be a non-public & **recursive** resolver. Otherwise, you might not be able ([see this Spamhaus post](https://www.spamhaus.org/faq/section/DNSBL%20Usage#365)) to make use of the block lists.
 
 ### Missing in the Current Implementation
 
--- a/docs/content/config/security/rspamd.md
+++ b/docs/content/config/security/rspamd.md
@@ -83,7 +83,7 @@ where `COMMAND` can be:
 
 !!! note "File Names & Extensions"
 
-    For command 1-3, we append the `.conf` suffix to the module name to get the correct file name automatically. For commands 4 & 5, the file name is fixed (you don't even need to provide it). For command 6, you will need to provide the whole file name (including the suffix) yourself!
+    For command 1 - 3, we append the `.conf` suffix to the module name to get the correct file name automatically. For commands 4 - 6, the file name is fixed (you don't even need to provide it). For command 7, you will need to provide the whole file name (including the suffix) yourself!
 
 You can also have comments (the line starts with `#`) and blank lines in `rspamd-modules.conf` - they are properly handled and not evaluated.
 
--- a/docs/content/config/security/rspamd.md
+++ b/docs/content/config/security/rspamd.md
@@ -2,7 +2,11 @@
 title: 'Security | Rspamd'
 ---
 
-!!! warning "Implementation of Rspamd into DMS is WIP!"
+!!! warning "The Current State of Rspamd's Integration Into DMS"
+
+    Recent pull requests have stabilized Rspamd's integration to a point where we encourage users to test the feature. We are confident that there are no major bugs in our code anymore that make using Rspamd infeasible. Please do note that there may still be breaking changes (made to the default configuration of Rspamd) as integration is still work in progress!
+
+    We expect to stabilize this feature with version `v12.1.0`.
 
 ## About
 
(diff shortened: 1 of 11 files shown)


target/scripts/startup/setup-stack.sh
```diff
@@ -141,6 +141,64 @@ EOF
...
+
+  local RSPAMD_CUSTOM_COMMANDS_FILE=/tmp/docker-mailserver/rspamd-commands
+  if [[ -f "${RSPAMD_CUSTOM_COMMANDS_FILE}" ]]
+  then
+    _log 'debug' "Found 'rspamd-commands' file - parsing and applying it"
+
+    while read -r COMMAND MODULE ARGUMENT1 ARGUMENT2
+    do
+      case "${COMMAND}"
+      in
```
polarathene, line 153:
```suggestion
      case "${COMMAND}" in
```

polarathene (changes requested):
Looks fine :)

The docs could probably do with some revision, but I lack the time to go through that. They're good enough.

I'm not sure about `rspamd-commands` as the config file name. Is it just for managing modules? Why not something like `rspamd.mod`, `rspamd-modules.conf` (_potentially gets syntax highlighting for comments in an editor_), or something else that looks more like a text file rather than a binary or folder at a glance? :sweat_smile: 

Likewise, `config/override-configs` was for a specific test to use. With the intent of having a cleaner separation of what files belong to what tests. I haven't given that too much thought, so it might make sense for some tests to share other config files that aren't in the common default config in tests.

test/helper/sending.bash
```diff
@@ -56,11 +56,12 @@ function _send_mail_and_get_id() {
 
   MAIL_ID=$(_exec_in_container tac /var/log/mail.log              \
     | grep -E -m 1 'postfix/smtpd.*: [A-Z0-9]+: client=localhost' \
-    | grep -E -o '[A-Z0-9]{11}')
+    | grep -E -o '[A-Z0-9]{9,12}' || true)
```
georglauterbach, line 59: Forgot to mention this (FYI): On aarch64, the ID postfix uses might not be 11 characters long, so I provide a bit of range here.

target/scripts/build/packages.sh
```diff
@@ -132,16 +132,31 @@ function _install_dovecot
...
+  local DEB_FILE='/etc/apt/sources.list.d/rspamd.list'
+  local RSPAMD_PACKAGE_NAME
+
+  # We try getting the most recent version of Rspamd for aarch64 (from an official source, which
+  # is the backports repository). The version for aarch64 is 3.2; the most recent version for amd64
+  # that we get with the official PPA is 3.4.
+  #
+  # Not removing it later is fine as you have to explicitly opt into installing a backports package
+  # which is not something you could be doing by accident.
+  if [[ $(uname --machine) == 'aarch64' ]]
```
polarathene, line 144: Not a big fan of this due to version difference, and the official rspamd advice is to not use the debian packages :sweat_smile: 

Hopefully the version disparity doesn't cause issues going forward if we take the approach you've chosen to support ARM64.

---

I had a browse of the [rspamd download docs](https://rspamd.com/downloads.html), and noticed they have support for Fedora (_and OpenSUSE via OBS, but TW isn't a good idea, and Leap releases are slow - not that Debian is any better_). 

At one point I wanted to consider Alpine as a base image offering, especially for the regular releases and newer packages, but I'm more reluctant now with all the `musl` compatibility and performance issues I've seen with it :sweat_smile: 

Fedora might be interesting, they have a minimal base image (slightly bigger than Debian), but hard to say if it's worthwhile or a smooth transition to change to. Those rspamd docs made it seem like it was supported but [it's been broken for a while](https://github.com/rspamd/rspamd.com/issues/604).

Maybe [OpenSUSE Leap image base](https://hub.docker.com/r/opensuse/leap/tags) would be worth looking into. It's fairly small at a glance, but I understand there is more familiarity with Debian :sweat_smile:  (_Leap is not likely to stay around either AFAIK as they're working towards a new container focused distro/spin_)

---

One other option is building rspamd ourselves with a multi-stage build, but that has it's own drawbacks (_maintenance + build time + CI cache?_).

polarathene (approved):
LGTM :)

target/scripts/build/packages.sh
georglauterbach: I agree, it's not very nice, but as of now, ARM64 users would have to use v2.7. With this change, they can use v3.2 which is only 2 minor versions behind the official PPA. I will rework this as soon as a better emerges :)

---

About the base image change: we should put this into our backlog for now I think :)

test/helper/sending.bash
georglauterbach: I don't have the slightest idea.. but this is what I saw today on my arm64 machine.

georglauterbach:
Added a comment about the mail ID in e5ab05f2a55ebcc6e447711964aa2c7709f51d0d.

target/scripts/build/packages.sh
polarathene: > About the base image change: we should put this into our backlog for now I think :)

I am in no rush, that's not a fun change to tackle and really needs to be worthwhile to bother with :sweat_smile: 

My main motivations were attempting to reduce image weight and switch to something with more frequent release cadence for us to not rely on all these external builds for newer versions, as reliance on that seems to be growing.

---

Debian Bookworm release should bump up the rspamd package release at least? :)

I [raised a request to get official ARM64 builds supported](https://github.com/rspamd/rspamd/issues/4390), but given the history of maintainer stances on that, I'm not sure if it'll be approved on their end.

I've also further investigated the Fedora situation further and no fedora release going back 5 years has support to install `rspamd` based on their instructions :confused: Their docs are terribly out of date for that support it seems.

polarathene (dismissed):

target/scripts/build/packages.sh
georglauterbach: This is going to be a tough one indeed. Ubuntu probably wont help here either as it's basically Debian.. We should discuss about this first on a maintainers discussion board or with discussions for this repository so users can post too. 

I thought Fedora was actually quite nice, while I don't have a great experience with OpenSUSE, but I'm open to listening to all proposals.

polarathene: No rush, we'll likely update to Bookworm before we change distro :)

georglauterbach:
@polarathene I rebased on `master` and added more commits after caabadad18b7623eabe9422fc046fa8b54fbabf2. Just FYI when reviewing again. Since I updated the setup file quite significantly, its probably worthwhile to revisit the whole file for review again though. Configuration wise, not much has changes, but two more module configuration commands were added (and one command name was adjusted). More than that, there were proper additions to the tests & revision of the documentation.

target/scripts/startup/setup-stack.sh
```diff
@@ -702,9 +612,15 @@ function _setup_dkim_dmarc
 
     # shellcheck disable=SC2016
     sed -i -E 's|^(smtpd_milters =.*)|\1 \$dmarc_milter|g' /etc/postfix/main.cf
+  else
+    sed -i '/.*inet:localhost:8891/d' /etc/postfix/main.cf
```
georglauterbach, line 616: Ah, forgot this in the commit message: our current `main.cf` still has the varibales `dkim_milter`/`dmarc_milter` and postconf will complain about them not being used when `ENABLE_OPENDKIM=0`/`ENABLE_OPENDMARC=0`. These changes (and line 619-623) get rid of the line and the warning.

docs/content/config/security/rspamd.md
```diff
@@ -2,37 +2,87 @@
 title: 'Security | Rspamd'
 ---
 
-!!! warning "Implementation of Rspamd into DMS is WIP!"
+!!! warning "The Current State of Rspamd's Integration Into DMS"
+
+    Recent pull requests have stabilized Rspamd's integration to a point where we encourage users to test the feature. We are confident that there are no major bugs in our code, that make using Rspamd infeasible, anymore. Please note that there may still be (breaking) changes ahead as integration is still work in progress!
```
polarathene, line 7:
```suggestion
!!! warning "The current state of Rspamd integration into DMS"

    Recent pull requests have stabilized integration of Rspamd to a point that we encourage users to test the feature. We are confident that there are no major bugs in our integration that make using Rspamd infeasible. Please note that there may still be (breaking) changes ahead as integration is still work in progress!
```

docs/content/config/security/rspamd.md
```diff
@@ -2,37 +2,87 @@
...
+
+### Manually
 
 DMS brings sane default settings for Rspamd. They are located at `/etc/rspamd/local.d/` inside the container (or `target/rspamd/local.d/` in the repository). If you want to change these settings and / or provide your own settings, you can
 
 1. place files at `/etc/rspamd/override.d/` which will override Rspamd settings and DMS settings
 2. (re-)place files at `/etc/rspamd/local.d/` to override DMS settings and merge them with Rspamd settings
 
-You can find a list of all Rspamd modules [on their website][modules].
+inside the container.
```
polarathene, line 52: Was this meant to be part of line 50?

target/scripts/helpers/setup-rspamd.sh
```diff
@@ -0,0 +1,172 @@
...
+  then
+    __rspamd__log 'warn' 'Running Amavis/SA & Rspamd at the same time is discouraged'
+  fi
+
+  if [[ ${ENABLE_CLAMAV} -eq 1 ]]
+  then
+    __rspamd__log 'debug' 'Enabling ClamAV integration'
+    sedfile -i -E 's|^(enabled).*|\1 = true;|g' /etc/rspamd/local.d/antivirus.conf
+    # RSpamd uses ClamAV's UNIX socket, and to be able to read it, it must be in the same group
+    usermod -a -G clamav _rspamd
```
polarathene, line 26: Should this not be handled the same way Amavis does in the Dockerfile? (_or the Amavis handling moved to it's setup script for enabling?_)

target/scripts/startup/setup-stack.sh
polarathene: Isn't this just for `dkim_milter` and `dmarc_filter`? Not better to match on those keys for deleting instead?

```suggestion
    sed -i '/^dkim_milter/d' /etc/postfix/main.cf
```

target/scripts/startup/setup-stack.sh
```diff
@@ -640,9 +612,15 @@ function _setup_dkim_dmarc
...
   fi
 
-  [[ ${ENABLE_OPENDKIM} -eq 1 ]] || return 0
+  if [[ ${ENABLE_OPENDKIM} -eq 0 ]]
+  then
+    sed -i '/.*inet:localhost:8893/d' /etc/postfix/main.cf
+    return 0
+  fi
```
polarathene, line 623: Same here, but note you're currently checking if OpenDKIM is disabled, while removing the `dmarc_milter` setting?

polarathene (changes requested):

target/scripts/helpers/setup-rspamd.sh
georglauterbach: I'd rather do it here since we only really need it when using ClamAV. I'd advocate for moving the Amavis-handling into scripts then (follow-up PR).

docs/content/config/security/rspamd.md
georglauterbach: Done in 62e528d380ac92f6cc2e361c4685c01780d38088

docs/content/config/security/rspamd.md
georglauterbach: Done in 62e528d380ac92f6cc2e361c4685c01780d38088

target/scripts/startup/setup-stack.sh
georglauterbach: Done in cef100363fd6765954a7dd0af3ed76b3da8b4d01. Took a slightly different approach now: commented the milter and uncomment them when DKIM/DMARC is enabled.

polarathene (dismissed):

georglauterbach:
Merged current `master` and resolved merge conflicts in [4a570ab](https://github.com/docker-mailserver/docker-mailserver/pull/3059/commits/4a570ab87321ec0ddc5eb525dd32d7d39c73879f). Added one more command (to adjust `options.inc`) in [96baf86](https://github.com/docker-mailserver/docker-mailserver/pull/3059/commits/96baf86acf41f36cd2bff0342d44364a86b4b3fe).

I will take care of resolving the conflicts that will arise when #3082 is merged. Then the PR is ready for review again :)

georglauterbach:
[1056876](https://github.com/docker-mailserver/docker-mailserver/pull/3059/commits/10568761479d6a84bc178ff20b73125adbe7428c) goes to show what I added to the docs a short while ago: "Program testing can be used to show the presence of bugs, but never to show their absence!" -- Dijkstra

polarathene (approved):
Praise be https://github.com/docker-mailserver/docker-mailserver/pull/3059/commits/96baf86acf41f36cd2bff0342d44364a86b4b3fe ! :rofl: 

> https://github.com/docker-mailserver/docker-mailserver/commit/10568761479d6a84bc178ff20b73125adbe7428c goes to show what I added to the docs a short while ago

Ha! I remember that quote from about a decade ago when I trained to be a software tester :sweat_smile: There was some terrifying bugs in medical equipment :grimacing:

georglauterbach:
> Praise be https://github.com/docker-mailserver/docker-mailserver/pull/3059/commits/96baf86acf41f36cd2bff0342d44364a86b4b3fe ! :rofl: 

😉

> > https://github.com/docker-mailserver/docker-mailserver/commit/10568761479d6a84bc178ff20b73125adbe7428c goes to show what I added to the docs a short while ago
>
> Ha! I remember that quote from about a decade ago when I trained to be a software tester :sweat_smile: There was some terrifying bugs in medical equipment :grimacing: 

Interesting! Never thought about hearing something linke this though.. actually a bit scary 🙈

georglauterbach:
Alright, after dismissing @polarathene's review for the one millionth time, I promise to not do it again (unless required by resolving another conflict - I do hope I don't need to..). I added proper commit messages to the last two commits to make the final review as easy as possible.

target/scripts/startup/setup-stack.sh
```diff
@@ -668,18 +645,27 @@ function _setup_dkim_dmarc
...
+      echo "Nameservers ${NAMESERVER_IPS}" >>/etc/opendkim.conf
       _log 'trace' "Nameservers added to '/etc/opendkim.conf'"
     fi
   fi
 
   if [[ ${ENABLE_OPENDMARC} -eq 1 ]]
   then
     _log 'trace' "Adding OpenDMARC to Postfix's milters"
-
+    # Make sure to append the milter _behind_ the OpenDKIM milter!
```
polarathene, line 658: Behind? After is probably more accurate? :)

```suggestion
    # Make sure to append the OpenDMARC milter _after_ the OpenDKIM milter!
```

polarathene (dismissed):
> I decided against the proposed solution of using an array/string that is appended to Postfix's milters at the end due to unnecessarry complexity it would add IMO.

I'm guessing the complication was with expanding the array values within the sed command? Here's two different ways that it could work:

```bash
local DKIM_DMARC=()
# In conditionals:
DKIM_DMARC+='$dkim_milter'
DKIM_DMARC+='$dmarc_milter'

# After conditionals retrieve setting value and expand array into it
# (Note querying settings will probably delay without calling our utility method to avoid that):
postconf "smtpd_milters = $(postconf -h smtpd_milters) ${DKIM_DMARC[@]}"
# Alternatively, create a custom setting to use as an interpolated variable (by Postfix):
postconf "dms_dkim_dmarc = ${DKIM_DMARC[@]}"
sed -i -E 's|^(smtpd_milters =.*)|\1 \$dms_dkim_dmarc|g' /etc/postfix/main.cf

# If you ever need to retrieve the value
# Query the custom setting for value (-h) directly:
postfix -h dms_dkim_dmarc
# Or expand (-x) to the interpolated value where it's used:
postfix -hx smtpd_milters
```

I won't block on it, but it seemed a bit more DRY, in that the related logic between the two ENV conditions handled that shared setting together in one place afterwards :sweat_smile:

target/scripts/startup/setup-stack.sh
georglauterbach: Yeah, you're right.. :D

georglauterbach: Done in [9756028](https://github.com/docker-mailserver/docker-mailserver/pull/3059/commits/97560282a7e640e80b9160df781366ec019744d0)

georglauterbach:
> > I decided against the proposed solution of using an array/string that is appended to Postfix's milters at the end due to unnecessarry complexity it would add IMO.
> 
> I'm guessing the complication was with expanding the array values within the sed command?

Actually, no :) I think your solution adds lines of code that are not necessary, since in your example you don't account for the extra calls to `_adjust_mtime_for_postfix_maincf` or the shellcheck comments. I implemented your solution, just to see what it looks like, but the extra complexity was not really worth it IMO. Moreover, there is no code handling the case in which `DKIM_DMARC` is empty, which adds another `if`..

> Here's two different ways that it could work: [...]

I do like some of the parts; especially adding things via `popstconf`. I have apllied some of this in the last PR feedback commit (97560282a7e640e80b9160df781366ec019744d0) :)

> I won't block on it, but it seemed a bit more DRY, in that the related logic between the two ENV conditions handled that shared setting together in one place afterwards 😅

I actually think it is really DRY now, and the current setup is easy to understand.

target/scripts/startup/setup-stack.sh
polarathene: The comment suggested change? It didn't seem to get applied :thinking:

polarathene (dismissed):

target/scripts/startup/setup-stack.sh
georglauterbach: Oh my, sorry! I know why, a stupid mistake. Will push one more commit..

polarathene (approved):

georglauterbach:
@polarathene this is a rather big PR, but I actually want to have this merged anyway. Do you think this PR is fine to merge with 1 review?

polarathene:
> Do you think this PR is fine to merge with 1 review?

I have been reasonably thorough with reviews. I know I sometimes miss things though. Pretty much our only other reviewer is Casper and he seems rather busy lately, probably swamped with notifications from this project to wade through too.

I don't see any harm in merging it, no release has been made, so there is still time to review even after a merge and provide feedback that can be addressed via a follow-up PR.

georglauterbach:
Alright, the I will go ahead :D

casperklein:
>  probably swamped with notifications from this project to wade through too.

Absolutely! lol

Issue #2970, opened 2023-01-01
[BUG] postfix: reject_unknown_client_hostname prevents legitimate mail from being received


mpldr:
### Miscellaneous first checks

- [X] I checked that all ports are open and not blocked by my ISP / hosting provider.
- [X] I know that SSL errors are likely the result of a wrong setup on the user side and not caused by DMS itself. I'm confident my setup is correct.

### Affected Component(s)

postfix

### What happened and when does this occur?

```Markdown
mailserver  | Jan  1 12:35:19 mail postfix/smtpd[1556]: NOQUEUE: reject: RCPT from unknown[54.240.85.218]: 450 4.7.25 Client host rejected: cannot find your hostname, [54.240.85.218]; from=<010101856d1d37d3-cb9c308b-0000-0000-0000-a21593c7994e-000000@bounce.stripe.com> to=<my-localpart+stripe@example.com> proto=ESMTP helo=<a85-218.smtp-out.us-west-2.amazonses.com>
```
```


### What did you expect to happen?

```Markdown
I expected an option to turn `reject_unknown_client_hostname` off, if it is not considered necessary. I can overwrite the config in the docker-compose, but that feels hacky.
```

### How do we replicate the issue?

```Markdown
I have absolutely no idea. I don't think this happens for all SES mails.
```

### DMS version

latest (d5db200744f1)

### What operating system is DMS running on?

Linux

### Which operating system version?

Archlinux

### What instruction set architecture is DMS running on?

AMD64 / x86_64

### What container orchestration tool are you using?

Docker Compose

### docker-compose.yml

```yml
moritz@pan ❯ cat docker-compose.yml                                 docker-mailserver -> master + ? ! $
services:
  mailserver:
    image: docker.io/mailserver/docker-mailserver:latest
      #entrypoint: ["/bin/sh", "-c", "while true; do sleep 10; done"]
    container_name: mailserver
    # If the FQDN for your mail-server is only two labels (eg: example.com),
    # you can assign this entirely to `hostname` and remove `domainname`.
    hostname: mail
    domainname: moritz.sh
    env_file: mailserver.env
    network_mode: host
    # More information about the mail-server ports:
    # https://docker-mailserver.github.io/docker-mailserver/edge/config/security/understanding-the-ports/
    # To avoid conflicts with yaml base-60 float, DO NOT remove the quotation marks.
    ports:
      - "25:25"    # SMTP  (explicit TLS => STARTTLS)
[...]

PR #2691, opened 2022-07-24
Add `reject_unknown_client_hostname` to main.cf

GoliathLabs:
# Description
As discussed in the below mentioned issue this adds the `reject_unknown_client_hostname` option.

Reference: http://www.postfix.org/postconf.5.html#reject_unknown_client_hostname

Fixes #2650 

## Type of change

- [X] Improvement (non-breaking change that does improve existing functionality)

## Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [X] If necessary I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes

PR #3061, opened 2023-02-05
Updated iPv6 entry in docs and removed not working fix.

### LINKED ISSUE #2927: IPv4 address must be set in docker-compose.yml explicitly
I recently noticed that gmail users were unable to send email to my docker-mailserver instance.  Messages were being bounced back with this message

```
450 4.7.25 Client host rejected: cannot find your hostname, [172.23.0.1]
```

In the docker logs: 

```
CONNECT from [172.23.0.1]:32968 to [172.23.0.1]:25
PASS OLD [172.23.0.1]:32968
connect from unknown[172.23.0.1]
Anonymouse TLS connection established from unknown[172.23.0.1]...
NOQUEUE: reject: RCPT from unknown[172.23.0.1]: 450 4.7.25 Client host rejected: cannot find your hostname, [172.23.0.1]
disconnect from unknown[172.23.0.1]...
```

`172.23.0.1` is the docker gateway address to the docker network.  What was weird is that *other* mail services, like yahoo, were sending mail correctly and the logs were showing CONNECT from the correct external IP: 

```
CONNECT from [170.249.216.138]:39102 to [172.23.0.3]:25
PASS OLD [170.249.216.138]:39102
connect from atl.slvit.us[170.249.216.138]
Anonymous TLS connection established from atl.slvit.us[170.249.216.138]...
prepend Received-SPF: Pass (mailfrom)...
...
```

It took some googling before I found out what was happening.  The diagnostics were done by @Rillke in [a comment](https://github.com/docker-mailserver/docker-mailserver/issues/1405#issuecomment-590106498) in #1405, which I quote: 

> Compose created an IPv4 network, since IPv6 is not enabled for Docker on that host. When another host connects to the host running the container using IPv6, the **Docker Proxy translates IPv6 to IPv4** and inserts itself/its **Gateway as Source IP Address**, essentially doing NAT.

The first thing I did was to enable IPv6 and assign an IPv6 address to the container.  But to my surprise, it was showing connections from the docker gateway's IPv6 address, instead of the external IPv6 address.  That is to say, it seems like the IPv6 implementation is broken in docker (as usual).  

```
CONNECT from [2001:db8:1::1:0:1]:32968 to [2001:db8:1::1:0:2]:25
PASS OLD [2001:db8:1::1:0:1]:32968
connect from unknown[2001:db8:1::1:0:1]
Anonymouse TLS connection established from unknown[2001:db8:1::1:0:1]...
NOQUEUE: reject: RCPT from unknown[2001:db8:1::1:0:1]: 450 4.7.25 Client host rejected: cannot find your hostname, [2001:db8:1::1:0:1]
disconnect from unknown[2001:db8:1::1:0:1]...
```

The only ~solution~ workaround that worked for me was @Rillke's: 

> As an interim solution I bound port 25 to the public IPv4 interface instead of any interface. If my host wants to send email, it has to authenticate now, but that's okay.
> 
> ```
>     ports:
> -      - "25:25"
> +      - "the_IP_v4:25:25"
> ```

This closes port 25 on the VPS's IPv6 address and works around the problem.  Now, seeing as: 

 - The only requirement to reproduce this is to have a server with IPv6 (pretty common)
 - The result is silently not receiving emails from gmail and other big services (very serious)
 - Docker is not likely to fix their IPv6 implementation anytime soon (unfortunately)

... I am proposing making the README and docker-compose.yml explicit about putting the server's external IP address in the ports section.  I'd be happy to PR it.

AlperShal:
# Description

This fix no longer works as mentioned in #2927 and with my same experience. @vedranmiletic had the [same experience](https://github.com/docker-mailserver/docker-mailserver/commit/854c1b52238af9922f73f8decfededc81efba8de#commitcomment-98661970) too.

Fixes #2927 

## Type of change

## Checklist:

- [No Need] I have commented my code, particularly in hard-to-understand areas
- [] If necessary I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/docs/content/config/advanced/ipv6.md
+++ b/docs/content/config/advanced/ipv6.md
@@ -2,47 +2,14 @@
 title: 'Advanced | IPv6'
 ---
 
-## Background
+## Not Available
+docker-mailserver doesn't currently support iPv6 networking! Trying to get mails using iPv6 will result with postfix not being able to find your hostname.
 
-If your container host supports IPv6, then `docker-mailserver` will automatically accept IPv6 connections by way of the docker host's IPv6. However, incoming mail will fail SPF checks because they will appear to come from the IPv4 gateway that docker is using to proxy the IPv6 connection (`172.20.0.1` is the gateway).
-
-This can be solved by supporting IPv6 connections all the way to the `docker-mailserver` container.
-
-## Setup steps
-
-```diff
-+++ b/serv/docker-compose.yml
-@@ -1,4 +1,4 @@
--version: '2'
-+version: '2.1'
-
-@@ -32,6 +32,16 @@ services:
-
-+  ipv6nat:
-+    image: robbertkl/ipv6nat
-+    restart: always
-+    network_mode: "host"
-+    cap_add:
-+      - NET_ADMIN
-+      - SYS_MODULE
-+    volumes:
-+      - /var/run/docker.sock:/var/run/docker.sock:ro
-+      - /lib/modules:/lib/modules:ro
-
-@@ -306,4 +316,13 @@ networks:
-
-+  default:
-+    driver: bridge
-+    enable_ipv6: true
-+    ipam:
-+      driver: default
-+      config:
-+        - subnet: fd00:0123:4567::/48
-+          gateway: fd00:0123:4567::1
-```
+## Fix
+To fix the issue, basicly remove the AAAA record for your mailserver. This will prevent email senders from using iPv6.
 
 ## Further Discussion
-
-See [#1438][github-issue-1438]
+See [#1438][github-issue-1438] and [#2927][github-issue-2927]
 
 [github-issue-1438]: https://github.com/docker-mailserver/docker-mailserver/issues/1438
+[github-issue-2927]: https://github.com/docker-mailserver/docker-mailserver/issues/2927


docs/content/config/advanced/ipv6.md
```diff
@@ -2,47 +2,14 @@
 title: 'Advanced | IPv6'
 ---
 
-## Background
+## Not Available
+docker-mailserver doesn't currently support iPv6 networking! Trying to get mails using iPv6 will result with postfix not being able to find your hostname.
```
georglauterbach, line 6: I don't think this is 100% correct. It is probably not even a DMS issue, but a Docker issue. I could try setting up a Kubernetes dual-stack cluster which supports IPv6 just fine and DMS should work in this setup too. @casperklein @polarathene do you think this really is accurate?

docs/content/config/advanced/ipv6.md
```diff
@@ -2,47 +2,14 @@
...
-+    driver: bridge
-+    enable_ipv6: true
-+    ipam:
-+      driver: default
-+      config:
-+        - subnet: fd00:0123:4567::/48
-+          gateway: fd00:0123:4567::1
-```
+## Fix
+To fix the issue, basicly remove the AAAA record for your mailserver. This will prevent email senders from using iPv6.
```
georglauterbach, line 9: > To fix the issue, basicly remove the AAAA record for your mailserver.

Haha, I'm all for that ❤️ This is probably not something someone who needs to use IPv6 wants to hear..

docs/content/config/advanced/ipv6.md
polarathene: > It is probably not even a DMS issue, but a Docker issue.
> @polarathene do you think this really is accurate?

The previous content was more accurate about the "why" and should remain really (_the gateway IP varies IIRC, dependent on what the docker network got assigned_).

The #2927 referenced issue is the one I was intending to respond to, but got side-tracked with all the test refactoring in jan :sweat_smile: I recently [shared details on how to get IPv6 working properly](https://github.com/docker-mailserver/docker-mailserver/pull/3057#issuecomment-1416706615), although I haven't applied it for DMS myself yet to verify that resolves the issue.

Probably better to update the docs with information from there as the IPv6 NAT support wasn't available when these docs were written IIRC.

polarathene (commented):
Background info seems relevant still. I'd rather not drop that. The alternative fix and additional issue to reference is good though :+1: 

It should be possible to configure your Docker host to support IPv6 NAT and _I think_ that will resolve the issue experienced. If anyone can confirm that I am happy to revise the docs here with that approach.

docs/content/config/advanced/ipv6.md
AlperShal: I didn't want to hear this too but that was the only fix I was able to find. :)

AlperShal:
> It should be possible to configure your Docker host to support IPv6 NAT and _I think_ that will resolve the issue experienced. If anyone can confirm that I am happy to revise the docs here with that approach.

If I hadn't started actively using my mail server I would definitely check it and validate but it's impossible for me to risk missing some mails right now.

vedranmiletic:
> Background info seems relevant still. I'd rather not drop that. The alternative fix and additional issue to reference is good though +1
> 
> It should be possible to configure your Docker host to support IPv6 NAT and _I think_ that will resolve the issue experienced. If anyone can confirm that I am happy to revise the docs here with that approach.

It might be possible, but the current documentation will fail to give that. So it should at least carry a big warning "this IPv6 NAT approach is only theoretically interesting as it is practically useless for mail reception".

polarathene:
> So it should at least carry a big warning "this IPv6 NAT approach is only theoretically interesting as it is practically useless for mail reception".

Docs state that IPv6 will connect, but mail will fail SPF check (_which causes rejection, at least if not authenticated IIRC?_):

> _If your container host supports IPv6, then `docker-mailserver` will automatically accept IPv6 connections by way of the docker host's IPv6._
> _However, incoming mail will fail SPF checks because they will appear to come from the IPv4 gateway that docker is using to proxy the IPv6 connection_

Referenced issue #2927  shows:

```
Client host rejected: cannot find your hostname, [172.23.0.1]
```

And [another related issue](https://github.com/docker-mailserver/docker-mailserver/issues/2970#issuecomment-1371935103) (_with another user already configured with IPv6 advice I gave_) had that same error (_but with an IP that wasn't a private address_). It's possible that [this PR](https://github.com/docker-mailserver/docker-mailserver/pull/2691) which made it to the `v11.2.0` release is another failure point being experienced, and thus the current docs are outdated as they don't account for that?

I've assigned myself to look into it. I'm currently focused on overhauling the test suite so that we can improve test coverage for issues like this, especially with DNS and more real-world style mail delivery. After that I'll try to reproduce and go from there.

vedranmiletic:
> Docs state that IPv6 will connect, but mail will fail SPF check (_which causes rejection, at least if not authenticated IIRC?_):

Correct.

> > _If your container host supports IPv6, then `docker-mailserver` will automatically accept IPv6 connections by way of the docker host's IPv6._
> > _However, incoming mail will fail SPF checks because they will appear to come from the IPv4 gateway that docker is using to proxy the IPv6 connection_

That is only correct if you use the stock configuration. If you enable robbertkl/ipv6nat per instructions, you will get the the address of the IPv6 gateway.

> And [another related issue](https://github.com/docker-mailserver/docker-mailserver/issues/2970#issuecomment-1371935103) (_with another user already configured with IPv6 advice I gave_) had that same error (_but with an IP that wasn't a private address_).

I haven't experienced the issue with the rejected public address, only private (both IPv4 as you wrote and IPv6 with robbertkl/ipv6nat). Hope this helps.

polarathene:
> If you enable robbertkl/ipv6nat per instructions, you will get the the address of the IPv6 gateway.

Since Docker supports IPv6 NAT via `/etc/docker/daemon.json` with `ip6tables: true` and `experimental: true`, this should replace the need for that extra container. I have heard that there is some minor differences, but I don't think they'd affect DMS differently. If we can resolve the issue the docs can be updated to suggest official support by Docker instead.

The quoted docs may only apply to stock configuration but they do describe the reason IPv6 is problematic out of the box, so I still see value in keeping that in the docs.

---

> I haven't experienced the issue with the rejected public address, only private (both IPv4 as you wrote and IPv6 with robbertkl/ipv6nat). Hope this helps.

So even with that container setup and providing proper IPv6 NAT, you don't receive mail? But no errors like others are seeing? Do you have any other logs that communicate failures?

vedranmiletic:
> So even with that container setup and providing proper IPv6 NAT, you don't receive mail? But no errors like others are seeing? Do you have any other logs that communicate failures?

I'm not sure what would proper IPv6 NAT mean in this case. Any configuration of IPv6 I found via web search and tried thus far, be it via `daemon.json` or via `docker-compose.yml`, resulted only in rejections. Either the rejected address was a private IPv4 address of the gateway or the private IPv6 address of the gateway. Therefore, I guess I haven't been able to get "proper NAT" working.

polarathene:
> I'm not sure what would proper IPv6 NAT mean in this case.

I meant that it was supported instead of the container seeing an IPv4 address it should be an IPv6 one. I think it can be a bit more problematic if there are other containers routing the traffic though. With reverse-proxies for example, they handle this fine, but if a container they bridge needs the external IP, they need to provide it via a separate header IIRC, as otherwise they get the wrong IP (internal container IP).

It would be helpful if there's a simple config + steps that I can try reproduce with when I get the time to investigate.

vedranmiletic:
> It would be helpful if there's a simple config + steps that I can try reproduce with when I get the time to investigate.

Sure, reset Docker config to default (i.e. remove `daemon.json` and restart the service) and then follow the IPv6 instructions from Docker-Mailserver docs. That should be enough.

> In that case, it's very much likely caused by [this PR](https://github.com/docker-mailserver/docker-mailserver/pull/2691). I think if you revert that change, it may fix the problem. If you try this and it does fix it for you, please let us know.

I believe this is the case. However, that PR is correct in its thinking so I don't want to revert it.

The container receiving the connection via IPv6 should see the public IPv6 of the sending side and not the private IPv4/IPv6 address of the gateway. The problem is that it doesn't happen and I have no idea whether this is due to incorrect configuration of Docker or robbertkl/ipv6nat. Either way, the documentation should mention the issue exists.

docs/content/config/advanced/ipv6.md
casperklein: > do you think this really is accurate?

Afaik, none of us is using IPv6. So we can't properly test/reproduce any errors that are reported. Given the amount of IPv6 related issue, I am not sure if we should advertise it to be functional. So a warning seems reasonable to me. The wording however is a bit harsh and might be adjusted.

georglauterbach:
As is, this PR can not be merged as maintainers decided it removes information still relevant. Since IPv6 is just an awfu.. complex protocol, this needs more work.

Moreover, this seems to have stalled, which I absolutely dislike, but I also understand it in this case.. so, @polarathene, what would you do now? I would like to do something about this PR - otherwise it will continue to rot.

polarathene:
> @polarathene, what would you do now? I would like to do something about this PR - otherwise it will continue to rot.

- I've been swamped, and need to prioritize other tasks in DMS backlog atm. If I can look into it this month I will.
- Still relevant to keep open as a reminder / todo. Alternatively an issue can be opened and reference this PR if you want to close it instead. So long as the concern is visible to be aware that it needs attention is all I care about right now.

georglauterbach:
Closing in favor if #3244 which superseeds this PR.

PR #3060, opened 2023-02-05
tests(refactor): `open_dkim.bats`

polarathene:
# Description

[2 years later](https://github.com/docker-mailserver/docker-mailserver/pull/1812#issuecomment-781208653), but we've finally done it :stuck_out_tongue: 

This test has been refactored heavily and should address the concerns raised in the comments and past discussions about it. There was a [mention of a `checkkey` feature that could be used, but that requires setting up DNS](https://github.com/docker-mailserver/docker-mailserver/pull/1812#issuecomment-779899562). In future once we do have a DNS service in the tests, I'd rather just do a test that sends and receives mail, while verifying DKIM was handled correctly.

I staged out changes across commits with commit messages in each if anyone wants to compare to the original with easier diffs to follow.

## Feature history

- 2016 - [Originally contributed](https://github.com/docker-mailserver/docker-mailserver/pull/56) in Jan 2016.
  - In Oct 2016 it was [revised along with OpenDMARC](https://github.com/docker-mailserver/docker-mailserver/pull/361).
  - Ports 8891 and 8893 chosen based on [this blog post](https://web.archive.org/web/20210802220251/https://www.stevejenkins.com/blog/2015/03/installing-opendmarc-rpm-via-yum-with-postfix-or-sendmail-for-rhel-centos-fedora/).
- 2017 - [Creating the `KeyTable` config file to avoid start-up issues with `opendkim`](https://github.com/docker-mailserver/docker-mailserver/pull/627) (_the [associated issue](https://github.com/docker-mailserver/docker-mailserver/issues/621) provides actual details_).
  - **NOTE:** I experienced similar with a change in this PR that had `supervisord` restart the `opendkim` process over 4k times spamming error logs with a related failure once the default opendkim config folder for tests was renamed.
- 2017 [`open-dkim domain` feature](https://github.com/docker-mailserver/docker-mailserver/pull/690).
  - Jan 2021 - This was [refactored to combine two separate but similar opendkim utils](https://github.com/docker-mailserver/docker-mailserver/pull/1753).
    - **NOTE:** DKIM default key length was [changed from 2048 to 4096 without proper justification](https://github.com/docker-mailserver/docker-mailserver/pull/1753#pullrequestreview-576855284).. March 2021 had a [discussion to revert back to `2048` default](https://github.com/docker-mailserver/docker-mailserver/issues/1854#issuecomment-800679841), but no action has been taken since (_I'll probably resolve that after this PR_).
[...]

PR #3072, opened 2023-02-08
ci(fix): Temporarily avoid specifying `provenance`

polarathene:
# Description

As the test workflow does not use the `docker-container` buildx driver (_our build and publish workflows do_), it uses the Docker Engine bundled BuildKit version (_BuildKit release over 2 years old_) which until the recent v23 release (Feb 2023) does not support attestations. Github should have v23 available in runners sometime this month.

This driver difference is causing the current CI failures for test workflow runs:

- `docker/build-push-action` v4 will only [add the `--provenance` option to buildx command, if explicitly configured in the action](https://github.com/docker/build-push-action/pull/784/files). By not adding a value we can avoid causing this failure happening in buildx.
- The option is only valid in buildx `0.10.0` when there is a version of BuildKit that supports it. The option will cause buildx to fail early if the feature is not available the detected BuildKit (_only when supported, does it acknowledge an opt-out_). This appears to be resolved in newer buildx releases, at least `0.10.2`.

---

As the test workflows don't do anything with Docker beyond re-use the build cache to import a local image for testing, the value of `--provenance` isn't relevant to this workflow regardless.

Alternative options considered were:
- Also use a `docker-container` driver to always use the same BuildKit version (_only during the rebuild of the image, not that relevant to this workflow_).
- Continue to allow CI failure until Github's updates to Docker Engine (_expected to arrive within a week_).
- Reverting the action version won't make a difference, [previously we used v3.3](https://github.com/docker-mailserver/docker-mailserver/pull/3008) for the past 3 weeks (_It by [default enabled provenance when supported](https://github.com/docker/build-push-action/pull/781#issue-1559752499)_). It was just the addition of `--provenance` option to buildx.

## References

- CI failure investigation: https://github.com/docker-mailserver/docker-mailserver/pull/3060#issuecomment-1421662676
- Upstream issue: https://github.com/docker/buildx/issues/1608
- Upgrade of `docker/build-push-action` to v4 PR that introduced the issue (_due to `provenance: false`, not v4_): https://github.com/docker-mailserver/docker-mailserver/pull/3066

## Type of change

### DIFF
--- a/.github/workflows/generic_test.yml
+++ b/.github/workflows/generic_test.yml
@@ -34,12 +34,14 @@ jobs:
           path: /tmp/.buildx-cache
           key: cache-buildx-${{ inputs.cache-key }}
 
+      # Configures buildx to use `docker-container` driver,
+      # Ensures consistent BuildKit version (not coupled to Docker Engine),
+      # and increased compatibility of the build cache vs mixing buildx drivers.
       - name: 'Set up Docker Buildx'
         uses: docker/setup-buildx-action@v2.4.0
 
       # Importing from the cache should create the image within approx 30 seconds:
-      # NOTE: Earlier `buildx` + `qemu` steps are not needed as no cache is exported,
-      # and only a single platform (AMD64) is loaded:
+      # NOTE: `qemu` step is not needed as we only test for AMD64.
       - name: 'Build AMD64 image from cache'
         uses: docker/build-push-action@v4.0.0
         with:
--- a/.github/workflows/generic_test.yml
+++ b/.github/workflows/generic_test.yml
@@ -34,8 +34,11 @@ jobs:
           path: /tmp/.buildx-cache
           key: cache-buildx-${{ inputs.cache-key }}
 
+      - name: 'Set up Docker Buildx'
+        uses: docker/setup-buildx-action@v2.4.0
+
       # Importing from the cache should create the image within approx 30 seconds:
-      # Earlier `buildx` + `qemu` steps are not needed as no cache is exported,
+      # NOTE: Earlier `buildx` + `qemu` steps are not needed as no cache is exported,
       # and only a single platform (AMD64) is loaded:
       - name: 'Build AMD64 image from cache'
         uses: docker/build-push-action@v4.0.0
@@ -47,9 +50,8 @@ jobs:
           # Rebuilds the AMD64 image from the cache:
           platforms: linux/amd64
           cache-from: type=local,src=/tmp/.buildx-cache
-          # TODO: Enable this once Docker Engine is v23 or buildx is consistently >=0.10.2
           # Disable provenance attestation: https://docs.docker.com/build/attestations/slsa-provenance/
-          # provenance: false
+          provenance: false
 
       - name: 'Run tests'
         run: make generate-accounts tests/${{ matrix.part }}
--- a/.github/workflows/generic_test.yml
+++ b/.github/workflows/generic_test.yml
@@ -47,8 +47,9 @@ jobs:
           # Rebuilds the AMD64 image from the cache:
           platforms: linux/amd64
           cache-from: type=local,src=/tmp/.buildx-cache
+          # TODO: Enable this once Docker Engine is v23 or buildx is consistently >=0.10.2
           # Disable provenance attestation: https://docs.docker.com/build/attestations/slsa-provenance/
-          provenance: false
+          # provenance: false
 
       - name: 'Run tests'
         run: make generate-accounts tests/${{ matrix.part }}


georglauterbach (approved):

wernerfred:
I'm not sure if i 100% get the point? Wouldn't that mean that that all tests should always have failed? So if not, why would they fail only sometimes? And how will we make sure to enable it in the future again (what i want to say is: if it will be available sometime this month isn't it easier to wait that 20days instead of disabling and then thinking of re-adding it again)?

casperklein:
>Alternative options considered were:
>Also use a docker-container driver to always use the same BuildKit 

I would opt for this, if that's easily possible.

polarathene:
@wernerfred 

> I'm not sure if i 100% get the point? Wouldn't that mean that that all tests should always have failed?

All tests will fail presently now since we added the `provenance: false` to our workflow.

As explained this adds the `--provenance` option to `buildx` command, which the current runners use buildx `0.10.0` and that fails early due to BuildKit in current Docker Engine being too old.

> So if not, why would they fail only sometimes?

I have seen some runners use buildx `0.10.2` which has a fix to acknowledge `--provenance false`, and not do a BuildKit compatibility check when `--provenance` option is provided (regardless of value), those runs will pass.

This wasn't an issue prior to us adding `provenance: false`, as the `--provenance` option was not added to buildx command, which then AFAIK has buildx only enable attestations (`--provenance true`) if the BuildKit version supported it. Basically `--provenance` with buildx `0.10.0` was buggy, but since been fixed, just Github CI runners haven't all been updated to it yet.

> how will we make sure to enable it in the future again
> what i want to say is: if it will be available sometime this month isn't it easier to wait that 20days instead of disabling and then thinking of re-adding it again

We've chosen to opt-out intentionally already in build and publish. It's not actually relevant to the test workflow, so opt-in/out explicitly (_or omit it for an implicit enable when supported_) won't matter. It's only presently an issue due to the bug with explicit `provenance` setting and the current Docker Engine + buildx versions in the CI.

Having an explicit `provenance: false` in the test workflow later would only be to maintain consistency with the other workflows if maintainers prefer that (_despite it not having any relevance to testing_).

Leaving it as-is, means our CI will fail the test suite for that duration...I'm not sure why you'd be okay with 20 days of tests failing due to this? If you're concerned about someone forgetting, we can raise an issue as a reminder and the stale bot will ping any maintainer subscribed as a reminder.

Or we just leave it disabled / removed :man_shrugging:

polarathene:
> > Alternative options considered were:
> > Also use a docker-container driver to always use the same BuildKit
> 
> I would opt for this, if that's easily possible.

@casperklein yeah no worries we can do that I think. `docker-container` driver is only necessary for some other benefits like supporting ARM64 builds, and exporting cache features IIRC. `COPY --link` also needs at least BuildKit `0.10.0` (_not to be confused with buildx version_) from March 2022. Docker Engine v23 will provide that. For the tests it's not important though.

---

So to clarify:

- test workflow does not need `docker-container` driver.
- nor does it need to opt out or in for attestations (`--provenance`) as the image is discarded after tests, and this feature does not affect tests, only image builds to publish to registries.
- Github CI versions of Docker Engine + buildx are the cause of failure with `--provenance` option, and that'll likely be resolved by Github within a week or so.

If you'd like more consistency, I can bring in the buildx action into the test workflow as well (_the qemu one isn't necessary_), which configures buildx with the `docker-container` driver by default. The only difference there is a newer version of BuildKit is used, but the build workflow has already built the image for us.

While it's not necessary, it may look easier to maintain that way if it's more consistent with the build and publish workflows. `docker/build-push-action` will still differ across all three though, as it should; but we can then have a consistent `provenance: false` if that is what other maintainers are comfortable with :+1:

polarathene:
Docker buildx docs:
- [`docker-container` driver](https://docs.docker.com/build/drivers/docker-container/#loading-to-local-image-store). Mentions `--load` being required. It's possible that using this driver to import the image build from the build container to the host Docker daemon will be slower due to a copy. CI time may increase.
- [`docker` driver](https://docs.docker.com/build/drivers/docker/) (default), doesn't actually seem to require `--load`, thus we probably could have avoided `load: true` in the action config, or it may have been a no-op for this driver. I recall CI reading in the build cache and loading the image within 30 sec.

Switching to `docker-container` driver is probably an improvement actually. I just reviewed some prior CI runs when they were working, and observed that while the build workflow cached the image and re-built withn 30 secs, the `docker` driver in the tests was not caching the same layers :sweat_smile: Instead taking 90+ sec to re-run package install scripts, while keeping the same cache key, thus not uploading a new cache. Whoops.

casperklein (approved):

wernerfred:
> All tests will fail presently now since we added the provenance: false to our workflow.

@polarathene thanks for your detailed answer and clarification! I've read through the other PR as well and now my original comment doesn't make sense anymore 😆

wernerfred (approved):

Issue #3098, opened 2023-02-17
[BUG] ClamAV security vulnerability


GoliathLabs:
### Miscellaneous first checks

- [X] I checked that all ports are open and not blocked by my ISP / hosting provider.
- [X] I know that SSL errors are likely the result of a wrong setup on the user side and not caused by DMS itself. I'm confident my setup is correct.

### Affected Component(s)

ClamAV

### What happened and when does this occur?

See: https://blog.clamav.net/2023/02/clamav-01038-01052-and-101-patch.html

### What did you expect to happen?

We should update our debian base image

### How do we replicate the issue?

1.
2.
3.
...

### DMS version

latest

### What operating system is DMS running on?

Linux

### Which operating system version?

Debian 11

### What instruction set architecture is DMS running on?

AMD64 / x86_64

### What container orchestration tool are you using?

Docker

### Other relevant information

https://blog.clamav.net/2023/02/clamav-01038-01052-and-101-patch.html

### What level of experience do you have with Docker and mail servers?

- [ ] I am inexperienced with docker
- [ ] I am rather experienced with docker
- [ ] I am inexperienced with mail servers
- [ ] I am rather experienced with mail servers
- [ ] I am uncomfortable with the CLI
- [ ] I am rather comfortable with the CLI

### Code of conduct

- [X] I have read this project's [Code of Conduct](https://github.com/docker-mailserver/docker-mailserver/blob/master/CODE_OF_CONDUCT.md) and I agree
- [X] I have read the [README](https://github.com/docker-mailserver/docker-mailserver/blob/master/README.md) and the [documentation](https://docker-mailserver.github.io/docker-mailserver/edge/) and I searched the [issue tracker](https://github.com/docker-mailserver/docker-mailserver/issues?q=is%3Aissue) but could not find a solution

georglauterbach:
Can you elaborate on what's _exactly_ the issue is / which CVE you're referring to? And is a base image Update really justified?

kub3let:
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2023-20032

https://blog.clamav.net/2023/02/clamav-01038-01052-and-101-patch.html

There is no patch in debian repositories yet https://security-tracker.debian.org/tracker/CVE-2023-20032

Solution would be to add clamav 1.0.1 from https://www.clamav.net/downloads, however there are no arm builds.

[...]

PR #3106, opened 2023-02-22
ci/docs: add vulnerability scanning workflow & security policy

georglauterbach:
# Description

Title says it all, commits descriptions have proper comments :)

## Type of change

## Checklist:

### DIFF
--- a/.github/workflows/generic_vulnerability-scan.yml
+++ b/.github/workflows/generic_vulnerability-scan.yml
@@ -36,6 +36,21 @@ jobs:
           path: /tmp/.buildx-cache
           key: cache-buildx-${{ inputs.cache-key }}
 
+      # Importing from the cache should create the image within approx 30 seconds:
+      # NOTE: `qemu` step is not needed as we only test for AMD64.
+      - name: 'Build AMD64 image from cache'
+        uses: docker/build-push-action@v4.0.0
+        with:
+          context: .
+          tags: mailserver-testing:ci
+          # Export the built image to the Docker host for later use:
+          load: true
+          # Rebuilds the AMD64 image from the cache:
+          platforms: linux/amd64
+          cache-from: type=local,src=/tmp/.buildx-cache
+          # Disable provenance attestation: https://docs.docker.com/build/attestations/slsa-provenance/
+          provenance: false
+
       - name: 'Run the Anchore Grype scan action'
         uses: anchore/scan-action@v3.3.4
         id: scan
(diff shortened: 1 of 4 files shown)


.github/workflows/generic_vulnerability-scan.yml
```diff
@@ -0,0 +1,64 @@
...
+        with:
+          context: .
+          tags: mailserver-testing:ci
+          # Export the built image to the Docker host for later use:
+          load: true
+          # Rebuilds the AMD64 image from the cache:
+          platforms: linux/amd64
+          cache-from: type=local,src=/tmp/.buildx-cache
+          # Disable provenance attestation: https://docs.docker.com/build/attestations/slsa-provenance/
+          provenance: false
```
polarathene, line 52: As per advice from the [associated issue that spurred this PR](https://github.com/docker-mailserver/docker-mailserver/issues/3098#issuecomment-1436117625), the cache is built with the `docker-container` buildx driver, this requires an additional step to configure (_which the `docker/build-push-action` then imports to the Docker host via `load: true`_). We had to do this with [`generic-test.yml` recently for consistent environment](https://github.com/docker-mailserver/docker-mailserver/pull/3072).

```suggestion
      # Get the cached build layers from the build job:
      # This should always be a cache-hit, thus `restore-keys` fallback is not used.
      # No new cache uploads should ever happen for this job.
      - name: 'Retrieve image built from build cache'
        uses: actions/cache@v3
        with:
          path: /tmp/.buildx-cache
          key: cache-buildx-${{ inputs.cache-key }}

      # Configures buildx to use `docker-container` driver,
      # Ensures consistent BuildKit version (not coupled to Docker Engine),
      # and increased compatibility of the build cache vs mixing buildx drivers.
      - name: 'Set up Docker Buildx'
        uses: docker/setup-buildx-action@v2.4.1

      # Importing from the cache should create the image within approx 30 seconds:
      # NOTE: `qemu` step is not needed as we only test for AMD64.
      - name: 'Build AMD64 image from cache'
        uses: docker/build-push-action@v4.0.0
        with:
          context: .
          tags: mailserver-testing:ci
          # Export the built image to the Docker host for later use:
          load: true
          # Rebuilds the AMD64 image from the cache:
          platforms: linux/amd64
          cache-from: type=local,src=/tmp/.buildx-cache
          # Disable provenance attestation: https://docs.docker.com/build/attestations/slsa-provenance/
          provenance: false
```

polarathene (changes requested):

polarathene (approved):
Thanks for taking the time to sort this out :grinning:

PR #3111, opened 2023-02-23
added plausibility checks for milter insertion

georglauterbach:
# Description

Makes the insertion of milters more robust.

Fixes #3109

## Type of change

- [x] Improvement (non-breaking change that does improve existing functionality)

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [ ] If necessary I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #3112, opened 2023-02-24
scripts: cleanup and prep for setup split

### LINKED ISSUE #3109: [BUG] mail from client to server slow, perhaps opendkim causing delays

### Affected Component(s)

it sometimes takes too long to send an email form thunderbird 

When sending email from thunderbird sometimes email sending takes 10 seconds. 
When analysing logs opendkim logs 4x for every outgoing email. Sometimes this is within 2 seconds, but sometimes this also takes 12 seconds. 

So question is why does opendkim log 4x times it is signing something? And can this be the cause for the delays? 

Could it be that because I have four active domains this is a bug? (wild quess of course)
 
```
root@mail [ ~ ]# egrep "E00D761FF6"  /var/lib/docker/volumes/mailserver_log/_data/mail/mail.log 
Feb 21 10:50:22 mail postfix/smtps/smtpd[599924]: E00D761FF6: client=unknown[192.168.0.185], sasl_method=PLAIN, sasl_username=line@pvandenboom.com
Feb 21 10:50:22 mail postfix/sender-cleanup/cleanup[599935]: E00D761FF6: message-id=<558ae5ed-50cb-f4d6-f042-89f28e6b7b4b@jvandenboom.com>
Feb 21 10:50:22 mail postfix/sender-cleanup/cleanup[599935]: E00D761FF6: replace: header MIME-Version: 1.0 from unknown[192.168.0.185]; from=<line@pvandenboom.com> to=<vanuden@gmail.com> proto=ESMTP helo=<[192.168.0.185]>: MIME-Version: 1.0
Feb 21 10:50:27 mail opendkim[651]: E00D761FF6: DKIM-Signature field added (s=mail, d=pvandenboom.com)
Feb 21 10:50:31 mail opendkim[651]: E00D761FF6: DKIM-Signature field added (s=mail, d=pvandenboom.com)
Feb 21 10:50:35 mail opendkim[651]: E00D761FF6: DKIM-Signature field added (s=mail, d=pvandenboom.com)
Feb 21 10:50:39 mail opendkim[651]: E00D761FF6: DKIM-Signature field added (s=mail, d=pvandenboom.com)
Feb 21 10:50:39 mail postfix/qmgr[228056]: E00D761FF6: from=<line@pvandenboom.com>, size=960, nrcpt=1 (queue active)
Feb 21 10:50:40 mail postfix/smtp[599997]: E00D761FF6: to=<vanuden@gmail.com>, relay=gmail-smtp-in.l.google.com[142.250.27.26]:25, delay=17, delays=17/0.01/0.16/0.62, dsn=2.0.0, status=sent (250 2.0.0 OK  1676972353 eu17-20020a170907299100b008b17fe8a06bsi12931520ejc.70 - gsmtp)
Feb 21 10:50:40 mail postfix/qmgr[228056]: E00D761FF6: removed
```

by the way, it looks like opendkim does not actually sign an email 4x, if I look at the email arriving at gmail. 

```
Delivered-To: hanscees.speel@gmail.com
Received: by 2002:a05:6638:3475:0:0:0:0 with SMTP id q53csp248368jav;
        Wed, 22 Feb 2023 13:10:31 -0800 (PST)
X-Google-Smtp-Source: AK7set9xRMzUytoinyieatCgy33Kp0wpb5NLtAI+C/Bx/sS7MhYOGnHhP6Du69X/vJNEalQxqRQA
X-Received: by 2002:a17:906:10da:b0:8b1:94cc:10b1 with SMTP id v26-20020a17090610da00b008b194cc10b1mr17180085ejv.7.1677100231346;
        Wed, 22 Feb 2023 13:10:31 -0800 (PST)
ARC-Seal: i=1; a=rsa-sha256; t=1677100231; cv=none;
        d=google.com; s=arc-20160816;
        b=R3TXZBdkskKsurXSlLg7xXjJIltwSV6vRUE3BUCmAAoN9fbXm5RIayBQLGptuxz/F+
         JUmhqcJhp12AE8U/qVttKSjQyfgQx8S4KPXzmt/0Ut53cGBEZfSvoVMEcUM+pGZCm65z
         SHEGXZUoCsc/DXy6BxpRZ1gmSSHykanqdcnL2X3Ub0H+na6BM4u7sszUJJhy+NkZRw25
         EWDUu8yDRb0Fj0fiKHzM35tZfPA4ZUg5LrWj2XN/0+nvr0NlHpNcT5ln9NrWscM5U5Zm
         FUzca9ZTbxVWD6olJrmWq8eEkol5FTtpNTkSA4WkXAJy+tvJbsnTISiSIx0aH1EWDBlD
         8Xeg==
ARC-Message-Signature: i=1; a=rsa-sha256; c=relaxed/relaxed; d=google.com; s=arc-20160816;
        h=content-transfer-encoding:subject:from:to:content-language
         :mime-version:date:dkim-signature:message-id;
        bh=C7aE0Y03L7U8Dn/rejS9nwNm3CWZUJigYYLvfep9ml4=;
        b=dsLG4m0GvIxSwTcO0/esCG2Bjz7Zsm28Qk9djcOgY8+UwWpq6kkYYlwRhfVbZRi6CF
         +kygImI3gU4B2CoCV4WKFoJAxEnW+/Drj6cSz+2cL6PrxIFitlwKxbRaXz70ZXbPmIpQ
         xYJN8+pCjRYQyVa446b02XdoXcXQCsU/Nz9nKeZThMTZ+mJM9QjGDQHsxKzBQzNJ6H3B
         BjwEQkzJX0MA14dXS0KTRgOoj3hYDy9nbWbME8ciM5IBctPedyq4yt5tpKaCkslyTghb
         TvVJykrap18PqZUQHByPgn3KJO6zjxdZSZuVn2KVTD0y8Xxsw/rOSNzFSXRmmjTraHdA
         lN/g==
ARC-Authentication-Results: i=1; mx.google.com;
       dkim=pass header.i=@hanscees.com header.s=mail header.b=mgmNQ0QQ;
       spf=pass (google.com: domain of hanscees@hanscees.com designates 213.233.217.138 as permitted sender) smtp.mailfrom=hanscees@hanscees.com
Return-Path: <hanscees@hanscees.com>
Received: from mail.hanscees.net (mail.hanscees.net. [213.233.217.138])
        by mx.google.com with ESMTPS id vs22-20020a170907139600b008df5a5088cesi5632683ejb.163.2023.02.22.13.10.30
        for <hanscees.speel@gmail.com>
        (version=TLS1_3 cipher=TLS_AES_256_GCM_SHA384 bits=256/256);
        Wed, 22 Feb 2023 13:10:31 -0800 (PST)
Received-SPF: pass (google.com: domain of hanscees@hanscees.com designates 213.233.217.138 as permitted sender) client-ip=213.233.217.138;
Authentication-Results: mx.google.com;
       dkim=pass header.i=@hanscees.com header.s=mail header.b=mgmNQ0QQ;
       spf=pass (google.com: domain of hanscees@hanscees.com designates 213.233.217.138 as permitted sender) smtp.mailfrom=hanscees@hanscees.com
Message-ID: <4fa8cb94-c7cd-1221-26c5-4b03a35ab1e0@hanscees.com>
DKIM-Signature: v=1; a=rsa-sha256; c=relaxed/simple; d=hanscees.com; s=mail;
	t=1677100920; bh=C7aE0Y03L7U8Dn/rejS9nwNm3CWZUJigYYLvfep9ml4=;
	h=To:From:Subject;
	b=mgmNQ0QQ9kS5ZcA/cBMGRbL1hYw/5YnwYsBr7+TAd9Ac942y9u3NuNibZRRdYGwc/
	 Zixh2Bof7uGyMvoAoYCx+Ti8mgoC4HWr3ZAt5M6a/B5e3En1uThxsJGtC6xICkacRM
	 YPumusJu/NxcKwPCraImT5GXt4KW/4atQwgxOIzjw3e/FCia3mnMx9X87+ODsJ4poJ
	 Rxx9ZwclnJ4PgSX25Iw8ILM41dHMCar3H+M+fLJfolCVLIT5LjSqmhLffbsVFTrKy0
	 Hmovs7rpLwUHdaAZJ1rnQrVmwmU9/rneT0K71EloA1KiTzIc3yrTFpAZqjG4WOz/Ja
	 jGRWxnhdWESsg==
Date: Wed, 22 Feb 2023 22:10:30 +0100
MIME-Version: 1.0
Content-Language: en-US
To: Hanscees <hanscees.speel@gmail.com>
From: HCImap <hanscees@hanscees.com>
Subject: test for opendkim
Content-Type: text/plain; charset=UTF-8; format=flowed
Content-Transfer-Encoding: 7bit



test for opendkim
```

 

I expect sending mail not to take 12 seconds. 
I expect opendkim only signs 1x in the logging

I dont know. 

You can try to add 4 domains and see if this perhaps causes the multiple signing. 

### DMS version

1301

Linux

vmware photon 4

AMD64 / x86_64

Docker

### docker-compose.yml

```yml
services:
  mailserver:
    image: docker.io/mailserver/docker-mailserver:edge
    #image: docker.io/mailserver/docker-mailserver:latest
    container_name: mailserver
    hostname: mail
    domainname: fanscees.net
    env_file: mailserver.env
    ports:
      - "995:995"  # pop3 (implicit TLS)
      - "110:110"  # pop3 (implicit TLS)
    volumes:
      #- etc:/etc
      - maildata:/var/mail/
      - mailstate:/var/mail-state/
      - log:/var/log/
      - Tconfig:/tmp/docker-mailserver/
      - ssl:/tmp/ssl
    restart: always
    stop_grace_period: 1m
    environment:
     # If you need SSL connection, you can provide your own certificates
     - SSL_TYPE=manual
    cap_add:
      - NET_ADMIN
    healthcheck:
      timeout: 3s
      retries: 0

volumes:
  maildata:
  mailstate:
  Tconfig:
  log:
  ssl:
```

### Relevant log output

```Text
see above
```

several domains with aliases are present

```
root@mail [ ~/containers/mailserver ]# ./setup.sh email list
* hanscees@fanscees.com
    [ aliases -> hanscees@fanscees.net, bomengidsnl@fanscees.net, bomengidsnl@bomengids.nl ]

* line@pvandenboom.com
    [ aliases -> info@pvandenboom.com ]

* root@fanscees.net
    [ aliases -> postmaster@fanscees.net ]

```

### Code of conduct

_No response_

georglauterbach:
# Description

A PR including changes I wanted to do for a longer time; the last before my vacation march and before v12 is released.

This PR opens the possibility to split the big `setup-stack.sh` into smaller pieces in a dedicated directory. **Moreover**, it simplifies scripts by a good amount (getting rid of the fixes and misc stacks completely).

The single new feature of this PR is important though: there is now a check whether the container was improperly restarted, and if so, the container will show an error and shut down. This makes sense, since we cannot guarantee anything in case of a bad restart, and IMO it is better to not run a possibly faulty configuration at all than to run with a faulty config (as seen in #3109).

This PR has its first commit based off of #3111. For the other commits I tried to provide a nice message so you see the rationale/changes more easily.

Fixes #3109

## Type of change

## Checklist:


target/scripts/startup/setup-stack.sh
```diff
@@ -940,16 +971,20 @@ EOF
 
       chmod +x "${SPAMASSASSIN_KAM_CRON_FILE}"
     fi
+  else
+    _log 'debug' 'Disabling SpamAssassin'
+    echo "@bypass_spam_checks_maps = (1);" >>"${DMS_AMAVIS_FILE}"
+    rm -f /etc/cron.daily/spamassassin
```
polarathene, line 977: No action required, can leave as-is. This is just a review comment note.

Both lines should be kept as-is, since there is no harm and if we were to allow bypassing the new fail fast check, then these fixes remain relevant (_and would need to be realized again_).

---

This `rm -f` along with the ones below for ClamAV are both migrated from `fixes-stack.sh`, which AFAIK was intended for addressing concerns with container restarts.

The intent of this PR is to disallow that, so this internal state should not be present to cleanup (_unless it exists from the `Dockerfile`_).

target/scripts/startup/setup-stack.sh
```diff
@@ -1259,3 +1294,110 @@ function _setup_timezone
...
+function _setup_apply_fixes
+{
+  _log 'trace' 'Removing leftover PID files from a stop/start'
+  find /var/run/ -not -name 'supervisord.pid' -name '*.pid' -delete
+  touch /dev/shm/supervisor.sock
+
+  _log 'debug' 'Checking /var/mail permissions'
+  _chown_var_mail_if_necessary || _shutdown 'Failed to fix /var/mail permissions'
+  _log 'trace' 'Permissions in /var/mail look OK'
+}
```
polarathene, line 1307: No action required, can leave as-is. This is just a review comment note.

---

These fixes are a bit different.

- Supervisor fix probably is similar to the SA and ClamAV ones, and not relevant to reload in `_setup_supervisor()`? This would be run before `daemons.stack.sh` before, but presumably could be part of `_start_daemons()`?
- `/var/mail` chown still relevant. Technically, this is specific to Dovecot AFAIK, but I'm not 100% sure.

target/scripts/startup/setup-stack.sh
```diff
@@ -1259,3 +1294,110 @@ function _setup_timezone
...
+    chown -R root /var/mail-state/spool-postfix/lib
+    chown -R root /var/mail-state/spool-postfix/pid
+    chown -R root /var/mail-state/spool-postfix/usr
+  elif [[  ${ONE_DIR} -eq 1 ]]
+  then
+    _log 'warn' "ONE_DIR is enabled (=1), but '${STATEDIR}' does not exist - is a volume mounted there?"
+  else
+    _log 'debug' "Mail state is not consolidated into one directory"
+  fi
+}
```
polarathene, line 1403: Suggested action: Extract into separate `setup.d` file, or keep `misc-stack.sh`

Technically fitting of `setup.d/one-dir.sh`, as it's an integration feature to support the `ONE_DIR` ENV paired with an external volume. As a purely integration focused feature though, it's a bit distinct from other `setup-stack.sh`, and is serving two functions (_`ONE_DIR=1` symlinks to `/var/mail-state`, and ensuring permissions/ownership is correct across restarts, notably during updating DMS releases when user and group IDs may have changed and need to adjust attributes of externally persisted state_).

---

The only change here from `misc-stack.sh` really is the extra conditional handling at the end to output a warning or debug log message.

I am a tad confused though. With the [commit expressing intent to split `setup-stack.sh` into smaller files for better scoping](https://github.com/docker-mailserver/docker-mailserver/pull/3112/commits/980d0a7b4662481cf51ed8835342abb1338cfdc4), you later [deleted `misc-stack.sh` to add it here in `setup-stack.sh`](https://github.com/docker-mailserver/docker-mailserver/pull/3112/commits/a65d9299bd451de4df8ebd5d05368fcf29a8a414), despite it being a good candidate as a separate file?

Personally, I'd envision this being refactored in future if we had better scoping of features. I preferred `misc-stack.sh` as a separate stage in it's current form, but would like services in setup to have a more clear cut scope where they'd have some convention to do their config setup, fix permissions/ownership like this method does, etc. At least for me it seems nicer if the bulk of Amavis + SA support and rspamd for example were mostly self-contained within their own folders.

test/tests/serial/sedfile.bats
```diff
@@ -36,38 +36,6 @@ function setup() {
...
-
-  # file unchanged?
-  _run_in_container cat "${TEST_FILE}"
-  assert_success
-  assert_output 'foo bar'
-
-  # recreate marker
-  _run_in_container touch '/CONTAINER_START'
-  assert_success
-}
```
polarathene, line 57: Are you sure this test case should be removed?

It's not related to DMS restarts. You'll notice the start and end of the test case ensures `CONTAINER_START` is not present during the test, so `SKIP_ERROR` condition would not have applied here regardless. It ensures that `sedfile` should fail when expected to.

It does look like this test would become redundant, but so does the whole `--strict` + `SKIP_ERROR` logic of `sedfile`. You'd want to [revert that feature](https://github.com/docker-mailserver/docker-mailserver/commit/29c2d975ec40a8a7d26814b473fa0c44d2b95b6a#diff-39578133bff98a4e812dd8ac4469fa3e0ee924e7d4028f549a100170dad0647c) entirely then?

target/bin/sedfile
```diff
@@ -26,7 +26,6 @@ then
   exit 1
 fi >&2
 
-[[ -f /CONTAINER_START ]] && SKIP_ERROR=1 # hide error if container was restarted
```
polarathene, line 29: You should technically be able to keep this and the two related sedfile test cases? The tests only simulate a restart by adjusting the presence of `CONTAINER_START`. Your fail early check wouldn't conflict with that.

polarathene (changes requested):
There is a lot of shuffling going on around here that is distinct from other logic being introduced into the PR.

That complicates review a little bit, but I understand that you're trying to get the changes through review faster (_it'd probably have been faster for me to review if it was staged out as separate PRs instead of only as commits :sweat_smile: _)

---

1. The new preventative check is interesting, but seems a bit heavy-handed personally. I'd like the perspective of another maintainer before I provide approval. I do understand the intent and value of it.
2. `sedfile` changes don't seem to be necessary to drop right now. Just like the `fixes-stack.sh` logic that is being kept in `setup-stack.sh`, which is mostly focused on fixes for persisted internal state for container restarts AFAIK (except `/var/mail`).
3. Not fond of the early methods added to `setup-stack.sh` and `check-stack.sh`, I don't think these are necessary or need to be there.
4. `misc-stack.sh` should live in `setup.d/` if you insist on moving it, not blended into `setup-stack.sh` methods, you have an entire commit in this PR introducing `setup.d/` for the purpose of avoiding a large `setup-stack.sh` file.
5. I have discussed in some review comments certain changes that may be better served by follow-up PRs to avoid all the other noise of this PR affecting relevance to their proposed changes. Such as when `_setup_user_patches()` may be better suited for running.

georglauterbach:
@polarathene you're right as usual; I will try to answer all of the comments now. I have opened #3115 though, which is the first part of a series of PRs that do the same as this PR but incrementally. This should make reviewing easier.

target/scripts/start-mailserver.sh
```diff
@@ -19,38 +13,17 @@ source /usr/local/bin/check-stack.sh
...
-# shellcheck source=./helpers/variables.sh
-source /usr/local/bin/helpers/variables.sh
-
-_setup_supervisor
-_obtain_hostname_and_domainname
-_environment_variables_backwards_compatibility
-_environment_variables_general_setup
-
-# ------------------------------------------------------------
-# ? << Early setup & environment variables setup
```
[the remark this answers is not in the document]
georglauterbach: With an upcoming PR after #3115 I have tried to find a solution (or a compromise if you will), for this issues. Stay tuned :)

target/scripts/startup/setup-stack.sh
```diff
@@ -1,5 +1,33 @@
 #!/bin/bash
 
+# shellcheck disable=SC2034
+declare -A VARS
```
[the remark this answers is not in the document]
georglauterbach: This was done due to a Bash having _very weird_ scoping. With #3115 and the follow-up, this is solved since `variables.sh` will become a proper "stack", so the sourcing will be done by `start-mailserver.sh`, which in turn makes declaring this array possible in `variables.sh` (then: `variables-stack.sh`).

target/scripts/startup/setup-stack.sh
```diff
@@ -1,5 +1,33 @@
...
+# shellcheck source-path=target/scripts/startup/setup.d
+
+# Runs early setup routines
+function _run_early_setup
+{
+  _log 'debug' 'Running early setup'
+
+  local PATH_TO_SCRIPTS='/usr/local/bin/setup.d/'
+
+  source "${PATH_TO_SCRIPTS}/rspamd.sh"
```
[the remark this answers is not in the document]
georglauterbach: My bad, the other sourcing should not happen. Resolved in #3115.

target/scripts/startup/setup-stack.sh
```diff
@@ -10,6 +38,10 @@ function _setup
 
   # All startup modifications to configs should have taken place before calling this:
   _prepare_for_change_detection
+  # Run user-patches
+  __setup_user_patches
```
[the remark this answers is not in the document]
georglauterbach: I have actually no clue why an underscore was prepended.. again, should not have happened.

target/scripts/startup/setup-stack.sh
```diff
@@ -10,6 +38,10 @@ function _setup
...
+  # Show the environment if the log level "demands" it
+  [[ ${LOG_LEVEL} =~ (debug|trace) ]] && print-environment
```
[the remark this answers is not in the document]
georglauterbach: I'm am willing to leave it in there; #3115 and follow-ups will not move it.

target/scripts/startup/setup-stack.sh
```diff
@@ -940,16 +971,20 @@ EOF
 
       chmod +x "${SPAMASSASSIN_KAM_CRON_FILE}"
     fi
+  else
+    _log 'debug' 'Disabling SpamAssassin'
```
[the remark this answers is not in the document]
georglauterbach: Nicd catch! After #3115, the follow-up PR will make sure to make all of these messages consistent (the messages itself, and consistent with the default ENV values).

target/scripts/startup/setup-stack.sh
georglauterbach: This will be moved in the follow-up after #3115 into a separate file in `setup.d`. I kept it in `setup.sh` because, yeah, TBH, I don't know either. Should have directly put it into `setup.d`.

test/tests/serial/sedfile.bats
georglauterbach: You're probably, right; in one of the last PRs in the series after #3115, I will try to remeber this and maybe `sedfile` does not need an adjust at all; as you've said.

target/bin/sedfile
georglauterbach: Right; I will test this in the last PR of the series after #3115.

test/tests/serial/sedfile.bats
polarathene: > maybe `sedfile` does not need an adjust at all; as you've said.

I don't think it does :)

It would only needs changes if we wanted to remove any support we've added to (partially) handle starting the container with persisted internal state.

No harm in leaving that around for a while though, easier to have a single PR specifically removing just that if it were a maintenance concern, so that it's self-documenting of what would need to be restored if we did want to bring back support for restarting or restoring stopped containers.

target/scripts/startup/setup-stack.sh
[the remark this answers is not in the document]
polarathene: > This was done due to a Bash having _very weird_ scoping.

Yeah I figured that was the case :sweat_smile: 

I'm particularly interested in your ENV parser rust project having potential to eventually manage not only the ENV documentation, but share the same struct / module for what `variables.sh` is effectively responsible for.

Then we won't have drift, as recently experienced with the `SA_KILL` update to `mailserver.env`.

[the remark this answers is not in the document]
georglauterbach: I was actually thinking about extending the parser this way too :D

georglauterbach:
Closing as #3121 and #3123 are merged.

PR #3115, opened 2023-02-25
scripts: split `setup-stack.sh`

georglauterbach:
# Description

Superseeds parts of #3112. Makes review easier. Split `setup-stack.sh`, did not add / remove any functionality. Basically a glorified copy-paste with minor renamings.

Fixes issue where reviewing #3112 is a chore, I guess.

I will leave #3112 open so I can gradually implement all the changes this single PR would bring. Having reviewed #3112 is good because you basically know what's coming @polarathene, so no time wasted here.

The first commit is basically just copy-paste everywhere, the rest is more renaming/minor adjustments.

This PR is the first part of a series to come about refatoring some parts of the setup with the final goal of disallowing faulty restarts.

---

This PR is **BIG**, but actually 95% copy-paste; there are **no** changes in functionality whatsoever.

## Type of change

## Checklist:


target/scripts/helpers/index.sh
```diff
@@ -10,6 +10,7 @@ function _import_scripts
   source "${PATH_TO_SCRIPTS}/accounts.sh"
   source "${PATH_TO_SCRIPTS}/aliases.sh"
   source "${PATH_TO_SCRIPTS}/change-detection.sh"
+  source "${PATH_TO_SCRIPTS}/dhparams.sh"
```
polarathene, line 13: This files contents could probably live in the `ssl.sh` helper as it's related to that support (ciphers)

polarathene (approved):
Manually compared diffs before/after, LGTM :sweat_smile: :+1: 

Cheers for handling this step as a separate PR :tada:

target/scripts/helpers/index.sh
georglauterbach: I will merge this PR (to get it through), and will apply this feedback in the next, smaller PR :)

target/scripts/startup/setup-stack.sh
```diff
@@ -1,7 +1,21 @@
...
+  _log 'trace' "${1}() registered"
+}
+
 function _setup
 {
+  while read -r FILE
+  do
+    # shellcheck source=/dev/null
+    source "${FILE}"
+  done < <(find /usr/local/bin/setup.d/ -type f)
```
casperklein, line 17: Generally bad practice - files including spaces for example will not be handled correctly. You should add `-print0` to the find command and define a delimiter `-d $'\0'` in the `read` command or just use globbing instead:

`for i in /usr/local/bin/setup.d/*; do source $i; done`

georglauterbach: See #3124.

casperklein: You are right, I overlooked that there is also a subdirectory and not just files.

target/scripts/helpers/index.sh
georglauterbach: See #3124

PR #3124, opened 2023-02-26
scripts: follow up of #3115 (feedback)

georglauterbach:
# Description

Follow-up of #3115 (branch name of this PR wrong, lol). See
<https://github.com/docker-mailserver/docker-mailserver/pull/3115#discussion_r1118077293> for example.

## Type of change

## Checklist:


target/scripts/start-mailserver.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+shopt -s globstar
```
casperklein, line 3: Can this introduce unwanted side-effects with our existing scripts that are called afterwards?

Another option would be, to just use it in the `_setup` function.

georglauterbach: I was wondering about this too. But I left it up to the tests. Seems there is one place we need to adjust. I will do this later.

georglauterbach: Nevermind, was a sporadic BATS issue.

polarathene: Probably does make sense to move it into `_setup` if that is where this is only relevant? I believe we're using it somewhere else for change detection in a similar fashion.

polarathene (approved):

target/scripts/start-mailserver.sh
georglauterbach: I decided against this for two reasons:

1. The effects of `shopt` are not scoped to a function, i.e.

```console
$ cat test.sh
#!/bin/bash

shopt -u globstar
shopt | grep "globstar"

function a() {
  shopt -s globstar
  shopt | grep "globstar"
}

a
shopt | grep "globstar"
$ bash test.sh
globstar     off
globstar     on
globstar     on
```

Hence, setting this inside the whole setup, i.e. changing the behavior at a certain point, is undesirable. We could turn it off again, but actually setting it as fine I think; we may be able to use it in the future and it does not hurt at all.

2. I want Bash adjustments like `shopt` or `set` to be used at the very beginning of scripts so everyone immediately notices them. I plan on introducing `set -u` and `shopt -s inherit_errexit` too, and these should go together at the beginning `start-mailserver.sh`.

casperklein (approved):

target/scripts/start-mailserver.sh
casperklein: I am aware of that. I didn't express it clearly enough. I meant something like this:

function foo() {
  shopt -s globstar
  for i in /foo/**/*; do
		bar
  done
  shopt -u globstar
}

Setting globstar in the function and revert it after usage.

georglauterbach: I actually think that this feature is useful in general, and having it available everywhere is a nice thing :)

PR #3127, opened 2023-02-28
config/ENV: improve Postfix config for spoof protection

### LINKED ISSUE #3117: [BUG] Submitting/Receiving a mail on port 25 generates warning when SPOOF_PROTECTION=1 is set

### Affected Component(s)

postfix

~~With `SPOOF_PROTECTION=1`, the following warning appears in `mail.log` when an email is send:~~
The following warning is shown, when submitting or receiving an email on port 25 and `SPOOF_PROTECTION=1` is set.

```Feb 25 22:47:39 mail postfix/smtpd[4800]: warning: restriction `reject_authenticated_sender_login_mismatch' ignored: no SASL support```

---

`reject_authenticated_sender_login_mismatch` is added here:

https://github.com/docker-mailserver/docker-mailserver/blob/ae05e6a7c379825fb464a648acc04b4074dd547f/target/scripts/startup/setup-stack.sh#L486-L491

No warning.

1. Send or receive an unauthenticated email
2.
3.
...

### DMS version

edge

Linux

Debian 11

AMD64 / x86_64

Docker

### docker-compose.yml

_No response_

### Relevant log output

_No response_

I think this is a regression introduced with https://github.com/docker-mailserver/docker-mailserver/pull/3006

### Code of conduct

_No response_

### PR DESCRIPTION
**Pull request by georglauterbach**
# Description

Make `SPOOF_PROTECTION=1` the default and improve Postfix configuration to eliminate the currently shown warning.

Reviewing commit-by-commit advised.

Fixes #3117

## Type of change

## Checklist:

### DIFF
--- a/target/postfix/main.cf
+++ b/target/postfix/main.cf
@@ -50,10 +50,12 @@ smtpd_helo_restrictions = permit_mynetworks, reject_invalid_helo_hostname, permi
 smtpd_relay_restrictions = permit_mynetworks permit_sasl_authenticated defer_unauth_destination
 smtpd_recipient_restrictions = permit_sasl_authenticated, permit_mynetworks, reject_unauth_destination, check_policy_service unix:private/policyd-spf, reject_unauth_pipelining, reject_invalid_helo_hostname, reject_non_fqdn_helo_hostname, reject_unknown_recipient_domain
 smtpd_client_restrictions = permit_mynetworks, permit_sasl_authenticated, reject_unauth_destination, reject_unauth_pipelining
-smtpd_sender_restrictions = permit_sasl_authenticated, permit_mynetworks, reject_unknown_sender_domain, reject_unknown_client_hostname
+smtpd_sender_restrictions = $dms_smtpd_sender_restrictions
 disable_vrfy_command = yes
 
+# Custom defined parameters for DMS:
 dms_smtpd_sender_restrictions = permit_sasl_authenticated, permit_mynetworks, reject_unknown_sender_domain, reject_unknown_client_hostname
+# Submission ports 587 and 465 support for SPOOF_PROTECTION=1
 mua_sender_restrictions = reject_authenticated_sender_login_mismatch, $dms_smtpd_sender_restrictions
 
 # Postscreen settings to drop zombies/open relays/spam early
--- a/target/postfix/main.cf
+++ b/target/postfix/main.cf
@@ -53,7 +53,8 @@ smtpd_client_restrictions = permit_mynetworks, permit_sasl_authenticated, reject
 smtpd_sender_restrictions = permit_sasl_authenticated, permit_mynetworks, reject_unknown_sender_domain, reject_unknown_client_hostname
 disable_vrfy_command = yes
 
-mua_sender_restrictions = reject_authenticated_sender_login_mismatch, $smtpd_sender_restrictions
+dms_smtpd_sender_restrictions = permit_sasl_authenticated, permit_mynetworks, reject_unknown_sender_domain, reject_unknown_client_hostname
+mua_sender_restrictions = reject_authenticated_sender_login_mismatch, $dms_smtpd_sender_restrictions
 
 # Postscreen settings to drop zombies/open relays/spam early
 postscreen_dnsbl_action = enforce
--- a/target/postfix/main.cf
+++ b/target/postfix/main.cf
@@ -53,6 +53,8 @@ smtpd_client_restrictions = permit_mynetworks, permit_sasl_authenticated, reject
 smtpd_sender_restrictions = permit_sasl_authenticated, permit_mynetworks, reject_unknown_sender_domain, reject_unknown_client_hostname
 disable_vrfy_command = yes
 
+mua_sender_restrictions = reject_authenticated_sender_login_mismatch, $smtpd_sender_restrictions
+
 # Postscreen settings to drop zombies/open relays/spam early
 postscreen_dnsbl_action = enforce
 postscreen_dnsbl_sites =
--- a/target/scripts/startup/setup.d/security/spoofing.sh
+++ b/target/scripts/startup/setup.d/security/spoofing.sh
@@ -25,6 +25,6 @@ function _setup_spoof_protection
   else
     _log 'debug' 'Spoof protection is disabled'
     # shellcheck disable=SC2016
-    postconf 'mua_sender_restrictions = $smtpd_sender_restrictions'
+    postconf 'mua_sender_restrictions = $dms_smtpd_sender_restrictions'
   fi
 }
--- a/target/scripts/startup/setup.d/security/spoofing.sh
+++ b/target/scripts/startup/setup.d/security/spoofing.sh
@@ -6,10 +6,6 @@ function _setup_spoof_protection
   then
     _log 'trace' 'Enabling and configuring spoof protection'
 
-    sed -i \
-      's|smtpd_sender_restrictions =|smtpd_sender_restrictions = reject_authenticated_sender_login_mismatch,|' \
-      /etc/postfix/main.cf
-
     if [[ ${ACCOUNT_PROVISIONER} == 'LDAP' ]]
     then
       if [[ -z ${LDAP_QUERY_FILTER_SENDERS} ]]
@@ -28,5 +24,7 @@ function _setup_spoof_protection
     fi
   else
     _log 'debug' 'Spoof protection is disabled'
+    # shellcheck disable=SC2016
+    postconf 'mua_sender_restrictions = $smtpd_sender_restrictions'
   fi
 }
--- a/docs/content/config/advanced/kubernetes.md
+++ b/docs/content/config/advanced/kubernetes.md
@@ -287,7 +287,7 @@ spec:
         - name: data
           persistentVolumeClaim:
             claimName: data
-        
+
         # certificates
         - name: certificates-rsa
           secret:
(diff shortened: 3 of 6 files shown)

### CODE REVIEW

**Comment by georglauterbach** [comment]
Just FYI: This and #3137 should have high prio for review, and they should be merged next.

**Comment by polarathene** [comment]
[This commit message](https://github.com/docker-mailserver/docker-mailserver/pull/3127/commits/87b4d949ba94b53df882aa4c8c93559cb621ce4b) has more detail regarding changes in this PR, probably should have been part of the PR description for better visibility :sweat_smile: 

The changes in `master.cf` and related scripts to disable that config seems to over-complicate config management.

I touched on the [naming convention from upstream in prior discussions](https://github.com/docker-mailserver/docker-mailserver/issues/3117#issuecomment-1445231943), I'd rather we use those or adopt some convention / section for `main.cf` where custom settings are more obvious. [Upstream advice](https://github.com/vdukhovni/postfix/blob/6876e420275bb44e98827c76306f7ff5481eb192/postfix/conf/master.cf#L26-L34) is to have the submission services configured with `mua` prefix instead of `smtpd`:

```
#     Instead of specifying complex smtpd_<xxx>_restrictions here,
#     specify "smtpd_<xxx>_restrictions=$mua_<xxx>_restrictions"
#     here, and specify mua_<xxx>_restrictions in main.cf (where
#     "<xxx>" is "client", "helo", "sender", "relay", or "recipient").
#  -o smtpd_client_restrictions=
#  -o smtpd_helo_restrictions=
#  -o smtpd_sender_restrictions=
#  -o smtpd_relay_restrictions=
#  -o smtpd_recipient_restrictions=permit_sasl_authenticated,reject
```

So for us that'd be:

```
### master.cf:
-o smtpd_sender_restrictions=$mua_sender_restrictions


### main.cf:
smtpd_sender_restrictions = permit_sasl_authenticated, permit_mynetworks, reject_unknown_sender_domain, reject_unknown_client_hostname

# Custom settings for submission ports 587 and 465:
mua_sender_restrictions = reject_authenticated_sender_login_mismatch, $smtpd_sender_restrictions


### target/scripts/startup/setup.d/security/spoofing.sh:
_log 'debug' 'Disabling spoof protection'
postconf 'mua_sender_restrictions = $smtpd_sender_restrictions'
```

That last change to use `postconf` with `main.cf` instead of `sed` on `master.cf` seems like a much better approach I think?

If we needed more control for the opposite of adding substrings, there was [an example helper I discussed here](https://github.com/docker-mailserver/docker-mailserver/pull/3111#discussion_r1116384635) (_alternatively `postconf -h` can read in the current value too, but needs to invoke [`_adjust_mtime_for_postfix_maincf()`](https://github.com/docker-mailserver/docker-mailserver/blob/b451742f0a6c23bd68fcc952fd3f6b6f9d291e34/target/scripts/helpers/utils.sh#L48-L59)_).

**Comment by polarathene** [comment]
**Let's delay the `SPOOF_PROTECTION=1` change to a future major release please.**

There is still good changes in this PR that look fine to approve, but feedback from my previous message should be addressed (_and adapted for the inverse since `SPOOF_PROTECTION=0` would be the default_).

---

## Justification to delay `SPOOF_PROTECTION=1` by default

I don't think we're ready to switch to `SPOOF_PROTECTION=1` by default? Not without some caveats at least..

Overview of notes in https://github.com/docker-mailserver/docker-mailserver/issues/2819

- **Section:** _`SPOOF_PROTECTION=1` handling for `smtpd_sender_login_maps`_:
  - `unionmap` config table - Inconsistency between LDAP and non-LDAP setups.
  - `/etc/postfix/regexp` always created at startup.
  - Sending via an alias must be permitted for the actual SASL account (_of which there could be multiple to an alias, or an alias of an alias_ :dizzy_face: ). Login via an alias apparently logs in via the aliased account (_unclear how that's determined if the alias is 1 to many? eg `marketing@example.com`_).
  - End of section, link and mention of eventual plan to make `SPOOF_PROTECTION=1` the default.
- **Section:** _Additional Notes_:
  - `SPOOF_PROTECTION=1` with a wildcard / catch-all alias `@example.com` could receive but not send mail. `/etc/postfix/regexp` feature added to resolve it.
    - Potentially a fix for the [recent user report with same problem](https://github.com/docker-mailserver/docker-mailserver/issues/3093) (_user need to report back if it worked, and docs need to cover the use-case well_).
  - A mention about support for LDAP users, linking several issues for reference.
  - Documented format and gotchas of the `smtpd_sender_login_maps` file, including regex support.
  - A mention of the `SPOOF_PROTECTION=1` ENV docs warn about extension delimiters no longer being usable (**breaking**). Changelog should highlight this in particular.
- **Section:** _`SPOOF_PROTECTION=1` only restricts the envelope `FROM` sender address, spoofing still feasible?_:
  - ENV name is not the most appropriate for what it does. A breaking change might as well adopt a better suited name.
  - `FROM` (mail envelope) protected from spoofing, not `from:` (mail message) header. The latter protection is handled separately from this feature.
  - [`magcks/milterfrom`](https://github.com/magcks/milterfrom) provides a Postfix milter that checks both `FROM` and `from:` are the same.
- **Section:** _Related_:
  - Potential concern for send-only addresses such as `noreply@external-domain.com` which have no mail storage and shouldn't be managed by Postfix, but with `SPOOF_PROTECTION=1` was causing issues.
  - A **concern that if `SPOOF_PROTECTION=1` is enabled by default**, that we should first support an alternative table, with improvements to docs.
  - Another scenario related to LDAP with a `noreply` account being affected.
  - A reference to docs about a workaround documented for `SPOOF_PROTECTION=1` with using intermediate aliases.

---

Beyond that, there is the _**Task Overview**_ section prior to all that, which summarizes what should ideally be tackled before defaulting to `SPOOF_PROTECTION=1`.

Given the concerns from past me, and the large amount of changes we're introducing into `v12` release, I'm inclined to say that I can't hand-wave away the `SPOOF_PROTECTION=1` change concerns. Neither of us have time to tackle those, and it would complicate any bug reports that arise from all other changes.

**Review by polarathene — APPROVED** [review]

**Review by polarathene — CHANGES_REQUESTED** [review]
Oops, just noticed the failed test suite :sweat_smile:

File: target/postfix/main.cf [hunk 1]
```diff
@@ -53,6 +53,8 @@ smtpd_client_restrictions = permit_mynetworks, permit_sasl_authenticated, reject
 smtpd_sender_restrictions = permit_sasl_authenticated, permit_mynetworks, reject_unknown_sender_domain, reject_unknown_client_hostname
 disable_vrfy_command = yes
 
+mua_sender_restrictions = reject_authenticated_sender_login_mismatch, $smtpd_sender_restrictions
```
polarathene, line 56: Wild guess that it Postfix doesn't like `smtpd_sender_restrictions` referencing itself :man_shrugging: 

```suggestion
dms_smtpd_sender_restrictions = permit_sasl_authenticated, permit_mynetworks, reject_unknown_sender_domain, reject_unknown_client_hostname
mua_sender_restrictions = reject_authenticated_sender_login_mismatch, $dms_smtpd_sender_restrictions
```

target/scripts/startup/setup.d/security/spoofing.sh
```diff
@@ -28,5 +24,7 @@ function _setup_spoof_protection
     fi
   else
     _log 'debug' 'Spoof protection is disabled'
+    # shellcheck disable=SC2016
+    postconf 'mua_sender_restrictions = $smtpd_sender_restrictions'
```
polarathene, line 28:
```suggestion
    postconf 'mua_sender_restrictions = $dms_smtpd_sender_restrictions'
```

polarathene (commented):
Hope you don't mind me committing this directly to test in CI as my local system isn't in good shape to try locally.

georglauterbach:
> Hope you don't mind me committing this directly to test in CI as my local system isn't in good shape to try locally.

Absolutely not - good thing you did :)

> **EDIT:** Ok that appears to resolve it. Maybe that is [what you were asking about earlier](https://github.com/docker-mailserver/docker-mailserver/issues/3117#issuecomment-1445352369)? 😅

👍🏼

polarathene (commented):
I added some comments if helpful, but the `dms_` prefix kinda makes it clear that it's not an upstream parameter setting.

To avoid duplication, `smtpd_sender_restrictions` will reference the same value and this avoids Postfix getting confused (_I would have thought it'd just expand what was in `main.cf` and not complained  :roll_eyes: sorry about that!_)

polarathene (approved):

target/postfix/main.cf
```diff
@@ -50,9 +50,14 @@ smtpd_helo_restrictions = permit_mynetworks, reject_invalid_helo_hostname, permi
 smtpd_relay_restrictions = permit_mynetworks permit_sasl_authenticated defer_unauth_destination
 smtpd_recipient_restrictions = permit_sasl_authenticated, permit_mynetworks, reject_unauth_destination, check_policy_service unix:private/policyd-spf, reject_unauth_pipelining, reject_invalid_helo_hostname, reject_non_fqdn_helo_hostname, reject_unknown_recipient_domain
 smtpd_client_restrictions = permit_mynetworks, permit_sasl_authenticated, reject_unauth_destination, reject_unauth_pipelining
-smtpd_sender_restrictions = permit_sasl_authenticated, permit_mynetworks, reject_unknown_sender_domain, reject_unknown_client_hostname
+smtpd_sender_restrictions = $dms_smtpd_sender_restrictions
```
casperklein, line 53: At this point, `$dms_smtpd_sender_restrictions` is unset?

Issue #3119, opened 2023-02-26
[TODO] Improve supervisor test coverage


polarathene:
## Description

As described in https://github.com/docker-mailserver/docker-mailserver/pull/3118#issuecomment-1445234066 the [supervisor tests](https://github.com/docker-mailserver/docker-mailserver/blob/15926986371a7f11de57805d24eaf0590b4fcbbc/test/tests/parallel/set3/container_configuration/process_check_restart.bats#L130) (_at the linked line location_) could be improved to better catch changes that aren't compatible with management via `supervisord`.

Should mostly involve looping through a list of service names associated to equivalent processes checked in the linked test. Those can be found in [this supervisor config](https://github.com/docker-mailserver/docker-mailserver/blob/master/target/supervisor/conf.d/supervisor-app.conf) (_there is also SASLAUTHD in a separate config](https://github.com/docker-mailserver/docker-mailserver/blob/master/target/supervisor/conf.d/saslauth.conf), but I'm not too fussed about that as we effectively only use the one for LDAP and another test for `rimap`, it's unclear how many users use SASLAUTHD for anything beyond LDAP support however_).

Up for grabs for anyone who'd like to contribute, no rush :)

PR #3128, opened 2023-02-28
scripts: remove PostSRSD wrapper

georglauterbach:
# Description

The setup is now completely done during _actual_ setup phase. The wrapper did not even catch signals (SIGINT, etc.), which I think is strange.

I also added all the ENVs the wrapper relied on (which previously could have been unset) to the variables script.

## Type of change

## Checklist:


polarathene (approved):
Thanks for tackling this! No more wrappers :tada:

target/supervisor/conf.d/supervisor-app.conf
```diff
@@ -148,7 +148,7 @@ autostart=false
 autorestart=true
 stdout_logfile=/var/log/supervisor/%(program_name)s.log
 stderr_logfile=/var/log/supervisor/%(program_name)s.log
-command=/usr/local/bin/postsrsd-wrapper.sh
+command=/etc/init.d/postsrsd start
```
casperklein, line 151: I am pretty sure this doesn't work, as the commands exits and does not run in foreground.

casperklein:
I just noticed, that the old wrapper also just called `/etc/init.d/postsrsd start`.

So this must be broken for a long time.. 😆

target/supervisor/conf.d/supervisor-app.conf
polarathene: > we can see an endless loop of supervisor starting postsrsd:

Could we better detect this issue in tests?

I touched on similar [concerns observed in the past with supervisor](https://github.com/docker-mailserver/docker-mailserver/pull/3134#issuecomment-1449598045). In addition to the other improvement [we have an existing issue for](https://github.com/docker-mailserver/docker-mailserver/issues/3119), I'll add mention of this problem too.

Good catch :)

polarathene:
> So this must be broken for a long time..

You might want to check with a release prior to `v12` / `edge` as [I did change supervisor config for PostSRSd](https://github.com/docker-mailserver/docker-mailserver/pull/3010) (_see the PR description for explanation_). Perhaps you've identified why it was configured differently and we can document that if reverting the change for `postsrsd` resolves the issue :sweat_smile: (_this would cause the test for the process restart to fail however_)

---

> Afaik `postsrsd` has no option to run in foreground.

I tried looking into that and that seems to be the case. We could probably open a feature request. The dev is fairly responsive :)

They've also released v2, not sure if that'll be part of the upcoming Debian 12 release later this year.

> So we need some kind of wrapper - I can provide one.

This is an issue specific to `supervisord`? I assume it's a non-issue with upstream official support for using systemd. Seems odd that there isn't support in `supervisord` to provide the functionality if it's fairly generic :thinking:

polarathene:
It looks like we can run the `postsrsd` command directly?

Although with `supervisord` I don't know how well that'd work as it expects ENV defined in config when initializing, not when running `supervisorctl start`, thus any changes from `start-mailserver.sh` onwards (_such as in `variables.sh`_) aren't applicable (_same for Postgrey mind you_).

Worth noting, v2 of PostSRSd uses a real config file but would presently require building from source AFAIK.

---

## Notes

From [v2 release of PostSRSd](https://github.com/roehling/postsrsd/releases/tag/2.0.0):
- They have removed CLI config options, and adopted an actual config file (_instead of sourcing ENV_).
- There is no longer a `/etc/init.d/postsrd` script supplied (_sysv service script?_), [only a service file for systemd](https://github.com/roehling/postsrsd/tree/main/data). I think [this explains why Upstart support has been dropped](https://github.com/roehling/postsrsd/issues/110#issuecomment-735474853), but systemd isn't an option for us in Docker :sweat_smile: 

---

For reference (_with some additional insights in each collapsed section_):

<details>
<summary>/etc/init.d/postsrsd (upstream init script)</summary>

[Same content as source](https://github.com/roehling/postsrsd/blob/f21c009c56d25c94bce998b26be3fa50e3c08470/init/postsrsd.sysv-lsb.in) (_after applying interpolation during install_) (_last updated Dec 2020 in `1.x` branch_):

```bash
#! /bin/sh
#
# postsrsd
#               start/stop the postsrsd daemon for Postfix
#
### BEGIN INIT INFO
# Provides:          postsrsd
# Required-Start:    $syslog $network $remote_fs
# Required-Stop:     $syslog $network $remote_fs
# Default-Start:     2 3 4 5
# Default-Stop:      0 1 6
# Short-Description: Start/stop the postsrsd daemon
# Description:       postsrsd offers Sender Rewriting Scheme support for Postfix
### END INIT INFO

set -e

PATH=/sbin:/bin:/usr/sbin:/usr/bin
DAEMON=/usr/sbin/postsrsd
NAME=postsrsd
DESC="Postfix Sender Rewriting Scheme daemon"

PIDFILE=/var/run/$NAME.pid
SCRIPTNAME=/etc/init.d/$NAME

# Gracefully exit if the package has been removed.
test -x $DAEMON || exit 0

. /lib/lsb/init-functions

# Default configuration
SRS_FORWARD_PORT=10001
SRS_REVERSE_PORT=10002
SRS_DOMAIN=`postconf -h mydomain || true`
SRS_SECRET=/etc/postsrsd.secret
SRS_SEPARATOR==
SRS_HASHLENGTH=4
SRS_HASHMIN=4
SRS_EXTRA_OPTIONS=
RUN_AS=postsrsd
CHROOT=/var/lib/postsrsd
SRS_LISTEN_ADDR=127.0.0.1
SRS_EXCLUDE_DOMAINS=

# Read config file if it is present
if [ -r /etc/default/$NAME ]
then
        . /etc/default/$NAME
fi

test -r "$SRS_SECRET" -a -n "$SRS_DOMAIN" || exit 1

ret=0

case "$1" in
  start)
        log_daemon_msg "Starting $DESC" "$NAME"
        if start-stop-daemon --start --oknodo --quiet \
                --pidfile $PIDFILE \
                --name $NAME \
                --startas $DAEMON \
                -- $SRS_EXTRA_OPTIONS -f "$SRS_FORWARD_PORT" -r "$SRS_REVERSE_PORT" \
                   -d "$SRS_DOMAIN" -s "$SRS_SECRET" -a "$SRS_SEPARATOR" \
                   -n "$SRS_HASHLENGTH" -N "$SRS_HASHMIN" -u "$RUN_AS" -p "$PIDFILE" \
                   -c "$CHROOT" -l "$SRS_LISTEN_ADDR" -D -X"$SRS_EXCLUDE_DOMAINS"
        then
            log_end_msg 0
        else
            ret=$?
            log_end_msg 1
        fi
        ;;
  stop)
        log_daemon_msg "Stopping $DESC" "$NAME"
        if start-stop-daemon --stop --oknodo --quiet \
                --pidfile $PIDFILE --exec $DAEMON --name $NAME
        then
            log_end_msg 0
        else
            ret=$?
            log_end_msg 1
        fi
        rm -f $PIDFILE
        ;;
  reload|force-reload|restart)
        $0 stop
        $0 start
        ret=$?
        ;;
  status)
        log_daemon_msg "postsrsd is running"
        if [ -s $PIDFILE ]; then
            PID=`cat $PIDFILE`
            if kill -0 "$PID" 2>/dev/null; then
                log_end_msg 0
            else
                log_end_msg 1
            fi
        else
            log_end_msg 1
        fi
        ;;
  *)
        echo "Usage: $SCRIPTNAME {start|stop|restart|reload|force-reload|status}" >&2
        exit 1
        ;;
esac

exit $ret
```

</details>

<details>
<summary>`postsrsd -h` (no longer valid in 2.x)</summary>

```
Sender Rewriting Scheme implementation for Postfix.

Implements two TCP lookup tables to rewrite mail addresses
as needed. The forward SRS is for sender envelope addresses
to prevent SPF-related bounces. The reverse SRS is for
recipient envelope addresses so that bounced mails can be
routed back to their original sender.

Usage: postsrsd -s<file> -d<domain> [other options]
Options:
   -s<file>       read secrets from file (required)
   -d<domain>     set domain name for rewrite (required)
   -a<char>       set first separator character which can be one of: -=+ (default: =)
   -n<num>        length of hash to be used in rewritten addresses (default: 4)
   -N<num>        minimum length of hash to accept for validation (default: 4)
   -l<addr>       set socket listen address (default: 127.0.0.1)
   -f<port>       set port for the forward SRS lookup (default: 10001)
   -r<port>       set port for the reverse SRS lookup (default: 10002)
   -p<pidfile>    write process ID to pidfile (default: none)
   -c<dir>        chroot to <dir> (default: none)
   -u<user>       switch user id after port bind (default: none)
   -t<seconds>    timeout for idle client connections (default: 1800)
   -X<domain>     exclude additional domain from address rewriting
   -A             always rewrite addresses
   -e             attempt to read above parameters from environment
   -D             fork into background
   -4             force IPv4 socket (default: any)
   -6             force IPv6 socket (default: any)
   -h             show this help
   -v             show version
```

</details>

<details>
<summary>/etc/default/postsrsd (out of sync with upstream)</summary>

We have been [supplying our own copy of the PostSRSd config](https://github.com/docker-mailserver/docker-mailserver/blob/9e2f96448ad4f5afd897621c54a731bbb18e12bb/target/postsrsd/postsrsd):

https://github.com/docker-mailserver/docker-mailserver/blob/9e2f96448ad4f5afd897621c54a731bbb18e12bb/Dockerfile#L104

Out of sync with [upstream config](https://github.com/roehling/postsrsd/blob/f21c009c56d25c94bce998b26be3fa50e3c08470/init/postsrsd.default.in). Debian `postsrsd` package already provides the `/etc/default/postsrsd` config with:
- `RUN_AS=postsrsd` instead of `nobody`
- `CHROOT=/var/lib/postsrsd`
- `SRS_SECRET=/etc/postsrsd.secret`
- `SRS_DOMAIN=localdomain` (_sourced from `postconf -h myhostname` with Postfix installed already, as config comments below document_)

Thus we're not making any changes relevant to DMS during the build and could just use the updated config from the package.

Despite being out of sync here, the `/etc/init.d/postsrsd` is not, and declares the same ENV default there. ~~I'm not 100% sure if supervisor inherits any ENV changes from `variables.sh`, or if the config modified by our PostSRSd scripts (_`setup.d/postfix.sh` in this PR_) is used.~~ (_**EDIT:** the `/etc/init.d/postsrsd` script sources `/etc/default/postsrsd` as ENV that would override the previous declarations in the init script, so all good_)

</details>

polarathene:
> It looks like we can run the `postsrsd` command directly?

Yes

## PostSRSd

So to wrap this up, the init script we're using presently calls `postsrsd` with `-D`, which if you look at the help output [daemonizes it](https://github.com/roehling/postsrsd/blob/bca3ec80698341a61ee0d0145c8d7be586a18bb8/src/main.c#L141-L154). 

But also there is an `-e` option that will attempt to populate all the options via ENV. At first I sourced `/etc/default/postsrsd` and tried calling `postsrsd -e`, that didn't work, probably needed to `export` each ENV in the file. 

We can [grab them all by filtering out comments and blank lines](https://gist.github.com/mihow/9c7f559807069a03e302605691f85572?permalink_comment_id=3944636#gistcomment-3944636), then call with `env` command, this works and `postsrsd` will start [without complaining](https://github.com/roehling/postsrsd/blob/f21c009c56d25c94bce998b26be3fa50e3c08470/postsrsd.c#L483-L487) about [not finding `SRS_DOMAIN`](https://github.com/roehling/postsrsd/blob/f21c009c56d25c94bce998b26be3fa50e3c08470/postsrsd.c#L427-L429) (`-d` value):

```bash
env $(grep -vE "^(#.*|\s*)$" /etc/default/postsrsd) postsrsd -e
```

As mentioned before, we don't need our own copy of `/etc/default/postsrsd`. That should resolve the problem.

- We probably don't need the chroot dir. Default is `-c none`.
- [Syslog appears to be implicitly used internally](https://github.com/roehling/postsrsd/blob/1.x/postsrsd.c).

---

## Postgrey

Similar could be done for Postgrey. `env` doesn't work I think here for `/etc/dms-settings` due to single quote wrapping values for shell use apparently, so a shell script to source the ENV (_after `variables.sh` processing_) then provides the vars to interpolate when calling the `postgrey` command:

```bash
#! /bin/bash
source /etc/dms-settings

postgrey \
  --inet=127.0.0.1:10023 \
  --syslog-facility=mail \
  --delay="${POSTGREY_DELAY}" \
  --max-age="${POSTGREY_MAX_AGE}" \
  --auto-whitelist-clients="${POSTGREY_AUTO_WHITELIST_CLIENTS}" \
  --greylist-text="${POSTGREY_TEXT}"
```

Then we don't need the extra ENV in `Dockerfile` since supervisor will start with a config that doesn't check for them. Postgrey does not support any config sourcing like PostSRSd.

I don't think that requires a wrapper, when the command stops (_eg if it `postgrey` process was killed, or if you ran this script and `CTRL+C`_), that exits, and supervisor should still restart that fine, along with start / stop commands.

- It's not running as a daemon (_supported via `-d` / `--daemonize`_).
- If needed we can add a PID file with `--pidfile` and use supervisor `pidproxy` to proxy the signals.
- `--syslog-facility` seems redundant (default), but perhaps worthwhile for visibility :sweat_smile: 
- There is also a `--hostname` option that looks like it defaults to calling `hostname` command, that we should probably be configuring explicitly with our setting to accommodate supporting `OVERRIDE_HOSTNAME`.
- Our Postgrey version is [1.36 from 2015](https://github.com/schweikert/postgrey/blob/c589bb9b727be9e1caaef6aa809a5ced688d9509/Changes#L4-L14), for some reason we didn't get the 1.37 released since 2016, and while they've updated whitelists in early 2022, no new release has been made. `postsrsd` also offers this with `-p<pidfile>`.

<details>
<summary>postgrey -h</summary>

```
Usage:
    postgrey [options...]

     -h, --help              display this help and exit
         --version           output version information and exit
     -v, --verbose           increase verbosity level
         --syslog-facility   Syslog facility to use (default mail)
     -q, --quiet             decrease verbosity level
     -u, --unix=PATH         listen on unix socket PATH
         --socketmode=MODE   unix socket permission (default 0666)
     -i, --inet=[HOST:]PORT  listen on PORT, localhost if HOST is not specified
     -d, --daemonize         run in the background
         --pidfile=PATH      put daemon pid into this file
         --user=USER         run as USER (default: postgrey)
         --group=GROUP       run as group GROUP (default: postgrey)
         --dbdir=PATH        put db files in PATH (default: /var/lib/postgrey)
         --delay=N           greylist for N seconds (default: 300)
         --max-age=N         delete entries older than N days since the last time
                             that they have been seen (default: 35)
         --retry-window=N    allow only N days for the first retrial (default: 2)
                             append 'h' if you want to specify it in hours
         --greylist-action=A if greylisted, return A to Postfix (default: DEFER_IF_PERMIT)
         --greylist-text=TXT response when a mail is greylisted
                             (default: Greylisted + help url, see below)
         --lookup-by-subnet  strip the last N bits from IP addresses, determined by ipv4cidr and ipv6cidr (default)
         --ipv4cidr=N        What cidr to use for the subnet on IPv4 addresses when using lookup-by-subnet (default: 24)
         --ipv6cidr=N        What cidr to use for the subnet on IPv6 addresses when using lookup-by-subnet (default: 64)
         --lookup-by-host    do not strip the last 8 bits from IP addresses
         --privacy           store data using one-way hash functions
         --hostname=NAME     set the hostname (default: `hostname`)
         --exim              don't reuse a socket for more than one query (exim compatible)
         --whitelist-clients=FILE     default: /etc/postgrey/whitelist_clients
         --whitelist-recipients=FILE  default: /etc/postgrey/whitelist_recipients
         --auto-whitelist-clients=N   whitelist host after first successful delivery
                                      N is the minimal count of mails before a client is 
                                      whitelisted (turned on by default with value 5)
                                      specify N=0 to disable.
         --listen-queue-size=N        allow for N waiting connections to our socket
         --x-greylist-header=TXT      header when a mail was delayed by greylisting
                                      default: X-Greylist: delayed <seconds> seconds by postgrey-<version> at <server>; <date>

     Note that the --whitelist-x options can be specified multiple times,
     and that per default /etc/postgrey/whitelist_clients.local and
     /etc/postgrey/whitelist_recipients.local are also read, so that you can put
     there local entries.
```

</details>

georglauterbach:
Nice analysis! I'd probably wait though, this does not have a high priority, right? Releasing v12.0.0 and then relaxing the whole situation is better I think.

polarathene:
> this does not have a high priority, right?

The whole point of looking into the above was because of a [bug from this PR](https://github.com/docker-mailserver/docker-mailserver/pull/3128/files#r1123885007) that @casperklein noticed. Presently `postsrsd` does not work, supervisor thinks it exited and constantly tries restarting it.

I am not able to try verifying the failure locally via a new build atm, but running `/etc/init.d/postsrsd start` directly in a build before this PR seemed to work fine :man_shrugging:

georglauterbach:
Oh, didn't notice! Yeah, this should be resolved before we publish v12.0.0. 🤣

polarathene:
> solves the issue?

I believe so, but like I said I can't verify that right now. I should be able to within 24-48 hours.

If you can reproduce the same failure that Casper demonstrated earlier, then apply these changes and the failure no longer reproduces, that'd be a fix :)

If I could rely on CI tests for this, I'd happily create a PR, but we don't have a test that catches this (_and it's been an issue I've seen with other services during the v12 cycle_).

casperklein:
> So running `/bin/bash -c 'exec env $(grep -vE "^(#.*|\s*)$" /etc/default/postsrsd) postsrsd -e -c none'` instead of `/etc/init.d/postsrsd start` and using the default `/etc/default/postsrsd` solves the issue?

I did a quick test and it does not work for me.

polarathene:
> I did a quick test and it does not work for me.

I am able to build and run locally again :sweat_smile: 

I tried to reproduce the failure you're experiencing, but I'm not seeing it? Something is different between our systems?

```console
$ docker run --rm -it --hostname example.test --env SMTP_ONLY=1 --env SUPERVISOR_LOGLEVEL=info --env ENABLE_SRS=1 mailserver-testing:ci

Restarted supervisord
2023-03-05 04:59:57,227 INFO Included extra file "/etc/supervisor/conf.d/saslauth.conf" during parsing
2023-03-05 04:59:57,227 INFO Included extra file "/etc/supervisor/conf.d/supervisor-app.conf" during parsing
2023-03-05 04:59:57,227 INFO Set uid to user 0 succeeded
2023-03-05 04:59:57,228 INFO RPC interface 'supervisor' initialized
2023-03-05 04:59:57,228 INFO supervisord started with pid 7
2023-03-05 04:59:58,230 INFO spawned: 'mailserver' with pid 30
[   INF   ]  Welcome to docker-mailserver 11.3.1
2023-03-05 04:59:58,266 INFO success: mailserver entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
[ WARNING ]  !! INSECURE !! SSL configured with plain text access - DO NOT USE FOR PRODUCTION DEPLOYMENT
[   INF   ]  Starting daemons
2023-03-05 04:59:58,968 INFO spawned: 'postsrsd' with pid 310
2023-03-05 04:59:58,969 INFO success: postsrsd entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
2023-03-05 04:59:59,114 INFO spawned: 'cron' with pid 317
2023-03-05 04:59:59,115 INFO success: cron entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
2023-03-05 04:59:59,266 INFO spawned: 'rsyslog' with pid 321
2023-03-05 04:59:59,267 INFO success: rsyslog entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
2023-03-05 04:59:59,414 INFO spawned: 'update-check' with pid 327
2023-03-05 04:59:59,415 INFO success: update-check entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
2023-03-05 04:59:59,713 INFO spawned: 'opendkim' with pid 355
2023-03-05 04:59:59,714 INFO success: opendkim entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
2023-03-05 04:59:59,866 INFO spawned: 'opendmarc' with pid 371
2023-03-05 04:59:59,867 INFO success: opendmarc entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
2023-03-05 05:00:00,016 INFO spawned: 'postfix' with pid 383
2023-03-05 05:00:00,016 INFO success: postfix entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
2023-03-05 05:00:00,173 INFO spawned: 'amavis' with pid 419
2023-03-05 05:00:00,174 INFO success: amavis entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
2023-03-05 05:00:00,334 INFO spawned: 'changedetector' with pid 480
2023-03-05 05:00:00,334 INFO success: changedetector entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
[   INF   ]  example.test is up and running
Mar  5 05:00:00 example amavis[419]: starting. /usr/sbin/amavisd-new at example.test amavisd-new-2.11.1 (20181009), Unicode aware, LC_CTYPE="C.UTF-8"
Mar  5 05:00:00 example amavis[419]: perl=5.032001, user=, EUID: 109 (109);  group=, EGID: 111 111 (111 111)
Mar  5 05:00:00 example amavis[419]: Net::Server: Group Not Defined.  Defaulting to EGID '111 111'
Mar  5 05:00:00 example amavis[419]: Net::Server: User Not Defined.  Defaulting to EUID '109'
Mar  5 05:00:00 example amavis[419]: No ext program for   .zoo, tried: zoo
Mar  5 05:00:00 example amavis[419]: No ext program for   .doc, tried: ripole
Mar  5 05:00:00 example amavis[419]: No decoder for       .F   
Mar  5 05:00:00 example amavis[419]: No decoder for       .doc 
Mar  5 05:00:00 example amavis[419]: No decoder for       .zoo
```

That is with no changes discussed here. Just git clone and `make build` with the `docker run` command shown above.

casperklein:
Very strange. On debian 11:

<details>
<summary>Details</summary>

```
git clone https://github.com/docker-mailserver/docker-mailserver && \
cd docker-mailserver && \
docker build -t test . && \
docker run --rm -it --hostname example.test --env SMTP_ONLY=1 --env SUPERVISOR_LOGLEVEL=info --env ENABLE_SRS=1 test
```

```
Restarted supervisord
2023-03-05 12:36:20,769 INFO Included extra file "/etc/supervisor/conf.d/saslauth.conf" during parsing
2023-03-05 12:36:20,769 INFO Included extra file "/etc/supervisor/conf.d/supervisor-app.conf" during parsing
2023-03-05 12:36:20,769 INFO Set uid to user 0 succeeded
2023-03-05 12:36:20,769 INFO RPC interface 'supervisor' initialized
2023-03-05 12:36:20,770 INFO supervisord started with pid 7
2023-03-05 12:36:21,772 INFO spawned: 'mailserver' with pid 30
[   INF   ]  Welcome to docker-mailserver 11.3.1
2023-03-05 12:36:21,793 INFO success: mailserver entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
[ WARNING ]  !! INSECURE !! SSL configured with plain text access - DO NOT USE FOR PRODUCTION DEPLOYMENT
[   INF   ]  Starting daemons
2023-03-05 12:36:22,069 INFO spawned: 'postsrsd' with pid 312
2023-03-05 12:36:22,070 INFO success: postsrsd entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
2023-03-05 12:36:22,163 INFO spawned: 'cron' with pid 319
2023-03-05 12:36:22,163 INFO success: cron entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
2023-03-05 12:36:22,259 INFO spawned: 'rsyslog' with pid 323
2023-03-05 12:36:22,260 INFO success: rsyslog entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
2023-03-05 12:36:22,353 INFO spawned: 'update-check' with pid 329
2023-03-05 12:36:22,354 INFO success: update-check entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
2023-03-05 12:36:22,539 INFO spawned: 'opendkim' with pid 363
2023-03-05 12:36:22,539 INFO success: opendkim entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
2023-03-05 12:36:22,633 INFO spawned: 'opendmarc' with pid 373
2023-03-05 12:36:22,633 INFO success: opendmarc entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
2023-03-05 12:36:22,735 INFO spawned: 'postfix' with pid 386
2023-03-05 12:36:22,735 INFO success: postfix entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
2023-03-05 12:36:22,763 INFO exited: postsrsd (exit status 0; expected)
2023-03-05 12:36:22,764 INFO spawned: 'postsrsd' with pid 394
2023-03-05 12:36:22,771 INFO success: postsrsd entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
2023-03-05 12:36:22,772 INFO exited: postsrsd (exit status 0; expected)
2023-03-05 12:36:22,772 INFO spawned: 'postsrsd' with pid 403
2023-03-05 12:36:22,781 INFO success: postsrsd entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
2023-03-05 12:36:22,781 INFO exited: postsrsd (exit status 0; expected)
2023-03-05 12:36:22,782 INFO spawned: 'postsrsd' with pid 413
2023-03-05 12:36:22,789 INFO success: postsrsd entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
2023-03-05 12:36:22,790 INFO exited: postsrsd (exit status 0; expected)

```

</details>

georglauterbach:
I am able to replicate @casperklein's behaviour on Ubuntu 22.04!

georglauterbach:
Could we just set `autorestart=false` for now? Not good, not terrible IMO.

casperklein:
> Could we just set `autorestart=false` for now? Not good, not terrible IMO.

This would fix the infinite loop, but would not bring the functionality to proper stop/restart this service.
Better to use a little wrapper again, which handles signals.

polarathene:
> Something is different between our systems?

As mentioned in #3160 I realized that this was due to my container running with `RLIMIT_NOFILE` / `ulimit` over 1 billion (`infinity`) which caused it to run for 8 minutes before failing as it iterates through that range of FDs calling `close()` on each one.

Running with a sane limit via `--ulimit "nofile=1024" reproduces the failure.

---

Adding this as the `postsrsd` service command in `/etc/supervisor/conf.d/supervisor-app.conf` works without a wrapper:

```
command=/bin/bash -c 'env $(grep -vE "^(#.*|\s*)$" /etc/default/postsrsd) postsrsd -e -p /var/run/postsrsd.pid'
```

This is with command:

```bash
docker run --rm -it --hostname example.test --env SMTP_ONLY=1 --env SUPERVISOR_LOGLEVEL=info --env ENABLE_SRS=1 --ulimit "nofile=$(ulimit -Sn):$(ulimit -Hn)" mailserver-testing:ci
```

It doesn't exit since it's running in the foreground (_unlike the `/etc/init.d/postsrsd start` that runs as daemon via `-D`_). `supervisorctl start` / `stop` / `restart` all working as expected and logs being output to `/var/log/supervisorctl/postsrsd.conf`.

PR #3125, opened 2023-02-27
config: ensure SASL socket file is not inside a volume mount

DavyLandman:
See #3110 for more details

# Description

The sasl socket was on a path that would end up getting volume mounted (`/var/mail-state`) in some setups, this could cause issues. The socket is not actually state, so it doesn't need to end up there. 

Fixes #3110

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [ ] If necessary I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

*note, I've marked it as Draft because:*

- [ ] not sure hot to test this locally
- [x] discussion is still open if we maybe should move to `/dev/shm` instead.

Issue #3116, opened 2023-02-25
[BUG] Restarting postfix fails


casperklein:
### Miscellaneous first checks

- [X] I checked that all ports are open and not blocked by my ISP / hosting provider.
- [X] I know that SSL errors are likely the result of a wrong setup on the user side and not caused by DMS itself. I'm confident my setup is correct.

### Affected Component(s)

`supervisorctl restart postfix` fails

### What happened and when does this occur?

Running `supervisorctl restart postfix` inside the container, leads to the following endless looping error messages in `mail.log`:

```
2023-02-25 22:51:45,196 INFO waiting for postfix to stop
2023-02-25 22:51:45,197 INFO stopped: postfix (terminated by SIGTERM)
2023-02-25 22:51:45,199 INFO spawned: 'postfix' with pid 4682
2023-02-25 22:51:45,200 INFO success: postfix entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
Feb 25 22:51:45 mail postfix/postfix-script[4688]: fatal: the Postfix mail system is already running
2023-02-25 22:51:46,223 INFO exited: postfix (exit status 1; not expected)
2023-02-25 22:51:47,228 INFO spawned: 'postfix' with pid 4689
Feb 25 22:51:47 mail postfix/postfix-script[4695]: fatal: the Postfix mail system is already running
2023-02-25 22:51:47,256 INFO success: postfix entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
2023-02-25 22:51:48,257 INFO exited: postfix (exit status 1; not expected)
2023-02-25 22:51:49,262 INFO spawned: 'postfix' with pid 4696
Feb 25 22:51:49 mail postfix/postfix-script[4702]: fatal: the Postfix mail system is already running
2023-02-25 22:51:49,287 INFO success: postfix entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
2023-02-25 22:51:50,288 INFO exited: postfix (exit status 1; not expected)
```

### What did you expect to happen?

Clean restart of postfix.

### How do we replicate the issue?

1. Run `supervisorctl restart postfix` inside the container
2.
3.
...

### DMS version

edge

### What operating system is DMS running on?

Linux

### Which operating system version?

Debian 11

### What instruction set architecture is DMS running on?

AMD64 / x86_64

### What container orchestration tool are you using?

Docker

### Other relevant information

[...]

PR #3033, opened 2023-01-27
chore: Remove the wrapper script for Postfix (and disable chroot in `master.cf`)

polarathene:
# Description

- Motivation is to drop the wrapper script mainly due to the `sleep` usage.
- Running Postfix in foreground for `supervisord` seems to be possible now, provided we drop the chroot jail config in `master.cf`.
- Added benefit of not being concerned with chroot copy of system files (notably DNS config changes) falling out of sync (_requires restarting Postfix process_). A concern that may be more likely problematic as our `changedetector` service in the next release switches to `postfix reload` instead of having the process killed to restart it fully.

This change seems to be fine from what I can assess :)

[Mailu `master.cf`](https://github.com/Mailu/Mailu/blob/179c62411603c50d4c554bc00bb52f62dd63611d/core/postfix/conf/master.cf) and [mailcow `master.cf`](https://github.com/mailcow/mailcow-dockerized/blob/640f535e9995f27fcc1541a60f0a56648d0d66cc/data/conf/postfix/master.cf) both are configured to not use chroot.

### DMS history for current config

- Postfix can run in foreground mode since 3.3 (Feb 2018).
- We've been using the `master.cf` config for Postfix in this image has been using the [chroot N default since April 2015 from an Ubuntu base image](https://github.com/docker-mailserver/docker-mailserver/commit/ef34462e34f2d6a46f8da9dc4c068c2186b8ea54). In [May 2017, this was changed to chroot Y](https://github.com/docker-mailserver/docker-mailserver/pull/612) for most services, but [an explicit `n` for Amavis which seems to be what motivated the change](https://github.com/docker-mailserver/docker-mailserver/issues/545#issuecomment-301740262).
- The [Postfix wrapper script was introduced in a PR from Feb 2018](https://github.com/docker-mailserver/docker-mailserver/pull/676) (_too new for the project to have as an updated package at the time_) for running via `supervisord`.

A [discussion between maintainers in June 2022](https://github.com/docker-mailserver/docker-mailserver/pull/2624#issuecomment-1155813516) referenced [the following](https://serverfault.com/questions/885828/is-there-any-way-to-run-postfix-in-foreground/916072#916072):

> Please note that if Postfix is launched in [chroot](https://en.wikipedia.org/wiki/Chroot) environment (_which is default in many distros_) **with** `postfix start-fg`, you will have the `/var/spool/postfix/etc` dir empty, postfix will not know your `localtime`, `hosts`, `resolv.conf`, **you will not be able to deliver a single message**.
[...]

Issue #329, opened 2016-09-21
Server is logging to both /var/log/mail/mail.* AND /var/log/syslog


tomav:
Am I right?
Perhaps we should only keep `/var/log/mail.*` to avoid log duplication / optimize disk space.

Any thoughts?

Josef-Friedrich:
You are right. The same issue in my container.

In `/etc/rsyslog.d/50-default.conf` `*.*;auth,authpriv.none		-/var/log/syslog` should be replaced with  `*.*;mail,auth,authpriv.none     -/var/log/syslog` (untested)

PR #3131, opened 2023-02-28
config: ensure SASL socket file is not inside a volume mount

### LINKED ISSUE #3110: [BUG] cannot send email, private/auth failed: permission denied

### Affected Component(s)

smtpd / sasl

Without any updates to the container, user credentials, or vps, I cannot send mail anymore.

I can still connect to the smtp ports (both 465 and 587), it's just that it will get rejected. IMAP receiving of email also still works.

I expect the SMTP server to accept my login and allow me to send emails.

Also, it seems to complain about SASL, but that's disabled for my server.

Uhm, I find this a bit hard to say right now, but I would like to help debug and figure out what's going on

### DMS version

v11.3.1

Linux

Alpine 3.16.4

AMD64 / x86_64

Docker

### docker-compose.yml

```yml
`  mailserver:
    image: docker.io/mailserver/docker-mailserver:11.3.1
    restart: unless-stopped
    stop_grace_period: 1m
    hostname: mail.XXX.cloud
    ports:
      - "25:25"     # SMTP  (explicit TLS => STARTTLS)
      - "143:143"   # IMAP4 (explicit TLS => STARTTLS)
      - "465:465"   # ESMTP (implicit TLS)
      - "587:587"   # ESMTP (explicit TLS => STARTTLS)
      - "993:993"   # IMAP4 (implicit TLS)
      - "4190:4190" # user seive
    volumes:
      - ./data/dms/data:/var/mail
      - ./data/dms/state:/var/mail-state
      - ./logs/dms:/var/log/mail
      # borrow traefik log
      - ./data/traefik/acme.json:/etc/letsencrypt/acme.json:ro
    mem_limit: 3.6G
    cpus: 1.9
    environment:
      - LOG_LEVEL=info
      - SUPERVISOR_LOGLEVEL=info
      - TZ=Europe/Amsterdam
      - POSTFIX_INET_PROTOCOLS=all
      - DOVECOT_INET_PROTOCOLS=all

      - ENABLE_SASLAUTHD=0

      # save all into single dir
      - ONE_DIR=1

      # security/filtering stuff
      # filter stuff
      - SPOOF_PROTECTION=1
      - ENABLE_CLAMAV=1
      - ENABLE_AMAVIS=1
      - ENABLE_POP3=0
      # show warnings from amavis filter
      - AMAVIS_LOGLEVEL=0
      # disable the dns blocklist for now
      - ENABLE_DNSBL=0
      # block attackers
      - ENABLE_FAIL2BAN=1
      - FAIL2BAN_BLOCKTYPE=drop
      # check valid connections with all tests
      - POSTSCREEN_ACTION=enforce
      # enalbe spam checker
      - ENABLE_SPAMASSASSIN=1
      - SPAMASSASSIN_SPAM_TO_INBOX=1
      - ENABLE_SPAMASSASSIN_KAM=1
      - "SA_SPAM_SUBJECT=[SPAM] "
      - ENABLE_MANAGESIEVE=1

      # we'll borrow traefik cretificates
      - SSL_TYPE=letsencrypt
      - SSL_DOMAIN=mail.XXX.cloud

      # limits
      # 50 mb
      - POSTFIX_MESSAGE_SIZE_LIMIT=52428800

    cap_add:
      - NET_ADMIN
      - SYS_PTRACE
    labels:
      - "traefik.enable=false"
```

### Relevant log output

I've looked through the logs (`docker-compose logs --follow`) and noticed this coming up when I tried to send an email (both with thunderbird and with android gmail client).

```Text
mailserver_1                       | Feb 23 19:40:38 mail postfix/submission/smtpd[1455]: warning: hostname XXXXX.connected.by.freedominter.net does not resolve to address 2a10:3781:XXX::1: Name or service not known
mailserver_1                       | Feb 23 19:40:38 mail postfix/submission/smtpd[1455]: connect from unknown[2a10:3781:XXX:1]
mailserver_1                       | Feb 23 19:40:38 mail postfix/submission/smtpd[1455]: Anonymous TLS connection established from unknown[2a10:3781:XXX::1]: TLSv1.3 with cipher TLS_AES_256_GCM_SHA384 (256/256 bits) key-exchange X25519 server-signature RSA-PSS (4096 bits) server-digest SHA256
mailserver_1                       | Feb 23 19:40:38 mail postfix/submission/smtpd[1455]: warning: SASL: Connect to private/auth failed: Permission denied
mailserver_1                       | Feb 23 19:40:38 mail postfix/submission/smtpd[1455]: fatal: no SASL authentication mechanisms
mailserver_1                       | Feb 23 19:40:39 mail postfix/master[1399]: warning: process /usr/lib/postfix/sbin/smtpd pid 1455 exit status 1
mailserver_1                       | Feb 23 19:40:39 mail postfix/master[1399]: warning: /usr/lib/postfix/sbin/smtpd: bad command startup -- throttling
```

I've also tried it over ipv4, the same happens.

_No response_

- [X] I am inexperienced with mail servers
- [ ] I am rather experienced with mail servers

### Code of conduct

_No response_

georglauterbach:
# Description

Change the SASL auth socket path. The SASL socket was on a path that would end up getting volume mounted (`/var/mail-state`) in some setups, this could cause issues. The socket is not actually state, so it doesn't need to end up there.

Fixes #3110
Superseeds #3125

## Type of change

## Checklist:


georglauterbach:
What even is this:

```txt
not ok 85 [General] system: postfix should not log to syslog in 123ms
#  in test file test/tests/serial/tests.bats, line 208)
#   `assert_failure' failed
#
# output : Feb 28 14:51:12 mail rsyslogd: cannot create '/var/spool/postfix/dev/log': No such file or directory [v8.2102.0 try https://www.rsyslog.com/e/2176 ]
# --
#
```

Can somebody please tell me what's happening here? 🤣 🤣 🤣

polarathene (dismissed):

polarathene:
> Can somebody please tell me what's happening here?    Why, if Postfix does not log to syslog, is syslog still trying to create `/var/spool/postfix/dev/log`? @casperklein any idea? I mean, the reason it fails at the moment is obvious (the directory does not exist in the first place)

This reminds me that I forgot to look into an extra change for [dropping the `chroot` Postfix config](https://github.com/docker-mailserver/docker-mailserver/pull/3033)... it was about logging :sweat_smile: 

There is some [Postfix docs about configuring _for_ `chroot`](https://www.postfix.org/BASIC_CONFIGURATION_README.html#chroot_setup):

> Additionally, you almost certainly need to configure syslogd so that it listens on a socket inside the Postfix queue directory.
> Examples of syslogd command line options that achieve this for specific systems:
> 
> Linux, OpenBSD: syslogd -a /var/spool/postfix/dev/log

So it's possibly due to `syslogd` being configured to setup logging there or something?

---

**EDIT:** And here it is:

```console
$ cat /etc/rsyslog.d/postfix.conf

# Create an additional socket in postfix's chroot in order not to break
# mail logging when rsyslog is restarted.  If the directory is missing,
# rsyslog will silently skip creating the socket.
$AddUnixListenSocket /var/spool/postfix/dev/log
```

And here is the [Postfix docs I recalled for adjusting the logs](https://www.postfix.org/MAILLOG_README.html#log-to-stdout). AFAIK, we'd want to set [`main.cf:maillog_file`](https://www.postfix.org/postconf.5.html#maillog_file) to `/dev/stdout` instead of a log file, this way supervisor service running `postfix start-fg` should be capturing the logs and writing that to a file AFAIK?

This would just require the following:

```
# /etc/postfix/master.cf
postlog   unix-dgram n  -       n       -       1       postlogd

# /etc/postfix/main.cf
maillog_file = /dev/stdout
```

Should I raise a separate PR for that, or do you want to bundle those changes into this PR?

georglauterbach:
> Should I raise a separate PR for that, or do you want to bundle those changes into this PR?

Since I can review quickly, and only 1 review requires ATM, please raise another PR :)

The removal of the syslog config file for postfix that you proposed to do in the Dockerfile, if you can also do that in the scripts, I'd do it in the scripts :)

polarathene:
> Since I can review quickly, and only 1 review is required ATM, please raise another PR :)

Done, feel free to merge after approving :+1:

casperklein:
Forgive me, I haven't read in detail the initial bug report. I was wondering if the other sockets that are also volume mounted are also problematic:

<details>
<summary>Details</summary>

```
./mailstate/lib-amavis/amavisd.sock
./mailstate/spool-postfix/public/cleanup
./mailstate/spool-postfix/public/qmgr
./mailstate/spool-postfix/public/showq
./mailstate/spool-postfix/public/sender-cleanup
./mailstate/spool-postfix/public/flush
./mailstate/spool-postfix/private/rewrite
./mailstate/spool-postfix/private/relay
./mailstate/spool-postfix/private/verify
./mailstate/spool-postfix/private/proxywrite
./mailstate/spool-postfix/private/maildrop
./mailstate/spool-postfix/private/anvil
./mailstate/spool-postfix/private/bsmtp
./mailstate/spool-postfix/private/tlsmgr
./mailstate/spool-postfix/private/error
./mailstate/spool-postfix/private/tlsproxy
./mailstate/spool-postfix/private/ifmail
./mailstate/spool-postfix/private/scalemail-backend
./mailstate/spool-postfix/private/defer
./mailstate/spool-postfix/private/proxymap
./mailstate/spool-postfix/private/retry
./mailstate/spool-postfix/private/mailman
./mailstate/spool-postfix/private/smtpd
./mailstate/spool-postfix/private/dnsblog
./mailstate/spool-postfix/private/auth
./mailstate/spool-postfix/private/virtual
./mailstate/spool-postfix/private/uucp
./mailstate/spool-postfix/private/local
./mailstate/spool-postfix/private/lmtp
./mailstate/spool-postfix/private/policyd-spf
./mailstate/spool-postfix/private/trace
./mailstate/spool-postfix/private/smtp
./mailstate/spool-postfix/private/bounce
./mailstate/spool-postfix/private/discard
./mailstate/spool-postfix/private/scache
./mailstate/spool-postfix/private/smtp-amavis
```

</details>

polarathene:
> I was wondering if the other sockets that are also volume mounted are also problematic

Are all those postfix files sockets? I wasn't sure why they seemed to be empty when I saw them. All I know is they're various programs that Postfix uses to run (_bulk are configured in `master.cf`_).

I know they're added in a fresh image without a volume mounted, but I didn't investigate if they were created as part of the package install or from Postfix service being started. I think it's the latter due to this being the `queue_directory` location where Postfix does it thing (_`chroot` jail or not_).

---

`amavisd.sock` would probably fall into the same category if the location was an issue. I'm fairly positive that it's unlikely anyone is running a host with an old linux kernel, so the original _Docker with sockets in locations using `overlayfs`_ bug should already be fixed (_since July 2016, but broader deployment adoption would take a while_).

AFAIK, something caused a users permissions for the socket to change from the configured `0666` to `0660`, which would fail due to the `docker` / `docker ` UID + GID ownership not applying to the `postfix` user trying to access it. The other failure case we know about seems related to rootless containers (several reports).

Correcting the user and group assigned to the socket should resolve the main issue. The new socket location is likely not necessary, just following the [same workaround that was used for supervisor](https://github.com/docker-mailserver/docker-mailserver/issues/3110#issuecomment-1445490507) (_Sep 2017, a time when the Docker bug was likely on systems with old enough kernels_).

casperklein:
> Are all those postfix files sockets? I wasn't sure why they seemed to be empty when I saw them.

Yes (`find -type s`).

> in some setups, this could cause issues

In detail, Podman rootless? Not going to block this PR, just curious about this, because afaik we didn't include patches for "not supported" stuff in the past?

polarathene:
Alright, I have since learned that `/dev/log` is created by `rsyslog` (_or any other log service that emits syslog format?_).

- If we replaced `rsyslog` in future, such as with Vector, it just needs to create a socket at `/dev/log` to receive logs from various processes emitting them there, and we're good. I've verified this works with Vector.
- The failing test was [implemented back in Sep 2016](https://github.com/docker-mailserver/docker-mailserver/pull/330) to [prevent duplicate content on disk](https://github.com/docker-mailserver/docker-mailserver/issues/329).
- Their solution to prevent it was to [opt-out of logs of a given facilitiy (such as `mail`) with the `.none` priority](https://unix.stackexchange.com/questions/108331/what-does-the-auth-authpriv-none-var-log-syslog-line-mean-in-rsyslog-confi#comment207591_108347), after the initial grab everything (`*.*`), which prevents those logs from going to `/var/log/syslog`.
- `rsyslog` (_at least in Debian, not the original Ubuntu base image_) was configured (_via `postfix` package install_) to create a separate socket to listen to, which Postfix used in `chroot` jail to receive logs from (_as it could not send to `/dev/log` due to the jail_). Presumably the Ubuntu base image may have been different, since the test was specifically to check for Postfix logs, so no chroot jail back then perhaps (_which has been the default for Postfix for a while, but Debian package defaults to `chroot` config_).

Presently this is what `/etc/rsyslog.conf` has configured for this log destination:

```
*.*;mail.none;mail.error;auth,authpriv.none             -/var/log/syslog
```

**EDIT:** Whoops, forgot to read the test content that it failed with :stuck_out_tongue_closed_eyes: 
- It's just `rsyslog` config not able to create a socket due to missing `dev/` directory :man_shrugging: 
- I have confirmed that Postfix without the `chroot` jail defaults to `/dev/log`, so everything is working fine otherwise :+1: 

---

**NOTE:** Github Actions service was acting up, I went to inspect the error logs, and got an error about the workflow run logs not being available anymore, so re-ran the test, and now it's only run the lint test instead of the test suite, hence the PR status of "All checks passed" :sweat_smile: (_hence the tests will still fail_)

polarathene:
> In detail, Podman rootless? Not going to block this PR, just curious about this

There is no confirmation yet that `/dev/shm` works any better for those users. The choice to move the socket there is due to:
- Avoiding creation in `/var/mail-state` since these have no relevance to persist, only relevant at runtime.
  - Potentially additional complications from volume mounting (_storage driver, host platform, container runtime, SELinux_).
- `/dev/shm` chosen only because we currently use it for the supervisor socket (_but this workaround is probably no longer relevant_).

---

Socket files could instead live at `/run` / `/var/run` or similar appropriate locations, in which case we probably should move the Supervisor one too.

The reports from users experiencing issues related to sockets AFAIK is specific to this Dovecot created auth socket and a lack of permissions for Postfix (_and Dovecot in some cases IIRC_) from accessing it. It's highly likely it was just related to bad config on our end with the ownership.

georglauterbach:
So what do we want to do now? (in one sentence please). Move all sockets to `/run/`? That's fine with me. How do we move the other socket files (that @casperklein mentioned) too?

polarathene:
> So what do we want to do now? (in one sentence please). Move all sockets to `/run/`?

No action, leave as-is with `/dev/shm` for now.

> How do we move the other socket files (that @casperklein mentioned) too?

This PR does not need to worry about that.

- Amavis socket could be part of the Amavis config in it's `/etc` config folder, or possibly a default with a setting that could change it.
- The others are for Postfix programs. Presumably they shouldn't be persisted, but these will be created at the `queue_directory` setting, where other relevant state is.
  - `/var/mail-state` could be more granular in symlinks to workaround that (_[expand collapsed sections for reference](https://github.com/docker-mailserver/docker-mailserver/issues/3116#issuecomment-1445240788) - just public + private dirs left that shouldn't persist AFAIK_).

Dockerfile
```diff
@@ -53,7 +53,7 @@ EOF
 # --- Dovecot -----------------------------------
 # -----------------------------------------------
 
-COPY target/dovecot/auth-passwdfile.inc target/dovecot/auth-master.inc target/dovecot/??-*.conf /etc/dovecot/conf.d/
+COPY target/dovecot/*.inc target/dovecot/??-*.conf /etc/dovecot/conf.d/
```
casperklein, line 56:
```suggestion
COPY target/dovecot/*.inc target/dovecot/*.conf /etc/dovecot/conf.d/
```

casperklein:
> How do we move the other socket files (that @casperklein mentioned) too?

I have no idea at the moment. Just mentioned it, while I stumbled upon it.

casperklein:
> Move all sockets to /run/? 

I would stick with "standards", so yes. I am not sure if switching to using a tmpfs could not introduce other issues in the future 🤷

However, there is no urgent need. We can do it later.

georglauterbach:
Auto-merge is on, just FYI :)

casperklein (approved):

polarathene:
> I have no idea at the moment.

AFAIK, they need to reside in `/var/spool/postfix` (_the `main.cf:queue_directory` location_).

I'm not sure why the socket files are created in the `public/` + `private/` dirs there if not using a `chroot` jail, I haven't found a way to configure that differently. Everything else AFAIK is state that should persist.

We could handle it as a special-case I guess, [here's two approaches](https://askubuntu.com/questions/438883/how-to-create-symbolic-link-to-a-directory-excluding-certain-files) to symlink subdirs and avoid some.

> I would stick with "standards", so yes.

v13 could look at changing that to `/run/dms/socket-name` (overlayFS) and moving any other sockets like Amavis there.

> I am not sure if switching to using a tmpfs could not introduce other issues in the future

```console
df -h

Filesystem           Size  Used Avail Use% Mounted on
overlay              120G  115G  5.1G  96% /
tmpfs                 64M     0   64M   0% /dev
shm                   64M     0   64M   0% /dev/shm
/dev/mapper/ventoy2  120G  115G  5.1G  96% /etc/hosts
tmpfs                 16G     0   16G   0% /proc/asound
tmpfs                 16G     0   16G   0% /proc/acpi
tmpfs                 16G     0   16G   0% /proc/scsi
tmpfs                 16G     0   16G   0% /sys/firmware
```

- `/etc/hosts` is from host OS, so whatever filesystem that belongs to on the host will be displayed for it in the container (_same with `Size` of my host partition_).
- These are the only `tmpfs` locations available, unless mounting a `tmpfs` storage driver volume.
- All I know is we've had supervisor socket moved to `/dev/shm` and that doesn't appear to be causing any problems.

casperklein:
> v13 could look at changing that to /run/dms/socket-name (overlayFS) and moving any other sockets like Amavis there.

👍 

> All I know is we've had supervisor socket moved to /dev/shm and that doesn't appear to be causing any problems.

Sure. I once had a problem with /dev/shm, but can't remember exactly atm what that was 😞

casperklein:
My memory returned 😆 Back then I could not execute files in `/dev/shm`. But I later figured out, that it wasn't a tmpfs issue, but how it was mounted (with the noexec option):

`tmpfs  on  /dev/shm  type  tmpfs  (rw,nosuid,nodev,noexec,relatime,size=3221180k)`

PR #3134, opened 2023-02-28
fix: Avoid creating an unnecessary syslog socket for Postfix

polarathene:
# Description

Details are covered [here](https://github.com/docker-mailserver/docker-mailserver/pull/3131#issuecomment-1449067744). This is a fix for Postfix change for `v12` to no longer use `chroot`, ~~as a part of that we're sending logs to `/dev/stdout` instead, which supervisor will maintain a log file of~~ (_no longer the case, continue using syslog as it's better suited than `/dev/stdout` for filtering / redirecting by **facility** and **severity**_).

Raised this PR as requested in linked issue.

<details>
<summary>Original message (no longer applicable)</summary>

I experienced some issues with the Postfix logs no longer being part of `/var/log/mail/mail.log` (_`rsyslog` is configured with a _`mail` facility_ that writes logs with that metadata to this location_), which is also what the console is tailing to stdout after startup completes.

All Postfix logs now live in the supervisor `postfix.log` instead, which is problematic if the logs aren't replicated into `/var/log/mail/mail.log`.

</details>

## Type of change

## Checklist:


georglauterbach:
I was investigating this issue as well, but noticed weird behaviour in some of the test helpers.

I'd be nice if you find a solution to this problem; I will tend to the helpers.

georglauterbach:
What if we just set Postif'x log file to `/var/log/mail/mail.log`?

polarathene:
> What if we just set Postfi'x log file to `/var/log/mail/mail.log` (too)?

As per concerns raised in investigation notes with Postgrey service logs by supervisor, I don't feel it's a good idea to have multiple writers to the same file. Technically I'd expect that wouldn't increase the existing processes writing since it'd just be Supervisor still (_with rsyslog being the other involved that comes to mind_).

We could try it as a workaround / quick fix, at the risk of the log being updated concurrently by both processes at some point messing up the log or corrupting it.

Something I didn't get around to yesterday is checking if without the `chroot` jail, would Postfix use `/dev/log` for syslog, instead of the relative path from it's `queue_directory`? If it would, then this would probably be the best choice for now, otherwise the next best option is to not attempt to remove this `/var/spool/postfix/dev` location (_at least keeping the folder, contents can be removed_).

---

### Vector for log management

One potential proper fix for the logging concern in this PR is to evaluate how viable overhauling our logging is handled. I was looking into [`vector`](https://vector.dev/) for this, which has been interesting to me for quite some time. Supervisor could manage this as a service that can take in [file sources](https://vector.dev/docs/reference/configuration/sources/file/) for the supervisor logs (_[log rotated inputs is supported](https://vector.dev/docs/reference/configuration/sources/file/#file-rotation) just like rsyslog_), as well as [syslog source](https://vector.dev/docs/reference/configuration/sources/syslog/#parsing), where we can do [any necessary transforms](https://vector.dev/docs/setup/quickstart/#hello-syslog).

The config and docs for Vector seems much more pleasant than rsyslog, the syntax is called [VRL](https://vector.dev/docs/reference/vrl/#vector-rust-native) and it's built off rust :stuck_out_tongue: 

Presently the output options (sinks) are lacking one for syslog style. There's a [tracking issue](https://github.com/vectordotdev/vector/issues/6863) related to that, and it seems that they've made progress late last year towards getting the `socket` sink to support syslog formatted logs, but not quite there yet. I came across [this 2021 discussion](https://github.com/vectordotdev/vector/discussions/8155) which might offer something in the interim if we need that. Alternatively, we could just use Vector to remove the syslog metadata that was an issue, and have rsyslog with `imfile()` process that instead of replacing rsyslogs `/var/log/mail/mail.log` file :man_shrugging: 

There is some potential concerns though with Vector and log rotated file sources:
- https://github.com/vectordotdev/vector/issues/15067
- https://github.com/vectordotdev/vector/issues/11742

While Vector can output to a file based log as well, it does not manage log rotation itself. I see that we're using `logrotate` already for that beyond what `supervisor` is doing for the logs it manages. I've come across [this helpful guide](https://www.datadoghq.com/blog/log-file-control-with-logrotate/) on `logrotate`, so can probably make this work :sweat_smile: 

---

I'll need to give the above some more thought, but it might work well for us.

target/scripts/build/packages.sh
```diff
@@ -37,6 +37,9 @@ function _install_postfix
   chmod +x /bin/hostname
   apt-get "${QUIET}" install --no-install-recommends postfix
   mv /bin/hostname.bak /bin/hostname
+
+  # Instead of syslog, master.cf and main.cf we supply are configured to log to /dev/stdout
+  rm /etc/rsyslog.d/postfix.conf
```
casperklein, line 42: We could do that already at build time?

polarathene: This is at build time, right after the `postfix` package has been installed and created it :sweat_smile:

polarathene: Or did you mean changing `master.cf` and `main.cf`? Those could be co-located here, but later in the `Dockerfile` we'd overwrite them with our own copies, not the packages.

This `rm` line could alternatively be moved to start-up scripts if you want co-location.

georglauterbach:
Vector seems very interesting! Keep us posted here, I read through all of it and this is nice!

As for now, we may want to apply "a quick fix" just proving `/var/log/mail/mail.log` for Postfix's log file? This way, we could merge the other PR.

target/scripts/build/packages.sh
casperklein: Oups, my fault. I just saw the script (`packages.sh`) instead of Dockerfile and forgot, that `packages.sh` is executed while building..

casperklein:
Afaik the only "issue" we currently have is this line appearing in /var/log/syslog:

`Mar  3 00:19:10 mail rsyslogd: cannot create '/var/spool/postfix/dev/log': No such file or directory [v8.2102.0 try https://www.rsyslog.com/e/2176 ]`

Just deleting `/etc/rsyslog.d/postfix.conf` should be enough?

Also, with logging to just one file we would lose rsyslog's feature to log by facility:

```
grep ^mail /etc/rsyslog.conf
mail.*                          -/var/log/mail/mail.log
mail.info                       -/var/log/mail/mail.info
mail.warn                       -/var/log/mail/mail.warn
mail.err                        /var/log/mail/mail.err
```

polarathene:
> Just deleting `/etc/rsyslog.d/postfix.conf` should be enough?

Yep, I wasn't paying attention before, and only just realized that myself :joy: 

> Also, with logging to just one file we would lose rsyslog's feature to log by facility:

I also had another misunderstanding there with `postlog` thinking it wasn't leveraging syslog severity levels :sweat_smile: 

Facility I don't think would matter in this sense, as Postfix just uses the `mail` facility? This PR was about writing Postfix logs to a separate file, and then thinking how to get a copy back into `/var/log/mail/mail.log`.

All good. I think there is benefit in bringing Vector in at a later point. But for now there's a quick and easy fix :+1: (_only removing `/etc/rsyslog.d/postfix.conf`_)

target/postfix/main.cf
```diff
@@ -17,6 +17,9 @@ recipient_delimiter = +
 inet_interfaces = all
 inet_protocols = all
 
+# Logging (instead of syslog default, output to stdout for supervisor)
+maillog_file = /dev/stdout
+
```
polarathene, line 22: No benefit, reverting:

```suggestion
```

target/postfix/master.cf
```diff
@@ -73,3 +73,4 @@ virtual        unix  -       n       n       -       -       virtual
 lmtp           unix  -       -       n       -       -       lmtp
 anvil          unix  -       -       n       -       1       anvil
 scache         unix  -       -       n       -       1       scache
+postlog  unix-dgram  n       -       n       -       1       postlogd
```
polarathene, line 76: This should be fine to keep, aligning with upstream `master.cf`. [`postlogd`](https://www.postfix.org/postlogd.8.html) explains that it's mostly for when you don't want to use syslog, by configuring `mail.cf:maillog_file`. [`postlog`](https://www.postfix.org/postlog.1.html) handles the logging, and defaults to syslog.

target/scripts/build/packages.sh
```diff
@@ -37,6 +37,9 @@ function _install_postfix
   chmod +x /bin/hostname
   apt-get "${QUIET}" install --no-install-recommends postfix
   mv /bin/hostname.bak /bin/hostname
+
+  # Instead of syslog, master.cf and main.cf we supply are configured to log to /dev/stdout
```
polarathene, line 41:
```suggestion

  # We are not using Debian's default `chroot` jail config for Postfix, this config is not needed:
```

polarathene (commented):
Revision based on discussions above and [here](https://github.com/docker-mailserver/docker-mailserver/pull/3131#issuecomment-1452768189).

georglauterbach (dismissed):

georglauterbach:
When @casperklein's review is finished, I can merge this even if GH actions is till bugged. Just tell me :)

target/scripts/build/packages.sh
```diff
@@ -37,6 +37,9 @@ function _install_postfix
   chmod +x /bin/hostname
   apt-get "${QUIET}" install --no-install-recommends postfix
   mv /bin/hostname.bak /bin/hostname
+
+  # We are not using Debian's default `chroot` jail config for Postfix, this config is not needed:
```
polarathene, line 41: Rewording to trigger CI:

```suggestion
  # Irrelevant - Debian's default `chroot` jail config for Postfix needed a separate syslog socket:
```

polarathene:
> When @casperklein's review is finished

I think it's finished, the concern was resolved :+1:

polarathene:
`setup_cli.bats` failed with:

```
not ok 61 [setup.sh] exit with error when wrong arguments provided in 264ms
# (from function `assert_line' in file test/test_helper/bats-assert/src/assert_line.bash, line 134,
#  in test file test/tests/parallel/set3/scripts/setup_cli.bats, line 37)
#   `assert_line --index 0 --partial "The command 'lol troll' is invalid."' failed
# /home/runner/work/docker-mailserver/docker-mailserver/test/test_helper/bats-assert/src/assert_line.bash: line 134: lines: parameter not set
```

https://github.com/docker-mailserver/docker-mailserver/blob/9e2f96448ad4f5afd897621c54a731bbb18e12bb/test/tests/parallel/set3/scripts/setup_cli.bats#L34-L38

Not sure what caused that failure. `assert_failure` was successful at detecting failure, but no error message was output?

Re-ran test and it passes now :man_shrugging:

georglauterbach (approved):

georglauterbach:
@casperklein when you approve, please go ahead and merge this straight away :)

casperklein (approved):

PR #3143, opened 2023-03-03
rspamd: rename `ENABLE_REDIS` & add persistence for Redis

### LINKED ISSUE #3138: [BUG] rspamd forgets it learned spam and ham when docker-compose down and up is issued

### Affected Component(s)

rspamd will not easily learn to recognize spam

I have trained rspamd with two email-folders containing spam or ham like so

First get stats on learned spam
`docker exec -it mailserver sh
rspamc stat

Statfile: BAYES_SPAM type: redis; length: 0; free blocks: 0; total blocks: 0; free: 0.00%; learned: 1; users: 1; languages: 0
Statfile: BAYES_HAM type: redis; length: 0; free blocks: 0; total blocks: 0; free: 0.00%; learned: 6; users: 1; languages: 0
Total learns: 7
`

Then I train rspamd with a folder full of spam, and a folder full of spam

`rspamc learn_spam /var/mail/fanscees.com/hanscees/.spam/cur
rspamc stat
Statfile: BAYES_SPAM type: redis; length: 0; free blocks: 0; total blocks: 0; free: 0.00%; learned: 413; users: 1; languages: 0
Statfile: BAYES_HAM type: redis; length: 0; free blocks: 0; total blocks: 0; free: 0.00%; learned: 6; users: 1; languages: 0
Total learns: 419
`
Same with ham training 
`
rspamc learn_ham /var/mail/hanscees.com/hanscees/.Ham/cur
rspamc stat
Statfile: BAYES_SPAM type: redis; length: 0; free blocks: 0; total blocks: 0; free: 0.00%; learned: 413; users: 1; languages: 0
Statfile: BAYES_HAM type: redis; length: 0; free blocks: 0; total blocks: 0; free: 0.00%; learned: 354; users: 1; languages: 0
Total learns: 767
`

I also have enabled that rspamd learns from users when they move email in and out the inbox.junk folder: see attached txt script 
[rspamd-patches.txt](https://github.com/docker-mailserver/docker-mailserver/files/10864574/rspamd-patches.txt)

Problem is now, after doing a docker-compose down, all learned material seems gone.

`docker-compose down
docker-compose up -d
Statfile: BAYES_SPAM type: redis; length: 0; free blocks: 0; total blocks: 0; free: 0.00%; learned: 0; users: 0; languages: 0
Statfile: BAYES_HAM type: redis; length: 0; free blocks: 0; total blocks: 0; free: 0.00%; learned: 0; users: 0; languages: 0
Total learns: 0
`

I think something should be added to the mailstate that is preserved, but no idea what. 

I expect learned spam to persist over a reset of the docker image

do what I did above

### DMS version

1301

Linux

photon 4

AMD64 / x86_64

Docker

### docker-compose.yml

```yml
services:
  mailserver:
    image: docker.io/mailserver/docker-mailserver:edge
    #image: docker.io/mailserver/docker-mailserver:latest
    container_name: mailserver
    hostname: mail
    domainname: fanscees.net
    env_file: mailserver.env
    ports:
      - "995:995"  # pop3 (implicit TLS)
      - "110:110"  # pop3 (implicit TLS)
    volumes:
      #- etc:/etc
      - maildata:/var/mail/
      - mailstate:/var/mail-state/
      - log:/var/log/
      - Tconfig:/tmp/docker-mailserver/
      - ssl:/tmp/ssl
    restart: always
    stop_grace_period: 1m
    environment:
     # If you need SSL connection, you can provide your own certificates
     - SSL_TYPE=manual
    cap_add:
      - NET_ADMIN
    healthcheck:
      timeout: 3s
      retries: 0
volumes:
  maildata:
  mailstate:
  Tconfig:
  log:
  ssl:
```

### Relevant log output

_No response_

I have also enabled the auto-learn rule on spam form your documentation

### Code of conduct

_No response_

georglauterbach:
# Description

This PR

1. renames `ENABLE_REDIS` -> `ENABLE_RSPAMD_REDIS`, see 064968362767be349abe3698ec9c78b295347ff7 for details
2. enables persistence for the Redis DB as detailed in #3138

Fixes #3138

## Type of change

## Checklist:


target/scripts/startup/setup.d/security/rspamd.sh
```diff
@@ -57,6 +57,19 @@ servers = "127.0.0.1:6379";
 expand_keys = true;
 
 EOF
+
+    # Here we adjust the Redis default configuration that we supply to Redis
+    # when starting it. Note that `/var/lib/redis/` is linked to
+    # `/var/mail-state/redis/` (for persisting it) if `ONE_DIR=1`.
+    sed -i -E                                  \
```
casperklein, line 64:
```suggestion
    sedfile -i -E                                  \
```

casperklein: Just this minor improvement, otherwise LGTM.

casperklein (approved):

polarathene (approved):
LGTM :+1: 

- Be mindful of Redis version and the major bump when Debian releases Bookworm, and reliance on implicit config (_especially when it differs from what Redis ships / documents_).
- **Persistence** - This PR enables (RDB type) implicitly (_and whatever else that upstream Redis defaults do not_).
- **Logging** - Discussed syslog instead of supervisor managed, but I wouldn't worry about it for now.
- **Docs** - Possible improvements, up to you. This review is self-documenting enough for me atm :sweat_smile: 

---

Bullseye is shipping v6 of Redis, note that the upcoming Bookworm release later this year bumps that to v7 (released April 2022).

The change from a command to using the default shipped config, was primarily motivated by the implicit config to enable persistence. Note that this is [not advised by upstream](https://redis.io/docs/management/config/):

> Redis is able to start without a configuration file using a built-in default configuration, however this setup is only recommended for testing and development purposes.
> 
> The proper way to configure Redis is by providing a Redis configuration file, usually called `redis.conf`.

I believe that is referring to just running `redis-server` and using args to override the defaults (_which presumably match the [upstream default config](https://redis.io/docs/management/config-file/), which Debian isn't aligned with_).

Just something to keep in mind, especially once the transition to v7 arrives and implicit config defaults are relied on.

### Persistence

[Persistence support](https://redis.io/docs/management/persistence/) has some behaviour differences between these two versions, but most seem to be related to AOF type which does seem nice at least for v7, but RDB type disadvantages don't appear to be a concern for our usage, and should be simpler to backup (_potentially not as much for some users, where Docker is wrapped in a VM such as on macOS, which IIRC uses file sync to manage volume mounts_).

Default config is using RDB (_Present in Debian shipped config, [upstream Redis doesn't](https://redis.io/docs/management/config-file/)_):

```
# Enable RDB
save 900 1
save 300 10
save 60 10000

# Disable AOF
appendonly no
```

Translates to persist triggers:
- After 900 sec (15 min) if at least 1 key changed
- After 300 sec (5 min) if at least 10 keys changed
- After 60 sec if at least 10000 keys changed

Also, as [mentioned in the associated issue](https://github.com/docker-mailserver/docker-mailserver/issues/3138#issuecomment-1453519599), when persistence is performing an update to disk, double the memory used for the in-memory store is required as a precaution (_but unlikely needed due to CoW_).

It might be worth [referencing that Redis FAQ entry](https://redis.io/docs/getting-started/faq/#background-saving-fails-with-a-fork-error-on-linux) in our docs for Redis, or just expecting a maintainer or user to know about it when troubleshooting. The link explains how to avoid the issue, but I think is required to be done on the host or via adding a capability to the compose config.

For now, it's at least documented in this review and associated issue.

---

### Logging

This section can probably can be ignored for this PR.

I'm not familiar with how to configure syslog to redirect logs to a specific file for a given syslog tag, and not too interested in maintaining that. May be better to wait until Vector is considered.

If we want to instead log to syslog instead of having supervisord write the `/dev/stdout` / `/dev/stderr` logs to a file it manages, we can use `syslog-enabled yes`. Here's relevant redis config:

```
# To enable logging to the system logger, just set 'syslog-enabled' to yes,
# and optionally update the other syslog parameters to suit your needs.
# syslog-enabled no

# Specify the syslog identity.
# syslog-ident redis

# Specify the syslog facility. Must be USER or between LOCAL0-LOCAL7.
# syslog-facility local0
```

[Here are the Syslog facility choices](https://en.wikipedia.org/wiki/Syslog#Facility) that can be used. ` syslog-facility` default is probably fine, as is `syslog-ident`.

Looking at the current `/etc/rsyslog.conf`, this would route logs to `/var/log/syslog` for that generic facility, and probably also to `/var/log/messages` (_intended for info, notice, warn logs, but excludes other facilities like `mail`_).

georglauterbach:
Thanks for the detailed comment @polarathene! I will make sure to check this out once we update the base image.

PR #3149, opened 2023-03-04
bugfix: special bits for maildrop and public directory

georglauterbach:
# Description

After changing the group, special bits are lost, but they should be set for the directories `/var/spool/postfix/{maildrop,public}`, otherwise you see the following error:

```txt
postfix/postdrop[17400]: warning: mail_queue_enter: create file maildrop/729504.17400: Permission denied
```

The web page <https://linux.m2osw.com/snapwebsites-postfixpostdrop18189-warning-mailqueueenter-create-file-maildrop25937318189-permission> provides the solution to restoring the correct permissions.

---

This is a blocker for v12.0.0.

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/startup/setup.d/mail_state.sh
+++ b/target/scripts/startup/setup.d/mail_state.sh
@@ -85,8 +85,8 @@ function _setup_save_states
     chgrp -R postdrop /var/mail-state/spool-postfix/{maildrop,public}
     # After changing the group, special bits (set-gid, sticky) may be stripped, restore them:
     # Ref: https://github.com/docker-mailserver/docker-mailserver/pull/3149#issuecomment-1454981309
-    chmod 1730 /var/spool/postfix/maildrop
-    chmod 2710 /var/spool/postfix/public
+    chmod 1730 /var/mail-state/spool-postfix/maildrop
+    chmod 2710 /var/mail-state/spool-postfix/public
   elif [[ ${ONE_DIR} -eq 1 ]]
   then
     _log 'warn' "'ONE_DIR=1' but no volume was mounted to '${STATEDIR}'"
--- a/target/scripts/startup/setup.d/mail_state.sh
+++ b/target/scripts/startup/setup.d/mail_state.sh
@@ -83,9 +83,10 @@ function _setup_save_states
     chown root:root /var/mail-state/spool-postfix
     # These two require the postdrop(103) group:
     chgrp -R postdrop /var/mail-state/spool-postfix/{maildrop,public}
-    # After changing the group, special bits (set-gid, sticky) are lost and we need to restore them:
-    chmod 3730 /var/spool/postfix/maildrop
-    chmod 2730 /var/spool/postfix/public
+    # After changing the group, special bits (set-gid, sticky) may be stripped, restore them:
+    # Ref: https://github.com/docker-mailserver/docker-mailserver/pull/3149#issuecomment-1454981309
+    chmod 1730 /var/spool/postfix/maildrop
+    chmod 2710 /var/spool/postfix/public
   elif [[ ${ONE_DIR} -eq 1 ]]
   then
     _log 'warn' "'ONE_DIR=1' but no volume was mounted to '${STATEDIR}'"
--- a/target/scripts/startup/setup.d/mail_state.sh
+++ b/target/scripts/startup/setup.d/mail_state.sh
@@ -83,6 +83,9 @@ function _setup_save_states
     chown root:root /var/mail-state/spool-postfix
     # These two require the postdrop(103) group:
     chgrp -R postdrop /var/mail-state/spool-postfix/{maildrop,public}
+    # After changing the group, special bits (set-gid, sticky) are lost and we need to restore them:
+    chmod 3730 /var/spool/postfix/maildrop
+    chmod 2730 /var/spool/postfix/public
   elif [[ ${ONE_DIR} -eq 1 ]]
   then
     _log 'warn' "'ONE_DIR=1' but no volume was mounted to '${STATEDIR}'"


casperklein (dismissed):

target/scripts/startup/setup.d/mail_state.sh
```diff
@@ -83,6 +83,9 @@ function _setup_save_states
     chown root:root /var/mail-state/spool-postfix
     # These two require the postdrop(103) group:
     chgrp -R postdrop /var/mail-state/spool-postfix/{maildrop,public}
+    # After changing the group, special bits (set-gid, sticky) are lost and we need to restore them:
+    chmod 3730 /var/spool/postfix/maildrop
+    chmod 2730 /var/spool/postfix/public
```
polarathene, line 88: These don't align with what I see when no `/var/mail-state` is present:

```console
$ exa -l --octal-permissions /var/spool/postfix

1730 drwx-wx--T - postfix  3 Mar 00:18 maildrop
0710 drwx--x--- - postfix  3 Mar 11:06 public
```

It's possible that this has changed upstream since that Aug 2018 article?

- `public/` dir only gives execute permissions for the group, and the contents there are sockets to executables, so that makes sense?
- `maildrop/` additionally allows a group to write files, with a [sticky bit](https://en.wikipedia.org/wiki/Sticky_bit) set (_[this is also a good reference](https://en.wikipedia.org/wiki/Setuid#Sticky_bit) for the various special permissions_):
  > For directories, when a directory's sticky bit is set, the filesystem treats the files in such directories in a special way so only the file's owner, the directory's owner, or [root user](https://en.wikipedia.org/wiki/Root_user) can rename or delete the file.

Increasing `maildrop/` permissions from `1730` to `3730` (_sticky bit + SGID_) doesn't seem neccessary? Nor should `public/` need to change from `0710` to `2730` (_write + SGID_).

```suggestion
    # After changing the group, special bits are lost and we need to restore them (sticky bit in this case):
    chmod 1730 /var/spool/postfix/maildrop
```

It's unlikely, but perhaps possible this was related to Postfix 2.10 changing from named pipes (FIFO) to unix sockets, but that was released in 2013, so should have been available in whatever distro the author used in 2018.

polarathene (changes requested):

polarathene:
> ```
> postfix/postdrop[17400]: warning: mail_queue_enter: create file maildrop/729504.17400: Permission denied
> ```

How does `postdrop` / `maildrop` get triggered btw? We [don't configure it in `master.cf`](https://github.com/docker-mailserver/docker-mailserver/blob/8ec5dbec74839e0460252bcc2115035310f49327/target/postfix/master.cf).

---

**EDIT:** This was from me recalling [removing `maildrop` in `master.cf` a while back](https://github.com/docker-mailserver/docker-mailserver/pull/2272), but looking that PR up I point out they're not the same thing :sweat_smile: 

They're explained in Postfix docs for anyone curious:

- https://www.postfix.org/MAILDROP_README.html
- https://www.postfix.org/QSHAPE_README.html#maildrop_queue (_[`postdrop`](https://www.postfix.org/postdrop.1.html) explained there, it requires SGID for `maildrop/`?_)

For reference `postfix` installed on Alpine has `maildrop/` permissions of `0730` :man_shrugging: 

Perhaps see if it works with `1730`, like it is by default in Debian, otherwise if really does need `3730` ok then :+1:

target/scripts/startup/setup.d/mail_state.sh
casperklein: Default permissions in Debian 11:

```
stat -c '%a %A %n' /var/spool/postfix/{maildrop,public}

1730 drwx-wx--T /var/spool/postfix/maildrop
2710 drwx--s--- /var/spool/postfix/public
```

If I stop postfix, remove those two directories and start postfix again, the directories are recreated with the following permissions:

```
stat -c '%a %A %n' /var/spool/postfix/{maildrop,public}

730 drwx-wx--- /var/spool/postfix/maildrop
710 drwx--x--- /var/spool/postfix/public
```

polarathene: I don't think I deleted the folders, forgot to mention that I did only check in an container I already had running restarted Postfix in among probably some other things, my bad! :sweat_smile: 

I have done a `git pull` and rebuilt the image, and see the same permissions @casperklein mentions without `/var/mail-state` present. I've also mounted with the state dir and exited the container to start it again with that volume still attached, permissions are untouched.

Only when removing those folders explicitly and starting Postfix did they lose those special permissions.

polarathene:
I found [this mailing list discussion from 8 years ago](https://mailing.postfix.users.narkive.com/iEFJACGh/ownership-permissions-of-var-spool-postfix#post6) which directs us to [`post-install`](https://github.com/vdukhovni/postfix/blob/2047994eeee1dbf0d8c0537ba00acb8986a8a370/postfix/conf/post-install#L569-L579) and [`postfix-files`](https://github.com/vdukhovni/postfix/blob/2047994eeee1dbf0d8c0537ba00acb8986a8a370/postfix/conf/postfix-files#L64-L65). It's a bit hard for me to follow how it gets `2710` from there instead of `0710` though. There's three commands that are set `2755` though, [one of them is `postdrop`](https://github.com/vdukhovni/postfix/blob/2047994eeee1dbf0d8c0537ba00acb8986a8a370/postfix/conf/postfix-files#L139) (_`$command_directory` maps to `/usr/sbin/` I think?_).

[`postfix check`](https://github.com/vdukhovni/postfix/blob/2047994eeee1dbf0d8c0537ba00acb8986a8a370/postfix/conf/postfix-script#L238-L243) doesn't raise any issues about the lack of special permissions.

---

No `/var/mail-state` volume, using `chgrp` isn't stripping the permission bits?:

```console
$ chgrp amavis /var/spool/postfix/{maildrop,public}
$ stat -c '%a %A %n' /var/spool/postfix/{maildrop,public}
1730 drwx-wx--T /var/spool/postfix/maildrop
2710 drwx--s--- /var/spool/postfix/public

$ ls -l /var/spool/postfix/ | grep -E 'maildrop|public'
drwx-wx--T 1 postfix amavis    6 Mar  5 03:03 maildrop
drwx--s--- 1 postfix amavis  110 Mar  5 03:28 public
```

I looked into why, and it turns out that this is due to a syscall, so is dependent upon the kernel rather than anything we control within the container :thinking: 

- https://unix.stackexchange.com/questions/313409/howto-prevent-chgrp-from-clearing-setuid-bit/313425#313425
- https://unix.stackexchange.com/questions/53665/chown-removes-setuid-bit-bug-or-feature/53675#53675

---

I'm having trouble finding a source on the proposed `3730` + `2730` changes. It seems we should just match what we're getting with Debian unless there is reason that doesn't work?

target/scripts/startup/setup.d/mail_state.sh
```diff
@@ -83,6 +83,10 @@ function _setup_save_states
     chown root:root /var/mail-state/spool-postfix
     # These two require the postdrop(103) group:
     chgrp -R postdrop /var/mail-state/spool-postfix/{maildrop,public}
+    # After changing the group, special bits (set-gid, sticky) may be stripped, restore them:
+    # Ref: https://github.com/docker-mailserver/docker-mailserver/pull/3149#issuecomment-1454981309
+    chmod 1730 /var/spool/postfix/maildrop
+    chmod 2710 /var/spool/postfix/public
```
polarathene, line 89: Should probably match `/var/mail-state/spool-postfix/` path directly like the other lines do?

```suggestion
    chmod 1730 /var/mail-state/spool-postfix/maildrop
    chmod 2710 /var/mail-state/spool-postfix/public
```

polarathene (approved):
LGTM :+1: 

I've also included a reference to this PR since refactoring efforts have not made `git blame` hunting a fun time :sweat_smile: 

Plenty of resources linked in this PR discussion to provide additional clarity (_like the syscall gotcha_) just in case future maintainers thought it was no longer needed :sweat_smile:

georglauterbach:
Fix seems to be working on my setup ❤️ - I will monitor it though :)

georglauterbach:
Turns out this issue was caused by `allowPrivilegeEscalation: false` in my setup, which caused the `postdrop` binary to run as another user.. Follow-up is #3619

Issue #3133, opened 2023-02-28
[BUG] Cannot find a valid DOMAIN for '/etc/letsencrypt/live/<DOMAIN>/', tried: '', 'subdomain.domain.tld', 'domain.tld'


alexanderadam:
### Miscellaneous first checks

- [X] I checked that all ports are open and not blocked by my ISP / hosting provider.
- [X] I know that SSL errors are likely the result of a wrong setup on the user side and not caused by DMS itself. I'm confident my setup is correct.

### Affected Component(s)

DMS

### What happened and when does this occur?

I'm trying to start DMS on unprivileged Podman but once I activate LetsEncrypt the server kills itself and I see this in the log:

```
[  ERROR  ]  Cannot find a valid DOMAIN for '/etc/letsencrypt/live/<DOMAIN>/', tried: '', 'subdomain.domain.tld', 'domain.tld'
[  ERROR  ]  _find_letsencrypt_domain | LETSENCRYPT_DOMAIN appears to be misconfigured, please verify.
[  ERROR  ]  Shutting down
[  ERROR  ]  _find_letsencrypt_key | LETSENCRYPT_DOMAIN appears to be misconfigured, please verify.
[  ERROR  ]  Shutting down
[  ERROR  ]  No difference after call to 'sed' in 'sedfile' (sed -i -r s|^(smtpd_tls_chain_files =).*|\1 /etc/letsencrypt/live//.pem /etc/letsencrypt/live//fullchain.pem| /etc/postfix/main.cf)
[  ERROR  ]  No difference after call to 'sed' in 'sedfile' (sed -i -r -e s|^(ssl_key =).*|\1 </etc/letsencrypt/live//.pem| -e s|^(ssl_cert =).*|\1 </etc/letsencrypt/live//fullchain.pem| /etc/dovecot/conf.d/10-ssl.conf)
```

I learned through issue #2405 that I shouldn't set `LETSENCRYPT_DOMAIN` (I just set it because of the error message).
The container can access the letsencryption configuration and I can also see that the directory and certs in `/etc/letsencrypt/live/subdomain.domain.tld` exist.

### What did you expect to happen?

It should start without errors or give at least a useful error message on misconfigurations

### How do we replicate the issue?

I just tried to start it

### DMS version

edge

### What operating system is DMS running on?

Linux

### Which operating system version?

Ubuntu 22.04

### What instruction set architecture is DMS running on?

AMD64 / x86_64

[...]

PR #3155, opened 2023-03-05
scripts: improve panic helpers

### LINKED ISSUE #3140: [BUG] `_shutdown` will continue when called in a subshell (causing UB)
# Description

Not too important.

[A user showed a particular failure scenario](https://github.com/docker-mailserver/docker-mailserver/issues/3133) where shutdown was called twice, and additionally two `sedfile` errors that came afterwards, but it should have stopped from the first shutdown.

The function involved likely continues running instead of exiting early on failure. This may be the case for other scripts that invoke shutdown on failure. A workaround could just involve the logger being aware of the shutdown call to silence it's logging from then on? :man_shrugging: 

Or we can just leave it as-is. I believe there's some desired changes in future for running bash scripts that handle failure better? Or the potential that we migrate away from bash scripts.

georglauterbach:
# Description

`_shutdown` wil now wait for Supervisord by default. Added option to exit immedaitely though. Adjusted panic handlers; made them more coherent and added one more panic type.

Fixes #3140

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/helpers/error.sh
+++ b/target/scripts/helpers/error.sh
@@ -22,9 +22,10 @@ function _exit_with_error
 # PANIC_SCOPE => Optionally provide a string for debugging to better identify/locate the source of the panic.
 function dms_panic
 {
-  local PANIC_TYPE=${1}
-  local PANIC_INFO=${2}
-  local PANIC_SCOPE=${3} #optional
+  local PANIC_TYPE=${1:-}
+  local PANIC_INFO=${2:-}
+  local PANIC_SCOPE=${3-} # optional, must not be :- but just -
+  local PANIC_STRATEGY=${4:-} # optional
 
   local SHUTDOWN_MESSAGE
 
@@ -49,6 +50,10 @@ function dms_panic
       SHUTDOWN_MESSAGE="Invalid value for ${PANIC_INFO}!"
       ;;
 
+    ( 'general' )
+      SHUTDOWN_MESSAGE=${PANIC_INFO}
+      ;;
+
     ( * ) # `dms_panic` was called directly without a valid PANIC_TYPE
       SHUTDOWN_MESSAGE='Something broke :('
       ;;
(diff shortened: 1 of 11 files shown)


casperklein:
Remember, when we once had the styling discussion? I wasn't a fan of putting everything in functions, even when they are only called once. This is the downside of it - the need for subshells to assign values 😉

But I have to admit, that I started liking it 😆

target/scripts/helpers/error.sh
```diff
@@ -49,34 +50,49 @@ function dms_panic
...
+  if [[ ${2:-wait} == 'immediate' ]]
+  then
+    # In case the user requested an immediate exit, he ensure he is not in a subshell
+    # call and exiting the whole script is safe. This way, we make the shutdown quicker.
+    exit 1
+  else
+    # We can simply wait until Supervisord has terminated all processes; this way,
+    # we do not return from a subshell call and continue as if nothing happened.
+    sleep 1000
+  fi
```
polarathene, line 97: Would it not be cleaner to approach with the first panic call (`_shutdown()` rather) creating some state like `/tmp/panic`, and just checking if this file exists to block with the `sleep 1000`?

No need for the `immediate` arg then, which avoids calls with empty arg before it for scope.

At least that seems simpler and pretty much always the behaviour desired? (_`exit 1` first call, any subsequent calls block_)

In the [cited failure case](https://github.com/docker-mailserver/docker-mailserver/issues/3133), the log should stop where it expected from the looks of it.

I understand the drawback would be scripts continue to run after the first `exit 1`? But I don't quite understand how that `exit 1` interacts with the scope of the `ssl.sh` helper, these are all panic calls within the same method, so what happens if `sleep 1000` was at the end of the panic call? It always hits the `sleep 1000`?

We are avoiding the `immediate` arg only for `ssl.sh` and `dns.sh`, if keeping the arg approach, it may be simpler to invert this arg to opt-in to blocking instead?

polarathene (approved):
It might make sense to use a different approach and not think about the arg dance, or if blocking is important inverting the arg opt-in.

Feel free to merge if you disagree :+1:

target/scripts/helpers/error.sh
georglauterbach: I chose not to take the file approach because I really dislike scripts continuing after `_shutdown` was called (potentially causing behaviour we do not expect at all). I would definitely avoid going to UB-lane.

This is also the reason for the arg dance; using `sleep` is the conservative but also very safe method. Hence, it should be the default. I am not a big fan of `''` args either, but it's working fine.

I'd keep it as is for now. If we ever feel the need to tackle this again, I will assist.

georglauterbach:
> It might make sense to use a different approach and not think about the arg dance, or if blocking is important inverting the arg opt-in.
> 

I provided feedback to your comment :) Hope that helps in explaining my choices.

> Feel free to merge if you disagree :+1: 

I think this is fine for now :)

PR #3160, opened 2023-03-05
fix: postsrsd restart loop

casperklein:
# Description

Fixes https://github.com/docker-mailserver/docker-mailserver/pull/3128#discussion_r1123885007

Can be tested with:

```bash
docker build -t test .
docker run --rm -it --hostname example.test --env SMTP_ONLY=1 --env ENABLE_SRS=1  --env SUPERVISOR_LOGLEVEL=info test
```

## Type of change

## Checklist:


georglauterbach:
I had really hoped not to see the return of wrappers 🤢 but it may currently be our only option..

casperklein:
Yes, unless someone comes up with a proper solution to run in foreground, this is the only way.

target/scripts/wrapper/postsrsd.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+function postsrsd_stop
+{
+  /etc/init.d/postsrsd stop # does not properly stop postsrsd process
+  # shellcheck disable=SC2046
+  kill $(pidof postsrsd)
+}
+
+trap postsrsd_stop EXIT
```
georglauterbach, line 10: I remember that other wrappers had traps for SIGINT and SIGTERM as well - I think Supervisord will send these signals actually, so shouldn't we have these trapped as well?

georglauterbach: 👍🏼

georglauterbach (dismissed):
@polarathene merge this when you're fine with the changes as well :)

polarathene:
I will review in approx 12 hours.

I found out why I could not reproduce the problem on my system... My current install has a different `docker.service` config for `LimitNOFILE=infinity`, causing that ulimit problem with container having over a billion FDs to close.

It appeared like `postsrsd` was running to me, but it was stalling for 8 minutes to startup as it iterated through all those syscalls :sweat_smile: So now I can reproduce :+1:

polarathene:
Sorry, went on a bit of a detour, I'll get back to this today :sweat_smile:

polarathene:
> Yes, unless someone comes up with a proper solution to run in foreground, this is the only way.

Tada :tada: https://github.com/docker-mailserver/docker-mailserver/pull/3128#issuecomment-1462062616

There isn't any tests in place that actually verify postsrsd is working, only config matches expectations, but I'm pretty sure that's working properly now.

What would you like to use?

target/supervisor/conf.d/supervisor-app.conf
```diff
@@ -148,7 +148,7 @@ autostart=false
 autorestart=true
 stdout_logfile=/var/log/supervisor/%(program_name)s.log
 stderr_logfile=/var/log/supervisor/%(program_name)s.log
-command=/etc/init.d/postsrsd start
+command=/bin/bash -l -c /usr/local/bin/postsrsd.sh
```
polarathene, line 151:
```suggestion
command=command=/bin/bash -c 'env $(grep -vE "^(#.*|\s*)$" /etc/default/postsrsd) postsrsd -e -p /var/run/postsrsd.pid'
```

georglauterbach:
> > Yes, unless someone comes up with a proper solution to run in foreground, this is the only way.
> 
> 
> 
> Tada :tada: https://github.com/docker-mailserver/docker-mailserver/pull/3128#issuecomment-1462062616
> 
> What would you like to use?

_Definitely_ your solution 🎉🚀

casperklein:
I can confirm this works 🚀

I noticed this difference between the old wrapper and your new foreground solution:

Old:

`/usr/sbin/postsrsd -f 10001 -r 10002 -d example.test -s /etc/postsrsd.secret -a = -n 4 -N 4 -u postsrsd -p /var/run/postsrsd.pid -c /var/lib/postsrsd -l 127.0.0.1 -D -X`

New:

`postsrsd -e -p /var/run/postsrsd.pid`

Can you confirm, that this does not introduce any regressions due to missing parameters? Are these all either default values or defined in `/etc/default/postsrsd`?

polarathene:
> Can you confirm, that this does not introduce any regressions due to missing parameters? Are these all either default values or defined in `/etc/default/postsrsd`?

You only need to provide the domain and secret I think to run the command, the defaults are shown with the `-h` command flag and can be compared to the init script and config file we ship (_slightly out of date from upstream and init script IIRC_).

I provided [captured snippets and relevant details](https://github.com/docker-mailserver/docker-mailserver/pull/3128#issuecomment-1452981719) for reference a while ago.

There's possibly something missing, I need to sleep but can double check tomorrow to confirm if parity is kept :+1: 

That said we may want to opt-out of the chroot (along with Dovecot) for v12?

casperklein:
> There's possibly something missing, I need to sleep but can double check tomorrow to confirm if parity is kept 👍

👍 

> That said we may want to opt-out of the chroot (along with Dovecot) for v12?

I see no problem with that. But let's do it separated from this PR.

polarathene:
Yep all looks good :+1:

`postsrsd` 1.x series will take either args to config, or with `-e` [match these ENV](https://github.com/roehling/postsrsd/blob/f21c009c56d25c94bce998b26be3fa50e3c08470/postsrsd.c#L427-L450), otherwise use the defaults listed via `-h` (_which is the case for anything we don't provide_).

I referenced the [`init` script + `-h` output I shared earlier](https://github.com/docker-mailserver/docker-mailserver/pull/3128#issuecomment-1452981719) , the [PostSRSd ENV config file we provide](https://github.com/docker-mailserver/docker-mailserver/blob/9e2f96448ad4f5afd897621c54a731bbb18e12bb/target/postsrsd/postsrsd) and [PostSRSd official config](https://github.com/roehling/postsrsd/blob/f21c009c56d25c94bce998b26be3fa50e3c08470/init/postsrsd.default.in) (_ours is slightly different, but aligns with the modifications Debian package already applies, ours just is outdated missing some added settings, but the defaults are the same_).

> I see no problem with that. But let's do it separated from this PR.

:100:

polarathene (approved):
LGTM :rocket:

Issue #3173, opened 2023-03-10
[FR] Make it easier for people to find the docs for the latest release


jrpear:
# Feature Request

Essentially, I'd like to change all user-facing documentation links to point to `latest` rather then `edge`. I also propose changing the docs theme to make it more obvious what documentation version you're using.

## Context

#3169 

### Is your Feature Request related to a Problem?

I was confused about the absence of an unreleased feature from the latest release because I was reading the docs for edge rather than the latest release.

I remember seeing the message on the landing page that "This documentation is versioned". I didn't know that I'd pulled "latest" rather than "edge" from docker hub. By the time I'd encountered the version-related issue, I'd forgotten that I was on the `edge` documentation rather than `latest`.

Here's what I did:

* I used the docs linked from project description, which took me to the docs for edge.

* I started on [the usage page](https://docker-mailserver.github.io/docker-mailserver/edge/usage/). That page had me download `docker-compose.yml` and `mailserver.env` from `master`.
  * The path of least resistance had me working with master/edge rather than the latest release, which is maybe not ideal, but it was consistent so far.

* I edited `mailserver.env` according to the comments in it (big fan of this workflow btw).

* I ran `docker-compose`. The `docker-compose.yml` from `master` pulls `latest`, not `edge`. So now I was working with the image from `latest`, but with the docs and commented `mailserver.env` from `edge`.

* Not wanting to bother with DKIM initially, I set `ENABLE_OPENDKIM=0`, an `edge`-only feature. It didn't work, so I opened 
#3169 in error.

#### Change the user-facing links

I think we should replace every user-facing link to the `edge` docs with a link to the `latest` docs. Everyone who is getting started with docker-mailserver will start with the latest release. So this would mean certainly changing the **project description**, **README**,  **docker hub description**, **`mailserver.env` comments**, and maybe more.

...But what about links in the docs? Right now the `edge` "Usage" page has people download `mailserver.env` and `docker-compose.yml` from master. How will each version after 12 tell people to download the right `docker-compose.yml` and the right `mailserver.env`? Does this problem show up anywhere else in the docs? Has it already been solved somehow? I don't know. Does anyone know?

#### Make the documentation version easier to identify

Change the mkdocs theme to display the version even after scrolling down. Now, if you're scrolled down, the only way to know what version you're looking at is the URL.

### Are you going to implement it?

Yes, because I know the probability of someone else doing it is low and I can learn from it.

### What are you going to contribute??

[...]

Issue #2688, opened 2022-07-22
[BUG] online documentation does not work completely with edge and 11.1 version


JiLleON:
### Miscellaneous first checks

- [X] I checked that all ports are open and not blocked by my ISP / hosting provider.
- [X] I know that SSL errors are likely the result of a wrong setup on the user side and not caused by DMS itself. I'm confident my setup is correct.

### Affected Component(s)

impossible to get acquainted with some examples from online documentation (FAQ -> How can I make SpamAssassin better recognize spam? ->  spoiler with example does not open); search does not work; quick navigation on the right does not work

### What happened and when does this occur?

```Markdown
at any time, you just need to visit online documentation (edge or 11.1) and try anything I have listed
```

### What did you expect to happen?

```Markdown
same work as with 11.0 and previous releases
```

### How do we replicate the issue?

```Markdown
1. visit https://docker-mailserver.github.io/docker-mailserver/edge/faq/#how-can-i-make-spamassassin-better-recognize-spam
2. try to open example in spoiler
----------
1. visit https://docker-mailserver.github.io/docker-mailserver/edge/
2. try to search smth with search
----------
1. visit FAQ or Introduction
2. try to quickly navigate through the headings on the right
```

### DMS version

edge and 11.1

### What operating system is DMS running on?

Other (unsupported)

### What instruction set architecture is DMS running on?

Other (unsupported)

### What container orchestration tool are you using?

Other (unsupported)

### What level of experience do you have with Docker and mail servers?

- [ ] I am inexperienced with docker
- [ ] I am inexperienced with mail servers
- [ ] I am uncomfortable with the CLI

### Code of conduct

- [X] I have read this project's [Code of Conduct](https://github.com/docker-mailserver/docker-mailserver/blob/master/CODE_OF_CONDUCT.md) and I agree
- [X] I have read the [README](https://github.com/docker-mailserver/docker-mailserver/blob/master/README.md) and the [documentation](https://docker-mailserver.github.io/docker-mailserver/edge/) and I searched the [issue tracker](https://github.com/docker-mailserver/docker-mailserver/issues?q=is%3Aissue) but could not find a solution

### Improvements to this form?

```Markdown
separate template for issues / requests in the documentation, which does not require mandatory checkboxes about ports and ssl errors
```

JiLleON:
I think it should be a low priority and fixed when any of you have free time for it. Also thank you for supporting this image, it made some tasks easier for me.

[...]

PR #3190, opened 2023-03-20
docs: Change `edge` version links to `latest` + fix links intended as relative not absolute

jrpear:
# Description

Change hard-coded links to point to the latest release rather than edge. Also change a few hard-coded links to be relative where I thought it made sense.

Finish addressing #3173

I don't know how the description on docker hub is maintained, maybe that needs a manual update. The github project description will certainly need a manual update.

## Type of change


jrpear:
Unrelated comment: I've been going through lots of the code, I appreciate how well commented it is and all of the links to relevant issues, I've come across a lot from @polarathene. Makes me want to work with this project more!

polarathene (changes requested):
Thanks for tackling contributing this update too! :heart: 

<details>
<summary>Original feedback (resolved)</summary>

Looks great :sunglasses: I've been a bit laxed on doc reviews these days and not enforced conventions introduced for consistently managing links.

The preference was to always have doc links use references with relative paths and prefix those references with `docs-`, usually bundling all those at the bottom of each docs page, but over time some have been co-located closer to the content which is ok too.

I've changed your `latest-docs` to `docs-latest` to stick with that convention.

---

You just need to update the workflow file so we also set the canonical link to `latest` instead of `edge` :+1: 

Whoops, just realized I made that a little bit annoying for you by committing the suggestion myself, sorry about that :grimacing: Remember to pull this branch first before committing a new change!

</details>

polarathene:
> Also change a few hard-coded links to be relative where I thought it made sense.

Fantastic, thanks for catching those! :grinning: 

> I don't know how the description on docker hub is maintained, maybe that needs a manual update.
> The github project description will certainly need a manual update.

I don't have rights to do either of those AFAIK. @wernerfred might?

Otherwise you'll need to wait on @georglauterbach I think, who is presently on vacation (_and will hopefully make a best effort to remember they're on vacation, and wait until they've returned as there is **no rush**_ :stuck_out_tongue: ).

---

> Unrelated comment: I've been going through lots of the code, I appreciate how well commented it is and all of the links to relevant issues

We certainly try! :muscle: 

It's something I tend to embrace a bit too excessively at times for the benefit of future maintainers (_and sometimes future me who forgets_ :joy: ) as there is quite a bit of niche knowledge and decisions in the project.

I have gotten a bit more lazy with formatting merge commit messages on PRs, so `git blame` should help track down the relevant PR for better details / insights :+1:

polarathene (approved):

polarathene:
Cheers, I'll wait for another approval for this change before merging.

jrpear:
Yeah, I believe you're right about the purpose of canonical links. Also the old versions of the docs will still have `edge` as their canonical versions unless we manually commit a change to those. Happy to add that to this PR or make another if that seems worthwhile.

jrpear:
> There's the shell scripts that glues most of the project together, but I can't say that's particularly enjoyable sweat_smile or researching parts of the project to troubleshoot / improve something (scripts aside). I know I've sunk a tonne of time going down rabbit holes on niche topics just to fix some config settings

I have a strange fondness for bash haha. Before I found this project, I spent probably 50 hours trying to create something which was nearly identical to dms with bash, systemd, and systemd-machined.

polarathene:
> Also the old versions of the docs will still have `edge` as their canonical versions unless we manually commit a change to those.

That's a pretty good point :sweat_smile: 

We could do that in a separate PR I think (_keep that noise separate from main focus of this PR_). (**EDIT:** As noted in 2nd section, this is probably not a PR we could review properly :disappointed: )

That reminds me of [this issue with `v11.1`](https://github.com/docker-mailserver/docker-mailserver/issues/2688). Bit of a hassle (_to build locally and adjust correctly_) presently since it needs to be built like it would be via the CI for that tagged commit, but our CI workflow isn't setup for manual trigger of the workflow with inputs to support that (one fix).

I was working on enabling manual trigger in the CI workflow but didn't get around to finishing it.

At the same time, I'm kind of tempted to archive our older versions of docs at some point as we can't realistically support them the older they get.

Also I don't think the preview deployments are ever routinely removed either and I probably should look into sorting that out :grimacing: 

---

> Happy to add that to this PR or make another if that seems worthwhile.

I'm a bit concerned about the diff size. It's probably not something I can review properly, and we can only trust ignoring large diffs from maintainers :sweat_smile: 

I could perhaps instead run a script to apply the change myself. (**EDIT: This is probably the best approach**)

Alternatively if we improve the workflow to support manual triggering of builds, it could be done that way too (_probably not a good idea as running a newer version of the docs generator will probably break old docs with deprecated syntax / features, I know that happened at some point_).

---

> I have a strange fondness for bash haha.
> Before I found this project, I spent probably 50 hours trying to create something which was nearly identical to dms with bash

Oh wow! You'd be very welcome to help with the scripts if you're comfortable :grin: 

The other two active maintainers are very good with bash, and we can all advise on the project if you have any questions.

wernerfred:
> I don't have rights to do either of those AFAIK. @wernerfred might?

Nope sorry. But that reminds me that we might need a list regarding all tools/portals we use and who is responsible/has permissions for what (github, dockerhub, netlify, ...)

wernerfred (commented):
I recall a long discussion we had when introducing the versioned docs. We left out latest (or switched to edge) on purpose. Rethinking this now imho it makes sense the way it is now as we also publish a `latest` tag together with the latest release. Only linking to specific docs site might break in future when sites gets removed or moved between sections. That is something that won't happen with strickt versioned docs. 

Looking at the `edge` docs we have a small tooltip that this is the latest documentation but not the latest stable one. Should we add a hint like this to the `latest` docs page as well stating that this is not the latest docs but the latest stable one?

Waiting for your opinion on that and if you both disagree with me i am happy to approve anyways.

Thanks @jrpear for putting this together and @polarathene i agree 100% on this "unrelated" comment:
>  I've been going through lots of the code, I appreciate how well commented it is and all of the links to relevant issues, I've come across a lot from @polarathene.

polarathene:
> Only linking to specific docs site might break in future when sites gets removed or moved between sections. That is something that won't happen with strickt versioned docs.

Hmm? Our docs are still versioned and you can link to them.

Recent contributions from @jrpear has introduced a `latest` alias to the current release (_all handled via the CI when a release is tagged_).

The plan was to default to directing users to those docs instead of `edge`, as we've received issues about users not paying attention to the version selector (_which is understandable to miss when new to our docs_), pulling `:latest` and getting confused by what the `:edge` docs say as those changes / features aren't yet released (_v12 has been a rather long cycle with some bigger changes_ :sweat_smile: )

---

I can't recall much about that earlier discussion, perhaps it related to `edge` docs reflecting the `master` branch as we don't have any separate dev branch.

I think our users are more likely to encounter the mismatch with image registries and the default `:latest` pull experience, than they are with the git repo mismatching the default docs version we link to?

polarathene:
> Looking at the `edge` docs we have a small tooltip that this is the latest documentation but not the latest stable one. Should we add a hint like this to the `latest` docs page as well stating that this is not the latest docs but the latest stable one?

AFAIK this is from refactoring that part of the docs during v12 by @georglauterbach 

I don't think you meant tooltip (popover / hover), but the admonition (coloured boxed content) we have on the main docs page?

Once we release v12, the `latest` will also inherit that. It's a bit difficult to bring back to previous docs as they build off tags, we could inject it directly into the existing HTML content on `gh-pages` branch, but the markup/CSS may differ from earlier doc versions.

---

An easier fix is to remove releases after they're a certain age, simplifying the version selector to recent versions only. The markdown would still be accessible via git history on `gh-pages`, and running them locally to access in the browser should be doable easily enough with something like `caddy`.

wernerfred:
> Hmm? Our docs are still versioned and you can link to them.

Let me give an example:

You link to `dms.docs/latest/some-site-a` which is an alias for `dms.docs/11.3/some-site-a`

Now in v12 site-a gets moved to site-b

Someone later in time using `dms.docs/latest/some-site-a` would get a `404` whereas someone using `dms.docs/11.3/some-site-a` would still be presented with the initial linked (but now outdated) docs.

This even gets more complicated if not the site name switches but huge amount of the content is changed. Refs to this docs in old Issues or Posts will link to docs that might be completely out of context content wise.

Do you get my point or am i overseeing something?

wernerfred:
> I don't think you meant tooltip (popover / hover), but the admonition (coloured boxed content) we have on the main docs page?

Yes i meant that admonition

> Once we release v12, the `latest` will also inherit that. It's a bit difficult to bring back to previous docs as they build off tags, we could inject it directly into the existing HTML content on `gh-pages` branch, but the markup/CSS may differ from earlier doc versions.

Thanks for clarifying, then i have no issue here. Just one further question. Will `latest` inherit the same box/info with `v12`? Asking because it states that `edge` is the default one but as i understand it it should be switched to `latest` then, correct?

polarathene:
> This even gets more complicated if not the site name switches but huge amount of the content is changed. Refs to this docs in old Issues or Posts will link to docs that might be completely out of context content wise.
> 
> Do you get my point or am i overseeing something?

I understand that issue, but that isn't different to defaulting to `edge`, nor is it resolved even if we did change to manually updating URLs each time to the current release tag, as users may still do the same with `edge` doc links, or `latest` if we keep that around.

`latest` will be just as prone to the issue as `edge`, and I can see how that may be more prone to happening than `edge` would, but I have often seen this done with `edge` links when users are actually discussing released versions. I personally try to use versioned links when relevant but it's often convenient to just grab the URL forgetting what version you're on (_in which case I'd end up with `edge`, and now `latest`, so nothing really changes - that's user error_).

---

> Will `latest` inherit the same box/info with `v12`?

- The version selector will have two entries for the latest release (`latest` + actual tag/version). Behind the scenes `latest` is just a symlink to the current release version docs folder.
- The admonition is not built dynamically and won't differ from `edge` docs. Presently it's hard-coded to `edge` and the description makes a mention of that, while **this PR changes the mention and URL versions to `latest`**, thus our `edge` docs will also reference `latest` in the admonition.

---

I think those were the answers you were after?

If you're still concerned about `latest` and would like a better solution, I believe I brought this concern up when we first introduced the docs.

Github Pages doesn't support redirects, but our preview service (Netlify) does. If we no longer used Github Pages (_to publish, the `gh-pages` branch could still be the source content served_), Netlify could redirect to the latest release instead, thus any `latest` URL should update to the current release version, and then any URL copied would be versioned correctly.

Only drawback is the host would change from Github to Netlify, breaking all the existing links (_unless we kept serving them with a banner about the new home_). Obviously not ideal, but that gets worse the longer we stay on Github Pages (_which we've used for approx 2 years now?_).

At the same time, I'd like to discard some older doc versions which'd have the same effect :sweat_smile: (_on the plus side, with Netlify I have a 404 page that would at least better communicate to users about content being unavailable and direct them to the newer docs, GH pages last I checked didn't support 404 pages that well_).

wernerfred (approved):
Thanks for clarifying!

In fact i would like the option you mentioned with netlify the most (from a technical and UX perspective) but regarding the dockerhub discussion i would prefer to stick with as many onboard solutions as possible. Imagine netlify cancelling their OSS plans then we are totally screwed ⚠️

georglauterbach:
Since today is [Nyepi](https://en.m.wikipedia.org/wiki/Nyepi) on Bali, and I alread spent a lot of time sleeping and being in the sun (since I am not allowed to go outside), I can warrant 10mins on GH even though I'm on vacation 😂🙂

---

The PR looks good! I'd stay with GH because it is unlikely we loose the ability to show out docs due to cancelled OSS plans.

---

@jrpear we're always looking for contributors, moderators and in a best-case scenario, maintainers.

If you are interested in DMS, I'd start by inviting you to the moderators team first. This will give you the ability to adjust issues and PRs. You'd also have access to the private discussions repo. Thereafter, I'd add you to the maintainers team :)

Just tell me whether you're interested 🙂

georglauterbach (approved):

georglauterbach:
PS: I will adjust the DockerHub README.

jrpear:
> @jrpear we're always looking for contributors, moderators and in a best-case scenario, maintainers.
> 
> If you are interested in DMS, I'd start by inviting you to the moderators team first. This will give you the ability to adjust issues and PRs. You'd also have access to the private discussions repo. Thereafter, I'd add you to the maintainers team :)

Yeah, I'd be happy to help moderate!

> PS: I will adjust the DockerHub README.

Great! Do you have permission to update the GitHub project description too? That still links to edge.

georglauterbach:
> Yeah, I'd be happy to help moderate!

You should have an invitation :)

> > PS: I will adjust the DockerHub README.
>
> Great! Do you have permission to update the GitHub project description too? That still links to edge.

Done :D

Issue #3191, opened 2023-03-21
Unit tests failing with messages like `No such container: dms-test_pop3`


jrpear:
# Subject

I would like to contribute to the project

## Description

I'm testing a change I'd like to propose, and I couldn't get the tests to pass on `master`.

I run these commands on my Debian (unstable) machine (copied from [contributing](https://docker-mailserver.github.io/docker-mailserver/edge/contributing/tests/)):

```bash
sudo apt-get -y install jq parallel

git clone https://github.com/docker-mailserver/docker-mailserver.git

cd docker-mailserver
git submodule update --init --recursive

make clean tests
```

I get lots of test failures on the form:

```
mail_pop3.bats
 ✗ setup_file failed []
   (from function `assert_success' in file test/test_helper/bats-assert/src/assert_success.bash, line 42,
    from function `_common_container_create' in file test/helper/setup.bash, line 174,
    from function `_common_container_setup' in file test/helper/setup.bash, line 195,
    from function `setup_file' in test file test/tests/serial/mail_pop3.bats, line 15)
     `_common_container_setup 'CUSTOM_SETUP_ARGUMENTS'' failed
   
   -- command failed --
   status : 1
   output (2 lines):
     Unable to find image 'mailserver-testing:ci' locally
     Error response from daemon: pull access denied for mailserver-testing, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
   --
   
   Error: No such container: dms-test_pop3
```

Any guidance?

<details>
<summary>Note to self</summary>
related: #2280
</details>

polarathene:
I think the error is a bit misleading here.

Your actual failure is due to no local image of `mailserver-testing:ci`. You get this image by building a new Docker image from the git project you just cloned, but you need to invoke `make build` to do that.

This is [mentioned on the same docs you've referenced](https://docker-mailserver.github.io/docker-mailserver/edge/contributing/tests/#executing-tests), you've just accidentally missed that part:

> You will first need to build the container image via `make build`.

---

Perhaps the docs should shift that down into the list as the first step instead for better visibility?

jrpear:
Ah yep, I did miss that. I was just reading the bullets :sweat_smile:. I think it would be slightly better if that instruction were in the list. But I should have just read better. I'll open a PR.

polarathene:
Closing as this has been resolved :+1:

PR #3193, opened 2023-03-21
docs: move `make build` instruction from paragraph into list

jrpear:
# Description

Move `make build` instruction from paragraph into list 

Clarifies confusion in #3191

## Type of change

### DIFF
--- a/docs/content/contributing/tests.md
+++ b/docs/content/contributing/tests.md
@@ -56,12 +56,13 @@ To run the test suite, you will need to:
 
 ### Executing Test(s)
 
-We use `make` to run commands. You will first need to build the container image via `make build`. You can then:
+We use `make` to run commands.
 
-1. Run all tests: `make clean tests`
-2. Run a single test: `make clean generate-accounts test/<TEST NAME WITHOUT .bats SUFFIX>`
-3. Run multiple unrelated tests: `make clean generate-accounts test/<TEST NAME WITHOUT .bats SUFFIX>,<TEST NAME WITHOUT .bats SUFFIX>` (just add a `,` and then immediately write the new test name)
-4. Run a whole set or all serial tests: `make clean generate-accounts tests/parallel/setX` where `X` is the number of the set or `make clean generate-accounts tests/serial`
+1. First, build the container image: `make build`. You can then:
+2. Run all tests: `make clean tests`
+3. Run a single test: `make clean generate-accounts test/<TEST NAME WITHOUT .bats SUFFIX>`
+4. Run multiple unrelated tests: `make clean generate-accounts test/<TEST NAME WITHOUT .bats SUFFIX>,<TEST NAME WITHOUT .bats SUFFIX>` (just add a `,` and then immediately write the new test name)
+5. Run a whole set or all serial tests: `make clean generate-accounts tests/parallel/setX` where `X` is the number of the set or `make clean generate-accounts tests/serial`
 
 ??? tip "Setting the Degree of Parallelization for Tests"
 
--- a/docs/content/contributing/tests.md
+++ b/docs/content/contributing/tests.md
@@ -58,7 +58,7 @@ To run the test suite, you will need to:
 
 We use `make` to run commands.
 
-1. First, build the container image: `make build`. You can then:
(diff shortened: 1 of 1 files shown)


jrpear:
The diff for this change is a bit ugly because I had to renumber the items in the list.

Markdown doesn't actually care what the number before the period is. These two lists render the same:

```
1. First, build the container image: `make build`. You can then:
2. Run all tests: `make clean tests`
3. Run a single test: `make clean generate-accounts test/<TEST NAME WITHOUT .bats SUFFIX>`
4. Run multiple unrelated tests: `make clean generate-accounts test/<TEST NAME WITHOUT .bats SUFFIX>,<TEST NAME WITHOUT .bats SUFFIX>` (just add a `,` and then immediately write the new test name)
5. Run a whole set or all serial tests: `make clean generate-accounts tests/parallel/setX` where `X` is the number of the set or `make clean generate-accounts tests/serial`
```

and

```
1. First, build the container image: `make build`. You can then:
1. Run all tests: `make clean tests`
1. Run a single test: `make clean generate-accounts test/<TEST NAME WITHOUT .bats SUFFIX>`
1. Run multiple unrelated tests: `make clean generate-accounts test/<TEST NAME WITHOUT .bats SUFFIX>,<TEST NAME WITHOUT .bats SUFFIX>` (just add a `,` and then immediately write the new test name)
1. Run a whole set or all serial tests: `make clean generate-accounts tests/parallel/setX` where `X` is the number of the set or `make clean generate-accounts tests/serial`
```

The latter makes for nicer diffs and editing. What do you think of changing every list in the docs to use the latter sort of list?

I suppose the downside is that the raw markdown is marginally less nice.

docs/content/contributing/tests.md
```diff
@@ -56,12 +56,13 @@ To run the test suite, you will need to:
...
 ### Executing Test(s)
 
-We use `make` to run commands. You will first need to build the container image via `make build`. You can then:
+We use `make` to run commands.
 
-1. Run all tests: `make clean tests`
-2. Run a single test: `make clean generate-accounts test/<TEST NAME WITHOUT .bats SUFFIX>`
-3. Run multiple unrelated tests: `make clean generate-accounts test/<TEST NAME WITHOUT .bats SUFFIX>,<TEST NAME WITHOUT .bats SUFFIX>` (just add a `,` and then immediately write the new test name)
-4. Run a whole set or all serial tests: `make clean generate-accounts tests/parallel/setX` where `X` is the number of the set or `make clean generate-accounts tests/serial`
+1. First, build the container image: `make build`. You can then:
```
polarathene, line 61:
```suggestion
1. Run `make build` to create or update the local `mailserver-testing:ci` Docker image (_using the projects `Dockerfile`_)
```

polarathene (changes requested):

polarathene (approved):
Thanks again! :grinning:

docs/content/contributing/tests.md
jrpear:
```
Run `make build` to create or update the local `mailserver-testing:ci` Docker image (_using the projects `Dockerfile`_)
```

doesn't say that the following commands depend on the first one. I think this might be better:

```
First, run `make build` to create or update the local `mailserver-testing:ci` Docker image (_using the projects `Dockerfile`_)
```

polarathene:
> The latter makes for nicer diffs and editing. What do you think of changing every list in the docs to use the latter sort of list?

My preference is to keep the order explicit, we rarely need to make such changes. Diff quality in Github web UI isn't great but I think some other tools do a better job (eg: Meld) usually at rendering the relevant changes.

I also know that our docs generator `mkdocs` has some GFM compatibility issues. This is also a problem with formatters like Prettier which would break this part of the docs (_I raised an issue upstream about it, but nothing has improved on that front AFAIK_).

IIRC, one of those is that indented/nested lists need to be 4 spaces wide, anything less and it's treated at the same level, and increments the items with the same logic that you bring up.

jrpear:
> My preference is to keep the order explicit, we rarely need to make such changes. Diff quality in Github web UI isn't great but I think some other tools do a better job (eg: Meld) usually at rendering the relevant changes.

Ah makes sense. :+1:

PR #3203, opened 2023-03-25
Fix: only chmod when there are files

casperklein:
# Description

Fix for:

```
[   INF   ]  Welcome to docker-mailserver 11.3.1
chmod: cannot access '/usr/lib/dovecot/sieve-filter/*': No such file or directory
chmod: cannot access '/usr/lib/dovecot/sieve-pipe/*': No such file or directory  
```

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/startup/setup.d/dovecot.sh
+++ b/target/scripts/startup/setup.d/dovecot.sh
@@ -87,8 +87,8 @@ function _setup_dovecot_sieve
   fi
 
   chown dovecot:root -R /usr/lib/dovecot/sieve-*
-  find /usr/lib/dovecot/sieve-* -type d -exec chmod 755 {} \;
-  find /usr/lib/dovecot/sieve-{filter,pipe} -type f -exec chmod +x {} \;
+  find /usr/lib/dovecot/sieve-* -type d -exec chmod 755 {} +
+  find /usr/lib/dovecot/sieve-{filter,pipe} -type f -exec chmod +x {} +
 }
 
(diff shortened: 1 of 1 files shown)


target/scripts/startup/setup.d/dovecot.sh
```diff
@@ -87,8 +87,8 @@ function _setup_dovecot_sieve
   fi
 
   chown dovecot:root -R /usr/lib/dovecot/sieve-*
-  find /usr/lib/dovecot/sieve-* -type d -exec chmod 755 {} \;
-  chmod +x /usr/lib/dovecot/sieve-{filter,pipe}/*
+  find /usr/lib/dovecot/sieve-*             -type d -exec chmod 755 {} +
```
polarathene, line 90: Curious why we set `chmod 755` and not `chmod -R 755` since this is only matching directories? If we're fixing permissions there, I assume that might be relevant to the contents too :thinking:

polarathene (approved):

target/scripts/startup/setup.d/dovecot.sh
georglauterbach: The fix appears to be correct to me, since we do not want to adjust the contents of the dirs, only the dirs, so the `-R` is not needed and not desired as well (as contents and dirs need different dirs).

If you were using `-R`, you'd be adjusting the contents as well, even though the `find` only matches dirs.

georglauterbach (approved):

Issue #3170, opened 2023-03-10
Disable spam/virus check for outbound messages


marcinkunert:
This is actually a follow-up for #1713. I've finally found some time to tinker on it. I would share my solution and ask a few questions about possible improvements.

I've enabled avamis and spamassasin in the env, so starting up both incoming and outgoing messages are checked for spam.

```
ENABLE_AMAVIS=1
ENABLE_SPAMASSASSIN=1
```

So basically every message goes through postfix. I don't quite understand why after receiving the message by `dovecot` it gets transferred to `postfix`, and not directly to the inbox. I guess it's about reusing some useful parts like queues, spamchecks etc.

The solution I've found is to route the traffic either to `amavis` or a dummy filter to send out the message.

What is important here is the `content_filter` in `/etc/postfix/main.cf`

```
content_filter = smtp-amavis:[127.0.0.1]:10024
```

So the traffic always goes to `smtp-amavis` for it to handle the check and send the message. We can override this by using a `FILTER` like this:

```
smtpd_sender_restrictions =
  check_sender_access FILTER smtp-amavis:[127.0.0.1]:10026
```

This would also use `smtp-amavis`, but on another port, which we could use to our advantage and disable some checks there, like this (config/amavis.cf):

```
$inet_socket_port = [10024, 10026];

#interface_policy{'10026'} = 'ORIGINATING';

$policy_bank{'ORIGINATING'} = {
    originating => 1, # declare that mail was submitted by our smtp client

    # don't perform spam/virus/header check.
    bypass_spam_checks_maps => [1],
    bypass_virus_checks_maps => [1],
    bypass_header_checks_maps => [1],

    # allow sending any file names and types
    bypass_banned_checks_maps => [1],

    final_spam_destiny   => D_PASS, # insure spam passes
    final_banned_destiny => D_PASS, # insure banned files pass
};
```

But this took ~400 ms per message, and I was not entirely happy with it. 

Ok, back to the topic. We need a conditional here. This can be used to switch between originating and foreign:

```
smtpd_sender_restrictions = 
  check_sender_access regexp:/etc/postfix/tag_as_originating.re
  permit_mynetworks
  permit_sasl_authenticated
  permit_tls_clientcerts
  reject_unknown_sender_domain
  reject_unknown_client_hostname
  check_sender_access regexp:/etc/postfix/tag_as_foreign.re
```

And the filters looks like this.
```
[...]

Issue #2929, opened 2022-12-11
Sending emails from within container network when behind "HaProxy"


niekvb:
# Subject

I would like some feedback concerning a use case
I have questions about TLS/SSL/STARTTLS/OpenSSL
Other

## Description

I've deployed Docker Mailserver on Kubernetes by following the recommendations in [the docs](https://docker-mailserver.github.io/docker-mailserver/edge/config/advanced/kubernetes/). This Mailserver is behind an Nginx ingress controller by using the proxy protocol. 

This all works well, until trying to send mail from within the cluster's network, which doesn't go through the ingress controller.

When I try to send mail from another pod/container on the network (through an ExternalName) it will fail for any protocol used (smtp:25, smtps:465/587) by logging the following to the Mailserver logs:

smtp:25 `postfix/postscreen[142071]: warning: haproxy read: time limit exceeded`
-- OR --
smtp:465
```bash
postfix/smtps/smtpd[147024]: warning: haproxy read: timeout error
postfix/smtps/smtpd[147024]: connect from 10-1-39-118.some-svc.some-svc.svc.cluster.local[10.1.39.118]
postfix/smtps/smtpd[147024]: disconnect from 10-1-39-118.some-svc.some-svc.svc.cluster.local[10.1.39.118] commands=0/0
```

## Question

The question here is: How would one configure the mailserver to accept mail from the cluster's network without (I suppose) trying to terminate the proxy protocol?

## Configuration

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: config
  namespace: mailserver
data:
  postfix-master.cf: |
    submission/inet/smtpd_upstream_proxy_protocol = haproxy
    smtps/inet/smtpd_upstream_proxy_protocol = haproxy

  postfix-main.cf: |
    mydestination = localhost.$mydomain, localhost
    postscreen_upstream_proxy_protocol = haproxy

  dovecot.cf: |
    # The following line should include the internal pod's network CIDR and localhost address
    haproxy_trusted_networks = 10.1.0.0/16, 127.0.0.0/8
    service imap-login {
      inet_listener imaps {
        haproxy = yes
      }
    }

  TrustedHosts: |
    127.0.0.1
    localhost
```

polarathene:
I'll reopen the issue, but I'm not sure if anyone will be able to assist :(

[...]

PR #3235, opened 2023-04-08
Postfix: rename "smtps" to "submissions"

georglauterbach:
# Description

@polarathene I am unsure whether we need to change even more. Please tell me if I forgot about something crucial.

## Type of change

- [x] Update (non-breaking change that does improve existing functionality)

## Checklist:


polarathene:
> Please tell me if I forgot about something crucial.

Nope, just replacing `smtps` for `submissions` in the project files should do it AFAIK.

We may want to add a minor test that greps 465  in `/etc/services` of the container to confirm it is assigned `submissions`. If the base image changes, that sometimes won't be the case (surprisingly with Fedora), so it'd be good to catch that instead of potential confusion troubleshooting it down to that issue :sweat_smile: 

---

I did want to get this into v12, but forgot about it. Probably should be considered a breaking change :man_shrugging: 

v13 release after v12.1 maybe?

georglauterbach:
> We may want to add a minor test that greps 465  in `/etc/services` of the container to confirm it is assigned `submissions`. If the base image changes, that sometimes won't be the case (surprisingly with Fedora), so it'd be good to catch that instead of potential confusion troubleshooting it down to that issue :sweat_smile: 

👍🏼

> I did want to get this into v12, but forgot about it. Probably should be considered a breaking change :man_shrugging: 
>
> v13 release after v12.1 maybe?

I thought this was fine for v12.1.0, because I see no reason that this is a breaking change. Why would you consider this a breaking change?

casperklein (approved):

polarathene:
> Why would you consider this a breaking change?

If someone depends on `smtps` being configured here?:

https://github.com/docker-mailserver/docker-mailserver/blob/adb38207ad35101bb01ce21621baa1c342fa205d/target/postfix/master.cf#L30-L31

Some issues that show this could happen:

- https://github.com/docker-mailserver/docker-mailserver/issues/2929 uses `postfix-master.cf` with `smtps/inet/smtpd_upstream_proxy_protocol = haproxy`
- https://github.com/docker-mailserver/docker-mailserver/issues/2601#issuecomment-1135520225 (_not directly, but semi-related advice_)
- Fail2Ban upstream config [regex may only recognize `smtps`, not `submissions`](https://github.com/fail2ban/fail2ban/blob/master/config/filter.d/postfix.conf#L13) (_I haven't looked into how relevant that is, just another example of where behaviour may change unexpectedly_)

Since we provide ways to target configuration for `smtps`/`submissions`, and it's used as the syslog name for monitoring logs, I wouldn't be surprised if some users find it a breaking change. @wernerfred usually views this sort of change as breaking as well.

---

For reference, a comment motivating this `smtps => submissions` change (_since none was provided in the PR description_): https://github.com/docker-mailserver/docker-mailserver/issues/3170#issuecomment-1465395403

casperklein:
> Some issues that show this could happen:

Good points. If there isn't a huge benefit, but possible (future) problems with stuff like fail2ban expecting smtps, I think it's not worth and we shouldn't change it.

polarathene:
> I think it's not worth and we shouldn't change it.

It's fine for introducing as a breaking change. Users can raise actual bug reports to surface the problems.

Like I said, I am not sure if fail2ban monitoring `smtps` was relevant, just an example that came to mind. I looked through `git blame` and realized that the regex still covers `submissions`, even though it was added back in 2016. So no issue there with this change.

Users may still have their own config / scripts or other tools that are more explicitly expecting `smtps`, which is my main point of respecting semver by only applying the change for a major release.

georglauterbach:
I had a glance at our F2B config as well, and it is using `submissions`, not `smtps`, as @polarathene said.

I am fine with waiting for the next major with this. It will probably not even take too long for v13.0.0 to come out, due to other changes I have in mind.

georglauterbach:
**INFO**: Marked this with `meta/feature-freeze` now to indicate this PR is not to be merged before v12.1.0 is released.

PR #3246, opened 2023-04-10
scripts: make `policyd-spf` configurable

georglauterbach:
# Description

Users of Rspamd do not need `policyd-spf`, hence it should be configurable. With `v13.0.0`, the default should change to `0` because `ENABLE_RSPAMD` will default to `1`.

Fixes #

## Type of change

## Checklist:


casperklein (approved):

polarathene:
> With `v13.0.0`, the default should change to `0` because `ENABLE_RSPAMD` will default to `1`.

If you are defaulting to Rspamd that soon, please get it into a state that it's more standardized for configuration like we support with other features already.

Last I saw the docs were being very vague on local storage and saying to mount content directly into `/etc` instead of leveraging our common `/tmp/docker-mailserver` config volume and copying from there like other features do (_and which supports the change-detection service to automatically update / reload services from config changes_).

There is still advice about manual config management (_and placement, including no strict file name, such as dkim signing module_). Along with a DSL of sorts config file you've designed, which I've not looked too much into.

I get the impression of the support being rushed / pushed a bit eagerly, with potential for burdening yourself with more maintenance as rspamd is largely your area of the project.

target/scripts/startup/setup.d/dmarc_dkim_spf.sh
```diff
@@ -84,3 +84,21 @@ function _setup_opendmarc
...
+    _log 'debug' 'Configuring policyd-spf'
+    cat >>/etc/postfix/master.cf <<EOF
+
+policyd-spf    unix  -       n       n       -       0       spawn
+    user=policyd-spf argv=/usr/bin/policyd-spf
+EOF
+  else
+    _log 'debug' 'Disabling policyd-spf'
+    sedfile -i -E 's|check_policy_service unix:private/policyd-spf, ||g' /etc/postfix/main.cf
+  fi
```
polarathene, line 103: Wouldn't it make sense to include this in the enable logic above?

```
# From (but only for smtpd_recipient_restrictions):
# https://github.com/docker-mailserver/docker-mailserver/blob/bbe1d2da3117264c484a87c75813b0d3da71161a/target/postfix/main.cf#L51
reject_unauth_destination, reject_unauth_pipelining

# To (sed value inbetween):
reject_unauth_destination, check_policy_service unix:private/policyd-spf, reject_unauth_pipelining
```

---

Bit more work to support, I'm not going to block since this snippet will be removed before the end of the year I assume.

polarathene (approved):
I'm not sure about the env/docs to communicate rspamd relevance, but will approve on the basis it'll no longer be relevant 6 months later.

docs/content/config/environment.md
```diff
@@ -110,6 +110,13 @@ Enables the OpenDMARC service.
 - **1** => Enabled
 - 0 => Disabled
 
+##### ENABLE_POLICYD_SPF
+
+Enabled `policyd-spf` in Postfix's configuration. You will likely want to set this to `0` in case you're using Rspamd ([`ENABLE_RSPAMD=1`](#enable_rspamd)).
```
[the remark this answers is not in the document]
georglauterbach: Fair point; I will be adding these checks soon. I think the Rspamd context does not hurt at all, and the only relevant setup where this setting really should be used as of now is `ENABLE_RSPAMD=1`.

target/scripts/startup/setup.d/dmarc_dkim_spf.sh
georglauterbach: I actually thought the same, but then I think it is easier with my initial way. You're right though: this should be removed by the end of the year :)

PR #3248, opened 2023-04-11
Rspamd: replace `reject_unknown_client_hostname` with Rspamd `HFILTER_HOSTNAME_UNKNOWN` and make it configurable

georglauterbach:
# Description

Please quickly read over #3228 :) Removes `reject_unknown_client_hostname` and (re-)add the test to Rspamd (via `HFILTER_HOSTNAME_UNKNOWN`).

Fixes #3228
Fixes #2970

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] Improvement (non-breaking change that does improve existing functionality)
- [x] This change requires a documentation update

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [x] If necessary I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #3255, opened 2023-04-12
Posfix: add option to re-enable `reject_unknown_client_hostname` after #3248

georglauterbach:
# Description

See <https://github.com/docker-mailserver/docker-mailserver/pull/3248#issuecomment-1504107878>.

Fixes concern raised in #3248

## Type of change

## Checklist:


georglauterbach:
@casperklein please review the first commit separately from the second one. I saw the setup file for Postfix and just could not help myself but end the trauma we have with these 1000 separate setup functions. Functionality-wise, nothing changes - if you will, this is just style.

Reviewing each commit on its own makes the diff less fuzzy.

casperklein:
LGTM 👍 

---

Serious question: In the past, there was lot of code refactoring where code like:

```bash
# do one thing
foo bar

# do another thing
baz bla
```

was converted to

```bash
function one_thing() {
	foo bar
}

function another_thing() {
	baz bla
}

one_thing
another_thing
```

We had a discussion about this, wether it's worth to create functions that are only called once or not. We sticked with functions. The first commit in this PR now reverts this more or less.
Previously I was team "no-functions". But after introducing them, I started to like them 😆
As you wrote, it's only styling. I am fine with both variant. But maybe consider this: we should follow and stick to one style, instead of changing in more or less short intervals.

---

`_setup_spoof_protection` could also be part of `_register_setup_function '_setup_postfix_late'`?

georglauterbach:
> Serious question: In the past, there was lot of code refactoring where code like:
> 
> ```shell
> # do one thing
> foo bar
> 
> # do another thing
> baz bla
> ```
> 
> was converted to
> 
> ```shell
> function one_thing() {
> 	foo bar
> }
> 
> function another_thing() {
> 	baz bla
> }
> 
> one_thing
> another_thing
> ```
> 
> We had a discussion about this, wether it's worth to create functions that are only called once or not. We sticked with functions. The first commit in this PR now reverts this more or less. Previously I was team "no-functions". But after introducing them, I started to like them 😆 As you wrote, it's only styling. I am fine with both variant. But maybe consider this: we should follow and stick to one style, instead of changing in more or less short intervals.

I actually do not remember 😆 But for me, the important part is _that there is at least one function commands are executed in_, and that we do not just execute commands in non-function-contexts. So this

```bash
# do one thing
foo bar

# do another thing
baz bla
```

should be this:

```bash
function one_thing() {
  # do one thing
  foo bar

  # do another thing
  baz bla
}

one_thing
```

I am not a big fan of one function executing one command, although that may be fine in single places. The Postfix setup did it a bit too much, hence my refactoring.

> `_setup_spoof_protection` could also be part of `_register_setup_function '_setup_postfix_late'`?

Probably 🤔 But I'd like to defer this, to move `_setup_spoof_protection` into the Postfix setup file.

casperklein (approved):

target/scripts/startup/setup.d/postfix.sh
```diff
@@ -46,40 +42,65 @@ EOF
...
-  [[ -z ${POSTFIX_DAGENT} ]] && return 0
+  _log 'debug' 'Configuring Postfix (late setup)'
 
-  _log 'trace' "Changing Postfix virtual transport to '${POSTFIX_DAGENT}'"
-  # Default value in main.cf should be 'lmtp:unix:/var/run/dovecot/lmtp'
-  postconf "virtual_transport = ${POSTFIX_DAGENT}"
+  __postfix__log 'trace' 'Configuring user access'
+  if [[ -f /tmp/docker-mailserver/postfix-send-access.cf ]]
+  then
+    sed -i 's|(smtpd_sender_restrictions =)|\1 check_sender_access texthash:/tmp/docker-mailserver/postfix-send-access.cf,|' /etc/postfix/main.cf
```
ap-wtioit, line 80: missing -E (pull request #3272)
```bash
sed -i -E 's|(smtpd_sender_restrictions =)|\1 check_sender_access texthash:/tmp/docker-mailserver/postfix-send-access.cf,|' /etc/postfix/main.cf
```

PR #3261, opened 2023-04-14
Rspamd: script stabilization pt. 1

georglauterbach:
# Description

1. added checks whether OpenDKIM/OpenDMARC/policyd-spf/Postgrey are enabled
2. added functions to check if VAR is 0/1 or an int
3. added more checks to Rspamd setup (mainly functions introduced in 2.)
4. added canonical directory for users to place files in

Related to #3249
Closes #3266

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] Improvement (non-breaking change that does improve existing functionality)
- [x] This change requires a documentation update

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [x] If necessary I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #3263, opened 2023-04-14
scripts: get all `policyd-spf` setup in one place

georglauterbach:
# Description

Basically <https://github.com/docker-mailserver/docker-mailserver/pull/3246#discussion_r1162156066>.

Looking at <https://github.com/docker-mailserver/docker-mailserver/actions/runs/4699505235/jobs/8333061595?pr=3261>, we can see that `postconf` is complaining:

```txt
postconf: warning: /etc/postfix/main.cf: unused parameter: policyd-spf_time_limit=3600
```

This PR resolves the matter and puts all the code that integrates policyd-spf in one place.

---

The log says `'Disabling policyd-spf'`, when actually doing nothing, but this is okay, since

1. the default value is `1`
2. we will change the default after v12.1.0 to 0 anyway and therefore, the log messages

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/startup/setup.d/dmarc_dkim_spf.sh
+++ b/target/scripts/startup/setup.d/dmarc_dkim_spf.sh
@@ -101,10 +101,8 @@ EOF
   sedfile -i -E \
     's|^(smtpd_recipient_restrictions.*reject_unauth_destination)(.*)|\1, check_policy_service unix:private/policyd-spf\2|' \
     /etc/postfix/main.cf
-  cat >>/etc/postfix/main.cf <<EOF
-
-# SPF policy settings
-policyd-spf_time_limit = 3600
+  # SPF policy settings
+  postconf 'policyd-spf_time_limit = 3600'
 EOF
   else
     _log 'debug' 'Disabling policyd-spf'
--- a/target/scripts/startup/setup.d/dmarc_dkim_spf.sh
+++ b/target/scripts/startup/setup.d/dmarc_dkim_spf.sh
@@ -97,8 +97,16 @@ function _setup_policyd_spf
 policyd-spf    unix  -       n       n       -       0       spawn
     user=policyd-spf argv=/usr/bin/policyd-spf
 EOF
+
+  sedfile -i -E \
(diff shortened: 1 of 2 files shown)


target/scripts/startup/setup.d/dmarc_dkim_spf.sh
```diff
@@ -96,9 +96,17 @@ function _setup_policyd_spf
...
     user=policyd-spf argv=/usr/bin/policyd-spf
+EOF
+
+  sedfile -i -E \
+    's|^(smtpd_recipient_restrictions.*reject_unauth_destination)(.*)|\1, check_policy_service unix:private/policyd-spf\2|' \
+    /etc/postfix/main.cf
+  cat >>/etc/postfix/main.cf <<EOF
+
+# SPF policy settings
+policyd-spf_time_limit = 3600
```
polarathene, line 107: Comment doesn't really need to be embedded into the `main.cf` for single setting. This suggestion would need to remove the `EOF` directly after it though.

```suggestion
  # SPF policy settings
  postconf 'policyd-spf_time_limit = 3600'
```

or

```suggestion

  cat >>/etc/postfix/main.cf <<EOF
# SPF policy settings
policyd-spf_time_limit = 3600
```

polarathene (approved):

target/scripts/startup/setup.d/dmarc_dkim_spf.sh
```diff
@@ -97,8 +97,14 @@ function _setup_policyd_spf
 policyd-spf    unix  -       n       n       -       0       spawn
     user=policyd-spf argv=/usr/bin/policyd-spf
 EOF
+
+  sedfile -i -E \
+    's|^(smtpd_recipient_restrictions.*reject_unauth_destination)(.*)|\1, check_policy_service unix:private/policyd-spf\2|' \
+    /etc/postfix/main.cf
+  # SPF policy settings
+  postconf 'policyd-spf_time_limit = 3600'
+EOF
```
polarathene, line 106: EOF was advised to be dropped for this suggestion.

georglauterbach: I should have read this more carefully, sorry 😨

Fixed in f62855f89dc61e272dd87791f518b8813a6f2b35 in PR #3261

PR #3265, opened 2023-04-16
scripts: improve shutdown function by making PANIC_STRATEGY obsolete

casperklein:
# Description

> In case the user requested an immediate exit, he ensure he is not in a subshell

This PR makes the decision obsolete, if the function is called from a subshell or not.

Inspired from: https://cravencode.com/post/essentials/exit-shell-script-from-subshell/

History:

- https://github.com/docker-mailserver/docker-mailserver/pull/3155
- https://github.com/docker-mailserver/docker-mailserver/issues/3140
- https://github.com/docker-mailserver/docker-mailserver/issues/3133

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/helpers/error.sh
+++ b/target/scripts/helpers/error.sh
@@ -25,7 +25,6 @@ function dms_panic
   local PANIC_TYPE=${1:-}
   local PANIC_INFO=${2:-}
   local PANIC_SCOPE=${3-} # optional, must not be :- but just -
-  local PANIC_STRATEGY=${4:-} # optional
 
   local SHUTDOWN_MESSAGE
 
@@ -61,9 +60,9 @@ function dms_panic
 
   if [[ -n ${PANIC_SCOPE:-} ]]
   then
-    _shutdown "${PANIC_SCOPE} | ${SHUTDOWN_MESSAGE}" "${PANIC_STRATEGY}"
+    _shutdown "${PANIC_SCOPE} | ${SHUTDOWN_MESSAGE}"
   else
-    _shutdown "${SHUTDOWN_MESSAGE}" "${PANIC_STRATEGY}"
+    _shutdown "${SHUTDOWN_MESSAGE}"
   fi
 }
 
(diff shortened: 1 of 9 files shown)


target/scripts/helpers/error.sh
```diff
@@ -77,22 +76,14 @@ function _dms_panic__general       { dms_panic 'general'       "${1:-}" "${2:-}"
 
 # Call this method when you want to panic (i.e. emit an 'ERROR' log, and exit uncleanly).
 # `dms_panic` methods should be preferred if your failure type is supported.
+trap "exit 1" SIGUSR1
+SCRIPT_PID="$$"
```
georglauterbach, line 80: To keep up with our conventions, this should be

```suggestion
SCRIPT_PID=${$}
```

which I agree looks interesting 😆

georglauterbach (commented):
LGTM 👍🏼

Just one minor nitpick :)

target/scripts/helpers/error.sh
```diff
@@ -25,7 +25,6 @@ function dms_panic
   local PANIC_TYPE=${1:-}
   local PANIC_INFO=${2:-}
   local PANIC_SCOPE=${3-} # optional, must not be :- but just -
```
casperklein, line 27: @georglauterbach 

Can you explain why `${3-}` must be used? `${3:-}` should do the same here.

georglauterbach: With your addition, both are actually fine. `${3-}` was required before because the (non-)nullity check that Bash does when using `:-` (instead of `-`) would lead to the situation in which `''` is treated as `null`, so `'' 'second_argument'` would become `'second_argument'`, which is not what we wanted.

Because now there is no 4th argument, we can make it `:-` again.

georglauterbach:
```suggestion
  local PANIC_SCOPE=${3:-}
```

georglauterbach (approved):
LGTM 👍🏼

A very fine addition, and something I did not think of - I learned someting new about Bash, and I'm very grateful for that ❤️

polarathene (approved):
Nice work! :grin:

target/scripts/helpers/error.sh
georglauterbach: Sorry, I was looking at it from the wrong side:

```console
$ function foo() { local ONE=${1:-lol} ; local TWO=${2:-} ; echo "|$ONE|$TWO|" ; }
$ foo '' 2
|lol|2|
$ function foo() { local ONE=${1-lol} ; local TWO=${2:-} ; echo "|$ONE|$TWO|" ; }
$ foo '' 2
||2|
```

Seems like that for our example, it really made no difference to begin with as we're assigning the empty string anyways.

PR #3259, opened 2023-04-14
issue with existing alias which doesn't exist

bibiak1:
# Description

Issue with adding new recipient to existing alias
If new recipient exists as alias in db then setup is throwing error.

Fixes part of script checks only the relevant alias and compares recipients for it.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Improvement (non-breaking change that does improve existing functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [x] If necessary I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #3270, opened 2023-04-17
scripts: disallow alias = account

### LINKED ISSUE #3262: bug report: alias code for group mail aliases
### Preliminary Checks

- [X] I searched the issue tracker but was unable to find my issue.
- [X] I read the [extended documentation in general](https://docker-mailserver.github.io/docker-mailserver/latest/) but found nothing to resolve the issue.
- [X] I read the [documentation on debugging](https://docker-mailserver.github.io/docker-mailserver/latest/config/debugging/), tried the proposed steps to debug the problem, but was still unable to resolve the issue.

### Affected Component(s)

Aliases setup

When I tried to create group mail aliases I came into problem with adding recipient to the group.
Problem appears if there's an email account and alias to redirect mails.

```console
root@mail:/# setup email list
* testuser@domain.com ( 0 / ~ ) [0%]

root@mail:/# setup alias list
[  ERROR  ]  '/tmp/docker-mailserver/postfix-virtual.cf' is empty, nothing to list
[  ERROR  ]  Aborting
root@mail:/# setup email add test1@domain.com
Enter Password: 
root@mail:/# setup email add test2@domain.com
Enter Password: 
root@mail:/# setup alias add test1@domain.com test1@other.com
root@mail:/# setup alias add test2@domain.com test2@other.com
root@mail:/# setup alias add group-1@domain.com test1@domain.com
root@mail:/# setup alias add group-1@domain.com test2@domain.com
root@mail:/# setup alias add group-2@domain.com test1@domain.com
root@mail:/# setup alias add group-2@domain.com test2@domain.com
[  ERROR  ]  'group-2@domain.com' is already an alias for recipient: 'test2@domain.com'
[  ERROR  ]  Aborting
root@mail:/# 
```

1. setup email add test1@domain.com
2. setup email add test2@domain.com
3. setup alias add test1@domain.com test1@other.com
4. setup alias add test2@domain.com test2@other.com
5. setup alias add group-1@domain.com test1@domain.com
6. setup alias add group-1@domain.com test2@domain.com
7. setup alias add group-2@domain.com test1@domain.com
8. setup alias add group-2@domain.com test2@domain.com

### DMS version

v12.0.0

Linux

Docker

AMD64 / x86_64

Docker Compose

### docker-compose.yml

```yml
services:
  mailserver:
    container_name: mailserver
    hostname: mail.example.com
    volumes:
    environment:
      - ENABLE_RSPAMD=1
      - ENABLE_CLAMAV=1
      - ENABLE_FAIL2BAN=1
    cap_add:
      - NET_ADMIN # For Fail2Ban to work
```

### Relevant log output

_No response_

Solution: https://github.com/docker-mailserver/docker-mailserver/pull/3259

### Code of conduct

_No response_

georglauterbach:
# Description

We do not support aliases that are accounts at the same time. This commit provides a guard that checks the condition.

@bibiak1 made us aware of the fact that we allow creating aliases as users (and vice versa). I think we do not support this, and at least to me, it does not make a lot of sense.

Fixes #3262
Closes #3259

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/helpers/database/manage/postfix-accounts.sh
+++ b/target/scripts/helpers/database/manage/postfix-accounts.sh
@@ -76,7 +76,7 @@ function _account_should_not_exist_yet
   __account_already_exists && _exit_with_error "'${MAIL_ACCOUNT}' already exists"
   if [[ -f ${DATABASE_VIRTUAL} ]] && grep -q "^${MAIL_ACCOUNT}" "${DATABASE_VIRTUAL}"
   then
-    _exit_with_error "'${MAIL_ACCOUNT}' is already defined to be an alias"
+    _exit_with_error "'${MAIL_ACCOUNT}' is already defined as an alias"
   fi
 }
 
--- a/target/scripts/helpers/database/manage/postfix-accounts.sh
+++ b/target/scripts/helpers/database/manage/postfix-accounts.sh
@@ -74,6 +74,10 @@ function _arg_expect_mail_account
 function _account_should_not_exist_yet
 {
   __account_already_exists && _exit_with_error "'${MAIL_ACCOUNT}' already exists"
+  if grep -q "^${MAIL_ACCOUNT}" "${DATABASE_VIRTUAL}"
+  then
+    _exit_with_error "'${MAIL_ACCOUNT}' is already defined to be an alias"
+  fi
 }
 
 # Also used by delmailuser, setquota, delquota
--- a/target/scripts/helpers/database/manage/postfix-accounts.sh
+++ b/target/scripts/helpers/database/manage/postfix-accounts.sh
@@ -74,7 +74,7 @@ function _arg_expect_mail_account
 function _account_should_not_exist_yet
 {
   __account_already_exists && _exit_with_error "'${MAIL_ACCOUNT}' already exists"
-  if grep -q "^${MAIL_ACCOUNT}" "${DATABASE_VIRTUAL}"
+  if [[ -f ${DATABASE_VIRTUAL} ]] && grep -q "^${MAIL_ACCOUNT}" "${DATABASE_VIRTUAL}"
   then
     _exit_with_error "'${MAIL_ACCOUNT}' is already defined to be an alias"
   fi
--- a/target/scripts/helpers/database/manage/postfix-virtual.sh
+++ b/target/scripts/helpers/database/manage/postfix-virtual.sh
@@ -27,7 +27,7 @@ function _manage_virtual_aliases
     ( 'update' )
       if [[ -f ${DATABASE_ACCOUNTS} ]] && grep -q "^${MAIL_ALIAS}" "${DATABASE_ACCOUNTS}"
       then
-        _exit_with_error "'${MAIL_ALIAS}' is already defined to be an account"
+        _exit_with_error "'${MAIL_ALIAS}' is already defined as an account"
       fi
       _db_entry_add_or_append "${DATABASE_VIRTUAL}" "${MAIL_ALIAS}" "${RECIPIENT}"
       ;;
--- a/target/scripts/helpers/database/manage/postfix-virtual.sh
+++ b/target/scripts/helpers/database/manage/postfix-virtual.sh
@@ -25,7 +25,7 @@ function _manage_virtual_aliases
   case "${ACTION}" in
     # Associate RECIPIENT to MAIL_ALIAS:
     ( 'update' )
-      if grep -q "^${MAIL_ALIAS}" "${DATABASE_ACCOUNTS}"
+      if [[ -f ${DATABASE_ACCOUNTS} ]] && grep -q "^${MAIL_ALIAS}" "${DATABASE_ACCOUNTS}"
       then
         _exit_with_error "'${MAIL_ALIAS}' is already defined to be an account"
       fi
--- a/target/scripts/helpers/database/manage/postfix-virtual.sh
+++ b/target/scripts/helpers/database/manage/postfix-virtual.sh
@@ -25,6 +25,10 @@ function _manage_virtual_aliases
   case "${ACTION}" in
     # Associate RECIPIENT to MAIL_ALIAS:
     ( 'update' )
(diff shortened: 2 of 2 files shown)


georglauterbach:
@polarathene I am not 100% certain whether we should disallow both directions, but to me that makes sense.

target/scripts/helpers/database/manage/postfix-accounts.sh
```diff
@@ -74,6 +74,10 @@ function _arg_expect_mail_account
 function _account_should_not_exist_yet
 {
   __account_already_exists && _exit_with_error "'${MAIL_ACCOUNT}' already exists"
+  if [[ -f ${DATABASE_VIRTUAL} ]] && grep -q "^${MAIL_ACCOUNT}" "${DATABASE_VIRTUAL}"
+  then
+    _exit_with_error "'${MAIL_ACCOUNT}' is already defined to be an alias"
```
polarathene, line 79:
```suggestion
    _exit_with_error "'${MAIL_ACCOUNT}' is already defined as an alias"
```

target/scripts/helpers/database/manage/postfix-virtual.sh
```diff
@@ -25,6 +25,10 @@ function _manage_virtual_aliases
   case "${ACTION}" in
     # Associate RECIPIENT to MAIL_ALIAS:
     ( 'update' )
+      if [[ -f ${DATABASE_ACCOUNTS} ]] && grep -q "^${MAIL_ALIAS}" "${DATABASE_ACCOUNTS}"
+      then
+        _exit_with_error "'${MAIL_ALIAS}' is already defined to be an account"
```
polarathene, line 30:
```suggestion
        _exit_with_error "'${MAIL_ALIAS}' is already defined as an account"
```

polarathene (changes requested):

polarathene (approved):
LGTM! :grinning: 

Preventing either DB adding a key that belongs to the other makes sense to me :+1:

target/scripts/helpers/database/manage/postfix-accounts.sh
```diff
@@ -74,6 +74,10 @@ function _arg_expect_mail_account
 function _account_should_not_exist_yet
 {
   __account_already_exists && _exit_with_error "'${MAIL_ACCOUNT}' already exists"
+  if grep -q "^${MAIL_ACCOUNT}" "${DATABASE_VIRTUAL}"
```
casperklein: I remember, that we had a similar issue in the past:

`postfix-virtual.cf` looks like:

```
myalias@example.com user@example.com
```

The grep statement above would also match on:

myalias@example.company.com

Adding a space for example should fix it:

`if grep -q "^${MAIL_ACCOUNT} " "${DATABASE_VIRTUAL}"`

polarathene: That should be handled in both lookups then? But for accounts table, instead of space a `|`?

casperklein: You are right.

KevinBrogan:
If you don't support it, then emit a warning. Call it an advanced option. something other than a pointless block.
The server software supports it without issue.
Now I'm just going to go delete this line from the container I'm hosting.

polarathene:
> If you don't support it, then emit a warning. Call it an advanced option. something other than a pointless block.

I'm fine with a change to allow it with a warning emitted, or a prompt for confirmation before continuing.

You're welcome to open a PR if the other maintainers are comfortable with the change. It should include test coverage that avoids [errors like this one](https://github.com/docker-mailserver/docker-mailserver/issues/3262).

FWIW we do make this concern [more prominent in our docs from DMS v15 onwards](https://docker-mailserver.github.io/docker-mailserver/edge/config/account-management/provisioner/file/#aliases):

> image

---

> The server software supports it without issue.

Yes but DMS is another layer on top of that.

Virtual aliases by default will be used to create dummy accounts in Dovecot. A side-effect is you can login with that (_and some users do this_), even though that is unintentional and not supported, it's the only option we have for Dovecot Quota support to work without writing and maintaining our own policy check service for Postfix.

Could you please add some context to why you need an alias to your mailbox account with the same address? We enforced this block due to various bug reports caused by misconfigurations (_[more context here](https://github.com/docker-mailserver/docker-mailserver/issues/3350#issuecomment-1550528898)_).

It's also possible you are dealing with an XY problem where you are dismissing our decision to forbid aliases sharing mailbox addresses to solve a problem that would be [better supported by other features from Postfix](https://github.com/docker-mailserver/docker-mailserver/pull/3259#issuecomment-1513294364). If so your concern is more for the lack of official support via DMS for such, but even if we had that in place you may have only been familiar with aliases and not considered it?

The valid use-cases that I'm aware of are:
- Wildcard catch-all alias has precedence as aliases are resolved before mailbox addresses. The workaround for that is to have an alias with the mailbox address which will have precedence over the wildcard, be a no-op and send to the same address as the alias, handing over to Dovecot for local delivery.
- Forwarding mail to an external mail server, but still retaining a copy locally. DMS doesn't officially support multiple recipients for an alias however (_the Dovecot Quotas workaround isn't compatible for example_).

Invalid use-cases tend to be forwarding to an external mail server, often `@gmail.com`.

georglauterbach:
> If you don't support it, then emit a warning. Call it an advanced option. something other than a pointless block. The server software supports it without issue. Now I'm just going to go delete this line from the container I'm hosting.

Thank you for chiming into this closed issue. I also appreciate the kindness with which this ~~command~~ comment has been put forward. As a consequence, I'm eagerly awaiting your well-tested contribution to this FOSS project, which is maintained by individuals in their spare time.

---

@polarathene was too kind when he gave a friendly answer here. But I have zero tolerance when it comes to impoliteness. Formulate what you want in a direct, precise, and friendly manner. You are not entitled to demand changes to this project, you can ask for them nicely. At the risk of repeating myself: This project is free, open-source, and maintained by individuals in their spare time. This happens again, and you'll be posting nowhere on this whole project, @KevinBrogan. I've made myself clear here, no discussion necessary anymore.

PR #3271, opened 2023-04-18
scripts: fix setting `SRS_EXCLUDE_DOMAINS` during startup

jamebus:
# Description

``mailserver  | sed: -e expression #1, char 302: invalid reference \1 on `s' command's RHS``

The sed command has a syntax error preventing a reference from being created. There's also an errant "=" that would output.

This fixes the issue and allows SRS_EXCLUDE_DOMAINS to be set in the postsrsd configuration.

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/startup/setup.d/postfix.sh
+++ b/target/scripts/startup/setup.d/postfix.sh
@@ -170,8 +170,8 @@ function _setup_SRS
 
   if [[ -n ${SRS_EXCLUDE_DOMAINS} ]]
   then
-    sed -i \
-      "s/^#\?(SRS_EXCLUDE_DOMAINS=).*$/\1=${SRS_EXCLUDE_DOMAINS}/g" \
+    sedfile -i -E \
+      "s|^#?(SRS_EXCLUDE_DOMAINS=).*|\1${SRS_EXCLUDE_DOMAINS}|" \
       /etc/default/postsrsd
   fi
 }


target/scripts/startup/setup.d/postfix.sh
```diff
@@ -171,7 +171,7 @@ function _setup_SRS
   if [[ -n ${SRS_EXCLUDE_DOMAINS} ]]
   then
     sed -i \
-      "s/^#\?(SRS_EXCLUDE_DOMAINS=).*$/\1=${SRS_EXCLUDE_DOMAINS}/g" \
+      "s/^#\?\(SRS_EXCLUDE_DOMAINS=\).*$/\1${SRS_EXCLUDE_DOMAINS}/g" \
```
georglauterbach, line 174:
```suggestion
    sedfile -i -E \
      "s|^#?(SRS_EXCLUDE_DOMAINS=).*|\1${SRS_EXCLUDE_DOMAINS}|" \
```

georglauterbach (changes requested):
Thank you very much for finding this bug - very much appreciated! ❤️

The solution you propose does seem to work, but the regular expression that `sed` was working with is strange to begin with. I have therefore proposed a cleaner and simpler solution that is coherent with the rest of the code.

target/scripts/startup/setup.d/postfix.sh
jamebus: Thanks! Yes, I like this better. Extended re's are also way easier to read.

georglauterbach (approved):

PR #3272, opened 2023-04-18
postfix.sh: add missing -E for extended regexes in smtpd_sender_restrictions

ap-wtioit:
# Description

Noticed that something was wrong when in our tests i got the following additional output:
```
sed: -e expression #1, char 111: invalid reference \1 on `s' command's RHS
```

## Type of change

## Checklist:

Info @wt-io-it

### DIFF
--- a/target/scripts/startup/setup.d/postfix.sh
+++ b/target/scripts/startup/setup.d/postfix.sh
@@ -77,7 +77,7 @@ function _setup_postfix_late
   __postfix__log 'trace' 'Configuring user access'
   if [[ -f /tmp/docker-mailserver/postfix-send-access.cf ]]
   then
-    sed -i 's|(smtpd_sender_restrictions =)|\1 check_sender_access texthash:/tmp/docker-mailserver/postfix-send-access.cf,|' /etc/postfix/main.cf
+    sed -i -E 's|(smtpd_sender_restrictions =)|\1 check_sender_access texthash:/tmp/docker-mailserver/postfix-send-access.cf,|' /etc/postfix/main.cf
   fi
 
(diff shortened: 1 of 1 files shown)


georglauterbach (approved):
Well, `sed` is one crude tool.. I thought the `-E` is only necessary when using more sophisticated regular expressions. Oh my..

ap-wtioit:
Well if you want to use unescaped brackets you have to use -E:
```bash
echo "abc" | sed 's/\(b\)/d/'
echo "abc" | sed -E 's/(b)/d/'
```
I think both grep and sed use [POSIX basic regular expressions](https://www.gnu.org/software/findutils/manual/html_node/find_html/posix_002dbasic-regular-expression-syntax.html) as default and we are not used to those any more.

georglauterbach:
That explains the weirdness of the "old code" in #3271 - thanks for the info :)

target/scripts/startup/setup.d/postfix.sh
```diff
@@ -77,7 +77,7 @@ function _setup_postfix_late
   __postfix__log 'trace' 'Configuring user access'
   if [[ -f /tmp/docker-mailserver/postfix-send-access.cf ]]
   then
-    sed -i 's|(smtpd_sender_restrictions =)|\1 check_sender_access texthash:/tmp/docker-mailserver/postfix-send-access.cf,|' /etc/postfix/main.cf
+    sed -i -E 's|(smtpd_sender_restrictions =)|\1 check_sender_access texthash:/tmp/docker-mailserver/postfix-send-access.cf,|' /etc/postfix/main.cf
```
casperklein, line 80: `sed` --> `sedfile` to get notified in the future, if no change takes place.

PR #3285, opened 2023-04-24
scripts: apply fixes to helpers when using `set -eE`

georglauterbach:
# Description

For an upcoming PR, these changes are required, because the script that is using the helpers uses `set -eE`. This leads to situations where errors are not properly handled in our helpers (yet; I plan on changing that in the future).

Fixes issues of an upcoming PR that I do not want to include in the upcoming PR because I am a nice maintainer and I try my best to separate concerns with PRs :D 😆

The `return 0` statements at the end of some functions seem superflous, but they are not, because the previos command (even when run inside an `if`-clause) set the return value to non-zero, which is not what we want.

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/helpers/dns.sh
+++ b/target/scripts/helpers/dns.sh
@@ -41,7 +41,7 @@ function _obtain_hostname_and_domainname
   # of this variable or if a more deterministic approach with `cut` should be relied on.
   if [[ $(_get_label_count "${HOSTNAME}") -gt 2 ]]
   then
-    if [[ -n ${OVERRIDE_HOSTNAME} ]]
+    if [[ -n ${OVERRIDE_HOSTNAME:-} ]]
     then
       # Emulates the intended behaviour of `hostname -d`:
       # Assign the HOSTNAME value minus everything up to and including the first `.`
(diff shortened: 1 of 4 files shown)


casperklein:
> The `return 0` statements at the end of some functions seem superflous, but they are not, because the previos command (even when run inside an `if`-clause) set the return value to non-zero, which is not what we want.

```bash
function _adjust_mtime_for_postfix_maincf
{
  if [[ $(( $(date '+%s') - $(stat -c '%Y' '/etc/postfix/main.cf') )) -lt 2 ]]
  then
    touch -d '2 seconds ago' /etc/postfix/main.cf
  fi
  return 0
}
```

If `touch` fails for some reason, why would we want to return success (0)?

georglauterbach:
> > The `return 0` statements at the end of some functions seem superflous, but they are not, because the previos command (even when run inside an `if`-clause) set the return value to non-zero, which is not what we want.
> 
> 
> 
> ```bash
> function _adjust_mtime_for_postfix_maincf
> {
>   if [[ $(( $(date '+%s') - $(stat -c '%Y' '/etc/postfix/main.cf') )) -lt 2 ]]
>   then
>     touch -d '2 seconds ago' /etc/postfix/main.cf
>   fi
>   return 0
> }
> ```
>
> If `touch` fails for some reason, why would we want to return success (0)?

It's not the touch - it's the condition in the `if`-clause that, when the `if` evaluated to false, would implicitly set the return value to false, i.e. 1.

Looking at it again, you're right about the `if`-is-true case. So a `return ${?}` after the touch should also be added.

georglauterbach:
You are indeed correct, there is a difference between

```bash
set -eE

function x() {
  if false && :; then :; fi
}

x
echo "this will show up"
```

and

```bash
set -eE

function x() {
  false && :
}

x
echo "this wont show up"
```

---

I will correct my error and remove unnecessary `return 0` statements.

PR #3231, opened 2023-04-08
docs: update DKIM/DMARC/SPF docs

georglauterbach:
# Description

Updates the DNS docs about DKIM, DMARC & SPF; we now have one location for setup of these services. I hope all links were properly updated.

**We may want to decide if we want to be even more precise / add even more details to the DKIM/DMARC/SPF page**, but I think it is fine for now, as it already contains more information that previous versions of our docs.

Follow-Up to #3219

## Type of change

- [x] This change is a documentation update

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own codey

PR #3286, opened 2023-04-24
scripts: add DKIM helper script for Rspamd

georglauterbach:
# Description

**DISCLAIMER**: I am sure you think I must be joking looking at the diff size, but I am really not sure how to put this into multiple PRs, because all these ~~changes~~ additions _really_ belong together.. If you think this PR can be split up into multiple PRs, please let me know.

---

This PR is the last PR before we can go into feature-freeze for v12.1.0. It adds support for DKIM signing keys via a helper script for Rspamd. The mode of operation is mostly identical to the already present `open-dkim` script.

As a good programmer, I added tests that show my ~~changes~~ additions are effective. I also updated the documentation for DKIM signing with Rspamd.

---

@polarathene I have not assigned you as a reviewer because I think for the script additions (and this PR mostly is one big script addition including tests) @casperklein will do a good job when it comes to reviewing. When you want to have a look at the docs changes though, I would not mind :) The docs preview should help, because the docs-diffs are always a bit fuzzy.

## Type of change

## Checklist:

### DIFF
--- a/target/bin/open-dkim
+++ b/target/bin/open-dkim
@@ -3,6 +3,12 @@
 # shellcheck source=../scripts/helpers/index.sh
 source /usr/local/bin/helpers/index.sh
 
+if [[ -f /etc/dms-settings ]] && [[ $(_get_dms_env_value 'ENABLE_RSPAMD') -eq 1 ]]
+then
+  /usr/local/bin/rspamd-dkim "${@}"
+  exit ${?}
+fi
+
 KEYSIZE=4096
 SELECTOR=mail
 DOMAINS=
--- a/target/bin/rspamd-dkim
+++ b/target/bin/rspamd-dkim
@@ -1,0 +1,295 @@
+#!/bin/bash
+
+# shellcheck source=../scripts/helpers/index.sh
+source /usr/local/bin/helpers/index.sh
+
+_trap_err_signal
+
+set -u -eE -o pipefail
+shopt -s inherit_errexit
+
+# shellcheck source=/dev/null
+source /etc/dms-settings
+
+function __usage
+{
+  _log 'trace' 'Showing usage message now'
+  echo -e "${PURPLE}RSPAMD-DKIM${RED}(${YELLOW}8${RED})
+
+${ORANGE}NAME${RESET}
+    rspamd-dkim - Configure DomainKeys Identified Mail (DKIM) via Rspamd
+
+${ORANGE}SYNOPSIS${RESET}
+    setup config dkim [ OPTIONS${RED}...${RESET} ]
+
+${ORANGE}DESCRIPTION${RESET}
+    This script aids in creating DKIM signing keys. The keys are created and managed by Rspamd.
+    OPTIONS can be used to configure a more complex setup.
+
+${ORANGE}OPTIONS${RESET}
+    ${BLUE}Generic Program Information${RESET}
+        -v        Enable verbose logging (setting the log level to 'debug').
+        -vv       Enable very verbose logging (setting the log level to 'trace').
+        help      Print the usage information.
+
+    ${BLUE}Configuration adjustments${RESET}
+        keytype   Set the type of key you want to use
+                  Possible values: rsa, ed25519
+                  Default: rsa
+        keysize   Set the size of the keys to be generated
+                  Possible values: 1024, 2048 and 4096
+                  Default: 2048
+                  Only applies when using keytype=rsa
+        selector  Set a manual selector for the key
+                  Default: mail
+        domain    Provide the domain for which keys are to be generated
+                  Default: primary domain name of DMS
+
+${ORANGE}EXAMPLES${RESET}
+    ${LWHITE}setup config dkim keysize 2048${RESET}
+        Creates keys of length 2048 bit in a default setup where domains are obtained from
+        your accounts.
+
+    ${LWHITE}setup config dkim keysize 512 selector 2023-dkim${RESET}
+        Creates keys of length 512 bit in a default setup where domains are obtained from
+        your accounts. The DKIM selector used is '2023-dkim'.
+
+    ${LWHITE}setup config dkim keysize 1024 selector 2023-dkim domain whoami.com${RESET}
+        Creates keys of length 1024 bit in a default setup where domains are obtained from your accounts.
+        The DKIM selector used is '2023-dkim'. The domain for which DKIM keys are generated is whoami.com.
+
+${ORANGE}EXIT STATUS${RESET}
+    Exit status is 0 if command was successful. If wrong arguments are provided or arguments contain
+    errors, the script will exit early with a non-zero exit status.
+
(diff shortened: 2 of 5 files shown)


polarathene:
I likely won't have time to participate in review. Just had a quick peek and noticed you're maintaining both opendkim and rspamd methods of generating the DKIM keys.

There was an [agnostic method I shared earlier](https://github.com/docker-mailserver/docker-mailserver/pull/3231#discussion_r1161355541). However, since you plan to drop opendkim eventually, perhaps that agnostic approach is not as useful.

georglauterbach:
@casperklein this PR is rather big, hence I wanted to ask you whether you actually have time to review this. If you're very short on time, I got to admit that I feel rather confident about these changes, so you could also do a very light review. They may not be perfect, but nothing it. Just tell me so I get the rough picture.

casperklein:
I'll review this tonight.

georglauterbach:
> I'll review this tonight.

Very nice 🚀👍🏼

target/bin/rspamd-dkim
```diff
@@ -0,0 +1,295 @@
...
+  DOMAIN=${DOMAINNAME}
+
+  _log 'trace' "Options given to this script: '${*}'"
+
+  while [[ ${#} -gt 0 ]]
+  do
+    case "${1}" in
+
+      ( 'keytype' )
+        if [[ -n ${2+set} ]]
```
casperklein, line 90: Personally I find this ("assigning" a value) confusing on first view. ${2:-} work just fine (and is shorter).

target/bin/rspamd-dkim
```diff
@@ -0,0 +1,295 @@
...
+  fi
+}
+
+function _check_permissions
+{
+  # shellcheck disable=SC2310
+  if ! __do_as_rspamd_user ls "${BASE_DIR}" >/dev/null
+  then
+    _log 'warn' "The Rspamd user ('_rspamd') seems to be unable to list files in the keys directory ('${BASE_DIR}') - Rspamd may experience permission errors later"
+  elif ! __do_as_rspamd_user cat "${PRIVATE_KEY_FILE}" >/dev/null
```
casperklein, line 220: If the command fails, STDERR should not be shown and redirected to `/dev/null` too?

```suggestion
  elif ! __do_as_rspamd_user cat "${PRIVATE_KEY_FILE}" &>/dev/null
```

target/bin/rspamd-dkim
```diff
@@ -0,0 +1,295 @@
...
+  : >"${PUBLIC_KEY_DNS_FILE}"
+  grep -o '".*"' "${PUBLIC_KEY_FILE}" | tr -d '"\n' >>"${PUBLIC_KEY_DNS_FILE}"
+  printf '\n' >>"${PUBLIC_KEY_DNS_FILE}"
+
+  _log 'info' "Here is the content of the TXT DNS record ${SELECTOR}._domainkey.${DOMAIN} that you need to create:\n"
+  if ! _log_level_is '(warn|error)'
+  then
+    cat "${PUBLIC_KEY_DNS_FILE}"
+    printf '\n'
+  fi
```
casperklein, line 274:
```suggestion
  if ! _log_level_is '(warn|error)'
  then
    _log 'info' "Here is the content of the TXT DNS record ${SELECTOR}._domainkey.${DOMAIN} that you need to create:\n"
    cat "${PUBLIC_KEY_DNS_FILE}"
    printf '\n'
  fi
```

target/bin/rspamd-dkim
```diff
@@ -0,0 +1,295 @@
...
+    chown _rspamd:_rspamd "${DEFAULT_CONFIG_FILE}"
+  fi
+}
+
+function _transform_public_key_file_to_dns_record_contents
+{
+  _log 'trace' 'Transforming DNS zone format to DNS record content now'
+  : >"${PUBLIC_KEY_DNS_FILE}"
+  grep -o '".*"' "${PUBLIC_KEY_FILE}" | tr -d '"\n' >>"${PUBLIC_KEY_DNS_FILE}"
+  printf '\n' >>"${PUBLIC_KEY_DNS_FILE}"
```
casperklein, line 267: Personal opinion: I don't get, why people are forcefully using `printf` in BASH. I know it's standard in other languages, where `echo` is not available.

`printf '%s %s' hello world` --> `echo "hello world"`

Unless there is a special need for `printf`, `echo` is in IMO always better/easier to read:

`echo >>"${PUBLIC_KEY_DNS_FILE}"`

casperklein (changes requested):
Nothing big. Some change requests and some comments you might aggree with and change the code or leave it as it is..

target/bin/open-dkim
```diff
@@ -3,6 +3,12 @@
 # shellcheck source=../scripts/helpers/index.sh
 source /usr/local/bin/helpers/index.sh
 
+if [[ -f /etc/dms-settings ]] && [[ $(_get_dms_env_value 'ENABLE_RSPAMD') -eq 1 ]]
+then
+  /usr/local/bin/rspamd-dkim "${@}"
+  exit ${?}
```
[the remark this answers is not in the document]
georglauterbach: I didn't rember - nice catch!

target/bin/rspamd-dkim
georglauterbach: No, I definitely want the error to be shown! In case of an error, I expect people to copy the log, and having the error shown is very valuable.

target/bin/rspamd-dkim
georglauterbach: Due to the way that `_log` works, it doesn't actually matter. But your version should be a tat more efficient :D 🙈

target/bin/rspamd-dkim
georglauterbach: Did it here to communicate better that a newline is added, but if all agree that `echo` communicates this as well, I have no issue in using `echo`.

georglauterbach, line 267:
```suggestion
  echo '' >>"${PUBLIC_KEY_DNS_FILE}"
```

target/bin/rspamd-dkim
georglauterbach: In this case, you're right, both work just fine. In cases where you'd require quotes though the `${...+sometext}` syntax is more robust. Due to the way that Bash's `[[ ]]` work, it'd be fine to use `:-` here instead of `+`, and I agree, it is easier to grasp.

I can rework that.

target/bin/rspamd-dkim
georglauterbach: Done in [6ab6212](https://github.com/docker-mailserver/docker-mailserver/pull/3286/commits/6ab6212a339c45e5e226f3ba95fe290657eeb797)

casperklein:
Looks good, but one test is failing.

casperklein (approved):

Issue #3289, opened 2023-04-24
other: open TODOs


georglauterbach:
### Subject

I would like to contribute to the project

### Description

I'd like v13.0.0 to be a release that barely adds new functionality, because I think DMS is already packed with functionality. For the far future, I'd like to remove unnecessary features as much as possible (i.e. OpenDKIM, OpenDMARC, Amavis (plus codecs), Spamassassin, Postgrey, Razor, Pyzor, fts-xapian and the likes). With v13.0.0, the default will change to Rspamd. Later, with upcoming major releases after v13.0.0, we can remove functionality.

The script are already high-class, but they lack Bash's integrity and error checks - this should changes as well. Then there are many TODO tickets in the issue tracker that should be resolved.

---

- [ ] Make Rspamd the default, and change default of `1` to `0` for OpenDKIM, OpenDMARC, policyd-spf, Amavis & SpamAssassin
- [x] clean up the issue tracker
- [x] clean up stale PRs (and force-close those that have stalled for a very long time)
- [x] ~~introduce `set -u` and `set -eE` to our scripts to make them more robust~~ (EDIT: could be done, but I lack the time) 
- [x] introduce `set -o pipefail` and `shopt -s inherit_errexit` to our scripts to make them more robust
- [x] ~~deprecate `OVERRIDE_HOSTNAME` and introduce `DMS_FQDN`~~

georglauterbach:
@polarathene could you give me a very rough (short) estimation of what it would take (also time wise) to update our CI so that we can take advantage of the Docker build cache for the ARM64 image as well?

polarathene:
@georglauterbach it's been a while since I thought about it. I recall two options:
- Switching to the beta / experimental [GHA cache driver](https://docs.docker.com/build/cache/backends/gha/). If you look up the PRs where we introduced caching workflow or improvements, I think I provided some additional info/insight there at the time.
- Using the current [local cache driver](https://docs.docker.com/build/cache/backends/local/) with separate job and build cache key (probably prefix with `arm64` to not overlap/mix with `amd64` cache?. Then build via separate job.

While only one build cache can be exported, [multiple can be imported](https://docs.docker.com/build/cache/backends/#multiple-caches). ~~If you perform these builds locally, or in your own fork on Github you should be able to inspect the cache location (_I think the artifact can be downloaded manually via Actions cache management page_). Depending on how it exports the cache, you might be able to upload the amd64 / arm64 build cache to separate keys.~~

[...]

Issue #2254, opened 2021-10-19
[BUG] instructions don't quite work under Unbuntu Linux


goblimey:
### Miscellaneous first checks

- [X] I checked that all ports are open and not blocked by my ISP / hosting provider.
- [X] I know that SSL errors are likely the result of a wrong setup on the user side and not caused by DMS itself. I'm confident my setup is correct.

### Affected Component(s)

readme.md

### What happened and when does this occur?

```Markdown
I installed the latest docker-compose as suggested and downloaded the docker-compose.yml.  It specifies version 3.8 at the top.  I ran the first docker command to create an email address, then ran the second to bring up the server permanently:


docker-compose up -d mailserver
```

I get an error saying that the version in the docker-compose.yml is wrong (see log)

Cranking the version down to 3.5 fixes the error.
```


### What did you expect to happen?

```Markdown
I guess this issue may be caused by different versions in different operating systems - I presume that the supplied docker-compose.yml has been tested and works in some environments.  It would be useful if it could be general-purpose.  If that's not possible then some guidance on this issue would be handy.
```

### How do we replicate the issue?

```Markdown
1. install the mail server on a clean ubuntu server.
2. docker-compose up -d mailserver
3.
...
```

### DMS version

latest

### What operating system is DMS running on?

Linux

### What instruction set architecture is DMS running on?

x86_64 / AMD64

### What container orchestration tool are you using?

Docker Compose

### docker-compose.yml

```yaml
version: '3.8'

services:
  mailserver:
    image: docker.io/mailserver/docker-mailserver:latest
    container_name: mailserver
    hostname: mail
    domainname: goblimey.com
    ports:
      - "25:25"
      - "143:143"
      - "587:587"
      - "993:993"
    volumes:
      - ./docker-data/dms/mail-data/:/var/mail/
      - ./docker-data/dms/mail-state/:/var/mail-state/
      - ./docker-data/dms/mail-logs/:/var/log/mail/
      - ./docker-data/dms/config/:/tmp/docker-mailserver/
      - /etc/letsencrypt:/etc/letsencrypt
      - /etc/localtime:/etc/localtime:ro
    environment:
      - SSL_TYPE=letsencrypt
      - ENABLE_SPAMASSASSIN=1
      - SPAMASSASSIN_SPAM_TO_INBOX=1
      - ENABLE_CLAMAV=1
      - ENABLE_FAIL2BAN=1
      - ENABLE_POSTGREY=1
[...]

PR #3295, opened 2023-04-26
docs/misc: update to align with Docker Compose v2

georglauterbach:
# Description

Align with upstream Docker Compose (v2). See <https://github.com/docker-mailserver/docker-mailserver/issues/3289#issuecomment-1521333815>. @polarathene pinging you here just so you know. I don't want to merge this without any maintainer knowing first. Hence, pinging @casperklein here too. Marked for v13.0.0 because I consider this to be breaking.

## Type of change

## Checklist:


docs/content/config/security/ssl.md
```diff
@@ -385,7 +385,7 @@ The following example is the [basic setup][acme-companion::basic-setup] you need
         container_name: nginx-proxy-acme
         restart: always
         environment:
-          # Only docker-compose v2 supports: `volumes_from: [nginx-proxy]`,
+          # Only Docker Compose v2 supports: `volumes_from: [nginx-proxy]`,
```
polarathene, line 388: This was about the v2 schema, which is probably going to be easy to confuse with Compose V2 (CLI).

[`volumes_from`](https://docs.docker.com/compose/compose-file/compose-file-v2/#volumes_from) (v2 schema) was [dropped in the v3 schema](https://docs.docker.com/compose/compose-file/compose-file-v3/#volumes). However... `docker compose` (V2) may actually support a mix of v2 + v3 schema :sweat_smile: 

**EDIT:** Yup, [here we go](https://docs.docker.com/compose/compose-file/):

> The Compose spec **merges the legacy `2.x` and `3.x` versions**, aggregating properties across these formats and is implemented by Compose 1.27.0+.

- This is when the `version` field no longer became necessary IIRC.
- The docs old compose v2 + v3 references are under the left side-bar legacy item, while the reference we want for this information [now belongs here](https://docs.docker.com/compose/compose-file/05-services/#volumes_from), where we can see `volumes_from` is supported again.

---

Thus AFAIK, this note was about an alternative way with `volumes_from` that worked (_and now does again_), while the documented alternative is some ENV based feature `nginx-proxy-acme` must implement :sweat_smile: 

I think it's fine to drop this line?

```suggestion
```

Or it could provide a bit more context:

```suggestion
          # When `volumes_from: [nginx-proxy]` is not supported,
```

polarathene (changes requested):
LGTM :+1: 

- May want to use upstream preferred `compose.yaml` convention.
- May want to revise a comment regarding `volumes_from`.

docs/content/config/advanced/mail-fetchmail.md
```diff
@@ -2,7 +2,7 @@
 title: 'Advanced | Email Gathering with Fetchmail'
 ---
 
-To enable the [fetchmail][fetchmail-website] service to retrieve e-mails set the environment variable `ENABLE_FETCHMAIL` to `1`. Your `docker-compose.yml` file should look like following snippet:
+To enable the [fetchmail][fetchmail-website] service to retrieve e-mails set the environment variable `ENABLE_FETCHMAIL` to `1`. Your `compose.yml` file should look like following snippet:
```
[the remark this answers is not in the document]
georglauterbach: Resolved in [08c04cb](https://github.com/docker-mailserver/docker-mailserver/pull/3295/commits/08c04cbf584b55c70ebd135c39e0a339e73e700f)

docs/content/config/security/ssl.md
georglauterbach: Resolved in [2f9358d](https://github.com/docker-mailserver/docker-mailserver/pull/3295/commits/2f9358d587e6f5b6ce332e9fff760c560b60a4d6)

polarathene (approved):
Nice! :grin:

georglauterbach:
**INFO**: Marked this with `meta/feature-freeze` now to indicate this PR is not to be merged before v12.1.0 is released.

casperklein (approved):

casperklein:
I am unsure about the  renaming of `docker-compose.yaml` to `compose.yaml`. I've looked around and couldn't find any other projects using `compose.yaml`.
`docker-compose.yaml` is not deprecated, well-know and I don't think it will go away anytime soon. To avoid confusion, I would keep the old file name for now.

polarathene:
> I am unsure about the renaming of `docker-compose.yaml` to `compose.yaml`. I've looked around and couldn't find any other projects using `compose.yaml`.

Just in case you missed them, I unresolved the earlier review feedback for visibility. It [quotes Docker docs](https://github.com/docker-mailserver/docker-mailserver/pull/3295#discussion_r1177545693) that from July 2023 they're encouraging `compose.yaml` as the preferred convention.

Other projects may not be aware of that, or may not see any reason to adjust. You'll likely see a mix with time, we can do our part, but it's not mandatory.

---

> `docker-compose.yaml` is not deprecated, well-know and I don't think it will go away anytime soon. To avoid confusion, I would keep the old file name for now.

We have had confusion over dropping the `version` field (_[here's a recent example](https://github.com/docker-mailserver/docker-mailserver/issues/2254#issuecomment-1528774201) from a user with Debian Bullseye_), but that hasn't stopped us even though the `version` field is optional and wouldn't cause breakage to have kept?

Once Docker updates their docs later this year, they'll be using `compose.yaml` and presumably with newer distro releases `docker-compose` will more commonly be used as V2 `docker compose`. New guides online for Docker will probably start using the new V2 conventions, and then we introduce added confusion to anyone new to Docker?

Up to you, but as long as our docs make mention about V2 along with changelog / release notes announcement, I don't see the confusion concern being relevant as it'll apply to different users regardless. Aligning with Dockers documented usage makes the most sense and something we would transition to at some point anyway.

We get various issues / PRs to "fix" support for versions of compose that are over 3 years old that latest releases like Debian package. You will probably experience the same either way with this change.

georglauterbach:
I'm definitely in favor of merging this as well. I think we should be up-to-date with our main container runtime. And users having issues with this due to using outdated packages is nothing new, and IMO no reason not to update :)

PR #3316, opened 2023-05-10
chore(main.cf): Add note advising caution changing `mydestination`

polarathene:
# Description

Changing the Postfix `mydestination` setting is fragile. Review feedback and insights have been linked in the `main.cf` config as a way to document awareness of this information better to maintainers/contributors in future.

Should a change be proposed again for `mydestination`, it is clear that it has a wider impact to keep in mind than it may originally appear to for DMS.

---

https://github.com/docker-mailserver/docker-mailserver/pull/3264 proposed modifying `mydestination`, but was decided that there was too much to take into consideration for an improvement.
- It would likely simplify config for bare domain users, but risk breakage for those using a subdomain (`mail.example.com`)
- `localhost` could be used, and some related changes were considered. However that would affect the address Postfix used for sending mail from (`MAILER-DAEMON@localhost`), which would be problematic. System mail and aliases also came into consideration.
- Nobody had the time available to go over all the concerns raised, and verify a change to `mydestination` would be an improvement without regressions.

Thus the current behaviour difference remains (neither ideal):
- Bare domain setups have a disadvantage when a virtual alias domain manages that same domains mail accounts, since `mydestination` is no longer configured to recognize local mail, the local aliases for `amavis`, `root`, `clamav`, etc would not resolve to local delivery. Setting virtual alias recipients should work, but is not ideal at creating a distinction between user and system accounts/aliases:
  - To receive Postmaster notifications from Postfix, an explicit virtual alias must now be created to direct the mail to a virtual mailbox managed by Dovecot, `/etc/aliases` may not be checked due to `mydestination` not matching (_eg: Postfix sender address is `postmaster@example.com`, `@example.com` is configured as vhost and thus queries Dovecot if no virtual alias was set_).
  - You probably don't want to set a virtual alias wildcard/regex for `postmaster`, since that would accept any recipient address matching, virtual aliases apply not only to inbound mail recipients, but also outbound recipients IIRC.
- Using our advised subdomain assignment for DMS (`mail.example.com`) will allow managing mail for `example.com` via virtual alias domain just like with bare domain setup. However, `mydestination` will represent the DMS MTA itself, recognizing mail destined to the DMS FQDN as specifically for that purpose. Typically this is just for system accounts (from `/etc/passwd`) to notify the postmaster/admin, and the `/etc/aliases` to add additional accounts like `postmaster` that will accept mail when the local part of the recipient address matches (_with the domain part qualifying as a match to `mydestination`_). If an alias is not added to forward that mail to an actual mailbox however, Postfix will deliver it to `/var/mail/username` in a mailbox file which Dovecot will not be aware of.

## Type of change

## Checklist:


georglauterbach (approved):

polarathene:
@georglauterbach I figured this was a better location than a page or FAQ entry in the docs which'd probably have less visibility to those wanting to actually contribute a PR to change it.

georglauterbach:
> @georglauterbach I figured this was a better location than a page or FAQ entry in the docs which'd probably have less visibility to those wanting to actually contribute a PR to change it.

Agreed 👍🏼

casperklein (approved):

PR #3317, opened 2023-05-10
ci: slim down bug report template

georglauterbach:
# Description

This PR eases using the bug report template. It had indeed become a bit heavy. @polarathene if you have any suggestions, you can apply them immediately without me having a second look. You can then go ahead and merge this straight away :)

## Type of change

- [x] This change more or less is a documentation update

## Checklist:

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -50,7 +50,7 @@ body:
       label: What computer architecture is the host running?
       options:
         - AMD64 (x86_64)
-        - ARM64 (ARMv8 and newer)
+        - ARM64 (AArch64)
         - Other (unsupported)
     validations:
       required: true
@@ -80,7 +80,7 @@ body:
       label: What happened and when does this occur?
       description: >
         Tell us what happened.
-        **Make sure to give us instruction on how we replicate the issue! Be precise and include as much information as possible!**
+        **It would also help immensely if you can share precise steps on how to reproduce the issue! ** :heart: 
         Use [fenced code blocks](https://docs.github.com/en/github/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks#fenced-code-blocks) when pasting lots of text!
       placeholder: Although `LOG_LEVEL=debug` is set, I see no debug output.
     validations:
@@ -111,7 +111,7 @@ body:
   - type: checkboxes
     id: experience
     attributes:
-      label: What level of experience do you have with containers, mail servers and command line interfaces?
+      label: What level of experience do you have with containers, mail servers, and using a terminal?
       description: >
         **You are not obliged to answer this question**.
         We do encourage answering though as it provides context to better assist you.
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -97,16 +41,16 @@ body:
...
   - type: dropdown
     id: isa
     attributes:
-      label: What instruction set architecture is DMS running on?
+      label: What computer architecture is the host running?
       options:
-        - AMD64 / x86_64
-        - ARM64 / AArch64 (ARM v8 and newer)
+        - AMD64 (x86_64)
+        - ARM64 (ARMv8 and newer)
```
polarathene, line 53: `AArch64` is equivalent to `x86_64` IIRC?  ARMv8 is the oldest supported, while ARMv7 is for `AArch32`: https://stackoverflow.com/questions/31851611/differences-between-arm64-and-aarch64

```suggestion
        - ARM64 (AArch64)
```

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -32,60 +18,18 @@ body:
...
-        - label: I read the [documentation on debugging](https://docker-mailserver.github.io/docker-mailserver/latest/config/debugging/), tried the proposed steps to debug the problem, but was still unable to resolve the issue.
-          required: true
-        - label: I have read this project's [Code of Conduct](https://github.com/docker-mailserver/docker-mailserver/blob/master/CODE_OF_CONDUCT.md) and I agree
+        - label: |
+            I read
+
+              - the [extended documentation in general](https://docker-mailserver.github.io/docker-mailserver/latest/) but found nothing to resolve the issue;
+              - the [documentation on debugging](https://docker-mailserver.github.io/docker-mailserver/latest/config/debugging/), tried the proposed steps to debug the problem, but was still unable to resolve the issue;
+              - this project's [Code of Conduct](https://github.com/docker-mailserver/docker-mailserver/blob/master/CODE_OF_CONDUCT.md) and I agree;
+              - the [documentation about filing a bug report](https://docker-mailserver.github.io/docker-mailserver/latest/contributing/issues-and-pull-requests/#filing-a-bug-report).
```
polarathene, line 27: who actually does that? :P

This will just get checked regardless, but here's hoping some at least gain awareness through the links like debugging docs :+1:

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -97,16 +41,16 @@ body:
...
       label: Which operating system version?
-      placeholder: e.g. Debian 11
+      placeholder: Debian 11 (Bullseye)
     validations:
       required: true
   - type: dropdown
     id: isa
     attributes:
-      label: What instruction set architecture is DMS running on?
+      label: What computer architecture is the host running?
```
polarathene, line 50: Assuming this change is because ISA won't make sense to many? :sweat_smile: (_the choices should regardless though_)

polarathene (changes requested):
:+1: :rocket:

polarathene (approved):

.github/ISSUE_TEMPLATE/bug_report.yml
georglauterbach: 😆 Absolutely right! We did our part though 👍🏼

PR #3332, opened 2023-05-12
scripts: improve `CLAMAV_MESSAGE_SIZE_LIMIT` usage

georglauterbach:
# Description

Addes sanity check for ClamAV size & adjusted MaxScanSize. The second part is of special importance! See <https://askubuntu.com/a/1448525>, which explains that the maximum scan size is important as well. We previously just set the maximum file size, which actually is pretty insecure as we silently not scan mile bigger than `MaxScanSize`. This is corrected now.

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/startup/setup.d/security/misc.sh
+++ b/target/scripts/startup/setup.d/security/misc.sh
@@ -188,7 +188,7 @@ function __setup__security__clamav
       _log 'trace' "Setting ClamAV message scan size limit to '${CLAMAV_MESSAGE_SIZE_LIMIT}'"
 
       # do a short sanity checks; ClamAV stops scanning at more that 4GB file size
-      if [[ $(numfmt --from=si "${CLAMAV_MESSAGE_SIZE_LIMIT}") > $(numfmt --from=si 4G) ]]
+      if [[ $(numfmt --from=si "${CLAMAV_MESSAGE_SIZE_LIMIT}") -gt $(numfmt --from=si 4G) ]]
       then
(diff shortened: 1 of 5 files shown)


georglauterbach:
I added this to the Rspmd tracking project, but this is primarily about ClamAV.

casperklein (approved):

target/scripts/startup/setup.d/security/misc.sh
```diff
@@ -186,8 +186,18 @@ function __setup__security__clamav
     if [[ ${CLAMAV_MESSAGE_SIZE_LIMIT} != '25M' ]]
     then
       _log 'trace' "Setting ClamAV message scan size limit to '${CLAMAV_MESSAGE_SIZE_LIMIT}'"
-      sedfile -i \
-        "s/^MaxFileSize.*/MaxFileSize ${CLAMAV_MESSAGE_SIZE_LIMIT}/" \
+
+      # do a short sanity checks; ClamAV stops scanning at more that 4GB file size
+      if [[ $(numfmt --from=si "${CLAMAV_MESSAGE_SIZE_LIMIT}") -gt $(numfmt --from=si 4G) ]]
```
ap-wtioit, line 191: sorry, for catching this now. i think clamav is not using the SI system for their limit variables.

i think the M / G in the clamav variables is actually MiB / GiB, so this check is not correct as it checks for 4000000000 instead of 4294967296.

Could it be that the correct check would involve:
```
if [[ $(numfmt --from=iec "${CLAMAV_MESSAGE_SIZE_LIMIT}") -gt $(numfmt --from=iec 4G) ]]
```

Issue #3358, opened 2023-05-21
bug report: `setup alias add` may fail when recipient exists for another alias


42l-neil-tmp:
### Preliminary Checks

- [X] I checked that all ports are open and not blocked by my ISP / hosting provider.
- [X] I know that SSL errors are likely the result of a wrong setup on the user side and not caused by DMS itself. I'm confident my setup is correct.
- [X] I searched the issue tracker but was unable to find my issue.
- [X] I read

  - the [extended documentation in general](https://docker-mailserver.github.io/docker-mailserver/latest/) but found nothing to resolve the issue;
  - the [documentation on debugging](https://docker-mailserver.github.io/docker-mailserver/latest/config/debugging/), tried the proposed steps to debug the problem, but was still unable to resolve the issue;
  - this project's [Code of Conduct](https://github.com/docker-mailserver/docker-mailserver/blob/master/CODE_OF_CONDUCT.md) and I agree;
  - the [documentation about filing a bug report](https://docker-mailserver.github.io/docker-mailserver/latest/contributing/issues-and-pull-requests/#filing-a-bug-report).

### What operating system is the host running?

Linux

### Which operating system version?

Debian 11

### What computer architecture is the host running?

AMD64 (x86_64)

### What container orchestration tool are you using?

Docker Compose

### DMS version

v12.1.0

### What happened and when does this occur?

This issue is similar to #3350 and #3262.

I want to create group aliases, as described in https://github.com/docker-mailserver/docker-mailserver/issues/3262#issuecomment-1509366419 : a fairly simple setup involving two existing mailboxes and an alias used to redirect a mail to those two mailboxes.

Those mailboxes are **not external**: they are **hosted by our DMS instance**, unlike the aforementioned bug reports, hence this new report.

In the following example, **both domain.org and other-domain.org are hosted on our DMS instance**.

```sh
$> ./setup.sh email add testemail1@domain.org
Enter Password:
$> ./setup.sh email add testemail2@domain.org
Enter Password:
$> ./setup.sh alias add group@domain.org testemail1@domain.org
$> ./setup.sh alias add group@domain.org testemail2@domain.org
$> ./setup.sh alias list | grep group
* group@domain.org testemail1@domain.org,testemail2@domain.org
$> ./setup.sh alias add group@other-domain.org testemail1@domain.org
$> ./setup.sh alias add group@other-domain.org testemail2@domain.org
[...]

PR #3359, opened 2023-05-22
fix: DB helper should properly filter entries

polarathene:
# Description

Previously it was assumed the sed operation was applying the sed expressions as a sequence, but it did not seem to filter entries being looked up correctly. Instead any line that matched either sed expression pattern was output (_value without matching key, values split by the delimiter_), then grep would match any of that causing false-positives.

Resolved by piping the first sed expression into the next.

Fixes: https://github.com/docker-mailserver/docker-mailserver/issues/3358

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/helpers/database/db.sh
+++ b/target/scripts/helpers/database/db.sh
@@ -132,11 +132,11 @@ function __db_list_already_contains_value
 {
   # Avoids accidentally matching a substring (case-insensitive acceptable):
   # 1. Extract the current value of the entry (`\1`),
-  # 2. If a value list, split into separate lines (`\n`+`g`) at V_DELIMITER,
+  # 2. Value list support: Split values into separate lines (`\n`+`g`) at V_DELIMITER,
   # 3. Check each line for an exact match of the target VALUE
-  sed -e "s/^${KEY_LOOKUP}\(.*\)/\1/" \
-      -e "s/${V_DELIMITER}/\n/g"      \
-      "${DATABASE}" | grep -qi "^${_VALUE_}$"
+  sed -ne "s/^${KEY_LOOKUP}\+\(.*\)/\1/p" "${DATABASE}" \
+    | sed -e "s/${V_DELIMITER}/\n/g" \
+    | grep -qi "^${_VALUE_}$"
 }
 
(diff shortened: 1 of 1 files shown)


polarathene:
Probably a case where an improved CI for patch releases would be handy :sweat_smile: 

I guess this will have to wait until v13?

georglauterbach (approved):

georglauterbach:
> Probably a case where an improved CI for patch releases would be handy :sweat_smile: 

I think there are no maintainer capacity limits left 😂

> I guess this will have to wait until v13?

Yes

polarathene:
> I think there are no maintainer capacity limits left

:100: 

> Yes. But there really isn't too much that's blocking v13.0.0.

:+1: 

> And then there is only the IPv6 PR.

Yep still working on that digging through upstream moby issues atm :sweat_smile: 

I recently learned that due to Docker setting `sysctl net.ipv4.ip_forward=1` and it's IPTables rules for published ports, you can route to `127.0.0.1` and docker network subnets private address space from another host within the same LAN. Technically another open relay :grimacing: 

Originally I was looking into a similar vulnerability for `userland-proxy: false` setting `route_localnet=1`, as this was an issue for k8s in the past, but doesn't appear to be reproducible with Dockers network management.

I've confirmed at least with `firewalld` active that the LAN hosts can't route to the hosts private subnets. Once the `userland-proxy` investigation is wrapped up (almost done), the IPv6 docs PR will resume :+1: 

---

I think this is fine to merge now.

Casper already gave a :+1: to the solution shared on the reported issue earlier, any changes can be done via follow-up PR if necessary :)

casperklein:
Maybe I've overlooked something, but afaik there were no breaking changes since `12.1.0`. As of now, the next release could be `12.2.0`?

polarathene:
Or just publish as v13, doesn't have to be a big release to warrant it :sweat_smile:

georglauterbach:
> Or just publish as v13, doesn't have to be a big release to warrant it :sweat_smile: 

I'd go for v13 as well. I reall don't mind having higher numbers - I'm not Linus Torvalds..

PR #3373, opened 2023-05-28
Dovecot: compile `fts_xapian` from source to match Dovecot ABI

### LINKED ISSUE #3247: other: fts-xapian-plugin does not work after upgrading dms to 12.0
### Subject

Other

### Description

If you use the fts-xapian-plugin specified in the documentation https://docker-mailserver.github.io/docker-mailserver/edge/config/advanced/full-text-search/#xapian , then after updating dms to 12.0, it will stop working for you. On 11.3.1 I am not getting any errors. I found [mention of a similar problem since 2022 in the repository of this plugin](https://github.com/grosjo/fts-xapian/issues/125). This was not fixed, and after a while the error was closed by the initiator. I think it would be right to remove the recommendation of this plugin from the documentation concerning 12.0, latest and edge, because everything works on 11.3.1 (I checked on my amd64 configuration, Debian 11, Docker 23.0.2, build 569dd73, Docker Compose v2.17.2). Or you can specify a link in the documentation to your answer [here](https://github.com/docker-mailserver/docker-mailserver/issues/3100#issuecomment-1435971388) to a similar error as an option for using this plugin.
when start container:
```
[  DEBUG  ]  Configuring Amavis
[  DEBUG  ]  Spam emails will be moved to the Junk folder
sievec: Fatal: Couldn't load required plugin /usr/lib/dovecot/modules/lib21_fts_xapian_plugin.so: Module is for different ABI version 2.3.ABIv13(2.3.13) (we have 2.3.ABIv20(2.3.20))
chown: cannot access '/usr/lib/dovecot/sieve-global/after/spam_to_junk.svbin': No such file or directory
```

tbutter:
# Description

compile fts_xapian from src to match dovecot abi version

Fixes #3247

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,17 +1,17 @@
 # syntax=docker.io/docker/dockerfile:1
 
-# This Dockerfile provides two stages: stage-base and stage-final
+# This Dockerfile provides three stages: stage-dep, stage-base and stage-final
 # This is in preparation for more granular stages (eg ClamAV and Fail2Ban split into their own)
-
-#
-# Base stage provides all packages, config, and adds scripts
-#
-
-FROM docker.io/debian:11-slim AS stage-base
 
 ARG DEBIAN_FRONTEND=noninteractive
 ARG DOVECOT_COMMUNITY_REPO=1
 ARG LOG_LEVEL=trace
+
+FROM docker.io/debian:11-slim AS stage-compile
+
+ARG DEBIAN_FRONTEND
+ARG DOVECOT_COMMUNITY_REPO
+ARG LOG_LEVEL
 
 SHELL ["/bin/bash", "-e", "-o", "pipefail", "-c"]
 
@@ -25,10 +25,51 @@ RUN <<EOF
   adduser --quiet --system --group --disabled-password --home /var/lib/clamav --no-create-home --uid 200 clamav
 EOF
 
-COPY target/scripts/build/* /build/
+COPY target/scripts/build/packages.sh /build/
 COPY target/scripts/helpers/log.sh /usr/local/bin/helpers/log.sh
 
 RUN /bin/bash /build/packages.sh
+
+# -----------------------------------------------
+# --- Compile deb packages ----------------------
+# -----------------------------------------------
+
+COPY target/scripts/build/compile.sh /build/
+RUN /bin/bash /build/compile.sh
+
+#
+# Base stage provides all packages, config, and adds scripts
+#
+
+FROM docker.io/debian:11-slim AS stage-base
+
+ARG DEBIAN_FRONTEND
+ARG DOVECOT_COMMUNITY_REPO
+ARG LOG_LEVEL
+
+SHELL ["/bin/bash", "-e", "-o", "pipefail", "-c"]
+
+# -----------------------------------------------
+# --- Install Basic Software --------------------
+# -----------------------------------------------
+
+COPY target/bin/sedfile /usr/local/bin/sedfile
+RUN <<EOF
+  chmod +x /usr/local/bin/sedfile
+  adduser --quiet --system --group --disabled-password --home /var/lib/clamav --no-create-home --uid 200 clamav
+EOF
+
+COPY target/scripts/build/packages.sh /build/
+COPY target/scripts/helpers/log.sh /usr/local/bin/helpers/log.sh
+
+RUN /bin/bash /build/packages.sh
+
+# -----------------------------------------------
+# --- Install xapian Software -------------------
+# -----------------------------------------------
+
+COPY --from=stage-compile dovecot-fts-xapian-1.5.5_1.5.5_amd64.deb /
+RUN apt install libxapian30 && dpkg -i /dovecot-fts-xapian-1.5.5_1.5.5_amd64.deb && rm /dovecot-fts-xapian-1.5.5_1.5.5_amd64.deb
 
 # -----------------------------------------------
 # --- ClamAV & FeshClam -------------------------
--- a/Dockerfile
+++ b/Dockerfile
@@ -59,8 +59,8 @@ SHELL ["/bin/bash", "-e", "-o", "pipefail", "-c"]
 # --- Install xapian Software -------------------
 # -----------------------------------------------
 
-COPY --from=stage-compile dovecot-fts-xapian-1.5.5_1.5.5_amd64.deb /
-RUN dpkg -i /dovecot-fts-xapian-1.5.5_1.5.5_amd64.deb && rm /dovecot-fts-xapian-1.5.5_1.5.5_amd64.deb
+COPY --from=stage-compile dovecot-fts-xapian-1.5.5_1.5.5_*.deb /
+RUN dpkg -i /dovecot-fts-xapian-1.5.5_1.5.5_*.deb && rm /dovecot-fts-xapian-1.5.5_1.5.5_*.deb
 
 # -----------------------------------------------
 # --- ClamAV & FeshClam -------------------------
--- a/Dockerfile
+++ b/Dockerfile
@@ -53,12 +53,14 @@ ARG DEBIAN_FRONTEND
 ARG DOVECOT_COMMUNITY_REPO
 ARG LOG_LEVEL
 
+SHELL ["/bin/bash", "-e", "-o", "pipefail", "-c"]
+
 # -----------------------------------------------
 # --- Install xapian Software -------------------
 # -----------------------------------------------
 
 COPY --from=stage-compile dovecot-fts-xapian-1.5.5_1.5.5_amd64.deb /
-RUN apt-get install -y libxapian30 && dpkg -i /dovecot-fts-xapian-1.5.5_1.5.5_amd64.deb && rm /dovecot-fts-xapian-1.5.5_1.5.5_amd64.deb
+RUN dpkg -i /dovecot-fts-xapian-1.5.5_1.5.5_amd64.deb && rm /dovecot-fts-xapian-1.5.5_1.5.5_amd64.deb
 
 # -----------------------------------------------
 # --- ClamAV & FeshClam -------------------------
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,45 +1,11 @@
 # syntax=docker.io/docker/dockerfile:1
 
-# This Dockerfile provides three stages: stage-dep, stage-base and stage-final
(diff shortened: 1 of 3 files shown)


georglauterbach:
How much larger will the final image be with these changes?

tbutter:
> How much larger will the final image be with these changes?

My local build is 870MB vs. latest release 616MB.
The issue with the community builds of dovecot is that any plugin needs to be built again with that version.

The only way I see to avoid image bloat is to split the docker build: First install dovecot from community, compile fts_xapian, build a .deb.

Then next step the full build copying only the dep from the first build.

But that would be a larger change.

Probably uninstalling the build dep packages will help a bit. I'll try that.

georglauterbach:
Good idea! Because 870MB is definitely too large.

tbutter:
> Good idea! Because 870MB is definitely too large. The multi-stage build may be the best option, but it'd be nice if you tinker a bit around 👍🏼

removing deps brings it down to 700MB

polarathene:
Perhaps put it behind a build flag? 80MB is a lot of extra weight for a feature not many are using AFAIK? (_definitely better than 870MB at least_) Not ideal since it requires users that do want the feature to build the image themselves though..

With Fedora and OpenSUSE this is easier at least, but I'm not that familiar with Debian. Would there be any way to reduce the image weight further?

- The situation might change with the upcoming Debian Bookworm release next month? (_Dovecot community build could be made opt-in if it's new enough, and thus the two may be compatible by default meeting most users needs?_)
- We could defer the build to a separate stage and build / tag releases with that (_but I don't think this one feature warrants that, plus it'd affect our own limited CI cache storage_)
- Changing base image has been discussed in the past which may also resolve the issue (_eg: Fedora releases more frequently and isn't holding back package updates like how Bookworm will still be stuck on PostSRSd <2_), however most active maintainers are in favour of sticking with Debian.

casperklein:
- fts-xapian-1.5.5
- dovecot-fts-xapian.tar.gz

Those can be deleted after compilation?

Are the *-dev packages needed after compilation?

tbutter:
the new patch brings it down to 619MB, but it changes the Dockerfile structure. Is this an acceptable solution?

target/scripts/build/packages.sh
```diff
@@ -205,6 +191,10 @@ function _install_getmail() {
   apt-get "${QUIET}" autoremove
 }
 
+function _install_xapian() {
```
casperklein, line 194:
```suggestion
function _install_dovecot_fts_xapian_dependency() {
```

target/scripts/build/packages.sh
```diff
@@ -228,5 +218,6 @@ _install_dovecot
 _install_rspamd
 _install_fail2ban
 _install_getmail
+_install_xapian
```
casperklein, line 221:
```suggestion
_install_dovecot_fts_xapian_dependency
```

target/scripts/build/compile.sh
```diff
@@ -0,0 +1,30 @@
...
+  curl -Lso dovecot-fts-xapian.tar.gz https://github.com/grosjo/fts-xapian/releases/download/1.5.5/dovecot-fts-xapian-1.5.5.tar.gz
+  tar xzvf dovecot-fts-xapian.tar.gz
+  cd fts-xapian-1.5.5
+  USER=root dh_make -p dovecot-fts-xapian-1.5.5 --single --native --copyright gpl2 -y
+  rm debian/*.ex debian/*.EX
+  cp PACKAGES/DEB/control debian/
+  cp PACKAGES/DEB/changelog debian/
+  cp PACKAGES/DEB/compat debian/
+  sed -i 's/1\.4\.11-6/1.5.5/g' debian/control
+  sed -i 's/1\.4\.11-6/1.5.5/g' debian/changelog
```
casperklein, line 25: Just wondering: Is this an upstream bug? Or why does the control file contains the wrong version?

casperklein (commented):
Just some minor suggestions, otherwise looks good 👍

target/scripts/build/compile.sh
tbutter: just me guessing:
the author had a ppa where he published the plugin. i think he stopped when it was included in debian and then just did not update the debian files anymore.
debian build tools need an up to date debian changelog (last entry of changelog == package version). since it is just a temporary package I change the version of the last entry.

georglauterbach (dismissed):
LGTM 👍🏼 Very nice!

casperklein (dismissed):

Dockerfile
```diff
@@ -25,11 +25,43 @@ RUN <<EOF
...
+RUN /bin/bash /build/compile.sh
+
+#
+# main stage provides all packages, config, and adds scripts
+#
+
+FROM stage-base AS stage-main
+
+ARG DEBIAN_FRONTEND
+ARG DOVECOT_COMMUNITY_REPO
```
polarathene, line 53: Not relevant? Only used for `build/packages.sh` in prior stage?

Dockerfile
```diff
@@ -25,11 +25,43 @@ RUN <<EOF
...
+ARG LOG_LEVEL
+
+SHELL ["/bin/bash", "-e", "-o", "pipefail", "-c"]
```
polarathene, line 56: Not 100% sure but I think this would be inherited from the prior stage `SHELL`.

Dockerfile
```diff
@@ -25,11 +25,43 @@ RUN <<EOF
...
+
+# -----------------------------------------------
+# --- Install xapian Software -------------------
+# -----------------------------------------------
+
+COPY --from=stage-compile dovecot-fts-xapian-1.5.5_1.5.5_*.deb /
+RUN dpkg -i /dovecot-fts-xapian-1.5.5_1.5.5_*.deb && rm /dovecot-fts-xapian-1.5.5_1.5.5_*.deb
```
polarathene, line 63: This is Dovecot specific right? May be better located within the dedicated Dovecot section With single line comment instead of section header.

Being so early into this stage, it also invalidates any other layers after it should the compile stage be invalidated. We could use `COPY --link` too, but that would increase the build cache that we upload into CI storage thus probably not worth it :sweat_smile:

Dockerfile
```diff
@@ -25,11 +25,43 @@ RUN <<EOF
...
 
 RUN /bin/bash /build/packages.sh && rm -r /build
 
+
+
+# -----------------------------------------------
+# --- Compile deb packages ----------------------
+# -----------------------------------------------
+
+FROM stage-base AS stage-compile
```
polarathene, line 39: May not be worth it, but presently this is dependent on `stage-base` installing packages, whereas that's all the prior stage is really doing differently, it could build in parallel if it wasn't dependent upon that (_may affect our CI build cache size though, so perhaps the slower build approach here is acceptable_).

target/scripts/build/compile.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# -eE :: exit on error (do this in functions as well)
+# -u  :: show (and exit) when using unset variables
+# -o pipefail :: exit on error in pipes
+set -eE -u -o pipefail
+
+# shellcheck source=../helpers/log.sh
+source /usr/local/bin/helpers/log.sh
+
+_log_level_is 'trace' && QUIET='-y' || QUIET='-qq'
```
polarathene, line 11: This is fixed as the `LOG_LEVEL` arg is missing in the stage running this script.

target/scripts/build/packages.sh
```diff
@@ -94,26 +94,12 @@ function _install_dovecot() {
   declare -a DOVECOT_PACKAGES
 
   DOVECOT_PACKAGES=(
-    dovecot-core dovecot-fts-xapian dovecot-imapd
+    dovecot-core dovecot-imapd dovecot-dev
```
polarathene, line 97: We are always using a compiled FTS Xapian now? How much does that add to build time :thinking: (_I suppose it's fine, cache should be fine so long as it's not invalidated by this `packages.sh` file being updated too often_ :sweat_smile: )

target/scripts/build/packages.sh
```diff
@@ -205,6 +191,10 @@ function _install_getmail() {
   apt-get "${QUIET}" autoremove
 }
 
+function _install_dovecot_fts_xapian_dependency() {
+  apt-get "${QUIET}" --no-install-recommends install libxapian30
+}
```
polarathene, line 196: Is this necessary for a single line?

Keeping isolation is fine, but I think that can just be handled as a separate line in `_install_dovecot` with adding the dependency to `DOVECOT_PACKAGES` array. It can have a comment above it to clarify it's a dependency specifically for the FTS Xapian support.

polarathene (approved):
Great work getting the size to be minimal :tada: 

I've left some feedback, some which is probably worth addressing, but I won't block on it :+1:

target/scripts/build/packages.sh
tbutter: Currently it would always use the compiled one. Did not measure the time, but packages.sh seems to be most of the build time.

Dockerfile
polarathene: Yeah I think there's a bug report about that on hadolint but they don't have a good solution for it :man_shrugging:

polarathene (approved):
Thanks, LGTM :+1:

georglauterbach (dismissed):

georglauterbach:
@casperklein you can go ahead and merge this when all is well and you approve

casperklein (approved):

casperklein:
Is the `dovecot-dev` package needed after compilation?

tbutter:
> Is the `dovecot-dev` package needed after compilation?

It contains the headers and /usr/share/dovecot/dovecot-abi - i am not 100% sure if the latter is needed. But it is just about 300kb.

casperklein:
Just for the record, ~3MB, not 300kb 😉 

```
root@mail:/# aptitude show dovecot-dev
Package: dovecot-dev
Version: 2:2.3.20-3+debian11
Architecture: amd64
Uncompressed Size: 3431 k
```

polarathene (dismissed):

tbutter:
> Just for the record, ~3MB, not 300kb 😉

oh, you are right. i looked up the wrong version. let me test if it works without the -dev.

tbutter:
> Is the `dovecot-dev` package needed after compilation?

works without the -dev package. i changed the build. image size 619MB->618MB

casperklein (approved):

polarathene (approved):

georglauterbach (approved):

Issue #3374, opened 2023-05-28
bug report: Rspamd crash when sending or receiving


FriderKlugser:
### Preliminary Checks

- [X] I checked that all ports are open and not blocked by my ISP / hosting provider.
- [X] I know that SSL errors are likely the result of a wrong setup on the user side and not caused by DMS itself. I'm confident my setup is correct.
- [X] I searched the issue tracker but was unable to find my issue.
- [X] I read

  - the [extended documentation in general](https://docker-mailserver.github.io/docker-mailserver/latest/) but found nothing to resolve the issue;
  - the [documentation on debugging](https://docker-mailserver.github.io/docker-mailserver/latest/config/debugging/), tried the proposed steps to debug the problem, but was still unable to resolve the issue;
  - this project's [Code of Conduct](https://github.com/docker-mailserver/docker-mailserver/blob/master/CODE_OF_CONDUCT.md) and I agree;
  - the [documentation about filing a bug report](https://docker-mailserver.github.io/docker-mailserver/latest/contributing/issues-and-pull-requests/#filing-a-bug-report).

### What operating system is the host running?

Linux

### Which operating system version?

Debian 11 (Bullseye)

### What computer architecture is the host running?

ARM64 (AArch64)

### What container orchestration tool are you using?

Docker Compose

### DMS version

v12.1.0

### What happened and when does this occur?

From time to time Rspamd crashs when sending or receiving an e-mail. And then doesn't check e-mails properly, as seen (missing) in the Web interface history.

I'm running DMS on an Raspberry Pi 4 (8 GB).

### What did you expect to happen?

No Rspamd crashs.

### Container configuration files

compose.yaml
```YAML
services:
  Mailserver:
    container_name: Mailserver
    hostname: mx.domain.tld
    image: docker.io/mailserver/docker-mailserver:latest
    pull_policy: always
    healthcheck: {disable: true}
    restart: unless-stopped
    stop_grace_period: 1m30s
    stdin_open: true
    tty: true
    networks: [bridge-mail]
    cap_add: [NET_ADMIN]
    ports:
    - "25:25"
    - "143:143"
    - "465:465"
    - "587:587"
    - "993:993"
    - "4190:4190" # ManageSieve
    - "11334:11334" # Rspamd
    volumes:
    - ./Mailserver/config:/tmp/docker-mailserver:rw
    - ./Mailserver/data:/var/mail:rw
    - ./Mailserver/state:/var/mail-state:rw
[...]

PR #3158, opened 2023-03-05
fix: SRS setup

casperklein:
# Description

This PR refactores the SRS setup, while fixing https://github.com/docker-mailserver/docker-mailserver/pull/3144#discussion_r1125664704 

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Improvement (non-breaking change that does improve existing functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [ ] If necessary I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #2064, opened 2021-07-03
add dovecot-fts-xapian

eleith:
# Description

this PR updates Docker to install dovecot-fts-xapian.

unfortunately, that package is only available on `buster-backports` so either Docker needs to be updated to pull from backports or the build can use `bullseye-slim` as the base image (which is nearing stable release in July 2021).

currently, this PR introduces `bullseye-slim` which is a rather large change.

I imagine this project will eventually adopt bullseye given its long testing cycle and relative stability. If so, this PR can be merged with that branch.

With that said, I would be happy to use `buster-backports` as well and i have a working branch, if desired. That approach merely requires updating Docker to add to apt a new source pointing to buster backports before installing dovecot and other software.

Lastly, this PR also updates relevant docs with instructions on how to enable fts-xapian
and what considerations to take into when deciding to turn on FTS and choose between xapian and solr.

Fixes #2061 

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

## Checklist:

- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)

PR #3403, opened 2023-06-20
general: update base image to Debian 12 ("Bookworm")

georglauterbach:
# Description

Update to Debian 12 ("Bookworm") as the base image and fix/adjust all parts affected by this change.

Superseeds #3402

## Type of change

## Checklist:


docs/content/config/security/rspamd.md
```diff
@@ -10,7 +10,7 @@ If you want to have a look at the default configuration files for Rspamd that DM
 
 !!! note "AMD64 vs ARM64"
 
-    We are currently doing a best-effort installation of Rspamd for ARM64 (from the Debian backports repository for Debian 11). The current version difference as of 23rd Apr 2023: AMD64 is at version 3.5 | ARM64 is at version 3.4.
+    When running DMS on AMD64, we are using a PPA (that does not yet provide a package for ARM64) to get the very latest version. As a result, there is a version discrepancy. The current version difference as of 20th June 2023: AMD64 is at version 3.5 | ARM64 is at version 3.4.
```
polarathene, line 13: Were you able to check the PPA for ARM64 btw? I do recall the maintainer saying they had added ARM64 builds somewhere.

polarathene:
- We will need to wait on https://repo.dovecot.org/#debian getting a Bookworm repo.
- Do we want to switch to Bookworm ASAP, or wait until the first point release? (_[based on Bullseye releases](https://en.wikipedia.org/wiki/Debian_version_history#Debian_11_(Bullseye)), those seem to be every 3 months_)

casperklein:
> * We will need to wait on [repo.dovecot.org/#debian](https://repo.dovecot.org/#debian) getting a Bookworm repo.
> 
>     * Do we want to switch to Bookworm ASAP, or wait until the first point release? (_[based on Bullseye releases](https://en.wikipedia.org/wiki/Debian_version_history#Debian_11_(Bullseye)), those seem to be every 3 months_)

Unless there is a killer feature, there is no need to upgrade ASAP IMO. Waiting for the first point release sounds reasonable to me. Minor bugs in the initial release will be fixed then + Dovecot hopefully provides a bookworm repo until then.

docs/content/config/security/rspamd.md
georglauterbach: I can check whether the official PPA now has arm64 support. Haven't done that yet.

georglauterbach: Checked: **no** arm64 version in the official PPA yet.

georglauterbach:
12.1 has dropped on July 23. Dovecot still has no repo for Bookworm. What should we do?

polarathene:
> What should we do?

Can continue to wait or just delay an image update to another release?

Alternatively migrate to another base image like Fedora was being evaluated in the past. Fedora provides stable releases more frequently with much newer packages. If you're interested in migrating to Fedora I can take a shot at it. I think there may be some subtle differences we implicitly have with Debian (like `/etc/services`) which may need to be accounted for though, so it's best delaying a base image switch to a separate breaking release without too much else going on to simplify troubleshooting if tests don't pick anything up.

georglauterbach:
I'm interested, but as you said, let's do this with a separate release entirely. And then I'd advocate for Alpine TBH.

Nevertheless, I do not want this to be open for too long. Hence, I'd continue with the Debian 11 Buster PPA for Dovecot and merge.

polarathene:
> And then I'd advocate for Alpine TBH.

I don't encourage Alpine. It's primary benefit is the small size and frequent releases (_although one of the most prolific package maintainers recently stepped down IIRC_). Common issues experienced are DNS handling and musl related (_instead of glibc_), of which I've experienced myself.

Unfortunately I don't recall specifics, but I have a vague memory of Alpine being a no-no for DMS, even though at an earlier point I really did want an Alpine base when I joined as a maintainer, something I experienced while working with Alpine at some point changed my opinion on it for DMS.

I definitely know that I contributed to the `testssl.sh` project improvements to their `Dockerfile` and explored changing it's base image, I did this with Alpine, openSUSE Leap and Fedora, moving from their Debian base. I think they had a problem with Alpine with `openssl` but it was a niche scenario and we ended up choosing openSUSE as the base. Not sure if they're still using it today or if they changed back to Debian.

I've also got WIP `Dockerfile` that I didn't get around to finishing / submitting to rspamd maintainers, which was a much better approach than their current CI repo building rspamd to link to various distro base images packages. Both Fedora and openSUSE can get fairly minimal in size, although I think I experimented with DMS and the savings wasn't that big.

At the very least I'd caution the switch to Alpine.

openSUSE vs Fedora I think may vary on available packages, but both do have the newer PostSRSd (v2) IIRC while Debian does not, Leap probably required adding a repo (like PPA). I think for Fedora at least one of the issues was rspamd package, I don't know if that actually improved upstream, or if it'd still require me to finish up the work I was doing there (_or we build rspamd ourselves since I've already done that.._).

This is a low priority for me, I'm not sure when I'll have time for taking a shot at it.

---

> Nevertheless, I do not want this to be open for too long. Hence, I'd continue with the Debian 11 Buster PPA for Dovecot and merge.

👍

georglauterbach:
Okay so using the Debian 11 PPA for Dovecot with Debian 12 as a base image does not work. Nice. Don't have time at the moment to further tackle this. Delayed for now.

georglauterbach:
@polarathene I was annoyed at work today so I gave this PR another shot. Can you get the getmail6 contributor to have a look at the failing test? I removed the custom install, and version is now 6.18.11 I think.

LucidityCrash:
I can't replicate that in some basic testing I did on a spare debian box BUT that machine isn't bookworm it is bullseye ... I installed the getmail version you listed from the bookworm repo and that assert looks good for me .  However looking at the output in that github step it would appear the first line is missing from the output so it would fail ... it says it has output 21 lines - that should be 22 ... you could try changing the tail -n +7 to tail -n +6 on line 18 of target/bin/debug-getmail

polarathene (changes requested):
Bulk of this review is just notes for my benefit, and optional / future changes.

- `/etc/cron.daily/spamassassin` (_maintenance run of `sa-update`_) will fail to run properly with Debian 12:
  - **Requires** `/etc/spamassassin/skip-timer-conversion` (empty file that is checked for existence), **or** the prevention of `systemd` subpackage being installed (by `amavisd-new`), which creates `/run/systemd`.
  - `Dockerfile` should include a comment referencing my PR feedback for details of the `'CRON=1' > /etc/default/spamassassin` + any additional added to resolve the above Debian 12 compatibility concern.
  - Drop [this run-time change to `/etc/cron.daily/spamassassin`](https://github.com/docker-mailserver/docker-mailserver/blob/c8a0bfd36195281b629916d64edf0a2c1fbd89d1/target/scripts/startup/setup.d/security/misc.sh#L82-L85), it's no longer applicable or useful.
- `Dockerfile` should include contextual comment for change to `/usr/lib/rsyslog/rsyslog-rotate`. Easier to grok that it's for our alternative process manager than part of a fix related to the change preceding it.
- **Optional:** `packages.sh` has feedback to consider.
- **Optional:** Changelog notes. You may want to directly update the changelog via the PR? See the `compatibility_level  3.6` comment for a suggestion.
- **Optional:** SA related test-case updates have suggested improvements to consider.

target/scripts/build/packages.sh
```diff
@@ -116,106 +120,24 @@ function _install_dovecot() {
...
-    echo "deb-src ${URL}" >>"${DEB_FILE}"
-    RSPAMD_PACKAGE_NAME='rspamd'
-  fi
-
-  _log 'debug' 'Installing Rspamd'
-  apt-get "${QUIET}" update
-  apt-get "${QUIET}" --no-install-recommends install "${RSPAMD_PACKAGE_NAME}" 'redis-server'
-}
-
-function _install_fail2ban() {
```
casperklein, line 147: I would keep fail2ban related code. The versions are now even, but in the future, they will drift away as happened in the past. For reason (bugfix) or another (missing feature), we then probably reintroduce it. So why not just keep it?

polarathene: I think @georglauterbach motivation was just a desire to trim out what he can if he's planning to continue as the primary maintainer of the project for his benefit? 🤷‍♂️ 

The method won't have as much relevance if moving to a different base image? If that happens it'd likely be a choice where the latest updates are available in faster release cycles. Depends how important it is to pull in the latest updates, but I understand both views here :)

I don't see any harm in leaving this method alone if it's not adding notable maintenance burden, the extra verbosity isn't an issue IMO.

Dockerfile
```diff
@@ -122,7 +120,7 @@ COPY \
 
 # hadolint ignore=SC2016
 RUN <<EOF
-  sedfile -i -r 's/^(CRON)=0/\1=1/g' /etc/default/spamassassin
+  echo 'CRON=1' >/etc/default/spamassassin
```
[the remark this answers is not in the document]
polarathene: **NOTE:** If in future SA will remain when Amavis is dropped, for integration with `rspamd`, apparently it should include an `sa-update` hook similar to the Amavis one to send a reload signal?:

https://github.com/rspamd/rspamd/issues/3562

target/scripts/build/packages.sh
```diff
@@ -116,106 +120,24 @@ function _install_dovecot() {
...
-  local FAIL2BAN_DEB_URL='https://github.com/fail2ban/fail2ban/releases/download/1.0.2/fail2ban_1.0.2-1.upstream1_all.deb'
-  local FAIL2BAN_DEB_ASC_URL="${FAIL2BAN_DEB_URL}.asc"
-  local FAIL2BAN_GPG_FINGERPRINT='8738 559E 26F6 71DF 9E2C  6D9E 683B F1BE BD0A 882C'
-  local FAIL2BAN_GPG_PUBLIC_KEY_ID='0x683BF1BEBD0A882C'
-  local FAIL2BAN_GPG_PUBLIC_KEY_SERVER='hkps://keyserver.ubuntu.com'
-
-  _log 'debug' 'Installing Fail2ban'
-  apt-get "${QUIET}" --no-install-recommends install python3-pyinotify python3-dnspython
```
polarathene, line 155: Almost forgot about this.

You've skipped these two packages, but they're added for a reason to support Fail2Ban better, they're still relevant, check `git blame`.

polarathene:
Review context regarding version changes in Debian 12 packages, plus some related history/notes for my benefit 😅 

Major version updates should be included in changelog.

---

- `fetchmail` no notable change, `6.4` on both Debian 11 and 12.
- `getmail6` (_`6.18` => [`6.18`](https://packages.debian.org/bookworm/mail/getmail6)_)
  - Debian 11 provided 6.14 in Bullseye repo and 6.18 in Bullseye-backports repo, but we were installing `getmail6` via the python package manager `pip`. Switching to Debian package with Debian 12.
- `openssl` (_[`1.1.1n`](https://packages.debian.org/bullseye/openssl) => [`3.0.9`](https://packages.debian.org/bookworm/openssl)_)
  - Tests caught a slight difference in some output. I haven't looked over the changelog for if this major release impacts anything else.
- `amavisd-new` (_[`2.11`](https://packages.debian.org/bullseye/amavisd-new) => [`2.13`](https://packages.debian.org/bookworm/mail/amavisd-new)_)
  - Adds dep of `systemd | systemd-standalone-sysusers | systemd-sysusers`. Installing [`systemd-standalone-sysusers`](https://packages.debian.org/bookworm/systemd-standalone-sysusers) is an appropriate non-default choice for our non-systemd requirement, **we should add that dependency explicitly**.
  - **NOTE:** Instead of `check_policy_service` integration to Postfix, [`amavisd-milter`](https://packages.debian.org/bookworm/mail/amavisd-milter) package could be used instead.
  - Suggested deps (_which DMS explicitly installs_):
    - `clamav` (_`0.103.8` => [`1.0.1`](https://packages.debian.org/bookworm/clamav)_)
    - `spamassassin` (_`3.4.6` => [`4.0.0`](https://packages.debian.org/bookworm/mail/spamassassin)_)
- `postfix` (_[`3.5.18`](https://packages.debian.org/bullseye/postfix) => [`3.7.6`](https://packages.debian.org/bookworm/mail/postfix)_)
- `redis-server` (_`6.0` => [`7.0`](https://packages.debian.org/bookworm/redis-server)_)
- `rsyslog` (_`8.2102` => [`8.2302`](https://packages.debian.org/bookworm/rsyslog)_)
- `supervisor` (_`4.2.2` => [`4.2.5`](https://packages.debian.org/bookworm/supervisor)_)
- `dovecot-core` (_`2.3.13` => [`2.3.19`](https://packages.debian.org/bookworm/dovecot-core)_)
  - The current default in DMS v12.1 is a community edition install that is presently `2.3.20`.
  - Debian 12 is expected to get `2.3.20` or when released `2.4.0` (_Dovecot devs [plan to release in 2023H2](https://www.mail-archive.com/dovecot@dovecot.org/msg87796.html)_) [via the bookworm backports repo later this year](https://github.com/docker-mailserver/docker-mailserver/pull/3403#issuecomment-1693657448).
- `dovecot-fts-xapian` (_`1.4.9` => [`1.5.5`](https://packages.debian.org/bookworm/dovecot-fts-xapian)_)
  - `libxapian30` dep `>= 1.4.19` + `dovecot-abi-2.3.abiv19` virtual dep (_provided via `dovecot-core`_)
  - Package previously broke compatibility with Debian 11 Bullseye packages when mixed with third-party Dovecot repo for a newer `dovecot-core` which could no longer satisfy the ABI virtual dependency. Solution was to [build package `dovecot-fts-xapian 1.5.5` from source](https://github.com/docker-mailserver/docker-mailserver/pull/3373) via introducing `build/compile.sh`. Originally [integrated into DMS in July 2021](https://github.com/docker-mailserver/docker-mailserver/pull/2064) via single package and a documented config guide.
  - When `dovecot-core 2.4.0` is available, [migrating to the `flatcurve` plugin becomes a viable option](https://github.com/docker-mailserver/docker-mailserver/issues/3247#issuecomment-1504247016) and may instead be a better replacement for this feature.
  - Additionally, the suggestion to migrate to [`flatcurve`](https://github.com/slusarz/dovecot-fts-flatcurve/) plugin was proposed by the contributor of support for [`xapian` plugin](https://github.com/grosjo/fts-xapian) in DMS, which was due to the existing [`solr` plugin having notable drawbacks](https://github.com/docker-mailserver/docker-mailserver/issues/3247#issuecomment-1504290556).
- `postsrsd` (_[`1.10`](https://packages.debian.org/bookworm/postsrsd) - no notable change from Debian 11_)
  - There is a bug within common container environments that have `containerd` or similar service on the host configured with `LimitNOFILE=infinity` where `fs.nr_open` is a value of `2^30` instead of `2^20` (Debian). That bug was fixed in Aug 2022 but only available with PostSRSd v1.12 / v2. Thus users of this feature in DMS may need to explicitly configure the `--ulimit` to `1024:524288` (_systemd implicit default since v240_). Also requires at least `glibc` of `2.35`, which [Debian 12 provides with `2.36`](https://packages.debian.org/bookworm/libc6). Host needs kernel `>= 5.9`.
- `postfix-policyd-spf-python` (_[`2.9.2`](https://packages.debian.org/bullseye/postfix-policyd-spf-python) => [`3.04`](https://packages.debian.org/bookworm/mail/postfix-policyd-spf-python)_)
  - Dropped the `python3-spf` dep, supposedly not needed anymore to provide Postfix with an SPF `check_policy_service`.
  - **NOTE:** There also exists an alternative milter-based package [`pyspf-milter`](https://packages.debian.org/bookworm/mail/pyspf-milter) which appears to be part of `python3-spf-engine` too (_available in [`spf-engine` since `2.9.0`](https://git.launchpad.net/spf-engine/tree/README.txt)_). See the [`spf-engine` README for advice](https://git.launchpad.net/spf-engine/tree/README.txt) on configuring Postfix with milter package.
  - `spf-engine` [upstream source](https://launchpad.net/spf-engine) and [Debian source](https://salsa.debian.org/python-team/packages/spf-engine).
  - Some resources online provide nicer docs to browse. [`policyd-spf` Postfix integration advice](https://www.mankier.com/1/policyd-spf#Postfix_Integration) (_states version 2.0, but this seems to match the [`python-policy-spf 2.0.0` mention in README source](https://salsa.debian.org/python-team/packages/spf-engine) ([upstream](https://git.launchpad.net/spf-engine/tree/README.txt)_) notes that when using the `check_policy_service` it [must come after `reject_unauth_destination`](https://salsa.debian.org/python-team/packages/spf-engine/-/blob/d06300cafb1f004576ab1bfb2e5eb343399bd148/debian/README.Debian#L22-23) to avoid creating an open relay (_DMS does this at present, but it's not clearly documented_). Similar [`pyspf-milter` docs](https://www.mankier.com/8/pyspf-milter).

georglauterbach:
I will come back to this PR. @polarathene thank you for your elaborate review; I will make sure to go all of it as soon as I have time again. Probably October.

polarathene:
> I will make sure to go all of it as soon as I have time again. Probably October.

Sure 👍 

I have a TODO list for DMS with over 30 items to wade through, but I may find time in-between to address the concerns/changes I raised in my review above if you're open to that? Not all of my TODO items are easy to resolve, nor suitable for v13, but since it's a breaking release I'll try squeeze in some that could benefit.

polarathene:
Are we agreeing to wait for Dovecot 2.4 (or `2.3.20`) landing in Debian 12 backports before updating to Debian 12? Supposedly that'll be available before 2024.

My understanding is that until that's available, we'd have to roll-back down to Dovecot `2.3.19` otherwise?

casperklein:
> My understanding is that until that's available, we'd have to roll-back down to Dovecot `2.3.19` otherwise?

Correct.

casperklein:
In doubt, we should stick with the community repo, as updates will appear there first.

georglauterbach:
So, I just got my master's degree from Technische Universität Dresden 🎉🎉🎉 Which means I will now have a bit more time again (after my vacation) 😄 🚀

polarathene:
> Which means I will now have a bit more time again (after my vacation) 😄 🚀

Grats on getting your masters! 💪 

Enjoy your vacation, chances are Dovecot backport may still not be ready for Debian 12 😛 

I'd still like to try get a few more things done for v13, and then the image update can be deferred to a future major release.

polarathene:
Was the recent force push just a rebase? No other changes?

georglauterbach:
Just a rebase

georglauterbach:
We should think about not making this a blocker for v13. v13 already has a lot of changes, and we can always just release a new major version. I'd like to see a new release so users of our stable release channel get the latest updates.

target/scripts/build/packages.sh
```diff
@@ -18,7 +21,14 @@ function _pre_installation_steps() {
   apt-get "${QUIET}" update
 
   _log 'trace' 'Installing packages that are needed early'
-  apt-get "${QUIET}" install --no-install-recommends apt-utils 2>/dev/null
+  apt-get "${QUIET}" install --no-install-recommends apt-utils ca-certificates curl gnupg 2>/dev/null
```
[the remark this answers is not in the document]
georglauterbach: I will add a comment.

target/scripts/build/packages.sh
```diff
@@ -116,106 +120,24 @@ function _install_dovecot() {
...
-  _log 'debug' 'Deleting default logwatch cronjob'
-  rm /etc/cron.daily/00logwatch
-}
-
 function _post_installation_steps() {
   _log 'debug' 'Running post-installation steps (cleanup)'
+  _log 'trace' 'Deleting sensitive files (secrets)'
+  rm /etc/postsrsd.secret
+  _log 'trace' 'Deleting default logwatch cronjob'
+  rm /etc/cron.daily/00logwatch
```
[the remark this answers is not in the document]
georglauterbach: You're right. I will re-order all of this.

test/tests/parallel/set1/spam_virus/amavis.bats
```diff
@@ -51,10 +51,10 @@ function teardown_file() {
   export CONTAINER_NAME=${CONTAINER1_NAME}
 
   # give Amavis just a bit of time to print out its full debug log
-  run _repeat_in_container_until_success_or_timeout 5 "${CONTAINER_NAME}" grep 'ANTI-SPAM-SA' /var/log/mail/mail.log
+  run _repeat_in_container_until_success_or_timeout 5 "${CONTAINER_NAME}" grep -E 'SA version: [0-9]\.[0-9]\.[0-9]' /var/log/mail/mail.log
+  assert_success
+  _run_in_container grep 'scanner SpamAssassin' /var/log/mail/mail.log
```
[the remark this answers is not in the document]
georglauterbach: Will commit this.

test/tests/parallel/set1/spam_virus/disabled_clamav_spamassassin.bats
```diff
@@ -29,8 +29,10 @@ function teardown_file() { _default_teardown ; }
 }
 
 @test "SA - Amavis integration should not be active" {
-  _run_in_container_bash "grep -i 'ANTI-SPAM-SA code' /var/log/mail/mail.log | grep 'NOT loaded'"
-  assert_success
+  _run_in_container grep -E 'SA version: [0-9]\.[0-9]\.[0-9]' /var/log/mail/mail.log
+  assert_failure
+  _run_in_container grep 'scanner SpamAssassin' /var/log/mail/mail.log
+  assert_failure
```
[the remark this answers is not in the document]
georglauterbach: Will commit this manually

target/scripts/build/packages.sh
georglauterbach: I would really like to get rid of this code, actually all the custom code that clutters `packages.sh` as of now. The same goes for Rspamd, where I also removed a lot of the custom code. It makes the whole script more verbose and harder to maintain. Using PPAs is fine as the package name stays consistent. From a maintenance PoV, removing the custom F2B code is very desirable.

target/scripts/build/packages.sh
georglauterbach: I will add this back.

polarathene:
I will need to delay review on this until later this week.

target/scripts/build/packages.sh
```diff
@@ -111,7 +146,7 @@ function _install_dovecot() {
     _log 'trace' 'Using Dovecot community repository'
     curl https://repo.dovecot.org/DOVECOT-REPO-GPG | gpg --import
     gpg --export ED409DA1 > /etc/apt/trusted.gpg.d/dovecot.gpg
-    echo "deb https://repo.dovecot.org/ce-2.3-latest/debian/bullseye bullseye main" > /etc/apt/sources.list.d/dovecot.list
+    echo "deb https://repo.dovecot.org/ce-2.3-latest/debian/${VERSION_CODENAME} ${VERSION_CODENAME} main" > /etc/apt/sources.list.d/dovecot.list
```
polarathene, line 149:
```suggestion
    # VERSION_CODENAME sourced from /etc/os-release
    echo "deb https://repo.dovecot.org/ce-2.3-latest/debian/${VERSION_CODENAME} ${VERSION_CODENAME} main" > /etc/apt/sources.list.d/dovecot.list
```

CHANGELOG.md
```diff
@@ -19,8 +33,14 @@ All notable changes to this project will be documented in this file. The format
 
 - **Documentation:**
   - Raise awareness in the troubleshooting page for a common misconfiguration when deviating from our advice by using a bare domain ([#3680](https://github.com/docker-mailserver/docker-mailserver/pull/3680))
+  - updated note about Rspamd on ARM64 lacking behind in version
```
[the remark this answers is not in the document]
polarathene: https://github.com/docker-mailserver/docker-mailserver/issues/3374#issuecomment-1844793514

> By now there are ARM64 builds in the official rspamd debian repository. So it may be installed from this source instead of debian backports.

Recently merged PR that removed the docs addition, these changes were for v13.1: https://github.com/docker-mailserver/docker-mailserver/pull/3686

The `package.sh` also reflects that update, so this changelog entry (with no PR) might as well be dropped?

```suggestion
```

polarathene (changes requested):
I've reviewed `CHANGELOG.md` and `packages.sh`, LGTM 👍 

I unfortunately don't have time to go over prior review feedback concerns I raised, so trust that anything of value from that is already applied.

I'll apply my change suggestions.

Might be good to still push out v13.1, but I understand if you want to merge this and go straight to a v14 release.

CHANGELOG.md
```diff
@@ -21,6 +38,11 @@ All notable changes to this project will be documented in this file. The format
   - Raise awareness in the troubleshooting page for a common misconfiguration when deviating from our advice by using a bare domain ([#3680](https://github.com/docker-mailserver/docker-mailserver/pull/3680))
 - **Internal:**
   - Postfix configures `virtual_mailbox_maps` and `virtual_transport` during startup instead of using defaults (configured for Dovecot) via our `main.cf` ([#3681](https://github.com/docker-mailserver/docker-mailserver/pull/3681))
+  - Updates to `packages.sh`:
+    - The script now uses `/etc/os-release` to determine the release name of Debian
+    - Removed custom installations of Fail2Ban, getmail6 and Rspamd
+    - Updated packages lists and added comments for maintainability
+    - Use official Rspamd PPA on AMD64; no PPA is used on ARM64
```
polarathene, line 45: This line isn't accurate anymore? Both ARM64 and AMD64 are used?

```suggestion
    - Use official Rspamd PPA
```

polarathene (approved):

CHANGELOG.md
```diff
@@ -6,6 +6,23 @@ All notable changes to this project will be documented in this file. The format
...
 
+The most noteworthy change of this release is the update of the container's base image from Debian 11 ("Bullseye") to Debian 12 ("Bookworm"). This update alone involves breaking changes and requires a careful update!
+
+### Breaking
+
+- **Updated base image to Debian 12**
+- Changed the default of `DOVECOT_COMMUNITY_REPO` to `0` (disabled) - the Dovecot community repo will (for now) not be the default when building the DMS.
+- Updated packages. For an overview, [we have a review comment on the PR that introduces Debian 12](https://github.com/docker-mailserver/docker-mailserver/pull/3403#issuecomment-1694563615)
+  - Notable major version bump: `openssl 3`, `clamav 1`, `spamassassin 4`, `redis-server 7`.
+  - Notable minor version bump: `postfix 3.5.18 => 3.7.6`
```
polarathene, line 17: rspamd minor version probably got bumped too, or at least consistent if you want to include that here in the changelog for visibility.

CHANGELOG.md
georglauterbach:
```suggestion
    - Use official Rspamd PPA for AMD64 and ARM64
```

CHANGELOG.md
georglauterbach: I think we already shop 3.7.5 on our `:edge` builds TBH, which is AFAIK the very latest version of Rspamd.

target/scripts/build/packages.sh
georglauterbach: PS: The same goes for here: https://github.com/docker-mailserver/docker-mailserver/pull/3403/files#diff-95c6b88025d3f0c1a18f85c2c05a14faa73ff59b57f6e938df5113221c8018f1R36

CHANGELOG.md
polarathene: > Rspamd will not see a version bump, so no note necessary.

~~Won't this be the first release where rspamd has the same version on AMD64 and ARM64?~~ (_**EDIT:** Nope that'd be 13.1 now_)

The [v13.1 PR removes from rspamd docs](https://github.com/docker-mailserver/docker-mailserver/pull/3686/files#diff-d6195d723286b5393e85a1f0f2127d1f72a1c814387f1f1b3f10b9eee85f98ddL13) stating rspamd 3.5 (AMD64) and 3.4 (ARM64).

Granted, since you've chosen to continue with a v13.1 release there won't be a version change to announce here, but it's [not mentioned explicitly in that PR changelog entry either](https://github.com/docker-mailserver/docker-mailserver/pull/3686/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR28).

polarathene (dismissed):

CHANGELOG.md
georglauterbach: We can add this to the release PR of the v13.1.0 release.

georglauterbach:
@casperklein Are you okay with merging this after v13.1.0 has been released (so we have the changes on `:edge`)?

casperklein:
Sure, I'll try to review asap. This might however take some days.

polarathene:
I ran out of time to get to this PR. The OAuth2 PR needs another revision, feel free to go for a v13.2 release, and follow up with or go straight to a v14 release here.

If you go with v13.2, I can resolve conflicts here tomorrow if that still needs to be done 😅

georglauterbach:
I will release v13.2; you do not need to take care of conflicts here, I'll do that. This PR is not important for v13.2 anyway.

polarathene:
> Do we need the special release freeze?

I don't think so? It's usually meant for giving `:edge` users some time to report any bugs which is useful when we've churned through quite a few changes.

The diff for 13.2 would be fairly small and users can still revert back to 13.1 if we need to push out a 13.2.1 (_so you may want to defer merging this PR for a few days_).

georglauterbach:
> Sure, I'll try to review asap. This might however take some days.

Do you have an ETA on the review?:) 🙈

casperklein:
Sorry for the delay, there was some unexpected stuff I needed to care about. I promise to finish review asap. Personal deadline is next sunday. Hopefully earlier ✌️ I hope that's fine.

georglauterbach:
Alright, nice!

Just FYI: I think we will release v13.3 in the meantime (see https://github.com/docker-mailserver/docker-mailserver/pull/3480#issuecomment-1891156792).

CHANGELOG.md
```diff
@@ -86,8 +106,10 @@ DMS is now secured against the [recently published spoofing attack "SMTP Smuggli
   - Debugging - Raise awareness of temporary downtime during certificate renewal that can cause a failure to deliver local mail ([#3718](https://github.com/docker-mailserver/docker-mailserver/pull/3718))
 - **Internal:**
   - Postfix configures `virtual_mailbox_maps` and `virtual_transport` during startup instead of using defaults (configured for Dovecot) via our `main.cf` ([#3681](https://github.com/docker-mailserver/docker-mailserver/pull/3681))
-- **Rspamd:**
-  - Upgraded to version `3.7.5`. This was previously inconsistent between our AMD64 (`3.5`) and ARM64 (`3.4`) images ([#3686](https://github.com/docker-mailserver/docker-mailserver/pull/3686))
+  - Updates to `packages.sh`:
+    - The script now uses `/etc/os-release` to determine the release name of Debian
+    - Removed custom installations of Fail2Ban, getmail6 and Rspamd
+    - Updated packages lists and added comments for maintainability
```
polarathene, line 112: Note this changelog addition incorrectly handled resolving the merge conflict / rebase?

Shouldn't be swapping out changelog entries from prior releases :P

CHANGELOG.md
```diff
@@ -34,7 +55,6 @@ All notable changes to this project will be documented in this file. The format
 - **Docs:**
   - Revised the SpamAssassin ENV docs to better communicate configuration and their relation to other ENV settings. ([#3756](https://github.com/docker-mailserver/docker-mailserver/pull/3756))
 
-
```
polarathene, line 37: Likewise, not something this PR should be modifying.

CHANGELOG.md
```diff
@@ -2,10 +2,31 @@
...
+
+The most noteworthy change of this release is the update of the container's base image from Debian 11 ("Bullseye") to Debian 12 ("Bookworm"). This update alone involves breaking changes and requires a careful update!
+
+### Breaking
+
+- **Updated base image to Debian 12**
+- Changed the default of `DOVECOT_COMMUNITY_REPO` to `0` (disabled) - the Dovecot community repo will (for now) not be the default when building the DMS.
+- Updated packages. For an overview, [we have a review comment on the PR that introduces Debian 12](https://github.com/docker-mailserver/docker-mailserver/pull/3403#issuecomment-1694563615)
+  - Notable major version bump: `openssl 3`, `clamav 1`, `spamassassin 4`, `redis-server 7`.
+  - Notable minor version bump: `postfix 3.5.18 => 3.7.6`
```
polarathene, line 19: Grouping these under the same base image update change, which now also references the PR that introduces it.

Postfix patch versions bumped too, based on the current [`-updates` repos for Bullseye and Bookworm](https://packages.debian.org/search?searchon=sourcenames&keywords=postfix).

```suggestion
- **Updated base image to Debian 12** ([#3403](https://github.com/docker-mailserver/docker-mailserver/pull/3403))
  - Changed the default of `DOVECOT_COMMUNITY_REPO` to `0` (disabled) - the Dovecot community repo will (for now) not be the default when building the DMS.
  - Updated packages. For an overview, [we have a review comment on the PR that introduces Debian 12](https://github.com/docker-mailserver/docker-mailserver/pull/3403#issuecomment-1694563615)
    - Notable major version bump: `openssl 3`, `clamav 1`, `spamassassin 4`, `redis-server 7`.
    - Notable minor version bump: `postfix 3.5.23 => 3.7.9`
```

CHANGELOG.md
```diff
@@ -2,10 +2,31 @@
...
+- **Postfix:**
+  - `compatibility_level` was raised from `2` to `3.6` ([#3403](https://github.com/docker-mailserver/docker-mailserver/pull/3403))
+    - Postfix has deprecated the usage of `whitelist` / `blacklist` in config parameters and logging in favor of `allowlist` / `denylist` and similar variations. ([#3403](https://github.com/docker-mailserver/docker-mailserver/pull/3403/files#r1306356328))
+      - This [may affect monitoring / analysis of logs output from Postfix](https://www.postfix.org/COMPATIBILITY_README.html#respectful_logging) that expects to match patterns on the prior terminology used.
+      - DMS `main.cf` has renamed `postscreen_dnsbl_whitelist_threshold` to `postscreen_dnsbl_allowlist_threshold` as part of this change.
+    - `smtpd_relay_restrictions` (relay policy) is now evaluated after `smtpd_recipient_restrictions` (spam policy). Previously it was evaluated before `smtpd_recipient_restrictions`. Mail to be relayed via DMS must now pass through the spam policy first.
+    - The TLS fingerprint policy has changed the default from MD5 to SHA256 (_DMS does not modify this Postfix parameter, but may affect any user customizations that do_).
```
polarathene, line 26: Likewise, all these postfix changes are bundled under the package upgrade. Adjusted accordingly:

```suggestion
- **Postfix:**
  - Postfix upgrade from 3.5 to 3.7 ([#3403](https://github.com/docker-mailserver/docker-mailserver/pull/3403))
    - `compatibility_level` was raised from `2` to `3.6`
    - Postfix has deprecated the usage of `whitelist` / `blacklist` in config parameters and logging in favor of `allowlist` / `denylist` and similar variations. ([#3403](https://github.com/docker-mailserver/docker-mailserver/pull/3403/files#r1306356328))
      - This [may affect monitoring / analysis of logs output from Postfix](https://www.postfix.org/COMPATIBILITY_README.html#respectful_logging) that expects to match patterns on the prior terminology used.
      - DMS `main.cf` has renamed `postscreen_dnsbl_whitelist_threshold` to `postscreen_dnsbl_allowlist_threshold` as part of this change.
    - `smtpd_relay_restrictions` (relay policy) is now evaluated after `smtpd_recipient_restrictions` (spam policy). Previously it was evaluated before `smtpd_recipient_restrictions`. Mail to be relayed via DMS must now pass through the spam policy first.
    - The TLS fingerprint policy has changed the default from MD5 to SHA256 (_DMS does not modify this Postfix parameter, but may affect any user customizations that do_).
```

target/scripts/build/packages.sh
```diff
@@ -1,10 +1,13 @@
 #!/bin/bash
 
-# -eE :: exit on error (do this in functions as well)
-# -u  :: show (and exit) when using unset variables
+# -eE         :: exit on error (do this in functions as well)
```
polarathene, line 3: Some of the changes in `packages.sh` here could probably be split out to a separate PR as it's not all related to the image upgrade.

The refactoring probably has been causing various conflicts to resolve too. Not a major concern anymore since this PR will finally get merged soon after final review :)

polarathene (changes requested):
Some small adjustments regarding Changelog

CHANGELOG.md
georglauterbach: You're right, but for this PR, not a problem now IMO.

CHANGELOG.md
georglauterbach: Was this just the line

```txt
    - Updated packages lists and added comments for maintainability
```

?

target/scripts/build/packages.sh
georglauterbach: Yeah, somehow typical for my PRs haha. Let's merge this PR anyway; reverting it now is something I can't affort time-wise.

polarathene (approved):

test/tests/parallel/set1/spam_virus/postgrey_enabled.bats
```diff
@@ -60,6 +60,10 @@ function teardown_file() { _default_teardown ; }
     'action=greylist' \
     'reason=new' \
     'client_address=127.0.0.1/32, sender=user@external.tld, recipient=user1@localhost.localdomain'
+
+  _repeat_until_success_or_timeout 10 _run_in_container grep \
+    'Recipient address rejected: Delayed by Postgrey' \
+    /var/log/mail/mail.log
```
polarathene, line 66: This addition looks like it was shifted up during your refactor for adding swaks.

The call directly above is failing at line 60. This is because the client address value needs to be `127.0.0.1` instead of `127.0.0.1/32` like the other related changes to this file below.

polarathene (changes requested):

test/tests/parallel/set1/spam_virus/postgrey_enabled.bats
georglauterbach: Done in 68a67c95e2a5b7ecc588a4c2acdd25f64d3e17ff

CHANGELOG.md
georglauterbach: Hopefully completely resolved in 7cdf318d4dbfa975c6d63eff434dbf8241ed048b

CHANGELOG.md
```diff
@@ -2,10 +2,36 @@
...
+
+The most noteworthy change of this release is the update of the container's base image from Debian 11 ("Bullseye") to Debian 12 ("Bookworm"). This update alone involves breaking changes and requires a careful update!
+
+### Breaking
+
+- **Updated base image to Debian 12** ([#3403](https://github.com/docker-mailserver/docker-mailserver/pull/3403))
+  - Changed the default of `DOVECOT_COMMUNITY_REPO` to `0` (disabled) - the Dovecot community repo will (for now) not be the default when building the DMS.
+  - Updated packages. For an overview, [we have a review comment on the PR that introduces Debian 12](https://github.com/docker-mailserver/docker-mailserver/pull/3403#issuecomment-1694563615)
+    - Notable major version bump: `openssl 3`, `clamav 1`, `spamassassin 4`, `redis-server 7`.
+    - Notable minor version bump: `postfix 3.5.23 => 3.7.9`
```
polarathene, line 19: Probably should highlight this downgrade due to community repo not being used.

```suggestion
    - Notable minor version bump: `postfix 3.5.23 => 3.7.9`
    - Notable minor version bump + downgrade: `dovecot 2.3.13 => 2.3.19` (_Previous release provided `2.3.21` via community repo, `2.3.19` is now the default_)
```

CHANGELOG.md
```diff
@@ -2,10 +2,36 @@
...
 > **Note**: Changes and additions listed here are contained in the `:edge` image tag. These changes may not be as stable as released changes.
 
+## [v14.0.0](https://github.com/docker-mailserver/docker-mailserver/releases/tag/v14.0.0)
+
+The most noteworthy change of this release is the update of the container's base image from Debian 11 ("Bullseye") to Debian 12 ("Bookworm"). This update alone involves breaking changes and requires a careful update!
+
+### Breaking
+
+- **Updated base image to Debian 12** ([#3403](https://github.com/docker-mailserver/docker-mailserver/pull/3403))
+  - Changed the default of `DOVECOT_COMMUNITY_REPO` to `0` (disabled) - the Dovecot community repo will (for now) not be the default when building the DMS.
```
polarathene, line 16: Probably worth highlighting this, although it probably won't break anything for our users.
- OAuth2 support just landed for v13.3 and neither of these fixes are related to the documented config.
- Lua support is only due to community contributed docs support.

```suggestion
  - Changed the default of `DOVECOT_COMMUNITY_REPO` to `0` (disabled) - the Dovecot community repo will (for now) not be the default when building the DMS.
    - While Debian 12 (Bookworm) was released in June 2023 and the latest Dovecot `2.3.21` in Sep 2023, as of Jan 2024 there is no [Dovecot community repo available for Debian 12](https://repo.dovecot.org).
    - This results in the Dovecot version being downgraded from `2.3.21` (DMS v13.3) to `2.3.19`, which [may affect functionality when you've explicitly configured for these features](https://github.com/dovecot/core/blob/30cde20f63650d8dcc4c7ad45418986f03159946/NEWS#L1-L158):
      - OAuth2 (_mostly regarding JWT usage, or POST requests (`introspection_mode = post`) with `client_id` + `client_secret`_).
      - Lua HTTP client (_DNS related_).
```

target/scripts/build/packages.sh
```diff
@@ -111,7 +154,8 @@ function _install_dovecot() {
     _log 'trace' 'Using Dovecot community repository'
     curl https://repo.dovecot.org/DOVECOT-REPO-GPG | gpg --import
     gpg --export ED409DA1 > /etc/apt/trusted.gpg.d/dovecot.gpg
-    echo "deb https://repo.dovecot.org/ce-2.3-latest/debian/bullseye bullseye main" > /etc/apt/sources.list.d/dovecot.list
+    # VERSION_CODENAME sourced from /etc/os-release
+    echo "deb https://repo.dovecot.org/ce-2.3-latest/debian/${VERSION_CODENAME} ${VERSION_CODENAME} main" > /etc/apt/sources.list.d/dovecot.list
```
polarathene, line 158: Just to clarify, this won't actually work anymore right? As Dovecot still does not provide a Debian Bookworm repo AFAIK?

I assume the next release is 2.4.0 / 3.0 from what I can tell. Perhaps they're trying to get that out the door first. Especially for a 3.x release that would probably make sense that they want to avoid the 2.x => 3.x update being implicit (_so prior Debian release repos would be stuck on the 2.x series_).

polarathene (changes requested):
Probably worth documenting this better :)

target/scripts/build/packages.sh
georglauterbach: As the Dovecot devs still have not bothered to update their community repo to include Debian 12, no, this won't work as of now. Personally I think this is a huge pain point; I guess it is what it is.

polarathene (dismissed):

target/scripts/build/packages.sh
paschun: Could try poking them on the email list? https://dovecot.org/mailman3/archives/list/dovecot@dovecot.org/

georglauterbach: That would be an option @paschun, thank you for noting this :) But truth be told, I do not have the time to do this, especially when considering that we'd then need to do this for all dependencies (to remain fair). Moreover, I'd need to clarify their ludicrous decision with 2.4 for community and 3 for "pro" first... In the meantime, staying with the Debian upstream is not too bad :)

georglauterbach:
THIS PR is now free to merge; I will resolve the conflicts in `CHANGELOG.md` and when @casperklein has approved, we can merge immediately.

target/scripts/build/packages.sh
casperklein: BTW: `-E` is not "exit on error in functions". `-e` also applies to functions.

```-E      If set, any trap on ERR is inherited by shell functions, command substitutions, and commands executed in a sub‐ shell environment. The ERR trap is normally not inher‐ ited in such cases.```

For the future, we could consider to add `shopt -s inherit_errexit` with applies `-e` to subshells.

File: target/scripts/build/packages.sh [hunk 22]
```diff
@@ -1,10 +1,13 @@
...
 
-# -eE :: exit on error (do this in functions as well)
-# -u  :: show (and exit) when using unset variables
+# -eE         :: exit on error (do this in functions as well)
+# -u          :: show (and exit) when using unset variables
 # -o pipefail :: exit on error in pipes
 set -eE -u -o pipefail
 
+# shellcheck source=/dev/null
+source /etc/os-release
```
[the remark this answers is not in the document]
casperklein: I am also a fan of just include what's needed instead of including all. This seems easy to do at first, but could lead to problems anytime later, when var names otherwise used are overlapping.

target/scripts/startup/setup.d/security/misc.sh
```diff
@@ -81,11 +81,6 @@ function __setup__security__spamassassin() {
     # shellcheck disable=SC2016
     sed -i -r 's|^\$sa_kill_level_deflt (.*);|\$sa_kill_level_deflt = '"${SA_KILL}"';|g' /etc/amavis/conf.d/20-debian_defaults
 
-    # fix cron.daily for spamassassin
-    sed -i \
-      's|invoke-rc.d spamassassin reload|/etc/init\.d/spamassassin reload|g' \
-      /etc/cron.daily/spamassassin
-
```
casperklein, line 88: Why was this fix removed?

If I understand correctly, the logic 

```
# Reload
if which invoke-rc.d >/dev/null 2>&1; then
    invoke-rc.d --quiet spamd status > /dev/null && \
      invoke-rc.d spamd reload > /dev/null
else
    /etc/init.d/spamd reload > /dev/null
fi
if [ -d /etc/spamassassin/sa-update-hooks.d ]; then
    run-parts --lsbsysinit /etc/spamassassin/sa-update-hooks.d
fi
```

moved to `/usr/sbin/spamassassin-maint` and should be fixed there now?

Suggestion:

```bash
# fix cron.daily for spamassassin
sed -i \
  's|invoke-rc.d spamd reload|/etc/init.d/spamd reload|g' \
  /usr/sbin/spamassassin-maint
```

casperklein:
Almost done with reviewing. I will do the last 10% tonight.

target/scripts/build/packages.sh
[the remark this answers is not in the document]
georglauterbach: The we could use `VERSION_CODENAME` instead to at least align with `/etc/os-release`.

target/scripts/startup/setup.d/security/misc.sh
georglauterbach: Good catch!

target/scripts/build/packages.sh
```diff
@@ -18,16 +21,44 @@ function _pre_installation_steps() {
...
+    systemd-standalone-sysusers # avoid problems with SA / Amavis (https://github.com/docker-mailserver/docker-mailserver/pull/3403#pullrequestreview-1596689953)
+  )
+  apt-get "${QUIET}" install --no-install-recommends "${EARLY_PACKAGES[@]}" 2>/dev/null
+
+  _log 'trace' 'Adding Rspamd PPA'
+  curl -sSfL https://rspamd.com/apt-stable/gpg.key | gpg --dearmor >/etc/apt/trusted.gpg.d/rspamd.gpg
+  echo "deb [signed-by=/etc/apt/trusted.gpg.d/rspamd.gpg] http://rspamd.com/apt-stable/ ${VERSION_CODENAME} main" >/etc/apt/sources.list.d/rspamd.list
+
+  _log 'trace' 'Updating package index after adding PPAs'
+  apt-get "${QUIET}" update
```
casperklein, line 39: Why did you move the rspam part here?

IMO the old function `_install_rspamd()` was better. All rspamd stuff in one function + that maked it clear, that rspam is not installed from the debian repo.

target/scripts/build/packages.sh
```diff
@@ -18,16 +21,44 @@ function _pre_installation_steps() {
...
 
   _log 'trace' 'Upgrading packages'
   apt-get "${QUIET}" upgrade
 }
 
+function _install_utils() {
+  _log 'debug' 'Installing utils sourced from Github'
+  _log 'trace' 'Installing jaq'
+  curl -sL "https://github.com/01mf02/jaq/releases/latest/download/jaq-v1.2.0-$(uname -m)-unknown-linux-gnu" -o /usr/bin/jaq && chmod +x /usr/bin/jaq
```
casperklein, line 48:
```suggestion
  curl -sSfL "https://github.com/01mf02/jaq/releases/latest/download/jaq-v1.2.0-$(uname -m)-unknown-linux-gnu" -o /usr/bin/jaq && chmod +x /usr/bin/jaq
```

Align with curl statement below

target/scripts/build/packages.sh
polarathene: @casperklein what would be a better wording for the doc comment then? :)

target/scripts/build/packages.sh
```diff
@@ -127,105 +171,23 @@ function _install_dovecot() {
...
-  rm -r "${SWAKS_RELEASE}"
-}
-
-function _remove_data_after_package_installations() {
-  _log 'debug' 'Deleting sensitive files (secrets)'
-  rm /etc/postsrsd.secret
-
-  _log 'debug' 'Deleting default logwatch cronjob'
-  rm /etc/cron.daily/00logwatch
-}
```
casperklein, line 212: Why are those rm statements obsolet?

target/scripts/startup/setup.d/security/misc.sh
polarathene: See my [earlier review feedback](https://github.com/docker-mailserver/docker-mailserver/pull/3403#pullrequestreview-1596689953), specifically the last part of this screenshot:

image


https://github.com/docker-mailserver/docker-mailserver/blob/c8a0bfd36195281b629916d64edf0a2c1fbd89d1/target/scripts/startup/setup.d/security/misc.sh#L82-L85

Verbose review comment related to those screenshot suggestions [can be found here](https://github.com/docker-mailserver/docker-mailserver/pull/3403#discussion_r1306282387). Here is the relevant context from that comment:

image

casperklein:
I have finished the review. Looks overall good to me, just a few questions/comments.

target/scripts/build/packages.sh
polarathene: This was [raised as a concern during my initial review](https://github.com/docker-mailserver/docker-mailserver/pull/3403#discussion_r1306372918) as well with @georglauterbach justifying it:

> _I believe the new spot to be more appropriate as the custom Rspamd installation function has been removed._

I too prefer keeping features better isolated and see no need to bundle them all into the same package install, although we could probably do that if we export an array variable that can be appended to by separate feature functions? 🤔

target/scripts/build/packages.sh
casperklein: `-E      inherit the ERR trap to functions, command substitutions and sub-shells`

Off-Topic: I've updated docker to version 25.0.0 and noticed, that the init script was changed. I checked the docker repo and a guy whose username sounds familiar to me added `ulimit -Hn 524288` 😆

target/scripts/build/packages.sh
polarathene: Initial review comment: https://github.com/docker-mailserver/docker-mailserver/pull/3403#discussion_r1306373757

The logwatch line is [still removed, just relocated](https://github.com/docker-mailserver/docker-mailserver/pull/3403/files#diff-95c6b88025d3f0c1a18f85c2c05a14faa73ff59b57f6e938df5113221c8018f1R177) (_no log during install of `logwatch` regarding the file, but it still gets created_). Removal of `postsrsd.secret` is no longer present, good catch there! (_this is still generated when installing the `postsrsd` package, so should it should still be removed_):

> _`postsrsd: Generating initial /etc/postsrsd.secret`_

---

Current removal (as quoted above to be removed):

https://github.com/docker-mailserver/docker-mailserver/blob/3cbcdb2d6517b6c61818afb49c0eb58ee82b6202/target/scripts/build/packages.sh#L206-L212

- From Dockerfile to `packages.sh`: https://github.com/docker-mailserver/docker-mailserver/pull/3158
- `rm /etc/postsrsd.secret` introduced to `Dockerfile` in March 2018: https://github.com/docker-mailserver/docker-mailserver/pull/891
- `rm /etc/cron.daily/00logwatch` introduced to `Dockerfile` in Sep 2019: https://github.com/docker-mailserver/docker-mailserver/pull/1262

target/scripts/build/packages.sh
```diff
@@ -43,12 +74,19 @@ function _install_postfix() {
...
-  ANTI_VIRUS_SPAM_PACKAGES=(
-    amavisd-new clamav clamav-daemon
-    pyzor razor spamassassin
+  local ANTI_VIRUS_SPAM_PACKAGES=(
+    clamav clamav-daemon
+    # spamassassin is used only with amavisd-new and pyzor/razor
+    # are used by spamassasin
+    amavisd-new spamassassin pyzor razor
+    # the following packages are all for Fail2Ban
+    fail2ban python3-pyinotify python3-dnspython
```
polarathene, line 83: Add context on the complimentary fail2ban packages for why they're installed so this information is not lost to any future maintenance that would also try to remove them: https://github.com/docker-mailserver/docker-mailserver/pull/3403#discussion_r1306581431

```suggestion
    # the following packages are all for Fail2Ban
    # https://github.com/docker-mailserver/docker-mailserver/pull/3403#discussion_r1306581431
    fail2ban python3-pyinotify python3-dnspython
```

target/scripts/build/packages.sh
```diff
@@ -43,12 +74,19 @@ function _install_postfix() {
...
   _log 'debug' 'Installing all packages now'
 
-  ANTI_VIRUS_SPAM_PACKAGES=(
-    amavisd-new clamav clamav-daemon
-    pyzor razor spamassassin
+  local ANTI_VIRUS_SPAM_PACKAGES=(
+    clamav clamav-daemon
+    # spamassassin is used only with amavisd-new and pyzor/razor
+    # are used by spamassasin
+    amavisd-new spamassassin pyzor razor
```
polarathene, line 81:
```suggestion
    # spamassassin is used only with amavisd-new, while pyzor + razor are used by spamassasin
    amavisd-new spamassassin pyzor razor
```

target/scripts/build/packages.sh
```diff
@@ -57,26 +95,33 @@ function _install_packages() {
...
+  local MAIL_PROGRAMS_PACKAGES=(
+    opendkim opendkim-tools
     opendmarc libsasl2-modules sasl2-bin
   )
 
+  # these packages are contributed by the community and not part
+  # of DMS' core set of package
+  local COMMUNITY_PACKAGES=(
+    fetchmail getmail6
+  )
```
polarathene, line 123:
```suggestion
  # These packages support community contributed features.
  # If they cause too much maintenance burden in future, they are liable for removal.
  local COMMUNITY_PACKAGES=(
    fetchmail getmail6
  )
```

target/scripts/build/packages.sh
```diff
@@ -127,105 +171,23 @@ function _install_dovecot() {
...
-  rm /etc/postsrsd.secret
-
-  _log 'debug' 'Deleting default logwatch cronjob'
-  rm /etc/cron.daily/00logwatch
-}
-
 function _post_installation_steps() {
   _log 'debug' 'Running post-installation steps (cleanup)'
+  _log 'debug' 'Deleting default logwatch cronjob'
+  rm /etc/cron.daily/00logwatch
```
polarathene, line 177:
```suggestion
  _log 'debug' 'Removing default generated secret from postsrsd package'
  rm /etc/postsrsd.secret
  _log 'debug' 'Removing default cronjob from logwatch package'
  rm /etc/cron.daily/00logwatch
```

polarathene (changes requested):

target/scripts/build/packages.sh
polarathene: Suggestion provided to bring back the postsrsd secret removal: https://github.com/docker-mailserver/docker-mailserver/pull/3403#discussion_r1461206904

target/scripts/build/packages.sh
georglauterbach: > Off-Topic: I've updated docker to version 25.0.0 and noticed, that the init script was changed. I checked the docker repo and a guy whose username sounds familiar to me added `ulimit -Hn 524288` 😆

Link please :D

target/scripts/build/packages.sh
georglauterbach: I will bring back the remove of `/etc/postsrsd.secret` in `_post_installation_steps`.

georglauterbach:
@polarathene can you please have a look at these test failures. They are related to bb64ead29d5ca3decec4cde2b89418510d997f7e which includes changes from #3780, which could be related to

```txt
not ok 7 [Process Management] (enabled ENV) should restart processes when killed in 40613ms
# (from function `_run_until_success_or_timeout' in file test/helper/common.bash, line 222,
#  from function `_should_restart_when_killed' in file test/tests/parallel/set3/container_configuration/process_check_restart.bats, line 151,
#  in test file test/tests/parallel/set3/container_configuration/process_check_restart.bats, line 118)
#   `_should_restart_when_killed "${PROCESS}"' failed
# Timed out on command: _check_if_process_is_running amavisd-new 4
# dms-test_process-check-restart_enabled
```

Moreover, we have

```txt
not ok 24 [OAuth2] should authenticate with XOAUTH2 in 243ms
# (from function `assert_output' in file test/test_helper/bats-assert/src/assert_output.bash, line 186,
#  from function `__dovecot_logs_should_verify_success' in file test/tests/serial/mail_with_oauth2.bats, line 111,
#  from function `__verify_auth_with_imap' in file test/tests/serial/mail_with_oauth2.bats, line 87,
#  from function `__should_login_successfully_with' in file test/tests/serial/mail_with_oauth2.bats, line 75,
#  in test file test/tests/serial/mail_with_oauth2.bats, line 62)
#   `__should_login_successfully_with 'XOAUTH2'' failed
#
# substring (1 lines):
#   dovecot: imap-login: Login: user=<user1@localhost.localdomain>, method=XOAUTH2
# output (5 lines):
#   2024-01-23T19:31:04.475330+00:00 mail dovecot: master: Dovecot v2.3.19.1 (9b53102964) starting up for imap, lmtp
#   2024-01-23T19:31:05.478720+00:00 mail dovecot: imap-login: Disconnected: Connection closed (disconnected before auth was ready, waited 0 secs): user=<>, rip=127.0.0.1, lip=127.0.0.1, secured, session=<Dn7J/KEPgI1/AAAB>
#   2024-01-23T19:31:05.547740+00:00 mail dovecot: imap-login: Disconnected: Connection closed (no auth attempts in 0 secs): user=<>, rip=127.0.0.1, lip=127.0.0.1, secured, session=<o4vK/KEPgo1/AAAB>
#   2024-01-23T19:31:05.666686+00:00 mail dovecot: imap-login: Login: user=<user1@localhost.localdomain>, method=OAUTHBEARER, rip=127.0.0.1, lip=127.0.0.1, mpid=885, secured, session=<80jM/KEPjo1/AAAB>
#   2024-01-23T19:31:05.668939+00:00 mail dovecot: imap(user1@localhost.localdomain)<885><80jM/KEPjo1/AAAB>: Disconnected: Logged out in=11 out=493 deleted=0 expunged=0 trashed=0 hdr_count=0 hdr_bytes=0 body_count=0 body_bytes=0
# --
#
```

I think you have the most expertise in these areas at the moment :D

target/scripts/build/packages.sh
casperklein: See [here](https://github.com/moby/moby/pull/45534/files#diff-329a7d3594b1ed91f76deded4734c7d9390c4dce74aeaa5a772fbf52b0a58c2e).

target/scripts/build/packages.sh
casperklein: 1. You forgot to add `_add_ppas` at the bottom of the file
2. IMO the function name should be more clear: `_add_rspamd_ppas`
    1. Or still my personal favorit: `_install_rspamd` with `apt-get "${QUIET}" --no-install-recommends install 'rspamd' 'redis-server'` added.

georglauterbach: Right, the line should be added indeed. Having the installation routine later install Rspamd makes the whole thing faster, though: we do not require a separate `apt-get install` and all PPAs that we potentially use in the future can be added to this method (hence the name; not being specific to Rspamd was my initial intention). What do you think?

target/scripts/build/packages.sh
polarathene: > I checked the docker repo and a guy whose username sounds familiar to me added `ulimit -Hn 524288` 😆

Yeaaah... it was well meant but apparently I made some assumptions on compatibility (_although in some cases, it [turns out the previous `ulimit` command and value could be just a no-op](https://github.com/docker/cli/issues/4807#issuecomment-1903950217)_), rather than serving the intended purpose and this change just surfaced that issue for those affected.

It's worse with containerd where my change is getting backlash from business users for breaking their production setups. I still think matching host limits by default is the right decision, but it's ruffled some feathers 😬

CHANGELOG.md
```diff
@@ -6,9 +6,36 @@ All notable changes to this project will be documented in this file. The format
 
 > **Note**: Changes and additions listed here are contained in the `:edge` image tag. These changes may not be as stable as released changes.
 
+The most noteworthy change of this release is the update of the container's base image from Debian 11 ("Bullseye") to Debian 12 ("Bookworm"). This update alone involves breaking changes and requires a careful update!
+
+### Breaking
+
+- **Updated base image to Debian 12** ([#3403](https://github.com/docker-mailserver/docker-mailserver/pull/3403))
+  - Changed the default of `DOVECOT_COMMUNITY_REPO` to `0` (disabled) - the Dovecot community repo will (for now) not be the default when building the DMS.
```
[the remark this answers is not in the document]
polarathene: This is mentioned directly after this line..

polarathene:
> @polarathene can you please have a look at these test failures
> `_check_if_process_is_running amavisd-new`

Should be a fairly obvious one to fix considering the related changes in this PR. Amavis process name changed, it needs to be adjusted in the test accordingly. The test only fails now from the revision because the coverage / specificity improved to catch it 😅 

> ``__should_login_successfully_with 'XOAUTH2'' failed`
> 
> `dovecot: imap-login: Login: user=<user1@localhost.localdomain>, method=OAUTHBEARER`

This one is odd.

The logs found in the output to match have `OAUTHBEARER` instead of `XOAUTH2`. However these two methods are split into separate test cases and `XOAUTH2` comes first. Bug with `curl` update?

polarathene (dismissed):
Looks like all feedback / concerns have now been addressed 👍 

Let's finally merge this! 💪 🥳 (_**EDIT:** After releasing v13.3.1, which I thought had already been done_ 😬 )

target/scripts/build/packages.sh
polarathene: Not sure if this was what @casperklein had in mind 🤷‍♂️ 

```suggestion
# -e          :: exit on error (do this in functions as well)
# -E          :: inherit the ERR trap to functions, command substitutions and sub-shells
```

target/scripts/build/packages.sh
```diff
@@ -17,17 +19,48 @@ function _pre_installation_steps() {
...
+    apt-utils # avoid useless warnings
+    apt-transport-https ca-certificates curl gnupg # required for adding PPAs
+    systemd-standalone-sysusers # avoid problems with SA / Amavis (https://github.com/docker-mailserver/docker-mailserver/pull/3403#pullrequestreview-1596689953)
+  )
+  apt-get "${QUIET}" install --no-install-recommends "${EARLY_PACKAGES[@]}" 2>/dev/null
+}
+
+function _add_ppas() {
+  _log 'debug' 'Adding PPAs'
+  _log 'trace' 'Adding Rspamd PPA'
```
polarathene, line 39: As earlier feedback concluded, leave this as-is is fine. It is unlikely that we'd have any new PPAs so I'd agree being more specific is preferred, but I'd rather avoid any more bike shedding on this PR 😝

polarathene (commented):
Optional - Neither need to be addressed by this PR though.

georglauterbach:
@polarathene could you please draft the v13.3.1 PR? I am currently facing issues on my dev system and I have no GUI like at all. I need to sort this out first.

georglauterbach:
I have applied your latest review comments about the Rspamd installation and will merge this PR now when tests are passing :)

casperklein:
The change of the time format ist mentioned [here](https://www.debian.org/releases/bookworm/amd64/release-notes/ch-information.en.html#rsyslog-creates-fewer-logfiles). There are other worth mentioning changes, e.g. the files /var/log/mail.{info,warn,err} are no longer created.

---

> Any idea how to revert this back to the old format?

Adding `$ActionFileDefaultTemplate RSYSLOG_TraditionalFileFormat` to /etc/rsyslog.conf brings back the old style.

georglauterbach:
It seems like rsyslog finally adopted ISO8601 as a time stamp format. I actually really appreciate that (I always thought the rsyslog format was weird and most-notable non-standard)!

We should provide an entry in the changelog though, indeed.

polarathene:
> It seems like rsyslog finally adopted ISO8601 as a time stamp format.

That's probably the newer syslog RFC format not just the timestamp change, whereas before it was BSD syslog which was less formalized in spec.

georglauterbach:
I understand that :) Can you additionally provide a PR that updates the CHANGELOG @casperklein?

Issue #3323, opened 2023-05-10
tracking: equivalent configs of SpamAssassin & Rspamd


ghnp5:
### Subject

I would like some feedback concerning a use case

### Description

It's not clear if we have equivalent configs for these, that apply to `rspamd`:

```
# deliver spam messages in the inbox (eventually tagged using SA_SPAM_SUBJECT)
SPAMASSASSIN_SPAM_TO_INBOX=1

# KAM is a 3rd party SpamAssassin ruleset, provided by the McGrail Foundation.
# If SpamAssassin is enabled, KAM can be used in addition to the default ruleset.
# - **0** => KAM disabled
# - 1 => KAM enabled
#
# Note: only has an effect if `ENABLE_SPAMASSASSIN=1`
ENABLE_SPAMASSASSIN_KAM=1

# spam messages will be moved in the Junk folder (SPAMASSASSIN_SPAM_TO_INBOX=1 required)
MOVE_SPAM_TO_JUNK=0

# add spam info headers if at, or above that level:
SA_TAG=-999.0

# add 'spam detected' headers at that level
SA_TAG2=5.0

# triggers spam evasive actions
SA_KILL=999.0

# add tag to subject if spam detected
# by being empty, "***SPAM*** " (with a space) will be used
SA_SPAM_SUBJECT=undef
```

I just want to make sure all emails reach the Inbox (and not 100% rejected or moved to Junk), independently of their Spam score, but with Headers added to them.

I see that you allow us to completely override config files, by adding files to `override.d`, but from what you understand, you already make use of `local.d`, which is the folder that can be used to override **some** configs.

So, if I want to update just a few configs, but keeping your changes in `local.d`, how can I do that, without overriding the **whole** file in `override.d` ?

Wouldn't it make more sense that the [add-line](https://docker-mailserver.github.io/docker-mailserver/latest/config/security/rspamd/#with-the-help-of-a-custom-file) custom command would add lines to `local.d` instead of `override.d`? We already can easily create files in `override.d`, hence I'm not understanding how `add-line` is useful.

Thank you very much!

georglauterbach:
[...]

PR #3440, opened 2023-07-23
rspamd: disable checks for authenticated users

georglauterbach:
# Description

Disable all checks for authenticated users. See <https://github.com/docker-mailserver/docker-mailserver/issues/3323#issuecomment-1645978207>.

The reason my PR deviates from the changes proposed by @williamdes is that enabling SPF, DKIM and policy checks for _outbound_ emails still does not make sense really (because DKIM is only evaluated for _incoming_ mail (remember, inbound emails are checked, outbound emails are signed), the same goes for SPF, and the `policies` group is just a combination of both). So we just disable all checks and hence, we can remove the line about the re-writes.

@williamdes **can you please verify this configuration for me?**:) I will keep all checks enabled on my machine, and I cannot test it locally (simply lacking time).

Related to #3323

## Type of change

## Checklist:


georglauterbach:
<https://mcgrail.com> seems to be unreachable at the moment, hence the CI failures.

mailserver.env
```diff
@@ -142,6 +142,13 @@ ENABLE_RSPAMD_REDIS=
 # 1     => enabled
 RSPAMD_LEARN=0
 
+# This settings controls whether checks should be performed on emails coming
+# from authenticated users (i.e. most likely outgoing emails). The default valu
```
williamdes, line 146:
```suggestion
# from authenticated users (i.e. most likely outgoing emails). The default value
```

target/rspamd/local.d/settings.conf
```diff
@@ -0,0 +1,12 @@
+# documentation: https://rspamd.com/doc/configuration/settings.html
+
+# DMS::SED_TAG::1::START
+# Disable all checks for authenticated users
+authenticated {
+    priority = high;
+    authenticated = yes;
+    apply {
+        groups_enabled = [];
```
williamdes, line 9: This looks okay to me, I will try to test it in production. No promises but I will do my best 

Not sure if some kind of forwarding outbound flow could be affected

georglauterbach: Nice. I'll merge when you sent me feedback about this :)

williamdes (approved):
Works fine in production !
See:
image

The score is 0 and there is no tokens in the list

casperklein (approved):

georglauterbach:
> Works fine in production !

Noice! Then I will merge this :D Thank you @williamdes for your feedback!

PR #3455, opened 2023-07-31
feat(setup): Add `fail2ban` sub-command `status <JAIL>`

nilshoell:
# Description

This PR adds a new option to the `setup.sh fail2ban` script to show a more detailed view on the status of all fail2ban jails using fail2ban's native `status <jail>` command. Effectively, the options just runs exactly that command for all jails already existing in the `JAILS` variable and outputs the result.
I wanted to have an option between just running the setup `fail2ban` sub-command without options that just returns the banned IPs and getting all details using `setup fail2ban log`. 
This might be a bit redundant to other options, but since it is a small change that breaks nothing, I figured I could move it back here, since its usage is probably very intuitive to any fail2ban user.

## Type of change

## Checklist:

- [X] If necessary I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/docs/content/config/security/fail2ban.md
+++ b/docs/content/config/security/fail2ban.md
@@ -58,6 +58,12 @@ setup fail2ban
 
 the script will show all banned IP addresses.
 
+To get Fail2ban's more detailed `status` view, run
+
+```bash
+setup fail2ban status
+```
+
 ### Managing Bans
 
 You can manage F2B with the `setup` script. The usage looks like this:
--- a/docs/content/config/security/fail2ban.md
+++ b/docs/content/config/security/fail2ban.md
@@ -58,7 +58,7 @@ setup fail2ban
 
 the script will show all banned IP addresses.
 
-To get Fail2ban's more detailed `status` view, run
+To get a more detailed `status` view, run
 
 ```bash
 setup fail2ban status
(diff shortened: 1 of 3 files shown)


target/bin/fail2ban
```diff
@@ -72,6 +73,13 @@ else
       cat /var/log/mail/fail2ban.log
       ;;
 
+    ( 'status' )
+      for JAIL in "${JAILS[@]}"; do
+        RESULT=$(fail2ban-client status "${JAIL}" 2>&1)
+        echo -e "$RESULT\n"
```
casperklein, line 79: There is no need to store the result in a variable. This should be enough:

```suggestion
        printf '%s\n\n' "$(fail2ban-client status "${JAIL}" 2>&1)"
```

PS: I've chosen `printf` over `echo` in that case, because the input is not exactly known before and could be falsely misinterpreted as option argument to `echo` - just to be safe.

docs/content/config/security/fail2ban.md
```diff
@@ -58,6 +58,12 @@ setup fail2ban
 
 the script will show all banned IP addresses.
 
+To get Fail2ban's more detailed `status` view, run
```
casperklein, line 61:
```suggestion
To get a more detailed `status` view, run
```

casperklein (approved):
Thank you for your contribution 👍

polarathene (approved):

PR #3465, opened 2023-08-07
docs: Guidance for binding outbound SMTP with multiple interfaces available

norrs:
# Description

Add tips section to the documentation, and include how I solved ensuring SMTP clients in postfix was bound to the right IP-addresses to match the entries in my SPF-record and reverse DNS (PTR) configuration on those specific IP-addresses.

Tricky part is that when you first involving the `smtp_bind_address` it also uses this as source IP address for forwarding to content_filter amavis. Since amavis is running on same node, enforce it to bind to localhost.

Interesting topics to maybe look into later:

Maybe you have even more advanced setup and want to look into specific IP-addresses used for particular domains (ie used for bulk-sending) and get some inspiration from https://serverfault.com/a/650326 . 

I were happy with only binding to the correct IP-addresses for now as I don't really have any bulk-needs in this setup.

## Type of change

## Checklist:

- [? ] My code follows the style guidelines of this project
- [-] I have commented my code, particularly in hard-to-understand areas
- [-] If necessary I have added tests that prove my fix is effective or that my feature works
- [-] New and existing unit tests pass locally with my changes


norrs:
@polarathene What I'm wondering about is if it should be filed under  `Examples -> `Use cases`` instead of `Configuration -> Tips`?

docs/content/config/advanced/tips/server-multiple-ip-addresses.md
```diff
@@ -0,0 +1,43 @@
...
+to the respective IP-address on the server you want to use.
+
+!!! info "IP-addresses shown below are reserved documentation IP-addresses by IANA, [RFC-5737](https://datatracker.ietf.org/doc/rfc5737/) and [RFC-3849](https://datatracker.ietf.org/doc/html/rfc3849)."
+
+Example `postfix-main.cf`:
+
+```
+smtp_bind_address = 198.51.100.10
+smtp_bind_address6 = 2001:DB8::10
+```
```
polarathene, line 28:
````suggestion
!!! example

    ```title="postfix-main.cf"
    smtp_bind_address = 198.51.100.10
    smtp_bind_address6 = 2001:DB8::10
    ```

    **NOTE:** IP-addresses shown above are placeholders, using reserved documentation IP-addresses by IANA, [RFC-5737](https://datatracker.ietf.org/doc/rfc5737/) and [RFC-3849](https://datatracker.ietf.org/doc/html/rfc3849).
````

File: docs/content/config/advanced/tips/server-multiple-ip-addresses.md [hunk 2]
```diff
@@ -0,0 +1,43 @@
...
+the `smtp_bind_address` as source address when "forwarding" email for filtering via amavis.
+This seems to be a better configuration than adding your bind-addresses to `mynetworks` configuration in
+Postfix.
+
+Example `postfix-master.cf`:
+
+```
+smtp-amavis/unix/smtp_bind_address=127.0.0.1
+smtp-amavis/unix/smtp_bind_address6=::1
+```
```
polarathene, line 41:
````suggestion
One problem with using `smtp_bind_address` is that the default configuration for `smtp-amavis` in
DMS needs to be updated to explicitly connect via loopback (localhost), which avoids using
the `smtp_bind_address` as source address when "forwarding" email for filtering via Amavis.

!!! example

    ```title="postfix-master.cf"
    smtp-amavis/unix/smtp_bind_address=127.0.0.1
    smtp-amavis/unix/smtp_bind_address6=::1
   ```

This seems to be a better approach than adding your bind-addresses to `mynetworks` parameter in
Postfix `postfix-main.cf`.
````

File: docs/content/config/advanced/tips/server-multiple-ip-addresses.md [hunk 3]
```diff
@@ -0,0 +1,43 @@
...
+
+If your host system is running multiple IPv4 and IPv6 IP-addresses, you probably have an interest to
+bind outgoing SMTP connections to specific IP-addresses to ensure MX records are aligned with
+PTR-records in DNS when sending emails to avoid getting blocked by SPF for example.
+
+We can use the tricks of [overriding defaults in postfix configuration](../override-defaults/postfix.md)
+by supplying a custom `docker-data/dms/config/postfix-master.cf` and `docker-data/dms/config/postfix-main.cf`.
+
+In `postfix-main.cf` you'll have to set the [smtp_bind_address](https://www.postfix.org/postconf.5.html#smtp_bind_address) and [smtp_bind_address6](https://www.postfix.org/postconf.5.html#smtp_bind_address6)
+to the respective IP-address on the server you want to use.
```
polarathene, line 19:
```suggestion
In `postfix-main.cf` you'll have to set the [`smtp_bind_address`](https://www.postfix.org/postconf.5.html#smtp_bind_address) and [`smtp_bind_address6`](https://www.postfix.org/postconf.5.html#smtp_bind_address6)
to the respective IP-address on the server you want to use.
```

polarathene (changes requested):
I'll apply these via the web UI, then pull the changes on your end to apply future feedback.

polarathene:
> What I'm wondering about is if it should be filed under `Examples -> `Use cases`` instead of `Configuration -> Tips`?

Yes, I think this is more niche. It would be better added to our Example docs for now. The config section of our docs is already quite big.

Please move the file and choose a smaller title like "Binding to a specific network". It's possible for DMS to have more than one network attached without using host network mode too, so the advice can still be helpful there.

---

Thanks for taking the time to contribute to DMS! It's very much appreciated 😁

docs/content/config/advanced/tips/server-multiple-ip-addresses.md
```diff
@@ -0,0 +1,43 @@
...
+One problem with using `smtp_bind_address` is that the default configuration for `smtp-amavis` in
+DMS needs to be updated to explicitly connect via loopback (localhost), which avoids using
+the `smtp_bind_address` as source address when "forwarding" email for filtering via Amavis.
+
+!!! example
+
+    ```title="postfix-master.cf"
+    smtp-amavis/unix/smtp_bind_address=127.0.0.1
+    smtp-amavis/unix/smtp_bind_address6=::1
+   ```
```
polarathene, line 38:
````suggestion
!!! example

    ```title="postfix-master.cf"
    smtp-amavis/unix/smtp_bind_address=127.0.0.1
    smtp-amavis/unix/smtp_bind_address6=::1
    ```
````

File: docs/content/config/advanced/tips/server-multiple-ip-addresses.md [hunk 5]
```diff
@@ -0,0 +1,43 @@
+---
+title: 'Advanced | Tips | Server with multiple IP-addresses'
+---
+
+## Server with multiple IP-addresses
```
polarathene, line 5:
```suggestion
---
title: 'Use Cases | Binding DMS to a specific network'
hide:
  - toc
---
```

No need to repeat the title :)

Also added extra metadata to disable the right-side navigation menu as that's not needed for this page.

polarathene (changes requested):
Sorry, last few changes to address should be good to go now 👍

polarathene:
Quick question, do you feel this information is relevant vs using the existing feature [`NETWORK_INTERFACE` ENV](https://docker-mailserver.github.io/docker-mailserver/latest/config/environment/#network_interface) we have?:

image

**UPDATE:** Yes. That doesn't appear to be as useful (_at least for IPv6 support_). But might still be a viable option with `PERMIT_DOCKER=container` for IPv4?

## Potential variant - `NETWORK_INTERFACE` + `PERMIT_DOCKER` (IPv4 only)

This ENV only seems to be used for supporting [`PERMIT_DOCKER` ENV](https://docker-mailserver.github.io/docker-mailserver/v12.1/config/environment/#permit_docker) (_where it queries for the specified interface and retrieves the IPv4 address only. The docs advise setting another ENV to configure Postfix to only listen on IPv4 for compatibility_):

https://github.com/docker-mailserver/docker-mailserver/blob/f28fce9cc432f1f447bd963d9e54e44bcf2c27dd/target/scripts/startup/setup.d/networking.sh#L16-L27

Above is extracted/used below for the different `PERMIT_DOCKER` settings:

- `CONTAINER_IP` is extracted for `PERMIT_DOCKER=container` (_IP assigned from the interface to the DMS container_):

   https://github.com/docker-mailserver/docker-mailserver/blob/f28fce9cc432f1f447bd963d9e54e44bcf2c27dd/target/scripts/startup/setup.d/networking.sh#L57-L59

- `CONTAINER_NETWORK` is extracted (_as a CIDR `/16` subnet (last two octets) of the `CONTAINER_IP` to belong to_) for `PERMIT_DOCKER=host`. While intended for host mode networking, `/16` may not always be an appropriate size? 🤷‍♂️ :

   https://github.com/docker-mailserver/docker-mailserver/blob/f28fce9cc432f1f447bd963d9e54e44bcf2c27dd/target/scripts/startup/setup.d/networking.sh#L61-L63

- `PERMIT_DOCKER=network` uses a standard docker IPv4 subnet, but assumes that the container is assigned to it:

   https://github.com/docker-mailserver/docker-mailserver/blob/f28fce9cc432f1f447bd963d9e54e44bcf2c27dd/target/scripts/startup/setup.d/networking.sh#L65-L67

- While `PERMIT_DOCKER=connected-networks` is the extended variant, iterating through all IPv4 networks visible to the container and whitelists these as `/16` subnets:

   https://github.com/docker-mailserver/docker-mailserver/blob/f28fce9cc432f1f447bd963d9e54e44bcf2c27dd/target/scripts/startup/setup.d/networking.sh#L25-L27

   https://github.com/docker-mailserver/docker-mailserver/blob/f28fce9cc432f1f447bd963d9e54e44bcf2c27dd/target/scripts/startup/setup.d/networking.sh#L50-L55

   https://github.com/docker-mailserver/docker-mailserver/blob/f28fce9cc432f1f447bd963d9e54e44bcf2c27dd/target/scripts/helpers/network.sh#L19-L23

Those all then get whitelisted by Postfix (`main.cf:mynetworks`) and the OpenDKIM/OpenDMARC services, but not others:

https://github.com/docker-mailserver/docker-mailserver/blob/f28fce9cc432f1f447bd963d9e54e44bcf2c27dd/target/scripts/startup/setup.d/networking.sh#L34-L43

Perhaps it could be adapted for handling IPv6 too? 🤷‍♂️ Seems a bit questionable though with the implicit subnet choices..

---

## Potential variant - Avoid `postfix-main.cf` only `postfix-master.cf`

**TL;DR:** Set `smtp_bind_address` / `smtp_bind_address6` via `postfix-master.cf` to target the `smtp` service instead of `smtp-amavis`? (_not sure if anything else is relevant if only concerned with outbound mail over port 25, maybe matters for relay connections? Although presently that diverts to `smtp` anyway_)

[`inet_interfaces`](https://www.postfix.org/postconf.5.html#inet_interfaces) is a single setting that could be changed to include both IPv4 and IPv6 addresses (including loopback `127.0.0.1` + `::1`). Both IPv4 + IPv6 settings for [`smtp_bind_address`](https://www.postfix.org/postconf.5.html#smtp_bind_address) would implicitly use their respective address, but only if there is one of their kind set in `inet_interfaces` (so no loopback).

The docs however infer that with the loopback addresses, similar to the default `all` value it will have the system choose the source IP (_unclear if that'd be constrained to those specified in `inet_interfaces`_), with another suggestion to have `smtp_bind_address` settings set to `0.0.0.0` / `::` (any interface) - which AFAIK would likewise defer to the system to choose source IP (_not constraining to IP_).

That said `smtp_bind_address` + `smtp_bind_address6` could possibly be set in `postfix-master.cf` to target the `smtp` service specifically, instead of Amavis (_and anything else that may be affected at some point/configuration_), if it's only a relevant setting for the `smtp` service (outbound mail)?:

https://github.com/docker-mailserver/docker-mailserver/blob/f28fce9cc432f1f447bd963d9e54e44bcf2c27dd/target/postfix/master.cf#L13

docs/content/config/advanced/tips/server-multiple-ip-addresses.md
```diff
@@ -0,0 +1,43 @@
...
+!!! example
+
+    ```title="postfix-master.cf"
+    smtp-amavis/unix/smtp_bind_address=127.0.0.1
+    smtp-amavis/unix/smtp_bind_address6=::1
+    ```
+
+This seems to be a better approach than adding your bind-addresses to `mynetworks` parameter in
+Postfix `postfix-main.cf`.
+
```
polarathene, line 42: Postfix `mynetworks` is for trusted networks (we manage with `PERMIT_DOCKER`, not really what interface to use as source for outbound connections. No need to bring it up as an alternative.

```suggestion
```

polarathene (changes requested):

norrs:
This is rather a very specific and special setup, because I host several things at one node with multiple IP addresses. 

Be free to to squash your commits and amend while insert your-self with `Co-authored-by: NAME <NAME@EXAMPLE.COM>` in the bottom of the commit message @polarathene .

Postfix is configured in DMS to deliver incoming mail and use a `content_filter` to forward mails with SMTP protocol to the smtp-amavis for scanning. This again is delivered back to postfix again on :10025 via localhost for further processing. 

> That said smtp_bind_address + smtp_bind_address6 could possibly be set in postfix-master.cf to target the smtp service specifically, instead of Amavis (and anything else that may be affected at some point/configuration), if it's only a relevant setting for the smtp service (outbound mail)?:

This is basically what I'm doing with the custom postfix-master.cf for smtp traffic targeting amavis. (which is used by the `content_filter`.) See `docker exec mailserver postconf -P` . 

The default DMS configuration for postfix-master.cf could probably contain this default binding to localhost (`smtp-amavis/unix/smtp_bind_address = 127.0.0.1`), as long as the setup doesn't change where amavis runs on another node (for scaling, ie requires multiple instances of amavis) and postfix requires to communicate with it by non-loopback. However, those who need that and uses DMS, probably know how to update these configuration by using the possibilities of overriding the main and master configs.

polarathene:
I'll wrap up this PR and merge it, although your feedback on advice below would be good to hear first :)

---

> This is rather a very specific and special setup, because I host several things at one node with multiple IP addresses.

Is there any reason you use host mode networking for DMS btw? If you want to target a specific interface, I believe you can specify that beside the port mapping in `compose.yaml` and it will bind only to that IP on the host. Presumably that would also restrict the interface used if the others are not needed in DMS.

If you could test that and let me know that'd be appreciated, we could add it in the docs page too.

> However, those who need that and uses DMS, probably know how to update these configuration by using the possibilities of overriding the main and master configs.

Yes, and DMS isn't really maintained for internal services to be scaled out individually (_although we do provide a means to disable/enable many of them specifically_).

If `smtp_bind_address ` was not already inherited, the `smtp-amavis` service could just set `inet_interfaces=loopback-only` I believe. Looking back on `master.cf`, on the right-most column is the process that will run, while on the left-most is the service definition (named unix socket or inet port); it'd probably be valid to set `smtp_bind_address` on the [`smtp` inet service](https://github.com/docker-mailserver/docker-mailserver/blob/8f971713360c4620046c7f223fe66df9ea800253/target/postfix/master.cf#L13) with `smtp-amavis/unix/smtp_bind_address = your_ip_here`, assuming no other service needs the change (_every other service in `master.cf` would inherit the `main.cf` default `smtp_bind_address` (empty) then AFAIK_):

```
smtp/inet/smtp_bind_address = 198.51.100.10
smtp/inet/smtp_bind_address6 = 2001:DB8::10
```

No `smtp-amavis` (`postfix-master.cf`) or `postfix-main.cf` changes would be required if the above works.

As `smtp-amavis` is for internal use in DMS (_and planned to eventually be replaced by rspamd AFAIK_), I'm happy to upstream the `smtp_bind_address=127.0.0.1` + `smtp_bind_address6=::1` settings to the amavis config in [`postfix-amavis.cf`](https://github.com/docker-mailserver/docker-mailserver/blob/8f971713360c4620046c7f223fe66df9ea800253/target/amavis/postfix-amavis.cf).

norrs:
> I'll wrap up this PR and merge it, although your feedback on advice below would be good to hear first :)
> 
> > This is rather a very specific and special setup, because I host several things at one node with multiple IP addresses.
> 
> Is there any reason you use host mode networking for DMS btw? If you want to target a specific interface, I believe you can specify that beside the port mapping in `compose.yaml` and it will bind only to that IP on the host. Presumably that would also restrict the interface used if the others are not needed in DMS.
> 
> If you could test that and let me know that'd be appreciated, we could add it in the docs page too.

I only ended up with host networking due to gmail rejecting emails with spf failures, and then I looked into why it was rejecting. I didn't notice right away that outbound smtp connections randomly bound to the "first" random IP address on the interface.

I started a new job this week, but I can probably test during the weekend if time allows it by temporarying adding a server and configure a domain for testing in https://github.com/norrs/ansible/tree/main/playbooks/mailserver . 

> 
> As `smtp-amavis` is for internal use in DMS (_and planned to eventually be replaced by rspamd AFAIK_), I'm happy to upstream the `smtp_bind_address=127.0.0.1` + `smtp_bind_address6=::1` settings to the amavis config in [`postfix-amavis.cf`](https://github.com/docker-mailserver/docker-mailserver/blob/8f971713360c4620046c7f223fe66df9ea800253/target/amavis/postfix-amavis.cf).

I believe those changes should work without issues, even in non host-mode in docker networking. Unsure what happens if you try to configure IPv6 binding and your system doesn't have IPv6 enabled tho.. 

Also IPv6 in docker probably needs to be configured like mentioned here ( https://docs.docker.com/config/daemon/ipv6/ ), which I actually haven't done myself .. :fearful:  

Which I guess is also why I went direct to networking host-mode in the first place :upside_down_face:

docs/content/config/advanced/tips/server-multiple-ip-addresses.md
```diff
@@ -0,0 +1,43 @@
...
+  - toc
+---
+
+!!! warning "Advice not extensively tested"
+
+    This configuration has only been used with `network: host`, where you have direct access to the host interfaces.
+
+If your host system is running multiple IPv4 and IPv6 IP-addresses, you probably have an interest to
+bind outgoing SMTP connections to specific IP-addresses to ensure MX records are aligned with
+PTR-records in DNS when sending emails to avoid getting blocked by SPF for example.
```
norrs, line 13: Maybe the phrase about SPF should be more clear:

bind outgoing SMTP connections to specific IP-addresses to ensure MX records are aligned with PTR and SPF records in DNS when sending (and receiving emails).

You expect at least the resolving IP-addresses from the MX records in DNS to have reverse DNS (PTR records) matching the MX records. 
And those IP-addresses for sending emails need to be specified in the SPF records in DNS. (should be correct for bounces etc as well, so affects receiving as well?) 

Agree @polarathene ?

polarathene:
> I started a new job this week, but I can probably test during the weekend if time allows it by temporarying adding a server and configure a domain for testing

Congrats on the new role! 🎉 

No rush. If you can find the time to test and get back to me (_in this PR is fine, even if it's closed_), that'd be appreciated 👍 

> Also IPv6 in docker probably needs to be configured like mentioned here ( https://docs.docker.com/config/daemon/ipv6/ ), which I actually haven't done myself .. 😨
> 
> Which I guess is also why I went direct to networking host-mode in the first place 🙃

I recently rewrote our IPv6 docs, until v13 is released you can only see them in the `edge` versioned docs: https://docker-mailserver.github.io/docker-mailserver/edge/config/advanced/ipv6/

I also contributed to the updated upstream Docker IPv6 docs you linked, as prior they were not very good. I'd advise giving our IPv6 docs a good look.

If you have a more advanced IPv6 setup where ULA is not appropriate for you, you'll find some extra tips linked to the IPv6 docs rewrite PR, as I didn't have the time to figure all that out and document it properly. IPv6 ULA should meet most users needs very well though (_just specify the IPv6 address on host to bind the containers internal ULA address to_).

docs/content/config/advanced/tips/server-multiple-ip-addresses.md
polarathene: I don't think the PTR records have to match MX records since that's only related to where mail should be delivered/received for that domain.

IIRC, when this type of check is done, it's not necessarily strict, it's usually acceptable to resolve the sending mail-server shares the same IP as the PTR record and the address it points to are the same.

Example setup:
- DMS server + config:
  - PTR `1.2.3.4` => `example.com`
  - A `example.com` => `1.2.3.4`
  - A `mail.example.com` => `1.2.3.4`
  - `compose.yaml` with hostname `mail.example.com` + `setup email add hello@example.org`
- DNS support for mail account `hello@example.org` (_mail accounts don't need to share the same domain or server as DMS_):
  - A `example.org` => `5.6.7.8` (_not important to DMS_)
  - MX `example.org` => `mail.example.com` (_Mail sent to `example.org` will be handled by DMS, not relevant to example below_)
  - TXT (SPF) `example.org` =>`"v=spf1 mx ip4:1.2.3.4 a:mail.example.com -all"` (_only the `mx` should be necessary, but could alternatively specify an IP address authorized to send mail on behalf of `example.org`, that IP address could implicitly be resolved for an `A` record lookup too if the `MX` of `example.org` is not appropriate_)

MTA connection example:
1.When connecting to another MTA to send mail to (_with sender `hello@example.org`_), DMS will greet the MTA with HELO / EHLO of `mail.example.com`.
2. The MTA we are sending the mail to (_eg: belonging to a recipient, or is a relay service_) may do a PTR lookup for the IP that `mail.example.com` points to (`1.2.3.4`), which the PTR record resolves to `example.com`, which in turn resolves back to `1.2.3.4`, same as DMS at `mail.example.com`, pass ✔️ 
3. SPF record can also be checked, verifying that `example.org` authorizes `mail.example.com` (_or in this case anything sending from that IP_).

I may be mistaken if I'm recalling the above incorrectly. But the `PTR` check is tied to DMS hostname (_that Postfix gets configured for_), like with the TLS certificate, the trust is in the mail-server here, the sender address is not relevant (_separately checked via DNS lookup for records like SPF that authorize DMS server to send_).

polarathene: So perhaps it should be advising in the docs here to match the IP that `hostname` (`compose.yaml`) would resolve to via DNS?

norrs:
For the record:

@polarathene I have not forgotten about this, but I havent had any free time the last weeks. Maybe I find some wriggle room next week or early october. 

As I plan to test https://github.com/docker-mailserver/docker-mailserver/pull/3465#issuecomment-1678159329 this in regards of `networking: bridge mode`

georglauterbach:
@norrs can you please resume working on this? I am trying to "clean" up old PRs because I know from experience that PRs like this tend to fade and rot 🙈

polarathene:
I'll do a review pass over this again sometime this week. If it's worthwhile to me then, I'll wrap this PR up and merge, otherwise will reject.

georglauterbach:
> I'll do a review pass over this again sometime this week. If it's worthwhile to me then, I'll wrap this PR up and merge, otherwise will reject.

👍🏼 Please close if you come to the conclusion that this may not be worthwhile :)

docs/content/config/advanced/tips/server-multiple-ip-addresses.md
polarathene, line 13:
```suggestion
If your Docker host is running multiple IPv4 and IPv6 IP-addresses, it may be beneficial to bind outgoing SMTP connections to specific IP-address / interface. When a mail is sent outbound from DMS, it greets the MTA it is connecting to with a EHLO (DMS FQDN) which might be verified against the IP resolved, and that a `PTR` record for that IP resolves an address back to the same IP. If DMS connections are inconsistent with the IP used here, these DNS checks are likely to fail.
```

[`reject_unknown_sender`](https://github.com/docker-mailserver/docker-mailserver/issues/3716#issuecomment-1868033234) is a setting DMS has which other mail servers may also check for, where a foreign sender address is checked for having a valid DNS MX or A record. The linked issue is an SPF failure related to that.

There are two related restrictions Postfix can configure related to the suggestion that other MTAs may also check for.

> [**`reject_unknown_reverse_client_hostname`**](https://www.postfix.org/postconf.5.html#reject_unknown_reverse_client_hostname):
> **Reject** the request when the **client IP address** has **no address -> name mapping**.
> This is a weaker restriction than the [`reject_unknown_client_hostname`](https://www.postfix.org/postconf.5.html#reject_unknown_client_hostname) feature, which requires not only that the **address -> name** and **name -> address** mappings exist, but _also_ that **the two mappings reproduce the client IP address**.

---

SPF only comes into play when it authorizes a mail server via `mx` / `ip4` / `ip6` / `a` etc, as [demonstrated in my prior review comment](https://github.com/docker-mailserver/docker-mailserver/pull/3465/files#r1294061262). Likewise those DNS records related changes would need to have consistent IP.

docs/content/config/advanced/tips/server-multiple-ip-addresses.md
```diff
@@ -0,0 +1,43 @@
...
+
+This can be configured by [overriding the default Postfix configurations](../override-defaults/postfix.md) DMS provides. Create `postfix-master.cf` and `postfix-main.cf` files for your config volume (`docker-data/dms/config`).
+
+In `postfix-main.cf` you'll have to set the [`smtp_bind_address`](https://www.postfix.org/postconf.5.html#smtp_bind_address) and [`smtp_bind_address6`](https://www.postfix.org/postconf.5.html#smtp_bind_address6)
+to the respective IP-address on the server you want to use.
```
polarathene, line 18:
```suggestion
This can be configured by [overriding the default Postfix configurations][docs::overrides-postfix] DMS provides. Create `postfix-master.cf` and `postfix-main.cf` files for your config volume (`docker-data/dms/config`).

In `postfix-main.cf` you'll have to set the [`smtp_bind_address`][postfix-docs::smtp-bind-address-ipv4] and [`smtp_bind_address6`][postfix-docs::smtp-bind-address-ipv6]
to the respective IP-address on the server you want to use.

[docs::overrides-postfix]: ../override-defaults/postfix.md
[postfix-docs::smtp-bind-address-ipv4]: https://www.postfix.org/postconf.5.html#smtp_bind_address
[postfix-docs::smtp-bind-address-ipv6]: https://www.postfix.org/postconf.5.html#smtp_bind_address6
```

docs/content/config/advanced/tips/server-multiple-ip-addresses.md
```diff
@@ -0,0 +1,43 @@
...
+
+!!! example
+
+    ```title="postfix-main.cf"
+    smtp_bind_address = 198.51.100.10
+    smtp_bind_address6 = 2001:DB8::10
+    ```
+
+    **NOTE:** IP-addresses shown above are placeholders, using reserved documentation IP-addresses by IANA, [RFC-5737](https://datatracker.ietf.org/doc/rfc5737/) and [RFC-3849](https://datatracker.ietf.org/doc/html/rfc3849).
```
polarathene, line 27:
```suggestion
    **NOTE:** IP addresses shown above are placeholders, they are IP addresses reserved for documentation by IANA (_[RFC-5737 (IPv4)][rfc-5737] and [RFC-3849 (IPv6)][rfc-3849]_). Replace with the IP addresses you want DMS to send mail through.
    
[rfc-5737]: https://datatracker.ietf.org/doc/html/rfc5737
[rfc-3849]: https://datatracker.ietf.org/doc/html/rfc3849
```

docs/content/config/advanced/tips/server-multiple-ip-addresses.md
```diff
@@ -0,0 +1,43 @@
...
+This can be configured by [overriding the default Postfix configurations](../override-defaults/postfix.md) DMS provides. Create `postfix-master.cf` and `postfix-main.cf` files for your config volume (`docker-data/dms/config`).
+
+In `postfix-main.cf` you'll have to set the [`smtp_bind_address`](https://www.postfix.org/postconf.5.html#smtp_bind_address) and [`smtp_bind_address6`](https://www.postfix.org/postconf.5.html#smtp_bind_address6)
+to the respective IP-address on the server you want to use.
+
+!!! example
+
+    ```title="postfix-main.cf"
+    smtp_bind_address = 198.51.100.10
+    smtp_bind_address6 = 2001:DB8::10
```
polarathene, line 24:
```suggestion
    smtp_bind_address = 198.51.100.42
    smtp_bind_address6 = 2001:DB8::42
```

docs/content/config/advanced/tips/server-multiple-ip-addresses.md
```diff
@@ -0,0 +1,43 @@
...
+One problem with using `smtp_bind_address` is that the default configuration for `smtp-amavis` in
+DMS needs to be updated to explicitly connect via loopback (localhost), which avoids using
+the `smtp_bind_address` as source address when "forwarding" email for filtering via Amavis.
+
+!!! example
+
+    ```title="postfix-master.cf"
+    smtp-amavis/unix/smtp_bind_address=127.0.0.1
+    smtp-amavis/unix/smtp_bind_address6=::1
+    ```
```
polarathene, line 38:
```suggestion
    !!! bug "Inheriting the bind from `main.cf` can misconfigure services"

        One problem when setting `smtp_bind_address` in `main.cf` is that it will be inherited by any services in `master.cf` that extend the `smtp` transport. One of these is `smtp-amavis`, which is explicitly configured to listen / connect via loopback (localhost / `127.0.0.1`).

        A `postfix-master.cf` override can workaround that issue by ensuring `smtp-amavis` binds to the expected internal IP:

        ```title="postfix-master.cf"
        smtp-amavis/unix/smtp_bind_address=127.0.0.1
        smtp-amavis/unix/smtp_bind_address6=::1
        ```
```

docs/content/config/advanced/tips/server-multiple-ip-addresses.md
```diff
@@ -0,0 +1,43 @@
+---
+title: 'Use Cases | Binding DMS to a specific network'
+hide:
+  - toc
+---
+
+!!! warning "Advice not extensively tested"
+
+    This configuration has only been used with `network: host`, where you have direct access to the host interfaces.
```
polarathene, line 9:
```suggestion
!!! warning "Advice not extensively tested"

    This configuration advice is a community contribution which has only been verified as a solution when using `network: host`, where you have direct access to the host interfaces.

    It may be applicable in other network modes if the container has control of the outbound IPs to bind to. This is not the case with bridge networks that typically bind to a private range network for containers which are bridged to a public interface via Docker.
```

polarathene (changes requested):
I'll apply these revisions.

docs/content/config/advanced/tips/server-multiple-ip-addresses.md
```diff
@@ -0,0 +1,46 @@
...
+    !!! bug "Inheriting the bind from `main.cf` can misconfigure services"
+
+        One problem when setting `smtp_bind_address` in `main.cf` is that it will be inherited by any services in `master.cf` that extend the `smtp` transport. One of these is `smtp-amavis`, which is explicitly configured to listen / connect via loopback (localhost / `127.0.0.1`).
+
+        A `postfix-master.cf` override can workaround that issue by ensuring `smtp-amavis` binds to the expected internal IP:
+
+        ```title="postfix-master.cf"
+        smtp-amavis/unix/smtp_bind_address=127.0.0.1
+        smtp-amavis/unix/smtp_bind_address6=::1
+        ```
```
polarathene, line 45: Using Content Tabs feature to share the alternative (unverified) discussed during review: https://github.com/docker-mailserver/docker-mailserver/pull/3465#issuecomment-1678107233

```suggestion
!!! example

    === "Contributed solution"

        ```title="postfix-main.cf"
        smtp_bind_address = 198.51.100.42
        smtp_bind_address6 = 2001:DB8::42
        ```

        !!! bug "Inheriting the bind from `main.cf` can misconfigure services"

            One problem when setting `smtp_bind_address` in `main.cf` is that it will be inherited by any services in `master.cf` that extend the `smtp` transport. One of these is `smtp-amavis`, which is explicitly configured to listen / connect via loopback (localhost / `127.0.0.1`).

            A `postfix-master.cf` override can workaround that issue by ensuring `smtp-amavis` binds to the expected internal IP:

            ```title="postfix-master.cf"
            smtp-amavis/unix/smtp_bind_address=127.0.0.1
            smtp-amavis/unix/smtp_bind_address6=::1
            ```

    === "Alternative (unverified)"

        During review of the contribution, it was identified that a better solution might be to [explicitly set the `smtp_bind_address` override on the `smtp` transport service][gh-pr::3465] instead:

        ```title="postfix-master.cf"
        smtp/inet/smtp_bind_address = 198.51.100.42
        smtp/inet/smtp_bind_address6 = 2001:DB8::42
        ```

        If that avoids the concern with `smtp-amavis`, you may still need to additionally override for the [`relay` transport][gh-src::postfix-master-cf::relay-transport] as well if you have configured DMS to relay mail.

!!! note "IP addresses for documentation"

    IP addresses shown in above examples are placeholders, they are IP addresses reserved for documentation by IANA (_[RFC-5737 (IPv4)][rfc-5737] and [RFC-3849 (IPv6)][rfc-3849]_). Replace them with the IP addresses you want DMS to send mail through.
    
[rfc-5737]: https://datatracker.ietf.org/doc/html/rfc5737
[rfc-3849]: https://datatracker.ietf.org/doc/html/rfc3849
[gh-pr::3465]: https://github.com/docker-mailserver/docker-mailserver/pull/3465#issuecomment-1678107233
[gh-src::postfix-master-cf::relay-transport]: https://github.com/docker-mailserver/docker-mailserver/blob/9cdbef2b369fb4fb0f1b4e534da8703daf92abc9/target/postfix/master.cf#L65
```

polarathene (changes requested):

docs/content/config/advanced/tips/server-multiple-ip-addresses.md
```diff
@@ -0,0 +1,63 @@
...
+  - toc
+---
+
+!!! warning "Advice not extensively tested"
+
+    This configuration advice is a community contribution which has only been verified as a solution when using `network: host`, where you have direct access to the host interfaces.
+
+    It may be applicable in other network modes if the container has control of the outbound IPs to bind to. This is not the case with bridge networks that typically bind to a private range network for containers which are bridged to a public interface via Docker.
+
+If your Docker host is running multiple IPv4 and IPv6 IP-addresses, it may be beneficial to bind outgoing SMTP connections to specific IP-address / interface. When a mail is sent outbound from DMS, it greets the MTA it is connecting to with a EHLO (DMS FQDN) which might be verified against the IP resolved, and that a `PTR` record for that IP resolves an address back to the same IP. If DMS connections are inconsistent with the IP used here, these DNS checks are likely to fail.
```
polarathene, line 13:
```suggestion
If your Docker host is running multiple IPv4 and IPv6 IP-addresses, it may be beneficial to bind outgoing SMTP connections to specific IP-address / interface.

- When a mail is sent outbound from DMS, it greets the MTA it is connecting to with a EHLO (DMS FQDN) which might be verified against the IP resolved, and that a `PTR` record for that IP resolves an address back to the same IP.
- If DMS connections are inconsistent with the IP used here, these DNS checks are likely to fail.
```

docs/content/config/advanced/tips/server-multiple-ip-addresses.md
```diff
@@ -0,0 +1,63 @@
...
+            A `postfix-master.cf` override can workaround that issue by ensuring `smtp-amavis` binds to the expected internal IP:
+
+            ```title="postfix-master.cf"
+            smtp-amavis/unix/smtp_bind_address=127.0.0.1
+            smtp-amavis/unix/smtp_bind_address6=::1
+            ```
+
+    === "Alternative (unverified)"
+
+        During review of the contribution, it was identified that a better solution might be to [explicitly set the `smtp_bind_address` override on the `smtp` transport service][gh-pr::3465] instead:
```
polarathene, line 46:
```suggestion
        A potentially better solution might be to instead [explicitly set the `smtp_bind_address` override on the `smtp` transport service][gh-pr::3465]:
```

polarathene (changes requested):

docs/mkdocs.yml
```diff
@@ -167,6 +167,7 @@ nav:
       - 'Customize IMAP Folders': examples/use-cases/imap-folders.md
       - 'iOS Mail Push Support': examples/use-cases/ios-mail-push-support.md
       - 'Lua Authentication': examples/use-cases/auth-lua.md
+      - 'Bind outbound SMTP to a specific network': examples/user-cases/bind-smtp-interface.md
```
polarathene, line 170:
```suggestion
      - 'Bind outbound SMTP to a specific network': examples/user-cases/bind-smtp-network-interface.md
```

docs/mkdocs.yml
```diff
@@ -167,6 +167,7 @@ nav:
       - 'Customize IMAP Folders': examples/use-cases/imap-folders.md
       - 'iOS Mail Push Support': examples/use-cases/ios-mail-push-support.md
       - 'Lua Authentication': examples/use-cases/auth-lua.md
+      - 'Bind outbound SMTP to a specific network': examples/user-cases/bind-smtp-network-interface.md
```
polarathene, line 170: Typo fix. Sorry can't run the docs build locally atm so relying on CI 😬 

```suggestion
      - 'Bind outbound SMTP to a specific network': examples/use-cases/bind-smtp-network-interface.md
```

docs/content/examples/use-cases/bind-smtp-network-interface.md
```diff
@@ -0,0 +1,66 @@
...
+
+- When a mail is sent outbound from DMS, it greets the MTA it is connecting to with a EHLO (DMS FQDN) which might be verified against the IP resolved, and that a `PTR` record for that IP resolves an address back to the same IP.
+- If DMS connections are inconsistent with the IP used here, these DNS checks are likely to fail.
+
+This can be configured by [overriding the default Postfix configurations][docs::overrides-postfix] DMS provides. Create `postfix-master.cf` and `postfix-main.cf` files for your config volume (`docker-data/dms/config`).
+
+In `postfix-main.cf` you'll have to set the [`smtp_bind_address`][postfix-docs::smtp-bind-address-ipv4] and [`smtp_bind_address6`][postfix-docs::smtp-bind-address-ipv6]
+to the respective IP-address on the server you want to use.
+
+[docs::overrides-postfix]: ../override-defaults/postfix.md
```
polarathene, line 23:
```suggestion
[docs::overrides-postfix]: ../../config/advanced/override-defaults/postfix.md
```

polarathene (approved):
Thanks for the contribution! ❤️

norrs:
> You sir, are awesome

A huge thanks @polarathene , I never managed to follow this up in a timely manner and I apology for this. In the end, the suggested changes and extra work you put into it made it even better :purple_heart:  

Next challenge, able to solve this with a non host networking mode as mentioned in https://github.com/docker-mailserver/docker-mailserver/pull/3465#issuecomment-1678159329 ? :thinking: 
Maybe I brew something together one day :sweat_smile:

polarathene:
> Next challenge, able to solve this with a non host networking mode as mentioned in [#3465 (comment)](https://github.com/docker-mailserver/docker-mailserver/pull/3465#issuecomment-1678159329) ? 🤔

Just letting you know we received a contribution recently that resolves that concern: https://github.com/docker-mailserver/docker-mailserver/pull/4330

Issue #2713, opened 2022-08-07
[FR] Improvements regarding OIDC


Westie:
I've tagged this as an issue as I've had to completely hack around with various parts of this docker project to get what I want it to do with regards to loading stuff how I want it to load, and I thought it would probably be nice to get this made far nicer.

I'd like to volunteer myself to at least partly solve this issue as I want to make what I've done far easier to main - at present I'm relocating init scripts in an attempt to modify fewer files.

c.f.: <https://github.com/docker-mailserver/docker-mailserver/issues/2312>

## Life so far

```
COPY conf/dovecot/09-auth.conf /etc/dovecot/conf.d/09-auth.conf
COPY conf/dovecot/auth-ldap.conf.ext /etc/dovecot/conf.d/auth-ldap.conf.ext
COPY conf/dovecot/auth-oauth2.conf.ext /etc/dovecot/conf.d/auth-oauth2.conf.ext
COPY conf/dovecot/dovecot-oauth2.conf.ext /etc/dovecot/dovecot-oauth2.conf.ext
```

### `scripts/setup-stack.sh`

Within my hijacked `start-mailserver.sh` I have the following:

```
configomat.sh OAUTH_ /etc/dovecot/dovecot-oauth2.conf.ext
```

Notwithstanding that I realise that this should actually be `OAUTH2_` (sue me, I wasn't thinking) - this is a rather easy addition to the project as we can just stick it within [setup-stack.sh](https://github.com/docker-mailserver/docker-mailserver/blob/master/target/scripts/startup/setup-stack.sh) somewhere.

### `dovecot/10-auth.conf`

At present I'm doing weird magic in order to ensure that the OAUTH2 driver is being loaded before the LDAP driver i.e., I'm creating a separate file called `09-auth.conf` and inserting my required config entries in there.

I had tried setting LDAP to specific authentication mechanisms but internally rewritten e-mails stopped getting delivered.

The project has its own `10-auth.conf` [file](https://github.com/docker-mailserver/docker-mailserver/blob/master/target/dovecot/10-auth.conf) which has the following order:

```
#!include auth-system.conf.ext
#!include auth-sql.conf.ext
#!include auth-ldap.conf.ext
!include auth-passwdfile.inc
#!include auth-checkpassword.conf.ext
#!include auth-vpopmail.conf.ext
#!include auth-static.conf.ext
```

[...]

PR #3480, opened 2023-08-13
feat: Auth - OAuth2 (Dovecot PassDB)

thechubbypanda:
## Description

I decided to start on an implementation of Oauth2 as described in #2713.

I am testing this locally with my [Authentik](https://goauthentik.io/) instance and a roundcube instance configured as per their [documentation](https://github.com/roundcube/roundcubemail/wiki/Configuration:-OAuth2).

<details>
<summary>Original details prior to PR feedback</summary>

I'm setting the following new variables as such:

```env
ACCOUNT_PROVISIONER=OAUTH2
OAUTH2_CLIENT_ID=id
OAUTH2_CLIENT_SECRET=secret
OAUTH2_TOKENINFO_URL=https://authentik.domain.com/application/o/userinfo/?access_token=
```

The current oauth2 login (that's all I've done so far) accepts xoauth2 tokens from something like roundcube using the dovecot passdb xoauth2 mechanism.
- It disables the default passdb and so users can't login with `login` or `plain` despite the fact that the configuration allows it and so all email clients that don't support generic oauth cannot login.
- This could have been mitigated by using [password grant](https://doc.dovecot.org/configuration_manual/authentication/oauth2/#proxy-with-password-grant) but this has been "deprecated" (is no longer best practice).

</details>

With the current changes, I am able to:
- Login with roundcube
- Send emails (albeit with a few errors discussed below)

## Implementation TODOs

- [ ] Figure out a system for provisioning accounts such that they can receive emails before login
- [ ] Figure out a system for querying account changes on a periodic basis
- [ ] Figure out a way to define aliases in the oauth2 provider such that the query can read them
- [ ] Figure out a way to define quotas in the oauth2 provider such that the query can read them

## Questions

Out of the [3 options listed](https://github.com/docker-mailserver/docker-mailserver/issues/2713#issuecomment-1581106813) in the tracking issue regarding account provisioning, which is going to be the best solution?

The first option is implemented here, using a static userdb.

The third option sounds like it would be the optimal one however there have been [points made](https://github.com/docker-mailserver/docker-mailserver/issues/2713#issuecomment-1581688454) against this.
- However Mailcow [has recently implemented oauth2](https://mailcow.email/posts/2023/mailcow-idp/) (in their [nightly branch](https://github.com/mailcow/mailcow-dockerized/tree/nightly) and they have gone about it in an interesting way.
- It appears they are adding userinfo and querying that using the already supplied credentials on a cron job or equivalent. This sounds like a very generic and simple solution to the problem and one I believe we can _yoink_.

## Type of change

- New feature (non-breaking change which adds functionality)
- This change requires a documentation update

## PR Checklist:


polarathene:
I'll possibly be able to contribute feedback / review sometime this month. The other maintainers are also a bit busy, so just a heads-up if it takes a bit of time 😅 (_I've got a bit of a backlog to address first in DMS and elsewhere_)

thechubbypanda:
I'm realising that there is really no way to solely use oauth2 if you also want to use any generic imap client like thunderbird or almost everything on mobile. I'm aware this was mentioned in the original issue.
The solution for this sounds like finding a way to do password authentication via the oauth2 provider or requiring a second form of auth such as ldap or static...
Authentik for example does have an LDAP provider but the whole idea for this was to not have to use LDAP.

Essentially this is just an update comment saying that I'm looking into proxying passwords to the auth backend. Unfortunately it doesn't look promising as [grant password](https://doc.dovecot.org/configuration_manual/authentication/oauth2/#proxy-with-password-grant) has [been deprecated](https://oauth.net/2/grant-types/password/)... although there might be some [light at the end of the tunnel](https://stackoverflow.com/a/65672931/4896472).

Until the next update o7

polarathene:
Hey sorry for the silence about this, caught up working elsewhere on the project.

While working with the LDAP support, I have been wondering if the OAuth2 / OIDC support is meant to be an account provisioner or just an alternative authentication feature DMS could support?

---

Probably nothing below that you don't already know; However, I am forgetful this is more of a refresher for me when I return to this 😝 

I don't have time to go over the existing discussion for the feature or this PR right now, but flow wise we have:
- Inbound mail delivered to Postfix, which queries Dovecot if the user exists (UserDB) to accept the mail.
- If user exists and mail is accepted, Dovecot will create the mailboxes in `/var/mail` as configured to do so.
- Mail submission to Postfix requires authentication, which again delegates to Dovecot (_even with LDAP by default, Dovecot will auth agains the LDAP provider_). This is not the case when using `ENABLE_SASLAUTHD=1`, which could be to an IMAP provider (_again Dovecot or any other alternative_) or LDAP (_no Dovecot indirection to LDAP this time_).
- Mail retrieval from Dovecot storage (`/var/mail`) involves authentication, which like with Postfix delegating auth to Dovecot, involves Dovecots PassDB, which may have overlap with the UserDB but doesn't need to.
- Both PassDB and UserDB can be configured to have multiple backends, trying one and falling back to another lookup if unsuccessful.

With the current LDAP and FILE provisioners, both auth and accounts are configured to use that provisioner. `ENABLE_SASLAUTHD=1` is only configured for Postfix at present, thus Dovecot is either configured by FILE lookups or LDAP queries, never a mix.

---

> I'm realising that there is really no way to solely use oauth2 if you also want to use any generic imap client like thunderbird or almost everything on mobile. I'm aware this was mentioned in the original issue.
> The solution for this sounds like finding a way to do password authentication via the oauth2 provider or requiring a second form of auth such as ldap or static...

From what I understand, the OAuth2 backend may be sufficient if mail clients are restricted to being compatible with that, however some deployments may want to support traditional means. Some IDPs like Keycloak can support both OIDC and LDAP IIRC, which should allow centralizing the account management/provisioning through that service, as you don't really want to try keep two different account sources in sync (_eg: static_) 😅 

Thus DMS probably needs to rethink the account management support, refactoring how we configure Dovecot PassDB/UserDB? I can try give that some thought after I bring LDAP support into better shape.

thechubbypanda:
Good consolidation, yeah. Most of that rings true, I have some ideas but life has caught up with me and I'll be busy for a while. I'll keep an eye out for mentions here in case others want some help taking over if I'm not back to finish this soon.

thechubbypanda:
The thing stopping oAuth2 from being it's own fully blown auth provider is that lack of query-ability. As such, for now it will likely just be a layer over top for something like roundcube to make use of, but still falling back to something like LDAP.

I'm thinking of just reducing the scope of this PR to that for now and creating another for the final form;

The final form being implementing adapters for different oAuth2 provider APIs such as Authentik, KeyCloak, etc. to allow user querying. This would allow for oAuth2 to be its own fully blown auth provider.

MohammedNoureldin:
Being its own fully OAuth2 provider and falling back to other providers like LDAP is the best solution and so should it be, IMO. Is there any time estimation when you will working on it?

thechubbypanda:
I have some time this weekend, hopefully to get the minimum use case done in this PR.

MohammedNoureldin:
Great! I am excited! Let me know if I can somehow support.

MohammedNoureldin:
The comment here is interesting:

https://github.com/nextcloud/mail/issues/5933#issuecomment-1026565034

Would this change the way you are going to implement this PR? It just attracted my attention, because it is talking about the most modern way of authenticating E-Mail users.

@polarathene could you please also take a look on it?

polarathene:
> @polarathene could you please also take a look on it?

I don't know what you want me to take from that. I don't see it adding anything helpful to the discussion here.

- This PR is tackling the Dovecot support for OAuth2 via `passdb`. That handles authentication.
- Dovecot `userdb` is the other part with actual accounts provisioned. SSO provides the authentication, and sometimes additional data about a user, but AFAIK not something we use to manage Dovecot specific account data such as quota limit, maildir, etc.

You would provision accounts and have same auth as we do currently AFAIK. This PR will provide an alternative authentication mechanism. There is another PR contributing Dovecot auth via Lua scripts as another alternative (_but won't be officially supported by DMS, only via a community guide in the docs_), it's been discussed there that in future DMS may need to rework this support to better separate the account provisioning choice and enabled auth backends for Dovecot to use.

It may be a couple months before I have time to spare for any of this. Backlog is already full.

thechubbypanda:
I'd agree with @polarathene, that doesn't really change the situation. For now it's just going to be enabling oauth2 in dovecot as an alternate login method allowing for things like SSO webmail login.

The Lua scripts sound ideal for the potential next phase of this implementation which is to make oAuth2 a full blown auth provider. They would be used to retrieve user lists for mailbox creation etc.

polarathene:
> Unfortunately most mail clients implement specific workarounds for each major provider and not a generic redirect that we could use here.

It's been a while since I looked into OAuth / OIDC, but I remember OIDC resolving that issue (_or it may have been OAuth2?_). I do recall before that I would see different libraries for languages that would maintain a massive list of OAuth provider integration, but AFAIK that became much simpler with OIDC (OpenID Connect).

Perhaps one option is to add a bridge of sorts that provides the OAuth2 providers if needed, but ideally OIDC if that has good compatibility these days, then for integrating with Dovecot I suppose it might be possible to route the OAuth2 feature through that internal DMS service? 🤔 

I've not worked on such myself before, so it may be too much work as-is.

I assume Dovecots mention of a [Proxy Backend](https://doc.dovecot.org/configuration_manual/authentication/oauth2/#proxy) for OAuth2 might be related to this for offloading the auth to a separate service / container.

---

If it's easier taking the per provider route, then this would be less valuable to DMS and might be better suited similarly to the Lua support via our docs? We can add some integration and some provider support, but rather than DMS maintaining providers it may be better as a BYO guide from the docs? 🤷‍♂️ 

That way the community can contribute provider support more easily and the providers that DMS users express enough interest in can be part of DMS.

**EDIT:** I assume this is about the [Dovecot OAuth2 "Backends" examples](https://doc.dovecot.org/configuration_manual/authentication/oauth2/#backend)?

---

**EDIT:** [Here's an old discussion about using KeyCloak for the OIDC service](https://www.reddit.com/r/selfhosted/comments/omwb2j/comment/hrtznxj/) to connect with any other services, and Dovecot integration (_which has some interesting information about their implementation_). KeyCloak would be too big for DMS to bundle, so they'd need to use a separate image, but I believe there are some Go or Rust based projects that might be an option if there is any benefit bundling such a service into DMS.

I may have misunderstood that interaction and it changes nothing to the Dovecot OAuth2 provider issue 😅 

**UPDATE:** Oh, there is already [OIDC support in Dovecot](https://doc.dovecot.org/configuration_manual/authentication/oauth2/#openid-discovery)? 🤔
- It was part of the [`2.3.16` release in Aug 2021](https://dovecot.org/pipermail/dovecot-news/2021-August/000463.html), but OIDC Discovery was apparently broken until a fix in [`2.3.21` release in Sep 2023](https://dovecot.org/mailman3/archives/list/dovecot-news@dovecot.org/thread/KYDR7WWPEQOBZA3IA4NL5XDSLODZLG6N/)
- If that changes anything, perhaps we should hold-off until we can get a newer Dovecot release?

thechubbypanda:
Ok I have this working as of now with authentik and roundcube. The user has to exist beforehand in dms for it to work as we are not providing dovecot a userdb, only a passdb. Ergo, either ldap or file provisioner must be used and the user added beforehand.
Regular login using username and password still works.
The 2 passwords are not linked in any way, they only grant access to the same account.

@MohammedNoureldin can you confirm this before I add tests and document?
My compose.override.yml looks like this for reference:
```
services:
  mailserver:
    image: docker.io/library/mailserver-testing:ci
    hostname: mail.domain.com
    environment:
      ENABLE_OAUTH2: 1
      OAUTH2_CLIENT_ID: verySecretId
      OAUTH2_CLIENT_SECRET: verySecretSecret
      OAUTH2_INTROSPECTION_URL: https://authentik.domain.com/application/o/userinfo/
```
and a roundcube configuration reference can be found [here](https://goauthentik.io/integrations/services/roundcube/)

MohammedNoureldin:
> Ok I have this working as of now with authentik and roundcube.

Sounds great! Thank you!

> The user has to exist beforehand in dms for it to work as we are not providing dovecot a userdb, only a passdb. Ergo, either ldap or file provisioner must be used and the user added beforehand.

Should this be the case? I mean, in case of LDAP, we are also not storing anything related to the user, but rather fetching the data directly from the LDAP server, and the mails are accordingly stored in the folder of the user. Why can't this be the case with OIDC? I mean it should be somehow possible to log in to the mail also for the first time. Or did I misunderstood the point here?

polarathene:
> Should this be the case? I mean, in case of LDAP, we are also not storing anything related to the user, but rather fetching the data directly from the LDAP server

LDAP service has the account information already, including mail specific data such as aliases for that user.

Postfix and Dovecot talks to the LDAP service to retrieve this information.

> and the mails are accordingly stored in the folder of the user.

That is a folder Dovecot manages responsibility for, nothing to do with LDAP.

To accept the mail, the account must exist from the provisioner. Dovecot UserDB provides that information, which for LDAP involves talking to the LDAP service to verify IIRC.

> Why can't this be the case with OIDC?
> I mean it should be somehow possible to log in to the mail also for the first time.

You can authenticate with Dovecot via the PassDB, this then optionally provides some information to UserDB IIRC.

Even if that works fine, if you have not logged in, then how should Dovecot handle the situation when Postfix provides some new mail for the account? Dovecot needs to know it exists to accept the mail, so how will it check via UserDB?

OIDC is just providing authentication AFAIK. Dovecot needs to know not only if a user exists, but any additional information such as mail storage location, quota limit, etc. Many of these we can provide defaults on. I know that OIDC allows supporting some complimentary user data, but not sure if it's that flexible for the Dovecot related user data.

LDAP support has Dovecot perform a lookup query to the LDAP service, usually with an admin account or similar permissions to grant querying user data.

LDAP also supports auth (PassDB), like our file provisioner. But AFAIK OIDC is only able to support the PassDB functionality, not UserDB, hence why you need something else providing that data to Dovecot.

thechubbypanda:
Yeah, nail on the head. 
We need to be able to query the oAuth2 provider irrespective of login events to fetch all the users so that in the case a new user is added, they can start receiving emails almost immediately. If we relied on login events, the user would have to log in before being able to receive any emails.

The solution to this is to use the API of the oAuth2 provider and create a userdb from that. As I mentioned previous, the lua scripts sound ideal to implement this but I'd have to look into it. A job for the next PR ;)

thechubbypanda:
Docs are done and could use a review. Tests are next, a job for this evening.

MohammedNoureldin:
> LDAP service has the account information already, including mail specific data such as aliases for that user.

> OIDC is just providing authentication AFAIK. Dovecot needs to know not only if a user exists, but any additional information such as mail storage location, quota limit.

OIDC can also provide the necessary information Postfix and Dovecot needs to work as same as LDAP does. That can be done via the **claims**, in which you can pass any information you specify in the scopes. These claims will be part of the ID or access token the user receives. So I see no reason that prevents OIDC to work also when the account has never been used before. Or am I missing something here?

Let us take for example this case: I have all my users created and managed only in my identity provider, e.g. KeyCloak or Authentik. Without any kind of federation, so I have no LDAP or anything else, only my IdP storing my users. How will the user be able to log in to his mail and use it? @thechubbypanda or did I misunderstood something here?

> Docs are done and could use a review, hopefully alongside your testing @MohammedNoureldin?

I will be able to test this on my K8s cluster next week, I just messed up my whole test env and I need a few days to set up it properly again. Is that OK?

thechubbypanda:
Yes, requesting claims works but it's per user if I'm not mistaken. I.e. that user has to log in for their information to be made available to the OAuth2 client (in this case, dovecot).
The problem is when a user exists in Keylcloak/Authentik, postfix (via dovecot) has no knowledge of them until they log in for the first time. Hence even if you create the user, they cannot receive email until they log in first which is what we're trying to avoid.
The OAuth2 spec doesn't provide a way for a client to get a list of users which would be needed for email receipt unrelated to login behaviour. LDAP can be queried for a list of users which is why it does not face this problem.

With regards to testing, that's fine. I doubt this will be merged any time soon due to @polarathene's lack of availability. I'm starting on testing now.

thechubbypanda:
I intend to run this with Authentik holding all the users, and 2 providers, one LDAP and one Oauth2. DMS will use the LDAP provider primarily but where the OAuth2 one can be used, it will prefer that; for things like roundcube that can http redirect and deal with tokens.

MohammedNoureldin:
@thechubbypanda OK I started grasping the issue. Let me just summarize here:

Considering that I have an IdP with LDAP federation (the users are actually stored in LDAP directory, but reached also via the IdP).

For all users that already have logged in at least once to their mail inbox, here the general SSO case can be applied, as my mail server will use the tokens received from the users to authenticate them, without any issue or limitation.

**For the users that have never yet logged in to their mail inbox, the users can still log in to their accounts normally and get the their inbox created**, and from now on, the users will be able to receive emails, but before logging in for the first time, the users will not be able to receive mails. Is that right? could you confirm the **bold** sentence particularly please?

An acceptable solution IMHO is as follows:

- When someone tries to send a mail to a user that already logged in before and has an inbox, then all is fine and the user will receive the mail.
- When someone tries to send a mail to a user what has never logged in before and has no inbox yet. The mail server might somehow be able to make an LDAP query to the LDAP server to make it possible mail server checks against LDAP to check if the user exists, which have the ability for new users, that have never logged in, to receive emails.
- Whatever the user is, as long as he exists, he should be able to login using OIDC without relying on LDAP (from the user-experience point of view). If the server needs at some point to reach LDAP server or not, it does not matter, as long as the user uses only OIDC.

Is this the case at the moment?

thechubbypanda:
Your "acceptable solution" seems correct. The only point that may be in contention is the last.

DMS must be connected to LDAP and the OAuth2 provider. It will provision the users by querying LDAP.
This process will be completely invisible to the user unless they decide to login with their LDAP username and password for something like a mobile app.
The user should be able to solely use SSO to access webmail at the least.

MohammedNoureldin:
> DMS must be connected to LDAP and the OAuth2 provider. It will provision the users by querying LDAP.
This process will be completely invisible to the user

@thechubbypanda thank you! That is exactly how it should be. Great work! I am doing my best to be able to test this ASAP.

Maybe after some thorough testing and docs, @polarathene might be able to prioritize this. At least to be part of the next release. But let us get it done, then we can argue about merging it.

thechubbypanda:
Testing is proving to be a real nightmare. While Roundcube logging in via imap works perfectly, doing the same manually with netcat results in:
> oauth2 failed: Introspection failed: Username 'user1@locahost.localdomain' did not match 'user1@localhost.localdomain'
I can't find any reason for this. I even went as far as to try (and fail) to compile and run dovecot locally to reproduce the problem.

polarathene:
> I can't find any reason for this

It reminds me of an issue with LDAP auth, at least when using SASLAuthD instead of Dovecot directly.

I haven't addressed it yet as I still have an LDAP related change PR to wrap up and merge. But IIRC:
- Postfix would output a similar confusing auth failure as the domain-part was appended implicitly if the auth username was only the local-part.
- This would then be sent to SASLAuthD service with the auth credentials, and then attempt to delegate auth via LDAP service directly or via Dovecot (_which would then delegate again to LDAP_).
- These are the two supported SASLAuthD mechanisms, `ldap` and `rimap`. `rimap` works fine against Dovecot because it's supervisord config uses the `-r` flag to strip (_or was it keep?_) the "realm" (_`@domain-part` is treated as a "realm"_) when provided IIRC, while `ldap` mechanism would not do this. Non-issue without SASLAuthD for LDAP as it'd have Postfix auth directly.

So to Postfix, you had two separate values with the same email address (_`auth-user@realm` vs `sender@domain`_), they just happen to share the same value but are distinct in purpose.
- If your auth username was `user1@localhost.localdomain`, then you'd be successful with `user1` as the auth username instead (_otherwise the other way around_).
- The failure is that the auth user lookup for the mail account failed to return a result, eg LDAP query was given input that the default configured query along with SASLAuthD `ldap` mechanism config failed to resolve to verify the account exists.

---

I have a bunch of notes on this and a branch with the change I think, it's just stuck in a VM that is a bit annoying to access right now since I'm low on storage and when I installed WSL2 my hypervisor was unable to launch the VM with the expected virtualization features (_Hyper-V now has those exclusively until I uninstall WSL2_). Honestly Windows has been quite problematic vs my previous Linux host, causing a fair amount of friction and frustration 😩 

If I'm too slow to sorting out the fix though, you can lookup the saslauthd supervisor config file we have, it's got rimap and ldap handled there. The ldap one just needs to match the usage of `-r` IIRC. This may additionally require a change to Postfix `main.cf` (_or related LDAP startup scripts we have_), or the Postfix LDAP `.cf` files (_which my open PR is refactoring_). It is a breaking change I wanted to bundle as a follow-up to the current LDAP PR, not sure if I'll find time to get that in for v13 yet.

Supposedly you'll avoid the issue if you can adjust auth username like I suggested. Otherwise feel free to stub out the test with some commented pseudo-code and we can review everything else 👍

thechubbypanda:
And it's working :tada: 
@polarathene @MohammedNoureldin

polarathene:
Awesome! I'll try allocate some time to review in the weekend, but most likely later this month as my plate is fairly full elsewhere right now 😅 

For DMS I need to try wrap up my LDAP PR to try squeeze it in for v13 as it's a breaking change, so your PR here is lower in priority as it should be fine for 13.1 follow-up release 👍 

Other maintainers may chime in with review in the meantime though 🙏 It would be good if @MohammedNoureldin can give it a spin too :)

MohammedNoureldin:
Sorry for my delay. I will test it thoroughly as soon as possible. I just have some issues, and thus I still need some days yet to be able to test it.

polarathene:
> And @polarathene how's your backlog looking?

Current queue is:
- Vector syslog codec PR review 2nd round of feedback
- `config-rs` (Rust crate) PRs to merge
- DMS PRs (_I missed the v13 deadline for my PR which is a breaking change, so yours will have higher priority now_)

Presently time for DMS is mostly related to user support and the recent release concerns.

I haven't forgotten your PR though! I should have time for it in December 👍

polarathene:
I am probably not going to be able to give this as decent of a review as I'd like to.
- I don't want to see this PR become stale however and appreciate your constant rebasing to prevent that.
- At a glance through the diff nothing stands out to me as a concern, I'm very happy with the quality of the contribution.
- I'd have appreciated someone else to chime in with feedback or testing the feature out, but like my pending review we're all strapped for time 😓 

I'll try to give this a better look tomorrow, but so far LGTM 👍 

---

Are there any remaining concerns or support needs for the functionality of this feature contribution? It may not tick all the boxes, but still provides value right?

docs/content/config/advanced/auth-oauth2.md
```diff
@@ -0,0 +1,60 @@
+---
+title: 'Advanced | Basic OAuth2 Authentication'
+---
+
+## Introduction
+
+**Warning** - This is only a supplement to the existing account provisioners; FILE and LDAP. Accounts must still be created using the `setup` command or added to the LDAP directory respectively. Reasoning for this can be found in [#3480](https://github.com/docker-mailserver/docker-mailserver/pull/3480). Future iterations on this feature may allow it to be a full account provisioner.
```
polarathene, line 7:
```suggestion
warning "This is only a supplement to the existing account provisioners"

    Accounts must still be managed via the configured [`ACCOUNT_PROVISIONER`][env::account-provisioner] (FILE or LDAP).

    Reasoning for this can be found in [#3480][gh-pr::oauth2]. Future iterations on this feature may allow it to become a full account provisioner.

[gh-pr::oauth2]: https://github.com/docker-mailserver/docker-mailserver/pull/3480
[env::account-provisioner]: ../environment.md#account_provisioner
```

docs/content/config/advanced/auth-oauth2.md
```diff
@@ -0,0 +1,60 @@
...
+
+For now, this adds the ability for a 3rd party application such as Roundcube to authenticate with DMS (dovecot) using a token obtained from an OAuth2 provider instead of passing passwords around.
```
polarathene, line 9:
```suggestion
The present OAuth2 support provides the capability for 3rd-party applications such as Roundcube to authenticate with DMS (dovecot) by using a token obtained from an OAuth2 provider, instead of passing passwords around.
```

docs/content/config/advanced/auth-oauth2.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Example (Authentik & Roundcube)
+
+???+ example "Authentik"
+    1. Create a new OAuth2 provider
+    2. Note the client id and client secret
+    3. Set the allowed redirect url to `https://roundcube.domain.com/index.php/login/oauth` (obviously changing your domain as needed)
```
polarathene, line 16:
```suggestion
    3. Set the allowed redirect url to the equivalent of `https://roundcube.example.com/index.php/login/oauth` for your RoundCube instance.
```

docs/content/config/advanced/auth-oauth2.md
```diff
@@ -0,0 +1,60 @@
...
+    # Specify the OAuth2 client ID
+    OAUTH2_CLIENT_ID=<insert client id here>
+
+    # empty => verySecretSecret
+    # Specify the OAuth2 client secret
+    OAUTH2_CLIENT_SECRET=<insert client secret here>
+
+    # empty => https://oauth2.domain.com/userinfo/
+    # Specify the user info endpoint URL of the oauth2 provider
+    OAUTH2_INTROSPECTION_URL=https://authentik.domain.com/application/o/userinfo/
```
polarathene, line 38:
```suggestion
    OAUTH2_INTROSPECTION_URL=https://authentik.example.com/application/o/userinfo/
```

docs/content/config/advanced/auth-oauth2.md
```diff
@@ -0,0 +1,60 @@
...
+
+???+ example "Roundcube `oauth2.inc.php` ([documentation](https://github.com/roundcube/roundcubemail/wiki/Configuration))"
+    ```php
+    $config['oauth_provider'] = 'generic';
+    $config['oauth_provider_name'] = 'Authentik';
+    $config['oauth_client_id'] = '<insert client id here>';
+    $config['oauth_client_secret'] = '<insert client secret here>';
+    $config['oauth_auth_uri'] = 'https://authentik.domain.com/application/o/authorize/';
+    $config['oauth_token_uri'] = 'https://authentik.domain.com/application/o/token/';
+    $config['oauth_identity_uri'] = 'https://authentik.domain.com/application/o/userinfo/';
```
polarathene, line 49:
```suggestion
    $config['oauth_auth_uri'] = 'https://authentik.example.com/application/o/authorize/';
    $config['oauth_token_uri'] = 'https://authentik.example.com/application/o/token/';
    $config['oauth_identity_uri'] = 'https://authentik.example.com/application/o/userinfo/';
```

mailserver.env
```diff
@@ -419,6 +418,26 @@ ENABLE_GETMAIL=0
...
+
+# empty => verySecretId
+# Specify the OAuth2 client ID
+OAUTH2_CLIENT_ID=
+
+# empty => verySecretSecret
+# Specify the OAuth2 client secret
+OAUTH2_CLIENT_SECRET=
+
+# empty => https://oauth2.domain.com/userinfo/
```
polarathene, line 437:
```suggestion
# empty => https://oauth2.example.com/userinfo/
```

target/dovecot/dovecot-oauth2.conf.ext
```diff
@@ -0,0 +1,3 @@
+client_id = verySecretId
+client_secret = verySecretSecret
+introspection_url = https://oauth2.domain.com/userinfo/
```
polarathene, line 3:
```suggestion
introspection_url = https://oauth2.example.com/userinfo/
```

polarathene (changes requested):

thechubbypanda:
No worries, I'm glad you got around to it.

@MohammedNoureldin Any chance of a test?

> Are there any remaining concerns or support needs for the functionality of this feature contribution? It may not tick all the boxes, but still provides value right? As per the docs addition:

I can't think of any concerns. 

Support _should_ be minimal but I'm happy for people to @ me on any problems. 

The main thing I expect to see asked about is "full" OAuth support as opposed to this minimal set. I do want to contribute something to get to that point but it'll be a while I reckon. 

This does indeed provide a slight bit of value and at the very least, it lays some groundwork for a full provider.

polarathene:
Requested review from other maintainers, I usually miss some things 😅 

---

At @georglauterbach did you want to add this into v13.1? I see you already started preparing for that release with a shorter than normal freeze period.

Can queue for v14.

georglauterbach:
I will review this :) 

---

This will be queued for v14.0.0 because I want to merge #3407.

polarathene (changes requested):

.gitignore
```diff
@@ -3,6 +3,7 @@
 #################################################
 
 .env
+compose.override.yaml
```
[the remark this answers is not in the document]
thechubbypanda: I used it extensively while developing to switch the image to the testing one, add env vars etc. Just means I didn't have to worry about not committing the original. Can remove if you want.

test/tests/serial/mail_with_oauth2.bats
```diff
@@ -0,0 +1,81 @@
...
+
+  # Link the test containers to separate network:
+  # NOTE: If the network already exists, test will fail to start.
+  docker network create "${DMS_TEST_NETWORK}"
+
+  # Setup local oauth2 provider service:
+  docker run --rm -d --name "${CONTAINER2_NAME}" \
+    --hostname "${FQDN_OAUTH2}" \
+    --network "${DMS_TEST_NETWORK}" \
+    --user "$(id -u):$(id -g)" \
```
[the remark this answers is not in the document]
thechubbypanda: This just ensures that the files in the developer's worktree don't get chowned to root.

test/tests/serial/mail_with_oauth2.bats
```diff
@@ -0,0 +1,81 @@
...
+    --volume "${REPOSITORY_ROOT}/test/config/oauth2/:/app/" \
+    --expose 80 \
```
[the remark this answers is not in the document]
thechubbypanda: (I thought this was necessary until now) Yeah mainly just for documentation purposes then. Happy to remove?

.gitignore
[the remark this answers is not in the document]
polarathene: Yeah, we've got a `docker run` dev image with some ENV override supported via a `make` command for similar needs of a maintainer now: https://github.com/docker-mailserver/docker-mailserver/pull/3663

I'll be adding a compose variant that is flexible for similar usage, might be worth considering the override ignore for that 🤔

I'm fine with keeping the ignore 👍

test/tests/serial/mail_with_oauth2.bats
```diff
@@ -0,0 +1,81 @@
...
+  _run_until_success_or_timeout 20 sh -c "docker logs ${CONTAINER2_NAME} 2>&1 | grep 'Starting server'"
+
+  #
+  # Setup DMS container
+  #
+
+  local ENV_OAUTH2_CONFIG=(
+    --env ENABLE_OAUTH2=1
+    --env OAUTH2_CLIENT_ID=mailserver
+    --env OAUTH2_CLIENT_SECRET=ah_yes___secret
```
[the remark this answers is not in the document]
thechubbypanda: This literally just mocks the tiny portion of the process that we actually need to test with this implementation.
A full blown sidecar.

I'll add some documentation for this and the python script.

test/tests/serial/mail_with_oauth2.bats
```diff
@@ -0,0 +1,81 @@
...
+    --env OAUTH2_INTROSPECTION_URL=http://provider.example.test/
+  )
+
+  local ENV_SUPPORT=(
+    --env PERMIT_DOCKER=container
+  )
```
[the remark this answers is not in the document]
thechubbypanda: Nope, dumping it.

thechubbypanda:
@polarathene I'm getting this error on a bunch of tests and also `make run-local-instance`:
```
Error: The directory named as part of the path /var/log/mail/mail.log does not exist in section 'program:postgrey' (file: '/etc/supervisor/conf.d/supervisor-app.conf')
For help, use /usr/bin/supervisord -h
make: *** [Makefile:37: run-local-instance] Error 2
```
Any idea what I might have screwed up?

polarathene:
> Any idea what I might have screwed up?
> I've tried rebuilding everything completely from scratch on master; same error :/

I haven't run DMS locally for a while, it might be related to that [makefile change](https://github.com/docker-mailserver/docker-mailserver/pull/3663) or similar change introduced for v13.1 🤔  

---

I can see that the postgrey service stdout/stderr config is not quite the same as the other services:

https://github.com/docker-mailserver/docker-mailserver/blob/3adb53eb1213c8848fdf44cf2343392c1817d094/target/supervisor/conf.d/supervisor-app.conf#L86-L87

Building the image with that config updated to look like others that create their own log file based on service name might resolve it? (_this is what every other service has defined_):

https://github.com/docker-mailserver/docker-mailserver/blob/3adb53eb1213c8848fdf44cf2343392c1817d094/target/supervisor/conf.d/supervisor-app.conf#L77-L78

@georglauterbach any input? I recalled bringing this log path up in [review feedback for a docs PR on log locations](https://github.com/docker-mailserver/docker-mailserver/pull/3640/files#r1393325507). The path is a symlink created in the `Dockerfile`, while all other services in this supervisor config use the supervisor location.

I would have expected others to have experienced the error too and reported it 😅

polarathene:
Great job addressing the feedback, I especially appreciate the added comments for test maintenance ❤️ 

---

I'm a bit annoyed at Github, not sure why it's swallowed up my main review comment, curious if that has happened in prior PRs now 😭 (_unless others can still see it and I can't for some reason. I only have this screenshot from an existing tab that was thankfully left open in another window_)

image

---

I guess that is due to the rebase of commits? 🤷‍♂️ 

Looks like all checklist items were addressed 🎉 (_except for changelog entry, which is due after v13.1 is released_)

polarathene (dismissed):

georglauterbach:
I will review after v13.1 is released.

mailserver.env
```diff
@@ -419,6 +418,26 @@ ENABLE_GETMAIL=0
...
+OAUTH2_CLIENT_ID=
+
+# empty => verySecretSecret
+# Specify the OAuth2 client secret
+OAUTH2_CLIENT_SECRET=
+
+# empty => https://oauth2.example.com/userinfo/
+# Specify the user info endpoint URL of the oauth2 provider. The trailing slash is MANDATORY (at least for Authentik)
+OAUTH2_INTROSPECTION_URL=
+
```
georglauterbach, line 440: This section needs to be added here as well: <https://docker-mailserver.github.io/docker-mailserver/latest/config/environment/>.

georglauterbach, line 440: This section needs to be added here as well: <https://docker-mailserver.github.io/docker-mailserver/latest/config/environment/>.

georglauterbach (changes requested):
LGTM 👍🏼 Very nice work! ❤️ 🚀 

You'll just need to update the doc's page on ENV variables and this can land in v14.0.0.

georglauterbach:
Feature-freeze is now over; as soon as final feedback is adresses, we can merge this PR immediately.

thechubbypanda:
Just gotten a touch busy. Will get to this in the next couple of days

docs/content/config/environment.md
```diff
@@ -605,9 +605,28 @@ Enable or disable `getmail`.
...
-#### LDAP
 
+#### OAUTH2
+
+##### ENABLE_OAUTH2
+
+- **empty** => OAUTH2 authentication is disabled
+- 1 => OAUTH2 authentication is enabled
+
+##### OAUTH2_CLIENT_ID
```
georglauterbach, line 616:
```suggestion
##### OAUTH2_CLIENT_ID

```

docs/content/config/environment.md
```diff
@@ -605,9 +605,28 @@ Enable or disable `getmail`.
...
+- **empty** => verySecretId
+- => Specify the OAuth2 client ID
 
+##### OAUTH2_CLIENT_SECRET
```
georglauterbach, line 620:
```suggestion
##### OAUTH2_CLIENT_SECRET

```

docs/content/config/environment.md
```diff
@@ -605,9 +605,28 @@ Enable or disable `getmail`.
...
+- **empty** => verySecretSecret
+- => Specify the OAuth2 client secret
+
+##### OAUTH2_INTROSPECTION_URL
```
georglauterbach, line 624:
```suggestion
##### OAUTH2_INTROSPECTION_URL

```

docs/content/config/environment.md
```diff
@@ -605,9 +605,28 @@ Enable or disable `getmail`.
...
+- **empty** => https://oauth2.example.com/userinfo/
+- => Specify the user info endpoint URL of the oauth2 provider. The trailing slash is MANDATORY (at least for Authentik)
+
+
```
georglauterbach, line 628:
```suggestion

```

georglauterbach (changes requested):

docs/content/config/environment.md
```diff
@@ -54,11 +54,11 @@ The Group ID assigned to the static vmail group for `/var/mail` (_Mail storage m
 
 Configures the provisioning source of user accounts (including aliases) for user queries and authentication by services managed by DMS (_Postfix and Dovecot_).
 
-User provisioning via OIDC is planned for the future, see [this tracking issue](https://github.com/docker-mailserver/docker-mailserver/issues/2713).
+User provisioning via OAUTH2 is planned for the future, see [this tracking issue](https://github.com/docker-mailserver/docker-mailserver/issues/2713). OAuth2.0 for machine-to-machine authentication (e.g. Roundcube login) has been implemented (see OAUTH2 section below).
 
 - **empty** => use FILE
 - LDAP => use LDAP authentication
-- OIDC => use OIDC authentication (**not yet implemented**)
+- OAUTH2 => use OAUTH2 authentication (**not yet implemented**)
```
georglauterbach, line 61: Wait - this does not seem to be valid anymore?

ref: https://github.com/docker-mailserver/docker-mailserver/pull/3480/files#diff-845c87c5018d576c8f4654f463f2d2a98a4a536fc2c0491cdc41d2680cc50413L65-L68

georglauterbach (changes requested):
One final question.

PS: please no more force-pushes at this point :)

docs/content/config/environment.md
thechubbypanda: This PR is only a supplement to a full Auth provider like LDAP/FILE. I'd like to contribute that next if I can.

georglauterbach: Oh, I read that wrong - all good!

georglauterbach (approved):

georglauterbach:
@polarathene  you can go ahead and merge if you approve as well :)

test/tests/serial/mail_with_oauth2.bats
```diff
@@ -0,0 +1,65 @@
...
+function teardown_file() {
+  docker rm -f "${CONTAINER1_NAME}" "${CONTAINER2_NAME}"
+  docker network rm "${DMS_TEST_NETWORK}"
+}
+
+
+@test "oauth2: imap connect and authentication works" {
+  _run_in_container_bash 'nc -w 1 0.0.0.0 143 < /tmp/docker-mailserver-test/auth/imap-oauth2-auth.txt'
+  assert_success
+}
```
polarathene, line 65: This seems like it always passes successfully, even when given bad data that fails to login?

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ All notable changes to this project will be documented in this file. The format
 
 > **Note**: Changes and additions listed here are contained in the `:edge` image tag. These changes may not be as stable as released changes.
 
+### Added
+
+  - **Basic OAuth2.0** 🎉
+    - You can now authenticate via apps like Roundcube using OAuth2.0 (LDAP/FILE provider still required)
+
```
polarathene, line 13:
```suggestion
### Features

  - **Authentication with OIDC / OAuth 2.0** 🎉
    - DMS now supports authentication via OAuth2 (_`XOAUTH` or `OAUTHBEARER`_) from capable services (_like Roundcube_).
    - This does not replace the need for an `ACCOUNT_PROVISIONER` (`FILE` / `LDAP`). Authentication (Dovecot PassDB) still requires an existing account (Dovecot UserDB) to be successful,.

```

polarathene (changes requested):
I have run out of time for the day, not able to share all my feedback for this PR but I have successfully run it locally with a different authentication server (Ory Hydra).

I think we should address some additional concerns first.

docs/content/config/environment.md
```diff
@@ -54,11 +54,11 @@ The Group ID assigned to the static vmail group for `/var/mail` (_Mail storage m
 
 Configures the provisioning source of user accounts (including aliases) for user queries and authentication by services managed by DMS (_Postfix and Dovecot_).
 
-User provisioning via OIDC is planned for the future, see [this tracking issue](https://github.com/docker-mailserver/docker-mailserver/issues/2713).
+User provisioning via OAUTH2 is planned for the future, see [this tracking issue](https://github.com/docker-mailserver/docker-mailserver/issues/2713). OAuth2.0 for machine-to-machine authentication (e.g. Roundcube login) has been implemented (see OAUTH2 section below).
```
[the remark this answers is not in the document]
thechubbypanda: Your understanding appears to match mine :)

I would prefer "_supplementary_ authentication method". "Alternative" might imply replacement.

docs/content/config/environment.md
thechubbypanda: I forgot that I removed that. Might it be better to leave it in `start-mailserver.sh` to solidify the difference between the supplement and a full OAUTH2 provider?

docs/content/config/environment.md
```diff
@@ -605,9 +605,26 @@ Enable or disable `getmail`.
...
+
+- => Specify the OAuth2 client ID
 
+##### OAUTH2_CLIENT_SECRET
+
+- => Specify the OAuth2 client secret
+
+##### OAUTH2_INTROSPECTION_URL
+
+- => Specify the user info endpoint URL of the oauth2 provider. E.g. `https://oauth2.example.com/userinfo/`, where the trailing slash is MANDATORY (at least for Authentik)
```
polarathene, line 618:
```suggestion
- => Specify the user info endpoint URL of the oauth2 provider. E.g. `https://oauth2.example.com/userinfo/`, where the trailing slash is MANDATORY (at least for Authentik)

```

polarathene: I'm not sure if we need to document the trailing slash. I believe this will vary by service.

It's a common feature with reverse proxies to support enforcing it or supporting a route without it. That could probably front Authentik too in a way that the reverse proxy supports either and handles any explicit requirement when handing off to Authentik.

It's also not always the `userinfo/` endpoint as I've found.

---

I'll come back to this after this review pass as I'm more inclined to drop these ENV.

docs/content/config/environment.md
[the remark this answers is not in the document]
polarathene: Supplementary is fine, and matches the referenced docs page.

Although I'm not quite sure of the value in this section documenting that entire tip. It's mostly a note that an `ACCOUNT_PROVISIONER` is planned 🤷‍♂️ Remainder is related to the current authentication feature state, but we're basically repeating what is covered there.

docs/content/config/environment.md
polarathene: There is no `OIDC` provider, just a stub that's been lingering for a while 🤷‍♂️ 

Since it's already there, I don't mind leaving it as is, but OAUTH2 is primarily meant to be for authorization AFAIK, with OIDC absorbing/extending that spec into proper authentication support?

Revert:

```suggestion
OIDC => use OIDC authentication (**not yet implemented**)
```

docs/content/config/advanced/auth-oauth2.md
```diff
@@ -0,0 +1,78 @@
...
+        OAUTH2_CLIENT_ID=<insert client id here>
+
+        # empty => verySecretSecret
+        # Specify the OAuth2 client secret
+        OAUTH2_CLIENT_SECRET=<insert client secret here>
+
+        # empty => https://oauth2.example.com/userinfo/
+        # Specify the user info endpoint URL of the oauth2 provider
+        OAUTH2_INTROSPECTION_URL=https://authentik.example.com/application/o/userinfo/
+        ```
```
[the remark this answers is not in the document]
polarathene: Apply this.

polarathene (changes requested):
Submitting this feedback now since you're actively pushing updates 😅 

- Thanks for addressing the test-case issue 🥳 
- Can you confirm that you do not need `client_id` and `client_secret` in the Dovecot config for your Roundcube + Authentik usage?
- Please share your opinion on dropping the ENV in favor of a config file (_volume mount or compose inline config mount_). Supporting the ENV prefix replacement may still be acceptable, the LDAP PR is reworking that support though.

docs/content/config/environment.md
thechubbypanda: I think the core is just auto-configuration using the `.well-known` path. At it's core, it's still OAuth2.0.
I don't have much of a leg to stand on since my knowledge is still somehow minimal despite doing all this work haha. So if you still think OIDC is better, we can change it.

test/tests/serial/mail_with_oauth2.bats
thechubbypanda: Ok this was very weird, I'm sure I tested it a while ago. Nevertheless, you are correct.

I've fixed it, it's a bit janky. If I don't have the first netcat and sleep, it complains that the auth system isn't ready. Unless you have a way of waiting for that specifically to be ready, we're stuck with whatever this is haha.

thechubbypanda:
@polarathene Thanks for the very in depth review. A second pair of eyes really sees everything.

I would never have figured out that the client id and secret weren't needed haha

target/dovecot/dovecot-oauth2.conf.ext
```diff
@@ -0,0 +1,3 @@
+client_id = verySecretId
+client_secret = verySecretSecret
+introspection_url = https://oauth2.example.com/userinfo/
```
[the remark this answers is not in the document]
thechubbypanda: 1 & 2 sorted, looking into the compose feature now. I didn't even know this was a thing.

docs/content/config/environment.md
```diff
@@ -54,11 +54,16 @@ The Group ID assigned to the static vmail group for `/var/mail` (_Mail storage m
...
+    Presently DMS supports OAuth2 only as an supplementary authentication method. 
+    - A third-party service must provide a valid token for the user which Dovecot validates with the authentication service provider. To enable this feature reference the [OAuth2 configuration example guide][docs::auth::oauth2-config-guide].
+    - User accounts must be provisioned to receive mail via one of the supported `ACCOUNT_PROVISIONER` providers.
+    - User provisioning via OAUTH2 is planned for the future, see [this tracking issue](https://github.com/docker-mailserver/docker-mailserver/issues/2713).
+[docs::auth::oauth2-config-guide]: ./advanced/auth-oauth2.md
 
 - **empty** => use FILE
 - LDAP => use LDAP authentication
-- OIDC => use OIDC authentication (**not yet implemented**)
+OIDC => use OIDC authentication (**not yet implemented**)
```
thechubbypanda, line 66:
```suggestion
- OIDC => use OIDC authentication (**not yet implemented**)
```

target/dovecot/dovecot-oauth2.conf.ext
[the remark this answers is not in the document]
thechubbypanda: You know what, especially since it's now just a one line file... This might be a good shout.

I don't have the experience to make the call though, I'll leave it up to you. Let me know if you'd like me to set it up using compose instead.

polarathene:
> @polarathene Thanks for the very in depth review. A second pair of eyes really sees everything.
> 
> I would never have figured out that the client id and secret weren't needed haha

No worries, took a few review passes to notice things myself 😅 

Finally found some time to try the feature locally and attempt to get a setup working with a service. I was going to use Authentik but it seemed to add more friction for what I wanted (avoiding browser interactions). However as I experienced so far, to get a token associated to a user login I had to manually interact via a browser, so the mock server is probably necessary for automated tests.

---

> Also, sorry if you're getting ping after ping on this. It must be exhausting. I'm still trying to figure out the most efficient way to do this open source business

No worries, I am used to constantly getting notifications on my phone from Github 😆 

You've been a great contributor here!

---

> Have I hit everything you've mentioned?

:shrug:

I'll go through it again when I come back to this PR today. I'm fairly happy with the PR, just the ENV concern to address.

docs/content/config/environment.md
polarathene: The trailing slash comment can be left to the guide docs. I'm suggesting with `eg` since I think that's used elsewhere even though it's not entirely correct compared to `e.g.:` 😅 

```suggestion
- => Specify the user info endpoint URL of the oauth2 provider (_eg: `https://oauth2.example.com/userinfo/`_)
```

polarathene (changes requested):

polarathene:
> Have I hit everything you've mentioned?

Everything is good now 🎉 

- [Docs could do with a little more work](https://github.com/docker-mailserver/docker-mailserver/pull/3480#discussion_r1443522926), but that's not required for this PR.
- [I do want to discard the reliance on ENV approach](https://github.com/docker-mailserver/docker-mailserver/pull/3480#discussion_r1441105098) -- unless there is a good reason to justify supporting another config with all/portion of it's settings mapped to ENV that we maintain/document and need to be aware of accidentally introducing conflicts against 😓 

---

**UPDATE:** There was a recent restructuring PR that was merged, the imap test file needs to be adjusted to that new location for the test to pass.

polarathene:
Below is probably only useful to me, but includes references related to the Ory Hydra service I verified the PR against, which uses slightly different config with Dovecot. I'll probably include another example in our docs with this provider.

Additionally for any future maintainers that may land here if troubleshooting, includes some expected outputs when using the `nc` test approach.

## Reference

### Ory Hydra config snippets (Incomplete for reproduction)

```
# /etc/dovecot/dovecot-oauth2.conf.ext

introspection_url = http://ory-hydra-example--hydra:4445/admin/oauth2/introspect
introspection_mode = post
username_attribute = sub
```

```python
# encode-token.py
# Small portion of the PR to encode the access token for use with Dovecot IMAP login against a real auth service instance
# Script is run with `token` below updated accordingly, and base64 output added to the a1 content of the Dovecot imap nc input

import base64
token = "ory_at_HVpNn6Rv_SXKG9-PBhKAMHdQlXpDu2LJGAimK_H9s58.vg08L4veMj_eaU2bj3TzclEIBClr2E0vNrkckjwDPrk"

# This is the string the user-facing client (e.g. Roundcube) should send via IMAP to Dovecot.
# We include the user and the above token separated by '\1' chars as per the XOAUTH2 spec.
xoauth2 = base64.b64encode(f"user=foo@bar.com\1auth=Bearer {token}\1\1".encode("utf-8"))

# If changing the user above, use the new output from the below line with the contents of the AUTHENTICATE command in test/test-files/auth/imap-oauth2-auth.txt
print("XOAUTH2 string: " + str(xoauth2))
```

```bash
# Manual approach, used when nc wasn't working well:
telnet localhost imap

# Querying the userinfo/ endpoint on the Docker host with the access token:
curl -H 'Authorization: Bearer ory_at_HVpNn6Rv_SXKG9-PBhKAMHdQlXpDu2LJGAimK_H9s58.vg08L4veMj_eaU2bj3TzclEIBClr2E0vNrkckjwDPrk' 'http://auth.localhost/userinfo'

# Querying the introspect/ endpoint:
curl -v -X POST -d 'token=ory_at_HVpNn6Rv_SXKG9-PBhKAMHdQlXpDu2LJGAimK_H9s58.vg08L4veMj_eaU2bj3TzclEIBClr2E0vNrkckjwDPrk' 'http://auth.localhost/admin/oauth2/introspect'
```

---

### IMAP auth logs output

Successful but expected `username_attribute = email` field was not returned in response received.
- This was due to my basic Ory Hydra service from their docs which assigned the account address as the Subject property (`sub: foo@bar.com`) for the account identity, where a proper setup would have that in `email` and `sub` would be a UUID-like value:

```
Jan  3 07:02:51 mail dovecot: auth: passwd-file(foo@bar.com,127.0.0.1,<aXMLNAUOwot/AAAB>): unknown user (SHA1 of given password: b58473) - trying the next passdb
Jan  3 07:02:51 mail dovecot: auth: Error: oauth2(foo@bar.com,127.0.0.1,<aXMLNAUOwot/AAAB>): oauth2 failed: Introspection failed: No username returned
Jan  3 07:05:30 mail dovecot: imap-login: Disconnected: Inactivity (client didn't finish SASL auth, waited 159 secs): user=<foo@bar.com>, method=XOAUTH2, rip=127.0.0.1, lip=127.0.0.1, secured, session=<aXMLNAUOwot/AAAB>
```

No DMS account provisioned, although the auth process itself would have been successful, fails due to Dovecot UserDB lookup not finding a matching account:

```
Jan  3 07:08:50 mail dovecot: auth: passwd-file(foo@bar.com,127.0.0.1,<Y2t4SQUOjLx/AAAB>): unknown user (SHA1 of given password: 89a7c4) - trying the next passdb
Jan  3 07:08:50 mail dovecot: auth: passwd-file(foo@bar.com,127.0.0.1,<Y2t4SQUOjLx/AAAB>): unknown user (SHA1 of given password: 89a7c4)
Jan  3 07:08:50 mail dovecot: auth: Error: passwd-file(foo@bar.com,127.0.0.1,<Y2t4SQUOjLx/AAAB>): user not found from userdb
Jan  3 07:08:50 mail dovecot: imap(5831): Error: auth-master: login: request [4270325761]: Login auth request failed: Authenticated user not found from userdb, auth lookup id=4270325761 (auth connected 0 msecs ago, request took 0 msecs, client-pid=5813 client-id=1)
Jan  3 07:08:50 mail dovecot: imap-login: Disconnected: Internal login failure (pid=5813 id=1): user=<foo@bar.com>, method=XOAUTH2, rip=127.0.0.1, lip=127.0.0.1, mpid=5831, secured, session=<Y2t4SQUOjLx/AAAB>
```

---

```console
# Successful:
$ nc -w 1 0.0.0.0 143 < ./imap-oauth2-login.txt

* OK Waiting for authentication process to respond..
* OK [CAPABILITY IMAP4rev1 SASL-IR LOGIN-REFERRALS ID ENABLE IDLE LITERAL+ AUTH=PLAIN AUTH=LOGIN AUTH=OAUTHBEARER AUTH=XOAUTH2] Dovecot ready.
a0 OK NOOP completed.
a1 OK [CAPABILITY IMAP4rev1 SASL-IR LOGIN-REFERRALS ID ENABLE IDLE SORT SORT=DISPLAY THREAD=REFERENCES THREAD=REFS THREAD=ORDEREDSUBJECT MULTIAPPEND URL-PARTIAL CATENATE UNSELECT CHILDREN NAMESPACE UIDPLUS LIST-EXTENDED I18NLEVEL=1 CONDSTORE QRESYNC ESEARCH ESORT SEARCHRES WITHIN CONTEXT=SEARCH LIST-STATUS BINARY MOVE SNIPPET=FUZZY PREVIEW=FUZZY PREVIEW STATUS=SIZE SAVEDATE LITERAL+ NOTIFY SPECIAL-USE QUOTA] Logged in
* FLAGS (\Answered \Flagged \Deleted \Seen \Draft)
* OK [PERMANENTFLAGS ()] Read-only mailbox.
* 0 EXISTS
* 0 RECENT
* OK [UIDVALIDITY 1704265956] UIDs valid
* OK [UIDNEXT 1] Predicted next UID
a2 OK [READ-ONLY] Examine completed (0.005 + 0.000 + 0.007 secs).
* BYE Logging out
a3 OK Logout completed (0.001 + 0.000 secs).
```

Earlier review feedback for the initial `nc` test where it always passed was caught with malformed base64 still providing an exit status of `0`:

```console
# Token base64 input is invalid intentionally:
$ nc -w 1 0.0.0.0 143 < ./imap-oauth2-login.txt

* OK Waiting for authentication process to respond..
* OK [CAPABILITY IMAP4rev1 SASL-IR LOGIN-REFERRALS ID ENABLE IDLE LITERAL+ AUTH=PLAIN AUTH=LOGIN AUTH=OAUTHBEARER AUTH=XOAUTH2] Dovecot ready.
a0 OK NOOP completed.
a1 BAD Invalid base64 data in initial response
a2 BAD Error in IMAP command received by server.
* BYE Logging out
a3 OK Logout completed.

# Exist status still considered successful:
echo $?
0

# Related logs:
Jan  3 07:18:43 mail dovecot: auth: xoauth2(?,127.0.0.1,<+QPPbAUOhqV/AAAB>): Invalid base64 data in initial response
Jan  3 07:18:43 mail dovecot: imap-login: Disconnected: Aborted login by logging out (sent invalid base64 in response): user=<>, method=XOAUTH2, rip=127.0.0.1, lip=127.0.0.1, secured, session=<+QPPbAUOhqV/AAAB>
```

polarathene (dismissed):

target/scripts/startup/setup.d/oauth2.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+function _setup_oauth2() {
+  _log 'debug' 'Setting up OAUTH2'
+
+  # Enable OAuth2 PassDB (Authentication):
+  sed -i -e '/\!include auth-oauth2\.conf\.ext/s/^#//' /etc/dovecot/conf.d/10-auth.conf
```
casperklein, line 7:
```suggestion
  sedfile -i -e '/\!include auth-oauth2\.conf\.ext/s/^#//' /etc/dovecot/conf.d/10-auth.conf
```

If for some reason the replacement fails, an error is logged.

casperklein (approved):

polarathene (dismissed):
Will probably merge this tomorrow 👍

target/dovecot/dovecot-oauth2.conf.ext
```diff
@@ -0,0 +1,2 @@
+introspection_url =
+introspection_mode = auth
```
polarathene, line 2:
```suggestion
introspection_url =
# Dovecot defaults:
introspection_mode = auth
username_attribute = email
```

polarathene (approved):

casperklein (dismissed):

thechubbypanda:
Damn, looks like I left a lot to be desired over the course of this PR haha. Thanks for fixing it up. Looking forward to using this soon :)

georglauterbach (approved):

test/tests/serial/mail_with_oauth2.bats
```diff
@@ -0,0 +1,66 @@
...
+  docker rm -f "${CONTAINER1_NAME}" "${CONTAINER2_NAME}"
+  docker network rm "${DMS_TEST_NETWORK}"
+}
+
+
+@test "oauth2: imap connect and authentication works" {
+  # An initial connection needs to be made first, otherwise the auth attempt fails
+  _run_in_container_bash 'nc -vz 0.0.0.0 143'
+
+  _nc_wrapper 'auth/imap-oauth2-auth' '-w 1 0.0.0.0 143'
```
polarathene, line 64: This will need an update after an open PR updating the method is merged.

```suggestion
  _nc_wrapper 'auth/imap-oauth2-auth.txt' '-w 1 0.0.0.0 143'
```

polarathene (changes requested):
I can apply these 👍 

Another PR may get merged before this one, then this will follow with these changes directly after :)

MohammedNoureldin:
Please execute my delay. I had some private circumstances and thus I could not do any progress. Now I available again and would be very glad to help! I see that you made a lot of progress here. Well done guys! Please let me know if I can support somehow.

test/config/oauth2/provider.py
```diff
@@ -0,0 +1,56 @@
...
+token = "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJodHRwOi8vcHJvdmlkZXIuZXhhbXBsZS50ZXN0OjgwMDAvIiwic3ViIjoiODJjMWMzMzRkY2M2ZTMxMWFlNGFhZWJmZTk0NmM1ZTg1OGYwNTVhZmYxY2U1YTM3YWE3Y2M5MWFhYjE3ZTM1YyIsImF1ZCI6Im1haWxzZXJ2ZXIiLCJ1aWQiOiI4OU4zR0NuN1M1Y090WkZNRTVBeVhNbmxURFdVcnEzRmd4YWlyWWhFIn0.zuCytArbphhJn9XT_y9cBdGqDCNo68tBrtOwPIsuKNyF340SaOuZa0xarZofygytdDpLtYr56QlPTKImi-n1ZWrHkRZkwrQi5jQ-j_n2hEAL0vUToLbDnXYfc5q2w7z7X0aoCmiK8-fV7Kx4CVTM7riBgpElf6F3wNAIcX6R1ijUh6ISCL0XYsdogf8WUNZipXY-O4R7YHXdOENuOp3G48hWhxuUh9PsUqE5yxDwLsOVzCTqg9S5gxPQzF2eCN9J0I2XiIlLKvLQPIZ2Y_K7iYvVwjpNdgb4xhm9wuKoIVinYkF_6CwIzAawBWIDJAbix1IslkUPQMGbupTDtOgTiQ"
+
+# This is the string the user-facing client (e.g. Roundcube) should send via IMAP to Dovecot.
+# We include the user and the above token separated by '\1' chars as per the XOAUTH2 spec.
+xoauth2 = base64.b64encode(f"user=user1@localhost.localdomain\1auth=Bearer {token}\1\1".encode("utf-8"))
+# If changing the user above, use the new output from the below line with the contents of the AUTHENTICATE command in test/test-files/auth/imap-oauth2-auth.txt
+print("XOAUTH2 string: " + str(xoauth2))
```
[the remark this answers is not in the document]
thechubbypanda: I think this is due to a lack of documentation on Authentik's part (the service I've been testing with). I have been barely aware that it actually has an `introspect` endpoint until now. The `userinfo` endpoint _worked_ so I had no reason to venture elsewhere.

Hence I will concur, the access token appears to be more correct.

polarathene:
> Well done guys! Please let me know if I can support somehow.

You can give it a try if you like. Let me know if you run into any problems, like the docs not providing enough information to setup for you.

How important the ENV (`OAUTH_*`) is to you (_since this only introduces one explicitly_), vs just providing the config directly (_slightly less convenient than it needs to be atm_).

This feature is otherwise approved and will get merged for v14 🎉 

---

One concern that might be worth raising awareness on is with this integration, if your third-party auth provider let's you configure the `email`, and there isn't any trust in that being verified by the user, that authentication to Dovecot grants access to the Dovecot account, even if you're not the actual owner of the address 🤔 

Usually there should be additional information like `email_verified`, and I think Dovecot can additionally match that, or select other claims to source a value as part of passdb attributes (_eg: some `id-hash` that Dovecot UserDB provides and the third-party auth provider needs to include in returned claims to establish trust. However that's not really something we can offer and would need custom integration AFAIK_ 😅 )

polarathene (approved):
Approving and merging for v14.

If I have time to revise an update to the oauth2 docs / ENV before then, I will but at this stage I may not have the time for it :(

---

Thanks for contributing this new feature! ❤️

MohammedNoureldin:
Thanks guys. I have been trying to get it working with NextCloud, still I have some points to clarify.

Just one point regarding its release, this is a sensitive feature, I would release it in a separate minor version and mention that it is under review rather than major version with many other big changes. What do you think?

polarathene:
> this is a sensitive feature

Why is it sensitive? You need to opt-in regardless.

---

> I would release it in a separate minor version (13.3.0) ASAP

Unfortunately our release process isn't well suited for that if we've already merged changes that would be semver incompatible:
- PRs get merged to `master` branch, that publishes the latest `:edge` image.
- `master` branch gets tagged which triggers a release. We do this via the GH releases feature which will automatically tag the latest commit and trigger CI publishing workflows once we publish a release manually / explicitly.
- To support what you propose requires adapting the CI to handle release branches, which would be nice but no one has had enough time to warrant prioritizing that.

Thus once breaking changes are added to `master`, we can only release a major version. Likewise once features are released we cannot push a bug fix release.

Looking at the [current `CHANGELOG.md`](https://github.com/docker-mailserver/docker-mailserver/blob/ce6ebcc0219eca96ead8d758973f14d94dbb7efe/CHANGELOG.md), if @georglauterbach doesn't consider the rspamd update a breaking change, it does look like we could publish a v13.3 release.

We have improved our review process to add changelog entries before merging now, so much of that friction to releases is now gone 💪 

---

> mention that it is under review rather than major version with many other big changes.

Presently the only big change due for v14 is the base image upgrade from Debian Bullseye to Debian Bookworm, and all the package updates that brings.

I would like to get the LDAP breaking change in as well, and I may be able to get that resolved in time.

Other than that, I don't think we have anything else that should be breaking / big.

---

> What do you think?

It would be good to get v13.3 out for the `jaq` ARM fix, so those users get notified of the v14 release (_presently broken with v13.2_).

I'm comfortable with v13.3 if the other maintainers are too.

georglauterbach:
I think the Rspamd changes are not breaking as we define it. I'm fine with releasing v13.3.

polarathene:
> I said sensitive because it is about authentication in general, which is the first part the user need to be able to interact with the email. And thus any small issue may potentially end up that users are not being able to login, for example. That is why it seemed sensitive to me.

Fair. In that case do you want the breaking changes LDAP PR to be it's own major release instead of bundled with the base image upgrade?

georglauterbach:
I'd prefer to have it in one release; less work on our side 🤣

polarathene:
> less work on our side 🤣

I think the reasoning was the release separation benefits maintainers too when it comes to troubleshooting issues related to releases that introduced breaking changes.

The LDAP changes are large enough that it'd be easier to troubleshoot when you rule out the changes the base image upgrade PR introduces. Thus the separate releases is probably less work maintenance wise😅

polarathene:
> I get dovecot: auth error failed operations error.

Set [`ENABLE_QUOTAS=0`](https://docker-mailserver.github.io/docker-mailserver/v13.2/config/environment/#enable_quotas) to opt-out of this Dovecot feature. It may help avoid some confusion if it introduces additional logs about `dovecot: auth error`, although with LDAP that quotas feature isn't available AFAIK, so probably not relevant to you 😅 

---

> for some reason I am not able to login even using LDAP, it is not related to this change as I am trying with 12.1.0

Are you saying you cannot login with LDAP? Nothing should have changed there and our tests pass auth:

https://github.com/docker-mailserver/docker-mailserver/blob/2bf52342500726e86362ee1ea5954b7d96533d32/test/tests/serial/mail_with_ldap.bats#L378-L384

It's possibly related to your LDAP config and DMS related settings for query filters (_and `result` attribute_). You do not need to enable SASLAuthd, just ensure configuration for Postfix and Dovecot is correct, refer to how we've done so in our LDAP test (_related LDAP config for users is by [`.ldif` files here](https://github.com/docker-mailserver/docker-mailserver/blob/2bf52342500726e86362ee1ea5954b7d96533d32/test/config/ldap/openldap/ldifs/02_user-email.ldif)_):

https://github.com/docker-mailserver/docker-mailserver/blob/2bf52342500726e86362ee1ea5954b7d96533d32/test/tests/serial/mail_with_ldap.bats#L74-L90

---

I've already verified that this feature is compatible with our LDAP auth tests, so I know it works :)

The LDAP config will be changed and the related docs page completely rewritten to accommodate it. Another change will affect SASLAuthd LDAP integration with Postfix as I'm not happy with it introducing an inconsistency.

As this is getting offtopic, feel free to open an issue/discussion when you have more information to share and ping me there 👍

georglauterbach:
> I think the reasoning was the release separation benefits maintainers too when it comes to troubleshooting issues related to releases that introduced breaking changes.
> 
> The LDAP changes are large enough that it'd be easier to troubleshoot when you rule out the changes the base image upgrade PR introduces. Thus the separate releases is probably less work maintenance wise 😝

I see, makes sense. Then let's please merge #3403 first because I really want to finally have Debian 12 (to change some things in my personal setup).

MohammedNoureldin:
I have no clue anymore. I am not able to test this because, as mentioned, for some reason I cannot connect to LDAP anymore, I tested also 12.1.0, so it is not related to the changes made here. Though, any help would be great so I can do extensive test to this feature. To keep this PR clean I opened a new discussion for it. https://github.com/orgs/docker-mailserver/discussions/3785

MohammedNoureldin:
> Thanks guys. I have been trying to get it working with NextCloud, still I have some points to clarify.

As an update, after a lot of tests I was NOT able to get it working with NextCloud. However, it turned out (I got an official statement regarding this), that Nextcloud Mail does not forward the access tokens apparently the the mail server. So I stopped working on it at the moment.

PR #3483, opened 2023-08-15
tests: Refactor LDAP tests to current conventions

polarathene:
# Description

Part of the big effort to migrate all tests to use the better helpers and conventions that support much better test maintenance (and a step towards better coverage).

- Migration has been staged out via smaller commits for easier review of diff. Each commit contains a detailed message of changes it covers.
- Various `TODO` have been added for future improvements.
- Due to size of refactor (_and relocation of files, which aren't as friendly with `git blame` history when changes are also applied_), I've held off adjusting some email addresses (`.tld` / `.localdomain`) until after this PR is merged.

<details>
<summary>All commit messages for easier reference</summary>

**tests: Switch to setup helper and conventions**

**tests: Adapt run command to new conventions**
- We have two helper methods with implicit `CONTAINER_NAME` reference, which is a bit more DRY and improves readability.
- `wc -l` + `assert_output 1` converted to use helper `_should_output_number_of_lines 1`
- `DOMAIN` var changed from `my-domain.com` to local testing domain `example.test`.

**tests: Refactor `setup_file()`**
- Test wide ENV defined at the top
- OpenLDAP build and run logic grouped together. Added notes, network alias and tty not required.
- Extracted out special LDAP Postfix/Dovecot ENV into separate array. LDAP specific provisioning / auth ENV also included, with comments + linebreak to better group related ENV.
- Likewise additional ENV to support test cases has been extracted to a separate array with additional comments for context.
- Those two arrays are expanded back into the main `CUSTOM_SETUP_ARGUMENTS` that configure hostname and network for the DMS container.

**tests: Refactor the LDAP account table query testcase**
- Covers 3 accounts to test from LDAP.
  - 2 are the same query against users/aliases/groups tables in Postfix, only differing by account queried (and expected as returned result).
  - 1 separate test to ensure a difference in config is supported correctly.
- Extracted repeated test logic into a helper method.
- Added additional context in comments about the creation source of these LDAP accounts and their related Postfix config / interaction. Direct reference to special case PR (_since `git blame` will be less useful_).

**tests: Use iteration for `grep` setting checks**
- More DRY approach. With a bit more helpful failure context via `assert_output` (_and only grepping the key_). Simpler to grok what's being covered.

**tests: DRY test email delivery**
- A bit more verbose with the new helper method. `test-email.txt` template is only used by the LDAP test, as is the `sendmail` command.
- Helper will take two args to support the testcases, but at a later date should be refactored to be consistent with the `_send_email()` helper (_which presently uses `nc` that is required for plain-text mail/auth, otherwise we'd have used `openssl`, bigger refactor required_).

**tests: Slight revisions and relocating testcases**
- Dovecot quota plugin testcase revised to check files exist instead of rely on `ls` failure.
- Moved Postfix quota plugin testcase into prior dovecot testcase for quota plugin check. Better error output by only querying the `smtpd_recipient_restrictions` setting (_which should be the only one configured for the service_).
- Moved the saslauthd and pflogsumm testcases (_no changes beyond revised comments_) above the `ATTENTION` comment, and one testcase below the comment that belonged to that group.

**tests: Simplify openldap `docker build` command**
- `--no-cache` was creating a new image on the Docker host each time the test is run. Shouldn't be any need to build without cache.
- No need to use `pushd` + `popd`, can just provide the path context directly, and the `./Dockerfile` is an implicit default thus `-f` not required either.
- Additionally removed the old `checking` prefix from testcase names.

**tests: Move LDAP specific config into `test/config/ldap/`**
- No changes to any of these config files, just better isolation as not relevant to any other tests.
- Section heading in `setup_file()` added to distinguish the remainder of the function is dedicated to the DMS container setup.
- Comment providing some context about the `mv` to maintainers, this should be done after defaults are initialized but before starting up the container.

</details>

## Type of change

## Checklist:


georglauterbach (commented):
Looks good at first glance, but I will give it a more thorough review as soon as I have a few minutes to spare again. I will not be blocking on this though, so earlier merging without my review is fine for me.

test/tests/serial/mail_with_ldap.bats
```diff
@@ -1,245 +1,314 @@
...
+    --hostname "${FQDN_MAIL}"
+    --network "${DMS_TEST_NETWORK}"
+
+    "${ENV_LDAP_CONFIG[@]}"
+    "${ENV_SUPPORT[@]}"
+  )
+
+  # Set default implicit container fallback for helpers:
+  export CONTAINER_NAME
+  CONTAINER_NAME=${CONTAINER1_NAME}
```
georglauterbach, line 86: Should be fine to just assign immediately, right?

```suggestion
  export CONTAINER_NAME=${CONTAINER1_NAME}
```

test/tests/serial/mail_with_ldap.bats
```diff
@@ -1,245 +1,314 @@
...
+  _run_in_container grep "\$mail_plugins quota" /etc/dovecot/conf.d/10-mail.conf
+  assert_failure
+  _run_in_container grep "\$mail_plugins imap_quota" /etc/dovecot/conf.d/20-imap.conf
+  assert_failure
 
-  run docker exec mail_with_ldap /bin/sh -c "grep 'start_tls = no' /etc/postfix/ldap-groups.cf"
+  # Dovecot Quota config only present with disabled extension:
+  _run_in_container [ -f /etc/dovecot/conf.d/90-quota.conf ]
+  assert_failure
+  _run_in_container [ -f /etc/dovecot/conf.d/90-quota.conf.disab ]
```
georglauterbach, line 207: While absolutely fine, I in Bash we should prefer

```suggestion
  _run_in_container [[ -f /etc/dovecot/conf.d/90-quota.conf ]]
  assert_failure
  _run_in_container [[ -f /etc/dovecot/conf.d/90-quota.conf.disab ]]
```

This works because `[[` is a valid program, but _actually_, I'd try

```suggestion
  _run_in_container_bash '[[ -f /etc/dovecot/conf.d/90-quota.conf ]]'
  assert_failure
  _run_in_container_bash '[[ -f /etc/dovecot/conf.d/90-quota.conf.disab ]]'
```

to be absolutely sure nothing breaks.

test/tests/serial/mail_with_ldap.bats
```diff
@@ -1,245 +1,314 @@
...
+  assert_output --partial 'Sender address rejected: not owned by user'
 }
 
-@test "checking dovecot: postmaster address" {
-  run docker exec mail_with_ldap /bin/sh -c "grep 'postmaster_address = postmaster@${FQDN_LOCALHOST_A}' /etc/dovecot/conf.d/15-lda.conf"
-  assert_success
-}
+@test "saslauthd: ldap smtp authentication" {
+  # Requires ENV `PERMIT_DOCKER=container`
+  _run_in_container_bash 'nc -w 5 0.0.0.0 25 < /tmp/docker-mailserver-test/auth/sasl-ldap-smtp-auth.txt'
```
georglauterbach, line 259: Should be

```suggestion
  _send_email 'auth/sasl-ldap-smtp-auth' '-w 5 0.0.0.0 25'
```

test/tests/serial/mail_with_ldap.bats
```diff
@@ -1,245 +1,314 @@
...
- run docker exec mail_with_ldap /bin/sh -c "grep '\$mail_plugins quota' /etc/dovecot/conf.d/10-mail.conf"
- assert_failure
- run docker exec mail_with_ldap /bin/sh -c "grep '\$mail_plugins imap_quota' /etc/dovecot/conf.d/20-imap.conf"
- assert_failure
- run docker exec mail_with_ldap ls /etc/dovecot/conf.d/90-quota.conf
- assert_failure
- run docker exec mail_with_ldap ls /etc/dovecot/conf.d/90-quota.conf.disab
- assert_success
-}
+  _run_in_container_bash 'openssl s_client -quiet -connect 0.0.0.0:465 < /tmp/docker-mailserver-test/auth/sasl-ldap-smtp-auth.txt'
```
georglauterbach, line 262: Just as a reminder for the future: `_send_email` should probably get a "ports" option and choose between `openssl s_client` and `nc` dynamically. Or, simpler, we provide a new helper `_send_mail_openssl()`.

test/tests/serial/mail_with_ldap.bats
```diff
@@ -1,245 +1,314 @@
...
 
+  # Link the test containers to separate network:
   # NOTE: If the network already exists, test will fail to start.
   docker network create "${DMS_TEST_NETWORK}"
 
-  docker run -d --name ldap_for_mail \
+  # Setup local openldap service:
+  # NOTE: Building via Dockerfile is required? Image won't accept read-only if it needs to adjust permissions for bootstrap files.
+  # TODO: Upstream image is no longer maintained, may want to migrate?
+  docker build -t dms-openldap test/config/ldap/docker-openldap/
```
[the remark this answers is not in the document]
polarathene: Yeah, I got a bit familiar with it while working through the refactoring of the test.

I tried a few other openldap images, but ran into problems carrying over the config `.ldif` + `.schema` we have. I'll give it a bit more time today but call it quits if I can't figure out a suitable replacement.

test/tests/serial/mail_with_ldap.bats
polarathene: Yep, initially I thought I was going to need it for the other container 😅 

There's the other LDAP PR open at present that does require introducing some additional DMS variants to test an ENV though.

test/tests/serial/mail_with_ldap.bats
polarathene: Sure, this was just referencing the manual SSL test which tested files in the container the same way.

test/tests/serial/mail_with_ldap.bats
polarathene: We could, but instead of switching between `openssl` / `nc` it might also be good to just use a CLI tool that sends mail and can support all ports?

Preferably it'd ingest a different mail template, I'd prefer we handle the auth ones with the account password in plaintext, instead of the encoded (hashed? base64?) format exchanged presently for `AUTH LOGIN`. We have an issue with that for the skipped test, and the `.ldif` files have `SSHA` hashed passwords themselves, so I've no clue what they were originally 😅

test/tests/serial/mail_with_ldap.bats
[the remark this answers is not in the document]
georglauterbach: 👍🏼

test/tests/serial/mail_with_ldap.bats
georglauterbach: Yeah, that might also be a good idea. But no rush, it works fine for now

georglauterbach (approved):

PR #3484, opened 2023-08-16
scripts: add wrapper to update Postfix configuration safely

### LINKED ISSUE #3482: bug report: Postfix `main.cf:smtpd_milters` appends `$rspamd_milter` on restarts
### 📝 Preliminary Checks

### 👀 What Happened?

A [user has identified a bug](https://github.com/orgs/docker-mailserver/discussions/3481#discussioncomment-6720861) similar to one recently resolved by https://github.com/docker-mailserver/docker-mailserver/pull/3380 where a setting is updated by appending a value, but repeated restarts do not check if it was already appended (_internal config is expected to be deterministic at startup, not modified from prior restart_).

https://github.com/docker-mailserver/docker-mailserver/blob/8f971713360c4620046c7f223fe66df9ea800253/target/scripts/startup/setup.d/security/rspamd.sh#L141

This `sed` approach has been used a fair bit and may result in similar bugs elsewhere.

---

### Internal state persisting across restarts is a known concern

This isn't too surprising; maintainers are well aware of DMS not playing well with restarts. We've given explicit advice in many issues to [properly restart DMS](https://github.com/docker-mailserver/docker-mailserver/issues/3462#issuecomment-1669120153) via either:
- `docker compose down` (_`ctrl +c` alone is not equivalent_) + `docker compose up`
- `docker compose up --force-recreate` (_no `down` command required_).

Our docs also try raise awareness in [Usage guide](https://docker-mailserver.github.io/docker-mailserver/v12.1/usage/#get-up-and-running) (_perhaps should also mention in [Debugging docs](https://docker-mailserver.github.io/docker-mailserver/v12.1/config/debugging/)_):

image

Internal configs being changed may not be deterministic with current scripts otherwise. While this approach won't help if any volume mounted configs (_that aren't copied internally_) were modified by DMS, but we tend to have less assumptions by scripts in those cases to avoid the problem.

### 👟 Reproduction Steps

_No response_

### 🐋 DMS Version

12.1 / edge

N/A

_No response_

### 📜 Relevant log output

_No response_

_No response_

georglauterbach:
# Description

Add functionality to safely and securely append to, prepend to or init options in `/etc/postfix/main.cf`.

Fixes #3482

## Open Tasks

- [ ] Manually check whether `docker restart` does not append the string twice.

## Type of change

## Checklist:


polarathene (dismissed):
I think I suggested somewhere using `postconf` instead of `sed` to set / append values for `main.cf`, but can't recall if there were any issues with that (_beyond the 2 sec delay to read a setting when not using the `mtime` workaround_),

Regardless this helper is awesome (and well documented!), great work 😁

target/scripts/startup/setup.d/security/rspamd.sh
```diff
@@ -132,13 +132,12 @@ EOF
 }
 
 # Adjust Postfix's configuration files. We only need to append Rspamd at the end of
-# `smtpd_milters` in `/etc/postfix/main.cf`.
+# `smtpd_milters` in `${POSTFIX_MAIN_CF}`.
```
casperklein, line 135: Was that change intended?

```suggestion
# `smtpd_milters` in `/etc/postfix/main.cf`.
```

target/scripts/helpers/postfix.sh
```diff
@@ -91,3 +94,58 @@ function _vhost_ldap_support() {
...
+# This function is not meant to be called directly from other scripts.
+# There are wrappers like `_append_to_option_in_postfix_main` that use
+# this function under the hood.
+function __add_to_or_update_postfix_main() {
+  if grep -E "^${1:?Option name is required}" "${POSTFIX_MAIN_CF}"; then
+    sed -i -E "${3:?Replacement string is required}" "${POSTFIX_MAIN_CF}"
+  else
+    # We have to use `sed` here, not `echo`, because
+    # the strings are escaped for `sed` already.
+    sed -i -E "\$a${1} = ${2:?String to append is required}" >>"${POSTFIX_MAIN_CF}"
```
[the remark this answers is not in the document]
casperklein: The `>>` are a typo?

target/scripts/startup/setup.d/dmarc_dkim_spf.sh
```diff
@@ -15,11 +15,8 @@ function _setup_opendkim() {
 
     _log 'trace' "Adding OpenDKIM to Postfix's milters"
     postconf 'dkim_milter = inet:localhost:8891'
-    # shellcheck disable=SC2016
-    sed -i -E                                            \
-      -e '/\$dkim_milter/! s|^(smtpd_milters =.*)|\1 \$dkim_milter|g'     \
-      -e '/\$dkim_milter/! s|^(non_smtpd_milters =.*)|\1 \$dkim_milter|g' \
-      /etc/postfix/main.cf
+    _append_to_option_in_postfix_main 'smtpd_milters' "\$dkim_milter"
```
casperklein, line 18:
```suggestion
    _append_to_option_in_postfix_main 'smtpd_milters' '$dkim_milter'
```

target/scripts/startup/setup.d/dmarc_dkim_spf.sh
```diff
@@ -63,8 +60,7 @@ function _setup_opendmarc() {
     _log 'trace' "Adding OpenDMARC to Postfix's milters"
     postconf 'dmarc_milter = inet:localhost:8893'
     # Make sure to append the OpenDMARC milter _after_ the OpenDKIM milter!
-    # shellcheck disable=SC2016
-    sed -i -E '/\$dmarc_milter/! s|^(smtpd_milters =.*)|\1 \$dmarc_milter|g' /etc/postfix/main.cf
+    _append_to_option_in_postfix_main 'smtpd_milters' "\$dmarc_milter"
```
casperklein, line 63:
```suggestion
    _append_to_option_in_postfix_main 'smtpd_milters' 'dmarc_milter'
```

target/scripts/startup/setup.d/security/rspamd.sh
```diff
@@ -132,13 +132,12 @@ EOF
...
 function __rspamd__setup_postfix() {
   __rspamd__log 'debug' "Adjusting Postfix's configuration"
 
   postconf 'rspamd_milter = inet:localhost:11332'
-  # shellcheck disable=SC2016
-  sed -i -E 's|^(smtpd_milters =.*)|\1 \$rspamd_milter|g' /etc/postfix/main.cf
+  _append_to_option_in_postfix_main 'smtpd_milters' "\$rspamd_milter"
```
casperklein, line 140:
```suggestion
  _append_to_option_in_postfix_main 'smtpd_milters' '$rspamd_milter'
```

target/scripts/startup/setup.d/dmarc_dkim_spf.sh
polarathene: I think the shellcheck lint will complain (_or did with `grep` unless I did `"\$value"` instead of `'$value'`_), and may affect `sed`? I recall some similar problem when passing strings like this to a function that'd use `sed` with a variable reference...

@georglauterbach has checked that this is working as intended with repeat runs? I don't think we have any specific test for it. My earlier advice on the original approach to this was to just ensure we reset the config files via a copy at startup, avoiding the issue (_so long as the change is not applied again via a `setup` command or triggered by `check-for-changes.sh`_).

target/scripts/helpers/postfix.sh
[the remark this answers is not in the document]
georglauterbach: > I get that this line is adding the setting + value when the setting does not already exist in the file.
> 
> However, I'm not familiar with the `\$a` portion, what's `sed` doing here with `$a`?

This is telling `sed` to append to the end of the file: `$` match the end, `a` append.

> The `>>` are a typo?

Absolutely! Nice catch.

target/scripts/startup/setup.d/security/rspamd.sh
georglauterbach: Actually, yes. In `postfix.sh`, I introduced `readonly POSTFIX_MAIN_CF=/etc/postfix/main.cf` and in case this ever changes (most unlikely, I know), I wanted to use the value that will always hold. If you think this is not a good idea, we can revert, I do not insist on this change.

target/scripts/helpers/postfix.sh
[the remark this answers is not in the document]
georglauterbach:
```suggestion
    # We have to use `sed` here, not `echo`, because
    # the strings are escaped for `sed` already.
    sed -i -E "\$a${1} = ${2:?String to append is required}" "${POSTFIX_MAIN_CF}"
```

target/scripts/startup/setup.d/dmarc_dkim_spf.sh
georglauterbach: @casperklein  The quoting is done in order to not use `# shellcheck disable=SC...` above.

@polarathene I can double-check, but it may also be a good idea to ask @wligtenberg whether #3380 was properly solved for him. If so, this would indicate success. I would check nevertheless, but don't know when exactly.

target/scripts/startup/setup.d/dmarc_dkim_spf.sh
georglauterbach: See https://github.com/docker-mailserver/docker-mailserver/pull/3484#discussion_r1296999539

polarathene (approved):

georglauterbach:
It turned out there were problem with my method indeed, because the escaping was lost on the way. I have revised the functionality again, and checked whether the fix works. While I was at it, it compacted the code and removed superflous functions again.

I rebased on master, please provide a dedicated review again.

target/scripts/startup/setup.d/dmarc_dkim_spf.sh
georglauterbach: See https://github.com/docker-mailserver/docker-mailserver/pull/3484#issuecomment-1684998323

target/scripts/helpers/postfix.sh
```diff
@@ -91,3 +94,27 @@ function _vhost_ldap_support() {
...
+    if [[ ${3:-append} == 'append' ]]; then
+      local SED_STRING="/${NEW_VALUE}/! s|^(${KEY} *=.*)|\1 ${NEW_VALUE}|g"
+    else
+      local SED_STRING="/${NEW_VALUE}/! s|^(${KEY}) *= *(.*)|\1 = ${NEW_VALUE} \2|g"
+    fi
+    sed -i -E "${SED_STRING}" "${POSTFIX_MAIN_CF}"
+  else
+    echo "${KEY} = ${NEW_VALUE}" >>"${POSTFIX_MAIN_CF}"
+  fi
+}
```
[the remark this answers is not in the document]
polarathene: FWIW, there is similar functionality already implemented (_although more thoroughly_) in the DB helpers:

https://github.com/docker-mailserver/docker-mailserver/blob/5bada0a83b0875b36934e1de856b724979c83346/target/scripts/helpers/database/db.sh#L33-L37

https://github.com/docker-mailserver/docker-mailserver/blob/5bada0a83b0875b36934e1de856b724979c83346/target/scripts/helpers/database/db.sh#L63-L76

---

It could potentially support `main.cf`, but the overlap introduced here may be better as `db.sh` is already quite complicated and makes some assumptions like:

https://github.com/docker-mailserver/docker-mailserver/blob/5bada0a83b0875b36934e1de856b724979c83346/target/scripts/helpers/database/db.sh#L44-L54

https://github.com/docker-mailserver/docker-mailserver/blob/5bada0a83b0875b36934e1de856b724979c83346/target/scripts/helpers/database/db.sh#L104-L111

with special handling for managing keys with list values (_notably removal, which is less necessary elsewhere_):

https://github.com/docker-mailserver/docker-mailserver/blob/5bada0a83b0875b36934e1de856b724979c83346/target/scripts/helpers/database/db.sh#L126-L135

https://github.com/docker-mailserver/docker-mailserver/blob/5bada0a83b0875b36934e1de856b724979c83346/target/scripts/helpers/database/db.sh#L78-L96

target/scripts/helpers/utils.sh
```diff
@@ -4,6 +4,15 @@ function _escape() {
...
 }
 
+# Replaces a string so that it can be used inside
+# `sed` safely.
+#
+# @param ${1} = string to escape
+# @output     = prints the escaped string
+function _escape_for_sed() {
+  sed -E 's/[]\/$*.^[]/\\&/g' <<< "${1:?String to escape for sed is required}"
+}
```
polarathene, line 14: Likewise, a similar function exists in `db.sh`:

https://github.com/docker-mailserver/docker-mailserver/blob/5bada0a83b0875b36934e1de856b724979c83346/target/scripts/helpers/database/db.sh#L166-L173

I am not familiar enough with the two to compare them. I barely recall how I came up with mine here at the time 😅 

That said, I'd be a bit cautious replacing / changing the `db.sh` method, if I avoided escaping certain characters with the method there, it may have been for good reason, but unfortunately is lacking in test coverage AFAIK.

Personally my future plans were to write a rust CLI tool instead of dealing with bash scripts where alot of the issues related to this was from. But that would be a larger effort to pursue :(

polarathene (approved):
LGTM 👍 

- 1 suggestion to consider (_optional, but seems like an improvement_)
- References to `db.sh` shared as similar functionality exists.
  - Although my stance is to avoid adapting `db.sh` to support `main.cf` if it'd complicate `db.sh` further; the PR version here is simpler to grok and maintain in the context of managing `main.cf`.
  - I caution changing `db.sh` usage of it's `sed` helper, but it may not be equivalent as yours? 🤷‍♂️

target/scripts/helpers/utils.sh
georglauterbach: There really is a difference: mine seems to be more comprehensive

```console
$ sed 's/\([/&]\)/\\\1/g' <<< '$\\'
$\\
$ sed -E 's/[]\/$*.^[]/\\&/g' <<< '$\\'
\$\\\\
```

Doesn't matter though, as I will not be fiddling with `db.sh` anyways :D No time 🚀

target/scripts/helpers/postfix.sh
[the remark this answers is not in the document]
georglauterbach: I agree that working with `db.sh` would introduce overhead (which I can currently not afford). I ahve further simplified `_add_to_or_update_postfix_main` after seeing your suggestion. I also added some breathing room to make everything more readable.

georglauterbach:
@polarathene I applied a patch that I derived after looking at your suggestion. This should include the error handling, but keep the whole function as simple as possible. We also use `postfconf` now.

target/scripts/helpers/utils.sh
polarathene: Might be worth adding a comment to the `db.sh` one referencing the newer helper you've contributed.

I would be open to switching to yours, I'm just overly cautious about a change to `db.sh` messing something up 😅 That is code I do not enjoy troubleshooting.

target/scripts/helpers/postfix.sh
```diff
@@ -91,3 +91,32 @@ function _vhost_ldap_support() {
...
+    elif [[ ${3:-append} != 'append' ]]; then
+      _log 'err' "Action '${3}' in _add_to_or_update_postfix_main is unknown"
+      return 1
+    fi
+
+    sed -i -E "${SED_STRING}" /etc/postfix/main.cf
+  else
+    postconf "${KEY} = ${NEW_VALUE}"
+  fi
+}
```
polarathene, line 122: I still find the following easier to grok:
- All parameters are covered up at the top for the method
- The default `ACTION` is clear and much nicer than the conditonals that keep setting it as default, the `case` statement reads better IMO.
- Your log call is for `err` but [the method expects `error`](https://github.com/docker-mailserver/docker-mailserver/blob/bb2038e8c65eaeaef20c61694fa68041de0a8d25/target/scripts/helpers/log.sh#L63), the `return 1` is fine though 👍 

```suggestion
function _add_to_or_update_postfix_main() {
  local KEY=${1:?Key name is required}
  local NEW_VALUE=${2:?New value is required}
  local ACTION=${3:-append}

  # If entry does not exist, add it - otherwise update with ACTION:
  if ! grep -q -E "^${KEY}" /etc/postfix/main.cf; then
    postconf "${KEY} = ${NEW_VALUE}"
  else
    KEY=$(_escape_for_sed "${KEY}")
    NEW_VALUE=$(_escape_for_sed "${NEW_VALUE}")
    local SED_STRING

    case "${ACTION}" in
      ('append')
        SED_STRING="/${NEW_VALUE}/! s|^(${KEY} *=.*)|\1 ${NEW_VALUE}|g"
        ;;
      ('prepend')
        SED_STRING="/${NEW_VALUE}/! s|^(${KEY}) *= *(.*)|\1 = ${NEW_VALUE} \2|g"
        ;;
      (*)
        _log 'error' "Action '${3}' in _add_to_or_update_postfix_main is unknown"
        return 1
        ;;
    esac

    sed -i -E "${SED_STRING}" /etc/postfix/main.cf
  fi
}
```

polarathene (changes requested):
I personally think the suggestion would be more maintainer friendly (_wishful thinking of the project getting additional maintainers some day.._).

If you disagree, @casperklein could weigh in with their opinion, if not I won't fight it and will approve (_requested changes instead of approved was only to provide a bit of time to consider the suggestion_).

polarathene (approved):
Thank you, much appreciated! ❤️

casperklein:
Merged too fast, couldn't review. Therefore, see my suggestions in the follow up PR #3508.

PR #3485, opened 2023-08-16
Add support for extracting certs from Caddy v2 data

am97:
# Description

This PR adds support for  extracting certificates from Caddy v2 data, as this is already supported for Traefik's `acme.json`. 
## Type of change

## Checklist:

### DIFF
--- a/docs/content/config/security/ssl.md
+++ b/docs/content/config/security/ssl.md
@@ -501,7 +501,7 @@ services:
       caddy.local_certs: # Remove this label when going to production
 
   dms:
-    image: registry.nuiton.org/codelutin/admsys/swarm-stack/docker-mailserver:with_caddy_support
+    image: ghcr.io/docker-mailserver/docker-mailserver:latest
     hostname: mail.example.com
     ports:
       - "25:25"
@@ -509,6 +509,7 @@ services:
       - caddy_data:/caddy_data
     environment:
       SSL_TYPE: letsencrypt
+      SSL_DOMAIN: mail.example.com # Ensure this is exactly the same as the domain you pass to Caddy
       LOG_LEVEL: trace
     labels:
       caddy_0: mail.example.com
--- a/docs/content/config/security/ssl.md
+++ b/docs/content/config/security/ssl.md
@@ -478,94 +478,44 @@ DSM-generated letsencrypt certificates get auto-renewed every three months.
 
 ### Caddy
 
-For Caddy v2 you can specify the `key_type` in your server's global settings, which would end up looking something like this if you're using a `Caddyfile`:
+[Caddy](https://caddyserver.com/) is an open-source web server with built-in SSL certificate generation. You can use the [official Docker image](https://hub.docker.com/_/caddy) and write your own Caddyfile, or use [lucaslorentz/caddy-docker-proxy](https://github.com/lucaslorentz/caddy-docker-proxy), which lets you add labels to your services in order to generate a Caddyfile.
 
-```caddyfile
-{
-  debug
-  admin localhost:2019
-  http_port 80
-  https_port 443
-  default_sni example.com
-  key_type rsa4096
-}
-```
-
-If you are instead using a json config for Caddy v2, you can set it in your site's TLS automation policies:
-
-??? example "Caddy v2 JSON example snippet"
-
-    ```json
-    {
-      "apps": {
-        "http": {
-          "servers": {
-            "srv0": {
-              "listen": [
-                ":443"
-              ],
-              "routes": [
-                {
-                  "match": [
-                    {
-                      "host": [
-                        "mail.example.com",
-                      ]
-                    }
-                  ],
-                  "handle": [
-                    {
-                      "handler": "subroute",
-                      "routes": [
-                        {
-                          "handle": [
-                            {
-                              "body": "",
-                              "handler": "static_response"
-                            }
-                          ]
-                        }
-                      ]
-                    }
-                  ],
-                  "terminal": true
-                },
-              ]
-            }
-          }
-        },
-        "tls": {
-          "automation": {
-            "policies": [
-              {
-                "subjects": [
-                  "mail.example.com",
-                ],
-                "key_type": "rsa2048",
-                "issuer": {
-                  "email": "admin@example.com",
-                  "module": "acme"
-                }
-              },
-              {
-                "issuer": {
-                  "email": "admin@example.com",
-                  "module": "acme"
-                }
-              }
-            ]
-          }
-        }
-      }
-    }
-    ```
-
-The generated certificates can then be mounted:
+DMS supports retrieving certificates from Caddy data when the local file system storage is used (the default). You just need to mount your Caddy data and set `SSL_TYPE: letsencrypt` in the DMS container. Here is a minimal example with `lucaslorentz/caddy-docker-proxy`:
 
 ```yaml
+version: "3.7"
+
 volumes:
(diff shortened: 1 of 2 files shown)


am97:
I would like to add a test for the new `_caddy_sanitize_domain` function (call it with multiple values and check that the output is what's expected). Where can I add it ? The function has no dependencies other than bash and sed, perhaps it's not necessary to spawn a container to test this function.

docs/content/config/security/ssl.md
```diff
@@ -478,94 +478,45 @@ DSM-generated letsencrypt certificates get auto-renewed every three months.
...
-            ]
-          }
-        }
-      }
-    }
-    ```
+[Caddy](https://caddyserver.com/) is an open-source web server with built-in SSL certificate generation. You can use the [official Docker image](https://hub.docker.com/_/caddy) and write your own Caddyfile, or use [lucaslorentz/caddy-docker-proxy](https://github.com/lucaslorentz/caddy-docker-proxy), which lets you add labels to your services in order to generate a Caddyfile.
 
-The generated certificates can then be mounted:
+DMS supports retrieving certificates from Caddy data when the local file system storage is used (the default). You just need to mount your Caddy data and set `SSL_TYPE: letsencrypt` in the DMS container. Here is a minimal example with `lucaslorentz/caddy-docker-proxy`:
```
polarathene, line 483: You removed the admonition (_example type, the purple collapsible section in our docs_): `??? example`

Adding that back requires 4 space indentation of the codeblock that follows, like the JSON snippet prior was.

docs/content/config/security/ssl.md
```diff
@@ -478,94 +478,45 @@ DSM-generated letsencrypt certificates get auto-renewed every three months.
...
-                  "module": "acme"
-                }
-              }
-            ]
-          }
-        }
-      }
-    }
-    ```
+[Caddy](https://caddyserver.com/) is an open-source web server with built-in SSL certificate generation. You can use the [official Docker image](https://hub.docker.com/_/caddy) and write your own Caddyfile, or use [lucaslorentz/caddy-docker-proxy](https://github.com/lucaslorentz/caddy-docker-proxy), which lets you add labels to your services in order to generate a Caddyfile.
```
polarathene, line 481: While the Caddy section was outdated a fair bit, you have dropped information about configuring the key type (Caddyfile and JSON examples). The self-generated and nginx-proxy sections both touch on this too, but it's most likely not too relevant / specific to DMS, so dropping these old examples is acceptable.

docs/content/config/security/ssl.md
```diff
@@ -478,94 +478,45 @@ DSM-generated letsencrypt certificates get auto-renewed every three months.
...
 
-The generated certificates can then be mounted:
+DMS supports retrieving certificates from Caddy data when the local file system storage is used (the default). You just need to mount your Caddy data and set `SSL_TYPE: letsencrypt` in the DMS container. Here is a minimal example with `lucaslorentz/caddy-docker-proxy`:
 
 ```yaml
+version: "3.7"
+
 volumes:
-  - ${CADDY_DATA_DIR}/certificates/acme-v02.api.letsencrypt.org-directory/mail.example.com/mail.example.com.crt:/etc/letsencrypt/live/mail.example.com/fullchain.pem
-  - ${CADDY_DATA_DIR}/certificates/acme-v02.api.letsencrypt.org-directory/mail.example.com/mail.example.com.key:/etc/letsencrypt/live/mail.example.com/privkey.pem
```
polarathene, line 568: And here in the old docs we have an example filepath that your PR presumably operates on and is still roughly the same?

Like with certbot section, it advises mounting the relevant cert folder to `/etc/letsencrypt/live` location that DMS will internally check for with `SSL_TYPE=letsencrypt`:

https://github.com/docker-mailserver/docker-mailserver/blob/8f971713360c4620046c7f223fe66df9ea800253/target/scripts/helpers/ssl.sh#L394-L412

Traefik has special support because of the `acme.json` format. Whereas I feel this PR is adding a very minor convenience to the user?

[Our docs do cover `SSL_TYPE=manual`](https://docker-mailserver.github.io/docker-mailserver/v12.1/config/security/ssl/#bring-your-own-certificates) as an alternative to `SSL_TYPE=letsencrypt`.  Both of which have `check-for-changes.sh` support as their locations are monitored:

https://github.com/docker-mailserver/docker-mailserver/blob/8f971713360c4620046c7f223fe66df9ea800253/target/scripts/helpers/change-detection.sh#L46-L63

However in your current PR state, your approach for Caddy would miss out on this and would not benefit. The Traefik `acme.json` support is hard-coded into our `/etc/letsencrypt` internal location which has been considered acceptable. While for this Caddy support we need to also rely on `/caddy_data` to monitor for changes (_ideally only for those relevant to DMS, which is better supported if mounting the full path to isolate if the reverse-proxy was using the same data folder for other services, as shown with rspamd_).

The difference with normal `SSL_TYPE=letsencrypt` is that unlike `SSL_TYPE=manual` there is no internal copy made. I'd like to simplify the SSL / TLS support, but haven't found time to. Ideally it'd be more generic in finding a cert from some location and then copying it to a fixed internal location. I understand the desire for convenience though.

polarathene (changes requested):
Hey thanks for the contribution! ❤️ 

There's a fair amount of changes to address, usually it's a better idea to open an issue and discuss them prior for approval.

If you can address the change requests I'm ok with continuing the review to approval, but keep in mind that a refactor in future may affect this feature.

target/scripts/helpers/ssl.sh
```diff
@@ -440,6 +469,54 @@ function _extract_certs_from_acme() {
...
+    s/+/_plus_/g;
+    s/*/wildcard_/g;
+    s/:/-/g;
+    s/\.\.//g;
+  ' <<< "${CERT_DOMAIN}")
+
+  # Remove all non-word characters and print result to stdout
+  # Note: \w in the original code is equivalent to [:alnum:]_
+  echo "${CERT_DOMAIN//[^[:alnum:]_@.-]/}"
+}
```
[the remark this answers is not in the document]
georglauterbach: Yes, just use `find`. It was written for this use case.

am97:
Hello and thanks for the extended review !

As you say I should have opened an issue to discuss this first. The changes required to support Caddy are bigger than I originally thought. I was not aware of the internals of DMS, such as `check-for-changes.sh`.

I think that support for multiple «certificate managers» may be implemented in a modular way with «extractors», I'm opening an issue to discuss this from scratch.

If you're okay with the approach, I can transform this PR to «Add a Caddy extractor» (or open a new one).

polarathene:
> If you're okay with the approach, I can transform this PR to «Add a Caddy extractor» (or open a new one).

A new PR would be better thanks. Feel free to close this one if you'd like to pursue the alternative discussed in your new issue.

There's a possibility I may be able to tackle it but if I don't find time by September I doubt it'll be any time soon 😅 

I outlined in the issue what a rework would look like, I have some familiarity with what commands can be used to accomplish the different tasks. In the meantime, I believe Caddy can be configured to restrict the providers to a single one if you choose to, which would avoid your concern with potential vendor changing allowing for `SSL_TYPE=manual` to work reliably.

am97:
Ok ! I'm Closing this PR then. I'll wait for your refactor of `check-for-changes.sh` and `ssl.sh` and add some notes on the PR.

Yes, Caddy can be configured to use a single CA, with the [`acme_ca` global option](https://caddyserver.com/docs/caddyfile/options#acme-ca) (I'm keeping the default of Lets Encrypt + ZeroSSL on my side, and I use an intermediate container to copy the certs, it does not account for all corner cases but should be sufficient for the moment)

PR #3426, opened 2023-07-14
feat: start implementing replication - Add DOVECOT_ITERATE_ATTRS/FILTER ENVs

williamdes:
# Description

I added two ENVs needed to list users: https://github.com/docker-mailserver/docker-mailserver/issues/2048#issuecomment-865180458

Ref #2048

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Improvement (non-breaking change that does improve existing functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [x] If necessary I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #1921, opened 2021-04-19
docs(ldap): Improve LDAP documentation

moqmar:
# Description
As discussed in https://github.com/docker-mailserver/docker-mailserver/pull/1902#issuecomment-821121746, this completely overhauls the LDAP documentation and actually explains what most of the variables do. Looking forward to feedback.

## Type of change

- [x] Documentation update
- [x] This change requires a documentation update

## Checklist:

- [x] My code follows the style guidelines of this project (probably?)
- [x] I have performed a self-review of my own code
- [x] ~~I have commented my code, particularly in hard-to-understand areas~~
- [x] ~~I have made corresponding changes to the documentation (README.md or ENVIRONMENT.md or the documentation)~~
- [x] ~~If necessary I have added tests that prove my fix is effective or that my feature works~~
- [x] ~~New and existing unit tests pass locally with my changes~~

PR #3494, opened 2023-08-21
tests: Change OpenLDAP image to `bitnami/openldap`

polarathene:
**TL;DR:**
- New image is actively maintained vs existing one that is over 5 years old. 
- Slight improvement to LDAP tree config via `.ldif` files.
- No more `Dockerfile` required to build, we can just rely on `docker run`.

---

# Full Details

Startup time for this new image is around 5 seconds? (_The LDAP test uses a standard 20 second timeout check to wait until the server is ready before continuing with starting the DMS image_).

The image [`osixia/openldap` has not seen any activity since Feb 2021](https://github.com/osixia/docker-openldap), while our `Dockerfile` was fixed to [v1.1.6` (Feb 2018)](https://github.com/osixia/docker-openldap/releases/tag/v1.1.6). It was [originally contributed to DMS in in Oct 2016](https://github.com/docker-mailserver/docker-mailserver/pull/352) with no real changes since then.

This PR migrates to `bitnami/openldap` which required modifying the `01_mail-tree.ldif` to also include adding the root object to start successfully.

The user account `.ldif` files have minimal changes:
- Lines moved around for better organization
- Additional comments for context
- Removal of inherited `objectClass` attributes (`person`, `top`) from the `orgnizationalPerson` class. Attribute `sn` changed to long form `surname` and values corrected with `givenName`. `changetype: add` was also not necessary.

Additionally, the image does not support the `.schema` format, they must be converted to `.ldif`. We need this for supporting the [`postfix-book.schema`](https://github.com/variablenix/ldap-mail-schema/blob/master/postfix-book.schema) (_supposedly the source of the file we carry, [docs have mentioned it since May 2021](https://github.com/docker-mailserver/docker-mailserver/pull/1921)_) which allows the LDAP user account objects (_defined in `.ldif` files_) to use extra attributes provided by the schema.
- I've converted the `postfix-book.schema` (_as [explained here](https://github.com/docker-mailserver/docker-mailserver/pull/3426#issuecomment-1681921407)_) into `postfix-book.ldif`.
- This schema isn't always an option for LDAP providers (_specifically ActiveDirectory apparently?_), and AFAIK isn't actually required to support LDAP with DMS, it just adds some convenience which our defaults assume it but can be configured differently as our LDAP docs describe.

---

**Benefit:** We no longer need to build an image with the LDIF files to use it for tests, just an initial pull. The `bitnami/openldap` image is actively maintained and one of the most popular OpenLDAP images on DockerHub.

**Concerns:**
- The issue tracker isn't too great as it's mixed in with all other bitnami images, and VMware has since acquired bitnami which branched off a commercial offering.
- Docs amount to the README which is roughly adequate but wasn't entirely straight-forward. I ran into problems initially trying to migrate to this image, and needed to learn about the ENV `BITNAMI_DEBUG=true` plus looking through the image source scripts to get a bet understanding.
- Actual issues raised for the openldap image also don't tend to pan out too well / helpful which isn't too reassuring..
- That said I'm not sure if there is any better maintained/documented/reputable options out there, at least for LDAP with OpenLDAP. It does get easier when you actually understand LDAP/OpenLDAP better in jargon, setup, and commands which took me a couple days.

**Alternative:** https://github.com/clayrisser/docker-openldap
- I did initially have [better success with migrating to an image variant based off of `bitnami/openldap`](https://github.com/docker-mailserver/docker-mailserver/pull/3426#issuecomment-1681921407) (_notes on trying to migrate, details this alternative and `bitnami/openldap`_).
- **Pro:** It adds some extra improvements including bundling of `postfix-book.schema` so that we don't have to.
- **Con:** It seems to differ with its Access Control List (ACL) policy. [Requiring a separate `.ldif` migration file](https://github.com/clayrisser/docker-openldap/issues/1#issuecomment-1681695098) to allow LDAP queries to authenticate for the `userPassword` attribute.
- **Con:** This variant is still fairly young, and I can't comment on what maintainer support is like. Other than submitting a detailed issue 5 days ago with no response. Presently that's the first issue opened on that repo, so it's difficult to gauge how long that would be actively maintained / supported vs upstream `bitnami/openldap`. For DMS LDAP tests at least, since I finally got `bitnami/openldap` to work, the value of this alternative image is low.
- I have noticed that `clayrisser` is or was a user of DMS with LDAP 😎 (_I came across a comment from them in 2020 on our issue tracker while looking through LDAP issues_).

## Type of change

## Checklist:


georglauterbach (approved):
Awesome work! 🚀

PR #3006, opened 2023-01-16
config: disable SMTP authentication on port 25

mazzz1y:
# Description

Fixes #2895

1) Remove `smtpd_sasl_auth_enable` parameter from postfix configuration file. It disabled by default: [docs](http://www.postfix.org/postconf.5.html#smtpd_sasl_auth_enable)
2) Tests: use openssl for smtp-auth and make sure that auth on 25 port is disabled

## Type of change

Improvement (non-breaking change that does improve existing functionality)

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [x] If necessary I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #3496, opened 2023-08-21
Docs: Drop mention of port 25 support for authenticated submission

HeySora:
# Description

In the same spirit as #3464, this PR fixes a documentation issue (in the FAQ page, and in "Understanding the Ports"), about using port 25 for mail clients, which is unsupported since #3006.

## Type of change

### DIFF
--- a/docs/content/config/security/understanding-the-ports.md
+++ b/docs/content/config/security/understanding-the-ports.md
@@ -16,7 +16,7 @@ Prefer ports with Implicit [TLS][wikipedia-tls] ports, they're more secure than 
 | IMAP4                    | 143                      | 993             | Retrieval            | Yes                |
 
 1. A connection _may_ be secured over TLS when both ends support `STARTTLS`. On ports 110, 143 and 587, DMS will reject a connection that cannot be secured. Port 25 is [required][ref-port25-mandatory] to support insecure connections.
-2. Receives email, DMS additionally filters for spam and viruses. For submitting email to the server to be sent to third-parties, you should prefer the _submission_ ports (465, 587) - which require authentication. Unless a relay host is configured (eg: SendGrid), outgoing email will leave the server via port 25 (_thus outbound traffic must not be blocked by your provider or firewall_).
+2. Receives email, DMS additionally filters for spam and viruses. For submitting email to the server to be sent to third-parties, you need to use the _submission_ ports (465, 587) - which require authentication. Unless a relay host is configured (eg: SendGrid), outgoing email will leave the server via port 25 (_thus outbound traffic must not be blocked by your provider or firewall_).
 3. A _submission_ port since 2018 ([RFC 8314][rfc-8314]).
 
 ??? warning "Beware of outdated advice on port 465"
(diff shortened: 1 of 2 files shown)


polarathene:
Not exactly the case.

You can still use port 25 to submit mail without requiring authentication.
- This is done internally for system mail (eg: cron), and can be supported when your client is trusted (_eg: Via `PERMIT_DOCKER` ENV_).
- While authentication would also enable trust (_working for the default `PERMIT_DOCKER=none`_), it has been intentionally disabled on Port 25 (_as it should_).

Perhaps we could better clarify that?

docs/content/faq.md
```diff
@@ -143,7 +143,7 @@ imap port:          143 or 993 with STARTTLS/SSL (recommended)
 imap path prefix:   INBOX
 
 # SMTP
-smtp port:          25 or 587/465 with STARTTLS/SSL (recommended)
+smtp port:          587 or 465 with STARTTLS/SSL (recommended)
```
polarathene, line 146: This is valid due to the suggestion of credentials to authenticate 👍

docs/content/config/security/understanding-the-ports.md
```diff
@@ -16,7 +16,7 @@ Prefer ports with Implicit [TLS][wikipedia-tls] ports, they're more secure than
 | IMAP4                    | 143                      | 993             | Retrieval            | Yes                |
 
 1. A connection _may_ be secured over TLS when both ends support `STARTTLS`. On ports 110, 143 and 587, DMS will reject a connection that cannot be secured. Port 25 is [required][ref-port25-mandatory] to support insecure connections.
-2. Receives email, DMS additionally filters for spam and viruses. For submitting email to the server to be sent to third-parties, you should prefer the _submission_ ports (465, 587) - which require authentication. Unless a relay host is configured (eg: SendGrid), outgoing email will leave the server via port 25 (_thus outbound traffic must not be blocked by your provider or firewall_).
+2. Receives email, DMS additionally filters for spam and viruses. For submitting email to the server to be sent to third-parties, you need to use the _submission_ ports (465, 587) - which require authentication. Unless a relay host is configured (eg: SendGrid), outgoing email will leave the server via port 25 (_thus outbound traffic must not be blocked by your provider or firewall_).
```
polarathene, line 19: While this one is not specific to authenticated submission.

You may have another container / server send mail to DMS on port 25 to relay externally elsewhere. Provided it's trusted via something like `PERMIT_DOCKER`, it will be able to send that mail.

Presently `SPOOF_PROTECTION=1` doesn't affect port 25, so such a scenario could send as anyone on port 25 when trusted even with this ENV enabled. Our method used is only related to authenticated SASL users being prevented from sending as anyone but themselves though, whereas a different Postfix option would be better for port 25 that prevents sending as a known SASL user (_without authenticating, which would be required over 465 or 587 instead_).

I'd still advise using authentication to send mail out, vs trusting an IP address where any process could from that source could use DMS as an open relay (_from that trusted source at least_).

polarathene (commented):
Here's some context if you want to further elaborate in the docs.

Otherwise, dropping the "Understanding the Ports" change, as it's not accurate. FAQ change is ok 👍

docs/content/config/security/understanding-the-ports.md
HeySora: Got it! Thank you for the clarification 😊  
This change has been reverted.

polarathene (approved):
Thanks for catching this and updating our docs ❤️

PR #3503, opened 2023-08-26
 scripts: add wrapper to update Postfix configuration safely (follow up)

casperklein:
# Description

While working on improving the function `_add_to_or_update_postfix_main()`, I noticed a problem we had before in other occasions.

`SED_STRING="/${NEW_VALUE}/!`

This is not really fail safe, when `$NEW_VALUE` is a substring of an existing value.

Now I simplified the whole function, by replacing `sed` with `postconf`.

Fixes #3484

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/helpers/postfix.sh
+++ b/target/scripts/helpers/postfix.sh
@@ -92,7 +92,7 @@ function _vhost_ldap_support() {
 # /etc/aliases is handled by `alias.sh` and uses `postalias` to update the Postfix alias database. No need for `postmap`.
 # http://www.postfix.org/postalias.1.html
 
-# Add an key with an value to Postfix's main configuration file
+# Add a key with a value to Postfix's main configuration file
 # or update an existing key. An already existing key can be updated
 # by either appending to the existing value (default) or by prepending.
 #
--- a/target/scripts/helpers/postfix.sh
+++ b/target/scripts/helpers/postfix.sh
@@ -114,10 +114,10 @@ function _add_to_or_update_postfix_main() {
 
     case "${ACTION}" in
       ('append')
-        SED_STRING="/${NEW_VALUE}/! s|^(${KEY}\s*=.*)|\1 ${NEW_VALUE}|g"
+        SED_STRING="/^${KEY}\s*=.*${NEW_VALUE}/! s|^(${KEY}\s*=.*)|\1 ${NEW_VALUE}|g"
         ;;
       ('prepend')
-        SED_STRING="/${NEW_VALUE}/! s|^(${KEY})\s*=\s*(.*)|\1 = ${NEW_VALUE} \2|g"
+        SED_STRING="/^${KEY}\s*=.*${NEW_VALUE}/! s|^(${KEY})\s*=\s*(.*)|\1 = ${NEW_VALUE} \2|g"
         ;;
       (*)
         _log 'error' "Action '${3}' in _add_to_or_update_postfix_main is unknown"
--- a/target/scripts/helpers/postfix.sh
+++ b/target/scripts/helpers/postfix.sh
@@ -98,7 +98,7 @@ function _vhost_ldap_support() {
 #
 # @param ${1} = key name in Postfix's main configuration file
 # @param ${2} = new value (appended or prepended)
-# @param ${3} = "append" (default) or "prepend" [OPTIONAL]
+# @param ${3} = action "append" (default) or "prepend" [OPTIONAL]
 function _add_to_or_update_postfix_main() {
   local KEY=${1:?Key name is required}
   local NEW_VALUE=${2:?New value is required}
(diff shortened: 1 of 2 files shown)


target/scripts/helpers/postfix.sh
```diff
@@ -92,39 +92,42 @@ function _vhost_ldap_support() {
...
+
+  # If key does not exist or value is empty, add it - otherwise update with ACTION:
+  if [[ -z ${CURRENT_VALUE} ]]; then
     postconf "${KEY} = ${NEW_VALUE}"
   else
-    KEY=$(_escape_for_sed "${KEY}")
-    NEW_VALUE=$(_escape_for_sed "${NEW_VALUE}")
-    local SED_STRING
+    # If $NEW_VALUE is already present --> nothing to do, skip.
+    if grep -qF " ${NEW_VALUE} " <<<" ${CURRENT_VALUE} "; then
```
casperklein, line 116: Pure BASH should also be fine?
```suggestion
    if [[ " ${CURRENT_VALUE} " == *" ${NEW_VALUE} "* ]]; then
```

target/scripts/helpers/postfix.sh
```diff
@@ -92,39 +92,42 @@ function _vhost_ldap_support() {
...
 function _add_to_or_update_postfix_main() {
   local KEY=${1:?Key name is required}
   local NEW_VALUE=${2:?New value is required}
   local ACTION=${3:-append}
+  local CURRENT_VALUE
 
-  # If entry does not exist, add it - otherwise update with ACTION:
-  if ! grep -q -E "^${KEY}" /etc/postfix/main.cf; then
+  # Get current value from /etc/postfix/main.cf
+  CURRENT_VALUE=$(postconf -h "${KEY}" 2>/dev/null)
```
polarathene, line 109: You'll probably want to call this helper first, as reading from `main.cf` frequently can compound the stalling effect otherwise, slowing down startup time:

https://github.com/docker-mailserver/docker-mailserver/blob/c8a0bfd36195281b629916d64edf0a2c1fbd89d1/target/scripts/helpers/utils.sh#L57-L66

I brought this up early in [the original PR review for this helper](https://github.com/docker-mailserver/docker-mailserver/pull/3484#pullrequestreview-1581477237) also thinking `postconf` would be a better way to go about it.

target/scripts/helpers/postfix.sh
polarathene: I am not a bash wizard like you two 😅 

Despite the glory of DMS bash scripts, we've not attracted anymore wizards as maintainers, so I'd probably go with the more familiar `grep` pattern from a maintenance perspective?

polarathene: I'm not sure how reliable this condition logic is though in practice.

- ~~I know you're using white-space to avoid accidents where value could belong to a prefix/suffix portion of something else.~~

...Nevermind! Being the smart wizard that you are, you're not using `grep` this way on `main.cf` contents which would have been riddled with a few gotchas 😅 (_I should probably do reviews once I've had a morning coffee_ 😝 )

polarathene (dismissed):
Nice improvement! 😁 

This is much simpler to grok now without all the `sed` syntax dance and escaping added into the mix.

target/scripts/helpers/postfix.sh
casperklein: Done [81b7785](https://github.com/docker-mailserver/docker-mailserver/pull/3503/commits/81b7785ff5182c379f48a57b63d712e3b2f22fdb)

target/scripts/helpers/postfix.sh
```diff
@@ -92,39 +92,44 @@ function _vhost_ldap_support() {
...
   local NEW_VALUE=${2:?New value is required}
   local ACTION=${3:-append}
+  local CURRENT_VALUE
 
-  # If entry does not exist, add it - otherwise update with ACTION:
-  if ! grep -q -E "^${KEY}" /etc/postfix/main.cf; then
+  _adjust_mtime_for_postfix_maincf
+
+  # Get current value from /etc/postfix/main.cf
+  CURRENT_VALUE=$(postconf -h "${KEY}" 2>/dev/null)
```
polarathene, line 111:
```suggestion
  # Get current value from /etc/postfix/main.cf
    _adjust_mtime_for_postfix_maincf
  CURRENT_VALUE=$(postconf -h "${KEY}" 2>/dev/null)
```

target/scripts/helpers/postfix.sh
georglauterbach: I also prefer `[[...` over `grep ...`.

georglauterbach (dismissed):
Not blocking on the check, but it'd be nice to see `[[` rather than `grep`.

polarathene (approved):

target/scripts/helpers/postfix.sh
casperklein: Switched to pure bash. Using build-ins should also a bit faster 😆 than calling external programs like `grep`.

polarathene (approved):

PR #3505, opened 2023-08-27
tests: add tests for helper function `_add_to_or_update_postfix_main()`

casperklein:
# Description

Some tests for the `_add_to_or_update_postfix_main()` helper function.

Fixes #

## Type of change

## Checklist:


test/tests/parallel/set3/scripts/postconf-helper.bats
```diff
@@ -0,0 +1,61 @@
+load "${REPOSITORY_ROOT}/test/helper/common"
+load "${REPOSITORY_ROOT}/test/helper/setup"
+
+BATS_TEST_NAME_PREFIX='[Scripts] (helper functions) (postfix.sh) (_add_to_or_update_postfix_main) '
+CONTAINER_NAME='dms-test-main.cf-changes'
```
polarathene, line 5: Convention is `dms-test` + container name + optional variant, where each value is kebab-case with `_` separator. Probably better to avoid `.` in the name too 🤷‍♂️ 

Let's just align with the test filename in this case?

```suggestion
CONTAINER_NAME='dms-test_postconf-helper'
```

test/tests/parallel/set3/scripts/postconf-helper.bats
```diff
@@ -0,0 +1,61 @@
...
+
+function setup_file() {
+  _init_with_defaults
+  _common_container_setup
+
+  # remove 'relayhost' from main.cf
+  _run_in_container postconf -X relayhost
+  assert_success
+
+  export CONFIG=/etc/postfix/main.cf
```
polarathene, line 17: Not necessary if you use extract out the common pattern in test cases to a method that can just reference this path directly.

```suggestion
```

polarathene (changes requested):
Nice test, thanks for putting this together! 😁 

I think we can make it a bit more DRY and it should be relocated to `tests/serial/` since test cases are dependent on running sequentially.

polarathene:
Actually, my bad. I forgot that our `parallel` tests only run entire test files concurrently, not their test cases, that will be sequential.

So it's only an issue when other tests have containers running that would affect each other, which barely applies to most of our tests.

test/tests/parallel/set3/scripts/postconf-helper.bats
```diff
@@ -0,0 +1,61 @@
...
+}
+
+function teardown_file() { _default_teardown ; }
+
+
+@test "check if initial value is empty" {
+  _run_in_container postconf -h relayhost
+  assert_output ""
+}
```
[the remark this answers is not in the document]
polarathene: Disregard this, BATS isn't configured to run the test-cases concurrently.

casperklein:
All suggestions applied 👍

test/tests/parallel/set3/scripts/postconf-helper.bats
```diff
@@ -0,0 +1,59 @@
...
+
+@test "append value (implicit)" {
+  _modify_postfix_main.cf "append-test-implicit"
+  assert_output "relayhost = prepend-test single-value-test append-test-explicit append-test-implicit"
+}
+
+@test "try to append already existing value" {
+  _modify_postfix_main.cf "append-test-implicit"
+  assert_output "relayhost = prepend-test single-value-test append-test-explicit append-test-implicit"
+}
```
polarathene, line 59: TIL that you can use `.` in function names, but just because we can we probably shouldn't 😅 

Not able to edit directly, so sending a full edit via suggestion:

```suggestion
BATS_TEST_NAME_PREFIX='[Scripts] (helper functions) (postfix - _add_to_or_update_postfix_main) '
CONTAINER_NAME='dms-test_postconf-helper'

# Various tests for the helper function `_add_to_or_update_postfix_main()`

function setup_file() {
  _init_with_defaults
  _common_container_setup

  # Begin tests without 'relayhost' defined in 'main.cf'
  _run_in_container postconf -X relayhost
  assert_success
}

function teardown_file() { _default_teardown ; }

# Add or modify in Postfix config `main.cf` a parameter key with the provided value.
# When the key already exists, the new value is appended (default), or prepended (explicitly requested).
# NOTE: This test-case helper is hard-coded for testing with the 'relayhost' parameter.
#
# @param ${1} = new value (appended or prepended)
# @param ${2} = action "append" (default) or "prepend" [OPTIONAL]
function _modify_postfix_main_config() {
  _run_in_container_bash "source /usr/local/bin/helpers/{postfix,utils}.sh && _add_to_or_update_postfix_main relayhost '${1}' '${2}'"
  _run_in_container grep '^relayhost' '/etc/postfix/main.cf'
}

@test "check if initial value is empty" {
  _run_in_container postconf -h 'relayhost'
  assert_output ''
}

@test "add single value" {
  _modify_postfix_main_config 'single-value-test'
  assert_output 'relayhost = single-value-test'
}

@test "prepend value" {
  _modify_postfix_main_config 'prepend-test' 'prepend'
  assert_output 'relayhost = prepend-test single-value-test'
}

@test "append value (explicit)" {
  _modify_postfix_main_config 'append-test-explicit' 'append'
  assert_output 'relayhost = prepend-test single-value-test append-test-explicit'
}

@test "append value (implicit)" {
  _modify_postfix_main_config 'append-test-implicit'
  assert_output 'relayhost = prepend-test single-value-test append-test-explicit append-test-implicit'
}

@test "try to append already existing value" {
  _modify_postfix_main_config 'append-test-implicit'
  assert_output 'relayhost = prepend-test single-value-test append-test-explicit append-test-implicit'
}
```

Revised some comments, changed the helper name to not use a `.`, and swapped double-quotes for single-quotes where appropriate. A slight change to the test prefix (which I still feel is quite lengthy).

Oh that diff is actually decent for once and self-explanatory 🤯

polarathene (changes requested):

test/tests/parallel/set3/scripts/postconf-helper.bats
polarathene: Oh, seems I can't apply the suggestion directly. I was going to do that and approve, leaving you to revert anything or merge 🤷‍♂️ 

Apply what you're comfortable with and I'll approve 👍

casperklein: > TIL that you can use . in function names, but just because we can we probably shouldn't 😅

I would have been surprised, if neither you or Georg would be triggered by that 😆 

> but just because we can we probably shouldn't

What's the problem? In that special case, I found it pretty suitable and precise. By calling the function, you exactly know, which file is modified.

PS: [You can use a lot of characters for functions, even emoticons](https://blog.dnmfarrell.com/post/bash-function-names-can-be-almost-anything/) 😆 . The limitations from variable naming does not apply. For example, `@test` is a perfectly valid function name.

polarathene: > What's the problem?

My distrust / paranoia of it eventually contributing to some unexpected gotcha 😅 

Probably some bias with it seeming off since a `.` usually has meaning elsewhere, especially other languages.

I know filenames can be problematic compatibility wise for example, and Apple allowed emoji when setting your password in the past but not inputting emoji for actual login, locking you out.

Emoji is actually a pretty bad one, have experienced my own gotchas when writing a hash function since some emoji can represent more than 20 bytes (_rendered as one emoji but like a ligature is actually many separate emoji components, can trip up some text editors too_).

So it's mostly down to playing it safe and predictable 😎

polarathene (approved):

PR #3515, opened 2023-08-29
ci: Revise PR template (Change checklist into a comment)

polarathene:
# Description

The checklist isn't too important to render, it is primarily to benefit the contributor as a reminder.

Keep it for that purpose, but treated as comment instead?

## Type of change

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,7 +1,9 @@
 # Description
 
-<!-- Include a summary of the change.
-     Please also include relevant motivation and context. -->
+<!--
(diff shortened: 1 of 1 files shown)


georglauterbach (changes requested):
I disagree; to mark an item in the checklist, the OP has to actively think of whether they did what the item requires them to do. Ticking the items may be nervewrecking when doing a lot of PRs, but also necessary. For long-time maintainers, I'm fine with just deleting them because I know you guys know, but for others I'd like to see it.

Issue #3517, opened 2023-08-30
[TODO]: Revise LDAP support


polarathene:
### Subject

Something else that requires developers attention

### Description

I have been working on getting familiar with the LDAP support in DMS, going over the history and issues / feature requests after revising the LDAP tests we have.

There's quite a few changes to work on, some for reduced maintenance burden, others breaking changes or feature work. I've made extensive notes for my own reference to support that work but as I revise them, I'll transfer them here as a potentially helpful reference for anyone else working on the project, especially if I run out of time for DMS.

## TODO

WIP

## DMS LDAP feature History

- May 2016 - Original request for LDAP / external auth: https://github.com/docker-mailserver/docker-mailserver/issues/187
- Aug 2016 - Earliest LDAP PR attempt: https://github.com/docker-mailserver/docker-mailserver/pull/251
- Oct 2016 - [LDAP support implemented for both Postfix and Dovecot + initial LDAP tests contributed](https://github.com/docker-mailserver/docker-mailserver/pull/352):
  - This PR was [built off an earlier PR attempt in Sep 2016](https://github.com/docker-mailserver/docker-mailserver/pull/335), which lacked Dovecot integration. The revised PR not only added support for Dovecot, but also adjusted the original PR content (_LDAP config files / ENV_) to use the `uniqueIdentifier` attribute for lookup queries (_as it was introduced via the LDIF files_).
  - `Dockerfile` added new packages: `dovecot-ldap`, `postfix-ldap`, `sasl2-bin` (_Provides cyrus `saslauthd`, aka `ENABLE_SASLAUTHD` ENV_).
  - `postfix-ldap-*.cf` files introduced, but only as example configs, with [support to copy your own to an internal location](https://github.com/docker-mailserver/docker-mailserver/pull/352/files#diff-6d88c6c2abe074135b8efac1cf0dca97dd288a71720cdcd6ac48c2a6f85dcf08R83-R86) (_via the DMS config volume `/tmp/docker-mailserver`_).
    - Nov 2016 - The [`postfix-` filename prefix was dropped and the example configs moved into the image](https://github.com/docker-mailserver/docker-mailserver/pull/379)  (_at `/etc/postfix/` via the `Dockerfile`_).
[...]

PR #3522, opened 2023-09-01
chore: LDAP config improvements

polarathene:
# Description

- Removes fallback compatibility ENV support for `ldap://` when no URI scheme is provided. v13 will better communicate this via error message in the deprecation checks as a precaution. Technically not a breaking change.
- The `SASLAUTHD_*` ENV are only used to generate a config file. I have removed logic for this in `variables-stack.sh`, so that only `saslauthd.sh` is handling it (_and only when using LDAP support_). The current defaults / inheritance is preserved (_with the exception of `ldap://` URI scheme fallback_).
- The current support to use ENV to target specific settings for overriding in LDAP config files (_with the `_replace_by_env_in_file()` helper method_) is presently not compatible with `/etc/saslauthd.conf` (_which we generate during startup_), due to different key/value delimiter format. I intend to refactor the approach for these configs and the related ENV support after this PR.
- LDAP tests have been updated for the explicit `ldap://` URI scheme and to relocate some LDAP ENV config that is generic, not just specific to Postfix.

By using `sed` to remove the unset settings, we simplify handling for a portion of this config that relied on updating an ENV variable to include the key (_and when unset, just created multiple empty lines_).

---

I believe a good approach going forward is to provide all settings in the supported configs upfront, with the `sed` removal approach. 

I have the configs already defined for this and would like to follow-up with another PR that implements such a change (_as [detailed here](https://github.com/docker-mailserver/docker-mailserver/issues/3517#issuecomment-1702053878)_).

Defaults that are non-standard will be dropped in a future PR. With consideration for sane defaults from DMS.

## Type of change

## Checklist:


target/scripts/startup/setup.d/saslauthd.sh
```diff
@@ -1,29 +1,29 @@
...
-
 function _setup_saslauthd() {
   _log 'debug' 'Setting up SASLAUTHD'
 
-  if [[ ! -f /etc/saslauthd.conf ]]; then
+  # NOTE: It's unlikely this file would already exist,
+  # Unlike Dovecot/Postfix LDAP support, this file has no ENV replacement
+  # nor does it copy from the DMS config volume to this internal location.
+  if [[ ${ACCOUNT_PROVISIONER} == 'LDAP' ]] \
+  && [[ ! -f /etc/saslauthd.conf ]]; then
```
polarathene, line 10: Technically instead of `ACCOUNT_PROVISIONER=LDAP`, this should be for `SASLAUTHD_MECHANISMS=ldap`, as the two features serve a separate purpose? 🤷‍♂️ (_although I'm not sure when you'd only have Postfix auth against LDAP while using a different `ACCOUNT_PROVISIONER`_).

target/scripts/startup/setup.d/saslauthd.sh
```diff
@@ -1,29 +1,29 @@
...
-ldap_start_tls: ${SASLAUTHD_LDAP_START_TLS}
-ldap_tls_check_peer: ${SASLAUTHD_LDAP_TLS_CHECK_PEER}
-
-${SASLAUTHD_LDAP_TLS_CACERT_FILE}
-${SASLAUTHD_LDAP_TLS_CACERT_DIR}
-${SASLAUTHD_LDAP_PASSWORD_ATTR}
-${SASLAUTHD_LDAP_MECH}
 
+    # Create a config based on ENV
+    sed '/^.*: $/d'> /etc/saslauthd.conf << EOF
```
georglauterbach, line 14: What does this `sed` actually do?

target/scripts/startup/setup.d/saslauthd.sh
georglauterbach: Should we take action here immediately?

georglauterbach (approved):
👍🏼

target/scripts/startup/setup.d/saslauthd.sh
polarathene: `key: ` is matched (_space could probably be `\s*`_), so that if no value was assigned via the HereDoc input, the line is removed from the output.

Before we had an awkward workaround to avoid keys with empty values, which resulted in multiple blank lines if unset.

target/scripts/startup/setup.d/saslauthd.sh
polarathene: Nah, this is getting reworked by a follow-up and can be addressed by that 👍

target/scripts/startup/setup.d/saslauthd.sh
casperklein: Only adding lines without empty vars.

casperklein: I just noticed, for empty vars, ${foo:-} should have been used. ATM that's not a problem. But a good practice, if we want to introduce "set -u" in the future.

PR #3526, opened 2023-09-04
refactor: Helper function `_replace_by_env_in_file()`

polarathene:
# Description

Some refactoring improvements and fixes to the escaping support:

-  `_replace_by_env_in_file()`:
   - Minor change to add support for usage with `saslauthd.conf` which differs with `:` for it's key/value delimiter.
   - Also adopts using the`_escape_for_sed()` helper function, instead of the inline sed pattern (_which for some reason escaped a `=`? but not actual regex tokens `(` + `)`, thus buggy if ever matching on input with those tokens_).
   - Since the value is used regardless in the final `sed` expression, no need to prepend white-space, and we can additionally preserve the white-space between key and value of the source input while being more DRY.
- `_escape_for_sed()`:
  - This was not escaping for `(` + `)` (_or even `|`_), which is required for proper escaping support when using `sed -E` / `sed -r` with this helper function. It does now.
  - Additionally added escaping support for the `&` _replacement segment_ token, which seems ok to escape in the _regexp segment_ too (_even though it has no meaning there_). Increased verbosity to better grok pattern matching expression (_easier when broken down, vs all in a single line_), clarified an escaping concern with the `sed` segment delimiter (_since the project is not consistent there_).

**NOTE:** Tests will fail, this PR can be rebased upon https://github.com/docker-mailserver/docker-mailserver/pull/3524 which for the `mail_with_ldap.bats` test at least should pass then, other tests can be updated here after that.

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/helpers/utils.sh
+++ b/target/scripts/helpers/utils.sh
@@ -130,7 +130,7 @@ function _replace_by_env_in_file() {
     ESCAPED_KEY=$(_escape_for_sed "${KEY,,}")
     ESCAPED_VALUE=$(_escape_for_sed "${VALUE}")
     _log 'trace' "Setting value of '${KEY}' in '${CONFIG_FILE}' to '${VALUE}'"
-    sed -i -E "s#^(${ESCAPED_KEY}[[:space:]]*${KV_DELIMITER}[[:space:]]*).*#\1${ESCAPED_VALUE}#g" "${CONFIG_FILE}"
+    sedfile -i -E "s#^(${ESCAPED_KEY}[[:space:]]*${KV_DELIMITER}[[:space:]]*).*#\1${ESCAPED_VALUE}#gi" "${CONFIG_FILE}"
   done < <(env | grep "^${ENV_PREFIX}")
 }
 
--- a/target/scripts/helpers/utils.sh
+++ b/target/scripts/helpers/utils.sh
@@ -129,9 +129,8 @@ function _replace_by_env_in_file() {
     KEY=${KEY#"${ENV_PREFIX}"} # strip prefix
     ESCAPED_KEY=$(_escape_for_sed "${KEY,,}")
     ESCAPED_VALUE=$(_escape_for_sed "${VALUE}")
-    [[ -n ${ESCAPED_VALUE} ]] && ESCAPED_VALUE=" ${ESCAPED_VALUE}"
     _log 'trace' "Setting value of '${KEY}' in '${CONFIG_FILE}' to '${VALUE}'"
-    sed -i -E "s#^${ESCAPED_KEY}[[:space:]]*${KV_DELIMITER}.*#${ESCAPED_KEY} ${KV_DELIMITER}${ESCAPED_VALUE}#g" "${CONFIG_FILE}"
+    sed -i -E "s#^(${ESCAPED_KEY}[[:space:]]*${KV_DELIMITER}[[:space:]]*).*#\1${ESCAPED_VALUE}#g" "${CONFIG_FILE}"
   done < <(env | grep "^${ENV_PREFIX}")
 }
 
--- a/target/scripts/helpers/utils.sh
+++ b/target/scripts/helpers/utils.sh
@@ -4,14 +4,30 @@ function _escape() {
   echo "${1//./\\.}"
 }
 
-# TODO: Not in use currently. Maybe in the future: https://github.com/docker-mailserver/docker-mailserver/pull/3484/files#r1299410851
-# Replaces a string so that it can be used inside
-# `sed` safely.
+# Replaces a string so that it can be used inside the `sed` regex segment safely.
+# WARNING: Only for use with `sed -E` / `sed -r` due to escaping additional tokens,
+# which are valid tokens when escaped in basic regex mode.
 #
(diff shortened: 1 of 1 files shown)


georglauterbach (approved):

georglauterbach (dismissed):

target/scripts/helpers/utils.sh
```diff
@@ -105,15 +122,16 @@ function _replace_by_env_in_file() {
   fi
 
   local ENV_PREFIX=${1} CONFIG_FILE=${2}
+  local KV_DELIMITER=${3:-'='}
```
casperklein, line 125: We normally don't quote here. Any reason to quote the equal sign?

target/scripts/helpers/utils.sh
```diff
@@ -105,15 +122,16 @@ function _replace_by_env_in_file() {
...
   local ESCAPED_VALUE ESCAPED_KEY
 
-  while IFS='=' read -r KEY VALUE; do
+  while IFS="${KV_DELIMITER}" read -r KEY VALUE; do
     KEY=${KEY#"${ENV_PREFIX}"} # strip prefix
-    ESCAPED_KEY=$(sed -E 's#([\=\&\|\$\.\*\/\[\\^]|\])#\\\1#g' <<< "${KEY,,}")
-    ESCAPED_VALUE=$(sed -E 's#([\=\&\|\$\.\*\/\[\\^]|\])#\\\1#g' <<< "${VALUE}")
+    ESCAPED_KEY=$(_escape_for_sed "${KEY,,}")
```
casperklein, line 130: Why is `$KEY` lower-cased? Currently this function is only used for LDAP. To keep it universal (able to handle other files which may have mixed cases), `$KEY` shouldn't be altered.

target/scripts/helpers/utils.sh
```diff
@@ -105,15 +122,15 @@ function _replace_by_env_in_file() {
...
+  while IFS="${KV_DELIMITER}" read -r KEY VALUE; do
     KEY=${KEY#"${ENV_PREFIX}"} # strip prefix
-    ESCAPED_KEY=$(sed -E 's#([\=\&\|\$\.\*\/\[\\^]|\])#\\\1#g' <<< "${KEY,,}")
-    ESCAPED_VALUE=$(sed -E 's#([\=\&\|\$\.\*\/\[\\^]|\])#\\\1#g' <<< "${VALUE}")
-    [[ -n ${ESCAPED_VALUE} ]] && ESCAPED_VALUE=" ${ESCAPED_VALUE}"
+    ESCAPED_KEY=$(_escape_for_sed "${KEY,,}")
+    ESCAPED_VALUE=$(_escape_for_sed "${VALUE}")
     _log 'trace' "Setting value of '${KEY}' in '${CONFIG_FILE}' to '${VALUE}'"
-    sed -i -E "s#^${ESCAPED_KEY}[[:space:]]*=.*#${ESCAPED_KEY} =${ESCAPED_VALUE}#g" "${CONFIG_FILE}"
+    sed -i -E "s#^(${ESCAPED_KEY}[[:space:]]*${KV_DELIMITER}[[:space:]]*).*#\1${ESCAPED_VALUE}#g" "${CONFIG_FILE}"
```
casperklein, line 133:
```suggestion
    sedfile -i -E "s#^(${ESCAPED_KEY}[[:space:]]*${KV_DELIMITER}[[:space:]]*).*#\1${ESCAPED_VALUE}#gi" "${CONFIG_FILE}"
```

target/scripts/helpers/utils.sh
polarathene: > Any reason to quote the equal sign?

Probably because `:-=` and `:=` are easy to mix up? 😅 

If you and @georglauterbach would like the quote-wrapping dropped to make the string value less distinct, I can do that. I just don't think it's helpful towards contributors that aren't as well-versed with bash and may confuse it as a bashism 🤷‍♂️

target/scripts/helpers/utils.sh
polarathene: > To keep it universal (able to handle other files which may have mixed cases), `$KEY` shouldn't be altered.

While I could agree with you there. I've not changed the existing behaviour, this was how the variable was handled previously.

It's only used for LDAP presently, but once my Config Templates PR is merged, this helper has no need to be tied to LDAP configs as it won't be used by anything.

I would ask if the lowercasing was dropped, how would it handle uppercase ENV keys matching lowercase config keys? It is probably better to defer that sort of change until a feature requires it. We'd then have an actual known use-case for it.

Presently it's an internal helper, so nothing but our scripts should be using it. I'm just providing the revision here as an extracted follow-up from the Config Templates PR. The functionality can remain without being used, but if it's dropped and later is brought back at least it'll be this improved/fixed version 😅

target/scripts/helpers/utils.sh
georglauterbach: I actually see the point here, and I am fine with keeping it. @casperklein is riht though,; we should not start quoting in other places.

casperklein (approved):

georglauterbach:
This is marked for v14.0.0. Will it be addresses for v14.0.0, or should we immediately postpone to v14.1.0?

PR #3527, opened 2023-09-05
fix: Ensure files are committed with `eol=lf` via `.gitattributes`

polarathene:
# Description

Slight project improvement, mainly beneficial if developing from a Windows machine.

Open to alternatives:
- A simpler file containing only `* text=auto`.
- Rejection due to existing EC linter (_preventing `CRLF` check-in is probably a good thing though?_).

---

Although our EditorConfig linter check can catch this mistake, it's probably useful to have a `.gitattributes` file to automatically convert to `LF` when necessary? (_Default allows checking in `CRLF` into the repo, as [this non-obvious CI test failure running on Linux demonstrates](https://github.com/docker-mailserver/docker-mailserver/pull/3524#issuecomment-1704744596)_).

The current commit is a bit overkill / verbose:
- We could trust `git` to make the decision with just `* text=auto` and rely on tests and linting to communicate failures? 🤷‍♂️ 
- Or drop most rules for niche files (_notably those that are extensionless_), I believe `git` will decide if a file is binary or not by it's content, and since we already git diffs generated for all relevant text files, those are not treated as binary.

The inline docs at the bottom, I'm not sure what your opinion is on keeping them. It's harmless and not really going to bother anyone, while providing a decent refresher/overview of the file format and rules (_vs the upstream docs_). They are [revised from an earlier `.gitattributes` I wrote](https://github.com/zellwk/zellwk.com/issues/17#issuecomment-481201054) years ago.

[Additional reference](https://github.com/zellwk/zellwk.com/pull/18#issuecomment-481203866):

> _Do note that if you do add more rules regarding text files, it will only affect new commits, it won't apply to the prior history, for that you'd need to rewrite the history, so it's good to commit this with new repos at the start of a project._
> 
> _Github has [a good resource on the topic](https://help.github.com/en/articles/dealing-with-line-endings), which also details how to rewrite history of the repo. At the end of the page the "Further reading" section links to a quality article: "[Mind the End of your Line](http://adaptivepatchwork.com/2012/03/01/mind-the-end-of-your-line/)" :)_

## Type of change

## Checklist:


georglauterbach (approved):
I actually do quite like the changes! 👍🏼

casperklein:
Disclaimer: Not going to block this / discuss to death 😆

I read a bit about .gitattributes and have mixed feelings. Personally I don't like it, when files are automatically changed (LF/CRLF in this case) on commit/checkout. If there are erroneous line endings, I prefer to get to know this by a failed lint test. So that I am informed that something did not fit and I can make it better in the future - instead of getting it silently fixed.

Every modern editor should be able to handle LF/CRLF correctly. It feels weird, to convert/alter all text files on checkout to another line ending CRLF (for example on windows). And later after the work is done, convert it back to LF when commiting. Instead, a proper editor should be used, so that converting beforehand is not necessary at all.

> I believe git will decide if a file is binary or not by it's content

If the first 1K bytes contains a null byte, it's considered binary.

BTW: Any idea, why the [mentioned test failure](https://github.com/docker-mailserver/docker-mailserver/pull/3524#issuecomment-1704744596) wasn't caught by eclint?

In .editorconfig we have

```
[*]
end_of_line = lf
```

I would have expected, that all text files are checked for the correct (LF) line ending.

polarathene:
**TL;DR:**
- `.gitattributes` sanitizes for the repo, only `eol=lf` should change check-out conversion, but this is enforced when it should be LF.
- On my current system (Windows), it seems to checkout without conversion, but it was also happy to commit without conversion, where VSCode was introducing the CRLF (_apparently for edited files sometimes too_).
- Nobody will want mixed line-endings, and for this project there's really no reason to want to commit with CRLF. This change should really only affect you if you're on Windows, but in a positive way.
- ECLint is doing it's job perfectly fine catching this. CI will skip it if ShellCheck fails first though.

---

> I read a bit about .gitattributes and have mixed feelings. Personally I don't like it, when files are automatically changed (LF/CRLF in this case) on commit/checkout.

I had read a lot on the topic myself and have had a few experiences where it was problematic over the years.

To be fair, I've been very verbose/explicit in the config here so you aren't really getting any surprises. It's similar to the linting that would catch it, but if you commit the change then the CI won't pick up on it via PR as the issue was prevented.

This just ensures that line endings are converted to LF during commit, as you really don't want CRLF (_especially mixed_) in the repo. If the file does not explicitly have `eol=lf`, but only `text` this allows for the local environment to checkout with CRLF if it wants to.

Especially since this is a project focused on files being copied into a linux container image, you have very little reason to want CRLF there. It will cause weird bugs like `.sh` scripts unable to find the binary from shebang hint due to that invisible CRLF, or as I recently experienced a test failure that told me `CONTENT` is not `CONTENT` 🤔 

---

> If there are erroneous line endings, I prefer to get to know this by a failed lint test. So that I am informed that something did not fit and I can make it better in the future - instead of getting it silently fixed.

The lint should work that way if you run it before committing. I often forget to, but since we contribute via PR instead of direct push to master it can be prevented sneaking into master.

I had no idea when the CI tests failed initially as the lint test was bailing at ShellCheck before EditorConfig, plus some tests were intermittent/unrelated test failures, so line nedings didn't cross my mind until EC Lint pointed it out.

That said I'm working on rewriting a docs page. Normally these say LF, but I just noticed that it's reporting CRLF in the editor right now, possibly it's mixed endings and after rewriting more than half of the page it detects as CRLF now, or perhaps it changed when I saved it 🤷‍♂️ 

If you're not working on Windows you're not going to have this surprise gotcha, it's better to prevent it for those that do, lints are great, but leveraging this git feature to fix all files when they're committed is even better.

---

> Every modern editor should be able to handle LF/CRLF correctly. It feels weird, to convert/alter all text files on checkout to another line ending (CRLF, for example on windows). And later after the work is done, convert it back to LF when commiting. Instead, a proper editor should be used, so that converting beforehand is not necessary at all.

I'm using VSCode, while for git it's one of the few other dev tools I like to use a GUI for instead of the CLI, I use GitKraken there. I consider both of these modern and quite capable.

In both cases I think they're just respecting the Git defaults, which is to use CRLF unless told otherwise. `.gitattributes` here as mentioned will only change the normalize to LF on commit. I could probably install an extension in VSCode for EditorConfig that might help prevent files getting changed to CRLF (_I thought it was only when I create new ones, but it seems to being doing that when editing docs too..._).

Regardless, we get other contributors and can't expect it'd be any better on their end. The CI lint tests will still catch that, but I can't think of a reason where `.gitattributes` isn't good for this project. Nothing should really need CRLF when it's all being used during a docker image build (_which I doubt converts back to LF?_).

---

> BTW: Any idea, why the [mentioned test failure](https://github.com/docker-mailserver/docker-mailserver/pull/3524#issuecomment-1704744596) wasn't caught by eclint?
> I would have expected, that all text files are checked for the correct (LF) line ending.

It did catch it, like mentioned above. Just requires the linter to reach that check before failing.

https://github.com/docker-mailserver/docker-mailserver/actions/runs/6070286374/job/16466057026

The test run before that one skipped ECLint due to ShellCheck failure:

https://github.com/docker-mailserver/docker-mailserver/actions/runs/6068011319/job/16460416288

While the test-case was failing it wasn't for that same reason as I hadn't yet adjusted for the white-space difference, so I wasn't confused at that point either.

casperklein:
Thanks for the detailed explanation 👍

Issue #3565, opened 2023-10-03
bug report: CI failure for `contributors.yml` (_due to files with `CRLF` conflicting with `.gitattributes`_)


polarathene:
### Description

The [`contributors.yml`](https://github.com/docker-mailserver/docker-mailserver/blob/8c0cfa0836bcf20e78a2d5d1ae02270004ee0aed/.github/workflows/contributors.yml) scheduled workflow [recently failed](https://github.com/docker-mailserver/docker-mailserver/actions/runs/6287887262) which caught a mishap (_my fault_) with `CRLF` line endings committed in these files:
- `test/config/ldap/openldap/schemas/postfix-book.ldif`
- `test/config/templates/dovecot-masters.cf`
- `test/config/templates/postfix-accounts.cf`

**Insights:**
- Those files were introduced by me in https://github.com/docker-mailserver/docker-mailserver/pull/3514 before [`.gitattributes` was added to prevent files checking in `CRLF` line endings](https://github.com/docker-mailserver/docker-mailserver/pull/3527).
- The Github Action did have a [successful run in-between those two PRs](https://github.com/docker-mailserver/docker-mailserver/actions/runs/6062330758)
- I also noticed [ECLint from the `linting.yml` workflow](https://github.com/docker-mailserver/docker-mailserver/blob/8c0cfa0836bcf20e78a2d5d1ae02270004ee0aed/.github/workflows/linting.yml#L24) did not catch this mistake either (_[even though config implies it should have](https://github.com/docker-mailserver/docker-mailserver/blob/8c0cfa0836bcf20e78a2d5d1ae02270004ee0aed/.editorconfig#L10)_).
  - **UPDATE:** This was because `make lint` actually uses [this `ecrc.json` config instead](https://github.com/docker-mailserver/docker-mailserver/blob/8c0cfa0836bcf20e78a2d5d1ae02270004ee0aed/test/linting/.ecrc.json#L7-L8) which excludes linting the `test/` directory.
  - The `test/` folder was excluded by: https://github.com/docker-mailserver/docker-mailserver/pull/2000
  - Full history was shared for eclint config changes in [my comment here](https://github.com/docker-mailserver/docker-mailserver/pull/2198#discussion_r711814762). I raised the concern that it was not apparent why these excludes continued to be expanded (_PRs introducing them did not provide much reasoning_).

---

## Logs

<details>
<summary>Contributors failing workflow log</summary>

[...]

PR #1046, opened 2018-10-07
Close Issue #998, Close Issue #999 - Support for additional postgrey options

millerjason:
Added support for:
- AUTO_WHITELIST_CLIENTS to whitelist hosts after multiple successes
- whitelist local (used for whitelists beyond defaults)
- whitelist recipients (used for whitelisting to specific accounts, like postmaster)

Updated documentation and tests.

PR #495, opened 2017-01-28
Added greylisting using postgrey

5ven:

PR #3566, opened 2023-10-03
ci(fix): Normalize for `.gitattributes` + improve `eclint` coverage

polarathene:
# Description

This is a follow-up PR to #3527 

- As per #3565 some files needed to fixed that had already been committed prior to #3527 , which was causing our `contributors.yml` workflow to fail.
- Additionally to resolve the problem in future with `linting.yml` + `make eclint`, I've revised the `.editorconfig` + `.ecrc.json` config files to be better aligned and provide improved coverage. Various files have been slightly modified to reflect the ECLint detected issues.

Postgrey had an init script [contributed in Feb 2017](https://github.com/docker-mailserver/docker-mailserver/pull/495) as part of the feature implementation.
- This was probably the file originally ignored with the `.init$` rule.
- In August 2017 we [introduced `supervisord` which made that redundant](https://github.com/docker-mailserver/docker-mailserver/commit/9e7959fafef48b0da620d0d04d0a547029785689#diff-6d88c6c2abe074135b8efac1cf0dca97dd288a71720cdcd6ac48c2a6f85dcf08L1210) (_we just [call it directly with options here](https://github.com/docker-mailserver/docker-mailserver/blob/bd96c1161e9f9b32bb0b3aaec4b930980a56945f/target/supervisor/conf.d/supervisor-app.conf#L88) now_), so I've dropped the file.
- That PR also contributed a `/etc/default/postgrey` file, which AFAIK isn't supporting anything either and we could drop it (_seems to just be a single ENV used by the init script for adding extra options_ 🤷‍♂️ ). A later [PR in 2018 added Postgrey support for whitelisting](https://github.com/docker-mailserver/docker-mailserver/pull/1046).
- Our [`postsrsd` integration has migrated from an `init` script](https://github.com/docker-mailserver/docker-mailserver/pull/3160#discussion_r1131053098) of it's own, while still [loading `/etc/default/postsrsd` config in it's supervisord command](https://github.com/docker-mailserver/docker-mailserver/blob/bd96c1161e9f9b32bb0b3aaec4b930980a56945f/target/supervisor/conf.d/supervisor-app.conf#L151). Postgrey could adopt something similar if worthwhile, although I think we plan to deprecate this service for removal sometime next year? 🤔 

Additional context between the two files that have meaningful changes `.ecrc.json` and `.editorconfig` can be found in the commit messages of their individual commits.

For reference:
- Default `eclint` configuration is [here](https://github.com/editorconfig-checker/editorconfig-checker#configuration).
- Default `eclint` excludes are [listed here](https://github.com/editorconfig-checker/editorconfig-checker#default-excludes).
- `test/test-files` is ignored, but it could be for the specific file affected if preferred. The failure is due to [this upstream `eclint` bug](https://github.com/editorconfig-checker/editorconfig-checker/issues/252). I've pinpointed the cause and deferred the fix to maintainers there.

Fixes https://github.com/docker-mailserver/docker-mailserver/issues/3565

## Type of change

## Checklist:


georglauterbach (approved):

rasa:
If my [PR](https://github.com/editorconfig-checker/editorconfig-checker/pull/284) is accepted, https://github.com/editorconfig-checker/editorconfig-checker/issues/252 will be fixed soon. @polarathene Thank you for the detailed error report and your patience.

Issue #3563, opened 2023-10-02
question: How to configure Traefik to proxy inbound mail connections via Port 25 (StartTLS)


vincentDcmps:
### Subject

I would like some feedback concerning a use case

### Description

HI I try to configure docker mail server behind traefik following thins works:
send internal mail from outside
send mail to outside
note: for that I must add "traefik.tcp.routers.esmtp.tls.passthrough=true" to esmtp and imap on traefik

but I can't receive mail from another domain behind traefik but telnet on port 25 works
I have try to put docker-mail server port 25 in front and in this case everything works
I have correctly put following parameter (using edge)

```postfix-master.cf (using edge)
submission/inet/smtpd_upstream_proxy_protocol=haproxy
submissions/inet/smtpd_upstream_proxy_protocol=haproxy
```
and 

```postfix-main.cf
postscreen_upstream_proxy_protocol = haproxy
```

try to change traefik service terminaison time but doesn't work too

I don't see any log for this request in DMS so I suposse that traefik drop request but don't know why or missing something in postscreen config for reverse proxy?

polarathene:
You've not filled a proper bug template. It's usually helpful not just for maintainers but others that arrive here with the same issue to have more information about your configuration.

Just to confirm, have you verified you can connect to Port 465 via Traefik? Two users are struggling with this currently with ThunderBird unable to connect via Traefik to submit mail, but successful if connecting to DMS port 465 directly:
- https://github.com/docker-mailserver/docker-mailserver/issues/3547#issuecomment-1743844660
- https://github.com/docker-mailserver/docker-mailserver/issues/3546

---

> HI I try to configure docker mail server behind traefik following thins works:
> send internal mail from outside
> send mail to outside
> note: for that I must add "traefik.tcp.routers.esmtp.tls.passthrough=true" to esmtp and imap on traefik

ESMTP is an improvement on SMTP, but it's not a specific port. You refer to sending mail, and that you must enable TLS passthrough for it to work.

Are you referring to port 465 or 587 here? Or also outbound port 25?
- Ports 587 (StartTLS) + 465 (_implicit TLS, this one needs passthrough_) are specifically for mail submission (_to be delivered internally, or sent outbound on port 25_).
- Port 25 inbound is for receiving mail only, not intended for submitting mail to be sent.
- Port 25 outbound is used when sending mail to another mail server without authentication.
[...]

PR #3568, opened 2023-10-05
docs: TLS - Include `passthrough=true` on implicit ports for Traefik example

vincentDcmps:
# Description

like discuss in #3563 I have add the passthrought option in reverse proxy doc

## Type of change

## Checklist:


docs/content/examples/tutorials/mailserver-behind-proxy.md
```diff
@@ -72,14 +72,15 @@ Feel free to add your configuration if you achieved the same goal using differen
           - "traefik.tcp.services.smtp.loadbalancer.server.port=25"
           - "traefik.tcp.services.smtp.loadbalancer.proxyProtocol.version=1"
           - "traefik.tcp.routers.smtp-ssl.rule=HostSNI(`*`)"
-          - "traefik.tcp.routers.smtp-ssl.tls=false"
           - "traefik.tcp.routers.smtp-ssl.entrypoints=smtp-ssl"
+          - "traefik.tcp.routers.smtp-ssl.tls.passthrough=true"
```
polarathene, line 76: 👍 

---

Reference:
- https://doc.traefik.io/traefik/routing/routers/#passthrough
- https://traefik.io/blog/traefik-2-tls-101-23b4fbee81f1/#what-about-passthrough

Doesn't require `tls=false`, but should hand the traffic over to DMS on port 465 encrypted, and DMS will handle the connection (_and presumably certs_).

Related:
- [`HostSNI` must either be an explicit value with no wildcard, or only `*`](https://stackoverflow.com/a/60858441).
- Ports that use StartTLS instead should not proxy through Traefik for TLS/certs, they need to establish a plain-text connection to DMS ports to upgrade to TLS explicitly, so `.tls=false` is appropriate for those TCP router port configs AFAIK as Traefik does not support StartTLS:
  - https://github.com/docker-mailserver/docker-mailserver/issues/3563#issuecomment-1744653719
  - https://github.com/docker-mailserver/docker-mailserver/issues/3563#issuecomment-1745728038

docs/content/examples/tutorials/mailserver-behind-proxy.md
```diff
@@ -72,14 +72,15 @@ Feel free to add your configuration if you achieved the same goal using differen
...
           - "traefik.tcp.routers.smtp-ssl.service=smtp-ssl"
           - "traefik.tcp.services.smtp-ssl.loadbalancer.server.port=465"
           - "traefik.tcp.services.smtp-ssl.loadbalancer.proxyProtocol.version=1"
           - "traefik.tcp.routers.imap-ssl.rule=HostSNI(`*`)"
           - "traefik.tcp.routers.imap-ssl.entrypoints=imap-ssl"
           - "traefik.tcp.routers.imap-ssl.service=imap-ssl"
+          - "traefik.tcp.routers.esmtp.tls.passthrough=true"
```
polarathene, line 83: I think you meant to reference the `imap-ssl` router not `esmtp`?:

```suggestion
          - "traefik.tcp.routers.imap-ssl.tls.passthrough=true"
```

I'm not familiar with why the `proxyProtocol.version` is sometimes 1 or 2, perhaps due to Postfix / Dovecot support? I understand it's to ensure that the original client IP is handed over correctly so it doesn't get misunderstood as directly from Traefik. But I'm not sure if that's required for each different supported way that Traefik routes to different DMS ports 🤷‍♂️

polarathene (changes requested):
I don't mind approving this change. Although I'd love it if there was a more consistent config experience with Traefik since several users reported recently having success and failure with different ports 🤔

polarathene (approved):

Issue #3558, opened 2023-09-30
bug report: Dovecot sieve error logged when timestamp of `.svbin` is shared with `.sieve` source file


polarathene:
### 📝 Preliminary Checks

- [X] I tried searching for an existing issue and followed the [debugging docs](https://docker-mailserver.github.io/docker-mailserver/latest/config/debugging/) advice, but still need assistance.

### 👀 What Happened?

- Dovecot attempts to use the `.svbin` that was compiled during startup, but notices it is "outdated" since it has the same timestamp as the `.sieve` file.
- Dovecot then compiles it again, and attempts to write the new compiled binary to the global dir but fails due to lack of permissions (_[this is expected](https://doc.dovecot.org/configuration_manual/sieve/usage/#manually-compiling-sieve-scripts), the vmail user should not have write access to this location_).
- Errors are logged and it continues to run the compiled binary successfully.

### 👟 Reproduction Steps

Failure seems to occur during this part of startup when the files share the same timestamp between `.sieve` and `.svbin` extensions:

https://github.com/docker-mailserver/docker-mailserver/blob/bd96c1161e9f9b32bb0b3aaec4b930980a56945f/target/scripts/startup/setup.d/dovecot.sh#L79-L81

In the `dovecot_sieve.bats` test, these two emails trigger a `before` + `after` global sieve scripts:

https://github.com/docker-mailserver/docker-mailserver/blob/89cb6d85b90f6070d53c15e455ebee916c899c7a/test/tests/parallel/set1/dovecot/dovecot_sieve.bats#L28-L31

```console
# Failure due to identical timestamps between `.sieve` + `.svbin`:
$ ls -l --time-style full-iso /usr/lib/dovecot/sieve-global/before/
-rw-r--r-- 1 dovecot root 107 2023-09-30 03:40:18.188085494 +0000 50-before.dovecot.sieve
-rw-r--r-- 1 dovecot root 309 2023-09-30 03:40:18.188085494 +0000 50-before.dovecot.svbin
$ ls -l --time-style full-iso /usr/lib/dovecot/sieve-global/after
-rw-r--r-- 1 dovecot root 149 2023-09-30 03:40:18.208085480 +0000 spam_to_junk.sieve
-rw-r--r-- 1 dovecot root 329 2023-09-30 03:40:18.208085480 +0000 spam_to_junk.svbin

# No failure (`.svbin` is newer):
[...]

PR #3550, opened 2023-09-27
feat: Allow changing the Dovecot vmail UID/GID via ENV

vincentDcmps:
# Description

There is a deployment scenario with remote storage (NFS / NAS) where the [appliance restricts what UID / GID values are allowed on the file system](https://github.com/docker-mailserver/docker-mailserver/issues/3548).

This causes a problem when using a volume bind mount for persisting the DMS `/var/mail` onto the host, where the UID/GID of `5000:5000` (_internally mapped to `docker:docker`_) that is used by Dovecot in DMS is not permitted.

Unlike a [similar issue with LDAP supporting per-user UID/GID (_where `chown -R` during startup and change detection adjusts ownership over `/var/mail`_)](https://github.com/docker-mailserver/docker-mailserver/issues/2238), this PR only needs to adjust the static vmail user configured for Dovecot. This can be supported easily with ENV.

Resolves: #3548 #405 #1611

## Type of change

Fixes a bug by introducing a new feature.

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

## Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [X] If necessary I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes

PR #3572, opened 2023-10-07
feat: Postfix permit DSN (Delivery Status Notification) only on authenticated ports (465 + 587)

allddd:
# Description

Checking the Postifx configuration, I noticed that there were no settings related to DSN.
In my opinion, it is always a good idea to partially disable this, as it can have some unexpected (negative) effects.

What I mean by partial disabling is disabling it on port 25, which will stop incoming email from requesting a DSN.
If the server simply sends such notifications/emails to anyone who requests them, this can easily be abused and redirected to any email address by spoofing.
Even if the intention was not to send email back to the spoofed address and it was just _normal_ spam, the backscatter can have an impact on reputation.

I have also noticed that such emails sent from a DMS instance reveal a little too much information.
For example, my tests showed that the email leaked whether an email address was an alias or not, as well as some of the software used on the server.

To ensure that this option works as intended, I have implemented some additional tests.
Apart from that, to make sure it wasn't interfering with anything else, I re-ran all the tests with `POSTFIX_DSN=0`, `1` and `2` as the default options.
This and manual testing showed no issues.

## Type of change

## Checklist:


polarathene (changes requested):
Thanks for spotting this issue and contributing a fix 😁  (_and with tests!_ 🥳 )

You can address the feedback or wait for input from other maintainers which might approve of dropping the ENV.

Would it make sense to restrict to authenticated by default for v13? @casperklein @georglauterbach 

---

If there is little concern for switching to this as default, we can probably drop the need for supporting an ENV with docs, (test is still relevant). Assuming the need to adjust the setting is niche, it should be supported well via our `postfix-master.cf` / `postfix-main.cf` overrides support, should anyone raise an issue about it that could be added to a FAQ entry?

polarathene:
## Reference

[Postfix docs on DSN support](https://www.postfix.org/DSN_README.html):

> _Delivery Status Notifications as described in [RFC 3464](https://tools.ietf.org/html/rfc3464). This gives senders control over successful and failed delivery notifications._

> _Just like reports of undeliverable mail, DSN reports of successful delivery can give away more information about the internal infrastructure than desirable._
> _Unfortunately, disallowing "success" notification requests requires disallowing other DSN requests as well. The RFCs do not offer the option to negotiate feature subsets._

> _This is not as bad as it sounds._
> _When you turn off DSN for remote inbound mail, remote senders with DSN support will still be informed that their mail reached your Postfix gateway successfully; they just will not get successful delivery notices from your internal systems._
> _Remote senders lose very little: they can no longer specify how Postfix should report delayed or failed delivery._

> - _Use the [`smtpd_discard_ehlo_keyword_address_maps`](https://www.postfix.org/postconf.5.html#smtpd_discard_ehlo_keyword_address_maps) feature if you wish to allow DSN requests from trusted clients but not from random strangers (see below for how to turn this off for all clients)_
> - _If you want to disallow all use of DSN requests from the network, use the [`smtpd_discard_ehlo_keywords`](https://www.postfix.org/postconf.5.html#smtpd_discard_ehlo_keywords) feature_

---

[`smtpd_discard_ehlo_keywords`](https://www.postfix.org/postconf.5.html#smtpd_discard_ehlo_keywords) setting:

> _A case insensitive list of EHLO keywords (pipelining, starttls, auth, etc.) that the Postfix SMTP server will not send in the EHLO response to a remote SMTP client._
> _Specify the `silent-discard` pseudo keyword to prevent this action from being logged._

---

[DSN backscatter article 2017](https://0xstubs.org/pitfalls-when-running-a-mail-server/):

> _Delivery status notification (DSN), widely known as bounces, inform the sender about emails that could not be delivered or are stuck in the mail queue. In [RFC 3464](https://tools.ietf.org/html/rfc3464) DSNs have been extended in that the sender can now request information about the delivery of their mails, including successful delivery. While this may be a nice idea, it makes generating backscatter trivial._
> _Spammers only need to send an email to a valid user of your mail server, set a certain sender address and request a DSN. Your server will happily send a confirmation to that sender address. Your server generates backscatter._
> _To avoid this, DSN should not be offered to foreign users/servers. The fact that the email was received by your server should be sufficient information for the sender requesting DSN._
> _But you may want to allow your own (authenticated) users to still use DSN. To achieve this, the option can be restricted to mails received via port 25 by not adding it to `main.cf` but `master.cf` instead_

[DSN backscatter article 2020](https://billauer.co.il/blog/2020/07/dsn-backscatter-spam-sendmail/):

> _Checking my own mail logs, the DSN is a virtually unused feature. So it’s probably just something spammers can take advantage of._
> _Turning off DSN is often recommended against in different sendmail guides, because it’s considered a “valuable feature” or so. As mentioned above, I haven’t seen it used by anyone else than spammers._
> _FAILURE doesn’t include the user not being known to the server, in which case the message is dropped anyhow without any DSN message generated. So as a spam trick, one can’t send mails to random addresses, and issue spam bounce messages because they failed._

[Postfix docs on Backscatter](http://www.postfix.org/BACKSCATTER_README.html) briefly cite a DSN issue with Outlook 2003.

[Microsoft 365 docs on DSNs](https://learn.microsoft.com/en-us/exchange/troubleshoot/email-delivery/ndr/non-delivery-reports-in-exchange-online):

> _When there's a problem delivering an email message that you sent, Microsoft 365 or Office 365 will generate an error code and often will send an email to let you know._
> _The email you receive is a delivery status notification, also known as a DSN or Bounce Message._
> _The most common type is called a non-delivery report (NDR) and they tell you that a message wasn't delivered._
> _Non-delivery can be caused by something as simple as a typo in an email address. NDRs include an error code that indicates why your email wasn't delivered_

Related DSN RFCs for reference:
- https://datatracker.ietf.org/doc/html/rfc3461
- https://datatracker.ietf.org/doc/html/rfc3464
- https://datatracker.ietf.org/doc/html/rfc6533

Dockerfile
```diff
@@ -201,7 +201,7 @@ RUN echo 'Reason_Message = Message {rejectdefer} due to: {spf}.' >>/etc/postfix-
 
 COPY target/fetchmail/fetchmailrc /etc/fetchmailrc_general
 COPY target/getmail/getmailrc /etc/getmailrc_general
-COPY target/postfix/main.cf target/postfix/master.cf /etc/postfix/
+COPY target/postfix/main.cf target/postfix/master.cf target/postfix/esmtp_access /etc/postfix/
```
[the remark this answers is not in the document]
allddd: Of course, as you say, the change is only visual.
My idea was to use the same name as it's called in the Postfix documentation.

mailserver.env
```diff
@@ -332,6 +332,17 @@ REPORT_SENDER=
...
+# Choose how Postfix handles DSN (Delivery Status Notification) requests
+# empty => Regardless of its origin, a DSN request is always fulfilled.
+#   1   => It can only be requested by authenticated users.
+#   2   => DSN is completely disabled. (not recommended)
+#
+# Note: If this option is left empty and you are using DMARC (reject, quarantine),
+# add an SPF record for the FQDN, since the email won't be signd with DKIM and will
+# be sent from mail.example.com instead of example.com.
+#
+POSTFIX_DSN=
```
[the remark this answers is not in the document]
allddd: * If other maintainers decide to keep the ENV, I will gladly to document it.

* This also makes sense and certainly looks nicer.

* As for the note, my idea was to inform users who don't know much about Postfix.
I figured you would just ignore it if you had already changed it to the bare domain, since you had to do it manually.

target/postfix/esmtp_access
```diff
@@ -0,0 +1,2 @@
+0.0.0.0/0   silent-discard, dsn
+::/0        silent-discard, dsn
```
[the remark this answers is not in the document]
allddd: You are right. The idea behind the file is to allow the user to mount it and customise / exclude some IPs if necessary. Using just `smtpd_discard_ehlo_keywords =` gives the same result, but a bit less flexibility.

target/scripts/startup/setup.d/postfix.sh
```diff
@@ -57,6 +57,15 @@ EOF
   __postfix__log 'trace' "Configuring virtual mailbox size limit to '${POSTFIX_MAILBOX_SIZE_LIMIT}'"
   postconf "virtual_mailbox_limit = ${POSTFIX_MAILBOX_SIZE_LIMIT}"
 
+  if [[ ${POSTFIX_DSN} -eq 1 ]]; then
+    __postfix__log 'trace' 'Disabling DSN for unauthenticated users'
+    sedfile -i -E '/^submissions?\>/a\  -o smtpd_discard_ehlo_keyword_address_maps=' /etc/postfix/master.cf
```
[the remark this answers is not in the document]
allddd: Your option is certainly the safer one in this case, and if the ENV is kept I have no problem with changing it.

test/test-files/email-templates/dsn-authenticated.txt
```diff
@@ -0,0 +1,13 @@
+EHLO mail
+AUTH PLAIN YWRkZWRAbG9jYWxob3N0LmxvY2FsZG9tYWluAGFkZGVkQGxvY2FsaG9zdC5sb2NhbGRvbWFpbgBteXBhc3N3b3Jk
+MAIL FROM: added@localhost.localdomain
+RCPT TO: user1@localhost.localdomain NOTIFY=success,failure
```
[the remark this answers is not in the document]
allddd: Yes, `NOTIFY=...` should be the only difference here.

test/tests/parallel/set3/mta/dsn.bats
```diff
@@ -0,0 +1,100 @@
...
+    # Required only for delivery via nc (_send_email)
+    --env PERMIT_DOCKER=container
+  )
+
+  _init_with_defaults
+  _common_container_setup 'CONTAINER_ARGS_ENV_CUSTOM'
+
+  _run_in_container setup email add 'added@localhost.localdomain' 'mypassword'
+  assert_success
+  _wait_until_change_detection_event_completes
```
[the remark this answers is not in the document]
allddd: Well, thank you very much, if only I had found it sooner.
I was unaware of any users other than `user1`.

allddd:
Thanks for the additional context and the review. 
As my setup would be permanently set to `POSTFIX_DSN=1` anyway, I am happy with whatever the maintainers decide.

If this is to be changed to the default, I would prefer to just change the default value and leave the ENV as additional configurability is not a bad thing.

polarathene:
> leave the ENV as additional configurability is not a bad thing.

We don't really like to incur more technical debt for ENV that can be better served by simple config adjustment. Otherwise we'd build up quite a bit of ENV contributions for the mere convenience of toggling one Postfix setting which has happened in the past.

It's more difficult to drop that support once we have it as we need to wait on a major release and issue deprecation notice usually. Prior to that it can add to maintenance burden.

allddd:
That makes sense, especially as this is not something that will be toggled on a regular basis.

polarathene:
Could either @casperklein or @georglauterbach chime in (_or express an intention to when time permits_), to provide your input for this contribution please?

[My review](https://github.com/docker-mailserver/docker-mailserver/pull/3572#pullrequestreview-1663118819) notes where your opinion is requested. If no activity is received in a weeks time from either of you, I'll assume you trust me to make the call myself so that the contributor can address the feedback and the PR can be merged.

Dockerfile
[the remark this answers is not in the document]
georglauterbach: Good point; I'd like to see the `.cidr` suffix as well :)

target/postfix/esmtp_access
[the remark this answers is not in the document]
georglauterbach: I prefer `smtpd_discard_ehlo_keywords =`, but won't block on this either way. Truth be told: most users will just set the variable value, and stop caring. Those that really care about being able to manually change the behavior do not need the flexibility anyway. But as I said, I really appreciate that you thought of this ❤️

georglauterbach (commented):
Very nice PR! Just a few touchups here and there. I mostly agree with @polarathene on the changes and added some feedback.

I highly appreciate the effort, and this is something we definitely need to take care of (since it may impact security).

polarathene (changes requested):
If @casperklein later chimes in with input to approach this differently, I'll handle those changes for you.

---

Please make the following changes.

- [ ] Update [`target/postfix/main.cf`](https://github.com/docker-mailserver/docker-mailserver/blob/894978ddd751f0d09e9f68e1a3de13926e17d432/target/postfix/main.cf) to include `smtpd_discard_ehlo_keywords = silent-discard, dsn`.
  - It will be the new default and enforced.
  - `git blame` should be sufficient for now to reference this PR for more information on the setting/decision if anyone needs to (_future maintainers / contributors_).
- [ ] Update `target/postfix/master.cf` to include `-o smtpd_discard_ehlo_keywords=` for both submission(s) services.
- [ ] Update the tests based on [review feedback](https://github.com/docker-mailserver/docker-mailserver/pull/3572/files#r1349595299) to simplify them:
  - `user1` account instead of creating `added` + remove change detection usage.
  - No `POSTFIX_DSN` ENV necessary.
  - While only the 2nd test case is needed, I'd like to keep the 1st and 3rd for robustness / reference.
    - This can be supported by [creating a `dsn` folder in the test config directory](https://github.com/docker-mailserver/docker-mailserver/tree/894978ddd751f0d09e9f68e1a3de13926e17d432/test/config), which can contain the `postfix-main.cf` and `postfix-master.cf` override files (_described in changelog entry below_).
    - The 1st + 3rd test cases can [selectively copy over the needed file](https://github.com/docker-mailserver/docker-mailserver/blob/894978ddd751f0d09e9f68e1a3de13926e17d432/test/tests/parallel/set3/mta/smtp_delivery.bats#L44-L46) which should be after `_init_with_defaults` line but before `_common_container_setup` line.
- [ ] Revert changes to:
  - `Dockerfile`
  - `mailserver.env`
  - `postfix.sh` + `variables-stack.sh`
  - Delete `esmtp_access`
- [ ] `dsn-authenticated.txt` - Replace the `AUTH PLAIN` line with the [`AUTH LOGIN` lines for `user1` account](https://github.com/docker-mailserver/docker-mailserver/blob/894978ddd751f0d09e9f68e1a3de13926e17d432/test/test-files/auth/smtp-auth-login.txt#L2C12-L3). Swap the `added` user in `MAIL FROM` and `From` lines to `user1`, it's still valid to login as `user1` and have their address as the sender and recipient.
- [ ] Document your change in [`CHANGELOG.md` under the "Unreleased => Breaking" section](https://github.com/docker-mailserver/docker-mailserver/blob/894978ddd751f0d09e9f68e1a3de13926e17d432/CHANGELOG.md?plain=1#L9) with:
    
  ````markdown
  - Postfix now defaults to supporting DSNs (_[Delivery Status Notifications](https://github.com/docker-mailserver/docker-mailserver/pull/3572#issuecomment-1751880574)_) only for authenticated users. This is a security measure to reduce spammer abuse of your DMS instance as a backscatter source.
    - If you need to modify this change, please let us know by opening an issue / discussion.
    - You can [opt-out (_enable DSNs_) via the `postfix-main.cf` override support](https://docker-mailserver.github.io/docker-mailserver/v12.1/config/advanced/override-defaults/postfix/) using the contents: `smtpd_discard_ehlo_keywords =`.
    - Likewise for authenticated users, the submission(s) ports (465 + 587) are configured internally via `master.cf` to keep DSNs enabled (_since authentication protects from abuse_). 
       
      If necessary, DSNs for authenticated users can be disabled via the `postfix-master.cf` override with the following contents:
       
      ```
      submission/inet/smtpd_discard_ehlo_keywords = silent-discard, dsn
      submissions/inet/smtpd_discard_ehlo_keywords = silent-discard, dsn
      ```
  ````

  Rendered preview:

   > - Postfix now defaults to supporting DSNs (_[Delivery Status Notifications](https://github.com/docker-mailserver/docker-mailserver/pull/3572#issuecomment-1751880574)_) only for authenticated users. This is a security measure to reduce spammer abuse of your DMS instance as a backscatter source.
   >   - If you need to modify this change, please let us know by opening an issue / discussion.
   >   - You can [opt-out (_enable DSNs_) via the `postfix-main.cf` override support](https://docker-mailserver.github.io/docker-mailserver/v12.1/config/advanced/override-defaults/postfix/) using the contents: `smtpd_discard_ehlo_keywords =`.
   >   - Likewise for authenticated users, the submission(s) ports (465 + 587) are configured internally via `master.cf` to keep DSNs enabled (_since authentication protects from abuse_). 
  >      
  >      If necessary, DSNs for authenticated users can be disabled via the `postfix-master.cf` override with the following contents:
  >      
  >     ```
  >     submission/inet/smtpd_discard_ehlo_keywords = silent-discard, dsn
  >     submissions/inet/smtpd_discard_ehlo_keywords = silent-discard, dsn
  >     ```

---

<details>
<summary>Prior feedback relevance (obsoleted by above checklist)</summary>

Feedback to apply:
- [Remove `esmtp_access` file in favor of `smtpd_discard_ehlo_keywords = silent-discard, dsn`](https://github.com/docker-mailserver/docker-mailserver/pull/3572/files#r1349585325). Users can easily adjust this via our supported Postfix overrides. Additional flexibility like this file (_and it's equivalent `*_address_maps` setting_) offers will be considered should users actually request a need for it.

Feedback to ignore:
- [`.cidr` extension for CIDR table config](https://github.com/docker-mailserver/docker-mailserver/pull/3572/files#r1349584361) (_no explicit demand anticipated, avoid maintenance burden_).
- [ENV docs](https://github.com/docker-mailserver/docker-mailserver/pull/3572/files#r1349585116) and feature logic (_level of configuration not likely needed_).
- [`postconf -P` instead of `sed` to adjust overrides in `master.cf`](https://github.com/docker-mailserver/docker-mailserver/pull/3572/files#r1349587161) (_`main.cf` will use the `smtpd_discard_ehlo_keywords` setting directly as a new default_).

</details>

polarathene:
> The only thing I had to do differently here was that the entries in `postfix-master.cf` had to be without spaces.

Good that we decided to test the other cases then 😁 

That wasn't something I thought would be an issue though, I'll have to look at our script for it or update our related docs to address that gotcha.

---

I'll return to this PR later today, my system is acting up and needs my attention to resolve (_WSL2 just ate a bunch of disk to the point I'm out of disk space + RAM and is unresponsive to shutdown, I can't seem to terminate the related process eating 10GB of memory_).

polarathene:
Apologies for the delay!

> The only thing I had to do differently here was that the entries in `postfix-master.cf` had to be without spaces.

Attempting to apply the setting with spaces (and double quote wrapped) via `postconf -P` fails with:

```
postconf: fatal: -Pe does not accept whitespace in parameter value
```

So doesn't seem like we can do anything about that 😅

---

Looks like all the feedback was addressed, so LGTM 👍 

Thank you for the excellent contribution to DMS! 😎

test/tests/parallel/set3/mta/dsn.bats
```diff
@@ -0,0 +1,85 @@
...
+@test "should never send a DSN" {
+
+  local LOG_DSN='delivery status notification'
+  local CONTAINER_ARGS_ENV_CUSTOM=(
+    # Required only for delivery via nc (_send_email)
+    --env PERMIT_DOCKER=container
+  )
+
+  _init_with_defaults
+  mv "${TEST_TMP_CONFIG}/dsn/postfix-master.cf" "${TEST_TMP_CONFIG}/postfix-master.cf"
```
polarathene, line 73:
```suggestion
  # Mirror default main.cf (disable DSN on ports 465 + 587 too):
  mv "${TEST_TMP_CONFIG}/dsn/postfix-master.cf" "${TEST_TMP_CONFIG}/postfix-master.cf"
```

test/tests/parallel/set3/mta/dsn.bats
```diff
@@ -0,0 +1,85 @@
...
+@test "should always send a DSN when requested" {
+
+  local LOG_DSN='delivery status notification'
+  local CONTAINER_ARGS_ENV_CUSTOM=(
+    # Required only for delivery via nc (_send_email)
+    --env PERMIT_DOCKER=container
+  )
+
+  _init_with_defaults
+  mv "${TEST_TMP_CONFIG}/dsn/postfix-main.cf" "${TEST_TMP_CONFIG}/postfix-main.cf"
```
polarathene, line 18:
```suggestion
  # Unset `smtpd_discard_ehlo_keywords` to allow DSNs by default on any `smtpd` service:
  mv "${TEST_TMP_CONFIG}/dsn/postfix-main.cf" "${TEST_TMP_CONFIG}/postfix-main.cf"
```

test/tests/parallel/set3/mta/dsn.bats
```diff
@@ -0,0 +1,85 @@
...
+
+  # A similar line is added to the log when a DSN (Delivery Status Notification) is sent:
+  #
+  # postfix/bounce[1023]: C943BA6B46: sender delivery status notification: DBF86A6B4CO
+  #
+  _run_in_container grep "${LOG_DSN}" /var/log/mail/mail.log
+  _should_output_number_of_lines 3
+}
+
+@test "should only send a DSN when requested from ports 465/587" {
```
polarathene, line 34:
```suggestion
# Defaults test case
@test "should only send a DSN when requested from ports 465/587" {
```

polarathene (changes requested):
I'll apply these, just some quick context for maintenance 👍

polarathene (approved):

polarathene:
Test cases 2 and 3 are failing when first sending the unauthenticated mail (`_send_email 'email-templates/dsn-unauthenticated'`).

I lack time to investigate right now.

---

[Test failure](https://github.com/docker-mailserver/docker-mailserver/actions/runs/6585549737/job/17892224465?pr=3572#step:6:32):

```
ok 27 [DSN] should always send a DSN when requested in 5774ms
not ok 28 [DSN] should only send a DSN when requested from ports 465/587 in 4079ms
#  from function `_send_email' in file test/helper/sending.bash, line 33,
#  in test file test/tests/parallel/set3/mta/dsn.bats, line 47)
#   `_send_email 'email-templates/dsn-unauthenticated'' failed
#
# -- command failed --
# status : 1
# output :
# --
#
# dms-test_dsn
not ok 29 [DSN] should never send a DSN in 4101ms
#  from function `_send_email' in file test/helper/sending.bash, line 33,
#  in test file test/tests/parallel/set3/mta/dsn.bats, line 79)
#   `_send_email 'email-templates/dsn-unauthenticated'' failed
#
# -- command failed --
# status : 1
# output :
# --
```

test/tests/parallel/set3/mta/dsn.bats
```diff
@@ -0,0 +1,88 @@
+load "${REPOSITORY_ROOT}/test/helper/setup"
+load "${REPOSITORY_ROOT}/test/helper/common"
+
+BATS_TEST_NAME_PREFIX='[DSN] '
+CONTAINER_NAME='dms-test_dsn'
```
polarathene, line 5: At a glance, I think the issue could be from here.

- The test goes straight into test cases, no `setup_file()`.
- The `teardown()` default should handle this inbetween without issue though, so a shared `CONTAINER_NAME` shouldn't be an issue due to sequential test run 🤷‍♂️ (_parallel is only relevant to separate test files concurrently, not their individual test cases within_)
- But given that the 2nd and 3rd test cases are the ones that fail, it does seem like the containers may have failed to start correctly, thus the attempt to send mail to port 25 fails.

allddd:
I have to say this is very strange, the failures seem to be completely random.
I ran all the tests twice before pushing without any problems.
Now I have tried again and it has failed, but then it succeeds two times after that.

This was not the case with the old configuration (using `/etc/postfix/esmtp_access`), so I tried changing it back to that and it just works, which makes it even stranger.
It would be good if someone more familiar with the test suite could look into this, I'm not sure exactly what the problem is.

polarathene:
> I have to say this is very strange, the failures seem to be completely random.
> I ran all the tests twice before pushing without any problems.
> Now I have tried again and it has failed, but then it succeeds two times after that.

That's interesting. I remember a recent PR a test case was revised for an error log check that was also proving to be spotty in behaviour unrelated to the PR:
- https://github.com/docker-mailserver/docker-mailserver/pull/3550#discussion_r1339424344
- https://github.com/docker-mailserver/docker-mailserver/pull/3550#issuecomment-1738509088

I don't seem to have mentioned it explicitly there, but I think it was related to [this report](https://github.com/docker-mailserver/docker-mailserver/issues/3558) for sieve errors. Basically there is a non-deterministic behaviour with file creation during startup, and if that creates two related files with a timestamp at the exact same time, instead of one expecting to be slightly newer, then it gets caught by the error log check added in that other test.

Doesn't seem to trip up our CI so far though AFAIK. This might be related, hard to say.

---

@georglauterbach you've worked on this helper IIRC, any advice? I assume we could maybe `assert_output` partially on something prior to the `assert_success` as that would give a bit more context to the failure when it doesn't match successful delivery expectations?

@allddd for the time being I think since this is the only test triggering the failure for that helper, we'll need to block merging the PR. It's not unusual for tests to be flimsy for some reason, but since it's for a common helper method used elsewhere something isn't right 😅 

---

If I find time I'll look into it.

My process is pretty simple, I just add a `sleep 600` prior to the failing part of the test, then `docker exec -it test_name_here bash` and then I can run the command (_or do that instead of `bash`_). A little annoying when the test fails inconsistently though.

Success or failure in this case should be fine as we should be able to use the output to `assert_output` / `refute_output` like other tests do with partial matching on that. When the `assert_output` / `refute_output` fails, it should log that failing output. Mostly curious what the delivery failure output is caused by, so we can address that.

test/tests/parallel/set3/mta/dsn.bats
```diff
@@ -0,0 +1,88 @@
...
+
+  local LOG_DSN='delivery status notification'
+  local CONTAINER_ARGS_ENV_CUSTOM=(
+    # Required only for delivery via nc (_send_email)
+    --env PERMIT_DOCKER=container
+  )
+
+  _init_with_defaults
+  # Unset `smtpd_discard_ehlo_keywords` to allow DSNs by default on any `smtpd` service:
+  mv "${TEST_TMP_CONFIG}/dsn/postfix-main.cf" "${TEST_TMP_CONFIG}/postfix-main.cf"
```
georglauterbach, line 19: Why us `mv` and not `cp` being used here?

test/tests/parallel/set3/mta/dsn.bats
```diff
@@ -0,0 +1,88 @@
...
+
+BATS_TEST_NAME_PREFIX='[DSN] '
+CONTAINER_NAME='dms-test_dsn'
+
+function teardown() { _default_teardown ; }
+
+@test "should always send a DSN when requested" {
+
+  local LOG_DSN='delivery status notification'
+  local CONTAINER_ARGS_ENV_CUSTOM=(
```
georglauterbach, line 12: Following our naming convention, this should be

```suggestion
local CUSTOM_SETUP_ARGUMENTS=(
```

test/tests/parallel/set3/mta/dsn.bats
```diff
@@ -0,0 +1,88 @@
...
+    # Required only for delivery via nc (_send_email)
+    --env PERMIT_DOCKER=container
+  )
+
+  _init_with_defaults
+  # Unset `smtpd_discard_ehlo_keywords` to allow DSNs by default on any `smtpd` service:
+  mv "${TEST_TMP_CONFIG}/dsn/postfix-main.cf" "${TEST_TMP_CONFIG}/postfix-main.cf"
+  _common_container_setup 'CONTAINER_ARGS_ENV_CUSTOM'
```
georglauterbach, line 20: Consequently

```suggestion
  _common_container_setup 'CUSTOM_SETUP_ARGUMENTS'
```

test/tests/parallel/set3/mta/dsn.bats
```diff
@@ -0,0 +1,88 @@
...
+
```
georglauterbach, line 21: I'd add

```suggestion

  _wait_for_service postfix
  _wait_for_smtp_port_in_container
```

test/tests/parallel/set3/mta/dsn.bats
```diff
@@ -0,0 +1,88 @@
...
+  _send_email 'email-templates/dsn-unauthenticated'
+  _send_email 'email-templates/dsn-authenticated' '0.0.0.0 465'
+  _send_email 'email-templates/dsn-authenticated' '0.0.0.0 587'
+  _wait_for_empty_mail_queue_in_container
```
georglauterbach, line 25: Very good, but checking here is the best way of doing it:

```suggestion
export MAIL_ID1=$(_send_email_and_get_id 'email-templates/dsn-unauthenticated')
  export MAIL_ID2=$(_send_email_and_get_id email-templates/dsn-authenticated' '0.0.0.0 465')
  export MAIL_ID3=$(_send_email_and_get_id 'email-templates/dsn-authenticated' '0.0.0.0 587')

  for ID in MAIL_ID{1,2,3}; do
    [[ -n ${!ID} ]] || { echo "${ID} is empty - aborting!" ; return 1 ; }
  done
```

This way, you can later use

```bash
_print_mail_log_for_id "${MAIL_ID1}"
```

and run asserts on the log for this email.

test/tests/parallel/set3/mta/dsn.bats
```diff
@@ -0,0 +1,88 @@
...
+  local LOG_DSN='delivery status notification'
+  local CONTAINER_ARGS_ENV_CUSTOM=(
+    # Required only for delivery via nc (_send_email)
+    --env PERMIT_DOCKER=container
+  )
+
+  _init_with_defaults
+  _common_container_setup 'CONTAINER_ARGS_ENV_CUSTOM'
+
+  _send_email 'email-templates/dsn-unauthenticated'
```
georglauterbach, line 47: Same here: the SMTP port should be checked, and we should wait for Postfix to come up (see my previous comments).

test/tests/parallel/set3/mta/dsn.bats
```diff
@@ -0,0 +1,88 @@
...
+    # Required only for delivery via nc (_send_email)
+    --env PERMIT_DOCKER=container
+  )
+
+  _init_with_defaults
+  # Mirror default main.cf (disable DSN on ports 465 + 587 too):
+  mv "${TEST_TMP_CONFIG}/dsn/postfix-master.cf" "${TEST_TMP_CONFIG}/postfix-master.cf"
+  _common_container_setup 'CONTAINER_ARGS_ENV_CUSTOM'
+
+  _send_email 'email-templates/dsn-unauthenticated'
```
georglauterbach, line 79: Same here (see comment above).

test/tests/parallel/set3/mta/dsn.bats
```diff
@@ -0,0 +1,88 @@
...
+  _send_email 'email-templates/dsn-authenticated' '0.0.0.0 465'
+  _send_email 'email-templates/dsn-authenticated' '0.0.0.0 587'
+
+  _run_in_container grep "${LOG_DSN}" /var/log/mail/mail.log
+  _should_output_number_of_lines 2
+}
+
+@test "should never send a DSN" {
+
+  local LOG_DSN='delivery status notification'
```
georglauterbach, line 68: This seems to be the same for all test cases, hence it can be made global I guess.

georglauterbach (changes requested):
Overall LGTM 👍🏼 Nice PR!

I have added review feedback that you can apply, but the test needs a bit more work. I know, annoying. Please have a look at [this test for Rspamd that I wrote](https://github.com/docker-mailserver/docker-mailserver/blob/master/test/tests/parallel/set1/spam_virus/rspamd_full.bats). Moreover, the [Postscreen test](https://github.com/docker-mailserver/docker-mailserver/blob/master/test/tests/parallel/set1/spam_virus/postscreen.bats) should provide you with a structure of how a test with multiple containers should look like.

I advocate for using `setup_file` here, using `export CONTAINER_NAME` and then setting it on-demand. This provides a unified place for starting containers. Moreover, we need to pay attention to using `_wait_for_service postfix` and `_wait_for_smtp_port_in_container` which could be important for this test.

With my proposals when using `export MAIL_IDX=$(_send_email_and_get_id ...) you'd also gain the advantage of better checks against logs for a specific email.

---

If you have questions about this, just ping me :)

allddd:
Thanks for the feedback @georglauterbach, it helped me get it stable again. :)
Now I was able to run the tests several times without any errors.
I will address all comments here as it is easier for me to do so:

---

> I advocate for using setup_file here, using export CONTAINER_NAME and then setting it on-demand.

It is indeed nicer this way, especially with multiple containers.

---

> Why is mv and not cp being used here?

No particular reason, it was used in [the example](https://github.com/docker-mailserver/docker-mailserver/blob/894978ddd751f0d09e9f68e1a3de13926e17d432/test/tests/parallel/set3/mta/smtp_delivery.bats#L44-L46) @polarathene sent, so I just copied it.

---

> Following our naming convention, this should be `CUSTOM_SETUP_ARGUMENTS`

I've tried to make it look as similar as possible. As for why I used it, I think it was because some of the first tests I looked at also used it.

```
grep -rl 'CONTAINER_ARGS_ENV_CUSTOM' docker-mailserver/

docker-mailserver/test/tests/parallel/set1/dovecot/dovecot_sieve.bats
docker-mailserver/test/tests/parallel/set3/mta/lmtp_ip.bats
docker-mailserver/test/tests/parallel/set3/mta/smtp_delivery.bats
docker-mailserver/test/tests/parallel/set3/container_configuration/process_check_restart.bats
docker-mailserver/test/tests/parallel/set3/scripts/setup_cli.bats
docker-mailserver/test/tests/serial/tests.bats
```

---

> I'd add `_wait_for_service postfix` & `_wait_for_smtp_port_in_container`

I think that this actually solved the problem, thank you very much.

---

> With my proposals when using export MAIL_IDX=$(_send_email_and_get_id ...) you'd also gain the advantage of better checks against logs for a specific email.

I thought this would be a great option as it would allow me to `grep` more accurately, but unfortunately it does not work in this case.

```
_run_in_container grep "${MAIL_ID}: ${LOG_DSN}" /var/log/mail/mail.log
```

As we currently send email using `nc`, this is not really possible as no email is sent at all, so there is no ID to get.
Here is the log output:

```
Oct 21 19:47:55 mail postfix/smtpd[704]: connect from localhost[127.0.0.1]
Oct 21 19:47:55 mail opendmarc[542]: ignoring connection from localhost
Oct 21 19:47:55 mail postfix/smtpd[704]: improper command pipelining after HELO from localhost[127.0.0.1]: MAIL FROM: user@external.tld\nRCPT TO: user1@localhost.localdomain NOTIFY=success,failure\nDATA\nFrom:
Oct 21 19:47:55 mail postfix/smtpd[704]: warning: non-SMTP command from localhost[127.0.0.1]: From: Docker Mail Server <dockermailserver@external.tld>
Oct 21 19:47:55 mail postfix/smtpd[704]: disconnect from localhost[127.0.0.1] helo=1 mail=1 rcpt=0/1 data=0/1 unknown=0/1 commands=2/5
```

Even if this was not the case, I'm not sure how accurate the ID extraction would be, as I send multiple emails and DSNs in the same container.

```
# This functions assumes **no concurrent sending of emails to the same container**!
# If two clients send simultaneously, there is no guarantee the correct ID is
# chosen. Sending more than one mail at any given point in time with this function
# is UNDEFINED BEHAVIOR!
```

test/tests/parallel/set3/mta/dsn.bats
polarathene: Normally we use `mv` as `_init_with_defaults` would copy the config dir into a temporary location for the container being setup (directory with the container name).

Since that is meant to be isolated, the pattern here is just moving from the dsn directory to the root test dir, as this whole directory is mounted into `/tmp/docker-mailserver`.

Not the greatest solution, but that's how it's been for quite some time. I'm not too bothered as it works, and the eventual plan is to switch to `compose.yaml` with `:ro` mounts where possible.

---

`mv` or `cp`, doesn't matter much either way 🤷‍♂️ (_`mv` just communicates the above a bit better_)

polarathene:
> > Following our naming convention, this should be `CUSTOM_SETUP_ARGUMENTS`
> 
> I've tried to make it look as similar as possible. As for why I used it, I think it was because some of the first tests I looked at also used it.

Good spotting, that'd be from an earlier iteration when this approach was only focused on ENV adjustment, but now just appends args to the [`docker create` command](https://github.com/docker-mailserver/docker-mailserver/blob/811a7698459ebd40cbd520740c2074f4ad7402a6/test/helper/setup.bash#L154).

> > I'd add `_wait_for_service postfix` & `_wait_for_smtp_port_in_container`
> 
> I think that this actually solved the problem, thank you very much.

Thanks @georglauterbach 🙏 

I've not been working on tests for a while that I thought this was implicit, but we don't always need it to minimize container startup time, so thanks for catching that 👍

test/tests/parallel/set3/mta/dsn.bats
```diff
@@ -0,0 +1,95 @@
...
+    --env PERMIT_DOCKER=container
+  )
+
+  export CONTAINER_NAME=${CONTAINER1_NAME}
+  _init_with_defaults
+  # Unset `smtpd_discard_ehlo_keywords` to allow DSNs by default on any `smtpd` service:
+  cp "${TEST_TMP_CONFIG}/dsn/postfix-main.cf" "${TEST_TMP_CONFIG}/postfix-main.cf"
+  _common_container_setup 'CUSTOM_SETUP_ARGUMENTS'
+  _wait_for_service postfix
+  _wait_for_smtp_port_in_container
```
polarathene, line 26: You could move these wait lines into each test case.

Not required, I'm fine with this as-is; I just thought I'd mention it (_slight benefit described below_).

---

The first test won't run until all 3 containers start sequentially with these blocking calls. Probably only a few seconds, so I'm not too fussed.

Some test files you'll notice handle this in each test case, as you had done before, which uses the simple `teardown()`. Lower memory overhead as well, but I don't think it's that much for a minimal container setup.

polarathene (approved):
Glad the testing issue was resolved, LGTM! 👍

test/tests/parallel/set3/mta/dsn.bats
allddd: If it's important I could change it, just not at the moment as it's a bit late and if I change it I'd want to also run some tests to make sure it doesn't break anything.

If you want to merge earlier, you can also change it if find have time, it should not be a problematic change.

polarathene: I don't mind, it's an easy optimization to spot and address later if someone cares to do that.

We'll wait for @georglauterbach to chime in, if they approve then this can be merged as-is :)

allddd: I tested it just to be sure:

`_wait_for_*` in `setup_file()`
```
real	0m14.120s
user	0m1.006s
sys	0m0.434s
```

`_wait_for_*` in `@test`
```
real	0m14.030s
user	0m0.987s
sys	0m0.493s
```

polarathene: Tested how? The concern was about all 3 being setup in `setup_file()`, this means:
- Container 1 start and wait.
- Container 2 start and wait after Container 1 is ready.
- Container 3 start and wait after Container 2 is ready.

There is a staggered wait sequence as a result. The waits could all be handled at the end of `setup_file()` which would minimize that. You'd still have a bit more delay / resources used spinning up multiple container instances, rather than one at a time per test-case (_since each test case has it's own container, rather than multiple test cases that can share the same container config_).

The wait time may vary a bit (_probably not by much, but can be a couple seconds IIRC, as the inconsistent test failures showed earlier_).

georglauterbach: No need to change this, it looks good. I may go over it later since the review feedback already was quite extensive. 👍🏼

georglauterbach (approved):
LGTM 👍🏼

Issue #3570, opened 2023-10-06
rspamd does not show rejected GKUBE mails in the history and spam "add header" is not delivered to junk


MohammedNoureldin:
### 📝 Preliminary Checks

- [X] I tried searching for an existing issue and followed the [debugging docs](https://docker-mailserver.github.io/docker-mailserver/latest/config/debugging/) advice, but still need assistance.

### 👀 What Happened?

I am not sure if this issue is specific to rspamd or to Docker mailserver.

When I try to send/receive spam message, in the statistics, I see an increasing in the rejected mails:

![image](https://github.com/docker-mailserver/docker-mailserver/assets/14913147/3d046439-5519-485d-b3fe-c46c40ad52d8)

and I see that the mail was rejected in the log):

```
 23:47:30 mail postfix/smtps/smtpd[16193]: connect from unknown[52.97.X.197]
Oct  6 23:47:30 mail postfix/smtps/smtpd[16193]: Anonymous TLS connection established from unknown[52.97.X.197]: TLSv1.3 with cipher TLS_AES_256_GCM_SHA384 (256/256 bits) key-exchange ECDHE (P-384) server-signature RSA-PSS (2048 bits) server-digest SHA256
Oct  6 23:47:30 mail postfix/smtps/smtpd[16193]: 8C020CE: client=unknown[52.97.X.197], sasl_method=LOGIN, sasl_username=administrator@DOMAIN.com
Oct  6 23:47:30 mail postfix/sender-cleanup/cleanup[16197]: 8C020CE: message-id=<AM0PR01MB547344B723EAB67EBA44DBA4A6C9A@AM0PR01MB5473.eurprd01.prod.exchangelabs.com>
Oct  6 23:47:30 mail postfix/sender-cleanup/cleanup[16197]: 8C020CE: replace: header MIME-Version: 1.0 from unknown[52.97.X.197]; from=<administrator@DOMAIN.com> to=<administrator@DOMAIN.com> proto=ESMTP helo=<AM0PR01MB5473.eurprd01.prod.exchangelabs.com>: MIME-Version: 1.0
[...]

PR #3576, opened 2023-10-10
refactor: `logrotate` setup + rspamd log path + tests log helper fallback path

georglauterbach:
# Description

~~In order to allow users to more easily go through log files, I added `most` to the set of packages we install. This adds +122kB in size, which I think is negligible.~~

~~Moreover, I added a section to the Rspamd documentation about where to find the corresponding log files.~~

I adjusted this PR to only include changes to logrotate and to the Rspamd log. In detail, I improved `_setup_logrotate` by simplifying it and making it more readable. Moreover, I adjusted the Rspamd log file location and added a logrotate config for the Rspamd log file.

I will provide a follow-up PR that includes changes about adding and removing packages, as discussed [here](https://github.com/docker-mailserver/docker-mailserver/pull/3576#discussion_r1353455195).

Associated: #3570

## Type of change

## Checklist:

### DIFF
--- a/docs/content/config/security/rspamd.md
+++ b/docs/content/config/security/rspamd.md
@@ -69,6 +69,10 @@ DMS does not supply custom values for DNS servers to Rspamd. If you need to use 
 
     This setting is enabled to not allow spam to proceed just because DNS requests did not succeed. It could deny legitimate e-mails to pass though too in case your DNS setup is incorrect or not functioning properly.
 
+### Logs
+
+You can find the Rspamd logs at `/var/log/mail/rspamd.log`, and the corresponding logs for [Redis](#persistence-with-redis), if it is enabled, at `/var/log/supervisor/rspamd-redis.log`. We recommend inspecting these logs (with `docker exec -ti <CONTAINER NAME> cat /var/log/mail/rspamd.log`) in case Rspamd does not work as expected.
+
 ### Modules
 
 You can find a list of all Rspamd modules [on their website][rspamd-docs-modules].
--- a/docs/content/config/security/rspamd.md
+++ b/docs/content/config/security/rspamd.md
@@ -71,7 +71,7 @@ DMS does not supply custom values for DNS servers to Rspamd. If you need to use 
 
 ### Logs
 
-You can find the Rspamd logs at `/var/log/mail/rspamd.log`, and the corresponding logs for [Redis](#persistence-with-redis), if it is enabled, at `/var/log/supervisor/rspamd-redis.log`. We recommend inspecting these logs (with `docker exec -ti <CONTAINER NAME> cat /var/log/mail/rspamd.log`) in case Rspamd does not work as expected.
+You can find the Rspamd logs at `/var/log/mail/rspamd.log`, and the corresponding logs for [Redis](#persistence-with-redis), if it is enabled, at `/var/log/supervisor/rspamd-redis.log`. We recommend inspecting these logs (with `docker exec -it <CONTAINER NAME> cat /var/log/mail/rspamd.log`) in case Rspamd does not work as expected.
 
 ### Modules
 
--- a/target/scripts/start-mailserver.sh
+++ b/target/scripts/start-mailserver.sh
@@ -91,7 +91,6 @@ function _register_functions() {
   _register_setup_function '_setup_dovecot_hostname'
 
   _register_setup_function '_setup_postfix_early'
-  _register_setup_function '_setup_logrotate'
   _register_setup_function '_setup_fetchmail'
   _register_setup_function '_setup_fetchmail_parallel'
 
@@ -106,6 +105,7 @@ function _register_functions() {
   fi
 
   _register_setup_function '_setup_postfix_late'
+  _register_setup_function '_setup_logrotate'
   _register_setup_function '_setup_mail_summary'
   _register_setup_function '_setup_logwatch'
 
--- a/target/scripts/start-mailserver.sh
+++ b/target/scripts/start-mailserver.sh
@@ -82,10 +82,8 @@ function _register_functions() {
   _register_setup_function '_setup_opendmarc' # must come after `_setup_opendkim`
   _register_setup_function '_setup_policyd_spf'
 
-  _register_setup_function '_setup_logrotate'
-
   _register_setup_function '_setup_security_stack'
-  _register_setup_function '_setup_rspamd' # must come after _setup_logrotate
+  _register_setup_function '_setup_rspamd'
 
   _register_setup_function '_setup_ssl'
   _register_setup_function '_setup_docker_permit'
@@ -93,6 +91,7 @@ function _register_functions() {
   _register_setup_function '_setup_dovecot_hostname'
 
   _register_setup_function '_setup_postfix_early'
+  _register_setup_function '_setup_logrotate'
   _register_setup_function '_setup_fetchmail'
   _register_setup_function '_setup_fetchmail_parallel'
 
--- a/target/scripts/start-mailserver.sh
+++ b/target/scripts/start-mailserver.sh
@@ -82,8 +82,10 @@ function _register_functions() {
   _register_setup_function '_setup_opendmarc' # must come after `_setup_opendkim`
   _register_setup_function '_setup_policyd_spf'
 
+  _register_setup_function '_setup_logrotate'
+
   _register_setup_function '_setup_security_stack'
-  _register_setup_function '_setup_rspamd'
+  _register_setup_function '_setup_rspamd' # must come after _setup_logrotate
 
   _register_setup_function '_setup_ssl'
   _register_setup_function '_setup_docker_permit'
@@ -105,7 +107,6 @@ function _register_functions() {
   fi
 
   _register_setup_function '_setup_postfix_late'
-  _register_setup_function '_setup_logrotate'
   _register_setup_function '_setup_mail_summary'
   _register_setup_function '_setup_logwatch'
 
--- a/target/scripts/startup/setup.d/log.sh
+++ b/target/scripts/startup/setup.d/log.sh
@@ -13,7 +13,7 @@ function _setup_logs_general() {
 function _setup_logrotate() {
   _log 'debug' 'Setting up logrotate'
 
-  if [[ ${LOGROTATE_INTERVAL} =~ ^(dai|week|month)ly$ ]]; then
+  if [[ ${LOGROTATE_INTERVAL} =~ ^(daily|weekly|monthly)$ ]]; then
     _log 'trace' "Logrotate interval set to ${LOGROTATE_INTERVAL}"
   else
     _log 'warn' "Value '${LOGROTATE_INTERVAL}' for LOGROTATE_INTERVAL is invalid - falling back to default"
--- a/target/scripts/startup/setup.d/log.sh
+++ b/target/scripts/startup/setup.d/log.sh
@@ -13,31 +13,25 @@ function _setup_logs_general() {
 function _setup_logrotate() {
   _log 'debug' 'Setting up logrotate'
 
-  LOGROTATE='/var/log/mail/mail.log\n{\n  compress\n  copytruncate\n  delaycompress\n'
+  if [[ ${LOGROTATE_INTERVAL} =~ ^(dai|week|month)ly$ ]]; then
+    _log 'trace' "Logrotate interval set to ${LOGROTATE_INTERVAL}"
+  else
+    _log 'warn' "Value '${LOGROTATE_INTERVAL}' for LOGROTATE_INTERVAL is invalid - falling back to default"
+    export LOGROTATE_INTERVAL='weekly'
+    # shellcheck disable=SC2034
+    VARS[LOGROTATE_INTERVAL]='weekly'
+  fi
 
-  case "${LOGROTATE_INTERVAL}" in
-    ( 'daily' )
-      _log 'trace' 'Setting postfix logrotate interval to daily'
-      LOGROTATE="${LOGROTATE}  rotate 4\n  daily\n"
-      ;;
-
-    ( 'weekly' )
-      _log 'trace' 'Setting postfix logrotate interval to weekly'
-      LOGROTATE="${LOGROTATE}  rotate 4\n  weekly\n"
-      ;;
-
-    ( 'monthly' )
-      _log 'trace' 'Setting postfix logrotate interval to monthly'
-      LOGROTATE="${LOGROTATE}  rotate 4\n  monthly\n"
-      ;;
-
-    ( * )
-      _log 'warn' 'LOGROTATE_INTERVAL not found in _setup_logrotate'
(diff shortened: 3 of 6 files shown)


target/scripts/build/packages.sh
```diff
@@ -65,7 +65,7 @@ function _install_packages() {
     apt-transport-https bind9-dnsutils binutils bsd-mailx
     ca-certificates curl dbconfig-no-thanks
     dumb-init ed gnupg iproute2 iputils-ping
-    libdate-manip-perl libldap-common
+    less libdate-manip-perl libldap-common
```
polarathene, line 68: We have a few packages being introduced specifically for debugging, could we instead group these like we do with others to make that distinction clear?

`bind9-dnsutils` was added just for `dig` IIRC, and `iputils-ping` for `ping`.

polarathene (changes requested):

target/scripts/build/packages.sh
georglauterbach: I can do that

georglauterbach: Done in [375d947](https://github.com/docker-mailserver/docker-mailserver/pull/3576/commits/375d947b762bd8726f1aa340c3b0ffc818eb5f8a)

georglauterbach:
> Instead of `less`, we could use `most`. It's the newest and feature rich pager tool:
> 
> `more < less < most`.

Done :D

target/scripts/build/packages.sh
```diff
@@ -43,10 +43,6 @@ function _install_postfix() {
 function _install_packages() {
   _log 'debug' 'Installing all packages now'
 
-  declare -a ANTI_VIRUS_SPAM_PACKAGES
-  declare -a CODECS_PACKAGES MISCELLANEOUS_PACKAGES
-  declare -a POSTFIX_PACKAGES MAIL_PROGRAMS_PACKAGES
-
```
georglauterbach, line 49: As @casperklein pointed out earlier (in another PR), `declate -a` is pretty much redundant, so I removed it here.

casperklein (dismissed):

target/scripts/build/packages.sh
```diff
@@ -82,12 +77,17 @@ function _install_packages() {
     opendmarc libsasl2-modules sasl2-bin
   )
 
+  DEBUG_PACKAGES=(
+    bind9-dnsutils ed iproute2 iputils-ping most
```
casperklein, line 81: I didn't know what `ed` is for. The internet told me, it's the oldest unix editor. I am wondering if it's just me, not knowing/using ed?

Here we could switch to something more modern, like `vim`.

polarathene: I actually looked into that a month or so ago, it was added a long time ago and isn't necessary.

`vim` isn't noob friendly IMO. If you want to include a text editor use `nano`?

casperklein: Noobs shouldn't edit files in a running container 😆

You might be right, I don't really care if `vim` or `nano` is included 👍

polarathene (approved):

reneploetz:
Just a general thought:

Is there a specific reason to put the rspamd.log inside the container instead of the central logging directory in `/var/log/mail` ?

I'm currently using the following patches to add logrotate + change the path to:

logging.inc
`filename = "/var/log/mail/rspamd.log";`

/etc/logrotate.d/rspamd
```/var/log/mail/rspamd.log
{
  compress
  copytruncate
  delaycompress
  rotate 4
  weekly
}

**Comment by polarathene** [comment]
> Is there a specific reason to put the rspamd.log inside the container instead of the central (usually volumed) logging directory in `/var/log/mail` ?

`supervisord` manages the services by running them in the foreground with stdout + stderr logs written to files in `/var/log/supervisor/${service}.log`. Often these logs are empty / minimal or repeated in the main mail log.

I imagine it's just been that way since supervisor was introduced. There has [been talk about refactoring how we manage logging](https://github.com/docker-mailserver/docker-mailserver/issues/3561), where Vector may be suitable. Ideally each service will keep it's logs separate and Vector would support configuration to output a log that combines those streams at whatever log level desired. It wouldn't be fixed to `/var/log/mail` then either.

**Comment by reneploetz** [comment]
Yeah, that sounds good.

My reasoning for the comment was somewhat along the lines of: If you write to `/var/log/supervisor` you will likely write inside the container as well as not performing any log rotation. This does not matter for small files, but might matter for larger one - especially on systems with many users/mail and/or very infrequent restarts.

If I look at the file sizes of the files in `/var/log/supervisor` on my system I can see that they are either empty or really small (with the exception of `saslauthd_ldap.log` for me). The `rspamd.log` also has a significant size.

For me it's just a bit odd that we use a somewhat centralized logging directory + logrotate for many things mail-related, but not here. But this is also the wrong ticket to discuss this anyway, so this is just a +1 for centralized logging.

**Comment by georglauterbach** [comment]
I'll come up with an update tomorrow

File: target/scripts/build/packages.sh (hunk 3 above)
Comment (georglauterbach) [reply target/scripts/build/packages.sh]: This will be resolved with another PR to have a separation of concerns here.

**Comment by georglauterbach** [comment]
I adjusted the PR so that the changes proposed by @reneploetz are incorporated. I will provide another PR that introduces updates to debug-packages later to have this PR only care about Rspamd and its logs.

The Rspamd logs are now managed by logrotate as well. Until Vector is used, I think this is the best way to move forward.

File: docs/content/config/security/rspamd.md [hunk 4]
```diff
@@ -69,6 +69,10 @@ DMS does not supply custom values for DNS servers to Rspamd. If you need to use
 
     This setting is enabled to not allow spam to proceed just because DNS requests did not succeed. It could deny legitimate e-mails to pass though too in case your DNS setup is incorrect or not functioning properly.
 
+### Logs
+
+You can find the Rspamd logs at `/var/log/mail/rspamd.log`, and the corresponding logs for [Redis](#persistence-with-redis), if it is enabled, at `/var/log/mail/rspamd-redis.log`. We recommend inspecting these logs (with `docker exec -ti <CONTAINER NAME> cat /var/log/mail/rspamd.log`) in case Rspamd does not work as expected.
```
reneploetz, line 74: As far as I can see: the `rspamd-redis.log` log file is still below `/var/log/supervisor/rspamd-redis.log` given the supervisor configuration.

reneploetz:
Thanks for the updated pull request!

docs/content/config/security/rspamd.md
georglauterbach: You're right, good catch!

georglauterbach: I resolved this :)

target/scripts/startup/setup.d/log.sh
```diff
@@ -13,31 +13,25 @@ function _setup_logs_general() {
...
-      ;;
-
-    ( * )
-      _log 'warn' 'LOGROTATE_INTERVAL not found in _setup_logrotate'
-      ;;
-
-  esac
-
-  echo -e "${LOGROTATE}}" >/etc/logrotate.d/maillog
+  if [[ ${LOGROTATE_INTERVAL} =~ ^(dai|week|month)ly$ ]]; then
```
polarathene, line 16: **Optional:** I just don't think the optimization here is worth it, better to be able to grep source files for these words and find this line than to skimp 2 chars via a regex?

```suggestion
  if [[ ${LOGROTATE_INTERVAL} =~ ^(daily|weekly|monthly)$ ]]; then
```

target/scripts/startup/setup.d/log.sh
```diff
@@ -13,31 +13,25 @@ function _setup_logs_general() {
...
+    _log 'trace' "Logrotate interval set to ${LOGROTATE_INTERVAL}"
+  else
+    _log 'warn' "Value '${LOGROTATE_INTERVAL}' for LOGROTATE_INTERVAL is invalid - falling back to default"
+    export LOGROTATE_INTERVAL='weekly'
+    # shellcheck disable=SC2034
+    VARS[LOGROTATE_INTERVAL]='weekly'
```
polarathene, line 22: The default is already set by this point?:

https://github.com/docker-mailserver/docker-mailserver/blob/82c38f242648290ab109de8e3994ea7564120deb/target/scripts/startup/variables-stack.sh#L138

https://github.com/docker-mailserver/docker-mailserver/blob/82c38f242648290ab109de8e3994ea7564120deb/target/scripts/start-mailserver.sh#L154

Recommending a [panic](https://github.com/docker-mailserver/docker-mailserver/blob/82c38f242648290ab109de8e3994ea7564120deb/target/scripts/helpers/error.sh#L21) for misconfiguration here during startup. We should clearly communicate to the user that they've explicitly set an incompatible ENV value for the setting and bail, just like we do with TLS certs during startup:

https://github.com/docker-mailserver/docker-mailserver/blob/82c38f242648290ab109de8e3994ea7564120deb/target/scripts/helpers/ssl.sh#L368

https://github.com/docker-mailserver/docker-mailserver/blob/82c38f242648290ab109de8e3994ea7564120deb/target/scripts/helpers/error.sh#L41-L47

---

**Mandatory:**

```suggestion
    _dms_panic__invalid_value 'LOGROTATE_INTERVAL' 'Setup -> Logrotate'
```

target/scripts/startup/setup.d/log.sh
```diff
@@ -13,31 +13,25 @@ function _setup_logs_general() {
...
+  cat >/etc/logrotate.d/maillog << EOF
+/var/log/mail/mail.log
+{
+  compress
+  copytruncate
+  delaycompress
+  rotate 4
+  ${LOGROTATE_INTERVAL}
+}
+EOF
```
polarathene, line 34: **No action required**

This is much more readable than the previous approach with the split `{` + `}` that looked almost like a typo 🚀 

Regarding [`rotate 4`](https://linux.die.net/man/8/logrotate), I guess that makes sense for `weekly`, while a bit odd for `daily` and `monthly`. I'm assuming `git blame` would show this was introduced with `weekly` in mind.

Given you've dropped the switch statement and no one seems to be bothered I guess leaving it at `4` here is fine, I was just a little curious what that setting did when I saw it here.

target/scripts/startup/setup.d/security/rspamd.sh
```diff
@@ -101,6 +102,19 @@ function __rspamd__run_early_setup_and_checks() {
...
+/var/log/mail/rspamd.log
+{
+  compress
+  copytruncate
+  delaycompress
+  rotate 4
+  ${LOGROTATE_INTERVAL}
+}
+EOF
+}
```
polarathene, line 116: **Optional:** While unlikely to introduce configuration drift, might be wise to add this comment for future maintenance?

```suggestion
# Keep in sync with `target/scripts/startup/setup.d/log.sh:_setup_logrotate()`
function __rspamd__setup_logfile() {
  cat >/etc/logrotate.d/rspamd << EOF
/var/log/mail/rspamd.log
{
  compress
  copytruncate
  delaycompress
  rotate 4
  ${LOGROTATE_INTERVAL}
}
EOF
}
```

test/helper/common.bash
```diff
@@ -429,8 +429,10 @@ function _filter_service_log() {
   local SERVICE=${1:?Service name must be provided}
   local STRING=${2:?String to match must be provided}
   local CONTAINER_NAME=$(__handle_container_name "${3:-}")
+  local FILE="/var/log/supervisor/${SERVICE}.log"
 
-  _run_in_container grep -E "${STRING}" "/var/log/supervisor/${SERVICE}.log"
+  [[ -f ${FILE} ]] || FILE="/var/log/mail/${SERVICE}.log"
```
polarathene, line 434: **Optional:** Took a moment for me to click what was going on 😅 

```suggestion
  # Fallback to alternative log location:
  [[ -f ${FILE} ]] || FILE="/var/log/mail/${SERVICE}.log"
```

target/scripts/start-mailserver.sh
```diff
@@ -82,8 +82,10 @@ function _register_functions() {
   _register_setup_function '_setup_opendmarc' # must come after `_setup_opendkim`
   _register_setup_function '_setup_policyd_spf'
 
+  _register_setup_function '_setup_logrotate'
+
   _register_setup_function '_setup_security_stack'
-  _register_setup_function '_setup_rspamd'
+  _register_setup_function '_setup_rspamd' # must come after _setup_logrotate
```
polarathene, line 88: I assume this is due to your addition of `export` reliance?

target/scripts/startup/setup.d/security/rspamd.sh
polarathene: At a glance beyond the `export` / `VARS` additions to `_setup_logrotate()` I don't see the dependency requirement?

If the ENV was misconfigured we'd bail with a panic now which is more correct. So there is no concern remaining about handling a misconfigured interval ENV reaching this point.

docs/content/config/security/rspamd.md
```diff
@@ -69,6 +69,10 @@ DMS does not supply custom values for DNS servers to Rspamd. If you need to use
 
     This setting is enabled to not allow spam to proceed just because DNS requests did not succeed. It could deny legitimate e-mails to pass though too in case your DNS setup is incorrect or not functioning properly.
 
+### Logs
+
+You can find the Rspamd logs at `/var/log/mail/rspamd.log`, and the corresponding logs for [Redis](#persistence-with-redis), if it is enabled, at `/var/log/supervisor/rspamd-redis.log`. We recommend inspecting these logs (with `docker exec -ti <CONTAINER NAME> cat /var/log/mail/rspamd.log`) in case Rspamd does not work as expected.
```
polarathene, line 74: **Optional:** `-it` is more consistent with other usage in the docs I think:

```suggestion
You can find the Rspamd logs at `/var/log/mail/rspamd.log`, and the corresponding logs for [Redis](#persistence-with-redis), if it is enabled, at `/var/log/supervisor/rspamd-redis.log`. We recommend inspecting these logs (with `docker exec -it <CONTAINER NAME> cat /var/log/mail/rspamd.log`) in case Rspamd does not work as expected.
```

polarathene (changes requested):

target/scripts/start-mailserver.sh
georglauterbach: Indeed, and it can be dropped now since we panic anyway.

polarathene (approved):

reneploetz (approved):

target/scripts/startup/setup.d/log.sh
casperklein: I stumbled upon the same while reviewing. We could introduce `$LOGROTATE_COUNT` or (not so readable) `$LOGROTATE_ROTATE` to make the rotation count configurable.

polarathene: It should be ok, no need to add new ENV to support this until we get actual requests to adjust the default.

PR #3578, opened 2023-10-12
chore: Add debug group (`packages.sh`) + more resilient rspamd setup

georglauterbach:
# Description

This PR does 2 things (and I was too lazy to provide two different PRs for these small changes):

1. Adjust `packages.sh` to install `most` and `nano` and remove `ed`, adding a dedicated set up packages for debugging (first commit)
2. Adjust the Rspamd setup script to make it even more robust (commit 2 and 4)

Follow up of #3576

## Type of change

## Checklist:

### DIFF
--- a/target/scripts/build/packages.sh
+++ b/target/scripts/build/packages.sh
@@ -43,10 +43,6 @@ function _install_postfix() {
 function _install_packages() {
   _log 'debug' 'Installing all packages now'
 
-  declare -a ANTI_VIRUS_SPAM_PACKAGES
-  declare -a CODECS_PACKAGES MISCELLANEOUS_PACKAGES
-  declare -a POSTFIX_PACKAGES MAIL_PROGRAMS_PACKAGES
-
   ANTI_VIRUS_SPAM_PACKAGES=(
     amavisd-new clamav clamav-daemon
     pyzor razor spamassassin
@@ -62,14 +58,13 @@ function _install_packages() {
   )
 
   MISCELLANEOUS_PACKAGES=(
-    apt-transport-https bind9-dnsutils binutils bsd-mailx
+    apt-transport-https binutils bsd-mailx
     ca-certificates curl dbconfig-no-thanks
-    dumb-init ed gnupg iproute2 iputils-ping
-    libdate-manip-perl libldap-common
-    libmail-spf-perl libnet-dns-perl
-    locales logwatch netcat-openbsd
-    nftables rsyslog supervisor
-    uuid whois
+    dumb-init gnupg libdate-manip-perl
+    libldap-common libmail-spf-perl
+    libnet-dns-perl locales logwatch
+    netcat-openbsd nftables rsyslog
+    supervisor uuid whois
   )
 
   POSTFIX_PACKAGES=(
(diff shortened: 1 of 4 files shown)


polarathene (dismissed):
LGTM 👍 

Mostly opinion below (_quoting specific commits_), I'm happy with the changes as-is.

---

> **add `readonly` to Rspamd setup script where appropriate**
> As a Rust-programmer, I expect immutability by default. Having it comes with nice properties. 
 In Bash, you have to explicitly make a variable immutable.
> This comment does exactly that for the Rspamd setup.

You probably meant `commit` there in the commit message? 😛 

While I agree immutability by default is nice, I'm not sure if I'd want to see `readonly` introduced throughout the codebase. We haven't hit any mutability issues yet right? Granted it's an easier workaround vs migrating from bash scripts to rust (_or [embeddable / script languages](https://github.com/rust-unofficial/awesome-rust#scripting) built off rust_).

To clarify, `readonly` after / during variable declaration is just locking it from accidental future modification?

---

> **refactor `packages.sh` and add proper text editor**
> I despise `nano`, but `(neo)vi(m)` has too many dependencies which would blow up image size.
> Moreover, `most` has been added, and I linked it to `less` because I am not aware of many users knowing `most` so linking it to less seemed appropriate to me.

Not sure about the symlink idea. It's with good intentions and probably harmless in the common scenario of just calling the binary by its name, but if trying to use some options to launch or navigate (eg: pattern search), this will probably be unexpected and a bit confusing initially to the user. Might result in bug report being raised too.

Personally, unless there's any notable advantage towards the UX for users with the command to troubleshoot within a container, it seems we'd be better off providing the familiar `less`. Alternatively a docs section that mentions the extra debugging packages we include can mention `most` there for discoverability and would fit right well on our debugging docs page.

I had a brief look at `less` vs `most` and I'm not entirely convinced that it's a better offering (_and since the SE discussion was from 2013, not exactly "new"_). Not blocking but [the suggestion to switch to `most`](https://github.com/docker-mailserver/docker-mailserver/pull/3576#issuecomment-1755720121) gave a different impression 😅 

- https://www.baeldung.com/linux/more-less-most-commands#comparison
- https://unix.stackexchange.com/questions/81129/what-are-the-differences-between-most-more-and-less/606397#606397

georglauterbach:
> To clarify, `readonly` after / during variable declaration is just locking it from accidental future modification?

Indeed:

```console
$ readonly --help
readonly: readonly [-aAf] [name[=value] ...] or readonly -p
    Mark shell variables as unchangeable.

    Mark each NAME as read-only; the values of these NAMEs may not be
    changed by subsequent assignment.  If VALUE is supplied, assign VALUE
    before marking as read-only.
```

---

I will switch to `less` now that I read through the websites you linked, because I concur that it is the better alternative as it well-known.

polarathene (approved):

casperklein (approved):

target/scripts/build/packages.sh
```diff
@@ -82,12 +77,17 @@ function _install_packages() {
     opendmarc libsasl2-modules sasl2-bin
   )
 
+  DEBUG_PACKAGES=(
+    bind9-dnsutils iproute2 iputils-ping less nano
+  )
```
polarathene, line 82: Is `iproute2` for the `ip` command?

Should we add some clarification here for `bind9-dnsutils` choice?:

```suggestion
  # `bind9-dnsutils` provides `dig` command
  # `iproute2` provides `ip` command
  DEBUG_PACKAGES=(
    bind9-dnsutils iproute2 iputils-ping less nano
  )
```

casperklein: > Is `iproute2` for the `ip` command?

Yes and it's mandatory and not just for debugging. We use the `ip` command in [networking.sh](https://github.com/docker-mailserver/docker-mailserver/blob/894978ddd751f0d09e9f68e1a3de13926e17d432/target/scripts/startup/setup.d/networking.sh#L16).

georglauterbach: Right - I will remove it from `DEBUG_PACKAGES` and add it back to the normal set.

georglauterbach: Done

casperklein (approved):

target/scripts/build/packages.sh
```diff
@@ -82,12 +77,19 @@ function _install_packages() {
     opendmarc libsasl2-modules sasl2-bin
   )
 
+  # `bind9-dnsutils` provides the `dig` command
+  # `iputils-ping` provides the `ping` command
```
polarathene, line 81: I figured that `ping` was a bit more self-explanatory, but not fussed either way 😛

polarathene (approved):

PR #3579, opened 2023-10-12
docs: Add Dovecot Lua auth guide + required package

### LINKED ISSUE #3575: feature request: Dovecot LUA support
### Context

Dovecot [supports Lua](https://doc.dovecot.org/admin_manual/lua/). One use case is [to extend its authentication options](https://doc.dovecot.org/configuration_manual/authentication/lua_based_authentication/#authentication-lua-based-authentication).

DMS currently does not include dovecot-lua. 

### Description

There is a dovecot-lua package available in Dovecot's deb repositories which are used to create the DMS image. Include this when building the image. The current version depends on Lua 5.3, which should be included as well. 

### Alternatives

Currently I patch Lua support into the docker-mailserver container upon initial startup (through user-patches.sh). I would like this to be included by default.

### Applicable Users

Any users who use an authentication backend that does not fit in any other category. One example from my use case is HTTP basic authentication against Nextcloud (DAV), over HTTPS of course.

I'll try to make a pull request for this later. It should not be too difficult considering that nothing new needs to be developed. It just needs to be enabled.

So:
- Enable Dovecot Lua support (include dovecot-lua and dependencies).
- Write some light documentation. I don't think it is necessary to create new environment variables to compose the container. Lua-based authentication is an advanced topic. To use it, it would be best [to rely on Dovecot configuration overrides](https://docker-mailserver.github.io/docker-mailserver/edge/config/advanced/override-defaults/dovecot/). I'll add a usage example in the documentation.

Zepmann:
# Description

This PR adds [Lua support for Dovecot](https://doc.dovecot.org/admin_manual/lua/) and documentation with an example usage for authentication. Lua authentication can be used to make Dovecot (and Postfix through Dovecot) authenticate against a web server or against anything else that can be used in Lua.

As discussed in #3575, a new `xapian` situation should be avoided. This PR relies on a standard Dovecot deb package that requires minimum change and maintenance as long as Dovecot deb packages are used. The documentation has a warning that there is minimal support for Lua, and that Lua support will be removed if it does not align with the development of DMS anymore for some reason.

Another warning in the documentation tells users that Lua authentication is an advanced topic, and that they should not open issues for questions about Lua scripting or bother DMS contributors about it.

Fixes #3575.

## Type of change

- [X] This change requires a documentation update - **The documentation is included.**

## Checklist:

### DIFF
--- a/docs/content/config/advanced/auth-lua.md
+++ b/docs/content/config/advanced/auth-lua.md
@@ -26,25 +26,23 @@ Each implementation of Lua-based authentication is custom. Therefore it is impos
 
 This scenario starts with [DMS being configured to use LDAP][docs::auth-ldap] for mailbox identification, user authorization and user authentication. In this scenario, [Nextcloud](https://nextcloud.com/) is also a service that uses the same LDAP server for user identification, authorization and authentication.
 
-The goal of this scenario is to have Dovecot not authenticate the user against LDAP, but against Nextcloud. Furthermore, the user should should only be able to authenticate using a randomly generated [Nextcloud application password](https://docs.nextcloud.com/server/latest/user_manual/en/session_management.html#managing-devices) and not the main password of the user account (stored in LDAP). The idea behind this is that a compromised mailbox password does not compromise the user's account entirely. To make this work, password reset through mail is disabled in Nextcloud.
+The goal of this scenario is to have Dovecot not authenticate the user against LDAP, but against Nextcloud using an [application password](https://docs.nextcloud.com/server/latest/user_manual/en/session_management.html#managing-devices). The idea behind this is that a compromised mailbox password does not compromise the user's account entirely. To make this work, Nextcloud is configured to [allow clients to only use application passwords for authentication](https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/config_sample_php_parameters.html#token-auth-enforced) and to [not allow password reset through mail](https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/config_sample_php_parameters.html#lost-password-link).
 
-If the application password is configured correctly, an adversary can only use it to access the user's mailbox (since it uses an application password that is compromised) and the user's CalDAV and CardDAV data on Nextcloud. File access through WebDAV can be disabled for the application password used to access mail. Having CalDAV and CardDAV compromised by the same password is a minor setback. If an adversary gets access to a Nextcloud application password through a device of the user, it is likely that the adversary also gets access to the user's calendars and contact lists anyway (locally or through the same account settings used by mail and Cal-/CardDAV synchronization). The user's stored files in Nextcloud, the LDAP account and any other services that rely on it would still be protected. A bonus is that a user is able to revoke and renew the mailbox password in Nextcloud for whatever reason, through a friendly user interface with all the security measures applies by the Nextcloud instance.
+If the application password is configured correctly, an adversary can only use it to access the user's mailbox, and the user's CalDAV and CardDAV data on Nextcloud. File access through WebDAV can be disabled for the application password used to access mail. Having CalDAV and CardDAV compromised by the same password is a minor setback. If an adversary gets access to a Nextcloud application password through a device of the user, it is likely that the adversary also gets access to the user's calendars and contact lists anyway (locally or through the same account settings used for mail and CalDAV/CardDAV synchronization). The user's stored files in Nextcloud, the LDAP account password and any other services that rely on it would still be protected. A bonus is that a user is able to revoke and renew the mailbox password in Nextcloud for whatever reason, through a friendly user interface with all the security measures with which the Nextcloud instance is configured.
 
-There is also a risk that a Nextcloud application password used for something else is compromised and is used to access the user's mailbox before it is revoked. Discussion of that risk falls outside of the scope of this scenario.
+A drawback of this method is that any (compromised) Nextcloud application password can be used to access the user's mailbox. This introduces a risk that a Nextcloud application password used for something else (e.g. WebDAV file access) is compromised and used to access the user's mailbox. Discussion of that risk falls outside of the scope of this scenario.
 
 To answer the questions asked earlier for this specific scenario:
 
 1. Do I want to use Lua to identify mailboxes and verify that users are are authorized to use mail services? **No. Provisioning is done through LDAP.**
-1. Do I want to use Lua to verify passwords that users authenticate with for IMAP/POP3/SMTP in their mail clients? **Yes. Password authentication is done through Lua.**
-1. If the answer is 'yes' to question 1 or 2: are there other methods that better facilitate my use case instead of custom scripts which rely on me being a developer and not just a user? **No. Only HTTP can be used to authenticate against Nextcloud, which is not supported by Dovecot or DMS.**
+1. Do I want to use Lua to verify passwords that users authenticate with for IMAP/POP3/SMTP in their mail clients? **Yes. Password authentication is done through Lua against Nextcloud.**
+1. If the answer is 'yes' to question 1 or 2: are there other methods that better facilitate my use case instead of custom scripts which rely on me being a developer and not just a user? **No. Only HTTP can be used to authenticate against Nextcloud, which is not supported natively by Dovecot or DMS.**
 
-While it is possible to extend what Nextcloud supports with [Nextcloud apps](https://apps.nextcloud.com/), there is currently a mismatch between what DMS supports and what Nextcloud applications support. This might change in the future. For now, Lua will be used to bridge the gap between DMS and Nextcloud for authentication only (Dovecot passdb), while LDAP will still be used to identify mailboxes and verify authorization (Dovecot userdb).
+While it is possible to extend the authentication methods which Nextcloud can facilitate with [Nextcloud apps](https://apps.nextcloud.com/), there is currently a mismatch between what DMS supports and what Nextcloud applications can provide. This might change in the future. For now, Lua will be used to bridge the gap between DMS and Nextcloud for authentication only (Dovecot passdb), while LDAP will still be used to identify mailboxes and verify authorization (Dovecot userdb).
 
 ## Modify Dovecot's configuration
 
-Add the following volume values to the relevant part of `compose.yaml`:
-
-???+ example
+???+ example "Add to DMS volumes in `compose.yaml`"
 
     ```yaml
         # All new volumes are marked :ro to configure them as read-only, since their contents are not changed from inside the container
@@ -57,25 +55,20 @@ Add the following volume values to the relevant part of `compose.yaml`:
           - ./docker-data/dms/config/dovecot/lua/:/etc/dovecot/lua/:ro
     ```
 
-Make the necessary changes on the filesystem:
+Create a directory for Lua scripts:
 ```bash
 mkdir -p ./docker-data/dms/config/dovecot/lua
-docker cp CONTAINER_NAME:/etc/dovecot/conf.d/auth-ldap.conf.ext ./docker-data/dms/config/dovecot/auth-ldap.conf.ext
 ```
 
-Edit configuration file `./docker-data/dms/config/dovecot/auth-ldap.conf.ext`. Comment out the passdb section. An excerpt of what that part would look like after you are done:
+Create configuration file `./docker-data/dms/config/dovecot/auth-ldap.conf.ext` for LDAP user provisioning:
 ```
-#passdb {
-#  driver = ldap
-#
-#  # Path for LDAP configuration file, see example-config/dovecot-ldap.conf.ext
-#  args = /etc/dovecot/dovecot-ldap.conf.ext
-#}
+userdb {
+  driver = ldap
+  args = /etc/dovecot/dovecot-ldap.conf.ext
+}
 ```
 
-Don't touch anything else in the file.
-
-Create configuration file `./docker-data/dms/config/dovecot/auth-lua-httpbasic.conf` with contents:
+Create configuration file `./docker-data/dms/config/dovecot/auth-lua-httpbasic.conf` for Lua user authentication:
 ```
 passdb {
   driver = lua
@@ -112,15 +105,7 @@ end
 function script_deinit()
 end
 
-function is_nextcloud_apppassword(password)
-  return string.find(password, "%w%w%w%w%w%-%w%w%w%w%w%-%w%w%w%w%w%-%w%w%w%w%w%-%w%w%w%w%w") ~= nil
-end
-
 function auth_passdb_lookup(req)
-  if not is_nextcloud_apppassword(req.password)
-  then
-    return dovecot.auth.PASSDB_RESULT_PASSWORD_MISMATCH, ""
-  end
   local auth_request = http_client:request {
     url = http_url;
     method = http_method;
@@ -149,14 +134,14 @@ end
 
 Replace the hostname in the URL to the actual hostname of Nextcloud.
 
-Dovecot [provides an HTTP client for use in Lua](https://doc.dovecot.org/admin_manual/lua/#dovecot.http.client). Aside of that, Lua by itself is pretty barebones. It chooses library compactness over included functionality. You can see that in the inefficiently typed regular expression in `is_nextcloud_apppassword()`, which is used because [Lua does not offer full support for regular expressions](https://www.lua.org/pil/20.2.html). You can also see Lua's limited functionality in that a separate library is referenced to add support for Base64 encoding, which is required for [HTTP basic access authentication](https://en.wikipedia.org/wiki/Basic_access_authentication). This library (also a Lua script) is not included. It must be downloaded and stored in the same directory:
+Dovecot [provides an HTTP client for use in Lua](https://doc.dovecot.org/admin_manual/lua/#dovecot.http.client). Aside of that, Lua by itself is pretty barebones. It chooses library compactness over included functionality. You can see that in that a separate library is referenced to add support for Base64 encoding, which is required for [HTTP basic access authentication](https://en.wikipedia.org/wiki/Basic_access_authentication). This library (also a Lua script) is not included. It must be downloaded and stored in the same directory:
 
 ```bash
 cd ./docker-data/dms/config/dovecot/lua
 curl -JLO https://raw.githubusercontent.com/iskolbin/lbase64/master/base64.lua
 ```
 
-Only use native (pure Lua) libraries as dependencies, such as `base64.lua` from the example. This ensures maximum compatibility. Performance is less of an issue since Lua scripts written for Dovecot probably won't be long or complex, and there won't be a lot of data processing by Lua itself.
+Only use native (pure Lua) libraries as dependencies if possible, such as `base64.lua` from the example. This ensures maximum compatibility. Performance is less of an issue since Lua scripts written for Dovecot probably won't be long or complex, and there won't be a lot of data processing by Lua itself.
 
 ## Debugging a Lua script
 
@@ -166,9 +151,7 @@ To see which Lua version is used by Dovecot if you plan to do something that is 
 docker exec CONTAINER_NAME strings /usr/lib/dovecot/libdovecot-lua.so|grep '^LUA_'
 ```
 
-While Dovecot logs the status of authentication attempts for any passdb backend, Dovecot will also log Lua scripting errors and messages sent to Dovecot's [Lua API log functions](https://doc.dovecot.org/admin_manual/lua/#dovecot.i_debug).
-
-The combined DMS log (including that of Dovecot) can be viewed using `docker logs CONTAINER_NAME`. If the log is too noisy (_due to other processes in the container also logging to it_), `docker exec CONTAINER_NAME cat /var/log/mail/mail.log` can be used to view the log of Dovecot and Postfix specifically.
+While Dovecot logs the status of authentication attempts for any passdb backend, Dovecot will also log Lua scripting errors and messages sent to Dovecot's [Lua API log functions](https://doc.dovecot.org/admin_manual/lua/#dovecot.i_debug). The combined DMS log (including that of Dovecot) can be viewed using `docker logs CONTAINER_NAME`. If the log is too noisy (_due to other processes in the container also logging to it_), `docker exec CONTAINER_NAME cat /var/log/mail/mail.log` can be used to view the log of Dovecot and Postfix specifically.
 
 If working with HTTP in Lua, setting `debug = true;` when initiating `dovecot.http.client` will create debug log messages for every HTTP request and response.
 
--- a/docs/content/config/advanced/auth-lua.md
+++ b/docs/content/config/advanced/auth-lua.md
@@ -6,23 +6,16 @@ title: 'Advanced | Lua Authentication'
 
 Dovecot has the ability to let users create their own custom user provisioning and authentication providers in [Lua](https://en.wikipedia.org/wiki/Lua_(programming_language)#Syntax). This allows any data source that can be approached from Lua to be used for authentication, including web servers. It is possible to do more with Dovecot and Lua, but other use cases fall outside of the scope of this documentation page.
 
-!!! warning
+!!! warning "Community contributed guide"
+    Dovecot authentication via Lua scripting is not officially supported in DMS. No assistance will be provided should you encounter any issues.
+    
+    DMS provides the required packages to support this guide. Note that these packages will be removed should they introduce any future maintenance burden.
 
-    DMS offers minimal support for Lua-based authentication due to it being an advanced method that can be used in many different ways. Do not open issues on GitHub or request support from DMS contributors for Lua scripts. The documentation on this page is all that is needed to get your own development started. Look elsewhere for Lua support.
+    The example in this guide relies on the current way in which DMS works with Dovecot configuration files. Changes to this to accommodate new authentication methods such as OpenID Connect will likely break this example in the future. This guide is updated on a best-effort base.
 
-!!! warning
+Dovecot's Lua support can be used for user provisioning (userdb functionality) and/or password verification (passdb functionality). Consider using other userdb and passdb options before considering Lua, since Lua does require the use of additional (unsupported) program code that might require maintenance when updating DMS.
 
-    Lua-based authentication relies on a Dovecot plugin. Dovecot is known to sometimes deprecate and remove support for their plugins (such as [CheckPassword](https://doc.dovecot.org/configuration_manual/authentication/checkpassword/)). DMS will drop support immediately if at some moment continued inclusion of the Lua plugin would not align with the DMS development process anymore.
-
-There are several questions you should ask yourself before you start:
-
-1. Do I want to use Lua to identify mailboxes and verify that users are are authorized to use mail services? This refers in the world of Dovecot to Lua providing 'userdb' functionality, as in a data source for user provisioning.
-1. Do I want to use Lua to verify passwords that users authenticate with for IMAP/POP3/SMTP in their (web) mail clients? This refers in the world of Dovecot to Lua providing 'passdb' functionality, as in a data source for user password verification.
-1. If the answer is 'yes' to question 1 or 2: are there other methods that better facilitate my use case aside of custom scripts which rely on me being a developer and not just a user?
-
-If the answer is 'no' to question 3, Lua-based authentication might just be the thing for you.
-
-Each implementation of Lua-based authentication is fully custom. Therefore it is impossible to write documentation that covers every scenario. Instead, this page describes a single scenario. If that scenario is followed, you will learn vital aspects that are necessary to kickstart your own Lua development:
+Each implementation of Lua-based authentication is custom. Therefore it is impossible to write documentation that covers every scenario. Instead, this page describes a single example scenario. If that scenario is followed, you will learn vital aspects that are necessary to kickstart your own Lua development:
 
 - How to override Dovecot's default configuration to disable parts that conflict with your scenario.
 - How to make Dovecot use your Lua script.
@@ -31,7 +24,7 @@ Each implementation of Lua-based authentication is fully custom. Therefore it is
 
 ## The example scenario
 
-This scenario starts with [DMS being configured to use LDAP][docs-authldap] for mailbox identification, user authorization and user authentication. In this scenario, [Nextcloud](https://nextcloud.com/) is also a service that uses the same LDAP server for user identification, authorization and authentication.
+This scenario starts with [DMS being configured to use LDAP][docs::auth-ldap] for mailbox identification, user authorization and user authentication. In this scenario, [Nextcloud](https://nextcloud.com/) is also a service that uses the same LDAP server for user identification, authorization and authentication.
 
 The goal of this scenario is to have Dovecot not authenticate the user against LDAP, but against Nextcloud. Furthermore, the user should should only be able to authenticate using a randomly generated [Nextcloud application password](https://docs.nextcloud.com/server/latest/user_manual/en/session_management.html#managing-devices) and not the main password of the user account (stored in LDAP). The idea behind this is that a compromised mailbox password does not compromise the user's account entirely. To make this work, password reset through mail is disabled in Nextcloud.
 
@@ -47,43 +40,27 @@ To answer the questions asked earlier for this specific scenario:
 
 While it is possible to extend what Nextcloud supports with [Nextcloud apps](https://apps.nextcloud.com/), there is currently a mismatch between what DMS supports and what Nextcloud applications support. This might change in the future. For now, Lua will be used to bridge the gap between DMS and Nextcloud for authentication only (Dovecot passdb), while LDAP will still be used to identify mailboxes and verify authorization (Dovecot userdb).
 
-## Container variables to adjust
-
-Since Docker Mailserver provides minimal support for Lua with Dovecot, there are no environment variables to configure specifically for Lua. Some environment variables that do exist for other aspects of DMS might need to be configured based on the scenario you want to follow.
-
-In the case of the example scenario, the environment variables must be configured as if users will be authenticated against LDAP (to support identification of mailboxes and verifying authorizations through LDAP). See [LDAP Authentication][docs-authldap] for more information. In addition, the following variables are required:
-
-???+ example
-
-    DMS configuration environment variables to let Postfix apply Lua authentication by proxy through Dovecot for authenticated SMTP.
-
-    ```yaml
-    - ENABLE_SASLAUTHD=1
-    - SASLAUTHD_MECHANISMS=rimap
-    - SASLAUTHD_MECH_OPTIONS=127.0.0.1
-    ```
-
-## Modifying Dovecot's configuration
+## Modify Dovecot's configuration
 
 Add the following volume values to the relevant part of `compose.yaml`:
 
 ???+ example
 
-    Override and add Dovecot configuration files and lua scripts.
-
     ```yaml
+        # All new volumes are marked :ro to configure them as read-only, since their contents are not changed from inside the container
         volumes:
+          # Configuration override to disable LDAP authentication
           - ./docker-data/dms/config/dovecot/auth-ldap.conf.ext:/etc/dovecot/conf.d/auth-ldap.conf.ext:ro
+          # Configuration addition to enable Lua authentication
           - ./docker-data/dms/config/dovecot/auth-lua-httpbasic.conf:/etc/dovecot/conf.d/auth-lua-httpbasic.conf:ro
+          # Directory containing Lua scripts
           - ./docker-data/dms/config/dovecot/lua/:/etc/dovecot/lua/:ro
     ```
 
-The first volume line [overrides][docs-dovecotoverrideconfiguration] Dovecot's standard LDAP authentication configuration file. The second line [adds][docs-dovecotaddconfiguration] a new configuration file for Lua authentication. The third line adds a directory which will contain Lua scripts. The files and directory will not be changed from inside the container, which is why they are configured as read-only.
-
-Make the necessary changes on the filesystem (*where `mailserver` is the container name*):
+Make the necessary changes on the filesystem:
 ```bash
 mkdir -p ./docker-data/dms/config/dovecot/lua
-docker cp mailserver:/etc/dovecot/conf.d/auth-ldap.conf.ext ./docker-data/dms/config/dovecot/auth-ldap.conf.ext
+docker cp CONTAINER_NAME:/etc/dovecot/conf.d/auth-ldap.conf.ext ./docker-data/dms/config/dovecot/auth-ldap.conf.ext
 ```
 
 Edit configuration file `./docker-data/dms/config/dovecot/auth-ldap.conf.ext`. Comment out the passdb section. An excerpt of what that part would look like after you are done:
@@ -120,12 +97,12 @@ local http_status_failure = 401
 local http_header_forwarded_for = "X-Forwarded-For"
 
 package.path = package.path .. ";/etc/dovecot/lua/?.lua"
-local base64 = require("base64")
+require("base64")
 
 local http_client = dovecot.http.client {
-    timeout = 1000;
-    max_attempts = 1;
-    debug = false;
+  timeout = 1000;
+  max_attempts = 1;
+  debug = false;
 }
 
 function script_init()
@@ -144,27 +121,30 @@ function auth_passdb_lookup(req)
   then
     return dovecot.auth.PASSDB_RESULT_PASSWORD_MISMATCH, ""
   end
-
   local auth_request = http_client:request {
     url = http_url;
     method = http_method;
   }
-  auth_request:add_header("Authorization", "Basic " .. base64.encode(req.user .. ":" .. req.password))
+  local base64 = require("base64")
+  auth_request:add_header("Authorization", "Basic " .. (base64.encode(req.user .. ":" .. req.password)))
   auth_request:add_header(http_header_forwarded_for, req.remote_ip)
   local auth_response = auth_request:submit()
+  local resp_status = auth_response:status()
+  local reason = auth_response:reason()
 
   local returnStatus = dovecot.auth.PASSDB_RESULT_INTERNAL_FAILURE
-  local returnDesc = http_method .. " - " .. http_url .. " - " .. auth_response:status() .. " " .. auth_response:reason()
-  if auth_response:status() == http_status_ok
+  local returnDesc = http_method .. " - " .. http_url .. " - " .. resp_status .. " " .. reason
+  if resp_status == http_status_ok
   then
     returnStatus = dovecot.auth.PASSDB_RESULT_OK
     returnDesc = "nopassword=y"
-  elseif auth_response:status() == http_status_failure
+  elseif resp_status == http_status_failure
   then
     returnStatus = dovecot.auth.PASSDB_RESULT_PASSWORD_MISMATCH
     returnDesc = ""
   end
   return returnStatus, returnDesc
+end
 ```
 
 Replace the hostname in the URL to the actual hostname of Nextcloud.
@@ -176,19 +156,25 @@ cd ./docker-data/dms/config/dovecot/lua
 curl -JLO https://raw.githubusercontent.com/iskolbin/lbase64/master/base64.lua
 ```
 
-Only use native (pure Lua) libraries as dependencies, such as `base64.lua` from the example. This ensures maximum compatibility. Performance is less of an issue since Lua scripts written for Dovecot probably won't be long or complex, and there won't be a lot of data processing by Lua itself. To see which Lua version is used by Dovecot if you plan to do something that is version dependent, run:
-
-```bash
-docker exec mailserver strings /usr/lib/dovecot/libdovecot-lua.so|grep '^LUA_'
-```
+Only use native (pure Lua) libraries as dependencies, such as `base64.lua` from the example. This ensures maximum compatibility. Performance is less of an issue since Lua scripts written for Dovecot probably won't be long or complex, and there won't be a lot of data processing by Lua itself.
 
 ## Debugging a Lua script
 
-Aside of succeeded and failed authentication attempts for any passdb backend, Dovecot also logs Lua scripting errors and messages send to Dovecot's [Lua API log functions](https://doc.dovecot.org/admin_manual/lua/#dovecot.i_debug). The combined DMS log (including that of Dovecot) can be viewed using `docker logs mailserver` (*where `mailserver` is the container name*). If the log is too noisy (due to other processes in the container also logging to it), `docker exec mailserver cat /var/log/mail/mail.log` can be used to view the log of Dovecot and Postfix specifically. If working with HTTP in Lua, setting `debug = true;` when initiating `dovecot.http.client` will create debug log messages for every HTTP request and response.
+To see which Lua version is used by Dovecot if you plan to do something that is version dependent, run:
 
-Note that Lua runs compiled bytecode, and that scripts will be compiled when they are initially started. Once compiled, the bytecode is cached and changes in the script will not be processed. [Restart sub-service Dovecot][docs-faqalterdms] using `docker exec mailserver supervisorctl restart dovecot` to have Dovecot load a changed Lua script.
+```bash
+docker exec CONTAINER_NAME strings /usr/lib/dovecot/libdovecot-lua.so|grep '^LUA_'
+```
 
-[docs-authldap]: ./auth-ldap.md
-[docs-dovecotoverrideconfiguration]: ./override-defaults/dovecot.md#override-configuration
-[docs-dovecotaddconfiguration]: ./override-defaults/dovecot.md#add-configuration
-[docs-faqalterdms]: ../../faq.md#how-to-alter-a-running-dms-instance-without-relaunching-the-container
+While Dovecot logs the status of authentication attempts for any passdb backend, Dovecot will also log Lua scripting errors and messages sent to Dovecot's [Lua API log functions](https://doc.dovecot.org/admin_manual/lua/#dovecot.i_debug).
+
+The combined DMS log (including that of Dovecot) can be viewed using `docker logs CONTAINER_NAME`. If the log is too noisy (_due to other processes in the container also logging to it_), `docker exec CONTAINER_NAME cat /var/log/mail/mail.log` can be used to view the log of Dovecot and Postfix specifically.
+
+If working with HTTP in Lua, setting `debug = true;` when initiating `dovecot.http.client` will create debug log messages for every HTTP request and response.
+
+Note that Lua runs compiled bytecode, and that scripts will be compiled when they are initially started. Once compiled, the bytecode is cached and changes in the Lua script will not be processed automatically. Dovecot will reload its configuration and clear its cached Lua bytecode when running `docker exec CONTAINER_NAME dovecot reload`. A (changed) Lua script will be compiled to bytecode the next time it is executed after running the Dovecot reload command.
+
+[docs::auth-ldap]: ./auth-ldap.md
+[docs::dovecot-override-configuration]: ./override-defaults/dovecot.md#override-configuration
+[docs::dovecot-add-configuration]: ./override-defaults/dovecot.md#add-configuration
+[docs::faq-alter-running-dms-instance-without-container-relaunch]: ../../faq.md#how-to-alter-a-running-dms-instance-without-relaunching-the-container
--- a/docs/content/config/advanced/auth-lua.md
+++ b/docs/content/config/advanced/auth-lua.md
@@ -1,0 +1,197 @@
+---
+title: 'Advanced | Lua Authentication'
+---
+
+## Introduction
+
+Dovecot has the ability to let users create their own custom user provisioning and authentication providers in [Lua](https://en.wikipedia.org/wiki/Lua_(programming_language)#Syntax). This allows any data source that can be approached from Lua to be used for authentication, including web servers. It is possible to do more with Dovecot and Lua, but other use cases fall outside of the scope of this documentation page.
+
+!!! warning
+
+    DMS offers minimal support for Lua-based authentication due to it being an advanced method that can be used in many different ways. Do not open issues on GitHub or request support from DMS contributors for Lua scripts. The documentation on this page is all that is needed to get your own development started. Look elsewhere for Lua support.
+
+!!! warning
+
+    Lua-based authentication relies on a Dovecot plugin. Dovecot is known to sometimes deprecate and remove support for their plugins (such as [CheckPassword](https://doc.dovecot.org/configuration_manual/authentication/checkpassword/)). DMS will drop support immediately if at some moment continued inclusion of the Lua plugin would not align with the DMS development process anymore.
+
+There are several questions you should ask yourself before you start:
+
+1. Do I want to use Lua to identify mailboxes and verify that users are are authorized to use mail services? This refers in the world of Dovecot to Lua providing 'userdb' functionality, as in a data source for user provisioning.
+1. Do I want to use Lua to verify passwords that users authenticate with for IMAP/POP3/SMTP in their (web) mail clients? This refers in the world of Dovecot to Lua providing 'passdb' functionality, as in a data source for user password verification.
+1. If the answer is 'yes' to question 1 or 2: are there other methods that better facilitate my use case aside of custom scripts which rely on me being a developer and not just a user?
+
+If the answer is 'no' to question 3, Lua-based authentication might just be the thing for you.
+
+Each implementation of Lua-based authentication is fully custom. Therefore it is impossible to write documentation that covers every scenario. Instead, this page describes a single scenario. If that scenario is followed, you will learn vital aspects that are necessary to kickstart your own Lua development:
+
+- How to override Dovecot's default configuration to disable parts that conflict with your scenario.
+- How to make Dovecot use your Lua script.
+- How to add your own Lua script and any libraries it uses.
+- How to debug your Lua script.
+
+## The example scenario
+
+This scenario starts with [DMS being configured to use LDAP][docs-authldap] for mailbox identification, user authorization and user authentication. In this scenario, [Nextcloud](https://nextcloud.com/) is also a service that uses the same LDAP server for user identification, authorization and authentication.
+
+The goal of this scenario is to have Dovecot not authenticate the user against LDAP, but against Nextcloud. Furthermore, the user should should only be able to authenticate using a randomly generated [Nextcloud application password](https://docs.nextcloud.com/server/latest/user_manual/en/session_management.html#managing-devices) and not the main password of the user account (stored in LDAP). The idea behind this is that a compromised mailbox password does not compromise the user's account entirely. To make this work, password reset through mail is disabled in Nextcloud.
+
+If the application password is configured correctly, an adversary can only use it to access the user's mailbox (since it uses an application password that is compromised) and the user's CalDAV and CardDAV data on Nextcloud. File access through WebDAV can be disabled for the application password used to access mail. Having CalDAV and CardDAV compromised by the same password is a minor setback. If an adversary gets access to a Nextcloud application password through a device of the user, it is likely that the adversary also gets access to the user's calendars and contact lists anyway (locally or through the same account settings used by mail and Cal-/CardDAV synchronization). The user's stored files in Nextcloud, the LDAP account and any other services that rely on it would still be protected. A bonus is that a user is able to revoke and renew the mailbox password in Nextcloud for whatever reason, through a friendly user interface with all the security measures applies by the Nextcloud instance.
+
+There is also a risk that a Nextcloud application password used for something else is compromised and is used to access the user's mailbox before it is revoked. Discussion of that risk falls outside of the scope of this scenario.
+
+To answer the questions asked earlier for this specific scenario:
+
+1. Do I want to use Lua to identify mailboxes and verify that users are are authorized to use mail services? **No. Provisioning is done through LDAP.**
+1. Do I want to use Lua to verify passwords that users authenticate with for IMAP/POP3/SMTP in their mail clients? **Yes. Password authentication is done through Lua.**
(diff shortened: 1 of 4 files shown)


Zepmann:
I am informed that this makes the ARM64 build fail. I'll look into it.

docs/content/config/advanced/auth-lua.md
```diff
@@ -0,0 +1,194 @@
...
+
+Don't touch anything else in the file.
+
+Create configuration file `./docker-data/dms/config/dovecot/auth-lua-httpbasic.conf` with contents:
+```
+passdb {
+  driver = lua
+  args = file=/etc/dovecot/lua/auth-httpbasic.lua blocking=yes
+}
+```
```
polarathene, line 107: Likewise, this may break for you in future when you're relying on DMS being configured for a specific `ACCOUNT_PROVISIONER` which makes certain assumptions. This is probably another caveat that could be mentioned in the disclaimer at the start of this guide, since breakage can work both ways.

When the OIDC feature is closer to being merged, we may be able to support an BYO config, but I think that'd still run into compatibility issues with some features of DMS (_eg: Dovecot sieve scripts, v13 recently received a fix to adjust the home directory within an accounts mailbox, LDAP doesn't have that and some users may explicitly configure their own location which won't be compatible_). Semver releases will hopefully at least prevent the likelihood of breakage for you between major releases.

---

BTW this section for Dovecot seems it's missing enabling the config? The LDAP config will be included, but you'll likely need a line similar to this:

https://github.com/docker-mailserver/docker-mailserver/blob/82c38f242648290ab109de8e3994ea7564120deb/target/dovecot/10-auth.conf#L125

I don't recommend volume mounting a replacement config for `10-auth.conf`, so `user-patches.sh` is probably beneficial for you. Keep in mind that if you just append the needed line, without a conditional to prevent adding it again this will pile up across container restarts (_unless the user follows our advice to always re-create the container, but many seem to miss this_), as internal state is persisted within the container until it's removed.

Eventually, we'll probably provide our customizations from a separate location, and always copy them over during container startup to minimize the issue so that config modifications are more deterministic.

docs/content/config/advanced/auth-lua.md
```diff
@@ -0,0 +1,194 @@
...
+local http_header_forwarded_for = "X-Forwarded-For"
+
+package.path = package.path .. ";/etc/dovecot/lua/?.lua"
+local base64 = require("base64")
+
+local http_client = dovecot.http.client {
+    timeout = 1000;
+    max_attempts = 1;
+    debug = false;
+}
```
polarathene, line 129: Indent is 4 spaces here instead of 2 like the rest of the lua file? Any reason for that?

docs/content/config/advanced/auth-lua.md
```diff
@@ -0,0 +1,194 @@
...
+function script_init()
+  return 0
+end
+
+function script_deinit()
+end
+
+function is_nextcloud_apppassword(password)
+  return string.find(password, "%w%w%w%w%w%-%w%w%w%w%w%-%w%w%w%w%w%-%w%w%w%w%w%-%w%w%w%w%w") ~= nil
+end
```
polarathene, line 140: **Optional**

Using `+` (_one or more_) and [makes the pattern a little easier to grok](https://gitspartv.github.io/lua-patterns/?pattern=%25w%2B-%25w%2B-%25w%2B-%25w%2B-%25w%2B). Although it's less specific (_[Lua pattern matching](https://riptutorial.com/lua/topic/5829/pattern-matching) doesn't seem to support repetition length like regex can_):

```suggestion
function is_nextcloud_apppassword(password)
  return string.find(password, "%w+-%w+-%w+-%w+-%w+") ~= nil
end
```

docs/content/config/advanced/auth-lua.md
```diff
@@ -0,0 +1,194 @@
...
+```bash
+cd ./docker-data/dms/config/dovecot/lua
+curl -JLO https://raw.githubusercontent.com/iskolbin/lbase64/master/base64.lua
+```
+
+Only use native (pure Lua) libraries as dependencies, such as `base64.lua` from the example. This ensures maximum compatibility. Performance is less of an issue since Lua scripts written for Dovecot probably won't be long or complex, and there won't be a lot of data processing by Lua itself. To see which Lua version is used by Dovecot if you plan to do something that is version dependent, run:
+
+```bash
+docker exec mailserver strings /usr/lib/dovecot/libdovecot-lua.so|grep '^LUA_'
+```
```
polarathene, line 183: Shift into Debugging section?

docs/content/config/advanced/auth-lua.md
```diff
@@ -0,0 +1,194 @@
...
+
+## Debugging a Lua script
+
+Aside of succeeded and failed authentication attempts for any passdb backend, Dovecot also logs Lua scripting errors and messages send to Dovecot's [Lua API log functions](https://doc.dovecot.org/admin_manual/lua/#dovecot.i_debug). The combined DMS log (including that of Dovecot) can be viewed using `docker logs mailserver` (*where `mailserver` is the container name*). If the log is too noisy (due to other processes in the container also logging to it), `docker exec mailserver cat /var/log/mail/mail.log` can be used to view the log of Dovecot and Postfix specifically. If working with HTTP in Lua, setting `debug = true;` when initiating `dovecot.http.client` will create debug log messages for every HTTP request and response.
```
polarathene, line 187:
```suggestion
While Dovecot logs the status of authentication attempts for any passdb backend, Dovecot will also log Lua scripting errors and messages sent to Dovecot's [Lua API log functions](https://doc.dovecot.org/admin_manual/lua/#dovecot.i_debug).

The combined DMS log (including that of Dovecot) can be viewed using `docker logs CONTAINER_NAME`. If the log is too noisy (_due to other processes in the container also logging to it_), `docker exec CONTAINER_NAME cat /var/log/mail/mail.log` can be used to view the log of Dovecot and Postfix specifically.

If working with HTTP in Lua, setting `debug = true;` when initiating `dovecot.http.client` will create debug log messages for every HTTP request and response.
```

The 2nd paragraph has a bit of [overlap with our dedicated debugging page](https://docker-mailserver.github.io/docker-mailserver/v12.1/config/debugging/#debug-a-running-container) on getting logs.

docs/content/config/advanced/auth-lua.md
```diff
@@ -0,0 +1,194 @@
...
+
+Note that Lua runs compiled bytecode, and that scripts will be compiled when they are initially started. Once compiled, the bytecode is cached and changes in the script will not be processed. [Restart sub-service Dovecot][docs-faqalterdms] using `docker exec mailserver supervisorctl restart dovecot` to have Dovecot load a changed Lua script.
```
polarathene, line 189: Could you confirm that restarting the Dovecot service completely is required? Or is it sufficient to run `dovecot reload`? (_which is what [`check-for-changes.sh` does](https://github.com/docker-mailserver/docker-mailserver/blob/82c38f242648290ab109de8e3994ea7564120deb/target/scripts/check-for-changes.sh#L57)_)

docs/content/config/advanced/auth-lua.md
```diff
@@ -0,0 +1,194 @@
...
+
+[docs-authldap]: ./auth-ldap.md
+[docs-dovecotoverrideconfiguration]: ./override-defaults/dovecot.md#override-configuration
+[docs-dovecotaddconfiguration]: ./override-defaults/dovecot.md#add-configuration
+[docs-faqalterdms]: ../../faq.md#how-to-alter-a-running-dms-instance-without-relaunching-the-container
```
polarathene, line 194: Convention for ref links is to categorize/group with the `::` delimiter, and use kebab-case for better readability during maintenance.

eg: `docs::faq-alter-dms` / `docs::faq::alter-dms`

Main gripe is the lack of a delimiter between words.

target/scripts/build/packages.sh
```diff
@@ -97,6 +97,7 @@ function _install_dovecot() {
     dovecot-core dovecot-imapd
     dovecot-ldap dovecot-lmtpd dovecot-managesieved
     dovecot-pop3d dovecot-sieve dovecot-solr
+    dovecot-lua
```
polarathene, line 100: Can you add a comment above `DOVECOT_PACKAGES=` to note that `dovecot-lua` is a community contributed package?

It helps to keep this distinction for maintenance. We'll need to do that for some prior community contributed packages too (_you don't need to worry about that though_).

polarathene (changes requested):
First review pass.

That was a much larger docs contribution than I expected! I appreciate it though 😁 

I'll provide some more revision after the current feedback is addressed. Great work so far!

polarathene:
> I am informed that this makes the ARM64 build fail. I'll look into it.

And I probably should have checked the test suite before diving into a review 😰 (_I think it took me roughly 2 hours_)

Here's the ARM64 failure:

> `Unable to locate package dovecot-lua`

This is one of those maintenance concerns I raised initially. The Debian package ~~doesn't have an ARM64 variant available~~. Actually it seems the package name is wrong, where did you get `dovecot-lua` from? It seems to be [`dovecot-auth-lua`](https://packages.debian.org/bullseye/dovecot-auth-lua) (_has full platform support_)?

The other test failure with Fail2Ban is unrelated, that's been acting up lately for unknown reasons.

---

I'm a bit surprised though, as for some reason our AMD64 build was successful. It should not have used a cache due to `packages.sh` being updated...

Zepmann:
Not  going into all of your extensive review points of the documentation at the moment. They are good points, and I will address them :smile:

> > I am informed that this makes the ARM64 build fail. I'll look into it.
> 
> And I probably should have checked the test suite before diving into a review 😰 (_I think it took me roughly 2 hours_)
> 
> Here's the ARM64 failure:
> 
> > `Unable to locate package dovecot-lua`
> 
> This is one of those maintenance concerns I raised initially. The Debian package ~doesn't have an ARM64 variant available~. Actually it seems the package name is wrong, where did you get `dovecot-lua` from? It seems to be [`dovecot-auth-lua`](https://packages.debian.org/bullseye/dovecot-auth-lua) (_has full platform support_)?

`tools/scripts/build/packages.sh` imports the [https://repo.dovecot.org/ce-2.3-latest/debian/bullseye](Dovecot deb community repo). This repository provides only x86_64 packages, and it uses package name `dovecot-lua` for Dovecot's Lua support (`dovecot-auth-lua` is just a dummy package in this repository).

If building on ARM64, the Debian repositories are used since Dovecot's community repository is unable to provide ARM64 packages. Debian repositories use package name `dovecot-auth-lua` for Dovecot's Lua support.

I will change packages.sh so that Dovecot's community repository is not used when not compiling for x86_64, and I will make it use the right package name for each situation.

> 
> The other test failure with Fail2Ban is unrelated, that's been acting up lately for unknown reasons.
> 

Thanks for the notification. I'll ignore these test failures if they pop up for me.

polarathene:
> `tools/scripts/build/packages.sh` imports the [Dovecot deb community repo](https://repo.dovecot.org/ce-2.3-latest/debian/bullseye). This repository provides only x86_64 packages, and it uses package name `dovecot-lua` for Dovecot's Lua support

Ah thanks for that clarification.

<details>
<summary>Resolved</summary>

I'm not fond of the community repo myself since that's been causing various issues, especially with preventing upgrading to Debian 12.
- One maintainer wants to push towards Alpine as a new base image, but I'm reluctant towards that from my own experience of many issues encountered with that.
- I'm more inclined towards Fedora as the base image going forward, but it lacks a reliable package repo for rspamd last I checked.

The only reason we use the community repo for Dovecot AFAIK, was due to needing a version newer than Debian 11 provided. While another maintainer is reluctant to use the Debian 12 Dovecot release as the version is behind the community repo, but Debian 12 won't be supported by the community repo until Dovecot 2.4/3.0 is released AFAIK.

Bit of a predicament 😅 

---

@casperklein @georglauterbach how do you want to resolve this?

- Defer until Debian 12 upgade.
- Defer until base image switch
- Conditionally build for AMD64

</details>

> I will change `packages.sh` so that Dovecot's community repository is not used when not compiling for x86_64, and I will make it use the right Lua package name for each situation.

I have a bad habit of replying too eagerly and didn't read this far 😑 Well that's sorted then 👍

Zepmann:
> The only reason we use the community repo for Dovecot AFAIK, was due to needing a version newer than Debian 11 provided. While another maintainer is reluctant to use the Debian 12 Dovecot release as the version is behind the community repo, but Debian 12 won't be supported by the community repo until Dovecot 2.4/3.0 is released AFAIK.

I read this as that there is a Dovecot version difference between DMS on x86_64 and ARM64. After some testing, it turns out this is true.

x86_64: Dovecot 2.3.21 (from the community repo)
ARM64: Dovecot 2.3.13 (from Debian's repo)

I do agree that this is quite a conundrum if stability is preferred and you wish to avoid community repos. Still, if you consider [the rspamd community repository for Fedora](https://rspamd.com/downloads.html), they **do** [provide ARM64-packages in their rpm repositores](https://rspamd.com/rpm/centos-7/), [unlike Dovecot](https://repo.dovecot.org/ce-2.3.17/centos/7/RPMS/). That would at least take care of the software version discrepancy between architectures, and any headaches that might cause.

> I have a bad habit of replying too eagerly and didn't read this far 😑 Well that's sorted then 👍

All is forgiven. Please provide the same courtesy if I do that at some point (which I try to avoid, but no guarantees!). :smile:

I'll keep working on this PR in the coming few days.

polarathene:
> Still, if you consider [the rspamd community repository for Fedora](https://rspamd.com/downloads.html), they **do** [provide ARM64 packages in their rpm repositores](https://rspamd.com/rpm/centos-7/),

Last I checked that wasn't compatible with any recent version of Fedora. Maybe it's changed since, but I'm a bit doubtful when it's coming from CentOS 7 repo that is a completely different distro (_released over 10 years ago and EOL next year, on a 4.4 kernel with extremely outdated systemd too, no cgroups v2 support for containers_). When I tried earlier this year the linked dependencies was a mismatch in Fedora.

I was able to build the AMD64 for Fedora myself, but had trouble with ARM64.

Zepmann:
> Last I checked that wasn't compatible with any recent version of Fedora. Maybe it's changed since, but I'm a bit doubtful when it's coming from CentOS 7 repo that is a completely different distro (_released over 10 years ago and EOL next year, on a 4.4 kernel with extremely outdated systemd too, no cgroups v2 support for containers_). When I tried earlier this year the linked dependencies was a mismatch in Fedora.

I am not that experienced with RH-based distributions, so forgive me for any ignorance in this. CentOS 7 was just an example. There are also packages [for CentOS 8 and CentOS 9](https://rspamd.com/rpm/). CentOS 9 is quite recent and well supported, but only in the shape of 'stream 9'. I am not sure what that does to compatibility with Fedora, but the claim on Wikipedia is that it is closer to RHEL and further away from Fedora...

Summarized, also less than ideal. I think the main contributors of DMS should agree on what they want:

* Running newer versions of software vs. low maintenance.
* Relying on community repositories or not.
* Compiling stuff yourselves or not.

Maybe a compromise? Ubuntu is Debian-based, so switching to that should be less trouble compared to switching to a distribution with a different background. It has all the packages that are currently used by DMS in its repositories, for both x86_64 and ARM64. Every 6 months there is a new release. From my experience, version upgrades are doable due to not changing much between Ubuntu releases (like package names). DMS would not be running bleeding edge, but also not fall so bar behind.

And yes, `dovecot-fts-xapian` is supported by Ubuntu. 😃

casperklein:
Some thoughts: I mainly chose DMS, because it's Debian and Bash based. Both things, I like and firm with.

Switching from Debian to any other Distro - I am very confident - will introduce other issues/problems, while fixing other things. So there is no perfect distro. Also switching from Bash to python, rust whatever.. is a red flag for me. The latter ones, I've almost no experience. I chose this project, because I can easily hack things together and contribute, which aren't supported out of the box (yet) etc.

Switching the distro and/or scripting language is a fundamental change to this project. A fork may be the better option.

---

>Running newer versions of software vs. low maintenance.
>Relying on community repositories or not.
>Compiling stuff yourselves or not.

I prefer rock solid/secure versions, also if that means that some features are missing. Things like the dovecot community forum where already implemented, when I started using DMS. When software vendors provide their own debian repos or packages and they are reliable, I've also no problems with that.

I probably don't rely on any additional feature, that dovecots community repo offers compared to the base Debian one. However, I think if it's relative easy to implement and use, it doesn't hurt to be up2date.
It's also not really hard to implement, to make those things optional. We already do so with the build argument `DOVECOT_COMMUNITY_REPO`.

At the moment, we already use additional repos (dovecot), install things from package files (fail2ban) and compile things on build time (fts_xapian). If that's done for good reasons, well documented/commented and in their own functions, to make those things optional, that's a good compromise.

So the only important questions "just" seems to be, what to make a projects default, while giving "self-builders" also their freedom.

> @georglauterbach @casperklein is this version mismatch a concern for us going forward?

Not for me. My take on this: the best version for each platform - also if that means there a version mismatches between platforms. As long as our tests passes and no ugly workarounds are needed I am fine with that.

polarathene:
> There are also packages [for CentOS 8 and CentOS 9](https://rspamd.com/rpm/). CentOS 9 is quite recent and well supported, but only in the shape of 'stream 9'. I am not sure what that does to compatibility with Fedora

The rspamd builds link to system dependencies, they're not very portable as the versions of some of these change. I'm not experienced with building C++ projects, it might be possible to bundle their own copies of the dependencies instead as you can do with other languages but I don't know how to do that with the build system rspamd uses (cmake), I tried to mess with it a bit but there is a lot going on which isn't documented 😅 

Chances are nothing packaged for CentOS with rspamd is going to be compatible with Fedora AFAIK. I may eventually get back to contributing official Fedora build support but other DMS tasks come first.

---

> Running newer versions of software vs. low maintenance.

It's a balance. However with Debian there has been a release of PostSRSd 2.0, and even in the existing 1.x series which had a bugfix for running in containers since 2022Q3, Debian 12 did not update this package at all as the debian packager has stepped down, and thus it's unlikely to be addressed any time soon.

New versions are important for security fixes, and sometimes feature support that is requested. Debian often conflicts with that and requires additional effort to support.

> Relying on community repositories or not.

My experience with these in DMS has been discouraging. Especially when it puts the AMD64 + ARM64 platforms out of sync, and with Dovecot eventually introduced a requirement to compile a package ourselves. I don't find it worthwhile, especially when a different base image resolves the concern.

> Compiling stuff yourselves or not.

If we don't maintain the project, then we don't really want to compile it. We may adopt rust in future but maintenance of compiling third-party software should be avoided. Again a better base distro choice would alleviate us of this.

> Maybe a compromise? Ubuntu is Debian-based, so switching to that should be less trouble compared to switching to a distribution with a different background.

I'm open to that if there were clear advantages, but I'm not sure I see any addressing the concerns with Debian?

However, [they're still relying on Debian package for `postsrsd`](https://packages.ubuntu.com/search?keywords=postsrsd&searchon=names&suite=all&section=all) (_[version `1.10`](https://github.com/roehling/postsrsd/releases/tag/1.10) from Dec 2020_). Since then there has been [`1.11`](https://github.com/roehling/postsrsd/releases/tag/1.11) (March 2022) and [`1.12`](https://github.com/roehling/postsrsd/releases/tag/1.12) (Aug 2022) with the bugfix I wanted. Meanwhile [since Jan 2023 there has been the `2.x` series](https://github.com/roehling/postsrsd/releases), which will need us to adapt configuration.

Same [situation with `rspamd`](https://packages.ubuntu.com/search?suite=all&section=all&arch=any&keywords=rspamd&searchon=names) and [Dovecot too](https://packages.ubuntu.com/search?suite=all&section=all&arch=any&keywords=dovecot&searchon=names). So nothing really improves there?

Other benefits of Ubuntu don't really apply to a container environment so I'm not sure if we'd gain much.

> Oh, and Ubuntu has their own Docker images, which is preferable over community images.

We use official debian image?:

https://github.com/docker-mailserver/docker-mailserver/blob/894978ddd751f0d09e9f68e1a3de13926e17d432/Dockerfile#L10

That wouldn't change with a new base image either.

polarathene:
> Some thoughts: I mainly chose DMS, because it's Debian and Bash based. Both things, I like and firm with.
> 
> Switching from Debian to any other Distro - I am very confident - will introduce other issues/problems, while fixing other things. So there is no perfect distro. Also switching from Bash to python, rust whatever.. is a red flag for me. The latter ones, I've almost no experience. I chose this project, because I can easily hack things together and contribute, which aren't supported out of the box (yet) etc.
> 
> Switching the distro and/or scripting language is a fundamental change to this project. A fork may be the better option.

That's fine, I don't think the base image will be changing to a different distro. None of the active maintainers are aligned with a preferred choice:
- I am advocating Fedora.
- @georglauterbach thinks Alpine would be a good choice (_which I am certain is risky and for a project like DMS will introduce more issues elsewhere_)
- @casperklein wants to remain with Debian (_but has had a clear history of maintenance issues, notably with packaging_)

While I can cite Debian and Alpine concerns. I am not sure what issues would be experienced with the Fedora base image and it is something I'd like to eventually explore. Personally, if issues are experienced at a similar scale, we could easily revert back to Debian, but otherwise I don't see the harm since Debian isn't exactly serving us as well as I'd like. I'm often tasked with catching / addressing image upgrade concerns too, both @casperklein and @georglauterbach have similar maintenance work involved with respective packages they are associated to for maintenance when that shouldn't be required IMO.

Migrating away from Bash is a little different and both myself and @georglauterbach agree on the maintenance benefits.
- Bash has been awkward and added notable friction in some cases with its limitations / quirks. We have some poor workarounds especially with the test suite regarding this.
- Bash has not really attracted any additional maintainers and rarely much contribution wise (_unless fairly simple_).
- When your experienced with Bash it's a bit easier to grok the codebase, but some of the features used are not easy to search for, nor immediately apparent what they're doing. This is less of a concern with real programming languages, especially when they offer solid official docs that are user-friendly.

I personally won't be pushing for either until I'm happy with the source for Debian + Bash. I also am not likely to remain a maintainer beyond that milestone so my opinion matters less. I don't mind making a fork from that point as a proof of concept, but it won't be something I'd continue to pursue.

I'd rather instill confidence in those changes to all active maintainers. I do my research and don't make decisions lightly, they're often backed with good reasons. I came to DMS with minimal bash knowledge, bash scripts intimidated me but I figured it out. If Rust were adopted, the hard part would already be taken care of for you, maintenance is much simpler and you'd have access to those with experience that are as friendly and supportive as has been the case with bash when I came onboard.

---

> Not for me. My take on this: the best version for each platform - also if that means there a version mismatches between platforms. As long as our tests passes and no ugly workarounds are needed I am fine with that.

This Lua PR reveals the mismatch with Dovecot, which in the past probably would have been a compatibility issue as I think we embraced the community repo for newer Dovecot to support some features that Debian release was unable to support (_I also recall something similar for the `getmail6` contribution?_).

Rspamd is another example of mismatch and probably issues reported like [these two segfaults mentioned](https://github.com/docker-mailserver/docker-mailserver/issues/3581#issuecomment-1762964600) are related to Debian packaging vs upstream official package. Unfortunately we can't easily test ARM64 builds (_probably with very slow emulation, but these aren't easy problems to troubleshoot_).

Rspamd isn't as simple to compile and maintain that just for DMS. Likewise we have `postsrsd` concern where we lack the bugfix and will likely continue to lack it, let alone the upgrade to `2.0`. I don't think the answer here is to justify compiling that as well, but we lack bugfixes and security patches that would land for that project as a result by sticking with Debian and not compiling it.

`postsrsd` concern isn't related to platform version mismatch, but definitely a general package concern with Debian. Other packaging concerns are like the Postfix post-install one which requires a workaround during image build that shouldn't be necessary.

The Debian 12 upgrade isn't going to be as smooth of a process as my review feedback on that makes clear. We have legacy service scripts still lingering and compatibility concerns within a container to address (_Debian 12 SA cron maintenance script, where systemd isn't viable in a container but installed/detected_). The Dovecot community repo / backports situation there is another blocker at present, Debian specific.

---

I value harmony with the maintainers though, so the best I can do is voice those frustrations and offer up solutions (_and the upfront effort to implement them if everyone is onboard_).

I contribute and fix quite a bit in the project, especially the more complicated issues so I'd hope that experience has some weight. I try to avoid any bias and focus on what is good for the project and its health.

Major decisions to change the project such as the base image or migrating away from bash make sense to me, but they're not worth the risk of losing a valued maintainer over.

Zepmann:
> This Lua PR reveals the mismatch with Dovecot

Just for expectation management:

While it reveals the mismatch for DMS as a whole, the mismatch is *not* a really an issue for this PR. The difference in package names can easily be fixed. I just need more time for it since I also work on some other things. Compared tot soms of the other fixed, this ons is quite minimal.

If the status quo is not changed, I would suggest that some effort is put in making the build system a bit more dynamic. Make it so that there is a default path (only Debian, no community repos), that new paths can be developed to be used next to the default path, and to allow each path to have support for optional 'quirks', such as the addition of community repos.

Combined with the test suite, this would allow you to:

- Pick a path that simply works. Software might be older, but a release would always be possible. This is the baseline. The baseline must always be able to produce an image that supports all (mandatory) tests, and it should keep platforms in sync (as far as the underlying distribution does this).
- Work on the development of quirks to support newer software without influencing the baseline if something breaks. Due to quirks being developed seperately, it should be easier to maintain them. Dovecot Xapian and Lua would be quirks in the current situation. Ideally quirks will keep platforms in sync, but it is not a requirement. Of course, test should still pass when quirks are applied.
- In time, use another distribution for the default baseline, so that less quirks are needed. Development of this can be done parallel to the current default baseline, and switching should be easy.

As for the Bash vs. Rust discussion, my two cents: It is unlikely that I would work on this PR if Rust was required due to my experience with Bash. In the suggested situation this would not matter if quirks can be written in whatever the maintainer is comfortable with.

polarathene:
I didn't expect this response to be so large. No obligation to respond to any of it 👍 (_it is going a bit off-topic_)

Take your time with updating this PR and applying the feedback, no rush :)

---

> Make it so that there is a default path (only Debian, no community repos), that new paths can be developed to be used next to the default path, and to allow each path to have support for optional 'quirks', such as the addition of community repos.

DMS only supports one base image, and that will remain Debian if there is a lack of consensus to justify otherwise. I would doubt that'll change any time in the near future.

Repo wise you have what is in the `packages.sh` file. IIRC this only really differs for:
- Dovecot (_which has a build arg to opt-out of the community repo, at the expense of building / maintaining it yourself instead of using the officially published images of DMS_). I'm not fond of the default personally given the maintenance burden it's contributed, but I don't have any interest to debate it further than what I've said above.
- Rspamd with and older ARM64 release (Debian package) vs official rspamd third-party repo. The ARM64 image of DMS is known to have problems including segfaults with rspamd and it's likely due to the packaging differences, not just version mismatch.
- This PR, which is effectively just a package name difference given the main issue being Dovecot repo / platform support.

You're welcome to debate the community repo being disabled by default, it used to be. You'd need to convince @casperklein as I think they're advocating for that strongly, and I assume some contributors/users that may only chime in only when aware of the change being made/released. We don't have any pressing user-facing issues being reported, when they exist @casperklein resolves them, so the maintenance burden is his which is acceptable for us, given all the other work we need to tackle.

---

> Pick a path that simply works. Software might be older, but a release would always be possible

A release is possible. Just no one is pushing for one right now. Our release process could be improved, but this is more of a problem with the lack of volunteer time across existing maintainers and no new maintainers / frequent contributors for years. We mostly get small drive-by contributions and never see the contributor again 🤷‍♂️ 

An improvement we would like is to better support releasing patch versions for prior `major.minor` semver releases. However the current release process is PRs pile up on `master` branch and a release is tagged.
- We'd need long-lived release branches with cherry-picking and CI updated accordingly.
- Related to that is conventional commits, but maintainer consistency aside, contributors like yourself don't respect that with your individual commits (non-issue since we squash merge) or PR title.
  - We could implement some CI check / bot to assist with this, but it'd probably add more friction which I prefer is minimized to encourage more contributions. It'd also require more time from someone to implement that CI support, in the past it's just been talked up, but no one has the time.
  - Likewise, we could lighten the load of releases by requiring PRs to contribute an update to `CHANGELOG.md` or at the very least as part of the PR description. Those aren't without gotchas either though 😅 (_they've been discussed_)

Not great when a vulnerability exists that gets patched or it's been a long time since release and the ClamAV database bundled could do with an update for the current release tag.
- We have this at least for the `:edge` build which benefits from scheduled builds, but lack the CI infrastructure to support versioned releases.
- Even then, the scheduled builds are affected by leveraging the existing image build cache, the cache is not skipped (_but probably should be in that scenario_), so it's dependent upon the cache for `master` being expired from stale age (7 days?) or freshly added (_via layer invalidation, but that'd vary based on where the invalidation occurs_).

All of that is a nice-to-have though, not a priority over other maintenance tasks.

---

> The baseline must always be able to produce an image that supports all (mandatory) tests, and it should keep platforms in sync (as far as the underlying distribution does this).

AFAIK, that's what our test suite already does? To the best of its ability at least.
- There are quite a few tests that are suboptimal/poor, but again volunteer time is a limited resource, addressing those concerns is lower priority than resolving others.
- No IPv6 tests (_would require modifying `/etc/docker/daemon.json`_) to do correctly, and that'd need to have special handling to not be done implicitly for local test runs (_where modifying your Docker config isn't something you'd likely welcome_), should only be done via CI. Docker networking itself is a "fun" topic with a history of problems (_and gotchas that still exist today_).
- Tests don't cover some environment conditions, notably different versions of Docker or hosts which are known to influence the container to fail in unexpected ways. Nor is it exactly practical to maintain such coverage.
- There is work towards adapting the test suite to be more flexible for matrix testing of different feature combinations that isn't practical to support / maintain variants manually.
  - This would also raise the quality of tests by configuring and using a local DNS resolver to support test coverage not presently possible with the current approach.
  - It will also improve confidence by not relying on bypassing some security/networking checks as a result. It will greatly help instill more confidence in changing some fragile networking logic and compatibility coverage for features like LDAP across the test suite instead of very narrow test coverage.
- ARM64 image lacks test coverage, it's not very practical for us with the CI available to enforce test coverage on that. Even if we did troubleshooting in a way that we can resolve it is likely low. We rely on:
  - Success that the ARM64 image at least builds successfully.
  - User reports where depending on the failure it might be resolvable. Sometimes it's very niche issues or misconfiguration / unsupported environment from the user, or they just never respond back when we request more information 🤷‍♂️ 

---

> Work on the development of quirks to support newer software without influencing the baseline if something breaks. Due to quirks being developed seperately, it should be easier to maintain them.
> Dovecot Xapian and Lua would be quirks in the current situation.
> Ideally quirks will keep platforms in sync, but it is not a requirement. Of course, test should still pass when quirks are applied.

While what you propose may make some sense, it should be clear from the above that it's not likely to be addressed any differently than what we currently have.
- We rely on ENV feature flags, and document what we can. Any common issues with quirks usually end up in the docs such as debugging page, FAQ or IPv6, sometimes other feature specific pages.
- Community contributions are tricky. A user wants a feature, they contribute it and often the review focuses on quality of functionality being implemented correctly, a review pass on docs, and ideally tests, but the tests are high level of friction (especially in the past), so we've been a bit more lenient there. `getmail` + `fetchmail` features for example only test that the config files are handled correctly when the ENV feature is toggled on, but doesn't actually have any test of the features functionality (_to bring in mail via POP3 for example_), mainly because neither party has the time to invest in resolving that.
- In the past there was a much lower quality bar for contributions, this introduced a variety of maintenance issues and bugs. I'm quite strict when I review (_but excessively verbose_), the other active maintainers are good too and spot things I don't while a bit more relaxed (_time is a valuable resource, especially when it's voluntary_).

Your PR here for example would probably not be too fun to write up tests for. You'd have more luck now than in the past as I've led an effort to better unify common test setup and modularize out tests. We have some good examples already for auth verification testing, you'd mostly need to configure another container as the auth provider, and leveraging the LDAP support helps, especially once I get around to finishing up my LDAP rework 😅 (_the other maintainers are less confident with maintaining LDAP however, as was I earlier in the year, that brings some maintenance burden as a result_)

---

> In time, use another distribution for the default baseline, so that less quirks are needed. Development of this can be done parallel to the current default baseline, and switching should be easy.

The project won't maintain / develop base image variants, no maintainer will be onboard with that AFAIK.

Switching to another base image is not a simple task for the project either.
- I know of reasons why changing the image to Fedora or Alpine would break even if you corrected `packages.sh` to use the right package manager and packages. Examples are `/etc/services` contents and `hostname` command (differs in behaviour), and I'm sure other fun gotchas that lack coverage.
- Due to the single image approach with many services / processes bundled, there is added complexity vs maintaining other Docker images that tend to avoid this. Our image also relies on the base image implicitly a bit as mentioned with examples, features that you normally wouldn't need to think about with the average Docker image to run a container successfully.

I'm reluctant to explore changing the base image until better test coverage is available personally.

---

> As for the Bash vs. Rust discussion, my two cents: It is unlikely that I would work on this PR if Rust was required due to my experience with Bash.

That's understandable and expected.

The issue for maintainers that want to migrate away from Bash is that it's barely proven to be an advantage luring in regular contributors, whilst having a variety of drawbacks that a more appropriate replacement like Rust would mitigate.

The expectation is that we'd just attract a different group of contributors. That may not prove to be any better than the contributor situation with bash. I was inexperienced with bash, but wanted to use DMS and contribute improvements to the TLS support, so I just figured it out.

When you have an existing codebase to reference and input from maintainers when you get stuck, it's not as difficult to contribute when you're inexperienced (_I've done this with a variety of languages I have never used prior in other projects_).

> In the suggested situation this would not matter if quirks can be written in whatever the maintainer is comfortable with.

That would be quite the burden on maintenance. None of us are really fond of other languages creeping in like Python, but we do have it (_and for that case, a bash equivalent wouldn't be better IMO_).

LDAP support almost got dropped due to lack of any maintainer understanding LDAP or wanting to, its support degraded over the years as a result creating a variety of LDAP specific issues which maintainers weren't able to really investigate and resolve.

You could see how that would probably apply to reducing contributor friction with language choice instead of prioritizing what works best for maintainers.

> With this, I think it would be easier for each maintainer to work on the parts they are comfortable with and which motivates them.

That's already the case. Off the top of my head (_probably missing some things_):
- @casperklein and @georglauterbach are bash wizards, while I grit my teeth usually and curse at anything beyond basic scripts 😆 
- Both have features they care about and pretty much own responsibility for (_Fail2Ban for @casperklein and Rspamd for @georglauterbach off the top of my head_) beyond general maintenance areas where they have great expertise too.
- @georglauterbach and I have revised the test suite quite a bit, and if you use `git blame` you'll probably run into big efforts by @georglauterbach to restructure and better organize the project (_I do similar with the docs_).
- I am usually found tackling stuff that nobody else wants to go near, I get OCD and hyper-focus on addressing niche topics. CI, networking/security and docs are usually my primary areas of contribution. Other than that, I'm quite involved with triage/troubleshooting and reviews.

We would love more maintainers or frequent contributors for DMS though. I sometimes burn myself out, so I'm not the most productive maintainer at times 😅 

---

> It would at least motivate me if I was asked to maintain Lua support. 😄

I think you'd have difficult time getting Lua introduced beyond the docs for reasons already mentioned.

Contributing `dovecot-lua`, you do get the joy of being responsible for keeping the feature around; if any issues arise you'll get pinged 😉 

Your experience with bash would be very welcome though! (_I really don't see DMS migrating away from it anytime soon due to the amount of effort to port away, especially when opposed by maintainer(s) which could include you!_ 😎 )

Zepmann:
@polarathene 

Thanks for the extensive explanation. I didn't know the history of DMS, and see that some tough choices were made and still have to be made.

> Your PR here for example would probably not be too fun to write up tests for.

I agree. That is why I push this as an advanced option with minimal support. Unlike LDAP,  there is no way to standardize this or to effectively test it thoroughly. I think the approach of 'it builds, and there are no problems unless people find them' is the correct one in this. The same for ARM64 support.

I don't think this place is the proper location to continue discussion about this. Perhaps open a separate issue about it (if it does not exist already) is more appropriate. Although I don't have the ideal answer, I would be happy to think with you guys.

In the meantime I'll work on this PR. Once it is done and approved I'll take a small break. After that, I'll see if I can help somewhere else.

georglauterbach:
I do not have the time to read through the whole thread, so here are my two cents:

The version discrepancy of Dovecot between amd64 and arm64 is similar to Rspamd, and not an issue if you ask me. It is not perfect, and I try to keep the differences for Rspamd at a minimum. We should not care about this at the moment if you ask me.

About the newly added package: this may sound harsh, but I really don't see why this could not be done with `user-patches.sh`. I really appreciate the Documentation effort, and I think we should merge that 🚀 But I see bigger maintenance issues on the horizon with the package addition. So, can we maybe only merge an adjusted version of the new documentation?

polarathene:
@georglauterbach 

> I really don't see why this could not be done with `user-patches.sh`.

This was explained in their response prior to two maintainer approvals: https://github.com/docker-mailserver/docker-mailserver/issues/3575#issuecomment-1756350338

> _ All I did was add `dovecot-lua` to the package list in packages.sh (and disable my `user-patches.sh`, of course). Lua support in Dovecot works immediately.
> I would not like to use `user-patches.sh`, since the deb packages in the image do not have to be in sync with the deb packages in Dovecot's repository.

Previously they had been using `user-patches.sh`, but the concern was due to third-party repo not being in sync with image release. Since we don't regularly update any tags beyond `:edge` due to current CI infrastructure this is a valid concern while the Dovecot community repo remains the default in builds.

In my docs review I have pointed out that they may need to rely on `user-patches.sh` in future for other changes to reliably be applied, as replacing entire Dovecot config files for small changes is not advisable, especially when it's a custom config we override with file replacement ourselves (_as has been problematic for some users when we changed the Dovecot SASL auth socket file path_).

But the package concern AFAIK is valid to warrant inclusion into the image instead of `user-patches.sh`.

---

> I see bigger maintenance issues on the horizon with the package addition.

As per the issue link with maintainer approvals and my docs review feedback above, maintainers have been very clear that @Zepmann is responsible for resolving any related maintenance issues that arise with the inclusion of this package.

If it causes us any problems that are not resolved to maintainer satisfaction in a timely manner, the package will be dropped in a future release. This is the stance for any community contributed packages / features that are deemed niche like `fetchmail` and `getmail` (_which also lack thorough testing_).

To further take that stance, I've also suggested like the debug group in `packages.sh` we do something similar for community contributed packages, so that the distinction for maintenance is clear.

target/scripts/build/packages.sh
Zepmann: A comment was added. Furthermore, a distinction is now made between x86_64 (use Dovecot community repository) and other architectures (use Debian's repository). A comment was added to clarify this.

docs/content/config/advanced/auth-lua.md
```diff
@@ -0,0 +1,194 @@
...
+
+Dovecot has the ability to let users create their own custom user provisioning and authentication providers in [Lua](https://en.wikipedia.org/wiki/Lua_(programming_language)#Syntax). This allows any data source that can be approached from Lua to be used for authentication, including web servers. It is possible to do more with Dovecot and Lua, but other use cases fall outside of the scope of this documentation page.
+
+!!! warning
+
+    DMS offers minimal support for Lua-based authentication due to it being an advanced method that can be used in many different ways. Do not open issues on GitHub or request support from DMS contributors for Lua scripts. The documentation on this page is all that is needed to get your own development started. Look elsewhere for Lua support.
+
+!!! warning
+
+    Lua-based authentication relies on a Dovecot plugin. Dovecot is known to sometimes deprecate and remove support for their plugins (such as [CheckPassword](https://doc.dovecot.org/configuration_manual/authentication/checkpassword/)). DMS will drop support immediately if at some moment continued inclusion of the Lua plugin would not align with the DMS development process anymore.
```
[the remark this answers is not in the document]
Zepmann: I made some slight alterations to the suggestion. It can concern multiple packages, depending on whether community repositories are used. The details of this are not relevant for this community guide, so I made it a bit more generic without taking anything away from the warning.

This will be resolved in the next documentation commit.

docs/content/config/advanced/auth-lua.md
```diff
@@ -0,0 +1,194 @@
...
+1. If the answer is 'yes' to question 1 or 2: are there other methods that better facilitate my use case instead of custom scripts which rely on me being a developer and not just a user? **No. Only HTTP can be used to authenticate against Nextcloud, which is not supported by Dovecot or DMS.**
+
+While it is possible to extend what Nextcloud supports with [Nextcloud apps](https://apps.nextcloud.com/), there is currently a mismatch between what DMS supports and what Nextcloud applications support. This might change in the future. For now, Lua will be used to bridge the gap between DMS and Nextcloud for authentication only (Dovecot passdb), while LDAP will still be used to identify mailboxes and verify authorization (Dovecot userdb).
+
+## Container variables to adjust
+
+Since Docker Mailserver provides minimal support for Lua with Dovecot, there are no environment variables to configure specifically for Lua. Some environment variables that do exist for other aspects of DMS might need to be configured based on the scenario you want to follow.
```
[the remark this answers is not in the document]
Zepmann: Removed as suggested. Change will show up in the next documentation commit.

docs/content/config/advanced/auth-lua.md
```diff
@@ -0,0 +1,194 @@
...
+
+???+ example
+
+    DMS configuration environment variables to let Postfix apply Lua authentication by proxy through Dovecot for authenticated SMTP.
+
+    ```yaml
+    - ENABLE_SASLAUTHD=1
+    - SASLAUTHD_MECHANISMS=rimap
+    - SASLAUTHD_MECH_OPTIONS=127.0.0.1
+    ```
```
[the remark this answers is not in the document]
Zepmann: I tested this, and confirmed what you wrote.

The whole section will be removed with the next documentation commit.

docs/content/config/advanced/auth-lua.md
Zepmann: Done for the part about finding out which Lua version is used by Dovecot. The part about using additional libraries I want to keep at the place it currently is, since it is essential for Lua development and a bit of a pain to figure out if starting from scratch.

This change is part of the next documentation commit.

docs/content/config/advanced/auth-lua.md
Zepmann: Agreed. Suggestion has been implemented in the next documentation commit.

docs/content/config/advanced/auth-lua.md
Zepmann: I based my references on those used in the [LDAP documentation](https://github.com/docker-mailserver/docker-mailserver/blob/master/docs/content/config/advanced/auth-ldap.md?plain=1). No matter. It is altered as suggested, and will show up in the next documentation commit.

docs/content/config/advanced/auth-lua.md
```diff
@@ -0,0 +1,194 @@
...
+    Override and add Dovecot configuration files and lua scripts.
+
+    ```yaml
+        volumes:
+          - ./docker-data/dms/config/dovecot/auth-ldap.conf.ext:/etc/dovecot/conf.d/auth-ldap.conf.ext:ro
+          - ./docker-data/dms/config/dovecot/auth-lua-httpbasic.conf:/etc/dovecot/conf.d/auth-lua-httpbasic.conf:ro
+          - ./docker-data/dms/config/dovecot/lua/:/etc/dovecot/lua/:ro
+    ```
+
+The first volume line [overrides][docs-dovecotoverrideconfiguration] Dovecot's standard LDAP authentication configuration file. The second line [adds][docs-dovecotaddconfiguration] a new configuration file for Lua authentication. The third line adds a directory which will contain Lua scripts. The files and directory will not be changed from inside the container, which is why they are configured as read-only.
```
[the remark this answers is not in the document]
Zepmann: For a Lua directory I would not use individual binding of files. That would require unnecessary edits of `compose.yaml` anytime a Lua library is added. A separate directory was created for Lua files to avoid conflicts with configuration files.

Also note that Lua is not just configuration, but also program code. I do not know if it would be a good idea to reload Dovecot's configuration anytime a Lua script changes. Since it concerns an unsupported/advanced feature, wouldn't it be better to leave reloading of the configuration to the user?

Since this concerns future changes, I would leave it for now as it is. This is community-contributed documentation. It can lag a bit behind official releases and a bit of possible breakage is expected between releases. Once DMS includes the relevant changes, the documentation can be updated to incorporate these where relevant/necessary.

The next documentation commit has the comments included inside the `configuration.yaml` example code.

docs/content/config/advanced/auth-lua.md
```diff
@@ -0,0 +1,194 @@
...
+
+Make the necessary changes on the filesystem (*where `mailserver` is the container name*):
+```bash
+mkdir -p ./docker-data/dms/config/dovecot/lua
+docker cp mailserver:/etc/dovecot/conf.d/auth-ldap.conf.ext ./docker-data/dms/config/dovecot/auth-ldap.conf.ext
+```
```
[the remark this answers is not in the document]
Zepmann: > * `*` should be `_` for our md italic convention.

Made this more consistent with the rest of the documentation, so the comment is not needed at all. Note that I base my documentation on other documentation pages, and some other pages also use `*` instead of `_`. It is not an issue anymore for this page.

>     * `compose.yaml` will create the lua directory itself (_although it'd be under the ownership of the user that `dockerd` runs as, which isn't always desirable_). Not necessary to hand-hold for that issue though I think.

The reason why I let the user make the directory beforehand is so that DMS does not have to be restarted (which would be overkill just to create a single directory). These files are read-only anyway. I agree that handholding is not necessary. 
 
>     * `docker cp` is fine, but this is referring to an existing container already existing.

As noted earlier, it is assumed that DMS is already up-and-running with LDAP authentication. Just referring to docker cp without any further hand-holding is fine. If a user does not realize a container needs to be running to copy files from it, the user would better not be messing around with Lua in the first place.

> When LDAP is enabled (`ACCOUNT_PROVISIONER=LDAP`), our startup scripts will disable our default Dovecot [passdb + userdb `passwd-file` config](https://github.com/docker-mailserver/docker-mailserver/blob/82c38f242648290ab109de8e3994ea7564120deb/target/dovecot/auth-passwdfile.inc) and enable the `ldap` config instead by adjusting comments in the `10-auth.conf` config:
> 
> https://github.com/docker-mailserver/docker-mailserver/blob/82c38f242648290ab109de8e3994ea7564120deb/target/scripts/startup/setup.d/ldap.sh#L54-L55

I am counting on this in this scenario.

> `user-patches.sh` could just as easily opt-out of the config by applying the reverse to the ldap config. Which may be relevant in future.
> 
> That has the following file contents:
> Click to view
> 
> In future for DMS we may switch to configs with separate passdb + userdb files as supporting OIDC may only be valid as a passdb AFAIK. If that happens we'll update your guide here.

That is allowed (of course!), but you can also give me a signal. I can help with the update. Lua authentication can be used for passdb + userdb, but the example only uses it for passdb. This makes it quite similar to OIDC, so it would be nice to implement it in the same way as OIDC would.
 
> Since there is anything special to copy from this config and you're advising to mount a different config anyway, there isn't much reason to include this step?

I disagree. The first line creates the lua directory so the user can work with it immediately. The second line copies the existing LDAP configuration file from the current container. This is better compared to including static configuration in the guide, in case the LDAP configuration file changes between now and when OIDC is implemented. Again, it is up to the user to make the relevant changes for the specific use case in the specific moment in time. The guide provides just an example which is kept up-to-date using best effort.

Of course, I realize an update of the guide will be necessary in the future. It'll happen when it is needed.

docs/content/config/advanced/auth-lua.md
Zepmann: > Likewise, this may break for you in future when you're relying on DMS being configured for a specific `ACCOUNT_PROVISIONER` which makes certain assumptions. This is probably another caveat that could be mentioned in the disclaimer at the start of this guide, since breakage can work both ways.

A good point. I added an additional line to the warning at the top of this guide to note this.

> When the OIDC feature is closer to being merged, we may be able to support an BYO config, but I think that'd still run into compatibility issues with some features of DMS (_eg: Dovecot sieve scripts, v13 recently received a fix to adjust the home directory within an accounts mailbox, LDAP doesn't have that and some users may explicitly configure their own location which won't be compatible_). Semver releases will hopefully at least prevent the likelihood of breakage for you between major releases.

We'll cross that bridge when we get there.

> BTW this section for Dovecot seems it's missing enabling the config? The LDAP config will be included, but you'll likely need a line similar to this:
> 
> https://github.com/docker-mailserver/docker-mailserver/blob/82c38f242648290ab109de8e3994ea7564120deb/target/dovecot/10-auth.conf#L125
> 
> I don't recommend volume mounting a replacement config for `10-auth.conf`, so `user-patches.sh` is probably beneficial for you. Keep in mind that if you just append the needed line, without a conditional to prevent adding it again this will pile up across container restarts (_unless the user follows our advice to always re-create the container, but many seem to miss this_), as internal state is persisted within the container until it's removed.

Unnecessary. `auth-lua-httpbasic.conf` has the `.conf` extension. All `.conf` files are loaded automatically. See:

```
dovecot.conf:!include conf.d/*.conf
```

> Eventually, we'll probably provide our customizations from a separate location, and always copy them over during container startup to minimize the issue so that config modifications are more deterministic.

I am looking forward to that. Please document this, and point me in the right direction when the time comes. I will be happy to adjust the current example to a better supported environment for community contributions.

docs/content/config/advanced/auth-lua.md
Zepmann: A sloppy copy. Fixed in the next documentation commit.

docs/content/config/advanced/auth-lua.md
```diff
@@ -0,0 +1,194 @@
...
+#passdb {
+#  driver = ldap
+#
+#  # Path for LDAP configuration file, see example-config/dovecot-ldap.conf.ext
+#  args = /etc/dovecot/dovecot-ldap.conf.ext
+#}
+```
+
+Don't touch anything else in the file.
```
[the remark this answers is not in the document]
Zepmann: I'll keep this in mind when the time comes to revise the documentation when the way DMS works with configuration files is changed in the future. Of course, you can also update it yourself then. Alternatively, hit me up.

docs/content/config/advanced/auth-lua.md
```diff
@@ -0,0 +1,194 @@
...
+
+There are several questions you should ask yourself before you start:
+
+1. Do I want to use Lua to identify mailboxes and verify that users are are authorized to use mail services? This refers in the world of Dovecot to Lua providing 'userdb' functionality, as in a data source for user provisioning.
+1. Do I want to use Lua to verify passwords that users authenticate with for IMAP/POP3/SMTP in their (web) mail clients? This refers in the world of Dovecot to Lua providing 'passdb' functionality, as in a data source for user password verification.
+1. If the answer is 'yes' to question 1 or 2: are there other methods that better facilitate my use case aside of custom scripts which rely on me being a developer and not just a user?
+
+If the answer is 'no' to question 3, Lua-based authentication might just be the thing for you.
```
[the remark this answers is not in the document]
Zepmann: I revised it myself. Please see the next documentation commit.

Zepmann:
I have updated the documentation based on comments from @polarathene. Furthermore, I changed `packages.sh` so that the Dovecot community repository is only used when relevant (i.e. the platform is x86_64). This fixes the ARM build issue.

Everything works with the current development version of DMS. As discussed, there will be changes to DMS in the future that will break the example in the documentation. I think it is best to start with a new PR when that time comes. This is a community effort, and as such it is no issue if it lags behind official development and breaks sometimes. The documentation also notes this as a warning for any users who want to use Lua-based authentication.

docs/content/config/advanced/auth-lua.md
```diff
@@ -0,0 +1,180 @@
...
+          - ./docker-data/dms/config/dovecot/auth-lua-httpbasic.conf:/etc/dovecot/conf.d/auth-lua-httpbasic.conf:ro
+          # Directory containing Lua scripts
+          - ./docker-data/dms/config/dovecot/lua/:/etc/dovecot/lua/:ro
+    ```
+
+Make the necessary changes on the filesystem:
+```bash
+mkdir -p ./docker-data/dms/config/dovecot/lua
+docker cp CONTAINER_NAME:/etc/dovecot/conf.d/auth-ldap.conf.ext ./docker-data/dms/config/dovecot/auth-ldap.conf.ext
+```
```
polarathene, line 64: Regarding the prior review feedback with your response: https://github.com/docker-mailserver/docker-mailserver/pull/3579#discussion_r1366689491

> The reason why I let the user make the directory beforehand is so that DMS does not have to be restarted (which would be overkill just to create a single directory).
> These files are read-only anyway. I agree that handholding is not necessary.

We have different perspectives. I assume the user does not already have a container running, especially since the earlier documentation requires configuration change to `compose.yaml` for volumes.

The lua directory will be created by Docker during that startup if it doesn't exist (_albeit with ownership of the Docker daemon user/group, which is sometimes inconvenient_).

If the user is providing Lua files here as you'd expect...they'd be capable of creating the directory themselves? So I questioned why `mkdir` was necessary to mention at all.

---

> As noted earlier, it is assumed that DMS is already up-and-running with LDAP authentication.
> Just referring to docker cp without any further hand-holding is fine.

Either I'm tired and missing the mention of a running DMS container with LDAP, or the guide to this point only mentions DMS will be configured to use LDAP, not that one is already setup and running.

> If a user does not realize a container needs to be running to copy files from it, the user would better not be messing around with Lua in the first place.

I'm not concerned about the user awareness with the `docker cp` command, but if it has any real relevance at all. It seems it can be dropped, as all you're doing is making a copy and editing it, you could just as easily create the file since it's being mounted back into the same place within the container. The user is reading a guide, copy/paste is fine.

---

> I disagree.
> The first line creates the lua directory so the user can work with it immediately.
> The second line copies the existing LDAP configuration file from the current container. This is better compared to including static configuration in the guide

The lua dir has been clarified above, I really don't see why `mkdir` is necessary but I'm not bothered leaving it there.

However since `docker cp` doesn't seem necessary, it really seems like this step is redundant.

You argue that it's better to copy than provide a static config, while the related [review feedback discussion](https://github.com/docker-mailserver/docker-mailserver/pull/3579#discussion_r1357902407) regarding this notes that `docker cp` won't be resilient to such:
- In an already existing setup after this guide, no change will be handled by the user (_unless it's documented in release notes and they read that when upgrading_), thus broken setup. This has happened with existing configs where users have replaced them entirely only to change one line.
- If such a change happens, new users following the guide may find the `docker cp` instruction is either copying over a config no longer used/enabled for DMS LDAP support, or with changes since this guide was last updated that may be incompatible (_nothing comes to mind, other than some runtime modification during startup_).

My main concern is that it's rather implicit on the expectation of the contents. The main goal is to disable the `passdb` config, but the docs could better communicate/document the expectation of what `userdb` config was at the time of the guide being last updated.

Not a major concern, since one can use `git blame` on the docs, and check what version of DMS it was written for, then pull that image to get the contents. If it's going to be fragile regardless, being more explicit seems better 🤷‍♂️ 

```
userdb {
  driver = ldap
  args = /etc/dovecot/dovecot-ldap.conf.ext
}
```

target/scripts/build/packages.sh
```diff
@@ -99,16 +99,24 @@ function _install_dovecot() {
...
-    _log 'trace' 'Using Dovecot community repository'
-    curl https://repo.dovecot.org/DOVECOT-REPO-GPG | gpg --import
-    gpg --export ED409DA1 > /etc/apt/trusted.gpg.d/dovecot.gpg
-    echo "deb https://repo.dovecot.org/ce-2.3-latest/debian/bullseye bullseye main" > /etc/apt/sources.list.d/dovecot.list
-
-    _log 'trace' 'Updating Dovecot package signatures'
-    apt-get "${QUIET}" update
+    # Dovecot's deb community repository only provides x86_64 packages, so do not include it
+    # when building for another architecture.
+    if [[ "$(uname --machine)" == "x86_64" ]]; then
```
polarathene, line 108: I don't think the added level of nesting is necessary here, you should be able to add it as a 2nd condition since there is no other conditional branches to handle.

```bash
# Dovecot's deb community repository only provides x86_64 packages,
# do not include it when building for another architecture
if [[ ${DOVECOT_COMMUNITY_REPO} -eq 1 ]] && [[ "$(uname --machine)" == "x86_64" ]]; then
```

target/scripts/build/packages.sh
```diff
@@ -99,16 +99,24 @@ function _install_dovecot() {
...
+      _log 'trace' 'Using Dovecot community repository'
+      curl https://repo.dovecot.org/DOVECOT-REPO-GPG | gpg --import
+      gpg --export ED409DA1 > /etc/apt/trusted.gpg.d/dovecot.gpg
+      echo "deb https://repo.dovecot.org/ce-2.3-latest/debian/bullseye bullseye main" > /etc/apt/sources.list.d/dovecot.list
+
+      _log 'trace' 'Updating Dovecot package signatures'
+      apt-get "${QUIET}" update
+
+      # Additional community package needed for Lua support if the Dovecot community repository is used.
+      DOVECOT_PACKAGES+=(dovecot-lua)
```
polarathene, line 118: Please apply similar handling + comment for the `dovecot-auth-lua` package.

polarathene (changes requested):
Great work! Thanks for taking the time to address and respond to feedback 😁 

Two changes remaining:
- Add title to admonition example.
- Address feedback for `packages.sh`.

The two large review comments are responses to prior discussion and can be ignored. One debates the relevance of `mkdir` and `docker cp` (_vs clearly documenting an explicit userdb config override for the versioned release at time of writing_), but I won't block on either of these.

---

Should the lua auth docs be part of the config docs, or would they be better suited in the [examples section](https://docker-mailserver.github.io/docker-mailserver/v12.1/examples/tutorials/basic-installation/)? (_tutorials / use-cases_)

Seems like it may be better suited there given the focus as a community guide? Up to you :)

Zepmann:
This PR is not forgotten. I'll pick it up later this week.

georglauterbach:
I've marked this for v13.1.0. The next release is v13.0.0, so this PR will not block on v13.0.0

docs/content/config/advanced/auth-lua.md
```diff
@@ -0,0 +1,180 @@
...
+1. Do I want to use Lua to verify passwords that users authenticate with for IMAP/POP3/SMTP in their mail clients? **Yes. Password authentication is done through Lua.**
+1. If the answer is 'yes' to question 1 or 2: are there other methods that better facilitate my use case instead of custom scripts which rely on me being a developer and not just a user? **No. Only HTTP can be used to authenticate against Nextcloud, which is not supported by Dovecot or DMS.**
+
+While it is possible to extend what Nextcloud supports with [Nextcloud apps](https://apps.nextcloud.com/), there is currently a mismatch between what DMS supports and what Nextcloud applications support. This might change in the future. For now, Lua will be used to bridge the gap between DMS and Nextcloud for authentication only (Dovecot passdb), while LDAP will still be used to identify mailboxes and verify authorization (Dovecot userdb).
+
+## Modify Dovecot's configuration
+
+Add the following volume values to the relevant part of `compose.yaml`:
+
+???+ example
```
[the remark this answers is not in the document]
Zepmann: Done. Part of the next commit.

docs/content/config/advanced/auth-lua.md
Zepmann: `docker cp` has been removed. Instead, the user creates a new file with the stated example `userdb` content. Change will be visible in the next commit.

target/scripts/build/packages.sh
Zepmann: A keen eye, and correct. Done. Change will be visible in next commit.

docs/content/config/advanced/auth-lua.md
```diff
@@ -0,0 +1,180 @@
...
+
+    ```yaml
+        # All new volumes are marked :ro to configure them as read-only, since their contents are not changed from inside the container
+        volumes:
+          # Configuration override to disable LDAP authentication
+          - ./docker-data/dms/config/dovecot/auth-ldap.conf.ext:/etc/dovecot/conf.d/auth-ldap.conf.ext:ro
+          # Configuration addition to enable Lua authentication
+          - ./docker-data/dms/config/dovecot/auth-lua-httpbasic.conf:/etc/dovecot/conf.d/auth-lua-httpbasic.conf:ro
+          # Directory containing Lua scripts
+          - ./docker-data/dms/config/dovecot/lua/:/etc/dovecot/lua/:ro
```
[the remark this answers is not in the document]
Zepmann: I now understand the feedback concerns the two added configuration files (for LDAP user provisioning and Lua user authentication). What I do not understand is if any more changes to this PR are required at this time. If there are, please note these concretely.

If not, I do realize that due to future changes to `check-for-changes.sh` there might be a better solution that does not depend on binding individual files.

Zepmann:
Finally had some time to work on this!

I think most of the comments have been solved adequately. If not, please let me know.

Also, the Lua documentation was simplified somewhat. By configuring Nextcloud (as the server) better, it is not necessary to test in Lua (as the client) if the provided password is a Nextcloud application password. The relevant Nextcloud configuration parameters are also included now in the documentation.

Finally, I am considering adding this documentation to the examples section instead of as a main page in the advanced configuration section, but I do not know if this would make the review process harder. Therefore, please review this first, but do not merge the PR if you think it is acceptable. Instead, let me know your comments and I'll make another change to move the documentation to the proper section.

polarathene:
> I am considering adding this documentation to the examples section instead of as a main page in the advanced configuration section, but I do not know if this would make the review process harder.
> I am considering adding the documentation to the 'Use Cases' subsection of Examples

Yes, this sounds like a good choice 👍 

Thanks for waiting for review before making the change, I had been thinking of relocating it from my initial review but I didn't want to deal with any added review friction :)

Great work, thanks for the contribution to DMS! ❤️ 

> Please notify me if Dovecot Lua support is ever considered problematic or a feature to be removed. This way, I can add my $ 0.02 and maybe help by solving the problem or by thinking of alternatives.

Will do 👍

polarathene (approved):
LGTM 👍 

Waiting on one last change before merge, relocation of docs to examples section.

Zepmann:
Thanks for the extensive feedback and an open mind, @polarathene. I'm looking forward to working together again in the future.

 Currently I'm busy with some other stuff. I'll move the documentation page later today (as in 'within 12 hours from now', but I'll do it ASAP within that timeframe).

For expectation management, is this PR to be included in release v13.0.0 or v13.1.0, considering the timeline and the v13.1.0 milestone?

georglauterbach:
> For expectation management, is this PR to be included in release v13.0.0 or v13.1.0, considering the timeline and the v13.1.0 milestone?

If this is to be merged in the next few days, it's going to be in v13.0.0.

Zepmann:
Moved the documentation. I have nothing more to add for this PR unless something new pops up. In that case, let me know! :smiley:

polarathene (approved):
LGTM 👍 

Thanks for your patience and the great contribution! 😁

PR #3585, opened 2023-10-17
docs: fix path to rspamd.log

ap-wtioit:
# Description

fix the remaining /var/log/supervisor/rspamd.log to also point the user to /var/log/mail/rspamd.log

Discovered this when i had to replace the log path in our local tests (using docker-mailserver with enabled rspamd to check if our mails conform to those non RFC rules). It seems this was missed in 894978ddd751f0d09e9f68e1a3de13926e17d432

## Type of change

## Checklist:

Info @wt-io-it


georglauterbach:
Good catch! I should have `grep`ed for this in my PR.

georglauterbach (approved):

PR #3592, opened 2023-10-21
feat: Add MTA-STS support for outbound mail

### LINKED ISSUE #3590: Support MTA-STS out of the box
### Context

MTA-STS is the equivalent to HSTS, e.g. a mechanism for a domain to declare that STARTTLS is supported and should be used, even if the initial SMTP exchange is intercepted by bad actors. The protocol is overcomplicated involving both DNS records and HTTPS, but the necessary glue in DMS is minimal.

### Description

Install `postfix-mta-sts-resolver` as part of the container image. Introduce a new flag ENABLE_OUTGOING_MTA_STS flag. If it is set, start `/etc/init.d/postfix-mta-sts-resolver` and add:

```
smtp_tls_policy_maps = socketmap:inet:127.0.0.1:8461:postfix
```

to `/etc/postfix/main.cf`.

### Alternatives

Running `postfix-mta-sts-resolver` in a separate container is possible, but messy. Current (temporary) workaround is doing it `user-patches.sh`.

### Applicable Users

Anyone sending mails to Gmail and Outlook.

If no one bets me to it, I'll sort out the necessary changes, but the description is pretty much all that is needed beside documentation.

jsonn:
# Description

Add optional MTA-STS support for outgoing mails. Domains can signal STARTTLS
support out-of-band using a combination of DNS and HTTPS to prevent
man-in-the-middle attacks that filter out the feature.

Fixes #3590

## Checklist:


mailserver.env
```diff
@@ -346,6 +346,13 @@ POSTFIX_REJECT_UNKNOWN_CLIENT_HOSTNAME=0
 # Note: More details at http://www.postfix.org/postconf.5.html#inet_protocols
 POSTFIX_INET_PROTOCOLS=all
 
+# If enabled, STARTTLS support is enforced for outgoing mails to domains
+# with MTA-STS records like Google Mail.
+# This can prevent man-in-the-middle that hide the STARTTLS feature.
+# - **0** ==> MTA-STS disabled
+# - 1 => MTA-STS enabled
+ENABLE_MTA_STS=0
```
polarathene, line 354: We have a similar page in the docs that should be updated as well.

While this is only outbound support for MTA-STS, it may warrant a bit more documentation.
- [Configuration for the daemon service](https://github.com/Snawoot/postfix-mta-sts-resolver#mta-sts-daemon-configuration). Defaults should be fine, so only mentioning it in this PR may be fine.
- [Configuration of Postfix](https://github.com/Snawoot/postfix-mta-sts-resolver#postfix-configuration), this PR covers that and we already configure the CA path location. Nothing needs to be covered regarding that.
- [DANE compatibility](https://github.com/Snawoot/postfix-mta-sts-resolver#warning-mta-sts-policy-overrides-dane-tls-authentication), this may be worth making users aware of.

Thus **minimal requirement** is to document this setting in ENV docs page, and reference the DANE concern.

Documenting support for MTA-STS inbound via policy file and DNS record (_neither requiring integration with DMS, but might be of interest to users interested in the feature for our docs to cover_) would be nice, but not a requirement.

polarathene (changes requested):
While I'd like to see a complimentary test (_eg: [as described here](https://github.com/Snawoot/postfix-mta-sts-resolver#operability-check)_), this is not an easy ask due to needing a DNS service setup if we want to avoid external network requests.

We could probably still implement the postmap query against gmail? Trusted delivery verification on the logs won't be viable though.

---

The PR is fine as-is, but should additionally update the ENV docs page, and reference the DANE issue for awareness in case any users have configured that.

target/scripts/startup/setup.d/mta_sts.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# Set up MTA-STS
+
+function _setup_mta_sts() {
+  _log 'trace' 'Adding MTA-STS lookup to the Postfix TLS policy map'
+  postconf 'smtp_tls_policy_maps = socketmap:inet:127.0.0.1:8461:postfix'
```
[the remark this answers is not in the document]
jsonn: It's messy as it requires inter-dependencies between containers. For something that I would consider an important security property, this problematic. I don't have a strong preference for communication locally via TCP oder UDS.

I'm still looking into the best option for testing. It seems to be a pity that there currently is no easy test domain in the wild and I'm trying to reach out to folks in the hope of something being willing to host one.

mailserver.env
jsonn: I'll work on the documentation side of thing.

target/scripts/startup/setup.d/mta_sts.sh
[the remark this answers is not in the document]
polarathene: It's not too difficult to do this all locally, we don't need an external test domain.

It's just something I've been putting off until the test suite is refactored more to switch to docker compose as I can more easily manage flexible configurations.

---

If you'd like to implement it, I'll allow that.
- Use a [Caddy container](https://hub.docker.com/_/caddy) to host the policy file. We can use our [existing local testing certs](https://github.com/docker-mailserver/docker-mailserver/tree/master/test/test-files/ssl/example.test), Caddy needs to be provided these along with a little config.
- Use a [CoreDNS container](https://hub.docker.com/r/coredns/coredns/) with a BIND file zone to handle the DNS records.
- Ensure that DMS has the CA root cert in the trust store for verifying trust against the Caddy container. The query CLI tool from the project your using packages that separately on Debian.
- I'm not sure if submitting mail to be sent to a domain the DMS container manages itself will work for the logs, it might optimize that delivery if it recognizes that it loops back into itself. That may require another DMS instance, but I don't think we're setup for that supporting that in tests yet 🤔(maybe)

I have familiarity with all of the above, so could probably contribute the Caddy + CoreDNS setup via a separate PR, then you could rebase to leverage that. I'll let you know if I get that handled, if you know of a better way I'm open to ideas :)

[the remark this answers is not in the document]
jsonn: The main reason I would like to see a test domain exist is to give the user a way to test the correct behavior on their own. E.g. for DKIM there are validation services and the report functions. Without this testing, #3593 might have gone undetected for a while. I was just trying to resist the setup so far :)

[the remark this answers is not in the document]
polarathene: I have a working local offline setup with caddy + coredns to resolve the lookup via `mta-sts-query` from the package. Uses the test CA + cert I linked for TLS.

A user could setup the same locally with two DMS instances to test setup is correct, then after that is successful use real public DNS + certs 🤷‍♂️ 

I'll put a test together tomorrow to handle the sending and logs verification.

---

Our test suite is slowly getting better, once it adopts the local DNS setup more broadly it can have better test coverage to catch issues like you've linked, assuming contributors have time to write the tests 😅 

> The main reason I would like to see a test domain exist is to give the user a way to test the correct behavior on their own.
> E.g. for DKIM there are validation services and the report functions.

I can't assist much with public validation concern, but I think being able to have an offline setup is a useful tool for getting confidence in getting everything working for new users :)

georglauterbach:
Just FYI: This PR and #3593 are the last blockers for [v13.0.0](https://github.com/docker-mailserver/docker-mailserver/milestone/21).

docs/content/config/advanced/mail-mta-sts.md
```diff
@@ -0,0 +1,27 @@
...
+title: 'Advanced | MTA-STS'
+---
+
+MTA-STS is an optional mechanism for a domain to signal support for
+STARTTLS. It can be used to prevent man-in-the-middle-attacks hiding the
+feature to force mail servers to send outgoing emails as plain text.
+MTA-STS is an alternative to DANE without the need of DNSSEC.
+
+MTA-STS is supported by some of the biggest mail providers like Google Mail
+and Outlook.
```
polarathene, line 11:
```suggestion
MTA-STS is an optional mechanism for a domain to signal support for STARTTLS.

- It can be used to prevent man-in-the-middle-attacks from hiding STARTTLS support that would force DMS to send outbound mail through an insecure connection.
- MTA-STS is an alternative to DANE without the need of DNSSEC.
- MTA-STS is supported by some of the biggest mail providers like Google Mail and Outlook.
```

docs/content/config/advanced/mail-mta-sts.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Supporting MTA-STS for outgoing mails
```
polarathene, line 13:
```suggestion
## Supporting MTA-STS for outbound mail
```

docs/content/config/advanced/mail-mta-sts.md
```diff
@@ -0,0 +1,27 @@
...
+
+This is enabled by setting `ENABLE_MTA_STS=1` [](../environment.md#enable_mta_sts)
+in the environment.
```
polarathene, line 16:
```suggestion
Enable this feature via the ENV setting [`ENABLE_MTA_STS=1`](../environment.md#enable_mta_sts).
```

docs/content/config/advanced/mail-mta-sts.md
```diff
@@ -0,0 +1,27 @@
...
+
+!!! warning
+
+    MTA-STS will by default override DANE if both are in used by a domain.
+    This can be partially addressed by configuring a dane-only policy resolver
+    before the MTA-STS entry in smtp_tls_policy_maps. See [the postfix-mta-sts-resolver documentation](https://github.com/Snawoot/postfix-mta-sts-resolver#warning-mta-sts-policy-overrides-dane-tls-authentication)
+    for further details.
```
polarathene, line 23:
```suggestion
!!! warning "If you have configured DANE"

    Enabling MTA-STS will by default override DANE if both are configured for a domain.

    This can be partially addressed by configuring a dane-only policy resolver before the MTA-STS entry in `smtp_tls_policy_maps`. See the [`postfix-mta-sts-resolver` documentation][postfix-mta-sts-resolver::dane] for further details.

[postfix-mta-sts-resolver::dane]: https://github.com/Snawoot/postfix-mta-sts-resolver#warning-mta-sts-policy-overrides-dane-tls-authentication
```

docs/content/config/advanced/mail-mta-sts.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Supporting MTA-STS for incoming mails
+
+A good introduction can be found on [dmarcian.com](https://dmarcian.com/mta-sts/).
```
polarathene, line 27:
```suggestion
## Supporting MTA-STS for inbound mail

While this feature in DMS supports ensuring STARTTLS is used when mail is sent to another mail server, you may setup similar for mail servers sending mail to DMS.

This requires configuring your DNS and hosting the MTA-STS policy file via a webserver. A good introduction can be found on [dmarcian.com](https://dmarcian.com/mta-sts/).

```

docs/content/config/environment.md
```diff
@@ -108,6 +108,15 @@ This enables DNS block lists in _Postscreen_. If you want to know which lists we
 - **0** => DNS block lists are disabled
 - 1     => DNS block lists are enabled
 
+##### ENABLE_MTA_STS
+
+Enables MTA-STS for outgoing mails.
```
polarathene, line 113:
```suggestion
Enables MTA-STS support for outbound mail.
```

mailserver.env
```diff
@@ -346,6 +346,13 @@ POSTFIX_REJECT_UNKNOWN_CLIENT_HOSTNAME=0
 # Note: More details at http://www.postfix.org/postconf.5.html#inet_protocols
 POSTFIX_INET_PROTOCOLS=all
 
+# If enabled, STARTTLS support is enforced for outgoing mails to domains
+# with MTA-STS records like Google Mail.
+# This can prevent man-in-the-middle that hide the STARTTLS feature.
```
polarathene, line 351:
```suggestion
# Enables MTA-STS support for outbound mail.
# More details: https://docker-mailserver.github.io/docker-mailserver/latest/config/advanced/mail-mta-sts/
```

target/scripts/startup/setup.d/mta_sts.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# Set up MTA-STS
+
+function _setup_mta_sts() {
+  _log 'trace' 'Adding MTA-STS lookup to the Postfix TLS policy map'
```
polarathene, line 6: This is not "adding", but rather setting/overriding. If there is any existing config for this, it's replaced to only be for MTA-STS. We have a [helper for this that will avoid this concern](https://github.com/docker-mailserver/docker-mailserver/blob/d988d8a8d1be8bd07c271eaa9911f15f5cd033c7/target/scripts/helpers/postfix.sh#L102) 👍

target/scripts/startup/setup.d/mta_sts.sh
polarathene, line 7:
```suggestion
  _add_to_or_update_postfix_main smtp_tls_policy_maps 'socketmap:inet:127.0.0.1:8461:postfix'
```

polarathene (changes requested):
I'll apply these.

docs/content/config/advanced/mail-mta-sts.md
```diff
@@ -0,0 +1,28 @@
+---
+title: 'Advanced | MTA-STS'
```
polarathene, line 2:
```suggestion
title: 'Best Practices | MTA-STS'
hide:
  - toc # Hide Table of Contents for this page
```

polarathene: Please also rename this file to `mta-sts.md`, and move the file to `docs/content/config/best-practices/`.

You'll need to also add an entry here for it to be added to the actual docs:

https://github.com/docker-mailserver/docker-mailserver/blob/d988d8a8d1be8bd07c271eaa9911f15f5cd033c7/docs/mkdocs.yml#L124-L126

Add this entry after the DKIM one:

```yaml
- 'MTA-STS': config/best-practices/mta-sts.md
```

docs/content/config/environment.md
```diff
@@ -108,6 +108,15 @@ This enables DNS block lists in _Postscreen_. If you want to know which lists we
...
 - 1     => DNS block lists are enabled
 
+##### ENABLE_MTA_STS
+
+Enables MTA-STS support for outbound mail.
+
+- **0** => Disabled
+- 1 => Enabled
+
+See [MTA-STS](advanced/mail-mta-sts.md) for further explanation.
```
polarathene, line 118:
```suggestion
See [MTA-STS](best-practices/mta-sts.md) for further explanation.
```

mailserver.env
```diff
@@ -346,6 +346,12 @@ POSTFIX_REJECT_UNKNOWN_CLIENT_HOSTNAME=0
 # Note: More details at http://www.postfix.org/postconf.5.html#inet_protocols
 POSTFIX_INET_PROTOCOLS=all
 
+# Enables MTA-STS support for outbound mail.
+# More details: https://docker-mailserver.github.io/docker-mailserver/latest/config/advanced/mail-mta-sts/
```
polarathene, line 350:
```suggestion
# More details: https://docker-mailserver.github.io/docker-mailserver/latest/config/best-practices/mta-sts/
```

target/scripts/startup/setup.d/mta_sts.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
polarathene, line 1: Could you please also rename this file to use kebab-case? `mta-sts.sh`, thanks!

polarathene (changes requested):
The docs contribution is quite nice, thanks for tackling that! I revised it a little bit and removed the line length limit (a11y concern).

These remaining changes are for you to handle:
- Update frontmatter title + add setting to hide ToC
- Rename docs file and relocate it to _Best Practices_ section.
- Update references to the new location.
- Add an entry to the docs config `mkdocs.yml` for the page to be accessible.
- Rename startup script to use kebab-case.

docs/mkdocs.yml
```diff
@@ -124,6 +124,7 @@ nav:
     - 'Best Practices':
       - 'DKIM, DMARC & SPF': config/best-practices/dkim_dmarc_spf.md
       - 'Auto-discovery': config/best-practices/autodiscover.md
+      - 'MTA-STS': config/best-practices/mta-sts.md
```
polarathene, line 127:
```suggestion
      - 'Auto-discovery': config/best-practices/autodiscover.md
      - 'DKIM, DMARC & SPF': config/best-practices/dkim_dmarc_spf.md
      - 'MTA-STS': config/best-practices/mta-sts.md
```

polarathene (dismissed):
LGTM 👍 

This can be merged or wait until a test is ready.

reneploetz:
If I read the config correctly, the resolver will be backed by an sqlite database in ```/var/lib/mta-sts/cache.db```. Does it make sense to move that file to our general mail state directory?

Also, the debian version is 1.1.2-1.1. Upstream has 1.4.0, though apart from documentation there is not much change. This issue seems to be the only relevant one fixed apart from the ability to run as a specific user (and not root): https://github.com/Snawoot/postfix-mta-sts-resolver/issues/90

jsonn:
I've pinged the Debian maintainer.

georglauterbach (dismissed):
Approving here as well as changes look good to me 👍🏼

polarathene:
> Does it make sense to move that file to our general mail state directory?

We could, not too important as it's just cache?

If @jsonn thinks it'd be good to persist, this is what you need to do for that support:

Add the conditional ENV here:

https://github.com/docker-mailserver/docker-mailserver/blob/cb62ce20e6f49d211a8f264b11feaed65582a338/target/scripts/startup/setup.d/mail_state.sh#L20-L22

next is restoring ownership since the UID/GID assigned aren't necessarily stable across upgrades, you'd need to know what the original ownership is:

https://github.com/docker-mailserver/docker-mailserver/blob/cb62ce20e6f49d211a8f264b11feaed65582a338/target/scripts/startup/setup.d/mail_state.sh#L83-L84

That should be it :)

polarathene:
> Related question: would documentation and configuration for DANE be useful as well? The world can't agree on a single scheme after all.

I rarely see interest in it expressed by users here, and last I recall there were issues with adoption or something with DANE?

You're welcome to contribute docs for that if you like, maybe it'll help our users adopt it, especially if they're not aware or feel uneasy with third-party advice? 🤷‍♂️

georglauterbach:
What's the status here?

jsonn:
No reaction from the Debian maintainer so far.  I've cleaned up the configuration a bit so that the daemon runs as non-root and the cache is preserved. I think that's it from my side.

georglauterbach:
Perfect! Please just quickly take care of the lint failure (Hadolint reports to consecutive `RUN` statements) and we're good to merge this. I would not wait for the Debian maintainer - I already have experience when it comes to waiting for Debian packages, not great fun.

georglauterbach (dismissed):

georglauterbach:
@polarathene when you also approve, go ahead and merge :)

polarathene:
This failure seems to be unrelated to changes of this PR.

---

[Test failure](https://github.com/docker-mailserver/docker-mailserver/actions/runs/6724202103/job/18275988451?pr=3592#step:6:62)

```
 not ok 57 [Change Detection] should detect and process changes (in both containers with shared config) in 2186ms
# (from function `assert_output' in file test/test_helper/bats-assert/src/assert_output.bash, line 186,
#  from function `_assert_has_standard_change_event_logs' in file test/tests/parallel/set3/scripts/changedetector.bats, line 106,
#  from function `_should_perform_standard_change_event' in file test/tests/parallel/set3/scripts/changedetector.bats, line 87,
#  in test file test/tests/parallel/set3/scripts/changedetector.bats, line 48)
#   `_should_perform_standard_change_event "${CONTAINER2_NAME}"' failed
#
# substring (1 lines):
#   Removed lock
# output (18 lines):
#   [   INF   ]  2023-11-01 19:34:33  Change detected
#   [  TRACE  ]  Creating lock '/tmp/docker-mailserver/check-for-changes.sh.lock'
#   [  TRACE  ]  2023-11-01 19:34:33  Regenerating accounts (Dovecot + Postfix)
#   [  TRACE  ]  Checking file line endings
#   [  TRACE  ]  Regenerating dovecot masters list
#   [  DEBUG  ]  Creating master user 'masterusername'
#   [  TRACE  ]  Checking file line endings
#   [  TRACE  ]  Regenerating postfix user list
#   [  DEBUG  ]  Creating user 'user1' for domain 'localhost.localdomain'
#   [  DEBUG  ]  Creating user 'user2' for domain 'otherdomain.tld'
#   [  DEBUG  ]  Creating user 'user3' for domain 'localhost.localdomain' with attributes 'userdb_mail=mbox:~/mail:INBOX=~/inbox'
#   [  DEBUG  ]  Adding alias 'alias1@localhost.localdomain' for user 'user1@localhost.localdomain' to Dovecot's userdb
#   [  DEBUG  ]  Alias 'alias2@localhost.localdomain' is non-local (or mapped to a non-existing account) and will not be added to Dovecot's userdb
#   [  DEBUG  ]  Adding alias '@localdomain2.com' for user 'user1@localhost.localdomain' to Dovecot's userdb
#   [  TRACE  ]  2023-11-01 19:34:33  Regenerating relay config (Postfix)
#   [  TRACE  ]  2023-11-01 19:34:33  Regenerating vhosts (Postfix)
#   [  DEBUG  ]  2023-11-01 19:34:33  Reloading services due to detected changes
#   [  DEBUG  ]  2023-11-01 19:34:33  Completed handling of detected change
# --
#
```

Fails on the last check here:

https://github.com/docker-mailserver/docker-mailserver/blob/f674232f7116acd1d5b2758c52ab5f5827d4d928/test/tests/parallel/set3/scripts/changedetector.bats#L39-L49

due to last check here:

https://github.com/docker-mailserver/docker-mailserver/blob/f674232f7116acd1d5b2758c52ab5f5827d4d928/test/tests/parallel/set3/scripts/changedetector.bats#L77-L88

because `Removed lock` assert failed here:

https://github.com/docker-mailserver/docker-mailserver/blob/f674232f7116acd1d5b2758c52ab5f5827d4d928/test/tests/parallel/set3/scripts/changedetector.bats#L103-L108

polarathene (changes requested):
I am mostly interested in the [change to non-root user commit](https://github.com/docker-mailserver/docker-mailserver/pull/3592/commits/0faed2f769b4bbc61503d955685776ec033be5ca).

- I am fine with running as the `_mta-sts` user (_I wasn't aware the package added that user_).
- Not sure why TCP bind address was swapped for the unix socket though? Seems like it should be ok, just curious on the motivation.



---

In a future release [user can be switched via CLI flags](https://github.com/Snawoot/postfix-mta-sts-resolver/pull/97) as [noted by prior review comment](https://github.com/docker-mailserver/docker-mailserver/pull/3592#issuecomment-1778095814).

The CLI also provides options to set [a log file, verbosity and PID file](https://github.com/Snawoot/postfix-mta-sts-resolver/blob/master/man/mta-sts-daemon.1.adoc) if we need that.

---

## Change Requests (Required for approval)

- [x] ~~`Dockerfile`: The `chown -R` should be for the `_mta-sts` group as well, to be consistent with the equivalent in `mail_state.sh`? (_Although I'm not sure if the `chown` is really necessary in `Dockerfile`? `mail_state.sh` will run before `supervisord` starts the daemon service_)~~ (_**UPDATE:** Ignore as [different content path](https://github.com/docker-mailserver/docker-mailserver/pull/3592#discussion_r1393307133) used_)
-  [x] Two change suggestions that include a comment referencing upstream docs.
- [ ] As this feature adds a process/service to manage, we should also update a relevant test that we have.

---

### Guidance for updating relevant test

Test update requires you adding the process name here:

https://github.com/docker-mailserver/docker-mailserver/blob/f674232f7116acd1d5b2758c52ab5f5827d4d928/test/tests/parallel/set3/container_configuration/process_check_restart.bats#L40-L52

Explicitly disable the ENV here:

https://github.com/docker-mailserver/docker-mailserver/blob/f674232f7116acd1d5b2758c52ab5f5827d4d928/test/tests/parallel/set3/container_configuration/process_check_restart.bats#L54-L68

Explicitly enable the ENV here:

https://github.com/docker-mailserver/docker-mailserver/blob/f674232f7116acd1d5b2758c52ab5f5827d4d928/test/tests/parallel/set3/container_configuration/process_check_restart.bats#L90-L106

The process name is used with `pgrep --list-full process_name_here` and `pkill --echo process_name_here`, refer to the comments in the test file with other processes if it helps. Usually it matches the binary, but not always. As this is python based it should be similar to fail2ban?

polarathene:
> @polarathene when you also approve, go ahead and merge :)

Once the remaining feedback is addressed I can do that 👍 

---

I still want to get the testing addressed, but juggling some other non-DMS tasks right now. My new approach for using `compose.yaml` seems like it'll work, but needs some thought with configuring DNS + network.

Since we run multiple tests concurrently, we can't just pin IP addresses to a single docker network/subnet and have them all use the `mail.example.test` FQDN in DNS either. We'd likely be creating new isolated networks, for each test file, or needing to leverage something like Traefik perhaps to route DNS to Traefik, then have Traefik connect the containers via container labels 🤔 

I'll get back to figuring that out later this month or next hopefully.

target/mta-sts-daemon/mta-sts-daemon.yml
```diff
@@ -0,0 +1,6 @@
+path: /var/run/mta-sts/daemon.sock
+mode: 0666
```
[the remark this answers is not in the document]
jsonn: Yes, the mode only applies to the UNIX socket. The socket doesn't expose anything secret, so I don't see a strong reason for making the access stricter.

target/supervisor/conf.d/supervisor-app.conf
```diff
@@ -157,3 +157,13 @@ autostart=false
 stdout_logfile=/var/log/supervisor/%(program_name)s.log
 stderr_logfile=/var/log/supervisor/%(program_name)s.log
 command=/bin/bash -l -c /usr/local/bin/update-check.sh
+
```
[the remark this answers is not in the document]
jsonn: Nothing else seems in the file seems to do that?

target/mta-sts-daemon/mta-sts-daemon.yml
```diff
@@ -0,0 +1,6 @@
+path: /var/run/mta-sts/daemon.sock
```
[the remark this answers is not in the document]
jsonn: You've asked about it two weeks ago. Generally, UDS is somewhat more light-weight than TCP, so as I needed to look at the details of the configuration anyway, it was easy to do it.

Dockerfile
```diff
@@ -191,6 +191,15 @@ COPY target/opendmarc/opendmarc.conf /etc/opendmarc.conf
 COPY target/opendmarc/default-opendmarc /etc/default/opendmarc
 COPY target/opendmarc/ignore.hosts /etc/opendmarc/ignore.hosts
 
+# --------------------------------------------------
+# --- postfix-mta-sts-daemon -----------------------
+# --------------------------------------------------
+COPY target/mta-sts-daemon/mta-sts-daemon.yml /etc/mta-sts-daemon.yml
+RUN <<EOF
+  mkdir /var/run/mta-sts
+  chown -R _mta-sts:root /var/run/mta-sts
```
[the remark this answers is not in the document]
jsonn: Not intentional, will adjust tomorrow.

target/mta-sts-daemon/mta-sts-daemon.yml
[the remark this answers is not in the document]
polarathene: > You've asked about it two weeks ago.

Yes, but I then shortly after expressed that there were upstream concerns with the support vs TCP 😅 

I try to be clear with the main review comment on what needs to be tackled. Thanks for looking into it though 👍 

Happy to keep it as UDS.

target/supervisor/conf.d/supervisor-app.conf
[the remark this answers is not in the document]
polarathene: Right, but I wasn't there when the file was originally created. Sometimes tracking down the docs for things is not straight-forward. This context can be a bit helpful to maintainers, especially when it's for features we rarely engage with over time.

It seemed like an appropriate place to add it. Any objections?

georglauterbach:
This is now the last PR blocking v13.0.0.

target/scripts/startup/setup.d/mta-sts.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# Set up MTA-STS
```
georglauterbach, line 3: Useless documentation :)

```suggestion
```

georglauterbach (changes requested):

georglauterbach (dismissed):

georglauterbach:
Okay, so at the moment, it looks like the `chown` issue has to be resolved, and a test has to be written. I would advise for removing the `chown` in the Dockerfile, because, as @polarathene said, `mail-state.sh` is run before Supervisor starts the processes anyway.

If you need help with the test, ping me @jsonn.

Dockerfile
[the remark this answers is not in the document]
jsonn: Looking at this again, they are different things. In `mail_state.sh`, the permissions of the state directory is set. In `Dockerfile`, the location for the socket is set. The permissions in `/var/run` sometimes keep the directory as root (e.g. clamav, fetchmail) and sometimes to the daemon group (opendkim, opendmarc, postgrey). So things are generally not consistent right now.

georglauterbach:
@jsonn I'd like to make a proposal; people reached out to me for v13.0.0, hence I'd like to ask you whether it'd be fine for you if this feature was included in v13.1.0 instead of v13.0.0. This way I can take care of the release now and there is no pressure with this PR.

jsonn:
There is no reason to rush this PR, so pushing it to 13.1.0 is perfectly fine.

polarathene:
I've troubleshooted the test failure. A separate PR will resolve that by changing the test. Then this PR can be merged 👍

target/supervisor/conf.d/supervisor-app.conf
```diff
@@ -157,3 +157,15 @@ autostart=false
...
+# Docs: https://github.com/Snawoot/postfix-mta-sts-resolver/blob/master/man/mta-sts-daemon.1.adoc
+[program:mta-sts-daemon]
+startsecs=0
+stopwaitsecs=55
+autostart=false
+autorestart=true
+stdout_logfile=/var/log/supervisor/%(program_name)s.log
+stderr_logfile=/var/log/supervisor/%(program_name)s.log
+command=/usr/bin/python3 /usr/bin/mta-sts-daemon --config /etc/mta-sts-daemon.yml
```
polarathene, line 169: This will work in the meantime, since the script called is like the fail2ban python script which has a python3 shebang.

The shebang is sufficient for the command to be run without directly running the script via the python3 command. That affects the pgrep matching without `--full`.

```suggestion
command=/usr/bin/mta-sts-daemon --config /etc/mta-sts-daemon.yml
```

polarathene (changes requested):
I'll apply this.

polarathene:
This just needs a changelog entry added under the features section.

I'll add within 12 hours time if this isn't handled when I return 👍 

---

While I'd like the improved testing (_dependent upon `compose.yaml` + DNS_), I'm presently short on time to handle that and rather see this feature merged.

Thanks for contributing this new feature! 😁

jsonn:
Feel to adjust. The indentation of the block looks off?

docs/content/config/best-practices/mta-sts.md
```diff
@@ -0,0 +1,30 @@
...
+    This can be partially addressed by configuring a dane-only policy resolver before the MTA-STS entry in `smtp_tls_policy_maps`. See the [`postfix-mta-sts-resolver` documentation][postfix-mta-sts-resolver::dane] for further details.
+
+[postfix-mta-sts-resolver::dane]: https://github.com/Snawoot/postfix-mta-sts-resolver#warning-mta-sts-policy-overrides-dane-tls-authentication
+
+## Supporting MTA-STS for inbound mail
+
+While this feature in DMS supports ensuring STARTTLS is used when mail is sent to another mail server, you may setup similar for mail servers sending mail to DMS.
+
+This requires configuring your DNS and hosting the MTA-STS policy file via a webserver. A good introduction can be found on [dmarcian.com](https://dmarcian.com/mta-sts/).
+
```
williamdes, line 30: just in case it is interesting to document, here is my very simple nginx no files config

```nginx
server {
    listen 80 default_server;
    listen [::]:80 default_server;
    server_name _;
    server_tokens off;
    autoindex off;

    location /robots.txt {
        return 200 "User-agent: *\nDisallow: /";
    }

    location / {
        return 204 "";
    }

    # 86400 = 1d
    location /.well-known/mta-sts.txt {
        return 200 "version: STSv1\nmode: testing\nmx: mx1.mails.wdes.eu\nmx: mx2.mails.wdes.eu\nmx: mx3.mails.wdes.eu\nmx: mx4.mails.wdes.eu\nmax_age: 86400\n";
    }
}
```

I open-sourced the repo right now for you: https://github.com/wdes/mta-sts.wdes.eu

williamdes: Should we also recommend tools like
- https://dane.sys4.de/

And nice CLI to perform tests yourself: https://github.com/PennockTech/smtpdane

polarathene: > just in case it is interesting to document

A docs contribution for handling this portion would be good in the meantime yes.

## Web server for policy file

I find Caddy is nicer than Nginx for such configuration, here's the Caddyfile config:

```caddyfile
# MTA-STS needs to serve the policy for supported mail domains:
mta-sts.example.test {
  tls /dms/certs/cert.ecdsa.pem /dms/certs/key.ecdsa.pem
  root * /dms/files
  file_server
}
```

The `mta-tsts.txt` file to serve, this can be inlined into the Caddyfile with it's HereDoc support to `respond` with instead.

```
version: STSv1
mode: testing
mx: mail.example.test
max_age: 604800
```

```yml
services:
  # Serves the MTA-STS policy:
  web:
    image: caddy:2.7
    container_name: mail-web
    volumes:
      - ./Caddyfile:/etc/caddy/Caddyfile:ro
      - ./mta-sts.txt:/dms/files/.well-known/mta-sts.txt:ro
      # Caddy data volume for persistence:
      - ./data/:/data
      # Sharing the same cert with DMS:
      - ./tls/wildcard:/dms/certs:ro
```

---

## DNS

```yml
services:
  # Optional, DNS service used for local testing
  dns:
    image: coredns/coredns:1.11.1
    container_name: mail-dns
    volumes:
      - ./Corefile:/Corefile:ro
      - ./zones/:/zones/:ro
    working_dir: /
```

CoreDNS `Corefile`:

```
example.test {
    file /zones/example.test.zone {
        reload 15s
    }
    log
}

. {
  forward . 1.1.1.1
  log
}
```

`zones/example.test.zone`:

```
$ORIGIN example.test.
$TTL 60

@       SOA    example.test. admin.example.test. 2023102034 4h 1h 14d 10m
@       NS     ns1.example.test.

; Mail
; SPF `mx`: Authorize the DMS mail-server at MX record
@          MX  10 mail.example.test.
@          TXT    "v=spf1 mx -all"

; MTA-STS
_mta-sts   TXT    "v=STSv1; id=2023102030"
_smtp._tls TXT    "v=TLSRPTv1; rua=mailto:tls-reports@example.test"

; A records
@       A      172.16.42.11
mail    A      172.16.42.12
ns1     A      172.16.42.10
mta-sts A      172.16.42.11
```

polarathene (approved):

bilogic:
@jsonn

I relay to a mix of SMTPs using 465 and 587.. 

`email postfix/smtp[2735]: SMTPS wrappermode (TCP port 465) requires setting "smtp_tls_wrappermode = yes", and "smtp_tls_security_level = encrypt" (or stronger)`

is there a way to overwrite the configuration for specific SMTP servers?

polarathene:
Relay isn't related to MTA-STS with the error you're showing.

Presently DMS only supports relay for port 587/25, as port 465 requires those settings to be changed to relay to port 465. To support both you'd need another relay service configured in `/etc/postfix/master.cf` that additionally enables those settings.

You should be able to find advice for such in issues about that topic, one of them may be open as a tracking issue. It's unclear when anyone will have time to contribute more flexible integration, but our docs should cover the technical details of the underlying verbose configuration that users can technically apply themselves (skipping our relay feature entirely).

PR #3597, opened 2023-10-22
bugfix: change Rspamd DKIM default config location

### LINKED ISSUE #3593: setup misplaces DKIM configuration for rspamd
### 📝 Preliminary Checks

### 👀 What Happened?

When `setup config dkim` is used with rspam, the configuration is put into `/etc/rspamd/override.d/dkim_signing.conf`. That location is not persisted, it would be better to put it directly into `/tmp/docker-mailserver/rspamd/override.d/dkim_signing.conf`. This was annoying to find as it was only visible after a restart of the container.

This could be seen either as incomplete documentation or suboptimal behavior of the setup script. I'd prefer if the script handles it better as it already creates the keys in the persistent location.

### 👟 Reproduction Steps

Set `ENABLE_RSPAMD=1` and `ENABLE_OPENDKIM=0`. Run `setup config dkim`. Notice the path setup for the default configuration.

### 🐋 DMS Version

git as of October 18

SuSE MicroOS

```yml
Default compose.yaml
```

### 📜 Relevant log output

```Text
[   INF   ]  Supplying a default configuration ('/etc/rspamd/override.d/dkim_signing.conf')
```
```


_No response_

### PR DESCRIPTION
**Pull request by georglauterbach**
# Description

Instead of using `/etc/rspamd/override.d/dkim_signing.conf`, we will now be using `/tmp/docker-mailserver/rspamd/override.d/dkim_signing.conf`. The new location is persisted (and linked again during startup) and hence better suited.

Fixes #3593

## Type of change

## Checklist:

### DIFF
--- a/target/bin/rspamd-dkim
+++ b/target/bin/rspamd-dkim
@@ -168,8 +168,8 @@ function _preflight_checks() {
   # 2. ${RSPAMD_OVERRIDE_D} has contents already
   #
   # If 1. is true, then we're good since DMS' default setup linked the directory already and we will save
-  # a persisted location in every case. If 1. is false, 2. should be false as well since by the default,
-  # ${RSPAMD_OVERRIDE_D} has not contents - we're good as well. What should logically never happen is
+  # a persisted location in every case. If 1. is false, 2. should be false as well since by default,
+  # ${RSPAMD_OVERRIDE_D} has no contents - we're good as well. What should logically never happen is
   # that 1. is false but 2. is true; this case is caught nevertheless and a warning is emitted.
   if [[ ! -h "${RSPAMD_OVERRIDE_D}" ]]; then
     if rmdir "${RSPAMD_OVERRIDE_D}" &>/dev/null; then
--- a/target/bin/rspamd-dkim
+++ b/target/bin/rspamd-dkim
@@ -154,11 +154,30 @@ function _preflight_checks() {
   # in other functions (after this function was called).
   # Also keep in sync with: target/scripts/startup/setup.d/security/rspamd.sh:__rspamd__run_early_setup_and_checks
   local RSPAMD_DMS_D='/tmp/docker-mailserver/rspamd'
-  RSPAMD_DMS_DKIM_D="${RSPAMD_DMS_D}/dkim"
-  RSPAMD_DMS_OVERRIDE_D="${RSPAMD_DMS_D}/override.d"
+  local RSPAMD_OVERRIDE_D='/etc/rspamd/override.d'
+  readonly RSPAMD_DMS_DKIM_D="${RSPAMD_DMS_D}/dkim"
+  readonly RSPAMD_DMS_OVERRIDE_D="${RSPAMD_DMS_D}/override.d"
 
   mkdir -p "${RSPAMD_DMS_DKIM_D}" "${RSPAMD_DMS_OVERRIDE_D}"
   chown _rspamd:_rspamd "${RSPAMD_DMS_DKIM_D}"
+
+  # Mimmick target/scripts/startup/setup.d/security/rspamd.sh:__rspamd__run_early_setup_and_checks where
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by georglauterbach** [comment]
The test `not ok 42 [Fail2Ban] ban ip on multiple failed login in 133164ms` is starting to get annoying.

**Comment by georglauterbach** [comment]
Just FYI: the reason this PR fixes the mentioned race-condition is because the keys are now directly created in the persisted directory (which is linked to `/etc/rspamd/overrid.d/` when DMS is rebooted at the latest). Previously, it could happen that `/tmp/docker-mailserver/rspamd/override.d/` was not present, hence the configuration is not persisted at all.

I will add a commit in a second that provides the users with more detail about a possible misconfiguration, and provide maintainers with more details on why the script does what it does (by adding a comment).

**Comment by polarathene** [comment]
Is it necessary to rely on config volume like this PR seems to adapt to?

You're doing ownership checks and `chown` at the config volume, is that necessary if like other config volume locations, startup scripts just copied user data to an internal location and adjusts permissions at the internal location? (like handled with certs)

During startup we're root and should be able to handle config for rspamd like we do elsewhere, just curious why runtime config should be sourced from config volume and not an internal one?

File: target/scripts/startup/setup.d/security/rspamd.sh [hunk 1]
```diff
@@ -70,7 +70,7 @@ function __rspamd__run_early_setup_and_checks() {
   readonly RSPAMD_OVERRIDE_D='/etc/rspamd/override.d'
   readonly RSPAMD_DMS_D='/tmp/docker-mailserver/rspamd'
 
-  local RSPAMD_DMS_OVERRIDE_D="${RSPAMD_DMS_D}/override.d/"
+  local RSPAMD_DMS_OVERRIDE_D="${RSPAMD_DMS_D}/override.d"
```
polarathene, line 73: Ok so here we have runtime setup referencing the config volume for rspamd, then later it symlinks the location instead of a copy?

If this is just config files, shouldn't that copy to an internal location instead? That way runtime isn't reliant on the config volume location which is more straight-forward?

Only benefit that comes to mind is avoiding `check-for-changes.sh` monitoring the config volume for rspamd config to sync over via copy? If those changes would require a reload of rspamd, you'd not have any benefit beyond being more implicit here by avoiding support via `check-for-changes.sh`?

target/bin/rspamd-dkim
```diff
@@ -186,8 +214,8 @@ function _create_keys() {
 
 function _check_permissions() {
   # shellcheck disable=SC2310
-  if ! __do_as_rspamd_user ls "${BASE_DIR}" >/dev/null; then
-    _log 'warn' "The Rspamd user ('_rspamd') seems to be unable to list files in the keys directory ('${BASE_DIR}') - Rspamd may experience permission errors later"
+  if ! __do_as_rspamd_user ls "${RSPAMD_DMS_DKIM_D}" >/dev/null; then
```
polarathene, line 217: Is it necessary to run as rspamd user? You use it for the `rspamadm dkim_keygen ...` at line 199 and other commands.

Is there an issue with using the root user to do this like opendkim, then correct permissions if necessary afterwards?

polarathene (approved):
Approving since rspamd integration is your domain and we often have conflicting opinions on how you approach that 😝 

I raised some questions / concerns, but no major issues 👍

target/bin/rspamd-dkim
georglauterbach: Here, it is necessary because at this point, we test whether the Rspamd user can list files inside the directory. The actual `chown _rspamd:_rspamd ...` comes earlier :)

target/scripts/startup/setup.d/security/rspamd.sh
georglauterbach: I am not sure why I didn't actually initially constructed it like that. I will merge this PR and provide a follow-up because then it is easier to track the changes later! Thank you for making me aware of the obvious (again) ❤️

PR #3603, opened 2023-10-28
docs: Clarify default for ENV `FETCHMAIL_PARALLEL`

jsonn:
Make this easier to find when brosing the default environment. Adjust documentation to properly mark the actual default value.

### DIFF
--- a/docs/content/config/environment.md
+++ b/docs/content/config/environment.md
@@ -578,7 +578,9 @@ Note: activate this only if you are confident in your bayes database for identif
 ##### FETCHMAIL_PARALLEL
 
 - **0** => `fetchmail` runs with a single config file `/etc/fetchmailrc`
-- 1 => `/etc/fetchmailrc` is split per poll entry. For every poll entry a separate fetchmail instance is started to allow having multiple imap idle configurations defined.
+- 1 => `/etc/fetchmailrc` is split per poll entry. For every poll entry a separate fetchmail instance is started to [allow having multiple imap idle connections per server][fetchmail-imap-workaround] (_when poll entries reference the same IMAP server_).
+
+[fetchmail-imap-workaround]: https://otremba.net/wiki/Fetchmail_(Debian)#Immediate_Download_via_IMAP_IDLE
 
 Note: The defaults of your fetchmailrc file need to be at the top of the file. Otherwise it won't be added correctly to all separate `fetchmail` instances.
 #### Getmail
--- a/docs/content/config/environment.md
+++ b/docs/content/config/environment.md
@@ -578,7 +578,7 @@ Note: activate this only if you are confident in your bayes database for identif
 ##### FETCHMAIL_PARALLEL
 
 - **0** => `fetchmail` runs with a single config file `/etc/fetchmailrc`
-- 1 => `/etc/fetchmailrc` is split per poll entry. For every poll entry a separate fetchmail instance is started  to allow having multiple imap idle configurations defined.
+- 1 => `/etc/fetchmailrc` is split per poll entry. For every poll entry a separate fetchmail instance is started to allow having multiple imap idle configurations defined.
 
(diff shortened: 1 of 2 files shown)


polarathene (approved):
Thanks for spotting this and contributing the improvement! ❤️

docs/content/config/environment.md
```diff
@@ -577,8 +577,8 @@ Note: activate this only if you are confident in your bayes database for identif
 
 ##### FETCHMAIL_PARALLEL
 
-  **0** => `fetchmail` runs with a single config file `/etc/fetchmailrc`
-  **1** => `/etc/fetchmailrc` is split per poll entry. For every poll entry a separate fetchmail instance is started  to allow having multiple imap idle configurations defined.
+- **0** => `fetchmail` runs with a single config file `/etc/fetchmailrc`
+- 1 => `/etc/fetchmailrc` is split per poll entry. For every poll entry a separate fetchmail instance is started  to allow having multiple imap idle configurations defined.
```
polarathene, line 581:
```suggestion
- 1 => `/etc/fetchmailrc` is split per poll entry. For every poll entry a separate fetchmail instance is started to allow having multiple imap idle configurations defined.
```

mailserver.env
```diff
@@ -397,6 +397,8 @@ ENABLE_FETCHMAIL=0
 
 # The interval to fetch mail in seconds
 FETCHMAIL_POLL=300
+# Workaround IMAP bug by splitting fetchmail.cf per poll entry
```
polarathene, line 400: Could you provide more context regarding the IMAP bug? This differs from the ENV docs which doesn't mention any IMAP bug to workaround.

polarathene (changes requested):

georglauterbach:
@jsonn please take a look at the review :) I want to have this in v13.0.0 🚀

mailserver.env
jsonn: It's mentioned already in https://docker-mailserver.github.io/docker-mailserver/latest/config/environment/#fetchmail_parallel

See also: https://otremba.net/wiki/Fetchmail_(Debian)#Immediate_Download_via_IMAP_IDLE

polarathene: > It's mentioned already

Ok, but that doesn't describe it as a workaround for an IMAP bug? Nor does the reader of the comment here necessary know the content is any different on the web docs for ENV, so the context matters more.

> See also

Yeah that is much better! 👍 

Ideally our web docs would link to that, and the ENV comment here would just reference that for more context.

```suggestion
# Use multiple fetchmail instances (1 per poll entry in fetchmail.cf)
# Supports multiple IMAP IDLE connections when a server is used across multiple poll entries
# https://otremba.net/wiki/Fetchmail_(Debian)#Immediate_Download_via_IMAP_IDLE
```

That conveys the feature much better with additional context right?

docs/content/config/environment.md
```diff
@@ -577,8 +577,8 @@ Note: activate this only if you are confident in your bayes database for identif
 
 ##### FETCHMAIL_PARALLEL
 
-  **0** => `fetchmail` runs with a single config file `/etc/fetchmailrc`
-  **1** => `/etc/fetchmailrc` is split per poll entry. For every poll entry a separate fetchmail instance is started  to allow having multiple imap idle configurations defined.
+- **0** => `fetchmail` runs with a single config file `/etc/fetchmailrc`
+- 1 => `/etc/fetchmailrc` is split per poll entry. For every poll entry a separate fetchmail instance is started to allow having multiple imap idle configurations defined.
```
polarathene, line 581:
```suggestion
- 1 => `/etc/fetchmailrc` is split per poll entry. For every poll entry a separate fetchmail instance is started to [allow having multiple imap idle connections per server][fetchmail-imap-workaround] (_when poll entries reference the same IMAP server_).

[fetchmail-imap-workaround]: https://otremba.net/wiki/Fetchmail_(Debian)#Immediate_Download_via_IMAP_IDLE
```

polarathene (approved):

polarathene:
Thanks for the contribution!

Issue #3557, opened 2023-09-30
[TODO]: Migrate away from legacy `docker` user + group


polarathene:
### Subject

Something else that requires developers attention

### Description

Bit of a history with this one:
- Considering our project is docker focused, this is a horrible choice for `grep` to track down 🤣 
- Presently, this represents our default `5000:5000` static vmail user and group and the relevance AFAIK is only for Dovecot. Possibly only when regarding `/var/mail` ownership.
- This topic relates to addressing [support for non-static UID+GID ownership in `/var/mail`](https://github.com/docker-mailserver/docker-mailserver/issues/3548#issuecomment-1736467572) which is mostly a [subset of our LDAP users](https://github.com/docker-mailserver/docker-mailserver/issues/2238). Ideally we can drop the `chown -R` logic (_or shift it into a `setup ...` command as a [troubleshooting tool we can document scenarios for](https://github.com/docker-mailserver/docker-mailserver/pull/2256#issuecomment-947566692)_).

The issue description is documenting any concerns / knowledge regarding this `docker` account, so that a future PR can justify renaming it to something more appropriate such as `vmail` / `dms-vmail`.
- May be a good idea to try tackle this first: https://github.com/docker-mailserver/docker-mailserver/issues/3554
- `5000:5000` is our default, but with v13 this becomes configurable via ENV.

---

### Before Dovecot - Creation of `docker` user + group (`Dockerfile`)

The [`docker` user and group created during the `Dockerfile` build](https://github.com/docker-mailserver/docker-mailserver/blob/89cb6d85b90f6070d53c15e455ebee916c899c7a/Dockerfile#L154) were added via [this Sep 2015 commit](https://github.com/docker-mailserver/docker-mailserver/commit/5137548acf0afc3ebe23bd5061474ed3dd136a48) from the original DMS project author. The only context provided was a reference to [this issue](https://github.com/docker-mailserver/docker-mailserver/issues/18), but it's lacking information on why the account was added.

Back then I think the mail account support was quite a different story, and it's likely become irrelevant for the original purpose?

---

### Dovecot configs referencing `docker` user + group

April 2016 - DMS [migrated from Courier + Cyrus to Dovecot](https://github.com/docker-mailserver/docker-mailserver/pull/123), which introduced various configs referencing the `docker` user and group: 
[...]

PR #3625, opened 2023-11-07
fix: Drop special bits from Postfix `maildrop/` and `public/` directory permissions

### LINKED ISSUE #3619: docs: Kubernetes requies `allowPrivilegeEscalation: true`
### 📝 Preliminary Checks

### 👀 What Happened?

I have setup DMS within kubernetes cluster utilizing sieve. I got the following error:

```
Nov  4 13:24:37 mailserver-0 dovecot: lmtp: Error: postdrop: warning: mail_queue_enter: create file maildrop/281592.964: Permission denied
Nov  4 13:24:37 mailserver-0 postfix/postdrop[964]: warning: mail_queue_enter: create file maildrop/281592.964: Permission denied
```

The following finding: 
- The postdrop was running with the docker user, thus checking setgid on postdrop, which was correctly set as well as directory permissions are correct.
- Had the kubernetes yaml setup as described in the [DMS k8s docs page](https://docker-mailserver.github.io/docker-mailserver/v12.1/config/advanced/kubernetes/).

According to [kubernetes documentation](https://kubernetes.io/docs/concepts/security/pod-security-standards/):

> _Privilege Escalation (v1.8+)_
> _Privilege escalation (such as via set-user-ID or set-group-ID file mode) should not be allowed. This is Linux only policy in v1.25+ (`spec.os.name != windows`)_

After changing to `allowPrivilegeEscalation: true` it works. Maybe you can adjust the configuration accordingly?

Best Regards

### 👟 Reproduction Steps

_No response_

### 🐋 DMS Version

v12.1.0

talos linux 1.4.1

```yml
containers:
        - name: mailserver
          imagePullPolicy: "IfNotPresent"
          securityContext:
            allowPrivilegeEscalation: true
            readOnlyRootFilesystem: false
            runAsUser: 0
            runAsGroup: 0
            runAsNonRoot: false
            privileged: false
            capabilities:
              add:
                # file permission capabilities
                - CHOWN
                - FOWNER
                - MKNOD
                - SETGID
                - SETUID
                - DAC_OVERRIDE
                # network capabilities
                # - NET_ADMIN  # needed for F2B
                # - NET_RAW    # needed for F2B
                - NET_BIND_SERVICE
                # miscellaneous  capabilities
                - SYS_CHROOT
                - KILL
              drop: [ALL]
            seccompProfile:
              type: RuntimeDefault
          ports:
            - name: transfer
              containerPort: 25
              protocol: TCP
            - name: esmtp-implicit
              containerPort: 465
              protocol: TCP
            - name: esmtp-explicit
              containerPort: 587
            - name: imap-implicit
              containerPort: 993
              protocol: TCP
            - name: repl-implicit
              containerPort: 9091
              protocol: TCP
          envFrom:
            - configMapRef:
                name: mailserver-conf
          volumeMounts:
            - name: dovecot-run
              mountPath: /var/run/dovecot
            - name: script-files
              subPath: user-patches.sh
              mountPath: /tmp/docker-mailserver/user-patches.sh
            - name: config-files
              subPath: postfix-sasl-password.cf
              mountPath: /tmp/docker-mailserver/postfix-sasl-password.cf
              readOnly: true
            - name: config-files
              subPath: postfix-relaymap.cf
              mountPath: /tmp/docker-mailserver/postfix-relaymap.cf
              readOnly: true
            - name: config-files
              subPath: postfix-main.cf
              mountPath: /tmp/docker-mailserver/postfix-main.cf
              readOnly: true
            - name: config-files
              subPath: sender_canonical.cf
              mountPath: /tmp/docker-mailserver/sender_canonical.cf
              readOnly: true
            - name: config-files
              subPath: fetchmail.cf
              mountPath: /tmp/docker-mailserver/fetchmail.cf
              readOnly: true
            - name: config-files
              subPath: dovecot.cf
              mountPath: /tmp/docker-mailserver/dovecot.cf
              readOnly: true
            - name: config-files
              subPath: before.dovecot.sieve
              mountPath: /tmp/docker-mailserver/before.dovecot.sieve
              readOnly: true
            - name: config-files
              subPath: sa-learn
              mountPath: /etc/cron.d/sa-learn
              readOnly: true
#            - name: config-files
#              subPath: fts-xapian-plugin.conf
#              mountPath: /etc/dovecot/conf.d/10-plugin.conf
#              readOnly: true
            - name: config-files
              subPath: 10-replication.conf
              mountPath: /etc/dovecot/conf.d/10-replication.conf
              readOnly: false
#            - name: config-files
#              subPath: 10-mail.conf - overwriting does not work - getting sed: cannot rename /etc/dovecot/conf.d/sedc3XhGc: Device or resource busy error - putting it in dovecot.cf config
#              mountPath: /etc/dovecot/conf.d/10-mail.conf
#              readOnly: false
            - name: config-files
              subPath: dovecot-index
              mountPath: /etc/cron.d/dovecot-index
              readOnly: true
            - name: maildata
              mountPath: /var/mail
              subPath: mail-data
              readOnly: false
            - name: maildata
              mountPath: /var/mail-state
              subPath: mail-state
              readOnly: false
            - name: maildata
              mountPath: /var/log/mail
              subPath: mail-log
              readOnly: false
            - name: servercert
              mountPath: /etc/ssl/mail/server.pem
              subPath: server.pem
              readOnly: true
            - name: serverkey
              mountPath: /etc/ssl/mail/key.pem
              subPath: key.pem
              readOnly: true
            - name: cacert
              mountPath: /etc/ssl/mail/server-ca.pem
              subPath: server-ca.pem
              readOnly: true
```

### 📜 Relevant log output

_No response_

georglauterbach:
# Description

1. Adjusts the K8s documentation, which requires `allowPrivilegeEscalation: true` as it turns out
2. Adjusts permsissions on `/var/spool/postfix/{maildrop,public}`. This is more tricky, but #3619 provides good insights. In short, we do not seem to require SUID/SGID/sticky bits on these directories as the binaries that place files in these directories already have the proper SGID bits set on them.

Fixes #3619

CC @innotecsol

## Type of change

## Checklist:

### DIFF
--- a/docs/content/config/advanced/kubernetes.md
+++ b/docs/content/config/advanced/kubernetes.md
@@ -190,7 +190,7 @@ spec:
           imagePullPolicy: IfNotPresent
 
           securityContext:
-            allowPrivilegeEscalation: false
+            allowPrivilegeEscalation: true # required for SUID/SGID to work
             readOnlyRootFilesystem: false
             runAsUser: 0
             runAsGroup: 0
--- a/docs/content/config/advanced/kubernetes.md
+++ b/docs/content/config/advanced/kubernetes.md
@@ -190,7 +190,10 @@ spec:
           imagePullPolicy: IfNotPresent
 
           securityContext:
-            allowPrivilegeEscalation: true # required for SUID/SGID to work
+            # Required to support SGID via `postdrop` executable
+            # in `/var/mail-state` for Postfix (maildrop + public dirs):
+            # https://github.com/docker-mailserver/docker-mailserver/pull/3625
+            allowPrivilegeEscalation: true
             readOnlyRootFilesystem: false
             runAsUser: 0
             runAsGroup: 0
--- a/target/scripts/startup/setup.d/mail_state.sh
+++ b/target/scripts/startup/setup.d/mail_state.sh
@@ -105,10 +105,10 @@ function _setup_save_states() {
     # These two require the postdrop(103) group:
     chgrp -R postdrop "${STATEDIR}"/spool-postfix/{maildrop,public}
 
-    # After changing the group, special bits (set-gid, sticky) may be stripped, restore them:
-    # Ref: https://github.com/docker-mailserver/docker-mailserver/pull/3149#issuecomment-1454981309
-    chmod 1730 "${STATEDIR}/spool-postfix/maildrop"
-    chmod 2710 "${STATEDIR}/spool-postfix/public"
+    # These permissions rely on the `postdrop` binary (and alike) having proper SUID/SGID bits set.
+    # Ref: https://github.com/docker-mailserver/docker-mailserver/issues/3619
+    chmod 730 "${STATEDIR}/spool-postfix/maildrop"
+    chmod 710 "${STATEDIR}/spool-postfix/public"
   elif [[ ${ONE_DIR} -eq 1 ]]; then
     _log 'warn' "'ONE_DIR=1' but no volume was mounted to '${STATEDIR}'"
   else
--- a/target/scripts/startup/setup.d/mail_state.sh
+++ b/target/scripts/startup/setup.d/mail_state.sh
@@ -105,8 +105,8 @@ function _setup_save_states() {
     # These two require the postdrop(103) group:
     chgrp -R postdrop "${STATEDIR}"/spool-postfix/{maildrop,public}
 
-    # These permissions rely on the `postdrop` binary (and alike) having proper SUID/SGID bits set.
-    # Ref: https://github.com/docker-mailserver/docker-mailserver/issues/3619
+    # These permissions rely on the `postdrop` binary having the SGID bit set.
+    # Ref: https://github.com/docker-mailserver/docker-mailserver/pull/3625
     chmod 730 "${STATEDIR}/spool-postfix/maildrop"
     chmod 710 "${STATEDIR}/spool-postfix/public"
   elif [[ ${ONE_DIR} -eq 1 ]]; then


target/scripts/startup/setup.d/mail_state.sh
```diff
@@ -105,10 +105,10 @@ function _setup_save_states() {
...
     chgrp -R postdrop "${STATEDIR}"/spool-postfix/{maildrop,public}
 
-    # After changing the group, special bits (set-gid, sticky) may be stripped, restore them:
-    # Ref: https://github.com/docker-mailserver/docker-mailserver/pull/3149#issuecomment-1454981309
-    chmod 1730 "${STATEDIR}/spool-postfix/maildrop"
-    chmod 2710 "${STATEDIR}/spool-postfix/public"
+    # These permissions rely on the `postdrop` binary (and alike) having proper SUID/SGID bits set.
+    # Ref: https://github.com/docker-mailserver/docker-mailserver/issues/3619
+    chmod 730 "${STATEDIR}/spool-postfix/maildrop"
+    chmod 710 "${STATEDIR}/spool-postfix/public"
```
georglauterbach, line 111: If maintainers think we should remove this altogether and just use the Debian defaults, I do not mind it either. But remember: these files are most likely in a persisted volume, hence, if we removed these lines, users would be stuck with the approach on `master`. Keeping these lines here is safer for now; we could add a TODO comment about removing this in future versions.

docs/content/config/advanced/kubernetes.md
```diff
@@ -190,7 +190,7 @@ spec:
           imagePullPolicy: IfNotPresent
 
           securityContext:
-            allowPrivilegeEscalation: false
+            allowPrivilegeEscalation: true # required for SUID/SGID to work
```
polarathene: Since you look after the k8s docs, is this enough context to the user?

Would it be helpful to add extra context?:

```suggestion
            # Required to support SGID via `postdrop` executable
            # in `/var/mail-state` for Postfix (maildrop + public dirs):
            # https://github.com/docker-mailserver/docker-mailserver/pull/3625
            allowPrivilegeEscalation: true
```

`git blame` will handle that implicitly, so up to you.

---

## Original feedback exploring possibility of improvement

I think this setting might be a bit of a [red flag to some users](https://aws.github.io/aws-eks-best-practices/security/docs/pods/#do-not-allow-privileged-escalation) as a workaround?

The equivalent for Docker appears to be opt-in via [`--security-opt=no-new-privileges`](https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html#rule-4-add-no-new-privileges-flag), although on a Docker host with SELinux enabled, that or similar restriction may already be default? 🤷‍♂️ 
- On non-root user currently it seems to [clear capabilities by default](https://github.com/moby/moby/issues/45491#issuecomment-1540933674).
- Recently a change was made for upcoming v25 release of Docker, and backported to bugfix releases of v23 and v24: https://github.com/moby/moby/pull/45511#pullrequestreview-1421294603

---

Is this config of any help?:
- [`fsGroup`](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#configure-volume-permission-and-ownership-change-policy-for-pods) (_**NOTE:** Support [depends on volume driver](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#delegating-volume-permission-and-ownership-change-to-csi-driver)_)
  
  > - By default, Kubernetes recursively changes ownership and permissions for the contents of each volume to match the `fsGroup` specified in a Pod's `securityContext` when that volume is mounted.
  > - For large volumes, checking and changing ownership and permissions can take a lot of time, slowing Pod startup.
  > - You can use the `fsGroupChangePolicy` field inside a `securityContext` to control the way that Kubernetes checks and manages ownership and permissions for a volume.
- Paired with a [Volume Subpath](https://kubernetes.io/docs/concepts/storage/volumes/#using-subpath):
  > Sometimes, it is useful to share one volume for multiple uses in a single pod.
  > The `volumeMounts.subPath` property specifies a sub-path inside the referenced volume instead of its root. 

[Similar resource from IBM](https://community.ibm.com/community/user/storage/blogs/gero-schmidt1/2022/04/01/advanced-static-volume-provisioning-on-ocp):

> **fsGroup:** set as group ID/SGID on the mounted volume & added as supplemental group ID
> **Note:** The `fsGroup` option must be supported by the storage provider for the volume. All files and directories created in this directory inherit the `fsGroup` ID through the special SGID bit.
> 
> **supplementalGroups:** (additional) supplemental groups for shared group access; must be set in the `securityContext` on pod level
> 
> A note on **fsGroup:**
> - The behavior of applying `fsGroup` by setting the `fsGroup` and SGID bit (`drwxrwsrwx`) on the mounted directory depends on the type of volume and its support of ownership management.
> - It differs for volumes managed by Kubernetes like `EmptyDir {}` and CSI volumes managed by a CSI driver.
> - By default, Kubernetes recursively changes ownership and permissions for the contents of each volume to match the `fsGroup` specified in a pod's `securityContext` when the volume is mounted.

If the two features make any sense to use together, [this seems to provide a config example](https://github.com/kubernetes/kubernetes/issues/61283) if helpful.

---

**EDIT:** I guess there is no granular "capability" that covers this? 😅 

> [**Use Case #3: Special Permissions (SUID, SGID)**](https://hackernoon.com/whats-the-big-deal-with-linux-capabilities-part-2)
> 
> - **Set User ID** (`setuid`) and **Set Group ID** (`sgid`) are special permissions for executable files.
> - When these permissions are assigned to a file, the file to be executed assumes the privileges of the file's owner or group.
> - `setuid` bit changes a program effective `uid` (`euid`) upon execution.

Actually, you seem to already have applied that via `SETGID`? capability below?

[`CAP_SETGID`](https://man7.org/linux/man-pages/man2/setgid.2.html):

> `setgid()` sets the effective group ID of the calling process.
> If the calling process is privileged (_more precisely: has the `CAP_SETGID` capability in its user namespace_), the real GID and saved `set-group-ID` are also set.

So was `allowPrivilegeEscalation: false` overriding allowing that capability? ~~Or was it due to another process running instead of the root user / entrypoint?~~ (_**EDIT:** [Overriding `SETGID` capability](https://github.com/docker-mailserver/docker-mailserver/issues/3619#issuecomment-1794668094)_)

polarathene: ## Follow-up adjustment comment regarding vol subpath + fsGroup

Oh I see below that a volume subpath is already used. I was wondering if it'd be good for the `maildrop` + `public` dirs if the `fsGroup` feature caters to that with `allowPrivilegeEscalation: false`?

Only issue then is the GID won't be static (deterministic) unless we create it before installing the `postfix` package (_like with ClamAV_) I think?

polarathene (changes requested):
- 2 suggestions
- 1 question: What changed since the original #3149 PR to motivate this for you?

---

I've heavily documented the topic in review comments for my own benefit while recapping on the prior PR. Should be helpful towards addressing: https://github.com/docker-mailserver/docker-mailserver/issues/3557#issuecomment-1741708401

target/scripts/startup/setup.d/mail_state.sh
georglauterbach: > **Question:** What changed since the original #3149 PR to motivate this for you?:
> 
> * [You commented recently](https://github.com/docker-mailserver/docker-mailserver/pull/3149#issuecomment-1798683390) there that `allowPrivilegeEscalation: false` caused the issue for you.
>       
>   * But that PR itself fixed the same problem you encountered correct?

That's a good point, indeed.

>   * What is different now with the report in [docs: Kubernetes requies `allowPrivilegeEscalation: true` #3619](https://github.com/docker-mailserver/docker-mailserver/issues/3619) ? Why did the existing approach not work for that k8s user like it had for you?

I cannot answer this question TBH, because I don't know the answer 😆 I can only state that the setup with this PR makes way more sense. I do not know why I stopped seeing this error, I can only guess: at some point, I started using a completely custom Postfix configuration - maybe this is the culprit.

> > If maintainers think we should remove this altogether and just use the Debian defaults, I do not mind it either.
> > Keeping these lines here is safer for now; we could add a TODO comment about removing this in future versions.
> 
> You can't drop this to rely on Debian defaults.
> 
> They're not carried over into a fresh volume mount for these locations IIRC. That was the whole point we explicitly added them 😅

👍🏼 

> > ```shell
> > # These permissions rely on the `postdrop` binary (and alike) having proper SUID/SGID bits set.
> > ```
> 
> * Doesn't look like there is any SUID bit involved at all?
>       
>   * Only `setgid` / SGID on the `postdrop` binary.
>   * Possibly also `postqueue` (_it has SGID set too_), but I'm not sure about it's involvement with either dir.

Indeed, it's only SGID.

> * I agree that these extra special bits (_Sticky bit on `maildrop/` and SGID on `public/`_) don't seem to be necessary, at least for DMS (_dropping them will however diverge from Debian and be inconsistent vs no volume mount for state_).
> 
> 
> **NOTE:** The sticky-bit on `maildrop/` was likely historical from the [previous PR referenced Postfix release notes](https://github.com/vdukhovni/postfix/blob/1f2df46e5211065a16f78739308131122e185c7c/postfix/RELEASE_NOTES-1.1#L32-L34). Debian just may never have adjusted to that on their end, or it may be for supporting third-party software like `mailx` writing to `maildrop/`

I am pretty sure too that the sticky bit is not required :)

polarathene (approved):

georglauterbach:
Not a single issue with these changes on my production machine so far.

georglauterbach:
All good on my machine; I will go ahead and merge this. 2 PRs remaining for v13 then,

PR #3629, opened 2023-11-08
Fix rspamd directory

jsonn:
The directory on disk is written in all-lowercase, so documentation should match.


georglauterbach (approved):
Agreed! Good catch 🚀

PR #3647, opened 2023-11-21
logs: improve format

georglauterbach:
# Description

This PR updates `log.sh` to add timestamps to all messages. Review commit by commit. The date command format string was taken [from here](https://unix.stackexchange.com/a/629504) where I searched for ISO8601 compliance.

Old format:

```txt
[  ERROR  ]  ndawkggdryg gydg ygdrrjdwjk
[ WARNING ]  adwdad ad  awdadawdaw
[   INF   ]  ndawkjdwjk
[  DEBUG  ]  da dadwadwda d aw
[  TRACE  ]  ndaw d adw da dd d dawkjdwjk
[  TRACE  ]  hdxhfdhfth fshfebhsjbf eh
[  DEBUG  ]  djaksdajdna nadjakw dakd awjkn
[  DEBUG  ]  danjdadaw
[  TRACE  ]  sads fshfebhsjbf eh
[  TRACE  ]  dgdht adwd eh
[  TRACE  ]  dadada  dadwddwadadwdaebhsjbf eh
[   INF   ]  dac dnajkd and akd jakwnd wjd nakjndwjdnak dwnjdajk
```

New format:

```txt
2023-11-21T16:40:14+01:00  ERROR  ndawkggdryg gydg ygdrrjdwjk
2023-11-21T16:40:14+01:00  WARN   adwdad ad  awdadawdaw
2023-11-21T16:40:14+01:00  INFO   ndawkjdwjk
2023-11-21T16:40:14+01:00  DEBUG  da dadwadwda d aw
2023-11-21T16:40:14+01:00  TRACE  ndaw d adw da dd d dawkjdwjk
2023-11-21T16:40:14+01:00  TRACE  hdxhfdhfth fshfebhsjbf eh
2023-11-21T16:40:14+01:00  DEBUG  djaksdajdna nadjakw dakd awjkn
2023-11-21T16:40:14+01:00  DEBUG  danjdadaw
2023-11-21T16:40:14+01:00  TRACE  sads fshfebhsjbf eh
2023-11-21T16:40:14+01:00  TRACE  dgdht adwd eh
2023-11-21T16:40:14+01:00  TRACE  dadada  dadwddwadadwdaebhsjbf eh
2023-11-21T16:40:14+01:00  INFO   dac dnajkd and akd jakwnd wjd nakjndwjdnak dwnjdajk
```

## Type of change

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,22 @@ clean: ALWAYS_RUN
 # --- Tests  ------------------------------------
 # -----------------------------------------------
 
+run-local-test: ALWAYS_RUN
+	docker run --rm -it \
+		--env OVERRIDE_HOSTNAME=mail.example.test \
+		--env LOG_LEVEL=trace \
+		--env POSTFIX_INET_PROTOCOLS=ipv4 \
+		--env DOVECOT_INET_PROTOCOLS=ipv4 \
+		--env ENABLE_RSPAMD=1 \
+		--env ENABLE_AMAVIS=0 \
+		--env ENABLE_CLAMAV=0 \
+		--env ENABLE_FAIL2BAN=0 \
+		--env ENABLE_OPENDMARC=0 \
+		--env ENABLE_OPENDKIM=0 \
+		--env ENABLE_POLICYD_SPF=0 \
+		--env TZ=Europe/Berlin \
+		mailserver-testing:ci
+
 tests: ALWAYS_RUN
 # See https://github.com/docker-mailserver/docker-mailserver/pull/2857#issuecomment-1312724303
 # on why `generate-accounts` is run before each set (TODO/FIXME)
--- a/target/scripts/helpers/log.sh
+++ b/target/scripts/helpers/log.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# shellcheck disable=SC2291 # Quote repeated spaces to avoid them collapsing into one.
+# shellcheck disable=SC2291 # Quote repeated SPACE to avoid them collapsing into one.
 # shellcheck disable=SC2034 # VAR appears unused.
 
 # Color variables for global usage
@@ -27,7 +27,6 @@ RESET=$(echo -ne   '\e[0m')
 # ### DMS Logging Functionality
 #
 # This function provides the logging for scripts used by DMS.
-# It adheres to the convention for log levels.
 # Valid values (in order of increasing verbosity) are: `error`,
 # `warn`, `info`, `debug` and `trace`. The default log level
 # is `info`.
@@ -41,20 +40,19 @@ RESET=$(echo -ne   '\e[0m')
 #
 # If the first argument is not set or invalid, an error
 # message is logged. Likewise when the second argument
-# is missing. Both failures will return with exit code '1'.
+# is missing. Both failures will result in exit code '1'.
 function _log() {
-  if [[ -z ${1+set} ]]; then
+  if [[ ! -v 1 ]]; then
     _log 'error' "Call to '_log' is missing a valid log level"
     return 1
   fi
 
-  if [[ -z ${2+set} ]]; then
+  if [[ ! -v 2 ]]; then
     _log 'error' "Call to '_log' is missing a message to log"
     return 1
   fi
 
-  local LEVEL_AS_INT
-  local MESSAGE="${RESET}["
+  local LEVEL_AS_INT LEVEL_STRING_WITH_COLOR SPACE MESSAGE
 
   case "$(_get_log_level_or_default)" in
     ( 'trace' ) LEVEL_AS_INT=5 ;;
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -40,6 +40,22 @@ clean: ALWAYS_RUN
...
+		--env ENABLE_RSPAMD=1 \
+		--env ENABLE_AMAVIS=0 \
+		--env ENABLE_CLAMAV=0 \
+		--env ENABLE_FAIL2BAN=0 \
+		--env ENABLE_OPENDMARC=0 \
+		--env ENABLE_OPENDKIM=0 \
+		--env ENABLE_POLICYD_SPF=0 \
+		--env TZ=Europe/Berlin \
+		mailserver-testing:ci
+
```
georglauterbach, line 58: Not sure if we want to leave this in this PR, but I frequently used this command during testing..

polarathene: The `TZ` is rather specific, as is the choice to use rspamd (_which eventually becomes redundant once it's default_).

Other than that, since it's not used for running actual test suite it's more of a local dev instance and doesn't really belong in the tests section here?

Most of the env you could provide yourself via `--env-file` arg? Or since this is tailored for you, perhaps an alias on your end works well too?

---

It's probably best to only add this if it meets the needs of others as well, unless `--env-file` and/or shell alias doesn't work well for your needs?

target/scripts/helpers/log.sh
```diff
@@ -67,35 +65,37 @@ function _log() {
...
 
     ( * )
       _log 'error' "Call to '_log' with invalid log level argument '${1}'"
       return 1
       ;;
   esac
 
-  MESSAGE+="${RESET}]  ${2}"
+  LEVEL_STRING_WITH_COLOR+="${RESET}"
+  MESSAGE="$(date +'%Y-%m-%dT%H:%M:%S%:z')  ${LEVEL_STRING_WITH_COLOR}  ${2}"
```
polarathene, line 98: Any reason for including date in all logs?

I think we only did this for change detector for the benefit of tests?

Some deployments however likely have logs managed in a way that timestamps are already part of the logger? So this would just double up on the date and increase disk/network usage? 🤷‍♂️ 

If your deployment logs don't handle this but you require all logs to have a timestamp this would be something Vector should be good for via configuration. There is progress on that front with the syslog sink PR now out of WIP status and waiting on review. Vector might be viable for DMS v14 🙏

polarathene: You've also changed the logging format here to swap log level and date, which does make sense, but if anyone is parsing these logs they'll need to update their logic/regex.

---

Oh and you changed the date format too 😝 

ISO8601 / RFC3339 (more constrained variant) is good. Syslog v1 format also uses RFC3339 format for the timestamp but with further constraints IIRC.

Still, this is probably something better deferred/configured by proper log management. If we're planning to adopt Vector, this change would be short-lived and perhaps introduce an inconvenience for users to adapt to briefly? (_assuming a release with Vector added isn't that far off after a release with this change_)

polarathene (commented):
I was taken straight to the changes and reviewed first before remembering to look at the actual PR description 😅 

I only have feedback with some concerns. Summary below:

- `Makefile` addition doesn't seem necessary unless others would like to see it. Additionally, as it's not related to running the test suite, thus should not give that impression to someone reading the `Makefile`.
- I'd like more motivation for the log change, especially since not only is the order of fields and timestamp format adjusted, but you've decided all logs should include the timestamp.
  - I'm concerned that will double up timestamps redundantly with production deployments that use loggers that already include timestamps.
  - They probably also parse the log message for the log level to transform that, but we may be introducing this in one release followed by Vector changing it again in the next?

target/scripts/helpers/log.sh
casperklein: > Oh and you changed the date format too 😝

I noticed the same. At the moment, there are already two log formats used:

```
2023-11-21 17:49:55,021 INFO spawned: 'amavis' with pid 2729
2023-11-21 17:49:55,021 INFO success: amavis entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
[   INF   ]  mail.example.test is up and running
Nov 21 17:49:55 mail amavis[2729]: No ext program for   .zoo, tried: zoo
Nov 21 17:49:55 mail amavis[2729]: No ext program for   .doc, tried: ripole
```

`2023-11-21 17:49:55,021` is from supervisord and `Nov 21 17:49:55` is from `mail.log`.

There shouldn't be introduced a third variant. I would go with the `mail.log` format.

polarathene: `mail.log` is using BSD-syslog IIRC (RFC 3164) and is obsoleted by (RFC 5424) that standardizes syslog format including the timestamp (_to RFC3339 with additional constraints_).

Not sure about `supervisord` format.

Once Vector is added into DMS, users (or test suite) can configure logging much more easily to their needs. So if the motivation for this change was related to an actual need I'd suggest deferring until the switch to Vector is ready.

georglauterbach:
@polarathene so if I understand your concerns correctly, Vector will take care of adding timestamps to our logs as well? So we would not have two (three with this PR) log formats but only one? If this is the case, we can close this PR. I thought vector was only about collecting logs :D

polarathene:
> Vector will take care of adding timestamps to our logs as well?
> So we would not have two (three with this PR) log formats but only one?

Yes, it'll handle that no problem.

Once the syslog sink feature lands and a new release of Vector is available, it should be able to replace rsyslog AFAIK. That will allow it to continue to output syslog format that should be compatible with how logs are managed presently with DMS.

Vector can continue to output some logs to files, along with the supervisord redirection of stdout / stderr to log files, which Vector can be configured to ingest too to become a more central logging system.

Our own internal logging in scripts can use the `logger` command which can be captured by Vector, and from that we can parse log level and have outputs that forward logs with metadata like log level and timestamps to other logging solutions, while being able to enable writing logs to a file / stdout as our current log function does.

---

>  I thought vector was only about collecting logs :D

Ingest log from input (stdout / file / syslog / etc), optionally parse/transform, output logs to sink(s).

Should be easier for users to provide their own configs too than to tamper with current logging in DMS.

georglauterbach:
I'll close this as Vector will take care of this then; I thought Vector would not 🚀

PR #3659, opened 2023-11-26
bug fix: remove `exit 0` from `update-check.sh`

vincentDcmps:
# Description

since release 13.0.0 commit my mailbox is flood by update mail.  I think when script is exit supervisord launch it again

## Type of change

## Checklist:

- [X] If necessary I have added tests that prove my fix is effective or that my feature works
- [X] **I have added information about changes made in this PR to `CHANGELOG.md`**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ All notable changes to this project will be documented in this file. The format 
 
 > **Note**: Changes and additions listed here are contained in the `:edge` image tag. These changes may not be as stable as released changes.
 
+- fixe check update mail loop when update is availlable
+
+
 ## [v13.0.0](https://github.com/docker-mailserver/docker-mailserver/releases/tag/v13.0.0)
 
 ### Breaking
--- a/target/scripts/update-check.sh
+++ b/target/scripts/update-check.sh
@@ -40,7 +40,7 @@ EOF
       _log_with_date 'info' "Update available [ ${VERSION} --> ${LATEST} ]"
 
       # only notify once
-      echo "${MAIL}" | mail -s "Mailserver update available! [ ${VERSION} --> ${LATEST} ]" "${POSTMASTER_ADDRESS}" && exit 0
+      echo "${MAIL}" | mail -s "Mailserver update available! [ ${VERSION} --> ${LATEST} ]" "${POSTMASTER_ADDRESS}"
     else
       _log_with_date 'info' 'No update available'
     fi


georglauterbach:
Hmmm.. Looking at the Supervisor configuration

https://github.com/docker-mailserver/docker-mailserver/blob/68e5c3426e033de2537aa0a646f2bfcbe63e4fd0/target/supervisor/conf.d/supervisor-app.conf#L153-L159

I do not think the check will run again.. The [documentation for Supervisor](http://supervisord.org/configuration.html) says the default is `autorestart: unexpected`. The `exit 0` should indicate a normal exit. @casperklein can you verify?

vincentDcmps:
edge is not update because cron is executed only fifth day of week

.github/workflows/scheduled_builds.yml
```shell
name: 'Deploy :edge on Schedule'

on:
  workflow_dispatch:
  schedule:
    - cron: 0 0 * * 5

``` 

I'am agree with you if I read supervisord docs my issue doesn't have to happen
the issue was that  I have received 5 mail by minute to inform that update is availlable, until to switch to latest tag with 13.0 version pushed.

georglauterbach:
I think the actual issue is here:

https://github.com/docker-mailserver/docker-mailserver/blob/2c602299136fcac8a8fd02ada86039687e6ae18c/.github/workflows/generic_publish.yml#L24-L35

My suggestions:

```diff
-             type=edge,branch=master
+             type=edge
```

I guess the issue is that `branch=master` is not identical to `branch=$repo.default_branch` (see [the action's docs](https://github.com/docker/metadata-action?tab=readme-ov-file#typeedge)). @polarathene can you check this for me?

georglauterbach:
Superseeded by #3662

@vincentDcmps Thank you for making us aware of this:) I have manually started a run on `master`, so `:edge` will be updated within the next half an hour and this problem should vanish.

polarathene:
@georglauterbach

Below is for reference, no need to read and respond 👍 

---

> My suggestions:
> 
> ```diff
> -             type=edge,branch=master
> +             type=edge
> ```

For reference, no that would be bad.

The logic here is that either the workflow was triggered by:
- A git tag for a release (_create multiple image tags based on semver_)
- No tag, just a push event to master branch (_tags aren't associated to branches, but commits_). This qualifies for `:edge`.

Your suggestion would publish `:edge` regardless of event.
- If the workflow was ever triggered to run on a different branch for example (_sometimes created for PRs by maintainers, or CI scripts like `CONTRIBUTORS.md` updates_), that'd accidentally attempt publishing a release.
- It shouldn't happen however, since we call this workflow in a controlled manner via `workflow_call` with explicit reference to the latest iteration of the workflow on master branch. But pushes to the master branch isn't the only trigger event, we have `workflow_dispatch` (manually invoked) and [scheduled workflow](https://github.com/docker-mailserver/docker-mailserver/blob/2c602299136fcac8a8fd02ada86039687e6ae18c/.github/workflows/scheduled_builds.yml). Best to keep safeguards in place 😅 

Multiple semver tags are useful for [services like `watchtower` to leverage](https://docker-mailserver.github.io/docker-mailserver/v13.0/config/advanced/maintenance/update-and-cleanup/), when DMS updates an existing tag these services notify, but otherwise unaware of newer tags as they don't actually understand semver (_some update services do_).

---

> But to me, the question remains: why didn't the default push (by merging the PR) trigger the action to build `:edge`?
> 
> UPDATE: Ah, now I get it:

To be fair, it's not a change relevant to `:edge` to warrant building the image again and publishing it.
- It's a bug with the current behaviour for `:edge`, it [should not check `VERSION`](https://github.com/docker-mailserver/docker-mailserver/pull/3662#pullrequestreview-1749456998) for update notifications.
- `:edge` users should use `watchtower` or configure compose (or similar) to always pull the latest `:edge` tag if they want to always run the current `:edge`.

polarathene:
@vincentDcmps 

Below is probably not important to you anymore, but if you're interested in better identifying / understanding the actual cause of the problem, give it a read.

---

> the issue was that I have received 5 mail by minute to inform me that update is availlable

May be something related to your configuration, you've not provided much context. `:edge` shouldn't behave any differently than a published release presently.

They all just check their internal copy of `VERSION` and do a request to compare it to the one on Github, which if found to be newer will send you one mail about the update once per `UPDATE_CHECK_INTERVAL`:

https://github.com/docker-mailserver/docker-mailserver/blob/b663e10841d7d95f25d51efb7696c0c5b588d55f/target/scripts/update-check.sh#L40-L45

For example I had a very simple container running, and it received the following log when the update was checked:

```console
$ docker run --rm -it --hostname mail.example.test --name dms mailserver/docker-mailserver

# Update check log:
$ docker exec dms cat /var/log/supervisor/update-check.log
[   INF   ]  2023-11-25 04:51:31  No update available
[   INF   ]  2023-11-26 13:51:04  Update available [ 12.1.0 --> 13.0.0 ]

# Default internal alias for local user accounts that belong to DMS hostname (`@mail.example.test`)
# Mail sent to root locally will be redirected to `postmaster@example.test`
$ docker exec dms cat /etc/aliases
root: postmaster@example.test
```

<details>
<summary>Full log for update-check script at 1.51PM Nov 26th</summary>
```
Nov 26 13:51:04 mail postfix/pickup[123308]: F16DF483E: uid=0 from=<root>
Nov 26 13:51:04 mail postfix/cleanup[130419]: F16DF483E: message-id=<20231126135104.F16DF483E@mail.example.test>
Nov 26 13:51:04 mail opendkim[439]: F16DF483E: localhost [127.0.0.1] not internal
Nov 26 13:51:04 mail opendkim[439]: F16DF483E: not authenticated
Nov 26 13:51:04 mail opendkim[439]: F16DF483E: no signature data
Nov 26 13:51:04 mail postfix/qmgr[570]: F16DF483E: from=<root@mail.example.test>, size=731, nrcpt=1 (queue active)
Nov 26 13:51:05 mail dovecot: master: Warning: Time moved forwards by 843.682211 seconds - adjusting timeouts.
Nov 26 13:51:05 mail dovecot: lmtp(130422): Connect from local
Nov 26 13:51:05 mail dovecot: auth: passwd-file(postmaster@example.test): unknown user
Nov 26 13:51:05 mail postfix/lmtp[130421]: F16DF483E: to=<postmaster@example.test>, relay=mail.example.test[/var/run/dovecot/lmtp], delay=0.06, delays=0.02/0.01/0.02/0.01, dsn=5.1.1, status=bounced (host mail.example.test[/var/run/dovecot/lmtp] said: 550 5.1.1 <postmaster@example.test> User doesn't exist: postmaster@example.test (in reply to RCPT TO command))
Nov 26 13:51:05 mail dovecot: lmtp(130422): Disconnect from local: Logged out (state=READY)
Nov 26 13:51:05 mail postfix/cleanup[130419]: 0B214484C: message-id=<20231126135105.0B214484C@mail.example.test>
Nov 26 13:51:05 mail postfix/bounce[130424]: F16DF483E: sender non-delivery notification: 0B214484C
Nov 26 13:51:05 mail postfix/qmgr[570]: 0B214484C: from=<>, size=2864, nrcpt=1 (queue active)
Nov 26 13:51:05 mail postfix/qmgr[570]: F16DF483E: removed
Nov 26 13:51:05 mail postfix/cleanup[130419]: 0C62B483E: message-id=<20231126135105.0B214484C@mail.example.test>
Nov 26 13:51:05 mail postfix/local[130425]: 0B214484C: to=<root@mail.example.test>, relay=local, delay=0.01, delays=0/0/0/0, dsn=2.0.0, status=sent (forwarded as 0C62B483E)
Nov 26 13:51:05 mail postfix/qmgr[570]: 0C62B483E: from=<>, size=2998, nrcpt=1 (queue active)
Nov 26 13:51:05 mail postfix/qmgr[570]: 0B214484C: removed
Nov 26 13:51:05 mail dovecot: lmtp(130422): Connect from local
Nov 26 13:51:05 mail dovecot: auth: passwd-file(postmaster@example.test): unknown user
Nov 26 13:51:05 mail postfix/lmtp[130421]: 0C62B483E: to=<postmaster@example.test>, orig_to=<root@mail.example.test>, relay=mail.example.test[/var/run/dovecot/lmtp], delay=0, delays=0/0/0/0, dsn=5.1.1, status=bounced (host mail.example.test[/var/run/dovecot/lmtp] said: 550 5.1.1 <postmaster@example.test> User doesn't exist: postmaster@example.test (in reply to RCPT TO command))
Nov 26 13:51:05 mail dovecot: lmtp(130422): Disconnect from local: Logged out (state=READY)
Nov 26 13:51:05 mail postfix/qmgr[570]: 0C62B483E: removed
```

</details>

In the above logs, the notification mail (ID `F16DF483E`) is sent from `root@mail.example.test` (DMS hostname) to the same address which by default DMS aliases to the assumed default postmaster account (`postmaster@example.test`).

However that is not a valid user to deliver mail to in this case so it bounces. In this case, the bounce is sent back to the sender ( `root@mail.example.test`), but since that loops back to `postmaster@example.test` again it bounces and Postfix knows to stop at double bounce, thus it doesn't bother trying to deliver the original notification to the intended recipient again, nor the bounce to the original sender.

---

If your `UPDATE_CHECK_INTERVAL` ENV is the default value of `1d`, you could provide more context, such as what the `update-check.log` shows (_if it was actually running more frequently than expected, thus a fault of the script or `supervisord`_), or if it was on the Postfix side (sounds more likely) with mail delivery.

It might just be that the update notification was sent to `postmaster@example.test` like above demonstrates, but that bounced, yet actually delivered the bounce to a `root@mail.example.test` account, so attempted to send the mail back to `postmaster@example.test` and repeated this failure?

## TL;DR

- Check your `UPDATE_CHECK_INTERVAL` ENV is the default value of `1d`
- Confirm if the script or `supervisord` are actually at fault: `docker exec dms cat /var/log/supervisor/update-check.log`
- Check `docker exec dms cat /etc/aliases` or related alias/ENV for the configured Postmaster account, that it actually exists. It's possible that Postfix repeatedly tried to deliver a mail due to some failure there.

casperklein:
The first update check is done on container start and then repeated infinite with the configured interval. If an update is found, the update-check exits (and won't be started by supervisor again). Your PR would effectively introduce notification on each interval when an update is found.
Did your container for some reason restarted frequently? That would explain your issue.

PR #3641, opened 2023-11-17
release: v13.0.0

georglauterbach:
# Description

This PR marks the feature freeze phase for the `v13.0.0` release.

The current `:edge` image will be tested for roughly **7 days** and if no bugs get reported, `v13.0.0` will be released. During this testing period, only PRs that do not change functionality will be merged.

**If you want to support us, give the latest `:edge` image a try and report any issues you encounter back to this PR.**

Documentation updates may be merged. **Feature freeze is in place until 24 Nov 2023**.

---

@polarathene @casperklein please apply suggestions immediately to the branch, no need to ask me if you want to change stuff up; already spent too much time on `CHANGELOG.md` and drafting the release

## Type of change

- [x] Release

## Release-Checklist:

- [x] Update CHANGELOG.md
- [x] Update version in VERSION
- [ ] Draft release (still needs proper text description for changes)

Issue #2952, opened 2022-12-23
[BUG] v11.3.1 was published with update check against the wrong version (v11.3.0)


Cl00e9ment:
### What happened and when does this occur?

`/VERSION` contains `11.3.0` in the image with the `:11.3.1` tag.

### What did you expect to happen?

`/VERSION` to contain `11.3.1`

### How do we replicate the issue?

```bash
# Outputs: `11.3.0`
docker run -it --rm --entrypoint cat mailserver/docker-mailserver:11.3.1 VERSION
```

### DMS version

11.3.1

### Improvements to this form?

My bug report can be summarized in 2 lines. The bug is OS agnostic and doesn't even require any service to be ran. This whole form is nonsense for this particular case. You should allow to bypass it.

vedranmiletic:
Can confirm.

polarathene:
Thanks for the bug report!

We are aware of this issue but it was deemed not likely to cause any real concerns and thus decided to leave as-is (_rather than mess with our git history to adjust the tagged commit or similar inconvenient workarounds_): https://github.com/docker-mailserver/docker-mailserver/pull/2944#issuecomment-1363421102

Closing as resolved (_unless there's an actual problem requiring a 11.3.2 release_).

----

> ### Improvements to this form?
> 
> My bug report can be summarized in 2 lines. The bug is OS agnostic and doesn't even require any service to be ran. This whole form is nonsense for this particular case. You should allow to bypass it.

In future you can raise an issue with the choice **Question / Other**. This will allow you to have control, we only ask you to use the bug report form when requiring support / troubleshooting.

Cl00e9ment:
The only problem that I can see is that docker-mailserver sends a mail with the subject `Mailserver update available! [ 11.3.0 --> 11.3.1 ]` even if it's false. I don't know if it sends it regularly (every day for example). If this is the case, it's a small annoyance/spam.

casperklein:
The update notification is only send once. However, if you restart DMS, you'll get notified again.

polarathene:
Keeping this open and pinning for awareness.

There has been discussion on resolving it, but it looks like due to recent changes with our CI it'll risk introducing a breaking change (platform support) among some other concerns to produce a release this way that we'd rather avoid.

[...]

PR #3662, opened 2023-11-26
bug fix: push `:edge` when `VERSION` is updated as well

georglauterbach:
# Description

Previously, we did not run the workflow on push on `master` when a release happened because the push on master is guarded by a check on which files were changed.

With this change, I added `VERSION` to the list of files to consider when updating `:edge`.

Superseeds #3659

## Type of change

## Checklist:

### DIFF
--- a/.github/workflows/default_on_push.yml
+++ b/.github/workflows/default_on_push.yml
@@ -11,6 +11,7 @@ on:
       - .gitmodules
       - Dockerfile
       - setup.sh
+      - VERSION # also update :edge when a release happens
     tags:
       - '*.*.*'
 


casperklein (approved):
BTW: would it hurt, to just build on all changes regardless of the path?

georglauterbach:
> BTW: would it hurt, to just build on all changes regardless of the path?

Probably not; this is what I was suggesting in https://github.com/docker-mailserver/docker-mailserver/pull/3659#issuecomment-1826887742 too.

georglauterbach:
UPDATE: I guess this PR solved the issue quite nicely; we should actually not remove `paths:` to not unnecessarily build an image (e.g., when the documentation is updated). @casperklein

.github/workflows/default_on_push.yml
```diff
@@ -11,6 +11,7 @@ on:
       - .gitmodules
       - Dockerfile
       - setup.sh
+      - VERSION # also update :edge when a release happens
```
polarathene, line 14: The comment is unnecessary:

```suggestion
      - VERSION
```

`VERSION` contributes no meaning to `:edge`. In the past we have once [missed bumping `VERSION` before tagging a release](https://github.com/docker-mailserver/docker-mailserver/issues/2952#issuecomment-1364304849), but since we [only publish images other than `:edge` by tagged commits](https://github.com/docker-mailserver/docker-mailserver/pull/2944#issuecomment-1364361134), we just published a new tag semver-patch release to resolve that.

I'd rather see the [`VERSION` file removed since it's unnecessary to support the functionality it's intended for](https://github.com/docker-mailserver/docker-mailserver/pull/3641#issuecomment-1826483637), but since it the [script in all prior releases since v10](https://github.com/docker-mailserver/docker-mailserver/pull/1951) will rely on checking `VERSION` we'd need to keep it around for a while 🤷‍♂️ 

---

If the better update check approach is taken, then the addition to build an image when `VERSION` changes would become redundant:

```suggestion
```

polarathene (commented):
We have a better fix IMO, `:edge` shouldn't care about this at all.

PR #3663, opened 2023-11-26
ci: Add `run-local-instance` target to `Makefile`

georglauterbach:
# Description

This adds a target to `Makefile` which runs a simple version of DMS (anti-spam services disabled) to test local builds. I use this frequently.

Fixes #

## Type of change

## Checklist:


Makefile
```diff
@@ -36,6 +36,22 @@ clean: ALWAYS_RUN
...
+		--env DOVECOT_INET_PROTOCOLS=ipv4 \
+		--env ENABLE_CLAMAV=0 \
+		--env ENABLE_AMAVIS=0 \
+		--env ENABLE_RSPAMD=0 \
+		--env ENABLE_OPENDKIM=0 \
+		--env ENABLE_OPENDMARC=0 \
+		--env ENABLE_POLICYD_SPF=0 \
+		--env ENABLE_SPAMASSASSIN=0 \
+		--env LOG_LEVEL=trace \
+		$(IMAGE_NAME)
```
polarathene, line 53: I'm still not a big fan of this in the `Makefile`.

You could just as easily have a minimal variant of our [`compose.yaml`](https://github.com/docker-mailserver/docker-mailserver/blob/68a43eb4970f2ab7680ffa470a1d26e37fa375f0/compose.yaml) example which:
- References a different `.env` file with `env_file` (_which a user can override to point to their own with the [`--env-file` option](https://docs.docker.com/compose/environment-variables/envvars-precedence/)_).
- The `${IMAGE_NAME}` should still be [supported with alternative fallback](https://docs.docker.com/compose/environment-variables/env-file/#parameter-expansion) with `image: ${IMAGE_NAME:-ghcr.io/docker-mailserver/docker-mailserver:latest}`.
- The need to make an account would also be avoided I think without worrying about volume mounting (_as read-only, or making a mutable copy like tests do_) by leveraging the [compose feature `configs`](https://docs.docker.com/compose/compose-file/08-configs/):

    ```yml
    configs:
      dms_accounts:
        content: |
        # user: "user1@example.test"
        # password: "mypassword"
        user1@example.test|{SHA512-CRYPT}$6$DBEbjh4I9P7aROk8$XosqE.YI2Z4bUkWD1/bedrSNpw79nsO60yiAKk04jARhPVX5VD/SaVM5HWFDQyzftESVDjbVdhzn/d4TJxFwg0
    ```

I don't know if we have anyone actually adjusting `IMAGE_NAME` tbh? So if you did want to take the compose approach, you could probably just create `compose.dev.yaml` with `image: mailserver-testing:ci` 🤷‍♂️ 

Then it's just `docker compose -f compose.dev.yaml up` or if you prefer, an additional `make run-local-instance` as a alias/proxy command if that is a little easier/friendly.

You'd then have far more flexibility for yourself and users due to all the extra compose features.
- Something the next refactor of the test suite will use.
- I actually have something very much like this locally that I could polish up, using the test certs and a local DNS service  (CoreDNS). I still want to integrate `octodns` for a better way to manage the DNS config, that recently had a release with all the fixes + features I've been waiting on.

I have for quite some time wanted to contributed something like what I've described as an example offline setup/playground. Along with the convenience you're aiming for resolving in this PR.

polarathene (approved):
LGTM 👍 

---

If you would like to take the `compose.yaml` approach, I can put together a PR for that. The more full-fledged offline setup would have to wait, but adjusting your current approach to `compose.yaml` is simple enough as a few files in an `example/` or similar folder.

`run-local-instance` might then be better as `run-example`. My future plans would be for having several common examples, which should be helpful to the community and reproductions for issues, and Docker Compose features really help simplify that 😎

Makefile
```diff
@@ -36,6 +36,22 @@ clean: ALWAYS_RUN
 	-@ while read -r LINE; do [[ $${LINE} =~ test/.+ ]] && FILES+=("/mnt$${LINE#test}"); done < .gitignore ; \
 		docker run --rm -v "$(REPOSITORY_ROOT)/test/:/mnt" alpine ash -c "rm -rf $${FILES[@]}"
 
+run-local-instance: ALWAYS_RUN
+	bash -c 'sleep 8 ; ./setup.sh email add postmaster@example.test 123' &
```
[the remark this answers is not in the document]
georglauterbach: Yeah, we should provide something like that :D But not with this PR. I will take care of this later.

Makefile
georglauterbach: I like where this is going. If you have the time, please provide your idea with `compose.yaml` :D

polarathene: It'll need to wait, the [`content` property for `configs` is very new (mid november)](https://github.com/compose-spec/compose-spec/pull/446).

I'm on Windows and relying on Docker Desktop with WSL2 (_which seems to regress VMware guests to unpleasant UX_). Otherwise the feature is usable with Docker Compose v2.23.1.

---

The `$` in the content also gets treated as variable for interpolation feature, so needs to be escaped as `$$`.

polarathene: Fine by me, you might want to wait on input from @casperklein

casperklein: Will review in the next 2 days.

casperklein (approved):

georglauterbach:
FYI: Auto-merge is enabled

polarathene (approved):

Issue #3630, opened 2023-11-08
[TODO]: Consider revising DKIM generation


polarathene:
### Subject

Something else that requires developers attention

### Description

Just adding another tracking issue publicly so the information is easy to locate for others 👍 

- We could generate the keypair with `openssl`, although I find the UX of `step` (Go-lang based) to be more friendly (_and useful elsewhere within DMS like dealing with certificates since it can output JSON for querying/lookup_). `step crypto keypair example.public example.private` [as mentioned here](https://github.com/docker-mailserver/docker-mailserver/pull/3231#discussion_r1161279784) should work at generating the actual keypair.
- Unified OpenDKIM / Rspamd keypair formatting instead of separate CLI generators: https://github.com/docker-mailserver/docker-mailserver/pull/3231#discussion_r1161355541
- For rspamd config could be generated as JSON instead of UCL: https://github.com/docker-mailserver/docker-mailserver/pull/3599#discussion_r1371198705
  - `yq` example with detailed approach for improved domain + selector layout hierarchy in the filesystem that is agnostic (_OpenDKIM could also share this location, `check-for-changes.sh` could probably support it well too if it's from the config volume and optionally copied over to internal location for either rspamd/opendkim to use_).
  - KCL example as an alternative that was explored to just YAML: https://github.com/docker-mailserver/docker-mailserver/pull/3599#discussion_r1371565205

No immediate need for this. Just an improvement like using Vector for logs, low priority.

polarathene:
As a part of this, when the test suite is capable of using a local DNS service we should also include better DKIM test coverage to catch failures in signing like v13.0.0 introduced: https://github.com/docker-mailserver/docker-mailserver/pull/3669#discussion_r1406874445

The future test will be backend agnostic to OpenDKIM / Rspamd, once the test suite uses `compose.yaml` configs and the unified DKIM source files work this issue is about lands, we can quite easily run the same DKIM tests against both supported services by appending a `compose.yaml` config include or override.

polarathene:
I wrote up a little [guide / revision in a discussion response](https://github.com/orgs/docker-mailserver/discussions/3803#discussioncomment-8204632).

## Snippets from link

Generation without custom args support:

```bash
# Create an RSA 2048-bit keypair for DKIM usage:
step crypto keypair mail.public mail.private --kty RSA --size 2048 --no-password --insecure

# Format multi-line split public key for DNS record:
cat mail.public \
[...]

PR #3627, opened 2023-11-08
Rspamd: add check for DKIM private key files' permissions

georglauterbach:
# Description

The newly added function `__rspamd__check_dkim_permissions` performs a check on DKIM private key files. This is useful to prevent issues like #3621 in the future. The function is deliberately kept simple and may not catch every single misconfiguration in terms of permissions and ownership, but it should be quite accurate.

Please note that the Rspamd setup does NOT change at all, and the checks will not abort the setup in case they fail. A simple warning is emmited.

Reviewing commit-by-commit is advised.

Fixes #3621

## Type of change

- [x] Improvement (non-breaking change that does improve existing functionality)

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [ ] If necessary I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #3666, opened 2023-11-27
fix: `update-check.sh` should query GH Releases

polarathene:
# Description

We have had 3 concerns encountered related to this feature in the past.

**Resolved by:**
- CI will now provide the `VERSION` as an image ENV (`DMS_RELEASE`) in the `Dockerfile` to build. No need to manually update `VERSION` (_well, until a future release warrants removing it_).
- Future releases will now check the GH Releases page instead of current `master` branch `VERSION` file.

**Past concerns addressed:**
1. `:edge` should not care about the update process, these concerns should not have been raised:
   - https://github.com/docker-mailserver/docker-mailserver/pull/3659
   - https://github.com/docker-mailserver/docker-mailserver/pull/3662#pullrequestreview-1749456998
2. A release was made [without bumping `VERSION` accordingly](https://github.com/docker-mailserver/docker-mailserver/issues/2952#issuecomment-1364304849):
   - The fix was to [publish a new patch version/tag](https://github.com/docker-mailserver/docker-mailserver/pull/2944).
   - Wouldn't be a problem when checking GH releases for tag version.
3. A concern when the [release PR is pushed but tagging an official release is delayed](https://github.com/docker-mailserver/docker-mailserver/pull/3641#issuecomment-1826478255):
   - `:edge` unaffected since it's not actually tied to the `/VERSION` file that it would compare to Github `VERSION`.
   - Tagged releases would avoid sending notification until a new tagged release is actually published.

---

## Additional context

Applies changes were originally proposed here:
- https://github.com/docker-mailserver/docker-mailserver/pull/3641#issuecomment-1826483637
- https://github.com/docker-mailserver/docker-mailserver/pull/3641#issuecomment-1827027920

**ℹ️ Note:**
- The curl response will be the tag value, which includes a `v` to strip away. Whereas the current `VERSION` file has no `v` prefix.
- `VERSION` should remain committed for now. Prior versions will still check it, notably v13. Perhaps by v14 or later removing the file would fine? 🤷‍♂️ 

**⚠️ Caution:**
- I have not verified the workflow.
  - I expect the value is correct from the Github Actions [`context` docs](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) + [conditional expression](https://docs.github.com/en/actions/learn-github-actions/expressions#example) example.
  - We do have an [existing workflow using `github.ref` / `${GITHUB_REF}`](https://github.com/docker-mailserver/docker-mailserver/blob/68a43eb4970f2ab7680ffa470a1d26e37fa375f0/.github/workflows/docs-production-deploy.yml#L33-L41), `github.ref_name` should be just the tag like in the JSON `tag_name` field.

---

### `yq` vs `jaq`

I have expressed interest in [`yq`](https://github.com/mikefarah/yq) previously for managing DKIM / rspamd config (_and processing JSON output from `step` CLI for TLS cert tools_):
- https://github.com/docker-mailserver/docker-mailserver/pull/3599#discussion_r1371202623
- https://github.com/docker-mailserver/docker-mailserver/pull/3627/files#r1387243971
- https://github.com/orgs/kcl-lang/discussions/804#discussioncomment-7430551
- https://github.com/docker-mailserver/docker-mailserver/issues/3630

However it weighs 9.3MB vs an alternative [`jaq`](https://github.com/01mf02/jaq) at 1.7MB, which in this case was simpler to query the JSON field for the tag and strip the `v` prefix character.

See the `jaq` link for preference of it over `jq`. Docs wise, if using `jaq` elsewhere in DMS scripts you can follow the `jq` docs, or see the `jaq` README for some additional features/capabilities.

---

## Type of change

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,11 +18,7 @@ BATS_PARALLEL_JOBS     ?= 2
 all: lint build generate-accounts tests clean
 
 build: ALWAYS_RUN
-	@ DOCKER_BUILDKIT=1 docker build \
-		--tag $(IMAGE_NAME) \
-		--build-arg VCS_VERSION=$(shell git rev-parse --short HEAD) \
-		--build-arg VCS_REVISION=$(shell cat VERSION) \
-		.
+	@ docker build --tag $(IMAGE_NAME) .
 
 generate-accounts: ALWAYS_RUN
 	@ cp test/config/templates/postfix-accounts.cf test/config/postfix-accounts.cf
--- a/docs/content/examples/tutorials/docker-build.md
+++ b/docs/content/examples/tutorials/docker-build.md
@@ -10,7 +10,7 @@ You'll need to retrieve the git submodules prior to building your own Docker ima
 
 ```sh
 git submodule update --init --recursive
-docker build -t <YOUR CUSTOM IMAGE NAME> .
+docker build --tag <YOUR CUSTOM IMAGE NAME> .
 ```
 
 Or, you can clone and retrieve the submodules in one command:
@@ -21,19 +21,26 @@ git clone --recurse-submodules https://github.com/docker-mailserver/docker-mails
 
 ### About Docker
 
-#### Version
+#### Minimum supported version
 
-We make use of build-features that require a recent version of Docker. Depending on your distribution, please have a look at [the official installation documentation for Docker](https://docs.docker.com/engine/install/) to get the latest version. Otherwise, you may encounter issues, for example with the `--link` flag for a [`#!dockerfile COPY`](https://docs.docker.com/engine/reference/builder/#copy) command.
+We make use of build features that require a recent version of Docker. v23.0 or newer is advised, but earlier releases may work.
 
-#### Environment
+- To get the latest version for your distribution, please have a look at [the official installation documentation for Docker](https://docs.docker.com/engine/install/).
+- If you are using a version of Docker prior to v23.0, you will need to enable BuildKit via the ENV [`DOCKER_BUILDKIT=1`](https://docs.docker.com/build/buildkit/#getting-started).
 
-If you are not using `make` to build the image, note that you will need to provide `DOCKER_BUILDKIT=1` to the `docker build` command for the build to succeed.
+#### Build Arguments (Optional)
 
-#### Build Arguments
+The `Dockerfile` includes several build [`ARG`][docker-docs::builder-arg] instructions that can be configured:
 
-The `Dockerfile` takes additional, so-called build arguments. These are
+- `DOVECOT_COMMUNITY_REPO`: Install Dovecot from the community repo instead of from Debian (default = 1) 
+- `DMS_RELEASE`: The image version (default = edge)
+- `VCS_REVISION`: The git commit hash used for the build (default = unknown)
 
-1. `VCS_VERSION`: the image version (default = edge)
-2. `VCS_REVISION`: the image revision (default = unknown)
+!!! note
 
-When using `make` to build the image, these are filled with proper values. You can build the image without supplying these arguments just fine though.
+    - `DMS_RELEASE` (_when not `edge`_) will be used to check for updates from our GH releases page at runtime due to the default feature [`ENABLE_UPDATE_CHECK=1`][docs::env-update-check].
+    - Both `DMS_RELEASE` and `VCS_REVISION` are also used with `opencontainers` metadata [`LABEL`][docker-docs::builder-label] instructions.
+
+[docs::env-update-check]: https://docker-mailserver.github.io/docker-mailserver/latest/config/environment/#enable_update_check
+[docker-docs::builder-arg]: https://docs.docker.com/engine/reference/builder/#using-arg-variables
+[docker-docs::builder-label]: https://docs.docker.com/engine/reference/builder/#label
--- a/target/scripts/start-mailserver.sh
+++ b/target/scripts/start-mailserver.sh
@@ -120,7 +120,7 @@ function _register_functions() {
 
   [[ ${SMTP_ONLY}               -ne 1 ]] && _register_start_daemon '_start_daemon_dovecot'
 
-  [[ ${ENABLE_UPDATE_CHECK} -eq 1 && ${DMS_RELEASE} != 'edge' ]] && _register_start_daemon '_start_daemon_update_check'
+  [[ ${ENABLE_UPDATE_CHECK} -eq 1 ]] && [[ ${DMS_RELEASE} != 'edge' ]] && _register_start_daemon '_start_daemon_update_check'
 
   # The order here matters: Since Rspamd is using Redis, Redis should be started before Rspamd.
   [[ ${ENABLE_RSPAMD_REDIS}     -eq 1 ]] && _register_start_daemon '_start_daemon_rspamd_redis'
--- a/target/scripts/start-mailserver.sh
+++ b/target/scripts/start-mailserver.sh
@@ -120,7 +120,7 @@ function _register_functions() {
 
   [[ ${SMTP_ONLY}               -ne 1 ]] && _register_start_daemon '_start_daemon_dovecot'
 
-  [[ ${ENABLE_UPDATE_CHECK}     -eq 1 ]] && _register_start_daemon '_start_daemon_update_check'
+  [[ ${ENABLE_UPDATE_CHECK} -eq 1 && ${DMS_RELEASE} != 'edge' ]] && _register_start_daemon '_start_daemon_update_check'
 
   # The order here matters: Since Rspamd is using Redis, Redis should be started before Rspamd.
   [[ ${ENABLE_RSPAMD_REDIS}     -eq 1 ]] && _register_start_daemon '_start_daemon_rspamd_redis'
--- a/target/scripts/update-check.sh
+++ b/target/scripts/update-check.sh
@@ -3,7 +3,7 @@
 # shellcheck source=./helpers/log.sh
 source /usr/local/bin/helpers/log.sh
 
-VERSION="${DMS_VERSION#v}"
+VERSION="${DMS_RELEASE#v}"
 VERSION_URL='https://github.com/docker-mailserver/docker-mailserver/releases/latest'
 CHANGELOG_URL='https://github.com/docker-mailserver/docker-mailserver/blob/master/CHANGELOG.md'
 
--- a/target/scripts/update-check.sh
+++ b/target/scripts/update-check.sh
@@ -4,7 +4,7 @@
 source /usr/local/bin/helpers/log.sh
 
 VERSION="${DMS_VERSION#v}"
-VERSION_URL='https://raw.githubusercontent.com/docker-mailserver/docker-mailserver/master/VERSION'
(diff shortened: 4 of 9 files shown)


target/scripts/start-mailserver.sh
```diff
@@ -120,7 +120,7 @@ function _register_functions() {
 
   [[ ${SMTP_ONLY}               -ne 1 ]] && _register_start_daemon '_start_daemon_dovecot'
 
-  [[ ${ENABLE_UPDATE_CHECK}     -eq 1 ]] && _register_start_daemon '_start_daemon_update_check'
+  [[ ${ENABLE_UPDATE_CHECK} -eq 1 && ${DMS_RELEASE} != 'edge' ]] && _register_start_daemon '_start_daemon_update_check'
```
georglauterbach, line 123:
```suggestion
  [[ ${ENABLE_UPDATE_CHECK} -eq 1 ]] && [[ ${DMS_RELEASE} != 'edge' ]] && _register_start_daemon '_start_daemon_update_check'
```

target/scripts/update-check.sh
```diff
@@ -17,7 +17,8 @@ fi
 
 while true; do
   # get remote version information
-  LATEST=$(curl -Lsf "${VERSION_URL}")
+  # JSON response provides a field for the release tag, the `v` prefix is removed with `[1:]`
+  LATEST=$(curl -sfL -H 'accept: application/json' "${VERSION_URL}" | jaq -r '.tag_name[1:]')
```
georglauterbach, line 21: We might want to alias `jq` to `jaq`? This way users could transparently use `jq` too. But I would not insist on this.

georglauterbach (changes requested):
LGTM 👍🏼 

One tiny nitpick (style) 🚀

target/scripts/update-check.sh
```diff
@@ -3,8 +3,8 @@
 # shellcheck source=./helpers/log.sh
 source /usr/local/bin/helpers/log.sh
 
-VERSION=$(</VERSION)
-VERSION_URL='https://raw.githubusercontent.com/docker-mailserver/docker-mailserver/master/VERSION'
+VERSION="${DMS_VERSION#v}"
```
georglauterbach, line 6: What is `DMS_VERSION`?

target/scripts/update-check.sh
polarathene: > This way users could transparently use `jq` too.

Same policy as usual, wait until there is sufficient interest expressed from users. If there's only a few they are probably quite happy to use `jaq`.

This is for internal use, we don't document it as available officially, so we're also free to change it to something else like `yq` if `jaq` isn't as good of a fit elsewhere I'd like to use it.

target/scripts/update-check.sh
polarathene: Good catch!

That was me being wishy/washy over `DMS_VERSION` or `DMS_RELEASE` and forgetting to update this one 😅

polarathene:
```suggestion
VERSION="${DMS_RELEASE#v}"
```

georglauterbach (dismissed):

polarathene:
Just spotted the `Makefile` sets the build args too, although they seem to be the wrong way around.

Do we even need to set the revision SHA when building with `make build`? I could fix those or drop them.

docs/content/examples/tutorials/docker-build.md
```diff
@@ -21,19 +21,26 @@ git clone --recurse-submodules https://github.com/docker-mailserver/docker-mails
 
 ### About Docker
 
-#### Version
+#### Minimum supported version
 
-We make use of build-features that require a recent version of Docker. Depending on your distribution, please have a look at [the official installation documentation for Docker](https://docs.docker.com/engine/install/) to get the latest version. Otherwise, you may encounter issues, for example with the `--link` flag for a [`#!dockerfile COPY`](https://docs.docker.com/engine/reference/builder/#copy) command.
```
polarathene, line 26: I removed the `--link` example as I think that has a fallback behaviour for compatibility and the information here with link doesn't provide that context.

- The HereDocs feature arrived earlier [around 2021H2](https://www.docker.com/blog/introduction-to-heredocs-in-dockerfiles/)
- The `COPY --link` feature [around 2022Q2](https://www.docker.com/blog/image-rebase-and-improved-remote-cache-support-in-new-buildkit/)

Docker [v23 was released Feb 2023](https://docs.docker.com/engine/release-notes/23.0/#2300). It enables BuildKit by default. It makes for a good baseline (_for building the image_), instead of just "recent version", while being a little vague on earlier version support for brevity.

docs/content/examples/tutorials/docker-build.md
```diff
@@ -21,19 +21,26 @@ git clone --recurse-submodules https://github.com/docker-mailserver/docker-mails
...
+We make use of build features that require a recent version of Docker. v23.0 or newer is advised, but earlier releases may work.
 
-#### Environment
+- To get the latest version for your distribution, please have a look at [the official installation documentation for Docker](https://docs.docker.com/engine/install/).
+- If you are using a version of Docker prior to v23.0, you will need to enable BuildKit via the ENV [`DOCKER_BUILDKIT=1`](https://docs.docker.com/build/buildkit/#getting-started).
 
-If you are not using `make` to build the image, note that you will need to provide `DOCKER_BUILDKIT=1` to the `docker build` command for the build to succeed.
```
polarathene, line 30: This page no longer makes mention of the `make build` command, it doesn't add much value to the reader so seemed like noise?

docs/content/examples/tutorials/docker-build.md
```diff
@@ -21,19 +21,26 @@ git clone --recurse-submodules https://github.com/docker-mailserver/docker-mails
...
-2. `VCS_REVISION`: the image revision (default = unknown)
+!!! note
 
-When using `make` to build the image, these are filled with proper values. You can build the image without supplying these arguments just fine though.
+    - `DMS_RELEASE` (_when not `edge`_) will be used to check for updates from our GH releases page at runtime due to the default feature [`ENABLE_UPDATE_CHECK=1`][docs::env-update-check].
+    - Both `DMS_RELEASE` and `VCS_REVISION` are also used with `opencontainers` metadata [`LABEL`][docker-docs::builder-label] instructions.
+
+[docs::env-update-check]: https://docker-mailserver.github.io/docker-mailserver/latest/config/environment/#enable_update_check
+[docker-docs::builder-arg]: https://docs.docker.com/engine/reference/builder/#using-arg-variables
+[docker-docs::builder-label]: https://docs.docker.com/engine/reference/builder/#label
```
polarathene, line 46: Revised this section, there is technically another ARG for log output but I've omitted that. The Dovecot one seemed relevant to the reader.

I've clarified the revision context, and to an extent `DMS_RELEASE`, with some insights that might be relevant to the reader for custom images.

Makefile
```diff
@@ -18,11 +18,7 @@ BATS_PARALLEL_JOBS     ?= 2
 all: lint build generate-accounts tests clean
 
 build: ALWAYS_RUN
-	@ DOCKER_BUILDKIT=1 docker build \
-		--tag $(IMAGE_NAME) \
-		--build-arg VCS_VERSION=$(shell git rev-parse --short HEAD) \
-		--build-arg VCS_REVISION=$(shell cat VERSION) \
-		.
```
polarathene, line 25: Since v23 is now the min version advised for building and the related docs make note of `DOCKER_BUILDKIT=1`, the ENV is removed from here.

The build args have also been dropped. The inputs were around the wrong way, and generally for building an image that isn't a release these inputs aren't useful. If users need them they should prefer a more explicit command like we do with CI.

Not part of the image releases, so I don't see this as a breaking change.

polarathene (commented):
Some review comments for context to the latest changes.

georglauterbach (approved):
LGTM 👍🏼 

I will wait for @casperklein before merging - please be advised that this is a high-priority PR - I'd like to publish v13.0.1 by the end of the day :)

Dubbeldrank:
> LGTM 👍🏼
> 
> I will wait for @casperklein before merging - please be advised that this is a high-priority PR - I'd like to publish v13.0.1 by the end of the day :)

Is it possible to release this DKIM hotfix without this merge?

cybermcm:
Issue is fixed currently with :edge version

Dubbeldrank:
> Issue is fixed currently with :edge version

It doesn't feel okay to run edge on a stable environment.

casperklein:
One regression I noticed:

``[[ ${ENABLE_UPDATE_CHECK} -eq 1 ]] && [[ ${DMS_RELEASE} != 'edge' ]]``

Previously, when you are on a tagged release that contained a bug of some kind and got advised by us to switch to edge, because it was fixed there already, you would still be notified via update-check, when a new tagged release is available, so you could switch back.
That's because in :edge, the VERSION file contained the latest tagged version at the time the image was pulled.

> :edge should not care about the update process, these concerns should not have been raised:
> https://github.com/docker-mailserver/docker-mailserver/pull/3659
> https://github.com/docker-mailserver/docker-mailserver/pull/3662#pullrequestreview-1749456998

No real issues if you ask me. In #3659 the root cause for the mail flood wasn't identified. But not related to the update-check IMO. #3662 is already fixed.

The initial question was, if we want to check against the latest release or the VERSION file in the master branch. Checking the release is the way to go.
However, replacing the VERSION file with an ENV (with other content --> "edge") and not doing an update-check at all when having a edge image is another story.

polarathene:
> That's because in :edge, the VERSION file contained the latest tagged version at the time the image was pulled.

I don't consider that desirable as a default.

If a user is switching to `:edge` temporarily and would like this kind of notification they can now set the `DMS_RELEASE` ENV with the release tag they want to monitor for newer updates since with (eg: `DMS_RELEASE='v12.1.0'), and they will get that functionality.

We've already had reports of the prior behaviour not being desired on `:edge` (_granted they could opt out via `UPDATE_CHECK=0`_).

If you'd prefer to source the version tag from a `git` command to find the latest tag, that could be done instead, likewise `UPDATE_CHECK` could then be given a build arg to default it to disabled on `:edge`.

Alternatively, if you'd like the `DMS_RELEASE` ENV to be documented for `:edge` users, I can add that.

---

> No real issues if you ask me. In #3659 the root cause for the mail flood wasn't identified. But not related to the update-check IMO. #3662 is already fixed.

There wasn't anything in #3659 to say they expected a notification on `:edge`, and I can't say that I'd want that behaviour on other software I'd run on a nightly/beta/edge like release channel. Update notifications on `:edge` should be opt-in.

#3662 shouldn't have been raised in the first place as `:edge` shouldn't need to care about stable releases. Regardless, with this PR it's resolved with an alternative approach.

---

> However, replacing the VERSION file with an ENV (with other content --> "edge") and not doing an update-check at all when having a edge image is another story.

If you can refer to other software that notifies users of stable updates on an `:edge` equivalent release channel, I'd be more supportive of that.

Otherwise I'd rather our users express actual demand for the functionality and I'll implement/document it.

polarathene:
I'll publish v13.0.1 shortly 🚀

PR #3702, opened 2023-12-16
docs: Rspamd DKIM config simplify via `path` setting

denisix:
using DKIM signing in Rspamd - automatically match keys for domains in dkim_signing.conf

# Description

in dkim_signing.conf added configuration lines to automatically match domain keys instead of manually adding config for each domain that configured in DMS

## Type of change

## Checklist:


docs/content/config/best-practices/dkim_dmarc_spf.md
```diff
@@ -156,6 +156,10 @@ DKIM is currently supported by either OpenDKIM or Rspamd:
         use_esld = true;
         check_pubkey = true; # you want to use this in the beginning
 
+        selector = "mail";
+        path = "/tmp/docker-mailserver/opendkim/keys/$domain/$selector.private"; # this will automatically match keys for domains
```
polarathene, line 160: ~~Can you please cite the source for this feature?~~ (_**EDIT:** Not necessary, cited below_)

I see on the [rspamd dkim docs](https://rspamd.com/doc/modules/dkim_signing.html#configuration) that `selector` is used as the default with the `$selector` variable, ~~but `$domain` is only mentioned in a comment, is that also a fixed fallback value or is it documented somewhere as dynamic?~~

**EDIT:** After reading the [first section](https://rspamd.com/doc/modules/dkim_signing.html#principles-of-operation) and going over the config example / defaults, I see that `$domain` is from `use_domain` setting, which defaults to dynamically sourced from `header` but could be configured as a fixed domain. While `selector` defaults to `dkim`.

polarathene, line 160: - I've moved the comment above the `path` setting and revised it for context.
- The `path` was pointing to `opendkim`, it was probably meant to point to `rspamd`? I've normalized it to `dkim` which we'd like to switch to (_perhaps for v14_) Feel free to set it to `rspamd` instead for now, which would be consistent with the `path` config that follows afterwards.

```suggestion
        selector = "mail";
        # The location is searched for a DKIM key with `$domain` sourced from the mail `FROM` header, while `$selector` is configured for `mail` as a default.
        path = "/tmp/docker-mailserver/dkim/keys/$domain/$selector.private";
```

docs/content/config/best-practices/dkim_dmarc_spf.md
```diff
@@ -156,6 +156,10 @@ DKIM is currently supported by either OpenDKIM or Rspamd:
...
+
+        # domain specific configurations can be provided below:
```
polarathene, line 162: Alternatively there is a simple [`selector_map` setting](https://rspamd.com/doc/modules/dkim_signing.html#using-maps) that references a file of `$domain $selector` pairs per line.

The `$domain` is used to lookup the `$selector` override (if any), similar to how the config here works. Assuming `selector_maps` is compatible with multiple selectors as well, then the only difference from this config below is configuring different `path` which we don't really need, and a similar setting `path_maps` supports.

polarathene (changes requested):
Nice improvement thanks!

polarathene:
Worth noting perhaps, the default for `$domain` in the config has `eSLD` normalization enabled, which means some configurations (like `hello@marketing.example.com`) [won't have a valid direct match](https://github.com/rspamd/rspamd/issues/1808#issuecomment-324560663) (normalizes `$domain` to `example.com`).

I think that applies regardless of config approach. Just mentioning for maintainers to be aware of if that ever comes up as a reported problem in future 😅

georglauterbach:
> Worth noting perhaps, the default for `$domain` in the config has `eSLD` normalization enabled, which means some configurations (like `hello@marketing.example.com`) [won't have a valid direct match](https://github.com/rspamd/rspamd/issues/1808#issuecomment-324560663) (normalizes `$domain` to `example.com`).
> 
> 
> 
> I think that applies regardless of config approach. Just mentioning for maintainers to be aware of if that ever comes up as a reported problem in future 😅 

This can probably be configured with `use_esld = true;`?

polarathene:
> This can probably be configured with `use_esld = true;`?

Yes, just raising awareness about it since it may show up some day as a bug report. If it does it's a simple documentation fix for rspamd/faq page(s) to communicate awareness to users.

It's probably unlikely that `use_esld` needs to be set to `false`? This is just storage of the private key within DMS for signing, and it's ok to use the same DKIM private key for various sender subdomains? (_AFAIK they can all point to the same public key DNS record, or individual DNS records with the same public key_)

georglauterbach:
@denisix please address the PR feedback; we can then go ahead and merge the PR.

georglauterbach:
last ping @denisix

docs/content/config/best-practices/dkim_dmarc_spf.md
polarathene: This feedback has been migrated to a separate TODO issue: https://github.com/docker-mailserver/docker-mailserver/issues/3778

docs/content/config/best-practices/dkim_dmarc_spf.md
polarathene: This context with source citing is now also part of the related TODO issue: https://github.com/docker-mailserver/docker-mailserver/issues/3778

polarathene (dismissed):
As the contributor is not engaging with review feedback, I'll apply my change suggestion.

DMS could better manage this internally and simplify on configuration a user needs to be aware of and manually manage via docs, but that'll need someone to contribute the change. A separate issue for tracking that with all the information discussed here has been opened.

Regarding the [concern with `use_esld = true` and DKIM related rspamd bug reports](https://github.com/docker-mailserver/docker-mailserver/pull/3702#issuecomment-1863480339), maintainers should be aware of this behaviour where it can affect some users that deviate from our happy path of `hostname: mail.example.com` with `use@example.com` mail domains if they have trouble with `$domain` expecting to match a `mail.example.com` (_eSLD normalization will look for `example.com` instead, which may not work if the user has `user@marketing.example.com` accounts?_).

Since that setting has nothing to do with the contribution here, it's already `true` in the docs above, I'll go ahead and merge 👍 (_we could improve on the docs a bit more by explaining the suggested config further though, especially with the `path` that isn't actually valid presently_)

docs/content/config/best-practices/dkim_dmarc_spf.md
```diff
@@ -156,6 +156,11 @@ DKIM is currently supported by either OpenDKIM or Rspamd:
         use_esld = true;
         check_pubkey = true; # you want to use this in the beginning
 
+        selector = "mail";
+        # The location is searched for a DKIM key with `$domain` sourced from the MIME mail message `From` header, while `$selector` is configured for `mail` (as a default fallback).
```
polarathene, line 160:
```suggestion
        # The path location is searched for a DKIM key with these variables:
        # - `$domain` is sourced from the MIME mail message `From` header
        # - `$selector` is configured for `mail` (as a default fallback)
```

polarathene (changes requested):

polarathene (approved):

docs/content/config/best-practices/dkim_dmarc_spf.md
```diff
@@ -156,6 +156,13 @@ DKIM is currently supported by either OpenDKIM or Rspamd:
         use_esld = true;
         check_pubkey = true; # you want to use this in the beginning
 
+        selector = "mail";
+        # The path location is searched for a DKIM key with these variables:
+        # - `$domain` is sourced from the MIME mail message `From` header
+        # - `$selector` is configured for `mail` (as a default fallback)
+        path = "/tmp/docker-mailserver/dkim/keys/$domain/$selector.private";
```
williamdes, line 163: This path is wrong, currently the keys are in: `/tmp/docker-mailserver/opendkim/keys/`

I have used `path = "/etc/opendkim/keys/$domain/$selector.private";` until now

polarathene: > This path is wrong

It's an example and was explicitly changed to that by me for the docs.

The [contributor used `/tmp/docker-mailserver/...`](https://github.com/docker-mailserver/docker-mailserver/pull/3702/files/02f09ef0e0570344248eadbcea39f2b595494762#diff-aee5f16ae135f9232103399f5670afb3c353405767d6020b38e51480c651d83f) instead of `/etc/...` but otherwise they also referred to `opendkim`.

Not sure why you'd want OpenDKIM as the path for rspamd though? I want this to be a unified location for DKIM keys, but someone will need to contribute that change.

williamdes: Well, it's more that rspamd wanted to do signing before I would even know it. So I patched the path and selector to use open dkim path. Since opendkim was the first dkim stuff I had setup
Just feels like there is too many moving components in DMS and without a diagram I can not decide to shut down one component or another.
Anyway this config allows to local dkim sign and that's all I needed to pass 👍

PR #3711, opened 2023-12-18
ci: Remove `VERSION` from `Dockerfile`

polarathene:
# Description

This line was meant to be dropped with the switch to using the `DMS_RELEASE` ENV.

No changelog entry seems necessary?


casperklein:
> No changelog entry seems necessary?

I would mention the file removal. There may be [custom stuff](https://github.com/casperklein/docker-mailserver-mailctl/blob/ab223542c03e64fce87c010d4aa59ff72cf4b9a1/mailctl#L240) from users depending on that file 😆

georglauterbach (dismissed):

polarathene:
> I would mention the file removal.

Technically it was intended to be implicit as part of the changelog for `13.0.1`, if it was handled then I don't think the changelog entry would have changed?

I can add an explicit mention of the file being removed from the `Dockerfile` if you'd like?

---

> There may be [custom stuff](https://github.com/casperklein/docker-mailserver-mailctl/blob/ab223542c03e64fce87c010d4aa59ff72cf4b9a1/mailctl#L240) from users depending on that file 😆

That's a pretty neat utility! 😎 

Kind of seeing a matryoshka hierarchy there: `mailctl` => `setup.sh` => `setup` 😆

georglauterbach (approved):

PR #3727, opened 2023-12-27
security(Postfix): Protect against "SMTP Smuggling" attack

### LINKED ISSUE #3719: security: postfix SMTP smuggling
### Subject

Something else that requires developers attention

### Description

A vulnerability affecting several mail server implementations, including postfix, has been disclosed recently.
See the related postfix documentation page for more details: https://www.postfix.org/smtp-smuggling.html

While a short-term workaround exists, it is only partial:
>  NOTE: This will stop only the published form of the attack. Other forms exist that will not be stopped in this manner.

Postfix should be updated to a version in which the vulnerability is fixed (3.8.4, 3.7.9, 3.6.13 or 3.5.23) and the new optional feature [`smtpd_forbid_bare_newline`](https://www.postfix.org/postconf.5.html#smtpd_forbid_bare_newline) be set to `yes`. The feature will be enabled by default for postfix >= 3.9.

If the `smtpd_forbid_bare_newline` feature cannot be enabled for backward-compatibility reasons, you could release a patch version of DMS with one of the patched postfix versions and let users decide whether they want to enable the feature [via their local configuration](https://docker-mailserver.github.io/docker-mailserver/latest/config/advanced/override-defaults/postfix/).

georglauterbach:
# Description

Prevent (currently known) attack vectors of SMTP smuggling, see #3719.

Fixes #3719

## Type of change

## Checklist:


georglauterbach:
Oh great, this messes with our test suite commands for sending and receiving e-mails, I think... I do not have time now to fix this though...

target/postfix/main.cf
```diff
@@ -56,6 +56,9 @@ smtpd_client_restrictions = permit_mynetworks, permit_sasl_authenticated, reject
 smtpd_sender_restrictions = $dms_smtpd_sender_restrictions
 smtpd_discard_ehlo_keywords = silent-discard, dsn
 disable_vrfy_command = yes
+smtpd_data_restrictions = reject_unauth_pipelining
```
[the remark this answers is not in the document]
polarathene: Might be worth adding some context of the restriction.

```suggestion
# Block clients that speak too early (don't wait their turn).
smtpd_data_restrictions = reject_unauth_pipelining
```

Keeping it in `main.cf` should be fine AFAIK, Amavis will inherit it and it may still make sense when Amavis is not used? (_although I kinda prefer to keep it explicit on the Amavis snippet, instead of implicitly inferring it from `main.cf`_)

---

The [documented short-term workaround](https://github.com/docker-mailserver/docker-mailserver/issues/3719#issuecomment-1870614251) for the SMTP smuggling vulnerability is:

```suggestion
smtpd_data_restrictions = reject_unauth_pipelining
# This will conflict with existing line:
# smtpd_discard_ehlo_keywords = silent-discard, dsn
smtpd_discard_ehlo_keywords = chunking
# Should replace original (and add contextual comment to PR for the dsn config)
# Also requires checking that the logic for that PR isn't broken from any assumptions that prepending/appending 'chunking' here may break:
# smtpd_discard_ehlo_keywords = chunking, silent-discard, dsn
```

but once we have v14 of DMS we'll have Postfix 3.7.6 and can use the original short-term workaround:

```suggestion
# Block clients that speak too early (don't wait their turn).
# Amavis config expects this (implicitly inherits from main.cf)
smtpd_data_restrictions = reject_unauth_pipelining

# Security - Prevent "SMTP Smuggling" exploit
# https://github.com/docker-mailserver/docker-mailserver/issues/3719
smtpd_forbid_unauth_pipelining = yes
# See prior suggestion for concerns
smtpd_discard_ehlo_keywords = chunking
```

Only benefit then is to get this in for 13.1, but you've got it marked for v14? In which case the Postfix 3.7.6 fix is more appropriate.

target/postfix/main.cf
```diff
@@ -56,6 +56,9 @@ smtpd_client_restrictions = permit_mynetworks, permit_sasl_authenticated, reject
...
+# TODO enable when possible, see https://github.com/docker-mailserver/docker-mailserver/issues/3719#issuecomment-1868287208
+#smtpd_forbid_bare_newline = yes
```
polarathene, line 61: "When possible" could better clarify that the blocker is waiting on Debian to provide an updated package of Postfix (_3.8.4, 3.7.9, 3.6.13 and 3.5.23_). For Bookworm that'd be `3.7.9`.

The comment also only refers to enabling the parameter, despite obsoleting the [short-term workarounds](https://www.postfix.org/smtp-smuggling.html#short).

---

The [longterm-fix](https://www.postfix.org/smtp-smuggling.html#long) also references an optional 2nd parameter to configure:

```
smtpd_forbid_bare_newline = yes
smtpd_forbid_bare_newline_exclusions = $mynetworks
```

We shouldn't configure it by default in the image though (_since Docker with IPv6 hosts but IPv4 only containers with default `user-proxy: true` daemon config_) can be part of the trusted docker subnet (_depending on our `PERMIT_DOCKER` ENV_). Those that need it (_like our test-suite_) could enable it, and we can document that in our docs somewhere.

polarathene (changes requested):
**Change overview** (_see review feedback for insights_):
- [ ] Add `smtpd_data_restrictions` to `main.cf`, remove the client and recipient restriction for `reject_unauth_pipelining`.
  - This defers rejection to a later stage (_assuming a permit rule didn't skip restrictions prior_), where any detected pipelining in the exchange would trigger the restriction, not just within `DATA` command.
- [ ] Keep `smtpd_data_restrictions` in Amavis config.
- [ ] For a v14 target, `smtpd_data_restrictions` is not necessary to workaround SMTP smuggling vulnerability.
  - We would have Postfix 3.7.6, this makes `smtpd_forbid_unauth_pipelining = yes` available. It should be adopted instead (but the `smtpd_data_restrictions` checklist above is valid (_perhaps as a separate PR_).
  - Both short-term workarounds also reference enabling `smtpd_discard_ehlo_keywords = chunking`. This conflicts with the same existing parameter already configured, prepending/appending to that should be fine, but it should be assessed that the feature doesn't have any logic that would accidentally remove `chunking`, the feature tests may not be sufficient for that, but should be obvious looking over the PR diff.
- [ ] Additional context in comments should be added regarding the long-term fix:
  - [ ] It replaces the short-term fix parameters as workarounds.
  - [ ] It's unclear when Debian Bookworm will update Postfix package with the available fix, but we can mention the min version needed to toggle the long-term fix.

---

**Overview for `smtpd_data_restrictions`:**
- `smtpd_data_restrictions = reject_unauth_pipelining` should probably remain on Amavis config, even though it'd be technically redundant (_since it would now inherit from `main.cf`_). Reasoning is for maintenance, where this is an explicit Amavis config (official docs), and DMS `main.cf` could change again as review feedback implies. Better for maintenance visibility to retain it.
- `smtpd_data_restrictions = reject_unauth_pipelining` on `main.cf` still may make sense, regardless of Amavis integration.
  - Presently we have this check at earlier restriction phases (client and recipient), but those seem unnecessary and can be dropped in favor of deferring to the data restriction which Postfix docs seem to indicate would be sufficient. Only difference would be rejecting due to the restriction slightly earlier (_at the `RCPT TO` SMTP command_).
  - Potentially breaking, but only in the sense that a client is not respecting waiting their turn at the `DATA` command, but for some reason was respectful earlier 🤷‍♂️

polarathene:
> Oh great, this messes with our test suite commands for sending and receiving e-mails, I think...

Technically an improvement 😝 

We skipped it in the past by having Amavis disabled in the test suite. Now that you've lifted it up into default `main.cf`, it gets caught. Which helps illustrate that the current restriction on client and recipient restrictions wasn't too effective vs applying at the data stage 🤔 

---

> I do not have time now to fix this though...

Should be fixed if we migrate away from `nc` / `openssl` based to using `swaks`. This is available as a package in Debian. Provided most tests use your helper to send mail it should reduce the friction to swap to using `swaks` and [something I've been meaning to do](https://github.com/docker-mailserver/docker-mailserver/pull/3480#pullrequestreview-1795498610).

An easier workaround of course is to just use our own override support with `postfix-main.cf` (_although that'd conflict with existing tests that provide their own_).

**EDIT:** There is an alternative short-term workaround for SMTP smuggling that you can use instead (_as per my review checklist_), that'll avoid the test failure issue. The `smtpd_data_restrictions` addition to `main.cf` is still worthwhile, but can be deferred until `swaks` is adopted.

georglauterbach:
Nice feedback, I will take care of all mentions over the course of the next days. This is marked for v14 still, so we do not need to fiddle with v13.1.0; makes my life easier.

georglauterbach:
**This PR currently depends on #3403**.

casperklein:
fyi: I just noticed, that the debian11 postfix package was bumped to version 3.5.23 and introduced this new default values:

> smtpd_forbid_bare_newline = no
> smtpd_forbid_bare_newline_exclusions = $mynetworks
> smtpd_forbid_unauth_pipelining = no
> tls_config_file = default
> tls_config_name =

However, the new version is not [listed](https://packages.debian.org/bullseye/postfix) yet.

georglauterbach:
Update: we are currently working on reviewing

- https://github.com/docker-mailserver/docker-mailserver/pull/3403 and
- https://github.com/docker-mailserver/docker-mailserver/pull/3732

which are both high-priority by now. Hence, a few breaking changes are going to be introduced with v14.0.0.

polarathene:
Since the current base image now has an [updated `postfix` with support for the same long-term fix we want to use](https://github.com/docker-mailserver/docker-mailserver/issues/3719#issuecomment-1873783044), would it make sense to address that first and get v13.1.1 released? Or do you want to focus on v14 straight away?

target/postfix/main.cf
```diff
@@ -57,6 +57,12 @@ smtpd_sender_restrictions = $dms_smtpd_sender_restrictions
 smtpd_discard_ehlo_keywords = silent-discard, dsn
 disable_vrfy_command = yes
 
+# Security - Prevent SMTP Smuggling attack
+# https://www.postfix.org/smtp-smuggling.html#long
+smtpd_forbid_bare_newline = yes
+# It is possible to exclude clients on trusted networks from this restriction:
+# smtpd_forbid_bare_newline_exclusions = $mynetworks
```
polarathene, line 64: The default is actually `$mynetworks` already, perhaps we want to disable that? Or rely on the default of `PERMIT_DOCKER=none` making that unnecessary?

Only concern is due to Docker and IPv6 capable hosts mistakenly routing through to an IPv4 only DMS container (_due to Docker daemon `user-proxy: true` default config_), where it'll be trusted in `$mynetworks` that cover the Docker network gateway address in a subnet (eg: `x.y.z.1`).

polarathene (approved):
Updated to use the [long-term fix](https://github.com/docker-mailserver/docker-mailserver/pull/3727#discussion_r1437273096) discussed in prior review feedback.

Prior [review feedback on the `smtpd_data_restrictions = reject_unauth_pipelining`](https://github.com/docker-mailserver/docker-mailserver/pull/3727#discussion_r1437264789) is no longer required as part of this PR. I've reverted that, but we can have a separate PR handle that _(add to `main.cf`, keep duplicate in `postfix-amavis.cf`_).

---

Technically if we make a v13.1.1 release, while it's a fix against the attack, it may be breaking for clients unintentionally? Which is why the default is `no` for existing releases.
- Release as v13.1.1?
- Release as v13.2?
- Release with feature disabled, but add a `postfix-main.cf` override suggestion in changelog?
- Delay until v14 release?

polarathene:
Tests presently fail due to the new parameter not being recognized. The image was not built with newer packages as our cache fallback provided all earlier layers of the `Dockerfile`:

https://github.com/docker-mailserver/docker-mailserver/blob/0889b0ff063a37b482113a684d934e4bd728a33c/Dockerfile#L204

I'll purge the CI cache to fix.

georglauterbach:
I'm still in favor of merging https://github.com/docker-mailserver/docker-mailserver/pull/3403 and https://github.com/docker-mailserver/docker-mailserver/pull/3732 first, then merge this PR, and then release v14.0.0. But I will not block on another solution either; someone will need to take care of this though...

polarathene:
> I'm still in favor of merging #3403 and #3732 first, then merge this PR

See my recent [review update](https://github.com/docker-mailserver/docker-mailserver/pull/3727#pullrequestreview-1800955336). Fail2Ban is failing consistently from this change.
- I suspect with `swaks` it's a non-issue, tests will pass.
- Since the `swaks` PR doesn't introduce any breaking changes, I think it may be worth merging that, then this exploit fix assuming tests pass. If tests here pass, then you can choose to go ahead with merging the base image update before this PR if you prefer that.

We can get v14 out quickly, so I don't mind. It may benefit users to stage this out as v13.2 before the base image update, but hopefully neither change will cause any new issues to troubleshoot 👍

georglauterbach:
I like this route:

1. Merge swaks
2. Merge this (if tests pass)
3. Maybe merge OAUTH2
4. Release v13.2
5. Merge base image update

polarathene (approved):

PR #3744, opened 2024-01-03
Postfix: add `smtpd_data_restrictions = reject_unauth_pipelining`

### LINKED ISSUE #3741: [TODO]: Add `smtpd_data_restrictions = reject_unauth_pipelining` to `main.cf`
### Description

- Add `smtpd_data_restrictions = reject_unauth_pipelining` to `main.cf`. Include context via comment? (_eg: "# Reject clients that speak to Postfix too early (don't wait their turn)."_)
- Keep the equivalent in `postfix-amavis.cf` for visibility as explicit config required by Amavis.
- Remove `reject_unauth_pipelining` from earlier restrictions (`smtpd_client_restrictions`, `smtpd_recipient_restrictions`).

---

[This change was originally bundled with an tracking issue for adding `swaks`](https://github.com/docker-mailserver/docker-mailserver/issues/3728#issue-2057839259) (_since that is required for tests to support the change_). That got resolved much earlier than expected, so I've extracted this out to it's own separate issue.

This change isn't too important, as it was originally motivated to support a short-term fix for the SMTP Smuggling attack that is no longer necessary.

`smtpd_data_restrictions = reject_unauth_pipelining` will defer rejection to a later stage (_assuming a permit rule didn't skip restrictions prior_), where any detected pipelining in the exchange would trigger the restriction, not just within the `DATA` SMTP command.

georglauterbach:
# Description

See #3741

Fixes #3741

## Type of change

## Checklist:

### DIFF
--- a/target/amavis/postfix-amavis.cf
+++ b/target/amavis/postfix-amavis.cf
@@ -23,7 +23,6 @@ smtp-amavis     unix    -       -       n       -       2       smtp
   -o smtpd_helo_restrictions=
   -o smtpd_sender_restrictions=
   -o smtpd_recipient_restrictions=permit_mynetworks,reject
-  -o smtpd_data_restrictions=reject_unauth_pipelining
   -o smtpd_end_of_data_restrictions=
   -o mynetworks=127.0.0.0/8
   -o smtpd_error_sleep_time=0
--- a/target/postfix/main.cf
+++ b/target/postfix/main.cf
@@ -51,10 +51,11 @@ smtpd_helo_required = yes
 smtpd_delay_reject = yes
 smtpd_helo_restrictions = permit_mynetworks, reject_invalid_helo_hostname, permit
 smtpd_relay_restrictions = permit_mynetworks permit_sasl_authenticated defer_unauth_destination
-smtpd_recipient_restrictions = permit_sasl_authenticated, permit_mynetworks, reject_unauth_destination, reject_unauth_pipelining, reject_invalid_helo_hostname, reject_non_fqdn_helo_hostname, reject_unknown_recipient_domain
-smtpd_client_restrictions = permit_mynetworks, permit_sasl_authenticated, reject_unauth_destination, reject_unauth_pipelining
+smtpd_recipient_restrictions = permit_sasl_authenticated, permit_mynetworks, reject_unauth_destination, reject_invalid_helo_hostname, reject_non_fqdn_helo_hostname, reject_unknown_recipient_domain
+smtpd_client_restrictions = permit_mynetworks, permit_sasl_authenticated, reject_unauth_destination
 smtpd_sender_restrictions = $dms_smtpd_sender_restrictions
 smtpd_discard_ehlo_keywords = silent-discard, dsn
+smtpd_data_restrictions = reject_unauth_pipelining
 disable_vrfy_command = yes
 
 # Security - Prevent SMTP Smuggling attack
--- a/test/tests/parallel/set1/spam_virus/postgrey_enabled.bats
+++ b/test/tests/parallel/set1/spam_virus/postgrey_enabled.bats
@@ -44,7 +44,7 @@ function teardown_file() { _default_teardown ; }
   # The other spam checks in `main.cf:smtpd_recipient_restrictions` would interfere with testing postgrey.
   _run_in_container sed -i \
     -e 's/permit_sasl_authenticated.*policyd-spf,$//g' \
-    -e 's/reject_unauth_pipelining.*reject_unknown_recipient_domain,$//g' \
+    -e 's/reject_invalid_helo_hostname.*reject_unknown_recipient_domain,$//g' \
     -e 's/reject_rbl_client.*inet:127\.0\.0\.1:10023$//g' \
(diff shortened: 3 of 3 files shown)


georglauterbach:
Okay, tests fail as expected; wherever we use `nc`,

polarathene:
> tests fail as expected; wherever we use `nc`

No worries, this change isn't a priority 👍 

We could probably relax configuration for those affected tests as a workaround with `postfix-main.cf`, which should work.

target/postfix/main.cf
```diff
@@ -51,10 +51,11 @@ smtpd_helo_required = yes
 smtpd_delay_reject = yes
 smtpd_helo_restrictions = permit_mynetworks, reject_invalid_helo_hostname, permit
 smtpd_relay_restrictions = permit_mynetworks permit_sasl_authenticated defer_unauth_destination
-smtpd_recipient_restrictions = permit_sasl_authenticated, permit_mynetworks, reject_unauth_destination, reject_unauth_pipelining, reject_invalid_helo_hostname, reject_non_fqdn_helo_hostname, reject_unknown_recipient_domain
-smtpd_client_restrictions = permit_mynetworks, permit_sasl_authenticated, reject_unauth_destination, reject_unauth_pipelining
+smtpd_recipient_restrictions = permit_sasl_authenticated, permit_mynetworks, reject_unauth_destination, reject_invalid_helo_hostname, reject_non_fqdn_helo_hostname, reject_unknown_recipient_domain
+smtpd_client_restrictions = permit_mynetworks, permit_sasl_authenticated, reject_unauth_destination
 smtpd_sender_restrictions = $dms_smtpd_sender_restrictions
 smtpd_discard_ehlo_keywords = silent-discard, dsn
+smtpd_data_restrictions = reject_unauth_pipelining
```
polarathene, line 58: Like the long-term fix for SMTP smuggling, it defaults trust to `$mynetworks` and we aren't likely to mind with authenticated connections, so this may work?

```suggestion
smtpd_data_restrictions = permit_mynetworks, permit_sasl_authenticated, reject_unauth_pipelining
```

I actually thought if they were permitted earlier that it'd cascade through the restrictions, but perhaps that isn't the case 🤔

polarathene (commented):
This may work.

target/amavis/postfix-amavis.cf
```diff
@@ -23,7 +23,6 @@ smtp-amavis     unix    -       -       n       -       2       smtp
   -o smtpd_helo_restrictions=
   -o smtpd_sender_restrictions=
   -o smtpd_recipient_restrictions=permit_mynetworks,reject
-  -o smtpd_data_restrictions=reject_unauth_pipelining
```
polarathene, line 26: This restriction should remain here explicitly, as the issue explains.

It's not necessary, but easier to maintain if we don't have to think about what the implicit config inherited here is, especially when we expect it to be this.

If the change to skip restriction by trust does work, then all the more reason for leaving this as-is for Amavis probably? (_I haven't given much thought to the Amavis config we have here, I'd just rather not tamper with it_)

polarathene (changes requested):

polarathene:
Tests pass now. Just restore the Amavis config line + changelog entry and this should be good to go 🚀 

There is an existing changelog line for the SMTP smuggling fix that is related to this change. Although we didn't try the adding trust in our previous attempt so it may not matter 🤷‍♂️

target/amavis/postfix-amavis.cf
georglauterbach: Done

casperklein (dismissed):

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ DMS is now secured against the [recently published spoofing attack "SMTP Smuggli
   - `swaks` handles pipelining correctly, hence we can now use `reject_unauth_pipelining` in Postfix's configuration.
   - `swaks` provides better CLI options that make many files superflous.
   - `swaks` can also replace `openssl s_client` and handles authentication on submission ports better.
+- The Postfix configuration was adjusted to reject unauthorized pipelining, now in `smtpd_data_restrictions` (i.e. at the end of the mail transfer transaction) ([#3744](https://github.com/docker-mailserver/docker-mailserver/pull/3744))
```
polarathene, line 28:
```suggestion
- **Postfix:**
  - We now defer rejection from unauthorized pipelining until the SMTP `DATA` command via `smtpd_data_restrictions` (_i.e. at the end of the mail transfer transaction_) ([#3744](https://github.com/docker-mailserver/docker-mailserver/pull/3744))
    - Prevously our configuration only handled this during the client and recipient restriction stages. Postfix will flag this activity when encountered, but the rejection now is handled at `DATA` where unauthorized pipelining would have been valid from this point.
    - If you had the Amavis service enabled (default), this restriction was already in place. Otherwise the concerns expressed with `smtpd_data_restrictions = reject_unauth_pipelining` from the security section above apply. We have permitted trusted clients (_`$mynetworks` or authenticated_) to bypass this restriction.
```

polarathene (changes requested):

polarathene (approved):

PR #3756, opened 2024-01-07
docs: SpamAssassin ENV docs refactor

polarathene:
# Description

- ENV name logged was incomplete (`SPAM_TO_INBOX=1`).
- SA related ENV docs revised.
  - Some use of two separate lists (_ordered/unordered mix or not_) would be merged by the markdown renderer (_mkdocs one isn't particularly good for lists_), even with the blank line to separate them. Worked around issue with content to separate the lists.
  - More details for settings, notably `SA_KILL`, linked references to other related ENV.
  - `SA_KILL` existing admonition about quarantine and `ONE_DIR` was [introduced in Aug 2019](https://github.com/docker-mailserver/docker-mailserver/pull/1233). Revised.
- An rspamd focused PR [introduced an invalid `'warning'` log level](https://github.com/docker-mailserver/docker-mailserver/commit/e58dd1b95b3552ea850dad6b2ef0b87f8a9bae78#diff-9136f4c9b13b5b917de16079fe75026c076a72bc276eaf58a0dd297df5285608R286) missed in review, this has been corrected to `'warn'`.

Fixes: #2491

---

**References:**
- Log line introduced in: https://github.com/docker-mailserver/docker-mailserver/pull/3360
- [`SPAMASSASSIN_SPAM_TO_INBOX` default changed from `0` to `1`](https://github.com/docker-mailserver/docker-mailserver/pull/2361) in [DMS `v10.5.0`](https://github.com/docker-mailserver/docker-mailserver/blob/0eb4ac771498e1eff91944e79c19c8605995bc07/CHANGELOG.md?plain=1#L574) (Mar 2022)
  - Reasoning to change to [preferring to pass spam instead of bounce/reject](https://github.com/docker-mailserver/docker-mailserver/issues/1396#issuecomment-597210154) seems valid.
- `SA_KILL` default changed in [DMS `v12.0.0`](https://github.com/docker-mailserver/docker-mailserver/blob/0eb4ac771498e1eff91944e79c19c8605995bc07/CHANGELOG.md?plain=1#L308) (April 2023), from `6.31` (_same as `SA_TAG2`_) to `10.0` (_thus not all mail marked as spam has a quarantine copy_).
  - With a comment of mine summarizing related history/insights: https://github.com/docker-mailserver/docker-mailserver/pull/3058#issuecomment-1420268148
[...]

PR #3767, opened 2024-01-10
fix: Revert quoting `SA_SPAM_SUBJECT` in `mailserver.env`

### LINKED ISSUE #3763: DKIM Setup with Rspamd Broken
### 📝 Preliminary Checks

### 👀 What Happened?

ENABLE_RSPAMD=1
ENABLE_OPENDKIM=0

Run:

```
root@mail:/usr/local/bin# setup config dkim help
/etc/dms-settings: line 68: : command not found
[  ERROR  ]  Unexpected error occured :: script = /usr/local/bin/rspamd-dkim  | function = none (global) | command = SA_SPAM_SUBJECT=''***SPAM*** '' | line = 68 | exit code = 127
```

`/usr/local/bin/rspamd-dkim` sources `/etc/dms-settings`. If you look at line 68:

```
SA_SPAM_SUBJECT=''***SPAM*** ''
```
Notice the duplicated single quotes.

### 👟 Reproduction Steps

_No response_

### 🐋 DMS Version

v13.0.1

Fedora 39 x64

```yml
I think the important bits are enabling rspamd and disabling opendkim:

ENABLE_RSPAMD=1
ENABLE_OPENDKIM=0
```

### 📜 Relevant log output

_No response_

_No response_

polarathene:
# Description

- This was introduced in https://github.com/docker-mailserver/docker-mailserver/pull/3756
- I'm reverting it due to https://github.com/docker-mailserver/docker-mailserver/issues/3763
- I don't know how to maintain the trailing space if quote wrapped value isn't supported by this file (_although I suspect the actual issue is more to do with how it's being parsed / [stored](https://github.com/docker-mailserver/docker-mailserver/issues/3763#issuecomment-1883817879)? I had fun with that when [tackling LDAP config replacements via ENV](https://github.com/docker-mailserver/docker-mailserver/pull/3524/files#diff-ec07d4a05d16c4dd121c2b1b233a866ae469c65e986e6f9f7ef8e40378586ba3)_)
- A linter for `mailserver.env` could be added to the lint process: https://github.com/dotenv-linter/dotenv-linter
  - Implies quotes are only ok if whitespace is between values: https://dotenv-linter.github.io/#/checks/quote_character
  - Forbids trailing spaces: https://dotenv-linter.github.io/#/checks/trailing_whitespace

Alternatively, commenting out the value in `mailserver.env` would allow for the implicit default (_which has been that way for over 2 years_):

https://github.com/docker-mailserver/docker-mailserver/blob/aba218e6d7f30fbf4717178581c3e04537bea178/target/scripts/startup/variables-stack.sh#L70

I'm personally not a fan of `mailserver.env` myself and find it more of a maintenance burden + noise with reports (_where barely any of it is changed and the values represent defaults...but not always historically_).

Fixes #3763

## Type of change

## Checklist:

Adding an entry to `CHANGELOG.md` would be redundant noise, so I've opted-out of that.
- Unreleased change that affected a user on an older release since it's a repo file not associated to our image releases.
- This PR and the related issue backlink reference the PR that introduced the change for v14.


casperklein:
I don't think we need/should revert the quoting:

- The syntax currently used is [valid](https://docs.docker.com/compose/environment-variables/env-file/).
- I cannot reproduce the issue using "docker compose". However, I can reproduce the issue with `podman-compose`.

This should be fixed upstream in `podman-compose`.

polarathene:
> This should be fixed upstream in [podman-compose](https://github.com/containers/podman-compose).

Ah ok, I don't think I had that context from the issue at the time I put this together 😅 

I'm fine leaving it with quotes then (_which I expected to work and don't know how you'd implement this otherwise, unless a trailing space without quotes works_). It's not a file that's part of our releases so users affected can change it directly 👍

cfis:
Seems like I'm missing something, but wouldn't you end up with 2 single quotes in docker compose too? And that is what breaks the bash script.

See https://github.com/docker-mailserver/docker-mailserver/issues/3763#issuecomment-1884188500

polarathene:
> wouldn't you end up with 2 single quotes in docker compose too? And that is what breaks the bash script.

You won't due to the referenced docs for how Docker Compose handles the `env_file` support.

The equivalent `--env-file` with `docker run` however does not parse the key/value pairs this way, which likewise introduces the single quotes into the ENV value string.

mailserver.env
```diff
@@ -395,7 +395,7 @@ SA_TAG2=6.31
 SA_KILL=10.0
 
 # add tag to subject if spam detected
-SA_SPAM_SUBJECT='***SPAM*** '
+SA_SPAM_SUBJECT=***SPAM***
```
polarathene, line 398: - I have verified with `docker run --env-file` where the same issue associated to `podman-compose` occurs.
- `docker compose` with a `compose.yaml` using `env_file` parses the ENV file properly.

Without quotes the trailing white-space is still part of the value.

```env
# Trailing white-space kept with `docker run --env-file`:
SA_SPAM_SUBJECT=***SPAM*** 

# `docker compose` strips the white-space and expects quote wrapping to preserve it, like a shell variable:
SA_SPAM_SUBJECT='***SPAM*** '
```

```suggestion
#SA_SPAM_SUBJECT='***SPAM*** '
```

```suggestion
# `undef` opts-out of this feature (useful since by default spam is delivered to junk folder already)
SA_SPAM_SUBJECT=undef
```

georglauterbach: I'm fine with changing the default to `undef`.

casperklein: > The equivalent --env-file with docker run however does not parse the key/value pairs this way, which likewise introduces the single quotes into the ENV value string.

Good finding 👍

---
> SA_SPAM_SUBJECT=undef

With this change, the actual Subject will be `'undef'`, instead of `'***SPAM*** '`

I propose one of these solutions (prefer the last one):

```
SA_SPAM_SUBJECT=
```

or
```
#SA_SPAM_SUBJECT='***SPAM*** '
```

Both work fine with `docker run / docker compose` and doesn't change our default value.

georglauterbach: Since this seems to be rather "old" code, `undef` seems to a special value:

https://github.com/docker-mailserver/docker-mailserver/blob/aba218e6d7f30fbf4717178581c3e04537bea178/target/scripts/startup/setup.d/security/misc.sh#L87-L93

and according to [our own docs](https://docker-mailserver.github.io/docker-mailserver/edge/config/environment/#sa_spam_subject), `undef` "opt[s]-out of modifying the subject for mail marked as spam".

georglauterbach:
Can we additionally apply this diff here:

```diff
--- a/target/scripts/startup/setup.d/security/misc.sh
+++ b/target/scripts/startup/setup.d/security/misc.sh
@@ -71,13 +71,13 @@ function __setup__security__spamassassin() {
     _log 'debug' 'Enabling and configuring SpamAssassin'
 
     # shellcheck disable=SC2016
-    sed -i -r 's|^\$sa_tag_level_deflt (.*);|\$sa_tag_level_deflt = '"${SA_TAG}"';|g' /etc/amavis/conf.d/20-debian_defaults
+    sed -i -r "s|^\$sa_tag_level_deflt (.*);|\$sa_tag_level_deflt = \"${SA_TAG}\";|g" /etc/amavis/conf.d/20-debian_defaults
 
     # shellcheck disable=SC2016
-    sed -i -r 's|^\$sa_tag2_level_deflt (.*);|\$sa_tag2_level_deflt = '"${SA_TAG2}"';|g' /etc/amavis/conf.d/20-debian_defaults
+    sed -i -r "s|^\$sa_tag2_level_deflt (.*);|\$sa_tag2_level_deflt = \"${SA_TAG2}\";|g" /etc/amavis/conf.d/20-debian_defaults
 
     # shellcheck disable=SC2016
-    sed -i -r 's|^\$sa_kill_level_deflt (.*);|\$sa_kill_level_deflt = '"${SA_KILL}"';|g' /etc/amavis/conf.d/20-debian_defaults
+    sed -i -r "s|^\$sa_kill_level_deflt (.*);|\$sa_kill_level_deflt = \"${SA_KILL}\";|g" /etc/amavis/conf.d/20-debian_defaults
 
     # fix cron.daily for spamassassin
     sed -i \
@@ -89,7 +89,7 @@ function __setup__security__spamassassin() {
       sed -i -r 's|^\$sa_spam_subject_tag (.*);|\$sa_spam_subject_tag = undef;|g' /etc/amavis/conf.d/20-debian_defaults
     else
       # shellcheck disable=SC2016
-      sed -i -r 's|^\$sa_spam_subject_tag (.*);|\$sa_spam_subject_tag = '"'${SA_SPAM_SUBJECT}'"';|g' /etc/amavis/conf.d/20-debian_defaults
+      sed -i -r "s|^\$sa_spam_subject_tag (.*);|\$sa_spam_subject_tag = \"${SA_SPAM_SUBJECT}\";|g" /etc/amavis/conf.d/20-debian_defaults
     fi
 
     # activate short circuits when SA BAYES is certain it has spam or ham.
```

While reading through the code, I saw the horrendous quotes just above this change. Maybe the `\"` isn't even required (when defining something like `$sa_kill_level_deflt = SOME_VALUE;` for example.

mailserver.env
casperklein: > Since this seems to be rather "old" code, undef seems to be a special value:

Thanks for pointing that out 👍

I still think the current SPAM subject line is useful and should not be emptied by setting `undef`.

> `undef` opts-out of this feature (useful since by default spam is delivered to junk folder already

This is only true when using IMAP. In POP3, there are no folders.

casperklein:
I think that will not work (not 100% sure, because not tested):

`sed -i -r "s|^\$sa_kill_level_deflt ...`

This will be exanded by bash to `'s|^$sa_kill_level_deflt ...'`

Bash didn't try to replace $sa_kill_level_deflt with a value, because of the backslash before the dollar sign. So far so good. But the $ sign is still there (indicating a line end, when not escaped), which will be a problem for `sed`.

To get the same functionality, you would have to use `sed -i -r "s|^\\$sa_kill_level_deflt ...` I think. If that's more clear then the current solution is probably a personal opinion for everyone.

PS: Every you touch an sed statement, think about switching to sedfile :-)

mailserver.env
polarathene: Ok, but as POP3 would also require explicit opt-out of the dovecot sieve rule to move spam to junk folder, wouldn't that make the concern irrelevant?

We could cover this on docs for changes to support POP3, but `undef` is perhaps a better default for most users?

polarathene:
> To get the same functionality, you would have to use `sed -i -r "s|^\\$sa_kill_level_deflt ...` I think

Need an extra `\` 😓 

```bash
# Current approach:
sed -i -r 's|^\$sa_tag_level_deflt (.*);|\$sa_tag_level_deflt = '"${SA_TAG}"';|g' /path/to/file

# Triple `\` required:
# 2 for escape the `$` regex for sed (\ needs to escape itself)
# 1 for escaping shell interpolation:
sed -i -r "s|^\\\$sa_tag_level_deflt (.*);|\\\$sa_tag_level_deflt = ${SA_TAG};|g" /path/to/file

# Escape $ regex on left(before), while right(after) escapes $ with $ + \ to escape $ from shell interpolation
sd '^\$sa_tag_level_deflt (.*);' "\$\$sa_tag_level_deflt = ${SA_TAG};" /path/to/file

# Alternatively, capture the prefix match to minimize escaping:
sed -i -r "s|^(\\\$sa_tag_level_deflt).*;|\1 = ${SA_TAG};|g" /path/to/file
sd '^(\$sa_tag_level_deflt).*;' "\$1 = ${SA_TAG};" /path/to/file
```

`sd` is a little nicer syntax wise, it's a [`sed` replacement written in rust](https://github.com/chmln/sd) (_see their README for comparisons to equivalent `sed` commands_).

The alternative `sed` approach isn't too bad either👍

mailserver.env
casperklein: > Ok, but as POP3 would also require explicit opt-out of the dovecot sieve rule to move spam to junk folder, wouldn't that make the concern irrelevant?

Not sure if I understand that correctly. If one goes with POP3 instead of IMAP, `MOVE_SPAM_TO_JUNK=1` makes no sense. There is no point in enabling it. So not indicating spam via the subject (undef), will make spam mails not easy to identify as spam mails on a first glance for pop3 users.
The current solution, to prefix `'***SPAM*** '` to the subject is a sane IMO. I think changing that and mention it in the docs so people can revert back, is not something the majority will benefit from. Currently only using `podman-compose` (which we not support) or `docker run` leads to an issue. I think not many people are using one of those. Otherwise there would be more reports.
Instead of changing IMO good defaults, to satisfy buggy implementations, I propose `#SA_SPAM_SUBJECT='***SPAM*** '`. This reliefs the pain for those affected by the bug and also keeps the status quo for all others.

I found [this](https://github.com/moby/moby/issues/26009) open issue for your `docker run --env-file` observation.

casperklein:
I would stick with the current approach. Three `\` makes it not clearer to me than the quotes 🤷 But I am fine with both.

While `sd` looks nice, introducing another dependency with a new syntax seems overkill to me 🚀

georglauterbach:
I see; I did not think it would be that complicated... thanks for the clarification!

mailserver.env
georglauterbach: I think the point that @casperklein is stressing is that with POP3, there are no directories, so `MOVE_SPAM_TO_JUNK` is useless (because there is no such thing as "Junk" in POP3)? Hence, the ENV `SA_SPAM_SUBJECT` is quite important with POP3 for the distinction. That's how I understood it.

polarathene: > That's how I understood it.

Yeah no worries. I could be misunderstanding it myself since I've not setup POP3.

I thought POP3 would only pull mail from the inbox, and folders like junk would be ignored (_thus no spam retrieved_). If all mail from the entire mailbox is pulled regardless of separate folders I guess that makes sense.

casperklein: > So a user that wants to use POP3 must explicitly disable these?

There is no explicit need to disable it. When enabled, it just have no effect, because there are no Folders in POP3. You assume, that IMAP users don't want the spam subject. I am not sure if that is really the case. I looked through various bug reports with their posted `compose.yml/mailserver.env` files and couldn't find a single one, where someone used `SA_SPAM_SUBJECT=` or `SA_SPAM_SUBJECT=undef`. I think the majority prefer the spam subject. If some don't want that, they can change it with less effort. Yes, the same would go for POP3 users to enable that.
But DMS should work for all users out of the box with sane defaults. Regardless of IMAP/POP3, marking SPAM messages for all users is what I consider sane. Everyone is free to opt out. This is better instead of every POP3 user must opt in to mark spam mails.

> I don't understand why adjusting the defaults to better suit IMAP is an issue.

How to you come to this confidence, that the change will better suits for IMAP users? Altering the subject is a common practice and I've never heard that this is a bad one.
BTW: I don't understand, why we even consider to change defaults (where there is no public demand of any kind for it), just because of a single issue, that we tracked down to a buggy implementation (not a DMS issue in general).

> so I'm not surprised that there's not been much reports, but the one we received wasn't that long after the change was made.

👍 Personally I still think, the largest part of the users use `docker compose`. Using `docker run` with lot's of ENVs passed to feels weird 😆 Never the less, the issue can be fixed without changing defaults.

---

Enough time invested for such a tiny change 😆 I think all thoughts were mentioned. Unless there is another killer reason to be mentioned, I stay with my opinion. You and Georg can decide how to proceed with this PR 👍

georglauterbach: Simple question? Can't we just insert the space after the spam prefix in our scripts? This way, you can write whatever you want in there, we could even leave the ENV file as is and we'd only need to insert a space in our scripts.

polarathene: **TL;DR:**
- POP3 does not retrieve spam that gets sent to the junk folder by default. This was my assumption and I've now verified it. Thus `SA_SPAM_SUBJECT` isn't providing value until opt-out of the spam to junk feature.
- I don't know of any major/popular services personally that modify the subject line of mail when they isolate spam to a junk folder.
- Some POP3 clients like ThunderBird can identify spam from the mail spam headers DMS adds, and can move that into a local junk folder. Reducing the value of `SA_SPAM_SUBJECT`.
- Just because you find no users changing `SA_SPAM_SUBJECT` to be `undef` does not mean users want the subject line modification. They may just not care about it or notice it with our defaults, additionally our docs weren't the best at communicating awareness of the feature.

---

> There is no explicit need to disable it. When enabled, it just have no effect, because there are no Folders in POP3.

```bash
# Run a basic DMS instance:
docker run --rm -it \
  --hostname mail.example.test \
  --env PERMIT_DOCKER=container \
  --env ENABLE_SPAMASSASSIN=1 \
  --env SA_TAG2=3 \
  --env ENABLE_POP3=1
  --name dms \
  -p 110:110 -p 143:143 -p 587:587 \
  mailserver/docker-mailserver

# In the container create an account and send a legit mail and a spam:
setup email add hello@example.test bad-password
# This will be legit, spam score was around 1.2:
swaks --server 0.0.0.0 --to hello@example.test --body 'hello, your code is 42'
# This will trigger spam, score was around 3-4:
swaks --server 0.0.0.0 --to hello@example.test --body 'spam' --data 'Date: %DATE%\nTo: %TO_ADDRESS%\nFrom: %FROM_ADDRESS%\nSubject: test\n%NEW_HEADERS%\n%BODY%\n' --header
'Message-Id: <spam@example.test>'
```

Setup ThunderBird with POP3 account, I have no custom DNS here, just the DMS container as above running from WSL2 and ThunderBird connecting to it over localhost port without TLS (_`None` was required to avoid TLS issue_).

image

Right-click on the account address/name on the sidebar and "Get Messages", only the mail that didn't land in the Junk folder is available:

image

Yet the DMS logs show both mails sent were delivered and stored as expected:

```
Jan 14 21:33:41 mail dovecot: lmtp(hello@example.test)<2300><cg/dHjVTpGX8CAAAUi6ngw>: sieve: msgid=<20240114213339.002286@mail.example.test>: stored mail into mailbox 'INBOX'
Jan 14 21:33:47 mail dovecot: lmtp(hello@example.test)<2300><sNGlNztTpGX8CAAAUi6ngw>: sieve: msgid=<spam@example.test>: fileinto action: stored mail into mailbox 'Junk'
```

This meets the assumption I had. POP3 users won't get such mail by default regardless. They would need to opt-out of the spam to junk defaults we have. Hence it's probably worth documenting that, as `SA_SPAM_SUBJECT` does not have any relevance to them by default until they opt-out of the junk folder.

`SA_SPAM_SUBJECT` is mostly redundant then, when you have all mail that would have that marking separated into the junk folder by default. That's why `SA_SPAM_SUBJECT=undef` made sense to me.

---

> You assume, that IMAP users don't want the spam subject. I am not sure if that is really the case.
> I looked through various bug reports with their posted `compose.yml` / `mailserver.env` files and couldn't find a single one, where someone used `SA_SPAM_SUBJECT=` or `SA_SPAM_SUBJECT=undef`.

The lack of it doesn't mean that users were happy with `SA_SPAM_SUBJECT` default. They may not have cared either way, or weren't aware that it could be easily customized. I only recently improved docs on this to help communicate that better.

Did you find anyone who actually customized the default to something else? I can only imagine it's very small amount, since there should be at least one from someone who wanted/implemented the feature in the first place. Along with the others that added additional contributions to the original wiki with advice on how to get white-space separator or the spam score. This is understandable since they are users actively interested in the feature.

Since such mail goes to the junk folder by default, I'd wager many users just don't think about it or see it much that it concerns them? Even by looking at the junk folder, they'll have all mails listed with a consistent prefix, so it's less likely to stand out or look odd to them. That doesn't mean they actually want that and would actively opt-in to that if we defaulted to `undef`.

---

> I think the majority prefer the spam subject. If some don't want that, they can change it with less effort.
> Yes, the same would go for POP3 users to enable that.
> But DMS should work for all users out of the box with sane defaults.

We can either document the better defaults for POP3, or you can conditionally adjust defaults 🤷‍♂️ 

I'd just like to know the value of the subject prefix adding `SPAM` to mail that gets lumped into a junk folder by default and as shown above does not get retrieved when POP3 is used.

---

> Regardless of IMAP/POP3, marking SPAM messages for all users is what I consider sane.
> Everyone is free to opt out. This is better instead of every POP3 user must opt in to mark spam mails.

POP3 is already an opt-in feature. It's not a big ask (_especially given the above findings_) to inform the user of other ENV to change. `SA_SPAM_SUBJECT` will be more relevant to the POP3 user, so it also makes more sense for them to customize that to what they prefer.

I don't mind the default being adjusted from `undef` when POP3 is enabled (_or rather spam to junk is disabled_). That makes more sense as sane defaults to me 🤔 

> How to you come to this confidence, that the change will better suits for IMAP users? Altering the subject is a common practice and I've never heard that this is a bad one.

I only use Gmail and Outlook/Live, but neither of these alter the subject line AFAIK? What popular services are doing this for you when they isolate detected spam mail to a junk folder?

In the case of ThunderBird, here are the Junk settings it offers which would be relevant to POP3 users. It has it's own spam analysis and moving spam to junk folder features. Notably (disabled by default) is that "Trust junk mail headers" feature, which will identify spam when DMS marks it with spam headers... that's much nicer, it's implicit and works just as well without modifying a subject line.

image

Other POP3 clients perhaps aren't as capable of such features, so having a clearly communicated SPAM prefix to your mails subject line can make sense. I'm just showing that it's not necessarily important to POP3 clients either.

polarathene: > Simple question: Can't we just insert the space after the spam prefix in our scripts? This way, you can write whatever you want in there

What if a user wants no trailing white-space? 🤷‍♂️ 

Probably unlikely though, so that would be nice... however the feature is already in use and this could introduce additional white-space that's already been accounted for, or you add maintenance burden with extra conditional logic :\

---

> I'll leave it up to @polarathene to decide what to do with this - I'll approve as I think we have multiple valid solutions.

I personally think `SA_SPAM_SUBJECT=undef` makes sense, it'd be more consistent with rspamd which will replace SA AFAIK (_unless there are plans to use the integration available with rspamd?_), so this sort of breaking change is probably going to land eventually anyway..

I'm not a fan of `mailserver.env`, I'd rather remove it entirely but I'll resolve this by commenting out the ENV.

polarathene, line 398:
```suggestion
# The value `undef` opts-out of this feature. The value shown below is the default.
# NOTE: By default spam is delivered to a junk folder, reducing the value of adding a subject prefix.
# NOTE: If not using Docker Compose, other CRI may require the single quotes removed.
#SA_SPAM_SUBJECT='***SPAM*** '
```

polarathene (commented):
I'll go with this change 👍 

For anyone landing on this PR, besides the ENV parser compatibility issue, there was a concern primarily regarding POP3 users benefiting from the feature, but [DMS defaults prevent `SA_SPAM_SUBJECT` being useful, unless opting out of the default spam to junk folder behavior](https://github.com/docker-mailserver/docker-mailserver/pull/3767/files#r1451811263).

georglauterbach (approved):

mailserver.env
casperklein: Thanks for taking the time and doing extended testing!

> POP3 does not retrieve spam that gets sent to the junk folder by default. This was my assumption and I've now verified it. Thus SA_SPAM_SUBJECT isn't providing value until opt-out of the spam to junk feature.

I didn't know that, I had always MOVE_SPAM_TO_JUNK=0. That's why I've never had problems and got the spam mails. So in a POP3 only setup, the user does not have access to the junk folder. Mails ending up there, are not accessable. That means the user is not aware of mails tagged as spam + misses false/positives. This could be fixed by printing a warning `if IMAP=0 and POP3=1 and MOVE_SPAM_TO_JUNK=1`

> Just because you find no users changing SA_SPAM_SUBJECT to be undef does not mean users want the subject line modification.

100% ack. But the same goes for the opposite :-D

> I'm not a fan of mailserver.env, I'd rather remove it entirely but I'll resolve this by commenting out the ENV.

Same here, I apply my ENVs via the compose file. Only while testing, I found it sometimes useful, because I didn't have to open the online documentation, when I am not exactly sure, what a ENV does. If we decide to remove the .env file, we could alternatively provide an additional `compose.override.yml` file (like [here](https://github.com/netbox-community/netbox-docker/blob/release/docker-compose.override.yml.example)) with just all ENVs listed. That way, you can see all ENVs available + work arround the quoting issues with `docker run ..` using an .env file

polarathene: > Thanks for taking the time and doing extended testing!

❤️ 

> That means the user is not aware of mails tagged as spam + misses false/positives. This could be fixed by printing a warning `if IMAP=0 and POP3=1 and MOVE_SPAM_TO_JUNK=1`

A warning is probably a good addition to raising that awareness yep 👍 

---

> If we decide to remove the .env file, we could alternatively provide an additional compose.override.yml file (like [here](https://github.com/netbox-community/netbox-docker/blob/release/docker-compose.override.yml.example)) with just all ENVs listed

I think that is a slight improvement, but my main gripe is maintaining both. IIRC they're not in sync with the docs (_which can also better convey extra information_). I also loathe user reports that dump the whole file in the report where we're left to figure out what they explicitly changed from the defaults (since many would assign a default value)

---

> Only while testing, I found it sometimes useful, because I didn't have to open the online documentation, when I am not exactly sure, what a ENV does

I can understand that convenience, but really it's the difference of browsing a file (via some app or terminal) vs a docs page.

The main issue with the docs page is there's a bit more noise going on, it doesn't feel as nice to navigate/explore for an overview of ENV.

We could better organize that further, and perhaps throw in a table overview at the top with key/value columns to show the defaults and each key linking to the respective ENV setting for more information.

---

> Probably because they want their users to use IMAP? As we found out, modifying the subject is mostly useful with POP3. But for security reasons, not everyone is fine with IMAP.

👍 I was just asking as you gave the impression it was common practice even with a Junk folder storing the spam 😝

georglauterbach:
I should have added the correct label; now this PR was merged even though we had a freeze in place for v13.3. What shall we do about this? IMI this change is small enough to not affect the freeze.

polarathene:
I don't consider `mailserver.env` part of our releases. It's only by association via the tag, but unrelated to the published images / docs.

PR #3779, opened 2024-01-15
fix: Ensure `.svbin` files are newer than `.sieve` source files

polarathene:
# Description

This change prevents operations applied to both files that adjusts their `mtime` attribute which sometimes results in the same timestamp which causes Dovecot to log errors.

Functionally there doesn't appear to be any issue. After logging the error, the existing svbin is used and works correctly.

This does help avoid false-positives in the test-suite that are asserting for errors in a log.

Fixes: https://github.com/docker-mailserver/docker-mailserver/issues/3558#issuecomment-1880392732

## Context

This fix prevents the two error log lines below that are preceding the log where the sieve script is run successfully regardless:

```console
# Each logline prefixed with
Jan  8 02:13:59 mail dovecot: lmtp(hello@example.test)<1319><34kpAWdam2UnBQAAUi6ngw>: 

$ grep -i junk /var/log/mail/mail.log
Error: sieve: binary /usr/lib/dovecot/sieve-global/after/spam_to_junk.svbin: save: failed to create temporary file: open(/usr/lib/dovecot/sieve-global/after/spam_to_junk.svbin.) failed: Permission denied (euid=5000(docker) egid=5000(docker) missing +w perm: /usr/lib/dovecot/sieve-global/after, dir owned by 110:0 mode=0755)
Error: sieve: The LDA Sieve plugin does not have permission to save global Sieve script binaries; global Sieve scripts like `/usr/lib/dovecot/sieve-global/after/spam_to_junk.sieve' need to be pre-compiled using the sievec tool

sieve: msgid=<20240108021357.001310@mail.example.test>: fileinto action: stored mail into mailbox 'Junk'
```

I've only witnessed this personally on my current system (AMD 7940HS with Windows 11 + WSL2), it's possibly related to the filesystem layer and changes happening within a certain window to be coalesced into the same timestamp as an optimization? 🤷‍♂️ 

If the `.sieve` file is newer (_or the same timestamp in this case_), Dovecot will try compile a new `.svbin` (_and presumably fallback to the existing one when that fails_).

We can see that the failure is due to the VMail user `docker:docker` / `5000:5000` lacking permissions to write `.svbin` as the directory ownership is `dovecot:root` / `110:0` with `755` permissions.

- As `misc.sh`, `dovecot.sh`, `rspamd.sh` all run `sievec` for `.sieve` files in either `/usr/lib/dovecot/sieve-pipe` or `/usr/lib/dovecot/sieve-global`. As such, while we could adjust the permission for the VMail static user, it won't support deployments with accounts that have their own unique UID (_some LDAP users are known to have this setup_).
- Instead it seems correct to ensure the `mtime` is adjusted as this PR does, as we are intentionally compiling these during startup scripts and there should be no need to compile again afterwards when mail is received and processed by these scripts.

## Type of change

## Checklist:


target/scripts/startup/setup-stack.sh
```diff
@@ -20,6 +20,16 @@ function _setup() {
...
 
+  _setup_post
+}
+
+function _setup_post() {
+  # Dovecot `.svbin` files must have a newer mtime than their `.sieve` source files,
+  # Modifications during setup to these files sometimes results in a common mtime value.
+  # Handled during post-setup as setup of Dovecot Sieve scripts is not centralized.
+  find /usr/lib/dovecot/ -iname *.sieve -exec touch -d '2 seconds ago' {} +
+  find /usr/lib/dovecot/ -iname *.svbin -exec touch -d '1 seconds ago' {} +
```
polarathene, line 31: Appeasing the linting gods 🙏 

```suggestion
  find /usr/lib/dovecot/ -iname '*.sieve' -exec touch -d '2 seconds ago' {} +
  find /usr/lib/dovecot/ -iname '*.svbin' -exec touch -d '1 seconds ago' {} +
```

georglauterbach (approved):

target/scripts/startup/setup-stack.sh
georglauterbach: This is actually very valuable because Bash would expand the star as it is a glob 🌝

georglauterbach:
I merged this right away so we can put it into v13.3.0.

PR #3783, opened 2024-01-16
fix: Ensure `setup dkim` generates DKIM keys with ownership matching the parent directory

ap-wtioit:
# Description

dkim key directories and files in /tmp/docker-mailserver/opendkim/keys/ are now created (chowned) with the same user / group as /tmp/docker-mailserver/opendkim/keys/. this means invoking `./setup.sh dkim domain example.com` no longer creates files as root:root when invoking inside a directory not owned by root.

Fixes #

## Type of change

Not 100% sure could also be just an improvement.

## Checklist:

### DIFF
--- a/target/bin/open-dkim
+++ b/target/bin/open-dkim
@@ -144,6 +144,9 @@ while read -r DKIM_DOMAIN; do
       --directory="/tmp/docker-mailserver/opendkim/keys/${DKIM_DOMAIN}"
   fi
 
+  # fix permissions to use the same user:group as /tmp/docker-mailserver/opendkim/keys
+  chown -R "$(stat -c '%U:%G' /tmp/docker-mailserver/opendkim/keys)" "/tmp/docker-mailserver/opendkim/keys/${DKIM_DOMAIN}"
+
   # write to KeyTable if necessary
   KEYTABLEENTRY="${SELECTOR}._domainkey.${DKIM_DOMAIN} ${DKIM_DOMAIN}:${SELECTOR}:/etc/opendkim/keys/${DKIM_DOMAIN}/${SELECTOR}.private"
   if [[ ! -f "/tmp/docker-mailserver/opendkim/KeyTable" ]]; then
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,8 @@ All notable changes to this project will be documented in this file. The format 
 
 ### Fixes
 
+- **Setup:**
+  - `setup` CLI - `setup dkim domain` now creates the keys files with the user owning the key directory ([#3783](https://github.com/docker-mailserver/docker-mailserver/pull/3783))
 - **Dovecot:**
   - During container startup for Dovecot Sieve, `.sievec` source files compiled to `.svbin` now have their `mtime` adjusted post setup to ensure it is always older than the associated `.svbin` file. This avoids superfluous error logs for sieve scripts that don't actually need to be compiled again ([#3779](https://github.com/docker-mailserver/docker-mailserver/pull/3779))
 - **Internal:**


georglauterbach:
We have seen the `ttrpc` in another PR recently as well (@polarathene). I will restart tests.

target/bin/open-dkim
```diff
@@ -144,6 +144,9 @@ while read -r DKIM_DOMAIN; do
       --directory="/tmp/docker-mailserver/opendkim/keys/${DKIM_DOMAIN}"
   fi
 
+  # fix permissions to use the same use as /tmp/docker-mailserver/opendkim/keys
```
georglauterbach, line 147:
```suggestion
  # fix permissions to use the same as /tmp/docker-mailserver/opendkim/keys
```

georglauterbach (approved):

target/bin/open-dkim
```diff
@@ -144,6 +144,9 @@ while read -r DKIM_DOMAIN; do
...
+  chown -R "$(stat -c '%U:%G' /tmp/docker-mailserver/opendkim/keys)" "/tmp/docker-mailserver/opendkim/keys/${DKIM_DOMAIN}"
```
polarathene, line 148: Just to clarify, earlier in this script at line 134:

```bash
mkdir -p "/tmp/docker-mailserver/opendkim/keys/${DKIM_DOMAIN}"
```

We create this parent location if it doesn't exist yet. So the only difference here with `chown` is that the user has  created this location in advance or adjusted the permissions since?

I am curious about the motivation for this as most content in the Docker config volume will be owned by `root:root` when it's created by DMS?

---

Docker hasn't landed official convenience options for the linux kernel mounting API feature (_[idmapped mounts](https://www.kernel.org/doc/html/latest/filesystems/idmappings.html#idmappings-when-creating-filesystem-objects)_), but [you can use it manually AFAIK](https://github.com/containers/podman/issues/10374#issuecomment-1559664583), while I think [podman implemented the feature for rootful volumes](https://github.com/containers/podman/issues/10374#issuecomment-1538909604).

For Docker I think it'll need to [wait for Containerd 2.0 release](https://github.com/containerd/containerd/issues/9310#issuecomment-1786033104), and then some release after Docker upgrades to that and implements the feature (_been a long wait since Kernel 5.12 introduced the feature in April 2021_).

ID mapped mounts will allow mapping `root` from the container easily to another user on the host with volume mounts, avoiding the need to chown effectively..

polarathene (dismissed):
I'm fine with the change 👍 

Note that it's desired in future to unify the DKIM key generation CLI helper for both OpenDKIM and Rspamd, which would also share a common location on the DMS Config volume. This sort of change would need to be carried over to that when that arrives in a future major release.

target/bin/open-dkim
ap-wtioit: in the [compose.yaml](https://github.com/docker-mailserver/docker-mailserver/blob/master/compose.yaml#L21) /tmp/docker-mailserver is mapped as a volume (we have an earlier version of the compose.yaml but it behaves the same) so when a user has a directory containing docker-data that belongs to him it is problematic if there is data created for root:root that the user cannot manage outside of the container (e.g. adding the dkim public keys to DNS).

target/bin/open-dkim
ap-wtioit: fixed with:
```
# fix permissions to use the same user:group as /tmp/docker-mailserver/opendkim/keys
```
somehow during testing i didn't fully finish this comment

target/bin/open-dkim
polarathene: Yes I understand, but this is admin not related to mail users per se. It can happen with other configs too, not just the DKIM keys generated.

The data should still be readable IIRC, and you rarely need to interact with this kind of data manually on the host for writes? Usually it's annoying if you want to move/rename or delete from the host due to the root ownership.

If you have ID mapped mounts however (_once they're in a release that makes it easier to configure_), this whole issue disappears.

polarathene (approved):

target/bin/open-dkim
ap-wtioit: yes, sorry my "user" is me the mail admin needing to manage the dkim keys ;-)

i wanted to make the smallest fix to fix my/our issue without affecting too much other stuff. ideally i would think DMS should respect the owner of /tmp/docker-mailserver when creating files there but that would be a much bigger change.

And not all data was readable (transfering private keys from the system i'm working on to the real server after testing was not working without using sudo chown locally, cleanup of CI systems failed without sudo chown).

polarathene: > ideally i would think DMS should respect the owner of /tmp/docker-mailserver when creating files there but that would be a much bigger change.

We could probably add ENV similar to the VMAIL_UID + VMAIL_GID actually and chown to that when necessary? Then it could work for all cases when you opt-in, without having to know the folder to create in advance (_or manually chown the first time_) which this fix relies on?

> cleanup of CI systems failed without sudo chown

Our tests have a `make clean` that is meant to handle that. If it doesn't we can fix that issue.

Our approach is to just run a container as root with the volume mounted and remove as root that way. I definitely remember the annoyance of relying on sudo from the Docker host in the past, while the tests prior did this on the host you'd have to provide your password upfront and I felt uncomfortable about the script removing content on the host system it had full access to vs removing with the restricted volume.

For CI it's less of an issue to just have a sudo step IMO 🤷‍♂️ 

Anyway this is fine and we can get this into v13.3 :)

polarathene:
Oh actually @georglauterbach may not approve of it getting into v13.3 🤔 

Is it ok, considering it's not core to DMS container, but rather a helper fix for the `setup cli`? Any existing users that have a root owned DKIM folder would still keep the root ownership, so it should be harmless to merge?

georglauterbach:
> Oh actually @georglauterbach may not approve of it getting into v13.3 🤔
>
> Is it ok, considering it's not core to DMS container, but rather a helper fix for the `setup cli`? Any existing users that have a root owned DKIM folder would still keep the root ownership, so it should be harmless to merge?

I'm fine with that, but this should really be the last PR to go into v13.3.0 :)

georglauterbach (approved):

polarathene:
> this should really be the last PR to go into v13.3.0 :)

I technically may try to squeeze in one for OAuth2 improvement to docs / test. Should be submitted within 48 hours, hopefully within 24 👍 (_docs + tests are valid to bypass the freeze, so that should be fine_).

PR #3747, opened 2024-01-04
tests: normalizations

georglauterbach:
# Description

See #3739 for what is fixed with this PR. Review commit by commit - I really put more effort into single commits this time, I promise @polarathene 😆 ❤️ 

Fixes #3739
Fixes #3733

## Type of change

- [x] Improvement (non-breaking change that does improve existing functionality)

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [ ] If necessary I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] **I have added information about changes made in this PR to `CHANGELOG.md`**

Issue #3554, opened 2023-09-28
[TODO]: Tests should share a more thorough helper method for mail delivery checks


polarathene:
### Subject

Something else that requires developers attention

### Description

Up for grabs for anyone who'd like to contribute, no rush :)

---

When [reviewing a PR that introduced widespread mail delivery failures](https://github.com/docker-mailserver/docker-mailserver/pull/3550/files#r1339564900) by changing UID/GID (_as Dovecot no longer had permission to write the mail to `/var/mail`_), I suggested a test-case that would better catch that common failure (_technically two, as [checking the mail logs for failures](https://github.com/docker-mailserver/docker-mailserver/pull/3550/files#r1339424344) would be complimentary at catching another set of errors_).

---

It would be nice to eventually do some cleanup across tests to have a common helper method for consistently checking successful mail delivery.
- At least when [only one mail is delivered in a test this approach works well](https://github.com/docker-mailserver/docker-mailserver/pull/3552) (_using a `timeout` with `tail` to avoid blocking on mail queue emptying, fails faster in a more helpful way_)
- An [alternative technique is to filter and group by account](https://github.com/docker-mailserver/docker-mailserver/blob/8c0cfa0836bcf20e78a2d5d1ae02270004ee0aed/test/tests/parallel/set3/mta/smtp_delivery.bats#L147-L166) (_ideally waiting for mail queue to empty first_)

[Failures were all shared here](https://github.com/docker-mailserver/docker-mailserver/pull/3550#issuecomment-1738509088), which should help track down which tests need the refactoring. I also identified some common keywords in the failure logs that could be checked. An [earlier comment](https://github.com/docker-mailserver/docker-mailserver/pull/3550#issuecomment-1738504868) provided more verbosity in logs (_for additional context_)

PR #3786, opened 2024-01-16
tests: new sending and filtering functions

georglauterbach:
# Description

This PR is intended to solve concerns raised in https://github.com/docker-mailserver/docker-mailserver/pull/3747#discussion_r1448558553 by @polarathene.

A new file for log filter helpers was created. Thereafter, `_send_email_and_get_id` was reworked to become `_send_email_with_mid`. I paid attention to using sub-shells only where necessary. This new way of sending + filtering allows for greater flexibility and the implementation has become easier. Last but not least, I updated the service log filters and applied them throughout all our tests to improve code quality.

Review commit by commit.

## Type of change

## Checklist:


test/helper/common.bash
```diff
@@ -413,62 +403,6 @@ function _should_have_content_in_directory() {
...
-# `_send_email_and_get_id`.
-#
-# @param ${1} = email ID
-# @param ${2} = container name [OPTIONAL]
-function _print_mail_log_for_id() {
-  local MAIL_ID=${1:?Mail ID must be provided}
-  local CONTAINER_NAME=$(__handle_container_name "${2:-}")
-
-  _run_in_container grep -E "${MAIL_ID}" /var/log/mail.log
-}
```
georglauterbach, line 470: This was dropped intentionally.

polarathene:
I'll review this after v13.3 is out, trying to wrap up some tasks in time for that 😅

georglauterbach:
> I'll review this after v13.3 is out, trying to wrap up some tasks in time for that 😅

Nice! This is basically just me implementing what you proposed in https://github.com/docker-mailserver/docker-mailserver/pull/3747#discussion_r1448558553.

test/helper/log_and_filtering.bash
```diff
@@ -0,0 +1,109 @@
...
+# in regular expressions, you need to escape them!
+function _service_log_should_not_contain_string() {
+  _filter_service_log "${@}"
+  assert_failure
+}
+
+# Filters the mail log according to MID (Message-ID) and prints lines
+# of the mail log that fit Postfix's queue ID for the given message ID.
+#
+# @param ${1} = message ID part before '@'
```
polarathene, line 77: `MSG_ID` is probably a bit clearer, so let's go with that 👍 

```suggestion
# Filters the mail log by the given MSG_ID (Message-ID) parameter,
# printing log lines which include the associated Postfix Queue ID.
#
# @param ${1} = The local-part of a Message-ID header value (`<local-part@domain-part>`)
```

test/helper/log_and_filtering.bash
```diff
@@ -0,0 +1,109 @@
...
+function _print_mail_log_of_queue_id_from_mid() {
+  # The unique ID Postfix (and other services) use may be different in length
+  # on different systems. Hence, we use a range to safely capture it.
+  local QUEUE_ID_REGEX='[A-Z0-9]{9,12}'
+
+  local MID=$(__construct_mid "${1:?Left-hand side of MID missing}")
+  shift 1
```
polarathene, line 84: Similar to comment on the other method below this one, this is Postfix specific with `message-id` field to find the Queue ID, unlike Dovecot which just matches on `msgid` field.

```suggestion
function _print_mail_log_of_queue_id_from_msgid() {
  # A unique ID Postfix generates for tracking queued mail as it's processed.
  # The length can vary (as per the postfix docs). Hence, we use a range to safely capture it.
  # https://github.com/docker-mailserver/docker-mailserver/pull/3747#discussion_r1446679671
  local QUEUE_ID_REGEX='[A-Z0-9]{9,12}'

  local MSG_ID=$(__construct_msgid "${1:?The local-part for MSG_ID was not provided}")
  shift 1
```

test/helper/log_and_filtering.bash
```diff
@@ -0,0 +1,109 @@
...
+
+  _wait_for_empty_mail_queue_in_container
+
+  QUEUE_ID=$(_exec_in_container tac /var/log/mail.log                    \
+    | grep -E "postfix/cleanup.*: ${QUEUE_ID_REGEX}:.*message-id=${MID}" \
+    | grep -E --only-matching --max-count 1 "${QUEUE_ID_REGEX}" || :)
```
polarathene, line 90:
```suggestion
  QUEUE_ID=$(_exec_in_container tac /var/log/mail.log                    \
    | grep -E "postfix/cleanup.*: ${QUEUE_ID_REGEX}:.*message-id=${MSG_ID}" \
    | grep -E --only-matching --max-count 1 "${QUEUE_ID_REGEX}" || :)
```

test/helper/log_and_filtering.bash
```diff
@@ -0,0 +1,109 @@
...
+# Filters the mail log according to MID (Message-ID) and prints lines
+# of the mail log that fit lines with the pattern `msgid=${1}@dms-test`.
+#
+# @param ${1} = message ID part before '@'
+function _print_mail_log_for_msgid() {
+  local MID=$(__construct_mid "${1:?Left-hand side of MID missing}")
+  shift 1
+
+  _filter_service_log 'mail' "msgid=${MID}"
+}
```
polarathene, line 109: Postfix uses `message-id`, you may prefer a regex, or to filter specifically for any logs with the `MSG_ID` value as a match, otherwise the current usage is Dovecot specific.

```suggestion
# A convenience method that filters for Dovecot specific logs with a `msgid` field that matches the MSG_ID input.
#
# @param ${1} = The local-part of a Message-ID header value (`<local-part@domain-part>`)
function _print_mail_log_for_msgid() {
  local MSG_ID=$(__construct_msgid "${1:?The local-part for MSG_ID was not provided}")
  shift 1

  _filter_service_log 'mail' "msgid=${MSG_ID}"
}
```

test/helper/sending.bash
```diff
@@ -100,74 +102,29 @@ function _send_email() {
   return "${RETURN_VALUE}"
 }
 
-# Like `_send_email` with two major differences:
-#
-# 1. this function waits for the mail to be processed; there is no asynchronicity
-#    because filtering the logs in a synchronous way is easier and safer;
-# 2. this function takes the name of a variable and inserts IDs one can later
-#    filter by to check logs.
+# Contruct the message ID for the 'Message-ID' header.
```
polarathene, line 105:
```suggestion
# Construct the value for a 'Message-ID' header.
# For tests we use only the local-part to identify mail activity in logs. The rest of the value is fixed.
#
# A Message-ID header value should be in the form of: `<local-part@domain-part>`
# https://en.wikipedia.org/wiki/Message-ID
# https://datatracker.ietf.org/doc/html/rfc5322#section-3.6.4
```

test/helper/sending.bash
```diff
@@ -100,74 +102,29 @@ function _send_email() {
...
-# ## Attention
-#
-# This function assumes `CONTAINER_NAME` to be properly set (to the container
-# name the command should be executed in)!
-#
-# ## Safety
+# @param ${1} = message ID part before '@' (later used when filtering logs again)
+function __construct_mid() {
+  echo "<${1:?Message-ID prefix is required}@dms-tests>"
+}
```
polarathene, line 110: Grep friendly with `MSG_ID` / `msgid`.

Defined a local variable for better clarity (_error message is noise to constructed value_), feel free to revert that since this has been better clarified in the parameter comment right above it:

```suggestion
# @param ${1} = The local-part of a Message-ID header value (`<local-part@domain-part>`)
function __construct_msgid() {
  local MSG_ID_LOCALPART=${1:?The local-part for MSG_ID was not provided}
  echo "<${MSG_ID_LOCALPART}@dms-tests>"
}
```

test/helper/sending.bash
```diff
@@ -100,74 +102,29 @@ function _send_email() {
...
+
+# Like `_send_email` but adds a "Message-Id: ${1}@dms-tests>" header, which
+# allows for filtering logs later.
```
polarathene, line 113:
```suggestion
# Like `_send_email` but adds a "Message-ID: ${1}@dms-tests>" header,
# which allows for filtering logs later.
```

test/helper/sending.bash
```diff
@@ -100,74 +102,29 @@ function _send_email() {
...
-  # Export the variable denoted by ${1} so everyone has access
-  export "${1:?Mail ID must be set for _send_email_and_get_id}"
-  # Get a "reference" to the content of the variable denoted by ${1} so we can manipulate the content
-  local -n ID_ENV_VAR_REF=${1:?}
-  # Prepare the message ID header here because we will shift away ${1} later
-  local MID="<${1}@dms-tests>"
-  # Get rid of ${1} so only the arguments for swaks remain
+# @param ${1} = message ID part before '@' (later used when filtering logs again)
+function _send_email_with_mid() {
+  local MID=$(__construct_mid "${1:?Left-hand side of MID missing}")
```
polarathene, line 117:
```suggestion
# @param ${1} = The local-part of a Message-ID header value (`<local-part@domain-part>`)
function _send_email_with_msgid() {
  local MSG_ID=$(__construct_msgid "${1:?The local-part for MSG_ID was not provided}")
```

test/helper/sending.bash
```diff
@@ -100,74 +102,29 @@ function _send_email() {
...
-  # Last but not least, we perform plausibility checks on the IDs.
-  assert_not_equal "${ID_ENV_VAR_REF}" ''
-  run echo "${ID_ENV_VAR_REF}"
-  assert_line --regexp "^${QUEUE_ID_REGEX}\|${MID}$"
 }
 
 # Send a spam e-mail by utilizing GTUBE.
 #
-# Extra arguments given to this function will be supplied by `_send_email_and_get_id` directly.
+# Extra arguments given to this function will be supplied by `_send_email_with_mid` directly.
```
polarathene, line 125:
```suggestion
# Extra arguments given to this function will be supplied by `_send_email_with_msgid` directly.
```

test/helper/sending.bash
```diff
@@ -100,74 +102,29 @@ function _send_email() {
...
 function _send_spam() {
-  _send_email_and_get_id MAIL_ID_SPAM "${@}" \
+  _send_email_with_mid 'spam' "${@}" \
```
polarathene, line 127:
```suggestion
  _send_email_with_msgid 'spam' "${@}" \
```

test/tests/parallel/set1/spam_virus/postscreen.bats
```diff
@@ -56,12 +56,11 @@ function teardown_file() {
 @test "should successfully pass postscreen and get postfix greeting message (respecting postscreen_greet_wait time)" {
   # Configure `send_email()` to send from the mail client container (CONTAINER2_NAME) via ENV override,
   # mail is sent to the DMS server container (CONTAINER1_NAME) via `--server` parameter:
-  # TODO: Use _send_email_and_get_id when proper resolution of domain names is possible:
   CONTAINER_NAME=${CONTAINER2_NAME} _send_email --expect-rejection --server "${CONTAINER1_IP}" --port 25 --data 'postscreen.txt'
-  # CONTAINER_NAME=${CONTAINER2_NAME} _send_email_and_get_id MAIL_ID_POSTSCREEN --server "${CONTAINER1_IP}" --data 'postscreen.txt'
-  # _print_mail_log_for_id "${MAIL_ID_POSTSCREEN}"
+  # TODO: Use _send_email_with_mid when proper resolution of domain names is possible:
+  # CONTAINER_NAME=${CONTAINER2_NAME} _send_email_with_mid 'postscreen' --server "${CONTAINER1_IP}" --data 'postscreen.txt'
+  # _print_mail_log_for_msgid 'postscreen'
```
polarathene, line 62:
```suggestion
  # TODO: Use _send_email_with_msgid when proper resolution of domain names is possible:
  # CONTAINER_NAME=${CONTAINER2_NAME} _send_email_with_msgid 'postscreen' --server "${CONTAINER1_IP}" --data 'postscreen.txt'
  # _print_mail_log_for_msgid 'postscreen'
```

test/tests/parallel/set1/spam_virus/rspamd_full.bats
```diff
@@ -45,16 +45,16 @@ function setup_file() {
 
   # We will send 4 emails:
   #   1. The first one should pass just fine
-  _send_email_and_get_id MAIL_ID_PASS
+  _send_email_with_mid 'pass'
```
polarathene, line 48:
```suggestion
  _send_email_with_msgid 'pass'
```

test/tests/parallel/set1/spam_virus/rspamd_full.bats
```diff
@@ -45,16 +45,16 @@ function setup_file() {
...
   #   2. The second one should be rejected (Rspamd-specific GTUBE pattern for rejection)
   _send_spam --expect-rejection
   #   3. The third one should be rejected due to a virus (ClamAV EICAR pattern)
   # shellcheck disable=SC2016
-  _send_email_and_get_id MAIL_ID_VIRUS --expect-rejection \
+  _send_email_with_mid 'virus' --expect-rejection \
```
polarathene, line 53:
```suggestion
  _send_email_with_msgid 'virus' --expect-rejection \
```

test/tests/parallel/set1/spam_virus/rspamd_full.bats
```diff
@@ -45,16 +45,16 @@ function setup_file() {
...
     --body 'X5O!P%@AP[4\PZX54(P^)7CC)7}$EICAR-STANDARD-ANTIVIRUS-TEST-FILE!$H+H*'
   #   4. The fourth one will receive an added header (Rspamd-specific GTUBE pattern for adding a spam header)
   #      ref: https://rspamd.com/doc/gtube_patterns.html
-  _send_email_and_get_id MAIL_ID_HEADER --body "YJS*C4JDBQADN1.NSBN3*2IDNEN*GTUBE-STANDARD-ANTI-UBE-TEST-EMAIL*C.34X"
+  _send_email_with_mid 'header' --body "YJS*C4JDBQADN1.NSBN3*2IDNEN*GTUBE-STANDARD-ANTI-UBE-TEST-EMAIL*C.34X"
```
polarathene, line 57:
```suggestion
  _send_email_with_msgid 'header' --body "YJS*C4JDBQADN1.NSBN3*2IDNEN*GTUBE-STANDARD-ANTI-UBE-TEST-EMAIL*C.34X"
```

test/tests/parallel/set1/spam_virus/rspamd_full.bats
```diff
@@ -120,9 +120,13 @@ function teardown_file() { _default_teardown ; }
   _service_log_should_contain_string 'rspamd' 'S \(reject\)'
   _service_log_should_contain_string 'rspamd' 'reject "Gtube pattern"'
 
-  _print_mail_log_for_id "${MAIL_ID_SPAM}"
+  _print_mail_log_of_queue_id_from_mid 'spam'
```
polarathene, line 123:
```suggestion
  _print_mail_log_of_queue_id_from_msgid 'spam'
```

test/tests/parallel/set1/spam_virus/rspamd_full.bats
```diff
@@ -131,11 +135,14 @@ function teardown_file() { _default_teardown ; }
   _service_log_should_contain_string 'rspamd' 'T \(reject\)'
   _service_log_should_contain_string 'rspamd' 'reject "ClamAV FOUND VIRUS "Eicar-Signature"'
 
-  _print_mail_log_for_id "${MAIL_ID_VIRUS}"
+  _print_mail_log_of_queue_id_from_mid 'virus'
```
polarathene, line 138:
```suggestion
  _print_mail_log_of_queue_id_from_msgid 'virus'
```

test/helper/log_and_filtering.bash
```diff
@@ -0,0 +1,109 @@
...
+
+# ? ABOUT: Functions defined here aid in working with logs and filtering them.
+
+# ! ATTENTION: This file is loaded by `common.sh` - do not load it yourself!
+# ! ATTENTION: This file requires helper functions from `common.sh`!
+
+# shellcheck disable=SC2034,SC2155
+
+# Assert that the number of lines output by a previous command matches the given
+# amount (${1}). `lines` is a special BATS variable updated via `run`.
```
polarathene, line 11:
```suggestion
# Assert that the number of lines output by a previous command matches the given amount (${1}).
# `lines` is a special BATS variable updated via `run`.
```

test/helper/log_and_filtering.bash
```diff
@@ -0,0 +1,109 @@
...
+#
+# @param ${1} = number of lines that the output should have
+function _should_output_number_of_lines() {
+  # shellcheck disable=SC2154
+  assert_equal "${#lines[@]}" "${1:?Number of lines not provided}"
+}
+
+# Filters a service's logs (under `/var/log/supervisor/<SERVICE>.log`) given
+# a specific string.
```
polarathene, line 20:
```suggestion
# Filters a service's logs (under `/var/log/supervisor/<SERVICE>.log`) given a specific string.
```

polarathene (changes requested):
I'll apply these.

Bulk of the changes are:
- `_mid` / `MID` => `_msgid` / `MSG_ID`
- Revised documentation / tooltip comments

test/tests/parallel/set3/mta/smtp_delivery.bats
```diff
@@ -227,23 +222,20 @@ function _successful() {
...
+  _service_log_should_contain_string 'mail' '<nouser@localhost.localdomain>: Recipient address rejected: User unknown in virtual mailbox table'
   _should_output_number_of_lines 1
 }
 
 @test "redirects mail to external aliases" {
-  _run_in_container_bash "grep 'Passed CLEAN {RelayedInbound}' /var/log/mail/mail.log | grep -- '-> <external1@otherdomain.tld>'"
-  assert_success
-  assert_output --partial '<user@external.tld> -> <external1@otherdomain.tld>'
+  _service_log_should_contain_string 'mail' 'Passed CLEAN {RelayedInbound}'
+  run bash -c "grep '<user@external.tld> -> <external1@otherdomain.tld>' <<< '${output}'"
```
polarathene, line 231: This is not viable like similar adjustments you made?

```suggestion
  assert_output --partial '<user@external.tld> -> <external1@otherdomain.tld>'
```

test/tests/serial/tests.bats
```diff
@@ -182,23 +182,15 @@ function teardown_file() { _default_teardown ; }
...
-  assert_failure
+@test "system: Mail log is error free" {
+  _service_log_should_not_contain_string 'mail' 'non-null host address bits in'
+  _service_log_should_not_contain_string 'mail' 'mail system configuration error'
+  _service_log_should_not_contain_string 'mail' ': error:'
+  _service_log_should_not_contain_string 'mail' 'is not writable'
+  _service_log_should_not_contain_string 'mail' 'permission denied'
+  _service_log_should_not_contain_string 'mail' '\(!\)connect'
+  _service_log_should_not_contain_string 'mail' 'using backwards-compatible default setting'
+  _service_log_should_not_contain_string 'mail' 'connect to 127.0.0.1:10023: Connection refused'
```
polarathene, line 193: Could probably retrieve the log and use `refute_output --partial` instead?

Then again some of the statements were relying on `grep -i`, so they potentially match multiple or casing may not be accurate 🤷‍♂️ Bit harder with failure assertions to determine 😓 (_although in this case I think I added context in the vmail-id test PR_)

**EDIT:** Feedback given on that related snippet. A future PR would replace this in favour of the `vmail-id` one. Tracking for that is in https://github.com/docker-mailserver/docker-mailserver/issues/3554

test/tests/serial/vmail-id.bats
```diff
@@ -41,35 +40,27 @@ function teardown_file() { _default_teardown ; }
...
 
   # Amavis: https://forum.howtoforge.com/threads/postfix-smtp-error-caused-by-clamav-cant-connect-to-a-unix-socket-var-run-clamav-clamd-ctl.81002/
-  _run_in_container grep -i '(!)connect' /var/log/mail/mail.log
-  assert_failure
+  _service_log_should_not_contain_string 'mail' '\(!\)connect'
 
   # Postfix: https://github.com/docker-mailserver/docker-mailserver/pull/2597
-  _run_in_container grep -i 'using backwards-compatible default setting' /var/log/mail/mail.log
-  assert_failure
+  _service_log_should_not_contain_string 'mail' 'using backwards-compatible default setting'
```
polarathene, line 61: The referenced PR has similar looking log lines, but it seems that the check here was referring to [log lines shown in this comment](https://github.com/docker-mailserver/docker-mailserver/pull/2598#issuecomment-1141176633):

```suggestion
   # Log line match example: https://github.com/docker-mailserver/docker-mailserver/pull/2598#issuecomment-1141176633
  _service_log_should_not_contain_string 'mail' 'using backwards-compatible default setting'
```

test/tests/serial/vmail-id.bats
```diff
@@ -41,35 +40,27 @@ function teardown_file() { _default_teardown ; }
...
-  assert_failure
-  _run_in_container grep -i 'permission denied' /var/log/mail/mail.log
-  assert_failure
+  _service_log_should_not_contain_string 'mail' 'not writable'
+  _service_log_should_not_contain_string 'mail' 'permission denied'
 
   # Amavis: https://forum.howtoforge.com/threads/postfix-smtp-error-caused-by-clamav-cant-connect-to-a-unix-socket-var-run-clamav-clamd-ctl.81002/
-  _run_in_container grep -i '(!)connect' /var/log/mail/mail.log
-  assert_failure
+  _service_log_should_not_contain_string 'mail' '\(!\)connect'
```
polarathene, line 58: This one is correct and doesn't seem like it needs any different casing (_as most probably don't with `-i`_).

It'd be good if this was not catering to regex 😅

test/tests/serial/vmail-id.bats
```diff
@@ -41,35 +40,27 @@ function teardown_file() { _default_teardown ; }
 # This test case is shared with tests.bats, but provides context on errors + some minor edits
 # TODO: Could improve in future with keywords from https://github.com/docker-mailserver/docker-mailserver/pull/3550#issuecomment-1738509088
 # Potentially via a helper that allows an optional fixed number of errors to be present if they were intentional
-@test '/var/log/mail/mail.log is error free' {
+@test 'Mail log is error free' {
```
polarathene, line 43: As [the PR comment above the test-case definition notes](https://github.com/docker-mailserver/docker-mailserver/pull/3550#issuecomment-1738509088), there's some extras another PR could check for.

- Below the `: error:` is likely the equivalent of the Dovecot error, where the `-i` may be relaxing the casing from `: Error:`.
- A [follow-up review comment on that link](https://github.com/docker-mailserver/docker-mailserver/pull/3550#discussion_r1339424344) also notes the same for `permission denied` as `failed: Permission denied`.

test/tests/serial/mail_pop3.bats
```diff
@@ -33,11 +33,9 @@ function teardown_file() { _default_teardown ; }
...
 }
 
-@test '/var/log/mail/mail.log is error-free' {
-  _run_in_container grep 'non-null host address bits in' /var/log/mail/mail.log
-  assert_failure
-  _run_in_container grep ': error:' /var/log/mail/mail.log
-  assert_failure
+@test 'Mail log is error-free' {
+  _service_log_should_not_contain_string 'mail' 'non-null host address bits in'
+  _service_log_should_not_contain_string 'mail' ': error:'
```
polarathene, line 38:
```suggestion
# TODO: Remove in favor of a common helper method, as described in vmail-id.bats equivalent test-case
@test 'Mail log is error free' {
  _service_log_should_not_contain_string 'mail' 'non-null host address bits in'
  _service_log_should_not_contain_string 'mail' ': error:'
```

test/tests/parallel/set1/spam_virus/rspamd_full.bats
```diff
@@ -279,9 +286,9 @@ function teardown_file() { _default_teardown ; }
   _nc_wrapper 'nc/rspamd_imap_move_to_inbox.txt' '0.0.0.0 143'
   sleep 1 # wait for the transaction to finish
 
-  _run_in_container cat /var/log/mail/mail.log
-  assert_success
-  assert_output --partial 'imapsieve: Matched static mailbox rule [2]'
+  _service_log_should_contain_string 'mail' 'imapsieve: Matched static mailbox rule \[2\]'
+
+  _filter_service_log 'mail' '.'
```
polarathene, line 291: This usage relying on regex, along with some other usage suggests you could have a method that just returns the log, or defaults the 2nd parameter to `.` if it's not provided? 🤔

test/tests/parallel/set1/spam_virus/rspamd_full.bats
```diff
@@ -45,16 +45,16 @@ function setup_file() {
 
   # We will send 4 emails:
   #   1. The first one should pass just fine
-  _send_email_and_get_id MAIL_ID_PASS
+  _send_email_with_msgid 'pass'
```
polarathene, line 48: All the `MSG_ID` header values provided in this test file could probably use a more grep friendly value, especially since `virus` / `header` / `pass` / `spam` are all very generic 😬

test/helper/log_and_filtering.bash
```diff
@@ -0,0 +1,108 @@
...
+  QUEUE_ID=$(_exec_in_container tac /var/log/mail.log                    \
+    | grep -E "postfix/cleanup.*: ${QUEUE_ID_REGEX}:.*message-id=${MSG_ID}" \
+    | grep -E --only-matching --max-count 1 "${QUEUE_ID_REGEX}" || :)
+
+  # We perform plausibility checks on the IDs.
+  assert_not_equal "${QUEUE_ID}" ''
+  run echo "${QUEUE_ID}"
+  assert_line --regexp "^${QUEUE_ID_REGEX}$"
+
+  _filter_service_log 'mail' "${QUEUE_ID}"
```
polarathene, line 97:
```suggestion
  # Postfix specific logs:
  _filter_service_log 'mail' "${QUEUE_ID}"
```

test/helper/log_and_filtering.bash
```diff
@@ -0,0 +1,108 @@
...
+}
+
+# A convenience method that filters for Dovecot specific logs with a `msgid` field that matches the MSG_ID input.
+#
+# @param ${1} = The local-part of a Message-ID header value (`<local-part@domain-part>`)
+function _print_mail_log_for_msgid() {
+  local MSG_ID=$(__construct_msgid "${1:?The local-part for MSG_ID was not provided}")
+  shift 1
+
+  _filter_service_log 'mail' "msgid=${MSG_ID}"
```
polarathene, line 107:
```suggestion
  # Dovecot specific logs:
  _filter_service_log 'mail' "msgid=${MSG_ID}"
```

polarathene (changes requested):
I'd personally prefer being more explicit about those that are using regex, potentially with a `_regex` variant of such methods 🤷‍♂️ 

Probably unlikely to be an issue but technically all those logs being checked for addresses will treat the `.` as any character rather than an actual `.`? Fixed string matching by default seems better suited?

Feel free to delay that to a separate PR.

---

Great work! 🥳

test/helper/sending.bash
```diff
@@ -100,74 +102,35 @@ function _send_email() {
...
+function _send_email_with_msgid() {
+  local MSG_ID=$(__construct_msgid "${1:?The local-part for MSG_ID was not provided}")
   shift 1
 
-  # The unique ID Postfix (and other services) use may be different in length
-  # on different systems. Hence, we use a range to safely capture it.
-  local QUEUE_ID_REGEX='[A-Z0-9]{9,12}'
-
-  _wait_for_empty_mail_queue_in_container
   _send_email "${@}" --header "Message-Id: ${MID}"
```
polarathene, line 126: I missed adjusting this one as the `MID` already existed prior to the PR 😓 

```suggestion
  _send_email "${@}" --header "Message-Id: ${MSG_ID}"
```

test/tests/parallel/set3/mta/smtp_delivery.bats
```diff
@@ -228,7 +228,7 @@ function _successful() {
 
 @test "redirects mail to external aliases" {
   _service_log_should_contain_string 'mail' 'Passed CLEAN {RelayedInbound}'
-  run bash -c "grep '<user@external.tld> -> <external1@otherdomain.tld>' <<< '${output}'"
+  assert_output --partial '<user@external.tld> -> <external1@otherdomain.tld>'
   _should_output_number_of_lines 2
```
georglauterbach, line 232: @polarathene this changes causes

```txt
not ok 49 [SMTP] (delivery) redirects mail to external aliases in 131ms
# (from function `assert_equal' in file test/test_helper/bats-assert/src/assert_equal.bash, line 40,
#  from function `_should_output_number_of_lines' in file test/helper/log_and_filtering.bash, line 16,
#  in test file test/tests/parallel/set3/mta/smtp_delivery.bats, line 232)
#   `_should_output_number_of_lines 2' failed
#
# -- values do not equal --
# expected : 2
# actual   : 13
# --
#
```

This is simply because `_should_output_number_of_lines 2` works on `_service_log_should_contain_string 'mail' 'Passed CLEAN {RelayedInbound}'` in your example and not on `run bash -c "grep '<user@external.tld> -> <external1@otherdomain.tld>' <<< '${output}'"` (which is the case in my example).

This may be confusing at first sight.

test/tests/serial/tests.bats
georglauterbach: It's probably a good idea to solve this in a PR that focuses on #3554

test/tests/serial/vmail-id.bats
georglauterbach: > This one is correct and doesn't seem like it needs any different casing (_as most probably don't with `-i`_).

Since I added `-i` to the `grep` statement in `_filter_service_log`, this PR introduces the maybe unwanted situation in which the casing does not matter at all. I'll play a bit around with that.

> It'd be good if this was not catering to regex 😅

Yes, but it was easier to implement it first for a regex. I will also play around with that.

test/tests/serial/vmail-id.bats
georglauterbach: Appropriate for a follow-up PR IMO :D

test/tests/parallel/set1/spam_virus/rspamd_full.bats
georglauterbach: > This usage relying on regex, along with some other usage suggests you could have a method that just returns the log

I explicitly opted against such a method because

1. we do not need it in many places really, filtering is always nicer;
2. it is probably a bit slower than filtering (with `grep`) (I haven't verified this, though);
3. I want future test writers to use filters instead of just `cat`ing the whole log, which may seem easy, but it is also error-prone.

> or defaults the 2nd parameter to `.` if it's not provided? 🤔

This is, IMO, a confusing default and hence, one I don't like. Especially with upcoming changes of not using a regex by default but a static string, we need to carefully think what to do here.

test/tests/parallel/set1/spam_virus/rspamd_full.bats
georglauterbach: 👍🏼

test/tests/parallel/set3/mta/smtp_delivery.bats
georglauterbach: I will revert this change.

test/tests/parallel/set1/spam_virus/rspamd_full.bats
georglauterbach: Done in 143ca0c27d686c44882b7e045e1c98574a20995d

test/tests/parallel/set3/mta/smtp_delivery.bats
georglauterbach: Done in 5622761c41abbf3f39bc9522d6cdfb6902ee0b88

test/tests/parallel/set1/spam_virus/rspamd_full.bats
georglauterbach: Done in aaeea8af8100fa52e5087abe87b007726e7826de

test/tests/serial/vmail-id.bats
georglauterbach: Partially implemented in aaeea8af8100fa52e5087abe87b007726e7826de
Converted all usage of helpers in b3a6a7b103862d601f7b9bd0b1cc33d4a632e337

test/tests/parallel/set3/mta/smtp_delivery.bats
polarathene: Related review comment: https://github.com/docker-mailserver/docker-mailserver/pull/3786#discussion_r1460360285

---

Original:

```bash
# `_should_output_number_of_lines 2` was compatible with these two lines:
_run_in_container_bash "grep 'Passed CLEAN {RelayedInbound}' /var/log/mail/mail.log | grep -- '-> <external1@otherdomain.tld>'"
assert_output --partial '<user@external.tld> -> <external1@otherdomain.tld>'
```

Replacement:

```bash
# assert_output passes but would cause `_should_output_number_of_lines 2`  to fail (13 lines vs 2),
# hence workaround with run bash grep instead of assert_output
_service_log_should_contain_string 'mail' 'Passed CLEAN {RelayedInbound}'
run bash -c "grep '<user@external.tld> -> <external1@otherdomain.tld>' <<< '${output}'"
```

I still don't follow why there is now 13 lines returned from the log method instead of the expected two..

**EDIT:** Ah.. Originally there was a 2nd grep to reduce the output from 13 to 2. The 2nd grep was to filter on the target value `external1@otherdomain.tld`, and use the `assert_output` for the additional context, so that if this test fails we'd get context on the mismatch for that left-side value of the assert - which would not be present by just relying on grep alone.

polarathene:
```suggestion
  # Filter to the specific subset of logs retrieved that we're interested in:
  run bash -c "grep -- '-> <external1@otherdomain.tld>' <<< '${output}'"
  # Now assert on the full expected content, if the left-side is a mismatch we'll get the value found via failure context
  assert_output --partial '<user@external.tld> -> <external1@otherdomain.tld>'
```

---

I believe this was done based on the commented out assertion below related to PostSRSd (if it were enabled):

```
assert_output --partial 'external.tld=user@example.test> -> <external1@otherdomain.tld>'
```

Where before it was not obvious why the test failed to match, and without that in the failure context (no results from grep instead), you'd have to do more work to reproduce the log output just to see the failure was due to hat difference in value.

test/tests/parallel/set1/spam_virus/rspamd_full.bats
```diff
@@ -108,34 +109,41 @@ function teardown_file() { _default_teardown ; }
...
 @test 'detects and rejects virus' {
-  _service_log_should_contain_string 'rspamd' 'T \(reject\)'
+  _service_log_should_contain_string 'rspamd' 'T (reject)'
   _service_log_should_contain_string 'rspamd' 'reject "ClamAV FOUND VIRUS "Eicar-Signature"'
 
-  _print_mail_log_for_id "${MAIL_ID_VIRUS}"
+  _print_mail_log_of_queue_id_from_msgid 'rspamd-test-email-virus'
   assert_output --partial 'milter-reject'
   assert_output --partial '5.7.1 ClamAV FOUND VIRUS "Eicar-Signature"'
   refute_output --partial "stored mail into mailbox 'INBOX'"
```
polarathene, line 142: Since this is a Dovecot line, this assertion will never fail. It serves no value since the content returned would only ever be related to the Queue ID, and the method has since changed to provide a specific MSG_ID, rather than rely on Postfix to generate one that contains the Queue ID.

Instead this is now covered by your `_print_mail_log_for_msgid` addition below asserting failure.

test/tests/parallel/set1/spam_virus/rspamd_full.bats
```diff
@@ -108,34 +109,41 @@ function teardown_file() { _default_teardown ; }
...
 
+  _print_mail_log_for_msgid 'dms-test-email-spam'
+  assert_failure
```
polarathene, line 145: While not exactly necessary, if the `assert_failure` did... fail, it would be good to catch this expected refute for context rather than a less helpful status failure. It also adds context on the log/action we're expecting to fail and test against.

```suggestion
  _print_mail_log_for_msgid 'dms-test-email-spam'
  refute_output --partial "stored mail into mailbox 'INBOX'"
  assert_failure
```

test/tests/parallel/set1/spam_virus/rspamd_full.bats
polarathene:
```suggestion
```

test/tests/parallel/set1/spam_virus/rspamd_full.bats
```diff
@@ -108,34 +109,41 @@ function teardown_file() { _default_teardown ; }
...
   _service_log_should_contain_string 'rspamd' 'reject "Gtube pattern"'
 
-  _print_mail_log_for_id "${MAIL_ID_SPAM}"
+  _print_mail_log_of_queue_id_from_msgid 'dms-test-email-spam'
   assert_output --partial 'milter-reject'
   assert_output --partial '5.7.1 Gtube pattern'
+  refute_output --partial "stored mail into mailbox 'INBOX'"
+
+  _print_mail_log_for_msgid 'dms-test-email-spam'
+  assert_failure
```
polarathene, line 130: Same here:

```suggestion

  _print_mail_log_for_msgid 'dms-test-email-spam'
  refute_output --partial "stored mail into mailbox 'INBOX'"
  assert_failure
```

test/tests/parallel/set1/spam_virus/postscreen.bats
```diff
@@ -56,12 +56,11 @@ function teardown_file() {
 @test "should successfully pass postscreen and get postfix greeting message (respecting postscreen_greet_wait time)" {
   # Configure `send_email()` to send from the mail client container (CONTAINER2_NAME) via ENV override,
   # mail is sent to the DMS server container (CONTAINER1_NAME) via `--server` parameter:
-  # TODO: Use _send_email_and_get_id when proper resolution of domain names is possible:
   CONTAINER_NAME=${CONTAINER2_NAME} _send_email --expect-rejection --server "${CONTAINER1_IP}" --port 25 --data 'postscreen.txt'
-  # CONTAINER_NAME=${CONTAINER2_NAME} _send_email_and_get_id MAIL_ID_POSTSCREEN --server "${CONTAINER1_IP}" --data 'postscreen.txt'
-  # _print_mail_log_for_id "${MAIL_ID_POSTSCREEN}"
+  # TODO: Use _send_email_with_msgid when proper resolution of domain names is possible:
+  # CONTAINER_NAME=${CONTAINER2_NAME} _send_email_with_msgid 'postscreen' --server "${CONTAINER1_IP}" --data 'postscreen.txt'
+  # _print_mail_log_for_msgid 'postscreen'
```
polarathene, line 62:
```suggestion
  # TODO: Use _send_email_with_msgid when proper resolution of domain names is possible:
  # CONTAINER_NAME=${CONTAINER2_NAME} _send_email_with_msgid 'msgid-postscreen' --server "${CONTAINER1_IP}" --data 'postscreen.txt'
  # _print_mail_log_for_msgid 'msgid-postscreen'
```

test/helper/sending.bash
```diff
@@ -100,74 +102,35 @@ function _send_email() {
...
-    | grep -E --only-matching --max-count 1 "${QUEUE_ID_REGEX}" || :)
-  # But we also requre potential Dovecot sieve output, which requires the mesage ID,
-  # so we need to provide the message ID too.
-  ID_ENV_VAR_REF+="|${MID}"
-
-  # Last but not least, we perform plausibility checks on the IDs.
-  assert_not_equal "${ID_ENV_VAR_REF}" ''
-  run echo "${ID_ENV_VAR_REF}"
-  assert_line --regexp "^${QUEUE_ID_REGEX}\|${MID}$"
+  _send_email "${@}" --header "Message-Id: ${MSG_ID}"
```
polarathene, line 126:
```suggestion
  _send_email "${@}" --header "Message-ID: ${MSG_ID}"
```

test/helper/log_and_filtering.bash
```diff
@@ -0,0 +1,112 @@
...
+#
+# @param ${1} = service name
+# @param ${2} = string to filter by
+# @param ${3} = container name [OPTIONAL]
+#
+# ## Attention
+#
+# The string given to this function is interpreted by `grep -E`, i.e.
+# as a regular expression. In case you use characters that are special
+# in regular expressions, you need to escape them!
```
polarathene, line 53: No longer the case:

```suggestion
```

test/helper/log_and_filtering.bash
```diff
@@ -0,0 +1,112 @@
...
```
polarathene, line 69: No longer the case:

```suggestion
```

polarathene (changes requested):
I'll apply some of these, and leave some others unresolved for better visibility to you 👍

test/tests/parallel/set1/spam_virus/rspamd_full.bats
```diff
@@ -108,34 +109,41 @@ function teardown_file() { _default_teardown ; }
...
   _count_files_in_directory_in_container /var/mail/localhost.localdomain/user1/new/ 1
 }
 
 @test 'detects and rejects virus' {
-  _service_log_should_contain_string 'rspamd' 'T \(reject\)'
+  _service_log_should_contain_string 'rspamd' 'T (reject)'
   _service_log_should_contain_string 'rspamd' 'reject "ClamAV FOUND VIRUS "Eicar-Signature"'
 
-  _print_mail_log_for_id "${MAIL_ID_VIRUS}"
+  _print_mail_log_of_queue_id_from_msgid 'rspamd-test-email-virus'
```
polarathene, line 139: Similar to the `dms-test-email-spam`, it might be nice to have the equivalent for virus. Not require to be addressed in this PR though since there's plenty of noise here already 😅

test/helper/log_and_filtering.bash
```diff
@@ -0,0 +1,100 @@
...
+# @param ${1} = service name
+# @param ${2} = string to filter by
+# @param ...  = options given to `grep` (which is used to filter logs)
+function _filter_service_log() {
+  local SERVICE=${1:?Service name must be provided}
+  local STRING=${2:?String to match must be provided}
+  shift 2
+
+  local FILE="/var/log/supervisor/${SERVICE}.log"
+  [[ -f ${FILE} ]] || FILE="/var/log/mail/${SERVICE}.log" # Fallback to alternative log location:
```
polarathene, line 30:
```suggestion
  # Alternative log location fallback:
  [[ -f ${FILE} ]] || FILE="/var/log/mail/${SERVICE}.log"
```

test/helper/log_and_filtering.bash
```diff
@@ -0,0 +1,100 @@
...
+  _run_in_container grep "${@}" "${STRING}" "${FILE}"
```
polarathene, line 31: `${@}` should default/fallback to `--fixed-strings` if possible, or use a separate `_regex` variant method.

The benefit of a `_regex` variant method is that it's easier to grep the test files for usage of where we're actually using that optional parameter, but that's a minor concern, I'm happy regardless of how you want to approach it.

polarathene (changes requested):
Just the prior review feedback and some additional feedback here to address and this looks good to go! 😁 🚀

polarathene:
Huzzah, a failure like I expected 😛 

```
not ok 86 [Rspamd] (full) RSPAMD_LEARN works in 3193ms
#  from function `_service_log_should_contain_string' in file test/helper/log_and_filtering.bash, line 50,
#  in test file test/tests/parallel/set1/spam_virus/rspamd_full.bats, line 290)
#   `_service_log_should_contain_string 'mail' 'imapsieve: Matched static mailbox rule [2]'' failed
#
# -- command failed --
# status : 1
# output :
# --
#
```

test/tests/parallel/set3/mta/smtp_delivery.bats
polarathene: Resolved with https://github.com/docker-mailserver/docker-mailserver/pull/3786#discussion_r1460693476

test/tests/parallel/set1/spam_virus/rspamd_full.bats
```diff
@@ -265,10 +273,10 @@ function teardown_file() { _default_teardown ; }
   _nc_wrapper 'nc/rspamd_imap_move_to_junk.txt' '0.0.0.0 143'
   sleep 1 # wait for the transaction to finish
 
-  _run_in_container cat /var/log/mail/mail.log
-  assert_success
-  assert_output --partial 'imapsieve: Matched static mailbox rule [1]'
-  refute_output --partial 'imapsieve: Matched static mailbox rule [2]'
+  _service_log_should_contain_string 'mail' 'imapsieve: Matched static mailbox rule \[1\]'
+  _service_log_should_not_contain_string 'mail' 'imapsieve: Matched static mailbox rule \[2\]'
```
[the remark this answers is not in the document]
georglauterbach: Good catch! I will adjust the PR.

test/helper/log_and_filtering.bash
georglauterbach: Having `_regex` will not be DRY with our helpers at all. But I guess the concern about searching for the method is a good one. I will implement that.

test/helper/log_and_filtering.bash
```diff
@@ -0,0 +1,119 @@
...
+#
+# @param ${1} = service name
+# @param ${2} = string to filter by
+function _service_log_should_not_contain_string() {
+  _filter_service_log "${1}" "${2}" --fixed-strings
+  assert_failure
+}
+
+# Like `_filter_service_log` but asserts that the string was found. Uses regular expressions
+# under the hood for pattern matching.
```
polarathene, line 63:
```suggestion
# Like `_filter_service_log` but asserts that the string was found.
# Uses regular expressions under the hood for pattern matching.
```

test/helper/log_and_filtering.bash
```diff
@@ -0,0 +1,119 @@
...
+#
+# @param ${1} = service name
+# @param ${2} = regular expression to filter by
+function _service_log_should_contain_string_regexp() {
+  _filter_service_log "${1}" "${2}" --extended-regexp
+  assert_success
+}
+
+# Like `_filter_service_log` but asserts that the string was not found. Uses regular expressions
+# under the hood for pattern matching.
```
polarathene, line 73:
```suggestion
# Like `_filter_service_log` but asserts that the string was _not_ found.
# Uses regular expressions under the hood for pattern matching.
```

test/helper/log_and_filtering.bash
```diff
@@ -0,0 +1,119 @@
...
+# Like `_filter_service_log` but asserts that the string was found.
+#
+# @param ${1} = service name
+# @param ${2} = string to filter by
+function _service_log_should_contain_string() {
+  _filter_service_log "${1}" "${2}" --fixed-strings
+  assert_success
+}
+
+# Like `_filter_service_log` but asserts that the string was not found.
```
polarathene, line 53:
```suggestion
# Like `_filter_service_log` but asserts that the string was _not_ found.
```

CHANGELOG.md
```diff
@@ -6,9 +6,14 @@ All notable changes to this project will be documented in this file. The format
...
 
+### Updates
+
+- **Tests:**
+  - Refactored helper methods for sending e-mails with specific `Message-ID` headers and the helpers for retrieving + filtering logs, which together help isolate logs relevant to specific mail when multiple mails have been processed within a single test.
+
 ### Fixes
 
-**Dovecot:**
+- **Dovecot:**
```
polarathene, line 16: Already handled in the `v13.3.1` release PR.

```suggestion
**Dovecot:**
```

It'd probably be a bit weird to squeeze this PR into a bug fix release 😅 (_it's marked for v14 too_)

polarathene (changes requested):

polarathene (dismissed):
LGTM 👍 

May want to merge this after the base image upgrade PR?

georglauterbach:
Yes, this is fine with me :)

georglauterbach:
I resolved merge conflicts and enabled auto-merged. When you approve @polarathene, this will automatically be merged into `master`.

polarathene (approved):
🚀

PR #3795, opened 2024-01-19
tests: Revise OAuth2 tests

polarathene:
# Description

This is a follow up to the original feature PR https://github.com/docker-mailserver/docker-mailserver/pull/3480

Each commit provides commit message for extra context. Changes have been staged into iterations towards the full review diff (_so there's a little bit of extra noise there_). Should be easy to follow if the full diff is too much.

## Overview

- Replaced the Python mock service with [Caddy](https://caddyserver.com), it'll be used in other tests such as for `MTA-STS` in future.
- Extended test coverage to both SASL mechanisms the feature enables:
  - `XOAUTH2` (_Googles widely adopted implementation_) and `OAUTHBEARER` (_the newer variant standardized by [RFC 7628](https://datatracker.ietf.org/doc/html/rfc7628) in 2015_).
  - Both are supported as different ways to verify authorization against the identity provider via the same OAuth Bearer Token (_[RFC 6750](https://datatracker.ietf.org/doc/html/rfc6750), 2012_).
- Better asserts in the test-cases, and more thorough / accurate documentation for maintainers (_I could probably improve the Caddy specific information, but I think it's roughly intuitive enough_).
- The Caddy service additionally provides an endpoint that handles generation of the credentials base64 encoded. 
  - Similar to what the Python script did except you can use `curl` against it with the two query string parameters for dynamic values.
  - This may be useful in future, but is simple to use with the complimentary `compose.yaml` for troubleshooting outside of a running test.

If time would permit, I'd extend the tests further for Dovecot configuration related to this feature. Caddy keeps the interactions fairly simple and transparent for communicating expectations to maintainers along with assistance in troubleshooting.

### Extra info

If needing to troubleshoot, adding the following ([global options](https://caddyserver.com/docs/caddyfile/options#log-credentials)) to the `Caddyfile` for more helpful information:

```caddyfile
{
  # Enables logging without omitting sensitive data
  # https://caddyserver.com/docs/caddyfile/directives/log
  # https://caddyserver.com/docs/caddyfile/options#log-credentials
  servers {
    log_credentials
  }

  # Slightly better output for console viewing:
  # https://caddyserver.com/docs/caddyfile/directives/log#format-modules
  log {
    format console {
      # This is a Go time format string, these values are not random but intentional:
      time_format "15:04:05"
    }
  }
}
```

## Type of change

## Checklist:


test/config/oauth2/Caddyfile
```diff
@@ -0,0 +1,87 @@
+# Mocked OAuth2 /userinfo endpoint normally provided via an Authorization Server (AS) / Identity Provider (IdP)
+#
+# Dovecot will query the mocked `/userinfo` endpoint with the OAuth2 bearer token it was provided during login.
+# If the session for the token is valid, a response returns an attribute to perform a UserDB lookup on (default: email).
+
+# `DMS_YWNjZXNzX3Rva2Vu` is the access token our OAuth2 tests expect for an authorization request to be successful.
+# - The token was created by base64 encoding the string `access_token`, followed by adding `DMS_` as a prefix.
+# - Normally an access token is a short-lived value associated to a login session. The value does not encode any real data.
```
polarathene, line 8:
```suggestion
# - Normally an access token is a short-lived value associated to a login session. The value does not encode any real data.
# It is an opaque token: https://oauth.net/2/bearer-tokens/
```

test/tests/serial/mail_with_oauth2.bats
```diff
@@ -19,11 +19,10 @@ function setup_file() {
...
     --hostname "${FQDN_OAUTH2}" \
     --network "${DMS_TEST_NETWORK}" \
-    --volume "${REPOSITORY_ROOT}/test/config/oauth2/:/app/" \
-    docker.io/library/python:latest \
-    python /app/provider.py
+    --volume "${REPOSITORY_ROOT}/test/config/oauth2/Caddyfile:/etc/caddy/Caddyfile:ro" \
+    caddy:2.7
 
-  _run_until_success_or_timeout 20 sh -c "docker logs ${CONTAINER2_NAME} 2>&1 | grep 'Starting server'"
+  _run_until_success_or_timeout 20 sh -c "docker logs ${CONTAINER2_NAME} 2>&1 | grep 'serving initial configuration'"
```
georglauterbach, line 25:
```suggestion
  _run_until_success_or_timeout 20 bash -c "docker logs ${CONTAINER2_NAME} 2>&1 | grep 'serving initial configuration'"
```

test/tests/serial/mail_with_oauth2.bats
```diff
@@ -61,5 +61,9 @@ function teardown_file() {
   _run_in_container_bash 'nc -vz 0.0.0.0 143'
 
   _nc_wrapper 'auth/imap-oauth2-auth.txt' '-w 1 0.0.0.0 143'
-  assert_output --partial 'Examine completed'
+
+  # Inspect the relevant Dovecot logs to catch failure / success:
+  _run_in_container grep 'dovecot:' /var/log/mail.log
```
georglauterbach, line 75: With #3786 merged, this should become

```suggestion
  _service_log_should_contain_string 'mail' 'dovecot:'

georglauterbach (dismissed):
LGTM 👍🏼

I am not too deep into OAUTH2, but might use it as well (later) when we have something like OIDC going. These changes look very good, and the additional information in the form of comments is appreciated!

casperklein (dismissed):

georglauterbach (approved):

Issue #3791, opened 2024-01-17
[TODO]: Replace `mailserver.env` in favor of improved ENV docs


polarathene:
### Subject

Something else that requires developers attention

### Description

As [discussed here](https://github.com/docker-mailserver/docker-mailserver/pull/3767#discussion_r1456502283). `mailserver.env` and the ENV docs aren't entirely in sync, and it adds a maintenance burden (_both in maintaining it and noise it often contributes to bug reports that copy/paste the whole thing instead of only sharing what was explicitly changed_).

We could convey most value with restructuring of the ENV docs, but the primary improvements for referencing/overview of ENV would be:
- A ENV overview table at the top with key/value columns. The value being the default.
  - A third column could technically list alternative values (enum variants), or a description of the value expected.
  - Key column would list each ENV with a link to the more detailed section on that page for the ENV to get more information.
- If using Content Tabs feature, or a collapsed admonition; we could also include a `.env` / `compose.yaml` of the ENV with empty values for easy copy/paste if that is considered helpful to users.

This would better centralize our ENV (and docs for it). Especially since the docs are better versioned, we've had users try the `master` branch `mailserver.env` instead of from a tagged release that matches their image version (_although still likely to happen with the `latest` docs we direct users to by default_ 🤷‍♂️ ).

Issue #3788, opened 2024-01-17
[TODO]: Improve POP3 docs


polarathene:
### Subject

Something else that requires developers attention

### Description

As detailed in [this review comment](https://github.com/docker-mailserver/docker-mailserver/pull/3767#discussion_r1451811263), enabling POP3 will fail to retrieve any mail stored in junk (_even if incorrectly flagged as spam_).

For those not using Rspamd but enabling SpamAssassin, the `SA_SPAM_SUBJECT` ENV does not assist the POP3 situation due to the default behaviour to move mail to junk folders. This ENV should default to `undef` (_avoids prepending anything to the mail subject header_), optionally using the current default as an alternative fallback instead of `undef` only when related ENV to direct spam to a junk folder is explicitly disabled.

---

Users interested in POP3 should be made aware of this implicit configuration gotcha regardless via an update to the docs.

PR #3820, opened 2024-01-24
spam: use Sieve for rewriting subject with Rspamd & SA/Amavis

### LINKED ISSUE #3323: tracking: equivalent configs of SpamAssassin & Rspamd

### LINKED ISSUE #3804: bug report: rspamd `X-Spam` header not applied to spam message with score 9
### 📝 Preliminary Checks

### 👀 What Happened?

I'm trying to use `MOVE_SPAM_TO_JUNK: 1`

Receiving a mail with a rspamd score of 9.00, the `X-Spam` header is missing, therefore the mail was not moved to the junk folder.

The (default) dms configuration looks fine:
- https://github.com/docker-mailserver/docker-mailserver/blob/master/target/rspamd/local.d/worker-proxy.inc#L18
- https://github.com/docker-mailserver/docker-mailserver/blob/master/target/rspamd/local.d/actions.conf#L8

```
greylist = 4;
add_header = 6;
rewrite_subject = 7;
reject = 11;
```

headers of this message:
```
X-Spamd-Result: default: False [9.00 / 11.00];
	URIBL_BLACK(7.50)[intertecsolar.com:url];
	R_SPF_NA(1.50)[no SPF record];
	MIME_GOOD(-0.10)[multipart/alternative,text/plain];
	BAD_REP_POLICIES(0.10)[];
	MIME_TRACE(0.00)[0:+,1:+,2:~];
	RCVD_COUNT_ZERO(0.00)[0];
	RCPT_COUNT_ONE(0.00)[1];
	MISSING_XM_UA(0.00)[];
	ARC_NA(0.00)[];
	ASN(0.00)[asn:201579, ipnet:91.237.124.0/24, country:GB];
	MID_RHS_MATCH_FROMTLD(0.00)[];
	TO_DN_ALL(0.00)[];
	FROM_HAS_DN(0.00)[];
	R_DKIM_ALLOW(0.00)[extintoresmg.com:s=root];
	FROM_EQ_ENVFROM(0.00)[];
	TO_MATCH_ENVRCPT_ALL(0.00)[];
	DMARC_POLICY_ALLOW(0.00)[extintoresmg.com,reject];
	DKIM_TRACE(0.00)[extintoresmg.com:+]
X-Rspamd-Server: <redacted>
X-Rspamd-Action: rewrite subject
X-Rspamd-Queue-Id: DEF36C38C2
```

I'm still reading the rspamd documentation to figure out why this is happening.

What I'm thinking right now: maybe a higher score action disables the lower actions? In this case the score is 9, so rspamd skips `greylist` and `add_header` and jumps directly to `rewrite_subject`?

**Update** I just received a message with a score of 6.3 and the `X-Spam` header is set. I think my assumption above is correct. `rewrite_subject` action breaks `MOVE_SPAM_TO_JUNK`.

### 👟 Reproduction Steps

_No response_

### 🐋 DMS Version

v13.3.0

Bookworm

```yml
environment:
      ONE_DIR: 1
      ENABLE_FAIL2BAN: 1
      ENABLE_RSPAMD: 1
      ENABLE_DNSBL: 1
      ENABLE_CLAMAV: 1
      ENABLE_OPENDKIM: 0
      ENABLE_OPENDMARC: 0
      ENABLE_POLICYD_SPF: 0
      ENABLE_AMAVIS: 0
      ENABLE_SPAMASSASSIN: 0
      SSL_TYPE: letsencrypt
      SSL_DOMAIN: ${HOSTNAME}
      POSTMASTER_ADDRESS: ${ADMIN_EMAIL}
      MOVE_SPAM_TO_JUNK: 1
      RSPAMD_LEARN: 1
      ENABLE_MTA_STS: 1
```

### 📜 Relevant log output

_No response_

_No response_

georglauterbach:
# Description

This PR reworks subject-rewriting. It disables the default `rewrite_subject` action in Rspamd and renames `SA_SPAM_SUBJECT` to `SPAM_SUBJECT`. Under the hood, I utilized Dovecot sieve scripts to adjust the `Subject` header, which seems to work quite nicely.

Review commit by commit.

@polarathene The commit history contains d0eccd322f1c967bfd3a1527ae906002067066c4, which is not strictly related to this PR, but the new tests (like 'SPAM_SUBJECT works') make use of it. I hope it's not too far out of the boundaries to justify another PR. I have only adjusted the Rspamd test file to keep the diff minimal and will provide a follow-up to adjust the whole test suite.

Fixes #3804
Fixes #3323

## Type of change

- [x] **This is a breaking change**

## Checklist:


target/scripts/startup/setup.d/security/rspamd.sh
```diff
@@ -304,6 +305,56 @@ function __rspamd__setup_check_authenticated() {
...
+    local DOVECOT_SIEVE_GLOBAL_BEFORE_DIR='/usr/lib/dovecot/sieve-global/before'
+    local DOVECOT_SIEVE_FILE='rspamd_spam_subject'
+    readonly DOVECOT_SIEVE_GLOBAL_BEFORE_DIR DOVECOT_SIEVE_FILE
+
+    mkdir -p "${DOVECOT_SIEVE_GLOBAL_BEFORE_DIR}"
+    # ref: https://superuser.com/a/1502589
+    cat >"${DOVECOT_SIEVE_GLOBAL_BEFORE_DIR}/${DOVECOT_SIEVE_FILE}.sieve" << EOF
+require ["editheader","variables"];
+
+if header :contains "X-Spam" "Yes"
```
williamdes, line 334: could it also handle `X-Spam-Flag` ?
https://is.oregonstate.edu/exchange-email-services/spamassassin

I am not sure this is right or not
I have this file that adds this header: https://github.com/wdes/mails.wdes.eu/blob/f40c1fec5a67b3448df4a536bf790d1ae215ec57/rspamd/local.d/milter_headers.conf#L16

georglauterbach: We could easily add something like `anyof (header :contains ...` to match multiple headers. The question is: how many do we want to support? One idea would be to have a new ENV like `RSPAMD_SPAM_HEADER_NAME` so we have a uniform name.

williamdes: Looks like a good idea to have `RSPAMD_SPAM_HEADER_NAME`

Also, the value may vary as I have observed on other repos, like https://github.com/mailcow/mailcow-dockerized/blob/2024-01/data/conf/dovecot/global_sieve_after#L11

Sometimes it is uppercase, some other times not. Depending on the example/tutorial/app that you are reading

georglauterbach: > PS: move to junk does not work at all for me for unknown reasons, maybe this integration will make it work better

It probably does not work because of this:

https://github.com/docker-mailserver/docker-mailserver/blob/00018e7e2b226fafb3a5ab50db078885ca6c8ddc/target/scripts/startup/setup.d/security/misc.sh#L249-L270

It's the same story as with this PR I guess.

---

Having `RSPAMD_SPAM_HEADER(_NAME)` is something I can implement in a follow-up PR.

docs/content/config/environment.md
```diff
@@ -446,6 +446,12 @@ Can be used to control the score when the [`HFILTER_HOSTNAME_UNKNOWN` symbol](#r
 
 Default: 6 (which corresponds to the `add_header` action)
 
+##### RSPAMD_SPAM_SUBJECT
+
+This variable defines a prefix for e-mails tagged with the `X-Spam: Yes` header. In general, it works similar to [`SA_SPAM_SUBJECT`](#sa_spam_subject). However, the default is handled differently.
+
+Default: empty (no prefix will be added to e-mails)
```
peterhirn, line 453: Big fan of changing this default 🙏

My main argument is that users can't edit subjects of received emails, and false-positives are annoying to reply/forward/etc.

georglauterbach: This was my initial instinct, and I agree with you, but we had a recent PR (I'll look it up) where we had a discussion on this default. I will need to take this into consideration.

I have an idea though: we should log a warning in case the user

1. did not set `MOVE_SPAM_TO_JUNK` **and**
2. did not set `RSPAMD_SPAM_SUBJECT`

in order to make everyone aware that spam might not be easily visible right away.

target/scripts/startup/setup.d/security/rspamd.sh
```diff
@@ -304,6 +305,56 @@ function __rspamd__setup_check_authenticated() {
...
 }
 
+# If `RSPAMD_SPAM_SUBJECT` is not empty, we create a Sieve script that alters the `Subject`
+# header, in order to prepend a user-defined string.
+function __rspamd__setup_spam_subject() {
+  if [[ -z ${RSPAMD_SPAM_SUBJECT} ]]
+  then
+    __rspamd__log 'debug' 'Spam subject is not set - no prefix will be added to spam e-mails'
+  else
+    __rspamd__log 'debug' "Spam subject is set - the prefix '${RSPAMD_SPAM_SUBJECT}' will be added to spam e-mails"
```
polarathene, line 315: This is for a major release where breaking changes could be introduced.

It should be viable to avoid introducing a separate rspamd equivalent here, and rename `SA_SPAM_SUBJECT` to `DMS_SPAM_SUBJECT` or similar.

Breaking:
- Switch to opt-in by default with explicit value. It doesn't make sense to use in IMAP, while POP3 has been proven to require other explicit config changes (_but I'm open to complicating this for POP3 to default to a fixed value_).
- ENV rename, we could just log warning or temporarily handle `SA_SPAM_SUBJECT` when it's configured explicitly.
- Handled in Dovecot instead of Postfix... the subject rewrite would only happen when delivering locally, other external MTAs mail is forwarded / relayed to would have to do their own spam subject modification on their end (_and maybe they do, but then our current DMS support would presumably prepend the subject twice then which is a bug IMO_).

---

I just don't see the point in separate logic for the two when it's effectively the same feature?

image

Granted.. the `_SCORE_` feature is an extra.
- Sieve can do that similar to how the subject header value was extracted. Presumably if `spamtest` is configured for a score, you'd also have generic sieve script access to that value (_or the confidence percent representation_).
- ~~Little bit more involved to support, but looking at [this sieve language reference](https://proton.me/support/sieve-advanced-custom-filters), it should be possible to match a variable containing `*_SCORE_*` where each `*` becomes `${1}` and `${2}` we can reconstruct the spam prefix.~~ Scratch that we have bash 😆 

```bash
# Use sed to replace the `_SCORE_` token to fixed string `${spam_score}` for referencing a sieve variable:
local PREFIX_SUBJECT_IF_SPAM=sed 's/_SCORE_/${spam_score}/' <<< "${DMS_SPAM_SUBJECT}"
```

```sieve
if header :matches "X-Spam-Score" "*" {
  set "spam_score" "\${1}";
}
....
addheader :last "Subject" "${PREFIX_SUBJECT_IF_SPAM}\${subject}";
```

For rspamd, the FAQ [states that the score isn't really intended for external use](https://www.rspamd.com/doc/faq.html#why-do-i-have-zero-score-for-a-spam-message), they encourage using actions.
- [`spam-header`](https://www.rspamd.com/doc/modules/milter_headers.html#spam-header) An alternative approach to having rspamd configure a header that a sieve script may operate on.
- [`x-spamd-result`](https://www.rspamd.com/doc/modules/milter_headers.html#x-spamd-result-158) (_which our docs touch on a little bit_), the bulky series of score values, where I didn't find a clear explanation for the two values in the `default` field it contains.
- [`x-spam-level`](https://www.rspamd.com/doc/modules/milter_headers.html#x-spam-level) similar to SpamAssassin header, compatible with one of the Dovecot Sieve spamtest configs (`strlen`).

If wanting to have this `_SCORE_` feature to be anti-spam service agnostic, that may need to be explored further 🤷‍♂️

polarathene (dismissed):
This is rspamd, so I won't block.

I don't particularly like some of the decisions here. We are going to publish this with v14, we have an opportunity to just be anti-spam service agnostic instead with a common ENV for this functionality.

Maybe @casperklein wants to weigh in, as he's expressed his own concerns/preferences with SpamAssassin, especially with this type of functionality.

docs/content/config/environment.md
polarathene: From what I understand from our rspamd docs, a score of 11, is similar to triggering `SA_KILL` (_a score that identifies "definitely spam", or at least is meant to be fairly confident_), which we have a separate ENV `SPAMASSASSIN_SPAM_TO_INBOX` to prevent the action of reject/bounce/discard, and always deliver mail regardless to spam... but there is no equivalent support for rspamd integration right?

For SpamAssassin:
- We deliver all spam by default based on [this](https://github.com/docker-mailserver/docker-mailserver/issues/1396#issuecomment-597210154) reasoning.
- I also compiled [these findings for SA scores](https://github.com/docker-mailserver/docker-mailserver/pull/3058#issuecomment-1420268148), one had a [legitimate mail with a score of `8`](https://github.com/docker-mailserver/docker-mailserver/issues/768). There was also another request to always deliver by default (_since it'd go into the junk folder anyway_).

---

> It was argued that the prefix is useful, and I concur.

It's not that useful without explicit ENV changes?
- [We concluded that the value was primarily for POP3 users](https://github.com/docker-mailserver/docker-mailserver/pull/3767#discussion_r1456140100), no known IMAP users have an interest in it with the default to store spam in a separate junk folder. Perhaps if they used the `_SCORE_` token with a custom subject set, but again that is opt-in.
- We agreed that emitting a warning would be good `if IMAP=0 and POP3=1 and MOVE_SPAM_TO_JUNK=1` (_a spam prefix to a subject won't be valid here, hence the warning_).

Follow-up issues for changes related to that discussion:
- https://github.com/docker-mailserver/docker-mailserver/issues/3788
- https://github.com/docker-mailserver/docker-mailserver/issues/3791

---

- Make this feature anti-spam service agnostic if possible.
- Not enabled by default (breaking change)
- `undef` needs to be set for SA, and communicated that'd be no longer necessary to set.
- When it is configured, it should emit a warning log if `MOVE_SPAM_TO_JUNK=1` && `ENABLE_POP3=1` (_not sure if `ENABLE_IMAP=1` has relevance to preventing that sort of warning_).

target/scripts/startup/setup.d/security/rspamd.sh
georglauterbach: I concur, having too many ENVs is problematic. Let's wait with tihs.

target/scripts/startup/setup.d/security/rspamd.sh
```diff
@@ -304,6 +305,56 @@ function __rspamd__setup_check_authenticated() {
...
+
+    # We can't "replace" a header, but we can delete (all instances of) it and
+    # re-add (a single instance of) it:
+    deleteheader "Subject";
+    # Append/prepend as you see fit
+
+    addheader :last "Subject" "${RSPAMD_SPAM_SUBJECT}\${subject}";
+    # Note that the header is added ":last" (so it won't appear before possible
+    # "Received" headers).
+}
```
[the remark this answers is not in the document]
georglauterbach: I did not hide it: https://github.com/docker-mailserver/docker-mailserver/pull/3820/files#diff-29f67ebf0da3c8d5c483168dd3cbc74a3d14f4752fb0c67962d19cb20ea9ff0fR330 :D

target/scripts/startup/setup.d/security/rspamd.sh
georglauterbach: I don't like the `_SCORE_` feature at all (because of maintainability concerns), but I agree that this should be agnostic to the choice of anti-spam software. I will rebase this PR.

target/scripts/startup/setup.d/security/rspamd.sh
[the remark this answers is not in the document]
polarathene: Oh I am blind 🙈 😂

target/scripts/startup/setup.d/security/rspamd.sh
polarathene: Perhaps add a deprecation notice with it then to encourage anyone that actually depends upon it to voice that concern.

Sometimes I think we carry features that only have a few users (namely the original contributor) who may no longer use DMS 🤷‍♂️ 

It doesn't seem too difficult to support, but at the same time with a sieve solution, we technically don't have to carry it anymore ourselves and can just document an example sieve script that could be added / adapted. Then we don't have to maintain any extra logic/ENV 👍

georglauterbach: I will add the deprecation notice in the changelog.

georglauterbach:
I have reset the changes and started afresh (hence the force push, and, a need to re-review everything 😆). The new var `SPAM_SUBJECT` is agnostic of the underlying anti-spam system. This should better reflect maintainer's views (including mine) and help transitioning between anti-spam systems.

---

PS: @polarathene noted there is no equivalent of `SA_KILL` in conjunction with `SPAMASSASSIN_SPAM_TO_INBOX`. This is because IMO this is really useless - just set `reject = 100;` in Rspamd's configuration. This way, no e-mail will be rejected if that's what you're after. Just a side note, unrelated to changes made in this PR.

target/scripts/startup/setup.d/security/rspamd.sh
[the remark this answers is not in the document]
georglauterbach: I will stick to the easier approach :D

CHANGELOG.md
```diff
@@ -32,11 +32,15 @@ The most noteworthy change of this release is the update of the container's base
       - DMS `main.cf` has renamed `postscreen_dnsbl_whitelist_threshold` to `postscreen_dnsbl_allowlist_threshold` as part of this change.
     - `smtpd_relay_restrictions` (relay policy) is now evaluated after `smtpd_recipient_restrictions` (spam policy). Previously it was evaluated before `smtpd_recipient_restrictions`. Mail to be relayed via DMS must now pass through the spam policy first.
     - The TLS fingerprint policy has changed the default from MD5 to SHA256 (_DMS does not modify this Postfix parameter, but may affect any user customizations that do_).
+- **Environment variables**:
+  - `SA_SPAM_SUBJECT` has been renamed into `SPAM_SUBJECT` to incorporate Rspamd (for which the default settings changed, disabling the `rewrite_subject` action); a notification is logged when using the old `SA_SPAM_SUBJECT` and the value is copied as a fallback ([3820](https://github.com/docker-mailserver/docker-mailserver/pull/3820))
```
polarathene, line 36: This is what I was expecting changelog wise, but as noted in other feedback it won't be entirely accurate with your current changeset. 

```suggestion
- **Environment Variables**:
  - `SA_SPAM_SUBJECT` has been renamed into `SPAM_SUBJECT` to become anti-spam service agnostic. ([3820](https://github.com/docker-mailserver/docker-mailserver/pull/3820))
    - As this functionality is now handled in Dovecot via a Sieve script instead of the respective anti-spam service during Postfix processing, this feature will only apply to mail stored in Dovecot. If you have relied on this feature in a different context, it will no longer be available.
    - Rspamd previously handled this functionality via the `rewrite_subject` action which as now been disabled by default in favor of the new approach with `SPAM_SUBJECT`.
    - `SA_SPAM_SUBJECT` is now deprecated and will log a warning if used. The value is copied as a fallback to `SPAM_SUBJECT`.
    - The default has changed to not prepend any prefix to the subject unless configured to do so. If you relied on the implicit prefix, you will now need to provide one explicitly.
    - `undef` was previously supported as an opt-out with `SA_SPAM_SUBJECT`. This is no longer valid, the equivalent opt-out value is now an empty value (_or rather the omission of this ENV being configured_).
    - The feature to include [`_SCORE_` tag](https://spamassassin.apache.org/full/4.0.x/doc/Mail_SpamAssassin_Conf.html#rewrite_header-subject-from-to-STRING) in your value to be replaced by the associated spam score is no longer available.
```

CHANGELOG.md
```diff
@@ -32,11 +32,15 @@ The most noteworthy change of this release is the update of the container's base
...
 
 ### Updates
 
 - **Tests:**
   - Refactored helper methods for sending e-mails with specific `Message-ID` headers and the helpers for retrieving + filtering logs, which together help isolate logs relevant to specific mail when multiple mails have been processed within a single test. ([#3786](https://github.com/docker-mailserver/docker-mailserver/pull/3786))
+- **Rspamd**:
+  - the `rewrite_subject` action, which was enabled by default, has been disabled; a Sieve script has been added in order to allow subject rewriting agnostic to the underlying anti-spam system ([3820](https://github.com/docker-mailserver/docker-mailserver/pull/3820))
```
polarathene, line 43:
```suggestion
  - The `rewrite_subject` action, is now disabled by default. It has been replaced with the new `SPAM_SUBJECT` environment variable, which implements the functionality via a Sieve script instead in favor of being anti-spam service agnostic ([3820](https://github.com/docker-mailserver/docker-mailserver/pull/3820))
```

mailserver.env
```diff
@@ -400,12 +400,6 @@ SA_TAG2=6.31
 # triggers spam evasive actions
 SA_KILL=10.0
 
-# add tag to subject if spam detected
-# The value `undef` opts-out of this feature. The value shown below is the default.
-# NOTE: By default spam is delivered to a junk folder, reducing the value of adding a subject prefix.
-# NOTE: If not using Docker Compose, other CRI may require the single quotes removed.
-#SA_SPAM_SUBJECT='***SPAM*** '
-
```
polarathene, line 408: This file doesn't have a `SPAM_SUBJECT` replacement?

The web UI won't allow me to provide a suggestions here, but you'd want to commit this:

```env
# Add the value as prefix to the mail subject when spam is detected
# NOTE: By default spam is delivered to a junk folder, reducing the value of a subject prefix for spam.
# NOTE: If not using Docker Compose, other CRI may require the single quotes removed.
SPAM_SUBJECT=
```

target/rspamd/local.d/actions.conf
```diff
@@ -6,7 +6,12 @@
 # and to be able to explain the impact on the whole system.
 greylist = 4;
 add_header = 6;
-rewrite_subject = 7;
 reject = 11;
 
-subject = "***SPAM*** %s"
+# The value `null` disabled the action. A subject rewrite is handled by
+# SPAM_SUBJECT
+# (https://docker-mailserver.github.io/docker-mailserver/edge/config/environment/#spam_subject)
```
polarathene, line 13:
```suggestion
# The value `null` disabled the action. A subject rewrite is handled by `SPAM_SUBJECT`:
# https://docker-mailserver.github.io/docker-mailserver/latest/config/environment/#spam_subject
```

target/scripts/startup/check-stack.sh
```diff
@@ -52,3 +52,12 @@ function _check_log_level() {
     LOG_LEVEL="${DEFAULT_LOG_LEVEL}"
   fi
 }
+
+function _check_spam_prefix() {
+  # This check should be independent of ENABLE_POP3 and ENABLE_IMAP
+  if _env_var_expect_zero_or_one MOVE_SPAM_TO_JUNK \
+  && [[ ${MOVE_SPAM_TO_JUNK} -eq 0 ]] \
```
polarathene, line 59: I don't see the value in this first condition:

https://github.com/docker-mailserver/docker-mailserver/blob/47f8d50bebaa0b1f1cb85d45c1d9210d9ce80a49/target/scripts/helpers/utils.sh#L155-L162

If you want to sanitize the variables, that should be handled in `variables-stack.sh`. Centralizing it there means we can keep our expectations simple elsewhere in the scripts that follow.

It should not be required at this point to check that the ENV if `-eq 0` to proceed.

```suggestion
  if [[ ${MOVE_SPAM_TO_JUNK} -eq 0 ]] \
```

target/scripts/startup/setup.d/security/rspamd.sh
```diff
@@ -304,6 +305,55 @@ function __rspamd__setup_check_authenticated() {
   fi
 }
 
+# If `SPAM_SUBJECT` is not empty, we create a Sieve script that alters the `Subject`
+# header, in order to prepend a user-defined string.
+function __rspamd__setup_spam_subject() {
```
polarathene, line 310: This is no longer rspamd specific, the method doesn't belong in `rspamd.sh` anymore, and it should not be reliant on it's related helper for logging?

target/scripts/startup/setup.d/security/misc.sh
```diff
@@ -81,12 +81,12 @@ function __setup__security__spamassassin() {
...
     sed -i -r 's|^\$sa_kill_level_deflt (.*);|\$sa_kill_level_deflt = '"${SA_KILL}"';|g' /etc/amavis/conf.d/20-debian_defaults
 
-    if [[ ${SA_SPAM_SUBJECT} == 'undef' ]]; then
+    if [[ -z ${SPAM_SUBJECT} ]]; then
       # shellcheck disable=SC2016
       sed -i -r 's|^\$sa_spam_subject_tag (.*);|\$sa_spam_subject_tag = undef;|g' /etc/amavis/conf.d/20-debian_defaults
     else
       # shellcheck disable=SC2016
-      sed -i -r 's|^\$sa_spam_subject_tag (.*);|\$sa_spam_subject_tag = '"'${SA_SPAM_SUBJECT}'"';|g' /etc/amavis/conf.d/20-debian_defaults
+      sed -i -r 's|^\$sa_spam_subject_tag (.*);|\$sa_spam_subject_tag = '"'${SPAM_SUBJECT}'"';|g' /etc/amavis/conf.d/20-debian_defaults
```
polarathene, line 89: I guess this is why you still have an rspamd specific approach with the sieve script. In which case my changelog suggestion for breaking changes is a bit inaccurate.

It should be fine to ensure the config is undef here, and use the same sieve script though, as the changelog suggests?

The main benefit of keeping this logic is `_SCORE_` tag will still be usable for Amavis+SpamAssassin, and if there was a DMS delivering spam to a different service rather than our bundled Dovecot service, it'd still keep the spam prefix in the subject.

target/scripts/startup/setup.d/security/rspamd.sh
polarathene: Given the context with Amavis+SA still being supported, this could be left as-is, and addressed in a follow-up PR separately.

I think it would be better to transition Amavis+SA to using the same approach as the plans for DMS is to switch to rspamd by default and potentially remove Amavis+SA at a later date, it'd help prep for that switch if we have user reports about the subtle functionality changes involved (_only our Dovecot service provides the functionality, no more `_SCORE_` tag support_).

polarathene (changes requested):

polarathene:
For visibility / discoverability, since the feedback has been hidden, related Sieve insights / advice related to spam:

- [Spam extensions](https://github.com/docker-mailserver/docker-mailserver/pull/3820/files#r1465773312) for better integration / logic and support if ever necessary.
- [Sieve example] for supporting `_SCORE_` tag functionality of Amavis+SA. Also includes references to several rspamd headers
- [Supporting both anti-spam services spam flag headers in Sieve](https://github.com/docker-mailserver/docker-mailserver/pull/3820#discussion_r1465780416)
- [Related issues / background](https://github.com/docker-mailserver/docker-mailserver/pull/3820#discussion_r1465850554), including a difference in support / behaviour between Amavis+SA and rspamd for ensuring delivery.

polarathene:
> there is no equivalent of `SA_KILL` in conjunction with `SPAMASSASSIN_SPAM_TO_INBOX`. This is, IMO, useless with Rspamd: just set `reject = 100;` in Rspamd's configuration. This way, no e-mail will be rejected if that's what you're after

`SA_KILL` is basically the equivalent of your `reject` rspamd setting, except it was agnostic to the action taken, it was a score threshold that triggered a different final action, however DMS only supported adjusting that action between `D_PASS` and `D_BOUNCE` via `SPAMASSASSIN_SPAM_TO_INBOX` toggle.

In the past users had done the same workaround approach by setting `SA_KILL` to `1000`. However `SA_KILL` also triggered some other behaviour such as adding a copy to quarantine. It was basically a marker for what is definitely spam, and the quarantine could be used for spam training, while still allowing the mail through to the junk folder.

I don't know how useful either of the two settings really are for users to configure though, just that we previously defaulted to `D_BOUNCE` and that was not what users were happy with, but at the time `SA_KILL` was set to `6.31` like `SA_TAG2`, which meant if it was considered as flagged for spam it'd be bounced without the user knowing. Legitimate mail was scoring `8` in some cases. `SA_KILL` was later bumped to `10`, which came after changing the default `D_BOUNCE` to `D_PASS` (deliver to users junk).

I don't know how well rspamd `reject` maps to SA scores, but if they're roughly equal I guess that's fine and we won't have any user complaints 🤷‍♂️ Just bringing up that difference in support between the two services in DMS for awareness.

mailserver.env
georglauterbach: I had #3791 in mind, but I guess doing it with 1 PR is the best course of action.

target/scripts/startup/setup.d/security/misc.sh
georglauterbach: I will change to only Sieve as the only mechanism with the next commits, hence your suggestions for the changelog are correct. The `_SCORE_` logic will be removed as well; users that require it can manually patch the configuration.

target/scripts/startup/setup.d/security/rspamd.sh
georglauterbach: This will be solved with the next commits.

mailserver.env
```diff
@@ -134,6 +134,11 @@ ENABLE_IMAP=1
 # **0** => Disabled
 ENABLE_CLAMAV=0
 
+# Add the value as prefix to the mail subject when spam is detected
+# NOTE: By default spam is delivered to a junk folder, reducing the value of a subject prefix for spam.
+# NOTE: If not using Docker Compose, other CRI may require the single quotes removed.
+SPAM_SUBJECT=
```
polarathene, line 140:
```suggestion
# Add the value as a prefix to the mail subject when spam is detected.
# NOTE: By default spam is delivered to a junk folder, reducing the value of a subject prefix for spam.
# NOTE: When not using Docker Compose, other CRI may not support quote-wrapping the value here to preserve any trailing white-space.
SPAM_SUBJECT=
```

test/tests/parallel/set1/spam_virus/amavis.bats
```diff
@@ -41,10 +45,9 @@ function teardown_file() {
   _run_in_container grep -F '127.0.0.1:10025' /etc/postfix/master.cf
   assert_success
 
-  _run_in_container_bash '[[ -f /etc/cron.d/amavisd-new.disabled ]]'
-  assert_failure
-  _run_in_container_bash '[[ -f /etc/cron.d/amavisd-new ]]'
+  _run_in_container_bash '[[ ! -f /etc/cron.d/amavisd-new.disabled ]]'
```
polarathene, line 48: Might warrant a `_file_should_not_exist_in_container`? 🤔 (_along with the other related checks in the file_)

test/tests/parallel/set1/spam_virus/amavis.bats
```diff
@@ -55,29 +58,25 @@ function teardown_file() {
...
   assert_success
-  assert_output --partial '= 6.31'
+  assert_output --partial '$sa_tag2_level_deflt = 2.0;'
 
-  # shellcheck disable=SC2016
-  _run_in_container grep '\$sa_kill_level_deflt' "${AMAVIS_DEFAULTS_FILE}"
+  _run_in_container grep 'sa_kill_level_deflt' "${AMAVIS_DEFAULTS_FILE}"
   assert_success
-  assert_output --partial '= 10.0'
+  assert_output --partial '$sa_kill_level_deflt = 3.0;'
```
polarathene, line 75: I think we've got two different tests here.

One that has been deleted is the default values being checked they were set/applied correctly as per our docs.

While the other ones you've carried over here are for verifying the ENV support to override (_although technically both go through that same flow these days_).

Do you not see value in the two distinct tests anymore? I don't mind either way.

test/tests/parallel/set1/spam_virus/amavis.bats
```diff
@@ -55,29 +58,25 @@ function teardown_file() {
...
 
-  # shellcheck disable=SC2016
-  _run_in_container grep '\$sa_spam_subject_tag' "${AMAVIS_DEFAULTS_FILE}"
+  _run_in_container grep 'sa_spam_subject_tag' "${AMAVIS_DEFAULTS_FILE}"
```
polarathene, line 77: Just adding a bit of context.

```suggestion
  # This feature is handled by our SPAM_SUBJECT ENV through a sieve script instead.
  # Thus the feature here should always be disabled via the 'undef' value.
  _run_in_container grep 'sa_spam_subject_tag' "${AMAVIS_DEFAULTS_FILE}"
```

polarathene (changes requested):
Few adjustments and it looks good to go 👍 

May want to wait to see if @casperklein has any input regarding this change impact to SpamAssassin / `SA_SPAM_SUBJECT`?

test/tests/parallel/set1/spam_virus/amavis.bats
georglauterbach: I was trying to keep the diff small so you wouldn't complain about this not belonging in this PR 😆😆😆 but I'll happily add `_file_should_not_exist_in_container` :D

test/tests/parallel/set1/spam_virus/amavis.bats
georglauterbach: I will add a third container that checks the defaults 👍🏼

georglauterbach:
@polarathene It seems like #3838 broke our docs preview deployment (<https://github.com/docker-mailserver/docker-mailserver/actions/runs/7679465094/job/20930390342>)?

test/tests/parallel/set1/spam_virus/amavis.bats
georglauterbach: Done in [ffe3370](https://github.com/docker-mailserver/docker-mailserver/pull/3820/commits/ffe33702270358ae12da227b29b722451015f98b)

test/tests/parallel/set1/spam_virus/amavis.bats
georglauterbach: Done in [c593999](https://github.com/docker-mailserver/docker-mailserver/pull/3820/commits/c593999da8e5c24c7be8d7f1a406977c4f7fdb08)

georglauterbach:
> Few adjustments and it looks good to go 👍

🚀 
 
> May want to wait to see if @casperklein has any input regarding this change impact to SpamAssassin / `SA_SPAM_SUBJECT`?

👍🏼

test/tests/parallel/set1/spam_virus/amavis.bats
polarathene: > so you wouldn't complain about this not belonging in this PR 😆😆😆

You were already halfway there with the inclusion anyway, and since I'm probably the only one providing a full review and we're at the end I don't mind as much 😅 (_but it should have all been handled separately_ **shakes fist** 😤 😝 )

polarathene:
> @polarathene It seems like #3838 broke our docs preview deployment

Oh damn, I didn't pay attention to the difference in similar parameter names `_` => `-`, whoops! 😬

polarathene (dismissed):
@casperklein should be given at least 48 hours to express an interest in any review feedback.
- Sorry about all the notification noise 😅
- See the changelog entry of this PR first, if you don't have any issue with what is proposed there, no further input required beyond an approval 👍

casperklein (changes requested):

casperklein:
I've read https://github.com/docker-mailserver/docker-mailserver/issues/3804. Did I understand it correctly, that rspam is not capable of adding a X-SPAM Header and at the same time changing the subject? That's a pity 😆 

I really don't like when features are reinvented (in sieve) that are build-in and working perfectly in SA 😞 IMO this is the job of the spam filter and not dovecots..

----

Code LGTM. Just one thing, see above.

target/scripts/startup/setup.d/security/misc.sh
```diff
@@ -245,6 +241,56 @@ function __setup__security__amavis() {
   fi
 }
 
+# If `SPAM_SUBJECT` is not empty, we create a Sieve script that alters the `Subject`
+# header, in order to prepend a user-defined string.
+function _setup_spam_subject() {
+  if [[ -z ${SPAM_SUBJECT} ]]
+  then
+    __rspamd__log 'debug' 'Spam subject is not set - no prefix will be added to spam e-mails'
```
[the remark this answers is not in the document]
georglauterbach: Good catch! Addresses in ebcdd961332e90e6282fdfce86732d1e20f21b52

georglauterbach:
> I've read #3804. Did I understand it correctly, that rspam is not capable of adding a X-SPAM Header and at the same time changing the subject? That's a pity 😆

Indeed; I originally assumed this was the case. It really is a shame..

> I really don't like when features are reinvented (in sieve) that are build-in and working perfectly in SA 😞 IMO this is the job of a spam filter and not dovecots..

I agree, but here I guess "it is what it is" fits most. With this change, we're one step closer to being agnostic to the underlying anti-spam (a silver lining).

ghnp5:
Hey!

Thanks very much for this.

Is there a summary somewhere of what the breaking changes will be for this?

Many thanks, again! 🥳

georglauterbach:
Yes, there is - have a look at our changelog (`CHANGELOG.md`) 👍🏼

williamdes:
For reference here is the rspamd issue report I was referring to when I said that I did think that this PR would probably fix my spam rewrote emails not being moved to Spam folder : https://github.com/rspamd/rspamd/issues/3078#issuecomment-1957729891

To be confirmed

PR #3821, opened 2024-01-24
Add warning when container was restarted

casperklein:
# Description

DMS does not (yet) support container restarts and a clean start is necessary. This PR adds a warning when DMS was restarted.

## Type of change

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ All notable changes to this project will be documented in this file. The format 
 
 The most noteworthy change of this release is the update of the container's base image from Debian 11 ("Bullseye") to Debian 12 ("Bookworm"). This update alone involves breaking changes and requires a careful update!
 
+### Added
+
+- **Internal:**
+  - Add warning when DMS was restarted, which is not supported ([#3821](https://github.com/docker-mailserver/docker-mailserver/pull/3821)).
+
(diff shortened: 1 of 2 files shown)


polarathene:
Your warning logs are better, but we technically already have this logic?

https://github.com/docker-mailserver/docker-mailserver/blob/ed1e1ebbd3eeb61b059967d94c54d8881112fba4/target/scripts/startup/check-stack.sh#L17-L24

casperklein:
You are right, I forgot about this 😆

If you and Georg thinks this one is better, I can remove `_check_improper_restart()`. Otherwise we can close this PR.

georglauterbach:
I don't mind this solution of using `if else`, I just like the wording in the current check better TBH. Technically it is also a check; having it in `check-stack.sh` seems most appropriate to me.

polarathene:
Several scripts rely on this file marker, so it's fine and appropriate to keep the check in `check-stack.sh` 👍 

Current wording is good, but the explicit guidance of `Use 'docker compose up --force-recreate'` is also a helpful addition to communicate (_plenty of users that raise bug reports don't seem to actually check the debug docs guidance_), but it's not clear if they're seeing that restart warning and just not sure how to correctly restart/destroy the container by re-recreating it.

PR #3823, opened 2024-01-25
docs(fix): New external link icon workaround for mkdocs-material 9.5.5

### LINKED ISSUE #3822: docs: CSS for external link broke with v13.3.1
### Description

The docs CI should only be using `MAJOR.MINOR`, so a potential bug was introduced via a patch release or the minor was bumped without verifying the docs preview.

Presently v13.3 docs are presenting all links as external (_this is a custom CSS rule we have added, which implies upstream added a change that broke it_).

image

https://github.com/docker-mailserver/docker-mailserver/blob/ed1e1ebbd3eeb61b059967d94c54d8881112fba4/docs/content/assets/css/customizations.css#L19-L20

polarathene:
# Description

This is the easiest to maintain workaround now available. Upstream continues to reject the value such a feature for accessibility.

**Approaches evaluated:**
- Previous CSS workaround has become a maintenance burden to adapt (_becomes very verbose to support deploy/preview/dev environments_).
- An unofficial approach via redirect template as a workaround has a bad UX (_at least when tested locally_) as it loads a placeholder page for the sole purpose of redirecting via a separate HTML document, just to display an icon 🙄 
- A similar approach to the current CSS was available, but requires to be a paid sponsor of the project until they reach $20k monthly funding milestone. That is $7k away and I doubt anyone here is interested in $15/month perpetually for a fix.
- The final alternative suggested that is minimal effort is a bit ugly, and **the chosen solution going forward**. The `mkdocs.yml` file gets some raw HTML included into the two external nav links, this just adds a span element to include a `class` of our own that we can reference in CSS.

Fixes #3822

---

Another alternative that was suggested by the primary mkdocs-material dev is to disable [Instant Navigation](https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/#instant-loading) feature, since that is what broke our current CSS by removing the relative URLs at runtime on the client (browser).

## Type of change

## Checklist:

I don't think this needs a changelog entry?

### DIFF
--- a/docs/content/assets/css/customizations.css
+++ b/docs/content/assets/css/customizations.css
@@ -16,12 +16,16 @@ If you want to append instead, switch `::before` to `::after`.
   src: url('../fonts/external-link.woff') format('woff');
 }
 
-/* Matches the two nav link classes that start with `http` `href` values, regular docs pages use relative URLs instead. */
-.md-tabs__link[href^="http"]::before, .md-nav__link[href^="http"]::before {
+/* 
+   Since mkdocs-material 9.5.5 broke support in our docs from DMS v13.3.1, we now use our own class name,
+   which has been included for the two external nav links in mkdocs.yml via workaround (insert HTML).
+*/
+.icon-external-link::before {
   display: inline-block; /* treat similar to text */
   font-family: 'external-link';
   content:'\0041'; /* represents "A" which our font renders as an icon instead of the "A" glyph */
   font-size: 80%; /* icon is a little too big by default, scale it down */
+  margin-right: 4px;
 }
 
 /* ============================================================================================================= */
--- a/docs/content/assets/css/customizations.css
+++ b/docs/content/assets/css/customizations.css
@@ -16,9 +16,9 @@ If you want to append instead, switch `::before` to `::after`.
   src: url('../fonts/external-link.woff') format('woff');
 }
 
-/* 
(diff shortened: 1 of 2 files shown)


georglauterbach (dismissed):
You will still need to appease the EdirConfig linting gods 🙏🏼🙏🏼🙏🏼

docs/content/assets/css/customizations.css
```diff
@@ -16,12 +16,16 @@ If you want to append instead, switch `::before` to `::after`.
   src: url('../fonts/external-link.woff') format('woff');
 }
 
-/* Matches the two nav link classes that start with `http` `href` values, regular docs pages use relative URLs instead. */
-.md-tabs__link[href^="http"]::before, .md-nav__link[href^="http"]::before {
+/* 
+   Since mkdocs-material 9.5.5 broke support in our docs from DMS v13.3.1, we now use our own class name,
+   which has been included for the two external nav links in mkdocs.yml via workaround (insert HTML).
+*/
```
polarathene, line 22: Praise be! 😨 🙏 

```suggestion
/*
  Since mkdocs-material 9.5.5 broke support in our docs from DMS v13.3.1, we now use our own class name,
  which has been included for the two external nav links in mkdocs.yml via workaround (insert HTML).
*/
```

georglauterbach (approved):

Issue #3753, opened 2024-01-06
bug report: 


MarcS1975:
### 📝 Preliminary Checks

- [X] I tried searching for an existing issue and followed the [debugging docs](https://docker-mailserver.github.io/docker-mailserver/latest/config/debugging/) advice, but still need assistance.

### 👀 What Happened?

DMS is running well but fail2ban is extremely aggressive. When I add new users, they naturally setup 4 or 5 devices with the new credentials and every single time dovecot blocks access, either because they have mistyped credentials or they are using a mail client with automated setup-wizzards (like Thunderbird) and they try different ways of submitting passwords (Normal/encrypted/etc.). How can I loosen the fail2ban parameters?

### 👟 Reproduction Steps

Users attempting to do first logins
after 2 or 3 attempts, fail2ban blocks the IP
 

### 🐋 DMS Version

12.1

### 💻 Operating System and Architecture

Docker

### ⚙️ Container configuration files

```yml
n/a
```

### 📜 Relevant log output

```Text
n/a
```

casperklein:
You can customize fail2ban to your likings:

- https://github.com/docker-mailserver/docker-mailserver/blob/master/config-examples/fail2ban-jail.cf
- https://docker-mailserver.github.io/docker-mailserver/latest/config/advanced/optional-config/

MarcS1975:
Thank you ! 

Do I just map these cfg files via my compose file?
`.MYPATH/config/fail2ban-jail.cf:/etc/fail2ban/jail.d/user-jail.local`

casperklein:
You probably already have mounted "config". If so, then just copy the file in there and restart DMS (`docker compose up --force-recreate`).

MarcS1975:
OK. understand. So I can keep all cfg files I use inside my mounted config folder and DMS will find the right location for each file inside the container at start up?

polarathene:
> DMS will find the right location where the cfg should go inside the container at start up?

[Most of these configs](https://docker-mailserver.github.io/docker-mailserver/v13.2/config/advanced/optional-config/) were introduced early into the project, so they just live in the same folder. In a future major release version they may get better organized.

If there is a nested structure with folders involved, our docs should communicate that 👍 

---

> I can keep all cfg files I use inside my mounted config folder 

[...]

Issue #3546, opened 2023-09-23
question: How to configure Traefik for port 465? (Postfix smtpd throws TLS library error)


mich2k:
### 📝 Preliminary Checks

- [x] I tried searching for an existing issue and followed the [debugging docs](https://docker-mailserver.github.io/docker-mailserver/latest/config/debugging/) advice, but still need assistance.

### 👀 What Happened?

Hello, after having carefully followed every step reported in #3063 I still do not manage in any way running implicit TLS SMTP meanwhile implicit imap works, wondering how to interpret these logs I get

### 👟 Reproduction Steps

Use DMS behind Traefik reverse proxy

### 🐋 DMS Version

v12.1.0

### 💻 Operating System and Architecture

Linux sbserver 5.15.0-83-generic #92-Ubuntu SMP Mon Aug 14 09:30:42 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

### ⚙️ Container configuration files

```yml
dovecot config override:


haproxy_trusted_networks = 172.25.0.2, 32
haproxy_timeout = 3 secs
service imap-login {
  inet_listener imaps {
    haproxy = yes
    ssl = yes
    port = 10993
  }
}
```
postfix-master

```
submission/inet/smtpd_upstream_proxy_protocol=haproxy
submissions/inet/smtpd_upstream_proxy_protocol=haproxy
```
postfix-main

```
postscreen_upstream_proxy_protocol = haproxy
```

container compose

```yml

  mailserver:
    image: ghcr.io/docker-mailserver/docker-mailserver:latest
    restart: always
    container_name: mailserver
    hostname: mail.domain.com
    env_file: mailserver.env
    volumes:
      - ${DOCKER_VOLUMES_FOLDER}/dms/mail-data/:/var/mail/
      - ${DOCKER_VOLUMES_FOLDER}/dms/mail-state/:/var/mail-state/
      - ${DOCKER_VOLUMES_FOLDER}/dms/mail-logs/:/var/log/mail/
      - ${DOCKER_VOLUMES_FOLDER}/dms/config/:/tmp/docker-mailserver/
      - /etc/localtime:/etc/localtime:ro
      - ${DOCKER_VOLUMES_FOLDER}/traefik/acme.json:/etc/letsencrypt/acme.json:ro
    cap_add:
      - NET_ADMIN
    stop_grace_period: 1m
    environment:
      - SSL_TYPE=letsencrypt
[...]

PR #3840, opened 2024-01-27
docs: Remove ENV `ONE_DIR`

polarathene:
# Description

I don't see this as a breaking change given the logic we have for it.

- I have brought this up a few times in the past, this PR finally contributes the suggestion to remove `ONE_DIR`.
- It doesn't really serve a purpose anymore, the `/var/mail-state` location existing is the real detection for triggering the runtime-state consolidation feature.
- I've updated all docs referencing it, and the first commit improves the documentation on our volumes (_I can split this out into a separate PR if preferred, although the `ONE_DIR` changes in docs relates to centralizing the docs for our DMS state volume_).

I've explained DMS Volumes several times in past issues ([example](https://github.com/docker-mailserver/docker-mailserver/issues/3753#issuecomment-1879919623)), so that should be a helpful addition to reference to link users to instead 👍 

## Type of change

## Checklist:


CHANGELOG.md
```diff
@@ -35,6 +35,9 @@ The most noteworthy change of this release is the update of the container's base
 
 ### Updates
 
+- **Environment Variables:**
+  - `ONE_DIR` has been removed (legacy ENV).
```
polarathene, line 39:
```suggestion
  - `ONE_DIR` has been removed (legacy ENV) ([#3840](https://github.com/docker-mailserver/docker-mailserver/pull/3840))
```

polarathene:
Just some thoughts on the lint failure as I didn't expect this (_feel free to ignore this comment_):
- It's only because I modified the last line in VSCode, while other files saved didn't adjust it.
- This contributes to a prior confusion about "final newline" being a line-ending only, but in VSCodes case the default is probably where I got the idea that it required a full blank line 😓 

---

image

```console
# 0 for missing final newline, 1 for present:
$ tail -c 1 docs/content/config/environment.md | wc -l
0
```

We already have that prevented by [a recent PR for user config parsing](https://github.com/docker-mailserver/docker-mailserver/pull/3819) (_where the [last report was related to VSCode defaults](https://github.com/docker-mailserver/docker-mailserver/issues/3546#issuecomment-1907617250)_), and the [ECLint check config](https://github.com/docker-mailserver/docker-mailserver/blob/0c7e49e654fb64a0b23abae2b423f54e8312beeb/test/linting/lint.sh#L19-L34) catches it here in PRs.

---

Perhaps a git pre-commit hook could better prevent it (_I haven't checked if that's implicit, or requires the contributor to [use an editor plugin like for `.editorconfig`](https://github.com/docker-mailserver/docker-mailserver/blob/v13.3.1/.editorconfig#L13)_), there doesn't seem to be an equivalent [`.gitattributes` feature like we do for preventing CRLF line-endings](https://github.com/docker-mailserver/docker-mailserver/pull/3527).

image

I'm wondering if there's any value in highlighting that default setting for VSCode in our docs (_with [required configuration change](https://stackoverflow.com/questions/50162952/vs-code-end-of-file-new-line-not-persisting/50330619#50330619)_) 🤔

VSCode otherwise expects a blank line (_a separate setting enabled by default optionally also renders it as a separate blank line visually_):

image

georglauterbach:
I have `Insert Final Newline` set to true, but I didn't know about `Render Final Newline` (which is set to `dimmed` for me by default).

georglauterbach (approved):
I like the first commit in particular; it always bothered me how we only document the volumes in the FAQ 👍🏼

polarathene:
I'll leave this a little bit longer for @casperklein to possibly chime in since I recall `ONE_DIR` discussions with him in the past too.

casperklein:
Short and sweet:

- Let's remove ONE_DIR 👍 
- I haven't enough time to think this completely through. If there are no drawbacks for people, currently still using ONE_DIR, then it's fine to merge. Otherwise I would wait for the next major release.

---

I am also using this in VSCode:

```
    "files.trimTrailingWhitespace": true,
    "files.insertFinalNewline": true,
```

casperklein (approved):

PR #3841, opened 2024-01-27
docs: Add Debian 12 breaking change for `opendmarc` package

### LINKED ISSUE #3828: v14.0.0: with switching to debian:12 bookworm opendmarc behaviour changes
### Subject

### Description

With switching to debian:12 opendmarc is updated from:
```
opendmarc: OpenDMARC Filter v1.4.0
	SMFI_VERSION 0x1000001
	libmilter version 1.0.1
	Active code options:
		WITH_SPF
		WITH_SPF2
```
to
```
opendmarc: OpenDMARC Filter v1.4.2
	SMFI_VERSION 0x1000001
	libmilter version 1.0.1
	Active code options:
		WITH_SPF
		WITH_SPF2
```

bringing the following changes:
* 1.4.1: https://github.com/trusteddomainproject/OpenDMARC/commit/376720be62dac35e2719e8822f562fc8e0f6bca2
* 1.4.2: https://github.com/trusteddomainproject/OpenDMARC/commit/6925ffe626f9a695d9e834e76b9471100969a06a

The following change in opendmarc:
> Added config option HoldQuarantinedMessages (default false), which
		controls if messages with p=quarantine will be passed on to 
		the mail stream (if False) or placed in the MTA's "hold" 
		queue (if True).  Issue [#105](https://github.com/trusteddomainproject/OpenDMARC/issues/105).  Patch by Marcos Moraes, on
		the OpenDMARC mailing list.

Changes the behaviour how DMS handles such failed messages from putting them in the hold queue to letting them into the INBOX of the user.

The change can be reverted with adding the following to /etc/opendmarc.conf either in user-patches or by building your own image and changing target/opendmarc/opendmarc.conf:
```
HoldQuarantinedMessages true
```

Should we add this behaviour change in the changelog or should we restore the old behaviour? (We are monitoring the hold queue for our servers and therefore prefer the hold queue to decide what's to be done with possible malicious mail but i know that those mails are kinda invisible to other people using DMS as their mailserver)

polarathene:
# Description

Part of the breaking changes for v14 with Debian 12 package upgrade, this config change was released in 1.4.2, while we previously had 1.4.0. Thus upstream would seem to treat it as a bugfix 👍 (_but we should probably still mention this, hence the addition to changelog for awareness_)

Can we also get confirmation from @georglauterbach on what the [dmarc equivalent behaviour in rspamd](https://www.rspamd.com/doc/modules/dmarc.html) has been configured for with DMS? (_regardless of explicitly / implicitly_). We should probably have this in sync.

Fixes #3828

## Type of change

- [x] This change requires a documentation update (CHANGELOG fix)

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,11 @@ The most noteworthy change of this release is the update of the container's base
     - The script now uses `/etc/os-release` to determine the release name of Debian
     - Removed custom installations of Fail2Ban, getmail6 and Rspamd
     - Updated packages lists and added comments for maintainability
+- OpenDMARC upgrade: `v1.4.0` => `v1.4.2`
+  - Previous versions of OpenDMARC would place incoming mail from domains announcing `p=quarantaine` (_that fail the DMARC check_) into the [Postfix "hold" queue](https://www.postfix.org/QSHAPE_README.html#hold_queue) until administrative intervention.
+  - OpenDMARC v1.4.2 has disabled that feature by default, but it can be enabled again by adding the setting `HoldQuarantinedMessages true` to [`/etc/opendmarc.conf`](https://github.com/docker-mailserver/docker-mailserver/blob/v13.3.1/target/opendmarc/opendmarc.conf) (_provided from DMS_).
+    - [Our `user-patches.sh` feature](https://docker-mailserver.github.io/docker-mailserver/latest/config/advanced/override-defaults/user-patches/) provides a convenient approach to updating that config file.
+    - Please let us know if you disagree with the upstream default being carried with DMS, or the value of providing alternative configuration support within DMS.
 - **Postfix:**
   - Postfix upgrade from 3.5 to 3.7 ([#3403](https://github.com/docker-mailserver/docker-mailserver/pull/3403))
     - `compatibility_level` was raised from `2` to `3.6`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,9 +24,9 @@ The most noteworthy change of this release is the update of the container's base
     - The script now uses `/etc/os-release` to determine the release name of Debian
     - Removed custom installations of Fail2Ban, getmail6 and Rspamd
     - Updated packages lists and added comments for maintainability
-- OpenDMARC upgrade: `v1.4.0` => `v1.4.2`
+- OpenDMARC upgrade: `v1.4.0` => `v1.4.2` ([#3841](https://github.com/docker-mailserver/docker-mailserver/pull/3841))
   - Previous versions of OpenDMARC would place incoming mail from domains announcing `p=quarantaine` (_that fail the DMARC check_) into the [Postfix "hold" queue](https://www.postfix.org/QSHAPE_README.html#hold_queue) until administrative intervention.
-  - OpenDMARC v1.4.2 has disabled that feature by default, but it can be enabled again by adding the setting `HoldQuarantinedMessages true` to [`/etc/opendmarc.conf`](https://github.com/docker-mailserver/docker-mailserver/blob/v13.3.1/target/opendmarc/opendmarc.conf) (_provided from DMS_).
+  - [OpenDMARC v1.4.2 has disabled that feature by default](https://github.com/trusteddomainproject/OpenDMARC/issues/105), but it can be enabled again by adding the setting `HoldQuarantinedMessages true` to [`/etc/opendmarc.conf`](https://github.com/docker-mailserver/docker-mailserver/blob/v13.3.1/target/opendmarc/opendmarc.conf) (_provided from DMS_).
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -24,6 +24,11 @@ The most noteworthy change of this release is the update of the container's base
     - The script now uses `/etc/os-release` to determine the release name of Debian
     - Removed custom installations of Fail2Ban, getmail6 and Rspamd
     - Updated packages lists and added comments for maintainability
+- OpenDMARC upgrade: `v1.4.0` => `v1.4.2`
```
polarathene, line 27:
```suggestion
- OpenDMARC upgrade: `v1.4.0` => `v1.4.2` ([#3841](https://github.com/docker-mailserver/docker-mailserver/pull/3841))
```

CHANGELOG.md
```diff
@@ -24,6 +24,11 @@ The most noteworthy change of this release is the update of the container's base
...
+  - Previous versions of OpenDMARC would place incoming mail from domains announcing `p=quarantaine` (_that fail the DMARC check_) into the [Postfix "hold" queue](https://www.postfix.org/QSHAPE_README.html#hold_queue) until administrative intervention.
+  - OpenDMARC v1.4.2 has disabled that feature by default, but it can be enabled again by adding the setting `HoldQuarantinedMessages true` to [`/etc/opendmarc.conf`](https://github.com/docker-mailserver/docker-mailserver/blob/v13.3.1/target/opendmarc/opendmarc.conf) (_provided from DMS_).
```
polarathene, line 29:
```suggestion
  - [OpenDMARC v1.4.2 has disabled that feature by default](https://github.com/trusteddomainproject/OpenDMARC/issues/105), but it can be enabled again by adding the setting `HoldQuarantinedMessages true` to [`/etc/opendmarc.conf`](https://github.com/docker-mailserver/docker-mailserver/blob/v13.3.1/target/opendmarc/opendmarc.conf) (_provided from DMS_).
```

georglauterbach:
> Can we also get confirmation from @georglauterbach on what the [dmarc equivalent behaviour in rspamd](https://www.rspamd.com/doc/modules/dmarc.html) has been configured for with DMS? (_regardless of explicitly / implicitly_). We should probably have this in sync.

DMARC is handled by Rspamd by applying a symbol (with a score) to the result of the DMARC check:

https://github.com/docker-mailserver/docker-mailserver/blob/9ac11021e1c33d87ec706f3f76b6c3197f1f000d/target/rspamd/scores.d/policies_group.conf#L76-L107

georglauterbach (approved):

polarathene:
> DMARC is handled by Rspamd by applying a symbol (with a score) to the result of the DMARC check

Soooo... it is like the behaviour OpenDMARC had before, or has now? Or neither?

We had a similar discussion recently for the `SA_KILL` / `SPAMASSASSING_SPAM_TO_INBOX=1` settings where it was possible to prevent rejection or similar actions by always allowing the mail into junk folder.

Is that the same "fix" workaround? `reject = 100`?

georglauterbach:
> > DMARC is handled by Rspamd by applying a symbol (with a score) to the result of the DMARC check
> 
> Soooo... it is like the behaviour OpenDMARC had before, or has now? Or neither?

I actually don't know whether you can compare this very well... OpenDMARC acts on a single check and "only" gives a decision to Postfix based on this check IIRC. Rspamd on the other hand handles multiple checks (SPF, DKIM, DMARC, denylists, etc.), and will provide a decision based on these multiple checks - the DMARC check is one of them.

> We had a similar discussion recently for the `SA_KILL` / `SPAMASSASSING_SPAM_TO_INBOX=1` settings where it was possible to prevent rejection or similar actions by always allowing the mail into junk folder.
> 
> Is that the same "fix" workaround? `reject = 100`?

Yes, `reject = 100;` is _the_ way of disabling rejection (only GTUBE will still trigger a rejection). Theoretically, a mail could have a score of >100, but the highest I have ever seen was 18. For anyone who can send me an e-mail that gets a score of >50, I'll buy you a 🍺 haha :D

ghnp5:
Hey

I didn't even know about postfix's "hold" queue!

Just checked, and I have loads of messages on hold.
I see they're all spam, so I didn't lose much.

Do you know if it's possible to make quarantined emails to either go to a specific folder (like Spam/Junk), or a separate mailbox?

It seems to be hard to see the emails on "hold", as we have to use `postqueue -p` and then `postcat`, etc.

Thanks!!

polarathene:
> Do you know if it's possible to make quarantined emails to either go to a specific folder (like Spam/Junk), or a separate mailbox?

I am not familiar with working with Postfix hold queue, I know you can check and manipulate the queue but not sure about redirecting that mail.

If the quarantined mail has specific information / metadata in the mail headers that can be matched then a sieve script could probably handle that like we do with redirecting spam delivered to junk folder by default.

Issue #3842, opened 2024-01-27
Setting `RELAY_HOST` forces expectation to provide relay credentials


cfis:
### 📝 Preliminary Checks

- [X] I tried searching for an existing issue and followed the [debugging docs](https://docker-mailserver.github.io/docker-mailserver/latest/config/debugging/) advice, but still need assistance.

### 👀 What Happened?

I have an internal email server that uses a postfix server I have setup on AWS as a relay. My configuration is:

```
RELAY_HOST=host.aws.com
RELAY_PORT=587
```

Notice I do not set `RELAY_USER` nor `RELAY_PASSWORD`. That is because the relay server is configured to only accept traffic  from the IP address of the internal server.

However, by setting RELAY_HOST, SMTP authentication is turned on here:

https://github.com/docker-mailserver/docker-mailserver/blob/9ac11021e1c33d87ec706f3f76b6c3197f1f000d/target/scripts/helpers/relay.sh#L165

Which is called from here:

https://github.com/docker-mailserver/docker-mailserver/blob/9ac11021e1c33d87ec706f3f76b6c3197f1f000d/target/scripts/helpers/relay.sh#L186

This code is a bit too aggressive. Instead, I it should enable SMTP authentication only if `RELAY_USER` or `RELAY_PASSWORD` are set.

### 👟 Reproduction Steps

See above.

### 🐋 DMS Version

v13.3.1

### 💻 Operating System and Architecture

Fedora 39 x64

polarathene:
Interestingly, I was actually touching up the relay host feature docs since they needed some love. unfortunately I don't have the time to refactor the actual scripts (which definitely need work).

![image](https://github.com/docker-mailserver/docker-mailserver/assets/5098581/a786ac06-a2db-4274-a72b-933444664013)

![image](https://github.com/docker-mailserver/docker-mailserver/assets/5098581/8c4dbeac-b05d-4d58-a942-80dad2a6c81f)

---

In your case, without any credentials you'd want to use `main.cf` setting `relayhost` which the ENV `DEFAULT_RELAY_HOST` can handle.

`RELAY_HOST` + `RELAY_PORT` + `RELAY_USER` + `RELAY_PASSWORD` are effectively an entirely different set of ENV for specifying multiple relayhosts with credential mappings. The current implementation for that is also poor 😞

polarathene:
[...]

Issue #3607, opened 2023-10-29
[TODO]: Refactor Relay host support


polarathene:
### Subject

Something else that requires developers attention

### Description

Creating an actual issue for tracking this.

---

I recently responded to a support question referencing the feature guidance in our docs: https://github.com/orgs/docker-mailserver/discussions/3605#discussioncomment-7418198

I've included a technical overview of the file in it's current state, while the file itself has plenty of information in comments and `git blame` should provide more from the original PR of mine that introduced them.

It would still be best to delay until the test suite can adopt `compose.yaml` + DNS approach for a proper relay test between DMS instances.

---

Related relay functionality issue to tackle after that: https://github.com/docker-mailserver/docker-mailserver/issues/3532

Issue #2813, opened 2022-10-04
[FR] Send-only domains


Giotino:
# Feature Request

There is a problem delivering emails to the correct mailserver when the domain exists in the sending mailserver.

## Context

If a domain exists in a mailserver, the server becomes its "final destination", catching all the emails for that domain and not relaying them (if needed).

### Is your Feature Request related to a Problem?

The fact that every domain ends up in the `virtual_mailbox_domains` list (https://www.postfix.org/postconf.5.html#virtual_mailbox_domains) create problems in the following case:

I have my domain `example.org` with all the mailboxes on `mailserver1`. I also use `mailserver2` only to send emails (like from `noreply@example.org`).

The problem is that when I send an email from `noreply@example.org` to `user@example.org` using `mailserver2` it fails becase `mailserver2` can't find `user`.

### Describe the Solution you'd like

There should be a way to prevent domains to be in the `virtual_mailbox_domains` list, marking them as send-only (as they wouldn't be able to receive any emails).

### Are you going to implement it?

Yes, I already have.

https://github.com/Giotino/docker-mailserver/commit/b651e3e93db22cfa0239430ad91d844c47d90b87

I created a new configuration file `sendonly-domains.cf` with the list of domains that shouldn't be in the `virtual_mailbox_domains` list.

### What are you going to contribute??

In addition to the code in the commit, I can add the option to add send-only domains to the setup.sh and then open a PR.

### Alternatives you've considered

I tried to set a relay for the domain to the mailserver with the mailboxes, but it's ignored.

An easy solution would be disabling the spoofing protection and using a mailbox with an "invalid" domain to send, but it will create a security problem where the user can send emails from any address.

### Who will that Feature be useful to?

See the case described before.

### What have you done already?

Implemented it.

### Drawbacks of the implementation

I don't know if it's really a drawback, but I think it should be taken in consideration.

Non-delivery notifications (for example if the destination address doesn't exists) are delivered to the mailserver specified in the MX record of the domain.

**Email aliases and dkim signature seems to not be applied to the mailserver generated notifications.**

polarathene:
> The problem is that when I send an email from `noreply@example.org` to `user@example.org` using `mailserver2` it fails becase `mailserver2` can't find `user`.

You can send a mail from a domain that your mail-server does not manage.

[...]

PR #3845, opened 2024-01-28
refactor: `relay.sh`

polarathene:
# Description

While revising the docs on this feature, I was reminded how frustratingly awkward the current implementation is 😮‍💨

So I've detoured to make some changes there:
- `relay.sh`: Much better maintainer comments + simplified logic + fixes.
- Partially contributes towards https://github.com/docker-mailserver/docker-mailserver/issues/3607
- Minor revisions to `check-for-changes.sh` + `postfix.sh`, primarily `VHOST_UPDATED` addition + comment revisions.
- Additionally allows `DEFAULT_RELAY_HOST` to support credentials without redundantly needing `RELAY_HOST`, which should better help cater to [this recent use-case bug report](https://github.com/docker-mailserver/docker-mailserver/issues/3842) (_the opposite, `RELAY_HOST` when set enforced credentials_).

---

Changes have been staged out into scoped commits with individual commit messages for added context. If you want an easier set of diffs to review through, that'd be a nicer experience to work through 👍 

Two other docs pages for the relay host feature will be revised as a separate follow-up PR.

## Context

- For reference my prior refactor on this functionality was in [May 2022](https://github.com/docker-mailserver/docker-mailserver/pull/2604) where full history of prior contributions was documented.
- In a past discussion (Oct 2023):
  - I [advised avoiding the `RELAY_PASSWORD` ENV for credentials](https://github.com/orgs/docker-mailserver/discussions/3605#discussioncomment-7418198) and just handling that directly in the user configuration file instead.
  - My response there also provides a breakdown of the current `relay.sh` implementation prior to this PR.
- Future work to support relays via implicit TLS (port 465) has [config changes required covered here](https://github.com/docker-mailserver/docker-mailserver/issues/2601#issuecomment-1135520225) (_and also now better described in `relay.sh` comments_) and [this follow-up comment](https://github.com/docker-mailserver/docker-mailserver/issues/2601#issuecomment-1137016558)
- A `compose.yaml` local test environment with several DMS instances configured for relaying is available:
  - Recent basic example, single `compose.yaml` file: https://github.com/docker-mailserver/docker-mailserver/issues/3842#issuecomment-1913380639
  - 2022 `compose.yaml` + multiple files for CoreDNS config: https://github.com/docker-mailserver/docker-mailserver/issues/2601#issuecomment-1135520225
  - Revised version of the 2022 example (_adapted to the reported issue_): https://github.com/docker-mailserver/docker-mailserver/issues/2813#issuecomment-1272223118

## Type of change

## Checklist:


polarathene:
Tests are a bit lacking for this feature presently, although I'm fairly confident the PR isn't doing anything not covered in the changelog.

target/scripts/helpers/relay.sh
```diff
@@ -83,90 +85,94 @@ function _relayhost_sasl() {
...
-    'smtp_sasl_security_options = noanonymous' \
-    'smtp_tls_security_level = encrypt'
+  # Configures each `SENDER_DOMAIN` to send outbound mail through the default `RELAY_HOST` + `RELAY_PORT`
+  # (by adding an entry in `/etc/postfix/relayhost_map`) provided it:
+  # - `/etc/postfix/relayhost_map` doesn't already have it as an existing entry.
+  # - `postfix-relaymap.cf` has no explicit opt-out (SENDER_DOMAIN key exists, but with no relayhost value assigned)
+  #
+  # NOTE: /etc/postfix/vhost represents managed mail domains sourced from `postfix-accounts.cf` and `postfix-virtual.cf`.
+  while read -r SENDER_DOMAIN; do
+    local MATCH_EXISTING_ENTRY="^@${SENDER_DOMAIN}\b"
```
polarathene, line 168: I think for the logic `\s+` is better than `\b`? `@example.co` vs `@example.co.nz`:
- `\b` would match the latter accidentally as `.` / `-` and other chars can introduce a boundary.
- `\s+` expects at least one white-space char present, which should be the case for an existing entry in `/etc/postfix/relayhost_map`.

```suggestion
    local MATCH_EXISTING_ENTRY="^@${SENDER_DOMAIN}\s+"
```

docs/content/config/environment.md
```diff
@@ -1019,36 +1019,84 @@ you to replace both instead of just the envelope sender.
...
+Support for configuring a different port than 25 for `RELAY_HOST` to use.
+
+!!! note
+
+    Requires `RELAY_HOST`.
 
-- **empty** => no default
-- default relay username (if no specific entry exists in postfix-sasl-password.cf)
+#### Relay Host Credentials
+
```
polarathene, line 1078:
```suggestion

!!! warn "Configuring relay host credentials make outbound authentication mandatory"

    Presently when `RELAY_USER` + `RELAY_PASSWORD` or `postfix-sasl-password.cf` are configured, all outbound mail traffic is configured to require a secure connection established and forbids the omission of credentials.
    
    Additional feature work is required to only enforce these requirements on mail sent through a configured relay host.

```

target/scripts/helpers/postfix.sh
```diff
@@ -43,21 +43,26 @@ function _vhost_collect_postfix_domains() {
...
+    while IFS=$'|' read -r FIRST_FIELD _; do
+      # It is expected valid lines have the format local-part@domain-part:
+      DOMAIN=$(echo "${FIRST_FIELD}" | cut -d @ -f2)
+
       echo "${DOMAIN}" >>"${TMP_VHOST}"
     done < <(_get_valid_lines_from_file "${DATABASE_ACCOUNTS}")
   fi
 
-  # getting domains FROM mail aliases
+  # Extract domains from virtual alias config:
```
polarathene, line 56:
```suggestion
  # TODO: Consider if virtual aliases should be configured to the same vhost file:
  # https://github.com/docker-mailserver/docker-mailserver/issues/2813#issuecomment-1272394563
  # Extract domains from virtual alias config:
```

polarathene:
## ENV docs visual comparison

### Before

> image

### After

> image

georglauterbach:
I will review in an hour or so 🚀

target/scripts/helpers/postfix.sh
```diff
@@ -43,21 +43,26 @@ function _vhost_collect_postfix_domains() {
   local DATABASE_VIRTUAL='/tmp/docker-mailserver/postfix-virtual.cf'
   local DOMAIN UNAME
 
-  # getting domains FROM mail accounts
+  # Extract domains from mail accounts:
   if [[ -f ${DATABASE_ACCOUNTS} ]]; then
-    while IFS=$'|' read -r LOGIN _; do
-      DOMAIN=$(echo "${LOGIN}" | cut -d @ -f2)
+    while IFS=$'|' read -r FIRST_FIELD _; do
```
georglauterbach, line 48: While `FIRST_FIELD` is absolutely correct, IMO it'd be more useful to state what the first field actually is and use this as a name.

target/scripts/helpers/postfix.sh
```diff
@@ -43,21 +43,26 @@ function _vhost_collect_postfix_domains() {
...
+      # It is expected valid lines have the format local-part@domain-part:
+      DOMAIN=$(echo "${FIRST_FIELD}" | cut -d @ -f2)
```
georglauterbach, line 50:
```suggestion
      DOMAIN=$(cut -d '@' -f 2 <<< "${FIRST_FIELD}")
```

target/scripts/helpers/postfix.sh
```diff
@@ -43,21 +43,26 @@ function _vhost_collect_postfix_domains() {
...
-  # getting domains FROM mail aliases
+  # Extract domains from virtual alias config:
+  # Aliases may have the forms: 'local-part@domain-part', only 'local-part', or '@domain-part' (wildcard catch-all)
   if [[ -f ${DATABASE_VIRTUAL} ]]; then
-    while read -r FROM _; do
-      UNAME=$(echo "${FROM}" | cut -d @ -f1)
-      DOMAIN=$(echo "${FROM}" | cut -d @ -f2)
+    while read -r FIRST_FIELD _; do
+      UNAME=$(echo "${FIRST_FIELD}" | cut -d @ -f1)
+      DOMAIN=$(echo "${FIRST_FIELD}" | cut -d @ -f2)
```
georglauterbach, line 61: While `FIRST_FIELD` is absolutely correct, IMO it'd be more useful to state what the first field actually is and use this as a name. Moreover:

```suggestion
    while read -r FIRST_FIELD _; do
      UNAME=$(cut -d '@' -f 1 <<< "${FIRST_FIELD}")
      DOMAIN=$(cut -d '@' -f 2 <<< "${FIRST_FIELD}")
```

target/scripts/helpers/relay.sh
```diff
@@ -146,20 +151,22 @@ function _populate_relayhost_map() {
...
     _list_domain_parts "${PRINT_DOMAIN_PART_VIRTUAL}" /tmp/docker-mailserver/postfix-virtual.cf
   } | sort -u | while read -r DOMAIN_PART; do
-    # DOMAIN_PART not already present in `/etc/postfix/relayhost_map`, and not listed as a relay opt-out domain in `postfix-relaymap.cf`
-    # `^@${DOMAIN_PART}\b` - To check for existing entry, the `\b` avoids accidental partial matches on similar domain parts.
-    # `^\s*@${DOMAIN_PART}\s*$` - Matches line with only a domain part (eg: @example.test) to avoid including a mapping for those domains to the RELAY_HOST.
-    if ! grep -q -e "^@${DOMAIN_PART}\b" /etc/postfix/relayhost_map && ! grep -qs -e "^\s*@${DOMAIN_PART}\s*$" "${DATABASE_RELAYHOSTS}"; then
+    local MATCH_EXISTING_ENTRY="^@${DOMAIN_PART}\b"
+    local MATCH_OPT_OUT_LINE="^\s*@${DOMAIN_PART}\s*$"
+
+    if ! grep -q -e "${MATCH_EXISTING_ENTRY}" /etc/postfix/relayhost_map && ! grep -qs -e "${MATCH_OPT_OUT_LINE}" "${DATABASE_RELAYHOSTS}"; then
```
georglauterbach, line 171: Is the `-e` required for this expression? If not, I'd like you to omit it, because I always need to look up this parameter just to discover that in invocations of `grep` with only one pattern, the parameter is not relevant at all.

target/scripts/helpers/relay.sh
georglauterbach: Agreed!

target/scripts/helpers/relay.sh
```diff
@@ -131,42 +139,28 @@ function _multiple_relayhosts() {
...
+  #
+  # NOTE: /etc/postfix/vhost represents managed mail domains sourced from `postfix-accounts.cf` and `postfix-virtual.cf`.
+  while read -r SENDER_DOMAIN; do
+    local MATCH_EXISTING_ENTRY="^@${SENDER_DOMAIN}\b"
+    local MATCH_OPT_OUT_LINE="^\s*@${SENDER_DOMAIN}\s*$"
 
     if ! grep -q -e "${MATCH_EXISTING_ENTRY}" /etc/postfix/relayhost_map && ! grep -qs -e "${MATCH_OPT_OUT_LINE}" "${DATABASE_RELAYHOSTS}"; then
-      _log 'trace' "Adding relay mapping for ${DOMAIN_PART}"
-      echo "@${DOMAIN_PART}    $(_env_relay_host)" >> /etc/postfix/relayhost_map
+      _log 'trace' "Configuring ${SENDER_DOMAIN} for the default relayhost '${RELAY_HOST}'"
```
georglauterbach, line 172: Consistency (picky, I know):

```suggestion
      _log 'trace' "Configuring '${SENDER_DOMAIN}' for the default relayhost '${RELAY_HOST}'"
```

target/scripts/helpers/relay.sh
```diff
@@ -82,14 +85,20 @@ function _relayhost_sasl() {
     postconf 'smtp_sender_dependent_authentication = yes'
   fi
 
-  # Add an authenticated relay host defined via ENV config:
-  if [[ -n ${RELAY_USER} ]] && [[ -n ${RELAY_PASSWORD} ]]; then
-    echo "$(_env_relay_host)    ${RELAY_USER}:${RELAY_PASSWORD}" >> /etc/postfix/sasl_passwd
+  # Support authentication to a primary relayhost (when configured with credentials via ENV):
+  if [[ -n ${DEFAULT_RELAY_HOST} || -n ${RELAY_HOST} ]] \
+  && [[ -n ${RELAY_USER} && -n ${RELAY_PASSWORD} ]]; then
+    echo "${DEFAULT_RELAY_HOST:-$(_env_relay_host)}    ${RELAY_USER}:${RELAY_PASSWORD}" >> /etc/postfix/sasl_passwd
```
georglauterbach, line 91: Picky again:

```suggestion
    echo "${DEFAULT_RELAY_HOST:-$(_env_relay_host)}    ${RELAY_USER}:${RELAY_PASSWORD}" >>/etc/postfix/sasl_passwd
```

target/scripts/helpers/relay.sh
```diff
@@ -147,6 +153,11 @@ function _multiple_relayhosts() {
 function _legacy_support() {
   local DATABASE_VHOST='/etc/postfix/vhost'
 
+  # Only relevant when `RELAY_HOST` is configured:
+  if [[ -z ${RELAY_HOST} ]]; then
+    return 1
+  fi
```
georglauterbach, line 159:
```suggestion
  [[ -z ${RELAY_HOST} ]] && return 1
```

target/scripts/helpers/relay.sh
```diff
@@ -179,24 +183,24 @@ function _setup_relayhost() {
...
+function _process_relayhost_configs() {
+  local DATABASE_RELAYHOSTS='/tmp/docker-mailserver/postfix-relaymap.cf'
 
-    _relayhost_configure_postfix
+  # One of these must configure a relayhost for the feature to relevant:
+  if [[ ! -f ${DATABASE_RELAYHOSTS} ]] \
+  && [[ -z ${DEFAULT_RELAY_HOST} ]] \
+  && [[ -z ${RELAY_HOST} ]]; then
+    return 1
   fi
```
georglauterbach, line 198:
```suggestion
  if [[ ! -f ${DATABASE_RELAYHOSTS} ]] && [[ -z ${DEFAULT_RELAY_HOST} ]] && [[ -z ${RELAY_HOST} ]]; then
    return 1
  fi
```

I will make sure my setup works with these "long lines" 😆

georglauterbach (changes requested):
Overall LGTM 👍🏼 Much cleaner to read now! I have some nitpicks in the scripts for you to address still :)

target/scripts/helpers/postfix.sh
polarathene: > While `FIRST_FIELD` is absolutely correct, IMO it'd be more useful to state what the first field actually is and use this as a name.

I think it's clear that we're taking the first field and don't care too much about being more descriptive about that as we clearly process that field directly after into the variable(s) `UNAME` / `DOMAIN`?

Before it was `LOGIN` which isn't that much more meaningful. There is a comment directly above the file name check (for the `postfix-accounts.cf` file, that states it's extracting domains from mail accounts / addresses. We know that the expectation is that the first field has a domain-part to extract.

target/scripts/helpers/relay.sh
polarathene: I assume it's fine to remove, I haven't checked but it's likely this was the original grep args before `relay.sh` was introduced.

target/scripts/helpers/relay.sh
polarathene: Personally the vertical arrangement is quicker to parse visually (minor), but it also fairs better for `git blame` history (also minor value in this context).

I can go with the single line if you want, did you want that to collapse the suggestion to a single line entirely with `&& return 1` like was suggested elsewhere?

polarathene:
It appears we do have some useful test coverage? 🤔 

One of your change suggestions caused a test to fail.

```
not ok 30 [Relay Host] custom mapping is added from file in 129ms
# (from function `assert_output' in file test/test_helper/bats-assert/src/assert_output.bash, line 178,
#  in test file test/tests/serial/mail_with_relays.bats, line 61)
#   `assert_output --regexp '^@domaintwo.tld[[:space:]]+\[other.relay.com\]:587$'' failed
#
# -- regular expression does not match output --
# regexp (1 lines):
#   ^@domaintwo.tld[[:space:]]+\[other.relay.com\]:587$
# output (2 lines):
#   @domaintwo.tld		[other.relay.com]:587
#   @domaintwo.tld    [default.relay.com]:2525
# --
#
```

target/scripts/helpers/relay.sh
georglauterbach: Have you checked that this is not the culprit of the current test failures? When looking over the diff of the last commit, this stood out to me.

georglauterbach:
I'd start by reverting the regex change: `\s+` -> `\b` to see what happens with the test.

target/scripts/helpers/relay.sh
```diff
@@ -83,90 +85,92 @@ function _relayhost_sasl() {
...
+  # (by adding an entry in `/etc/postfix/relayhost_map`) provided it:
+  # - `/etc/postfix/relayhost_map` doesn't already have it as an existing entry.
+  # - `postfix-relaymap.cf` has no explicit opt-out (SENDER_DOMAIN key exists, but with no relayhost value assigned)
+  #
+  # NOTE: /etc/postfix/vhost represents managed mail domains sourced from `postfix-accounts.cf` and `postfix-virtual.cf`.
+  while read -r SENDER_DOMAIN; do
+    local MATCH_EXISTING_ENTRY="^@${SENDER_DOMAIN}\s+"
+    local MATCH_OPT_OUT_LINE="^\s*@${SENDER_DOMAIN}\s*$"
+
+    if ! grep -q -e "${MATCH_EXISTING_ENTRY}" /etc/postfix/relayhost_map && ! grep -qs -e "${MATCH_OPT_OUT_LINE}" "${DATABASE_RELAYHOSTS}"; then
```
polarathene, line 169: Good spotting @georglauterbach 👍 this is still the correct change, it was just the grep default regex syntax isn't compatible, either [requires escaping (`\+`) or using `-E`](https://linuxize.com/post/regular-expressions-in-grep/#grep-regular-expression).

```suggestion
    # NOTE: `-E` is required for `\s+` syntax to avoid escaping `+`
    if ! grep -q -E "${MATCH_EXISTING_ENTRY}" /etc/postfix/relayhost_map && ! grep -qs "${MATCH_OPT_OUT_LINE}" "${DATABASE_RELAYHOSTS}"; then
```

target/scripts/helpers/relay.sh
casperklein: > I will make sure my setup works with these "long lines" 😆

Yeah, these old (VT100) terminals with a max line length of 80 are a PITA nowadays 🤣 You should really consider an upgrade in the near future.

target/scripts/helpers/postfix.sh
georglauterbach: According to the comment, would `MAIL_ACCOUNT` make sense?

target/scripts/helpers/relay.sh
polarathene: > But I have no objection to the current way either.

I choose the `<100` length multi-line split condition then 😝

georglauterbach (approved):

PR #3852, opened 2024-01-29
fix: `packages.sh` - Download `jaq` via release `tag` not `latest`

polarathene:
# Description

As the filename includes the version / tag, we cannot rely on the latest URL to be stable.

This additionally bumps `jaq` from `v1.2.0` to [`v1.3.0`](https://github.com/01mf02/jaq/releases/tag/v1.3.0).

---

I performed a build locally with `make build` and while it didn't fail on the build, I was watching the log output and noticed a curl 404 line, which turned out to be:

```
7.203 [  TRACE  ]  Installing jaq
7.848 curl: (22) The requested URL returned error: 404
```

The build AFAIK failed for other reasons (Debian package repos being unavailable, which is common experience lately for me), so I'm not sure if the `jaq` curl failure above would normally be caught (_**EDIT:** Had a successful build, it wasn't caught as a failure_):

<details>
<summary>Debian package build log failure snippets</summary>

```
61.90 Ign:215 http://deb.debian.org/debian bookworm/main amd64 whois amd64 5.5.17
65.90 Err:177 http://deb.debian.org/debian bookworm/main amd64 libio-string-perl all 1.08-4
65.90   Could not connect to debian.map.fastlydns.net:80 (151.101.166.132), connection timed out Unable to connect to deb.debian.org:http: [IP: 151.101.166.132 80]

...

65.90 Err:215 http://deb.debian.org/debian bookworm/main amd64 whois amd64 5.5.17
65.90   Unable to connect to deb.debian.org:http: [IP: 151.101.166.132 80]
65.90 Fetched 56.9 MB in 47s (1207 kB/s)
65.90 E: Failed to fetch http://deb.debian.org/debian/pool/main/libi/libio-string-perl/libio-string-perl_1.08-4_all.deb  Could not connect to debian.map.fastlydns.net:80 (151.101.166.132), connection timed out Unable to connect to deb.debian.org:http: [IP: 151.101.166.132 80]

...

65.90 E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
------
Dockerfile:31
--------------------
  29 |     COPY target/scripts/helpers/log.sh /usr/local/bin/helpers/log.sh
  30 |
  31 | >>> RUN /bin/bash /build/packages.sh && rm -r /build
  32 |
  33 |     # -----------------------------------------------
--------------------
ERROR: failed to solve: process "/bin/bash -e -o pipefail -c /bin/bash /build/packages.sh && rm -r /build" did not complete successfully: exit code: 100
[...]

PR #3853, opened 2024-01-29
fix: abort when (jaq) curl fails

casperklein:
# Description

https://github.com/docker-mailserver/docker-mailserver/pull/3852#issuecomment-1915680328

## Type of change

## Checklist:


target/scripts/build/packages.sh
```diff
@@ -143,7 +144,7 @@ function _install_dovecot() {
   # when building for another architecture.
   if [[ ${DOVECOT_COMMUNITY_REPO} -eq 1 ]] && [[ "$(uname --machine)" == "x86_64" ]]; then
     _log 'trace' 'Using Dovecot community repository'
-    curl https://repo.dovecot.org/DOVECOT-REPO-GPG | gpg --import
+    curl -sSfL https://repo.dovecot.org/DOVECOT-REPO-GPG | gpg --import
```
polarathene, line 147: So a pipe operator won't cause the problem, only `&&` / `||`?

polarathene (approved):

target/scripts/build/packages.sh
casperklein: Correct.

Pipes are no problem as we use `set -o pipefail`. This setting prevents errors in a pipeline from being masked. If any command in a pipeline fails, that return code will be used as the return code of the whole pipeline. By default, the pipeline's return code is that of the last command even if it succeeds.

casperklein: PS: Using `&&` the way it was used, makes no sense (no benefit) when `set -e` is already in place.

georglauterbach (dismissed):

polarathene (approved):

Issue #1854, opened 2021-03-16
Default DKIM keylength of 4096 is too long


JaapD:
# Default DKIM keylength of 4096 is too long

The default DKIM is 4096, but according to https://tools.ietf.org/html/rfc6376 , Signers MUST use RSA keys of at least 1024 bits for long-lived keys.  Verifiers MUST be able to validate signatures with keys ranging from 512 bits to 2048 bits, and they MAY be able to validate signatures with larger keys.

The consequence of 4096 bit keys is that some mailservers deny the mail.

aendeavor:
Although this is very true, you are able to set the length of your DKIM keys manually, of course.

You have chosen to use the BUG issue template.This is more of a question and not a bug - I will mark it appropriately.

---

> The consequence of 4096 bit keys is that some mailservers deny the mail.

I do not deny that this does not happen, but DKIM failures are not automatically rejected AFAIK, only when the mail server does not understand the key length itself, right?

---

We as maintainers have chosen to use 4096 as the default, and we are likely going to stick with it. The reason being that we try to do a best-effort at using the forefront of security measures. This, of course, leads to situations where (especially old) mail servers will reject mail. But we believe this to be better than security vulnerabilities. In the end, as already mentioned, we will always make sure all of our users can fall back to more common settings (i.e. 2048 DKIM key size, intermediate TLS level, etc.).  

Unless @wernerfred or @casperklein or @polarathene think decreasing the default, I would stick with the current default.

casperklein:
From my personal TLS experience (Webserver):

1024 is good and not broken yet.
2048 is better/recommended, but generates slightly more CPU load.
4096 is obviously the "best", but generates significant more CPU load. Compared to the increased CPU load, one must decided if it's worth it. Keeping in mind, that 1024 is still unbroken, 4096 is probably overkill in most situations.

I don't see this as a big problem, as everybody can choose the key-length they want, while generating the keys.

The only thing we could discuss is the aspect, what we want to provide:

1. Max security --> sticking with 4096 as default
2. Max compatibility, but still security in mind --> change to 2048 as default.

wernerfred:
Imho you can't make a right decision. If we choose 2048 for default we will receive Issues asking why we support 4096 but don't set it as default and vice versa. 

> I don't see this as a big problem, as everybody can choose the key-length they want, while generating the keys.

[...]

PR #3875, opened 2024-02-05
docs: fix 404 in mailserver.env and default to RSA 2048 for TLS certs

rahilarious:
# Description
title

## Type of change

### DIFF
--- a/mailserver.env
+++ b/mailserver.env
@@ -362,7 +362,7 @@ POSTFIX_REJECT_UNKNOWN_CLIENT_HOSTNAME=0
 POSTFIX_INET_PROTOCOLS=all
 
 # Enables MTA-STS support for outbound mail.
-# More details: https://docker-mailserver.github.io/docker-mailserver/latest/config/advanced/mail-mta-sts/
+# More details: https://docker-mailserver.github.io/docker-mailserver/v13.3/config/best-practices/mta-sts/
 # - **0** ==> MTA-STS disabled
 # - 1 => MTA-STS enabled
 ENABLE_MTA_STS=0
--- a/docs/content/config/security/ssl.md
+++ b/docs/content/config/security/ssl.md
@@ -408,7 +408,7 @@ The following example is the [basic setup][acme-companion::basic-setup] you need
 
     - `LETSENCRYPT_TEST=true`: _Recommended during initial setup_. Otherwise the default production endpoint has a [rate limit of 5 duplicate certificates per week][letsencrypt::limits]. Overrides `ACME_CA_URI` to use the _Let's Encrypt_ staging endpoint.
     - `LETSENCRYPT_EMAIL`: For when you don't use `DEFAULT_EMAIL` on `acme-companion`, or want to assign a different email contact for this container.
-    - `LETSENCRYPT_KEYSIZE`: Allows you to configure the type (RSA or ECDSA) and size of the private key for your certificate. Default is RSA 4096.
+    - `LETSENCRYPT_KEYSIZE`: Allows you to configure the type (RSA or ECDSA) and size of the private key for your certificate. Default is RSA 4096, but RSA 2048 is recommended.
     - `LETSENCRYPT_RESTART_CONTAINER=true`: When the certificate is renewed, the entire container will be restarted to ensure the new certificate is used.
 
     [`acme-companion` ENV for default settings][acme-companion::env-config] that apply to all containers using `LETSENCRYPT_HOST`:
@@ -450,8 +450,8 @@ The following example is the [basic setup][acme-companion::basic-setup] you need
     # Optional variables:
     LETSENCRYPT_mail_TEST=true
     LETSENCRYPT_mail_EMAIL='admin@example.com'
-    # RSA-4096 => `4096`, ECDSA-256 => `ec-256`:
-    LETSENCRYPT_mail_KEYSIZE=4096
+    # Supported values are `2048`, `3072` and `4096` for RSA keys, and `ec-256` or `ec-384` for elliptic curve keys.
+    LETSENCRYPT_mail_KEYSIZE=2048
     ```
 
     Unlike with the equivalent ENV for containers, [changes to this file will **not** be detected automatically][acme-companion::standalone-changes]. You would need to wait until the next renewal check by `acme-companion` (_every hour by default_), restart `acme-companion`, or [manually invoke the _service loop_][acme-companion::service-loop]:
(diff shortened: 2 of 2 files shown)


mailserver.env
```diff
@@ -362,7 +362,7 @@ POSTFIX_REJECT_UNKNOWN_CLIENT_HOSTNAME=0
 POSTFIX_INET_PROTOCOLS=all
 
 # Enables MTA-STS support for outbound mail.
-# More details: https://docker-mailserver.github.io/docker-mailserver/latest/config/advanced/mail-mta-sts/
+# More details: https://docker-mailserver.github.io/docker-mailserver/latest/config/best-practices/mta-sts/
```
polarathene, line 365:
```suggestion
# More details: https://docker-mailserver.github.io/docker-mailserver/v13.3/config/best-practices/mta-sts/
```

polarathene (changes requested):
Thanks for spotting the wrong MTA-STS docs link, it should use the version tag (optional) for better reliability?

---

The other changes from RSA 2048-bit to 4096-bit are not needed and can be reverted.

If you disagree, please justify why 110-bit of symmetric equivalent security strength is insufficient, and why you need roughly 40 bits more?

- Math: https://gist.github.com/polarathene/9e3cd791091dc2bd075a2ec4cf793aaa
- Reasoning: https://github.com/docker-mailserver/docker-mailserver/issues/1854#issuecomment-806280929
  - You need enough energy to boil all the oceans in the world to pull this off...
  - You can have all the hardware compute power for crypto-mining currently and it's still not happening.
  - Typically when you renew a certificate such as with LetsEncrypt as the CA, it's common to also change the private key in that process, that's a <90 day window, good luck 🤷‍♂️ (_as if the financial cost without the time constraint wasn't a large enough barrier alone_)

I have looked into this quite seriously in the past as you can see. If you have anything to explain why you think 4096-bit is adding any actual value, I'm always open to new information 👍
- Often though this is just security paranoia / parroting without understanding the topic well
- Some also wrongly assume that the difference between 2048-bit and 4096-bit RSA is only 2x, when it's actually more like 1 trillion times as much (_roughly 40-bit symmetric delta_). You don't need that, switch to ECC if you want more entropy.

mailserver.env
rahilarious: done

polarathene:
> I also agree RSA 2048 is fine, I just wanted to have consistent keysize on ssl.md page.

👍 

Apologies about verbosity below. You may be aware of a bunch of that, but I like being thorough 😎 

---

> I'm bit confused here, when using MTA-STS (encrypted email using TLS)

MTA-STS isn't affecting email directly, it's only to remember compatible MTAs DMS sent mail to previously that used TLS to know that something is fishy if it can't use TLS (via STARTTLS) in future connections to that MTA. It's more of a bandaid over the fact that it's kind of expected (via an RFC) that mail sent over port 25 is permitted to be sent without encryption for compatibility / delivery reasons IIRC.

You only get that guarantee between the direct connection between DMS and that MTA, there may me multiple hops after that which you cannot ensure the traffic is encrypted, it's not quite the same as HTTPS due to how mail works. Although you could perhaps liken it to a reverse proxy that terminates TLS for HTTPS, then connects to another service unencrypted over HTTP, usually on the same system.

---

> which key pair is used for encryption, env SSL_CERT_PATH? or DKIM key pair?

DKIM isn't about encryption, just verifying that your mail content wasn't modified in transit, since as mentioned above it could go through additional hops insecurely where someone tampers with it. This was a problem in some countries due to governments and also malicious use with airports/cafes IIRC.

If DKIM was compromised, it has no impact on past mail traffic since it's already been delivered untampered. The attack only benefits with mail signed that is sent going forward.

You provide a certificate for TLS that DMS will use to help secure the connection, it's primary role however is verifying trust. This differs a bit from the DKIM keypair as there is additional metadata beyond private/public keys, the public key is part of the [X.509 certificate](https://en.wikipedia.org/wiki/X.509#Certificates).

---

Your confusion with my link is understandable. The title implies it's about DKIM but my comment there is about [public-key cryptography](https://en.wikipedia.org/wiki/Public-key_cryptography) in general, but regarding RSA key size strength.

For TLS this is only really relevant for verifying authenticity against a third-party (CA), so that the client knows the server is who they claim to be, establishing trust that negotiating a secure connection is to the correct party. At least when using a cipher with [perfect forward secrecy](https://en.wikipedia.org/wiki/Transport_Layer_Security#Forward_secrecy) (PFS), that should be the only purpose for the RSA cert.

Provided PFS is in use, then the client and server exchange a bit of information to derive a session key for securing that connection with symmetric encryption (_eg: 128-bit or 256-bit AES-GCM_), this is done with fancy math that prevents someone listening to that earlier client/server exchange from having enough information to create the session key (_since they're missing the secret part unique to client and server, see [Diffie-Helman exchange](https://en.wikipedia.org/wiki/Diffie%E2%80%93Hellman_key_exchange)_).

An attacker would need to record that encrypted traffic, and then crack the session key which is roughly equivalent to those RSA key size metrics I referenced earlier (little bit stronger IIRC). We use [standardized DHE params](https://security.stackexchange.com/questions/149811/why-is-mozilla-recommending-predefined-dhe-groups), [presently 4096-bit](https://github.com/docker-mailserver/docker-mailserver/tree/d5efaf95c3a9239fa5ad3728be1525f505cc38b9/target/shared) (_although I've been open to lowering it to 2048-bit / 3072-bit_). AFAIK, since these are params for DHE used widely, it'd be similar in impact to attacking a CA like LetsEncrypt (_which for RSA uses 4096-bit IIRC_), these are also used by TLS 1.3 and 2048-bit will be preferred I believe.

However... most connections should be using ECDHE instead (_compatible with RSA cert_), where EC-256 is equivalent to 128-bit symmetric strength, thus DHE is less attractive. ECDHE is more difficult (_remember that we're talking about absolutely massive cost to attack 128-bit, over 65k times as much my previous statements...that's a lot of ocean boiling!_)

Since the session key and symmetric encryption are at least 128-bit these days, there's obviously more value in compromising the DHE/ECDHE parameters for a wider breach. At the limits of my understanding with the security aspects and impact at this point though 😅 

Some security standards like by NIST will advise higher than I do, but that's usually in their interest to do so, and somewhat caters to a specific audience that likes bigger numbers / comfort over understanding the math. For the most part, what I advise is sane and secure; there is more affordable alternatives to getting information from a target than these attacks.

polarathene (approved):

rahilarious:
> At the limits of my understanding with the security aspects and impact at this point though 😅

hahahah. You're funny. 

I LOVE your verbosity. Never apologize for it. Genuinely appreciate the efforts you put in well-thought out deep explanation. Thank you for kindness.

I'm more curious about you. How can I know more?

polarathene:
> I'm more curious about you. How can I know more?

You can reach me via [LinkedIn](https://nz.linkedin.com/in/brennan-kinney-a5454953)👍

PR #3880, opened 2024-02-10
fix: `postfix-main.cf` may depend upon `postfix-master.cf`

polarathene:
# Description

Custom parameters must be referenced to be retained when `postconf -n` is run.

If `postfix-main.cf` defines custom parameters that are referenced by `postfix-master.cf`, that needs to first update `master.cf` before updating `main.cf` (_where `postconf -n` is relied upon_).

---

Failure was a little tricky to troubleshoot as it only occurred during initial startup, reproducing the logic afterwards manually, or via container restart (with internal state persisted) worked fine as `master.cf` was already configured with the custom parameters.

Due to `postconf -n` outputting any errors to `/dev/null`, this sort of information is omitted from the logs:

```
postconf: warning: /etc/postfix/main.cf: unused parameter: some_custom_parameter=value
```

Discovered while answering: https://github.com/orgs/docker-mailserver/discussions/3847

## Type of change

## Checklist


polarathene:
Should we add a test case for this? Or is the inline comment about order dependency sufficient?

georglauterbach:
I think having a test for this is a good idea :)

target/scripts/startup/setup.d/postfix.sh
```diff
@@ -109,29 +109,30 @@ function _setup_postfix_late() {
...
+
+  # NOTE: Must come after postfix-master.cf as custom parameter support only works provided either config references it.
   local OVERRIDE_CONFIG_POSTFIX_MAIN='/tmp/docker-mailserver/postfix-main.cf'
   if [[ -f ${OVERRIDE_CONFIG_POSTFIX_MAIN} ]]; then
     cat "${OVERRIDE_CONFIG_POSTFIX_MAIN}" >>/etc/postfix/main.cf
     _adjust_mtime_for_postfix_maincf
 
-    # do not directly output to 'main.cf' as this causes a read-write-conflict
+    # Do not directly output to 'main.cf' as this causes a read-write-conflict
     postconf -n >/tmp/postfix-main-new.cf 2>/dev/null
```
casperklein, line 129: > Due to postconf -n outputting any errors to /dev/null, this sort of information is omitted from the logs

As long as annoying errors are not expected, `2>/dev/null` should be removed:

```suggestion
    postconf -n >/tmp/postfix-main-new.cf
```

polarathene: > As long as annoying errors are not expected

What do you consider annoying?

I gave an example log line (warning), you get those as well for any duplicate parameters configured IIRC (_due to `postfix-main.cf` appending overrides_).

I can commit your suggestion if you think the added noise is worthwhile? (_[here is why we originally did this](https://github.com/docker-mailserver/docker-mailserver/pull/2598#issuecomment-1140813123)_)

casperklein: Probably I misunderstood something.

By moving the postfix master part at the top, the mentioned error should be fixed? Therefore `postconf -n` should run fine without errors. If that's the case, stderr should not be redirected to dev/null.

polarathene: > By moving the postfix master part at the top, the mentioned error should be fixed?

The log line example shown in this PR would be avoided if the setting is in `master.cf` yes.



> Therefore `postconf -n` should run fine without errors. If that's the case, stderr should not be redirected to dev/null.

See the link to the original reason we added `2>/dev/null` to avoid log spam due to a different condition. I assume that behaviour can still be reproduced and don't see how it would be desirable to show it.

target/scripts/startup/setup.d/postfix.sh
```diff
@@ -109,29 +109,33 @@ function _setup_postfix_late() {
...
   if [[ -f ${OVERRIDE_CONFIG_POSTFIX_MAIN} ]]; then
     cat "${OVERRIDE_CONFIG_POSTFIX_MAIN}" >>/etc/postfix/main.cf
     _adjust_mtime_for_postfix_maincf
 
-    # do not directly output to 'main.cf' as this causes a read-write-conflict
-    postconf -n >/tmp/postfix-main-new.cf 2>/dev/null
+    # Do not directly output to 'main.cf' as this causes a read-write-conflict.
+    # `postconf` output is filtered to skip expected warnings regarding overrides:
+    # https://github.com/docker-mailserver/docker-mailserver/pull/3880#discussion_r1510414576
+    postconf -n >/tmp/postfix-main-new.cf 2> >(grep -v 'overriding earlier entry')
```
polarathene, line 132: This is the main change I added since the last round of reviews.

It should allow for retaining other output from `postconf` that may be of interest when something unexpected happens, as opposed to the more blunt `2>/dev/null` approach we have 👍 

You're both more comfortable with shell syntax than I am, so if you have any feedback here, happy to go with that instead 😎

test/config/override-configs/postfix-main.cf
```diff
@@ -2,3 +2,6 @@ max_idle = 600s
 # this is a comment
     # this is also a comment
 readme_directory = /tmp
+
+# This parameter is referenced by an override in `postfix-master.cf`:
+custom_parameter = cidr:{{!172.16.0.42 REJECT}}, permit_sasl_authenticated, reject
```
polarathene, line 7: The test adds this `postfix-main.cf` custom parameter, which `postfix-master.cf` uses for `smtpd_client_restrictions` value. This is the same as what the referenced discussion was using, it just prepends that `cidr` inline table to reject any IP that is not `172.16.0.42`.

With the `postfix.sh` change to swap the order we apply `postfix-main.cf` and `postfix-master.cf`, the test demonstrates that we can correctly resolve/expand the `$custom_parameter` value in the updated `/etc/postfix/master.cf` file 👍

test/tests/parallel/set1/config_overrides.bats
```diff
@@ -15,6 +15,9 @@ function setup_file() {
 
 function teardown_file() { _default_teardown ; }
 
+# The `postconf` command can query both `main.cf` and `master.cf` at `/etc/postfix/`.
+# Reference: http://www.postfix.org/postconf.1.html
```
polarathene, line 19: Maintainer docs reference for this test file, since there's a few different options used with the `postconf` command in the test cases below, probably helpful to have here 😅

polarathene (commented):
Sorry for the long delay on getting back to this!

Test case added. I've done some more revisions, but the new comments as part of this review is all you should need AFAIK. Feel free to look over the test case itself, it should be easy to grok👍 

---

@georglauterbach ping for review, since I can't renew an existing request 😓

target/scripts/startup/setup.d/postfix.sh
georglauterbach: > You're both more comfortable with shell syntax than I am, so if you have any feedback here, happy to go with that instead 😎

No, looks all good to me!

georglauterbach (approved):
LGTM 👍🏼 Nice work!

casperklein (approved):

polarathene:
Cheers for the reviews/approvals! 😁

PR #3882, opened 2024-02-11
docs: Complete rewrite of PROXY protocol guide

polarathene:
# Description

Based on the rather long discussion at https://github.com/docker-mailserver/docker-mailserver/issues/3866

The existing docs are a bit dated:
- Postfix has support for PROXY protocol v2 now (Since Postfix 3.5).
- Incorrect information at the end regarding an IMAPS port conflicting with SMTP related services `amavis` / `postscreen`.
- Dovecot override config was a bit lacking (_earlier services, no need for timeout, trusted network setting using invalid syntax `,`_)

**Added:**
- More technical insights and caveats for those it'd benefit.
- Includes a security section about forgery risk and monitoring gotchas (Roundcube + Fail2Ban) with advice to resolve.
- Small section on verifying working configuration.
- Dovecot + Postfix config for enabling PROXY protocol support + warning and alternative config advice for when separate ports may be needed (_someone may contribute that as a feature to DMS v15+_)

## Type of change

## Checklist

Can bundle in changelog entry referencing various docs PRs 👍

### DIFF
--- a/docs/content/examples/tutorials/mailserver-behind-proxy.md
+++ b/docs/content/examples/tutorials/mailserver-behind-proxy.md
@@ -2,126 +2,410 @@
 title: 'Tutorials | Mail Server behind a Proxy'
 ---
 
-## Using DMS behind a Proxy
+## Using a Reverse Proxy
 
-### Information
+Guidance is provided via a Traefik config example, however if you're only familiar with configuring a reverse proxy for web services there are some differences to keep in mind.
 
-If you are hiding your container behind a proxy service you might have discovered that the proxied requests from now on contain the proxy IP as the request origin. Whilst this behavior is technical correct it produces certain problems on the containers behind the proxy as they cannot distinguish the real origin of the requests anymore.
+- A security concern where preserving the client IP is important but needs to be handled at Layer 4 (TCP).
+- TLS will be handled differently due protocols like STARTTLS and the need to comply with standards for interoperability with other MTAs.
+- The ability to route the same port to different containers by FQDN can be limited.
 
-To solve this problem on TCP connections we can make use of the [proxy protocol](https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt). Compared to other workarounds that exist (`X-Forwarded-For` which only works for HTTP requests or `Tproxy` that requires you to recompile your kernel) the proxy protocol:
+This reduces many of the benefits for why you might use a reverse proxy, but they can still be useful.
 
-- It is protocol agnostic (can work with any layer 7 protocols, even when encrypted).
-- It does not require any infrastructure changes.
-- NAT-ing firewalls have no impact it.
-- It is scalable.
+Some deployments may require a service to route traffic (kubernetes) when deploying, in which case the below advice is important to understand well.
 
-There is only one condition: **both endpoints** of the connection MUST be compatible with proxy protocol.
+## What can go wrong?
 
-Luckily `dovecot` and `postfix` are both Proxy-Protocol ready softwares so it depends only on your used reverse-proxy / loadbalancer.
+Without a reverse proxy involved, a service is typically aware of the client IP for a connection.
 
-### Configuration of the used Proxy Software
+However when a reverse proxy routes the connection this information can be lost, and the proxied service mistakenly treats the client IP as the reverse proxy handling the connection.
 
-The configuration depends on the used proxy system. I will provide the configuration examples of [traefik v2](https://traefik.io/) using IMAP and SMTP with implicit TLS.
+- That can be problematic when the client IP is meaningful information for the proxied service to act upon, especially when it [impacts security](#security-concerns).
+- The [PROXY protocol][networking::spec:proxy-protocol] is a well established solution to preserve the client IP when both the proxy and service have enabled the support.
 
-Feel free to add your configuration if you achieved the same goal using different proxy software below:
+??? abstract "Technical Details - HTTP vs TCP proxying"
 
-??? "Traefik v2"
+    A key difference for how the network is proxied relates to the [OSI Model][networking::osi-model]:
 
-    Truncated configuration of traefik itself:
+    - Layer 7 (_Application layer protocols: SMTP / IMAP / HTTP / etc_)
+    - Layer 4 (_Transport layer protocols: TCP / UDP_)
 
-    ```yaml
+    When working with Layer 7 and a protocol like HTTP, it is possible to inspect a protocol header like [`Forwarded`][networking::http-header::forwarded] (_or it's predecessor: [`X-Forwarded-For`][networking::http-header::x-forwarded-for]_). At a lower level with Layer 4, that information is not available and we are routing traffic agnostic to the application protocol being proxied.
+
+    A proxy can prepend the [PROXY protocol][networking::spec:proxy-protocol] header to the TCP/UDP connection as it is routed to the service, which must be configured to be compatible with PROXY protocol (_often this adds a restriction that connections must provide the header, otherwise they're rejected_).
+
+    Beyond your own proxy, traffic may be routed in the network by other means that would also rewrite this information such as Docker's own network management via `iptables` and `userland-proxy` (NAT). The PROXY header ensures the original source and destination IP addresses, along with their ports is preserved across transit.
+
+## Configuration
+
+### Reverse Proxy
+
+The below guidance is focused on configuring [Traefik][traefik-web], but the advice should be roughly applicable elsewhere (_eg: [NGINX][nginx-docs::proxyprotocol], [Caddy][caddy::plugin::l4]_).
+
+- Support requires the capability to proxy TCP (Layer 4) connections with PROXY protocol enabled for the upstream (DMS). The upstream must also support enabling PROXY protocol (_which for DMS services rejects any connection not using the protocol_).
+- TLS should not be terminated at the proxy, that should be delegated to DMS (_which should be configured with the TLS certs_). Reasoning is covered under the [ports section](#ports).
+
+???+ example "Traefik service"
+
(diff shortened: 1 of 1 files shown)


polarathene:
Direct links for comparison:
- https://docker-mailserver.github.io/docker-mailserver/v13.3/examples/tutorials/mailserver-behind-proxy/
- https://pullrequest-3882--dms-doc-previews.netlify.app/examples/tutorials/mailserver-behind-proxy/

casperklein (approved):

georglauterbach (approved):

docs/content/examples/tutorials/mailserver-behind-proxy.md
```diff
@@ -2,126 +2,410 @@
...
 
-    Truncated configuration of traefik itself:
+    - Layer 7 (_Application layer protocols: SMTP / IMAP / HTTP / etc_)
+    - Layer 4 (_Transport layer protocols: TCP / UDP_)
 
-    ```yaml
+    When working with Layer 7 and a protocol like HTTP, it is possible to inspect a protocol header like [`Forwarded`][networking::http-header::forwarded] (_or it's predecessor: [`X-Forwarded-For`][networking::http-header::x-forwarded-for]_). At a lower level with Layer 4, that information is not available and we are routing traffic agnostic to the application protocol being proxied.
+
+    A proxy can prepend the [PROXY protocol][networking::spec:proxy-protocol] header to the TCP/UDP connection as it is routed to the service, which must be configured to be compatible with PROXY protocol (_often this adds a restriction that connections must provide the header, otherwise they're rejected_).
+
```
cfis, line 36: I would remove "(_often this adds a restriction that connections must provide the header, otherwise they're rejected_)." I don't think its often, I think its always. But that is already captured by the sentence above.

docs/content/examples/tutorials/mailserver-behind-proxy.md
```diff
@@ -2,126 +2,410 @@
...
+    ---
+
+    Postfix config now requires [our `user-patches.sh` support][docs::overrides::user-patches] to add new services in `/etc/postfix/master.cf`:
+
+    ```bash title="docker-data/dms/config/user-patches.sh"
+    #!/bin/bash
+
+    # Duplicate the config for the submission(s) service ports (587 / 465) with adjustments for the PROXY ports (10587 / 10465) and `syslog_name` setting:
+    postconf -Mf submission/inet | sed -e s/^submission/10587/ -e 's/submission/submission-proxyprotocol/' >> /etc/postfix/master.cf
```
cfis, line 248: I assume you run this inside the container? Wouldn't it be lost when the container is restarted?

cfis:
Added  a few comments, but overall looks godo to me!

docs/content/examples/tutorials/mailserver-behind-proxy.md
polarathene: > I don't think its often, I think its always.

What source do you have to say always? My statement is valid, here is why it varies:

- [Traefik entrypoints support the PROXY header as optional](https://doc.traefik.io/traefik/routing/entrypoints/#proxyprotocol). I'm able to enable it and connect without PROXY protocol in use from curl.
- Whereas the [equivalent for Caddy](https://caddyserver.com/docs/json/apps/http/servers/listener_wrappers/proxy_protocol/) will allow normal connections without PROXY protocol, unless they're in the `allow` list in which case they're expected to only use PROXY protocol.
  - I can allow Traefik to connect with PROXY protocol, and in a separate container route through Traefik or alternatively connect to Caddy directly without PROXY protocol. So long as I didn't allow PROXY protocol for the curl container by configuring the entire subnet, that direct connection via curl works fine.
- Postfix and Dovecot however only accept connections using PROXY protocol for ports that enable it. They're not as flexible.

docs/content/examples/tutorials/mailserver-behind-proxy.md
polarathene: > I assume you run this inside the container?

Yes, as the docs mention prior, this is `user-patches.sh`. I've annotated the codefence and provided a link reference to the relevant docs page, so the reader should already have this context.

image

> Wouldn't it be lost when the container is restarted?

It would be lost when the container is recreated, which is what we advise for "restarting" (destroy the previous instance).

Otherwise it'd be appended each time since the container persists internal state regardless of a volume mount when it's restarted (_or a more manual stop/start_). Which is pretty bad, we have several issues open about such bugs.

Ideally we'd have our internal base config stored elsewhere to be copied to the proper locations at each restart so that we can avoid that issue. It'll be a breaking change though as some users volume mount configs directly to the internal location and some of those are files we may modify, addressing the issue may wipe their direct mounted volume content (_one would hope they have a backup, or at least read breaking changes in our changelog_).

polarathene:
> Added a few comments, but overall looks good to me! Thanks for writing this up.

Thanks for the feedback! 😁

I know it's quite a bit to go over, so I really appreciate it! Writing docs is an exhausting process for me to tackle 😩 

I think my replies have addressed the feedback you gave, but if you feel I could improve on the content or disagree in anyway let me know :)

docs/content/examples/tutorials/mailserver-behind-proxy.md
cfis: I base my comment on this paragraph from the spec (https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt):

> The receiver MUST be configured to only receive the protocol described in this
> specification and MUST not try to guess whether the protocol header is present
> or not. This means that the protocol explicitly prevents port sharing between
> public and private access. Otherwise it would open a major security breach by
> allowing untrusted parties to spoof their connection addresses. The receiver
> SHOULD ensure proper access filtering so that only trusted proxies are allowed
> to use this protocol.

docs/content/examples/tutorials/mailserver-behind-proxy.md
cfis: Oh I see, missed it was part of user-patches.sh

docs/content/examples/tutorials/mailserver-behind-proxy.md
polarathene: > This means that the protocol explicitly prevents port sharing between public and private access.
> Otherwise it would open a major security breach by allowing untrusted parties to spoof their connection addresses.
> The receiver SHOULD ensure proper access filtering so that only trusted proxies are allowed to use this protocol.

The `SHOULD` making it advised rather than mandatory resolves the issue though?

If the receiver only allows a connection from a trusted host, you could require the PROXY protocol header only on those connections. But what is the risk in making the header optional when you're trusting that host regardless?

Likewise, if you reject a connection from a host providing a PROXY header when they're not trusted, what is the risk in allowing them to connect without the PROXY header present?

---

The upcoming Caddy 2.8 release changed their PROXY protocol dependency to the same that Traefik is using. Thus they'll also have the capability for ports to accept an optional PROXY header that's accepted when the remote host is trusted by the reverse proxy software.

The Go library used is [`go-proxyproto`](https://github.com/pires/go-proxyproto), and this flexibility is offered via the [Policy](https://github.com/pires/go-proxyproto/blob/v0.7.0/policy.go#L18) feature (_landed [roughly a year ago](https://github.com/pires/go-proxyproto/issues/95)_). You can see the [implementation for Caddy here](https://github.com/caddyserver/caddy/pull/5915/files).

There wasn't much discussion about security concerns or the spec you're quoting, but there doesn't seem to be any major concerns with that flexibility via policy since it still implements the `SHOULD` to ensure PROXY protocol is only valid with trusted connections.

---

Regardless, the spec doesn't reflect actual popular software support in the wild which our docs are noting that the restriction is not always imposed.

So agree to leave as-is?

polarathene:
There is an AWS specific gotcha with their load balancer with PROXY protocol that affects SMTP, not sure if it's relevant for our docs though. I've not got the time to verify if it's still accurate.

https://github.com/pires/go-proxyproto#special-notes

It was added in June 2020: https://github.com/pires/go-proxyproto/pull/25

docs/content/examples/tutorials/mailserver-behind-proxy.md
cfis: I think the first sentence is pretty clear:

> The receiver MUST be configured to only receive the protocol described in this
> specification and MUST not try to guess whether the protocol header is present
> or not.

I read the SHOULD as a secondary concern, it does not impact that first sentence in anyway. Thus, I understand why Postfix and Dovecot have implemented proxy support the way they have.

I think the Caddy code you posted is clearly in violation of the spec (not that it matters what I think of course). I would guess it was implemented by someone who didn't read the spec and likely didn't consider the security implications.

polarathene: > I would guess it was implemented by someone who didn't read the spec and likely didn't consider the security implications.

What security implications?

If you only follow the MUST requirements of the spec and ignore the SHOULD, you have less security 🤷‍♂️ 

I see no security concerns with the policy approach that `go-proxyproto` implements.

---

> Either leave as is, or maybe add text "according to the spec the port should always ... but some implementation allow ... "

Leaving as-is 👍 

The docs are focused on PROXY protocol compatibility with different reverse proxies / software relevant to DMS. I don't see any value in the expectations of the spec when they're not well justified (_specifications can have flaws or become outdated too_).

I have [raised an issue at HAProxy](https://github.com/haproxy/haproxy/issues/2450), perhaps they'll update it accordingly or weigh in with something that justifies the current wording.

cfis: The spec has been updated 11 times in ten years...guessing its not a flaw in the spec.  Excellent that you raised an issue, hope it gets a response.

PR #3899, opened 2024-02-20
fix(`setup`): `open-dkim` log for conflicting implementations

### LINKED ISSUE #3896: bug report: log: command not found
### 📝 Preliminary Checks

### 👀 What Happened?

The command should not output an error and print me the help
Present since a bunch of versions of DMS.

```sh
/usr/local/bin/open-dkim: line 8: log: command not found
```

### 👟 Reproduction Steps

```sh
docker exec -it mailserver-1 bash
root@emails:/# setup config dkim
/usr/local/bin/open-dkim: line 8: log: command not found
```

### 🐋 DMS Version

13.3.1

Debian 12

_No response_

### 📜 Relevant log output

_No response_

_No response_

polarathene:
# Description

Fixes #3896

## Type of change

### DIFF
--- a/target/bin/open-dkim
+++ b/target/bin/open-dkim
@@ -5,7 +5,7 @@ source /usr/local/bin/helpers/index.sh
 
 if [[ -f /etc/dms-settings ]] && [[ $(_get_dms_env_value 'ENABLE_RSPAMD') -eq 1 ]]; then
   if [[ $(_get_dms_env_value 'ENABLE_OPENDKIM') -eq 1 ]]; then
-    log 'error' "You enabled Rspamd and OpenDKIM - OpenDKIM will be implicitly used for DKIM keys"
+    _log 'warn' "Conflicting DKIM support, both Rspamd and OpenDKIM enabled - OpenDKIM will manage DKIM keys"
   else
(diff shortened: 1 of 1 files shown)


georglauterbach (approved):

williamdes (approved):
Nice, this is more obvious that the command is continuing

I would suggest also adding an output of the public key at the end or a success log with the key path
what do you think?

polarathene:
> I would suggest also adding an output of the public key at the end or a success log with the key path
> what do you think?

The rspamd version supports this AFAIK.

Nobody is interested in contributing changes to these existing two scripts for DKIM AFAIK, it's better to replace them with the single keypair generation method [I've proposed using smallstep `step-cli`](https://github.com/docker-mailserver/docker-mailserver/issues/3630#issuecomment-1903401956). You'll find that supports the public key output too.

I'll tackle that for v15 release, I need to first wrap up a task unrelated to DMS before I can come return to DMS LDAP PR work (docs + refactor). The docs are blocking on v14 so hopefully I can sort that out by the weekend 😅 

As for key path, not that important. New approach will have a predictable location and storage convention for the DMS config volume. Those will be copied to an internal location in DMS at runtime with config generation handled.

PR #3901, opened 2024-02-20
fix: fetchmail environment variables

casperklein:
# Description

Executing `supervisor restart fetchmail` causes:

```
Feb 20 17:12:45 mail fetchmail[2543]: /root/.netrc: cannot open file for reading: Permission denied
```

Fetchmail checks the wrong location to see if `.netrc` exists. That's because $HOME is set to `/root`. fetchmail does not have permission to access `/root`.

http://supervisord.org/subprocess.html#subprocess-environment

> No shell is executed by supervisord when it runs a subprocess, so environment variables such as USER, PATH, HOME, SHELL, LOGNAME, etc. are not changed from their defaults or otherwise reassigned. This is particularly important to note when you are running a program from a supervisord run as root with a user= stanza in the configuration

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,7 @@ The most noteworthy change of this release is the update of the container's base
   - Restarting DMS should not be required when configuring relay hosts without these ENV, but solely via `setup relay ...`, as change detection events now apply relevant Postfix setting changes for supporting credentials too.
 - Rspamd configuration: Add a missing comma in `local_networks` so that all internal IP addresses are actually considered as internal ([#3862](https://github.com/docker-mailserver/docker-mailserver/pull/3862))
 - Ensure correct SELinux security context labels for files and directories moved to the mail-state volume during setup ([#3890](https://github.com/docker-mailserver/docker-mailserver/pull/3890))
+- Use correct environment variable for fetchmail ([#3901](https://github.com/docker-mailserver/docker-mailserver/pull/3901))
 
 ## [v13.3.1](https://github.com/docker-mailserver/docker-mailserver/releases/tag/v13.3.1)
 
--- a/target/supervisor/conf.d/supervisor-app.conf
+++ b/target/supervisor/conf.d/supervisor-app.conf
@@ -122,6 +122,7 @@ autorestart=true
 stdout_logfile=/var/log/supervisor/%(program_name)s.log
 stderr_logfile=/var/log/supervisor/%(program_name)s.log
 user=fetchmail
+environment=HOME="/var/lib/fetchmail",USER="fetchmail"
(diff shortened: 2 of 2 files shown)


casperklein:
@georglauterbach @polarathene Not related to this PR and just cosmetic: What is your opinion about renaming `target/supervisor/conf.d/supervisor-app.conf` to something like `DMS.conf` or similar?

georglauterbach:
> @georglauterbach @polarathene Not related to this PR and just cosmetic: What is your opinion about renaming `target/supervisor/conf.d/supervisor-app.conf` to something like `DMS.conf` or similar?

Not an issue if you ask; we should classify it as breaking, though.

georglauterbach (approved):

polarathene:
> What is your opinion about renaming `target/supervisor/conf.d/supervisor-app.conf` to something like `DMS.conf` or similar?

I don't see any value in doing so (especially as uppercase). I'd rather a motivation for it, as we do have a separate config for sasl login mechanisms too, you could probably split it into separate service groups if that makes more sense? 🤷‍♂️

casperklein:
I think "supervisor-app" is completely generic. The "app" (all services combined) we are using is called docker-mailserver (DMS). So it makes more sense to me, to name the config accordingly.

georglauterbach:
Having service groups is something I'd support as well. Maybe for anti-spam, then dovecot-specific, and then dms.conf for the rest?

casperklein:
Having it all in one place, like now, is fine for me. The file isn't that big or confusing that would justify a splitting IMO.
It is really just the name supervisor-app, that triggers me for some time now 😆 If you both want to keep that name, that's fine 👍

georglauterbach:
Nah, go ahead and change it to `dms.conf`, I like that one better too.

polarathene:
> The file isn't that big or confusing that would justify a splitting IMO.

I was thinking more in terms of feature isolation. I probably won't get around to it, but thought about exploring a file structure that groups scripts related to features/services together, just gets a bit messy when integration is involved. Probably not worth it.



> It is really just the name supervisor-app, that triggers me for some time now 😆 If you both want to keep that name, that's fine 👍

I don't mind either way. `dms-services.conf` might be more appropriate than just `dms.conf` if you do go forward with such a change. Might be best to schedule it for v15, as v14 breaking changes is already quite a large list in changelog 😅

casperklein:
`dms-services.conf` sounds perfect 👍

> Might be best to schedule it for v15, as v14 breaking changes is already quite a large list in changelog

One more tiny change makes no big difference IMO. But I don't care if this is done in v14 or 15.

georglauterbach:
I agree that this can easily go into v14.

polarathene:
Fair enough 👍

PR #3908, opened 2024-02-24
Rename supervisor-app.conf to dms-services.conf

casperklein:
# Description

Discussed in https://github.com/docker-mailserver/docker-mailserver/pull/3901

## Type of change

## Checklist


casperklein:
I've added the necessary info to `CHANGELOG.md`. But personally, I don't consider such changes as breaking. Breaking for me means, a user can't update a default setup from one version to another without making active changes to keep existing functionality. This is not the case here. Every default (supported) setup will just work fine.
Otherwise every (internal) file changes (renaming functions etc), would also considered as breaking, because someone could have scripted something that uses an old function name for example.

georglauterbach (approved):

polarathene:
👍 

---

> Otherwise every (internal) file changes (renaming functions etc), would also considered as breaking, because someone could have scripted something that uses an old function name for example.

It's only meant to be considered a breaking internal change if we had some public documentation about it? Only appears to be a separate config though:

> https://docker-mailserver.github.io/docker-mailserver/latest/faq/#how-to-adjust-settings-with-the-user-patchessh-script
> 
> image

I thought we had advice about modifying the `supervisord.conf` file, but that was done internally for a specialized log ENV we support:

https://github.com/docker-mailserver/docker-mailserver/blob/aa9465773c99a25dea7ca9534f1928ed6301693a/target/scripts/startup/setup-stack.sh#L37-L49

I thought the config name might have mattered, but also see that it's not used explicitly 😅 

https://github.com/docker-mailserver/docker-mailserver/blob/aa9465773c99a25dea7ca9534f1928ed6301693a/target/supervisor/supervisord.conf#L27-L31

PR #3925, opened 2024-03-04
fix: Move spam to mailbox associated to the `\Junk` special-use attribute (#3918)

### LINKED ISSUE #3918: feature request: Move spam to mailbox associated to the `\Junk` special-use attribute
### Context

For now, I can rename Junk mailbox (and others) as described in [documentation](https://docker-mailserver.github.io/docker-mailserver/latest/examples/use-cases/imap-folders/). But additional changes may be necessary, namely replacing 'Junk' string to selected mailbox name in two other files:
1. If `MOVE_SPAM_TO_JUNK=1`, in file `/usr/lib/dovecot/sieve-global/after/spam_to_junk.sieve`
2. If `RSPAMD_LEARN=1`, in file `/etc/dovecot/conf.d/90-sieve.conf`

This can be done in `user-patches.sh` using, for example, `sed` utility. But I think we can do better.

### Description

Dovecot 2.3 added support for [RFC 8579](https://datatracker.ietf.org/doc/html/rfc8579.html). If we enable Sieve extension 'special-use', we can rewrite `spam_to_junk.sieve` as follows:
```
require ["fileinto","special-use"];

if anyof (header :contains "X-Spam-Flag" "YES",
          header :contains "X-Spam" "Yes") {
    fileinto :specialuse "\Junk" "Junk";
}
```
This will eliminate the need to patch the file `spam_to_junk.sieve`.

Unfortunately, some changes in IMAPSieve plugin required to solve the problem with `90-sieve.conf` (allowing to specify special-use attribute in `imapsieve_mailboxXXX_name` and `imapsieve_mailboxXXX_from` settings). I've already [suggested](https://dovecot.org/mailman3/hyperkitty/list/dovecot@dovecot.org/thread/QDU2BIJFTU7YRUJUZFKXG5COUQYTREME/) this feature in Dovecot mailing list and waiting for reply.

### Alternatives

Of course, instead we can add new configuration parameter like `JUNK_MAILBOX_NAME` in docker-mailserver. But what will happen, if user rename Junk mailbox and forget to set/change this parameter?

### Applicable Users

Users, who rename Junk mailbox on their servers (for translation purpose or because they just prefer 'Spam' over 'Junk').

I've tested changes to `spam_to_junk.sieve` on recent version of docker-mailserver, and it works.
For some reason I can not run project tests locally. I can successfully build image, but when I run `make clean tests` I get this error:
_The directory named as part of the path /var/log/mail/rspamd.log does not exist in section 'program:rspamd' (file: '/etc/supervisor/conf.d/dms-services.conf')_
What am I doing wrong?

UltraCoderRU:
If `MOVE_SPAM_TO_JUNK=1`, spam messages now will be moved to one mailbox with `\Junk` special-use attribute. If there is no such mailbox, messages will be moved to 'Junk' mailbox, as earlier.
This change simplifies configuration in case Junk mailbox needs to be renamed.

Fixes #3918

## Type of change

## Checklist

- [x] I have made corresponding changes to the documentation (README.md or the documentation under docs/)
- [x] I have added information about changes made in this PR to CHANGELOG.md

### DIFF
--- a/test/tests/parallel/set1/spam_virus/spam_junk_folder.bats
+++ b/test/tests/parallel/set1/spam_virus/spam_junk_folder.bats
@@ -49,7 +49,7 @@ function teardown() { _default_teardown ; }
   _should_receive_spam_at '/var/mail/localhost.localdomain/user1/new/'
 }
 
-@test "(enabled + MOVE_SPAM_TO_JUNK=1) should deliver spam message into Junk folder" {
+@test "(enabled + MOVE_SPAM_TO_JUNK=1) should deliver spam message into folder with \Junk attribute" {
   export CONTAINER_NAME=${CONTAINER3_NAME}
 
   local CUSTOM_SETUP_ARGUMENTS=(
@@ -62,12 +62,16 @@ function teardown() { _default_teardown ; }
   )
   _init_with_defaults
   _common_container_setup 'CUSTOM_SETUP_ARGUMENTS'
+  _run_in_container sed -i -e 's/mailbox Junk/mailbox Spam/' /etc/dovecot/conf.d/15-mailboxes.conf
+  assert_success
+  _run_in_container dovecot reload
+  assert_success
 
   _should_send_spam_message
   _should_be_received_by_amavis 'Passed SPAM {RelayedTaggedInbound,Quarantined}'
 
   # Should move delivered spam to Junk folder
-  _should_receive_spam_at '/var/mail/localhost.localdomain/user1/.Junk/new/'
+  _should_receive_spam_at '/var/mail/localhost.localdomain/user1/.Spam/new/'
 }
 
 # NOTE: Same as test for `CONTAINER3_NAME`, only differing by ENV `MARK_SPAM_AS_READ=1` + `_should_receive_spam_at` location
--- a/test/tests/parallel/set1/spam_virus/spam_junk_folder.bats
+++ b/test/tests/parallel/set1/spam_virus/spam_junk_folder.bats
@@ -52,6 +52,8 @@ function teardown() { _default_teardown ; }
 @test "(enabled + MOVE_SPAM_TO_JUNK=1) should deliver spam message into Junk mailbox" {
   export CONTAINER_NAME=${CONTAINER3_NAME}
 
+  _init_with_defaults
+
   local CUSTOM_SETUP_ARGUMENTS=(
     --env ENABLE_AMAVIS=1
     --env ENABLE_SPAMASSASSIN=1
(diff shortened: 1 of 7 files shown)


test/tests/parallel/set1/spam_virus/spam_junk_folder.bats
```diff
@@ -62,12 +62,16 @@ function teardown() { _default_teardown ; }
   )
   _init_with_defaults
   _common_container_setup 'CUSTOM_SETUP_ARGUMENTS'
+  _run_in_container sed -i -e 's/mailbox Junk/mailbox Spam/' /etc/dovecot/conf.d/15-mailboxes.conf
+  assert_success
+  _run_in_container dovecot reload
+  assert_success
```
polarathene, line 68: This should be handled via `user-patches.sh`:

- Store it in `test/config/junk-mailbox/user-patches.sh` similar to [this example](https://github.com/docker-mailserver/docker-mailserver/blob/d227d6dc7368274fa5c2b206e77b33274b4569bc/test/config/dovecot-lmtp/user-patches.sh). You will not need the `dovecot reload` this way, and it will better reflect testing the equivalent of where a user would introduce the change.
- Here is the associated test file](https://github.com/docker-mailserver/docker-mailserver/blob/d227d6dc7368274fa5c2b206e77b33274b4569bc/test/tests/parallel/set3/mta/lmtp_ip.bats#L26-L29) that uses that referenced `user-patches.sh`. Note how it's configured with the `mv` _before_ `_common_container_setup`, and _after_ `_init_with_defaults`.
- Use the blank lines like the reference example does for `mv`. Please include a brief comment for context to the reader so it's quicker to grok at a glance that [you're targeting this line](https://github.com/docker-mailserver/docker-mailserver/blob/d227d6dc7368274fa5c2b206e77b33274b4569bc/target/dovecot/15-mailboxes.conf#L54) / context. Suggested comment: "Adjust 'Junk' mailbox name to verify special-use flag delivers to modified mailbox folder name"

test/tests/parallel/set1/spam_virus/spam_junk_folder.bats
```diff
@@ -62,12 +62,16 @@ function teardown() { _default_teardown ; }
...
 
   _should_send_spam_message
   _should_be_received_by_amavis 'Passed SPAM {RelayedTaggedInbound,Quarantined}'
 
   # Should move delivered spam to Junk folder
```
polarathene, line 73: A little more context since we're testing more than one configuration now:

```suggestion
  # Should move delivered spam to the Junk mailbox (adjusted to be located at '.Spam/')
```

georglauterbach:
I've restarted the CI so we can see whether the failure was related to this PR, which I doubt. Sometimes, our CI is not 100% accurate, i.e., it is a bit "flaky".

test/tests/parallel/set1/spam_virus/spam_junk_folder.bats
UltraCoderRU: Thank you for providing an example. I was not sure how to make changes to configuration files the best way, but found `_run_in_container` helper function and used it.

> Suggested comment: "Adjust 'Junk' mailbox name to verify special-use flag delivers to modified mailbox folder name"

Sounds a little weird to me. How about this?
"Adjust 'Junk' mailbox name to verify delivery to Junk mailbox based on special-use flag instead of mailbox's name"

polarathene: Yeah that revision is good too!

I guess you're just changing to `Spam` for the mailbox folder name to ensure it's not so similar to the special-use flag `\Junk`?

polarathene (dismissed):
LGTM! 👍 

If you'd like to add that revision you suggested that's ok, I'll hold off merging until I hear back from you about that.

No need to rewrite commit history cleanly if that's a hassle, we squash merge PRs so commit history is only relevant to PR context 👍

UltraCoderRU:
> If you'd like to add that revision you suggested that's ok, I'll hold off merging until I hear back from you about that.

Done.

polarathene (approved):

PR #3929, opened 2024-03-09
Better support regular container restarts

### LINKED ISSUE #3886: bug report: Improper restarts
### 📝 Preliminary Checks

### 👀 What Happened?

I'm doing a docker restart (e.g. `docker restart <container>` or `docker compose restart <service>`) but on startup it logs:

```
[ WARNING ]  This container was (likely) improperly restarted which can result in undefined behavior
[ WARNING ]  Please destroy the container properly and then start DMS again
```

If not this, then what would be a clean way to restart? Of course `docker compose up -d --force-recreate <service>` works, but that recreates the container, not really a restart.

Additionally, as a result of this "improper restart" I also get these errors that I don't get on normal startup:

```
mv: cannot stat '/etc/cron.d/dovecot-purge.disabled': No such file or directory
mv: cannot stat '/etc/cron.d/amavisd-new': No such file or directory
ln: failed to create symbolic link '/usr/lib/dovecot/sieve-pipe/rspamc': File exists
```

After this, is it in a messed up state? And should I recreate the container?

### 👟 Reproduction Steps

_No response_

### 🐋 DMS Version

v13.3.1

FedoraCoreOS 39 x86_64

```yml
  mail:
    container_name: mail
    image: mailserver/docker-mailserver
    hostname: ######
    restart: always
    cap_add:
      - NET_ADMIN
    ports:
      - '25:25'
      - '465:465'
      - '587:587'
      - '993:993'
    volumes:
      - mail-data:/var/mail
      - mail-state:/var/mail-state
      - mail-logs:/var/log/mail
      - mail-config:/tmp/docker-mailserver
      - letsencrypt:/etc/letsencrypt
    environment:
      ENABLE_RSPAMD: 1
      ENABLE_FAIL2BAN: 1
      ENABLE_CLAMAV: 0
      ENABLE_OPENDMARC: 0
      ENABLE_OPENDKIM: 0
      ENABLE_POLICYD_SPF: 0
      ENABLE_AMAVIS: 0
      ENABLE_SPAMASSASSIN: 0
      ENABLE_UPDATE_CHECK: 0
      DOVECOT_MAILBOX_FORMAT: mdbox
      SSL_TYPE: letsencrypt
      SPOOF_PROTECTION: 1
      MOVE_SPAM_TO_JUNK: 1
      MARK_SPAM_AS_READ: 1
      RSPAMD_LEARN: 1
      TZ: Europe/Amsterdam
```

### 📜 Relevant log output

```Text
[   INF   ]  Starting daemons
[   INF   ]  ###### is up and running
2024-02-13 00:09:51,296 WARN received SIGTERM indicating exit request
[   INF   ]  Welcome to docker-mailserver v13.3.1
[ WARNING ]  This container was (likely) improperly restarted which can result in undefined behavior
[ WARNING ]  Please destroy the container properly and then start DMS again
mv: cannot stat '/etc/cron.d/dovecot-purge.disabled': No such file or directory
mv: cannot stat '/etc/cron.d/amavisd-new': No such file or directory
ln: failed to create symbolic link '/usr/lib/dovecot/sieve-pipe/rspamc': File exists
[   INF   ]  Starting daemons
[   INF   ]  ###### is up and running
```

_No response_

casperklein:
# Description

Described [here](https://github.com/docker-mailserver/docker-mailserver/issues/3886#issuecomment-1943416349).

Not covered by this PR:

- Documentation
- Additional test(s)

Fixes #3886

## Type of change

## Checklist

### DIFF
--- a/target/scripts/start-mailserver.sh
+++ b/target/scripts/start-mailserver.sh
@@ -185,7 +185,7 @@ else
   # container was restarted
   _early_variables_setup
 
-  _log 'info' 'Container was restarted. Skipping setup routines..'
+  _log  'info' 'Container was restarted. Skipping setup routines.
   _log 'info' "Welcome to docker-mailserver ${DMS_RELEASE}"
 
   _register_functions
--- a/target/scripts/start-mailserver.sh
+++ b/target/scripts/start-mailserver.sh
@@ -185,7 +185,7 @@ else
   # container was restarted
   _early_variables_setup
 
-  _log  'info' 'Container was restarted. Skipping setup routines.
+  _log 'info' 'Container was restarted. Skipping setup routines.'
   _log 'info' "Welcome to docker-mailserver ${DMS_RELEASE}"
 
   _register_functions
--- a/target/scripts/start-mailserver.sh
+++ b/target/scripts/start-mailserver.sh
@@ -170,20 +170,33 @@ function _register_functions() {
 # ? >> Executing all stacks / actual start of DMS
 # ------------------------------------------------------------
 
-_early_supervisor_setup
-_early_variables_setup
+# fresh container?
+if [[ ! -f /CONTAINER_START ]]; then
+  # first container start
+  _early_supervisor_setup
+  _early_variables_setup
 
-_log 'info' "Welcome to docker-mailserver ${DMS_RELEASE}"
+  _log 'info' "Welcome to docker-mailserver ${DMS_RELEASE}"
 
-_register_functions
-_check
-_setup
-[[ ${LOG_LEVEL} =~ (debug|trace) ]] && print-environment
-_run_user_patches
-_start_daemons
+  _register_functions
+  _check
+  _setup
+  _run_user_patches
+else
+  # container was restarted
+  _early_variables_setup
+
+  _log 'info' 'Container was restarted. Skipping setup routines..'
+  _log 'info' "Welcome to docker-mailserver ${DMS_RELEASE}"
+
+  _register_functions
+fi
 
 # marker to check if container was restarted
 date >/CONTAINER_START
+
+[[ ${LOG_LEVEL} =~ (debug|trace) ]] && print-environment
+_start_daemons
 
 _log 'info' "${HOSTNAME} is up and running"
 
--- a/target/scripts/start-mailserver.sh
+++ b/target/scripts/start-mailserver.sh
@@ -169,9 +169,9 @@ function _register_functions() {
 # ? >> Executing all stacks / actual start of DMS
 # ------------------------------------------------------------
 
-# fresh container?
+# Ensure DMS only adjusts config files for a new container.
+# Container restarts should skip as they retain the modified config.
 if [[ ! -f /CONTAINER_START ]]; then
-  # first container start
   _early_supervisor_setup
   _early_variables_setup
 
--- a/target/scripts/start-mailserver.sh
+++ b/target/scripts/start-mailserver.sh
@@ -201,6 +201,4 @@ _start_daemons
 _log 'info' "${HOSTNAME} is up and running"
 
 touch /var/log/mail/mail.log
-tail -Fn 0 /var/log/mail/mail.log
-
-exit 0
+exec tail -Fn 0 /var/log/mail/mail.log
--- a/target/scripts/start-mailserver.sh
+++ b/target/scripts/start-mailserver.sh
@@ -33,7 +33,6 @@ function _register_functions() {
 
   # ? >> Checks
 
-  _register_check_function '_check_improper_restart'
   _register_check_function '_check_hostname'
   _register_check_function '_check_log_level'
   _register_check_function '_check_spam_prefix'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,9 @@ The most noteworthy change of this release is the update of the container's base
   - `LOGROTATE_COUNT` defines the number of files kept by logrotate ([#3907](https://github.com/docker-mailserver/docker-mailserver/pull/3907))
     - The fail2ban log file is now also taken into account by `LOGROTATE_COUNT` and `LOGROTATE_INTERVAL` ([#3915](https://github.com/docker-mailserver/docker-mailserver/pull/3915), [#3919](https://github.com/docker-mailserver/docker-mailserver/pull/3919))
 
+- **Internal:**
+  - Support container restart ([3929](https://github.com/docker-mailserver/docker-mailserver/pull/3929))
+
 ### Updates
 
 - **Environment Variables:**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,7 +75,7 @@ The most noteworthy change of this release is the update of the container's base
     - The fail2ban log file is now also taken into account by `LOGROTATE_COUNT` and `LOGROTATE_INTERVAL` ([#3915](https://github.com/docker-mailserver/docker-mailserver/pull/3915), [#3919](https://github.com/docker-mailserver/docker-mailserver/pull/3919))
 
 - **Internal:**
-  - Support container restart ([3929](https://github.com/docker-mailserver/docker-mailserver/pull/3929))
+  - Regular container restarts are now better supported. Setup scripts that ran previously will now be skipped ([#3929](https://github.com/docker-mailserver/docker-mailserver/pull/3929))
 
 ### Updates
 
--- a/target/scripts/startup/check-stack.sh
+++ b/target/scripts/startup/check-stack.sh
@@ -14,15 +14,6 @@ function _check() {
   done
 }
 
-function _check_improper_restart() {
-  _log 'debug' 'Checking for improper restart'
-
-  if [[ -f /CONTAINER_START ]]; then
-    _log 'warn' 'This container was (likely) improperly restarted which can result in undefined behavior'
-    _log 'warn' "Please use 'docker compose up --force-recreate' or equivalent (view our troubleshooting docs)"
-  fi
-}
-
 function _check_hostname() {
   _log 'debug' 'Checking that hostname/domainname is provided or overridden'
 
--- a/target/scripts/startup/setup-stack.sh
+++ b/target/scripts/startup/setup-stack.sh
@@ -40,7 +40,7 @@ function _early_supervisor_setup() {
   if ! grep -q "loglevel = ${SUPERVISOR_LOGLEVEL}" /etc/supervisor/supervisord.conf; then
     case "${SUPERVISOR_LOGLEVEL}" in
       ( 'critical' | 'error' | 'info' | 'debug' )
-        sed -i -E \
+        sedfile -i -E \
           "s|(loglevel).*|\1 = ${SUPERVISOR_LOGLEVEL}|g" \
           /etc/supervisor/supervisord.conf
 


target/scripts/start-mailserver.sh
```diff
@@ -170,24 +169,35 @@ function _register_functions() {
 # ? >> Executing all stacks / actual start of DMS
 # ------------------------------------------------------------
 
-_early_supervisor_setup
-_early_variables_setup
+# fresh container?
+if [[ ! -f /CONTAINER_START ]]; then
+  # first container start
```
polarathene, line 174: Might be better to communicate the intent more explicitly? I know we all know why, but it may not be obvious to other readers (_and hopeful future maintainers_)

```suggestion
# Ensure DMS only adjusts config files for a new container.
# Container restarts should skip as they retain the modified config.
if [[ ! -f /CONTAINER_START ]]; then
```

target/scripts/start-mailserver.sh
```diff
@@ -170,24 +169,35 @@ function _register_functions() {
...
+  _setup
+  _run_user_patches
+else
+  # container was restarted
+  _early_variables_setup
+
+  _log 'info' 'Container was restarted. Skipping setup routines..'
+  _log 'info' "Welcome to docker-mailserver ${DMS_RELEASE}"
+
+  _register_functions
```
polarathene, line 191: It seems like the two `_early` methods could be handled before this conditional branch? `_early_supervisor_setup` isn't doing anything of concern with a container restart:

https://github.com/docker-mailserver/docker-mailserver/blob/267fc552d220414be9f8ba7dbb2f0ed2431f3228/target/scripts/startup/setup-stack.sh#L37-L62

While the only reason you're calling `_early_variables_setup` + `_register_functions` here is for the later `_start_daemons`? (_the later `print-environment` call is probably less meaningful with restarts, it may even be more helpful for us knowing that it's only output when the user ensures a fresh container?_)

---

So realistically, if the container is fresh we want to run:

```bash
_check
_setup
_run_user_patches

[[ ${LOG_LEVEL} =~ (debug|trace) ]] && print-environment
```

Otherwise:

```bash
_log 'info' 'Container was restarted. Skipping setup routines..'
```

Do you think that makes more sense to reduce the conditional scope to just that?

target/scripts/start-mailserver.sh
```diff
@@ -170,24 +169,35 @@ function _register_functions() {
...
+[[ ${LOG_LEVEL} =~ (debug|trace) ]] && print-environment
+_start_daemons
+
 _log 'info' "${HOSTNAME} is up and running"
 
 touch /var/log/mail/mail.log
-tail -Fn 0 /var/log/mail/mail.log
-
-exit 0
+exec tail -Fn 0 /var/log/mail/mail.log
```
polarathene, line 203: 👍

polarathene (changes requested):

target/scripts/start-mailserver.sh
```diff
@@ -170,24 +169,35 @@ function _register_functions() {
 # ? >> Executing all stacks / actual start of DMS
 # ------------------------------------------------------------
 
-_early_supervisor_setup
-_early_variables_setup
+# Ensure DMS only adjusts config files for a new container.
+# Container restarts should skip as they retain the modified config.
+if [[ ! -f /CONTAINER_START ]]; then
+  _early_supervisor_setup
+  _early_variables_setup
```
polarathene, line 176:
```suggestion
_early_supervisor_setup
_early_variables_setup
_register_functions
_log 'info' "Welcome to docker-mailserver ${DMS_RELEASE}"
  
# Ensure DMS only adjusts config files for a new container.
# Container restarts should skip as they retain the modified config.
if [[ ! -f /CONTAINER_START ]]; then
  _check
  _setup
  _run_user_patches

  # Only displayed on fresh containers (better for bug reports)
  [[ ${LOG_LEVEL} =~ (debug|trace) ]] && print-environment
else
  _log 'info' 'Container was restarted. Skipping setup routines..'
fi
```

target/scripts/start-mailserver.sh
```diff
@@ -170,24 +169,35 @@ function _register_functions() {
...
 
-_log 'info' "Welcome to docker-mailserver ${DMS_RELEASE}"
+  _log 'info' "Welcome to docker-mailserver ${DMS_RELEASE}"
```
polarathene, line 178:
```suggestion
```

target/scripts/start-mailserver.sh
```diff
@@ -170,24 +169,35 @@ function _register_functions() {
...
+  _setup
+  _run_user_patches
+else
+  # container was restarted
+  _early_variables_setup
+
+  _log 'info' 'Container was restarted. Skipping setup routines..'
+  _log 'info' "Welcome to docker-mailserver ${DMS_RELEASE}"
+
+  _register_functions
```
polarathene, line 191:
```suggestion
```

target/scripts/start-mailserver.sh
```diff
@@ -170,24 +169,35 @@ function _register_functions() {
...
+fi
```
polarathene, line 192:
```suggestion
```

target/scripts/start-mailserver.sh
```diff
@@ -170,24 +169,35 @@ function _register_functions() {
...
 
 # marker to check if container was restarted
 date >/CONTAINER_START
 
+[[ ${LOG_LEVEL} =~ (debug|trace) ]] && print-environment
```
polarathene, line 197:
```suggestion
```

polarathene (changes requested):
Earlier suggestion for simplifying the conditional branch. Suggestion is slightly awkward due to web UI functionality 😅

casperklein:
> Earlier suggestion for simplifying the conditional branch. Suggestion is slightly awkward due to web UI functionality 😅

Thanks, but I already got what you proposed 👍  I just had no time yesterday answering you.

The reason I did it my way is: 

Depending on the log level defined in `SUPERVISOR_LOGLEVEL`, `_early_supervisor_setup` modifies the supervisor configuration and runs a reload afterwards. When supervisord is reloaded, all services (only [mailserver](https://github.com/docker-mailserver/docker-mailserver/blob/2133b51e782ae207475788a9b168b0008a5b450b/target/supervisor/conf.d/dms-services.conf#L7-L16) at this point) are restarted. That means `start-mailserver.sh` is run a second time. That is not necessary, especially on a container restart. That's why `_early_supervisor_setup` should only be called run once (on first container start).

After your comment, I read the source again and noticed, that for that situation there is already a [safe-guard](https://github.com/docker-mailserver/docker-mailserver/blob/2133b51e782ae207475788a9b168b0008a5b450b/target/scripts/startup/setup-stack.sh#L40) not to try to modify the configuration (and reloading supervisor), when not needed. But I think the intention for that in the first place was not in regard of a container restart.

Even with that safe-guard in place, there is no need to call that function on container restarts.

Regarding `_early_variables_setup`: We could put that above the `if [[ ! -f /CONTAINER_START ]]; then` condition. In the case of `SUPERVISOR_LOGLEVEL=info` for example, supervisord is reloaded later on the first container start, which then runs `start-mailserver.sh` again. `_early_variables_setup` would then be called again, which makes the first invocation useless.

I hope you understand what I tried to say. Summery: I did it that way, with a view to effectiveness (avoiding double functions calls on some circumstances) --> saving some CPU cycles 😆

polarathene:
> Depending on the log level defined in `SUPERVISOR_LOGLEVEL`, `_early_supervisor_setup` modifies the supervisor configuration and runs a reload afterwards.

Ideally it shouldn't be a problem for container restarts since ENV doesn't change.

> When supervisord is reloaded, all services (only [mailserver](https://github.com/docker-mailserver/docker-mailserver/blob/2133b51e782ae207475788a9b168b0008a5b450b/target/supervisor/conf.d/dms-services.conf#L7-L16) at this point) are restarted.
> That means `start-mailserver.sh` is run a second time. That is not necessary, especially on a container restart.
> That's why `_early_supervisor_setup` should only be called run once (on first container start).

This is the actual problem, and is due to:

https://github.com/docker-mailserver/docker-mailserver/blob/2133b51e782ae207475788a9b168b0008a5b450b/Dockerfile#L316-L317

https://github.com/docker-mailserver/docker-mailserver/blob/2133b51e782ae207475788a9b168b0008a5b450b/target/supervisor/conf.d/dms-services.conf#L7-L16

https://github.com/docker-mailserver/docker-mailserver/blob/2133b51e782ae207475788a9b168b0008a5b450b/target/scripts/start-mailserver.sh#L190-L191

What is the technical reason for needing to run the startup script as a separate service? If we didn't then this early supervisord support wouldn't be necessary and the script would not need to restart it and exit itself to re-run?

https://github.com/docker-mailserver/docker-mailserver/blob/2133b51e782ae207475788a9b168b0008a5b450b/target/scripts/startup/setup-stack.sh#L47-L48

---

> I hope you understand what I tried to say. 

Yes thanks, I hadn't considered it.

Although it should be fine for our default `CMD` to be `start-mailserver.sh`, so long as it starts supervisor via `exec` at the end it'd be PID 1? (_although that'd probably require refactoring of [`daemon-stack.sh`](https://github.com/docker-mailserver/docker-mailserver/blob/2133b51e782ae207475788a9b168b0008a5b450b/target/scripts/startup/daemons-stack.sh), and may complicate the log output a bit_ 🤷‍♂️)

polarathene (changes requested):

polarathene (approved):

target/scripts/start-mailserver.sh
```diff
@@ -170,24 +169,35 @@ function _register_functions() {
 # ? >> Executing all stacks / actual start of DMS
 # ------------------------------------------------------------
 
-_early_supervisor_setup
-_early_variables_setup
+# Ensure DMS only adjusts config files for a new container.
+# Container restarts should skip as they retain the modified config.
+if [[ ! -f /CONTAINER_START ]]; then
```
georglauterbach, line 174: I think reading the non-negated statement is easier. Because we have an `else`, having `-f /CONTAINER_START` is feasible.

target/scripts/start-mailserver.sh
```diff
@@ -170,24 +169,35 @@ function _register_functions() {
...
-_start_daemons
+  _register_functions
+  _check
+  _setup
+  _run_user_patches
+else
+  # container was restarted
+  _early_variables_setup
+
+  _log 'info' 'Container was restarted. Skipping setup routines..'
```
georglauterbach, line 188:
```suggestion
  _log  'info' 'Container was restarted. Skipping setup routines.
```

georglauterbach (changes requested):
Two small nitpicks. Otherwise, LGMT 👍🏼 Why doesn't this PR ship the corresponding documentation updates?

target/scripts/start-mailserver.sh
casperklein: I did it that way, because I wanted the "fresh container" routines to come first and the "container was restarted" stuff later in the else block.
We can change that, but the reasoning that the if statement is hard to read because of the negation is thin 😉 (IMO).

georglauterbach: Absolutely fine with me :)

georglauterbach:
Oops, my suggestion contained an additional space - praise be linters 😝

polarathene:
> I don't care about this one and can change that, but I am wondering what your reasoning for that is.
> If someone enables debugging, why should we omit debug information on container restarts?

If they enable the ENV it'll do a fresh start AFAIK. Unless there's other surprising cases like how compose will keep a `VOLUME` (_implicit anonymous volume in `Dockerfile`_) even with `--force-recreate` or a new tag upgrade (_it seems to key it to the service name and `VOLUME` instruction regardless of image_).

Sure there might be a legitimate need for it with a container restart, I just figured we'd prefer bug reports for fresh containers only and this would be more evident in their log output without us having to do extra back and forth?

If there is a legit bug that is reproduced only with a container restart, they could manually get the equivalent output since we write it to a file anyway?

> There are solutions/logging drivers where the log size is limited and you can't scroll back to the very beginning and only get the last X lines.

In that sense, this output is even less helpful? Personally I'd rather we had this limited to `trace` if we need early output for a startup failure, otherwise we could just instruct the user to run a `setup debug ...` command to get the same output separate from the main log output?

Often when ENV matters, the user doesn't provide such log output, and we're just given a `mailserver.env` reference and have to ask for that, which is often a copy/paste dump with plenty of default values for us to wade through, even at times when I ask for them to only share the non-default values they changed 😩

polarathene (approved):

georglauterbach:
LGTM 👍🏼 Still, why isn't this PR adjusting our docs?

casperklein:
> If they enable the ENV it'll do a fresh start AFAIK

? If you change ENVs and want to take them place, you'll have to recreate the container. "start/stop" does not take changed variables into account.

> .. I just figured we'd prefer bug reports for fresh containers only and this would be more evident in their log output without us having to do extra back and forth?

I haven't thought about this at all 👍 For me, enabling debugging is not mutual exclusive to opening an issue. Enabling any feature, I would not expect, that some functions are only run on first container start. That's why I find it in this case inconsistent, when printing the ENVs, would only take place on the first container start.

> Often when ENV matters, the user doesn't provide such log output

That's unfortunately nothing we can fix scripting wise 😆

---

> LGTM 👍🏼 Still, why isn't this PR adjusting our docs?

I haven't checked it, but I guess that there are many places that need to be adjusted. I simply have no time for that right now. We can leave this PR open if you or @polarathene want to do the docu stuff.

On the other hand, starting with a fresh container is still a good advice. That "wrong" information does not hurt in any way. We could wait some time to see, how well this PR ages and adjust the docs, when we are 100% confident, that container restarts are fully supported and there are no more issues with that.

polarathene:
> Enabling any feature, I would not expect, that some functions are only run on first container start.

But that's exactly what this PR does, conditionally run startup logic based on if this is a new container instance or not?

> That's why I find it in this case inconsistent, when printing the ENVs, would only take place on the first container start.

This is related to troubleshooting right? That's the main purpose of that output, yet by skipping the output from running initial container setup, you're lacking any helpful insights that come from that which can emit warnings/errors.

It's better to encourage a fresh container instance for this and by ensuring the ENV print method is also only run for that it will be communicated much better to us with bug reports, and for the user who's more likely to see the skipped log message this PR adds when there is less noise being output?

> That's unfortunately nothing we can fix scripting wise 😆

Disagree, we know what are default ENV values are, we could absolutely output only non-defaults, either at startup via log or a `setup debug` command.

> ? If you change ENVs and want to take them place, you'll have to recreate the container. "start/stop" does not take changed variables into account.

That was kind of my point. With `docker start` you're resuming a container and the ENV will carry over as you can't adjust them with that command.

With Docker Compose, IIRC adjusting ENV via `compose.yaml` would get detected and recreate the container, at least with the common `docker compose up` command?

So if a user explicitly changes ENV for a container, they'll need a fresh container instance regardless.

Not sure what value we're providing to the user by keeping ENV printing at container startup for restarts/resumes?

---

> I haven't checked it, but I guess that there are many places that need to be adjusted. I simply have no time for that right now. We can leave this PR open if you or @polarathene want to do the docu stuff.

I've not chimed in on the docs as I don't know what docs are being referred to that need updating. There's two pages (usage and debugging) with the `docker compose up --force-recreate` advice.

Does this PR completely address the internal state issue?

> We could wait some time to see, how well this PR ages and adjust the docs, when we are 100% confident, that container restarts are fully supported and there are no more issues with that.

👍 

I don't see issue with leaving doc as-is for now either. Rather have that there for now and we can remove it later if it doesn't provide value. I think it's still good advice for the debugging page.

casperklein:
Let's stop the discussion at this point about printing the ENVs on startup if that's fine for you. I see you have valid points - me too. I was about to reply to your points and noticed it was getting quiet lengthy again. Surely you would reply again to my answers. I think it's not worth to invest that much time for 1 LOC. If you or Georg demand a change, I can do it or you can change it directly 👍

---

> Does this PR completely address the internal state issue?

I hope so, but I wouldn't bet any money on this 😆

georglauterbach:
Agreed, let's not make a lengthy discussion out of it 🚀 I am fine with not updating the docs for now, but only if it does not become the default behavior. We also need to update the docs when we are confident that all is well with this feature.

georglauterbach (approved):

PR #3937, opened 2024-03-13
docs: Update links for account management in `README.md`

rahilarious:
# Description

## Type of change


polarathene (approved):
Thanks for spotting this ❤️  (_broken for almost a year!_ 🙈 )

I normally would keep the version at `latest`, but plan to change to make an update to this page so I'll approve and merge as-is with the `v13.3` pin 👍

PR #1859, opened 2021-03-22
yaml base-60 float warning added

casperklein:
# Description

Removing the quotation marks in the docker-compose port section, can lead to unexpected behaviour.

More Information:

- https://github.com/compose-spec/compose-spec/blob/master/spec.md#ports
- https://yaml.org/type/float.html

Fixes #1858 and other similar

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Improvement (non-breaking change that does improve existing functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (README.md or ENVIRONMENT.md or the Wiki)
- [ ] If necessary I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

Issue #1858, opened 2021-03-21
Incoming e-mails from external servers are not received


DeveloperPad:
# Subject

Incoming e-mails from external servers are not received.

## Description

Hey there,

I recently installed your mailserver and it works pretty well.
The only problem I just cannot solve is the receival of incoming e-mails from external servers.
Intra-server mail transmission works fine though, so I think that the problem is somehow related to my DNS settings or maybe a misconfiguration of the mailserver (e.g. the hostname).

Listening on port 25 for inter-mailserver communication traffic results in this output directly after sending an e-mail from another e-mail domain to my mailserver domain.
``` log
Capturing on 'eth0'
1 0.000000000  40.92.75.97 → 75.119.129.6 TCP 66 15591 → 25 [SYN, ECN, CWR] Seq=0 Win=64240 Len=0 MSS=1460 WS=256 SACK_PERM=1
2 0.000067356 75.119.129.6 → 40.92.75.97  TCP 54 25 → 15591 [RST, ACK] Seq=1 Ack=1 Win=0 Len=0
3 0.523886062  40.92.75.42 → 75.119.129.6 TCP 66 38734 → 25 [SYN] Seq=0 Win=64240 Len=0 MSS=1460 WS=256 SACK_PERM=1
4 0.523974447 75.119.129.6 → 40.92.75.42  TCP 54 25 → 38734 [RST, ACK] Seq=1 Ack=1 Win=0 Len=0
5 1.044128701  40.92.75.65 → 75.119.129.6 TCP 66 16822 → 25 [SYN] Seq=0 Win=64240 Len=0 MSS=1460 WS=256 SACK_PERM=1
6 1.044193733 75.119.129.6 → 40.92.75.65  TCP 54 25 → 16822 [RST, ACK] Seq=1 Ack=1 Win=0 Len=0
7 1.067185129 2a01:111:f400:fe0e::829 → 2a02:c206:3006:9356::1 TCP 86 30336 → 25 [SYN, ECN, CWR] Seq=0 Win=64800 Len=0 MSS=1440 WS=256 SACK_PERM=1
8 1.067225966 2a02:c206:3006:9356::1 → 2a01:111:f400:fe0e::829 TCP 74 25 → 30336 [RST, ACK] Seq=1 Ack=1 Win=0 Len=0
[...]

PR #3982, opened 2024-04-19
chore: Remove base-60 port quote warning from example `compose.yaml`

polarathene:
This should not be relevant to users of `docker compose` which is the primary demographic.

However, I know that we have users with other compose clients that parse `compose.yaml` where it may still be a gotcha, so feel free to reject this PR 👍 

---

## Additional information

`docker compose` (v2) [uses the v3 version of `go-yaml`](https://github.com/docker/compose/blob/4ae2e0ec551d5707da2639fb7e1783cc1a21892e/go.mod#L56C2-L56C18) for parsing YAML where **this gotcha will not occur**. This package [intentionally avoids base-60 syntax](https://github.com/go-yaml/yaml?tab=readme-ov-file#compatibility) that we have raised awareness in our example `compose.yaml`. The Go rewrite of `docker compose` did not exist back when we added this AFAIK, and the earlier Python based version of Docker Compose was more prevalent (now EOL), which is where the bug was probably being encountered.

It's worth noting that while base-60 syntax (_aka sexagesimal_) is valid for YAML 1.1, it's been [**removed with YAML 1.2** of the spec](https://yaml.org/spec/1.2.2/ext/changes/). YAML 1.2 spec was **published in 2009Q3**, so there's been adequate time for parsers in modern maintained software to update to that.

The **last bug report** to [motivate the PR to add this](https://github.com/docker-mailserver/docker-mailserver/pull/1859) warning comment [was in March 2021](https://github.com/docker-mailserver/docker-mailserver/issues/1858#issuecomment-803700403), but the reporter did not provide much details about their environment. Presumably it'd be the fault of the YAML parser distributed with the `docker-compose` (Python) package though 🤷‍♂️ (_plus maintainers had confirmed it on their own systems_). So 12 years roughly may not have been adequate time 😅 

The [referenced **compose spec** still raises this awareness](https://github.com/compose-spec/compose-spec/blob/c35d19323fdfd7dbf9db30022498e29ef5ebeab2/spec.md?plain=1#L1623), which is fair as it's more generic towards implementers parsing YAML. It was [added to the spec directly](https://github.com/compose-spec/compose-spec/commit/dc687ad6c7d63ece2996f40fc9310c6679ee3b00#diff-bc6661da34ecae62fbe724bb93fd69b91a7f81143f2683a81163231de7e3b545R865) (Jan 2020) with many other changes and no PR process, no context provided.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -7,7 +7,6 @@ services:
     env_file: mailserver.env
     # More information about the mail-server ports:
     # https://docker-mailserver.github.io/docker-mailserver/latest/config/security/understanding-the-ports/
-    # To avoid conflicts with yaml base-60 float, DO NOT remove the quotation marks.
     ports:
       - "25:25"    # SMTP  (explicit TLS => STARTTLS, Authentication is DISABLED => use port 465/587 instead)
       - "143:143"  # IMAP4 (explicit TLS => STARTTLS)


georglauterbach:
That's a tough one. I always appreciate getting rid of old stuff that nobody uses anymore. In this case, I think it is warranted, too. I am just raising the concern that, in case we do see this issue popping back up, we should revert.

georglauterbach (approved):

casperklein (approved):

PR #4018, opened 2024-05-16
fix: `/var/mail-state` should not symlink non-existing directories

### LINKED ISSUE #4015: bug report: getmail not set up
### 📝 Preliminary Checks

### 👀 What Happened?

I set `ENABLE_GETMAIL` to 1 in my config file and create a `getmail-example.cf` file in my `dms/config/` folder. Getmail is not enabled and instead I recieve the following error to the postmaster address:

```
Error: Specified config/data dir "/var/lib/getmail" does not exist - create or specify alternate directory with --getmaildir option
```

If I examine the file structure I can indeed see that the `/var/lib/getmail` symlink links to `/var/mail-state/lib-getmail` which does not exist.

### 👟 Reproduction Steps

I tried recreating this with a fresh local instance. First I ran the commands to get the compose file and env file.

```
DMS_GITHUB_URL="https://raw.githubusercontent.com/docker-mailserver/docker-mailserver/master"
wget "${DMS_GITHUB_URL}/compose.yaml"
wget "${DMS_GITHUB_URL}/mailserver.env"
  ```

I ran the container and I created a user and and alias:
```
docker compose up -d
docker exec -ti mailserver setup email add user@example.com
docker exec -ti mailserver setup alias add postmaster@example.com user@example.com
```
 
I then edited the `mailserver.env` file to set 
  
```
LOG_LEVEL=debug
ENABLE_GETMAIL=1
```

I ran recreate the container with:
```
docker-compose down
docker-compose up
```

I can now see in `/var/mail/example.com/user/new/` the error emails related to the getmail folder not existing.

### 🐋 DMS Version

v13.3.1

Ubuntu 20.04.6 LTS

```yml
services:
  mailserver:
    container_name: mailserver
    hostname: mail.example.com
    env_file: mailserver.env
    # https://docker-mailserver.github.io/docker-mailserver/latest/config/security/understanding-the-ports/
    ports:
      - "25:25"    # SMTP  (explicit TLS => STARTTLS, Authentication is DISABLED => use port 465/587 instead)
    volumes:
      - ./docker-data/dms/mail-data/:/var/mail/
      - ./docker-data/dms/mail-state/:/var/mail-state/
      - ./docker-data/dms/mail-logs/:/var/log/mail/
      - ./docker-data/dms/config/:/tmp/docker-mailserver/
    restart: always
    stop_grace_period: 1m
    # Uncomment if using `ENABLE_FAIL2BAN=1`:
    # cap_add:
    #   - NET_ADMIN
    healthcheck:
      timeout: 3s
      retries: 0
```

### 📜 Relevant log output

```Text
[  DEBUG  ]  Handling general environment variable setup
[   INF   ]  Welcome to docker-mailserver v13.3.1
[  DEBUG  ]  Registering functions
[  DEBUG  ]  Checking for improper restart
[  DEBUG  ]  Checking that hostname/domainname is provided or overridden
[  DEBUG  ]  Domain has been set to example.com
[  DEBUG  ]  Hostname has been set to mail.example.com
[  DEBUG  ]  Setting up general log files
[  DEBUG  ]  Setting up Dovecot
[  DEBUG  ]  Enabling IMAP services
[  DEBUG  ]  Setting up Dovecot dhparam
[  DEBUG  ]  Setting up Dovecot quota
[  DEBUG  ]  Spam emails will be moved to the Junk folder
[  DEBUG  ]  Spam emails will not be marked as read
[  DEBUG  ]  Setting up Dovecot Local User
[  DEBUG  ]  Creating user 'user' for domain 'example.com'
[  DEBUG  ]  Configuring DKIM
[  DEBUG  ]  OpenDKIM enabled but no DKIM key(s) provided
[  DEBUG  ]  Configuring OpenDMARC
[  DEBUG  ]  Configuring policyd-spf
[  DEBUG  ]  Setting up Security Stack
[  DEBUG  ]  Postgrey is disabled
[  DEBUG  ]  Configuring Postscreen
[  DEBUG  ]  Disabling Postscreen DNSBLs
[  DEBUG  ]  SpamAssassin is disabled
[  DEBUG  ]  Disabling ClamAV
[  DEBUG  ]  Fail2Ban is disabled
[  DEBUG  ]  Configuring Amavis
[  DEBUG  ]  Rspamd is disabled
[  DEBUG  ]  Setting up SSL
[  DEBUG  ]  TLS configured with 'modern' ciphers
[ WARNING ]  !! INSECURE !! SSL configured with plain text access - DO NOT USE FOR PRODUCTION DEPLOYMENT
[  DEBUG  ]  Setting up PERMIT_DOCKER option
[  DEBUG  ]  Setting up mailname and creating '/etc/mailname'
[  DEBUG  ]  Applying hostname to Dovecot
[  DEBUG  ]  Configuring Postfix (early setup)
[  DEBUG  ]  '/tmp/docker-mailserver/postfix-virtual.cf' not provided - no mail alias/forward created
[  DEBUG  ]  Setting up Postfix dhparam
[  DEBUG  ]  Spoof protection is disabled
[  DEBUG  ]  Configuring Postfix (late setup)
[  DEBUG  ]  Setting up Postfix Relay Hosts
[  DEBUG  ]  (Postfix setup) Overriding / adjusting configuration with user-supplied values
[  DEBUG  ]  Fetchmail is disabled
[  DEBUG  ]  Fetchmail parallel is disabled
[  DEBUG  ]  Setting up logrotate
[  DEBUG  ]  Postfix log summary reports disabled
[  DEBUG  ]  Logwatch reports disabled.
[  DEBUG  ]  Consolidating all state onto /var/mail-state
[  DEBUG  ]  Checking /var/mail permissions
[  DEBUG  ]  Removing files and directories from older versions
[  DEBUG  ]  Ensuring  is owned by '_rspamd:_rspamd'
chown: cannot access '': No such file or directory
[  DEBUG  ]  Exporting environment variables now (creating '/etc/dms-settings')
[  DEBUG  ]  Setting up configuration checksum file
[  DEBUG  ]  Printing environment variables. Make sure no sensitive data is copied.
ACCOUNT_PROVISIONER='FILE'
AMAVIS_LOGLEVEL='0'
CLAMAV_MESSAGE_SIZE_LIMIT='25M'
DEFAULT_RELAY_HOST=''
DMS_VMAIL_GID='5000'
DMS_VMAIL_UID='5000'
DOVECOT_INET_PROTOCOLS='all'
DOVECOT_MAILBOX_FORMAT='maildir'
DOVECOT_TLS='no'
ENABLE_AMAVIS='1'
ENABLE_CLAMAV='0'
ENABLE_DNSBL='0'
ENABLE_FAIL2BAN='0'
ENABLE_FETCHMAIL='0'
ENABLE_GETMAIL='1'
ENABLE_IMAP='1'
ENABLE_MANAGESIEVE='0'
ENABLE_OAUTH2='0'
ENABLE_OPENDKIM='1'
ENABLE_OPENDMARC='1'
ENABLE_POLICYD_SPF='1'
ENABLE_POP3='0'
ENABLE_POSTGREY='0'
ENABLE_QUOTAS='1'
ENABLE_RSPAMD='0'
ENABLE_RSPAMD_REDIS='0'
ENABLE_SASLAUTHD='0'
ENABLE_SPAMASSASSIN='0'
ENABLE_SPAMASSASSIN_KAM='0'
ENABLE_SRS='0'
ENABLE_UPDATE_CHECK='1'
FAIL2BAN_BLOCKTYPE='drop'
FETCHMAIL_PARALLEL='0'
FETCHMAIL_POLL='300'
GETMAIL_POLL='5'
LOGROTATE_INTERVAL='weekly'
LOGWATCH_INTERVAL='none'
LOGWATCH_RECIPIENT='postmaster@example.com'
LOGWATCH_SENDER='mailserver-report@mail.example.com'
LOG_LEVEL='debug'
MARK_SPAM_AS_READ='0'
MOVE_SPAM_TO_JUNK='1'
NETWORK_INTERFACE='eth0'
ONE_DIR='1'
OVERRIDE_HOSTNAME=''
PERMIT_DOCKER='none'
PFLOGSUMM_RECIPIENT='postmaster@example.com'
PFLOGSUMM_SENDER='mailserver-report@mail.example.com'
PFLOGSUMM_TRIGGER='none'
POSTFIX_DAGENT=''
POSTFIX_INET_PROTOCOLS='all'
POSTFIX_MAILBOX_SIZE_LIMIT='0'
POSTFIX_MESSAGE_SIZE_LIMIT='10240000'
POSTFIX_REJECT_UNKNOWN_CLIENT_HOSTNAME='0'
POSTGREY_AUTO_WHITELIST_CLIENTS='5'
POSTGREY_DELAY='300'
POSTGREY_MAX_AGE='35'
POSTGREY_TEXT='Delayed by Postgrey'
POSTMASTER_ADDRESS='postmaster@example.com'
POSTSCREEN_ACTION='enforce'
RELAY_HOST=''
REPORT_RECIPIENT='postmaster@example.com'
REPORT_SENDER='mailserver-report@mail.example.com'
RSPAMD_CHECK_AUTHENTICATED='0'
RSPAMD_GREYLISTING='0'
RSPAMD_HFILTER='1'
RSPAMD_HFILTER_HOSTNAME_UNKNOWN_SCORE='6'
RSPAMD_LEARN='0'
SA_KILL='10.0'
SA_SPAM_SUBJECT='***SPAM*** '
SA_TAG2='6.31'
SA_TAG='2.0'
SMTP_ONLY='0'
SPAMASSASSIN_SPAM_TO_INBOX='1'
SPOOF_PROTECTION='0'
SRS_DOMAINNAME='example.com'
SRS_EXCLUDE_DOMAINS=''
SRS_SECRET=''
SRS_SENDER_CLASSES='envelope_sender'
SSL_TYPE=''
SUPERVISOR_LOGLEVEL='warn'
TLS_LEVEL='modern'
TZ=''
UPDATE_CHECK_INTERVAL='1d'
VIRUSMAILS_DELETE_DELAY='7'
[   INF   ]  Starting daemons
[  DEBUG  ]  Starting cron
[  DEBUG  ]  Starting rsyslog
[  DEBUG  ]  Starting dovecot
[  DEBUG  ]  Starting update-check
[  DEBUG  ]  Starting opendkim
[  DEBUG  ]  Starting opendmarc
[  DEBUG  ]  Starting postfix
[  DEBUG  ]  Starting amavis
[  DEBUG  ]  Starting changedetector
[   INF   ]  mail.example.com is up and running
```

N/A

polarathene:
# Description

- Log an error when the expected service state directory doesn't exist.
- The location `/var/lib/getmail/` doesn't seem like it should have been introduced. Drop it in favor of `/tmp/docker-mailserver/getmail`. It appears to be for storing remote mail that was retrieved if not configured to send to Dovecot like our docs advise. This location was never valid anyway (_as referenced issue covers_).
- Additional getmail insights are covered in a later comment of mine below.
- ~~Fetchmail debug command force enabled the feature to avoid requiring the ENV. Probably before a release when we moved the `setup` CLI into the container itself. Since the feature setup script called already gates on the feature ENV toggle and outputs a debug log when disabled... don't implicitly enable the feature.~~ (_**EDIT:** Sorry lacking time to support this unrelated change due to a test case reliance on it, reverted_)

<details>
<summary>Original message</summary>

- The Getmail service doesn't have a directory in `/var/lib/` by default, yet `mail-state` is configured to symlink it, and the service in DMS is configured to expect `/var/lib/getmail` exists.
- ~~Handled in `mail-state` logic to ensure it doesn't silently fail.~~ (_**EDIT:** Just realized this won't solve it due to the condition it's in_ 😂 )

It should probably skip the symlink in this case and omit an error to the logs? Quick fix would be to have the Getmail service startup script itself ensure the directory is configured.

</details>

Fixes #4015

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -109,7 +109,7 @@ The most noteworthy change of this release is the update of the container's base
 - Rspamd configuration: Add a missing comma in `local_networks` so that all internal IP addresses are actually considered as internal ([#3862](https://github.com/docker-mailserver/docker-mailserver/pull/3862))
 - Ensure correct SELinux security context labels for files and directories moved to the mail-state volume during setup ([#3890](https://github.com/docker-mailserver/docker-mailserver/pull/3890))
 - Use correct environment variable for fetchmail ([#3901](https://github.com/docker-mailserver/docker-mailserver/pull/3901))
-- When `ENABLE_GETMAIL=1` the expected state directory `/var/lib/getmail/` is now created for the Getmail service ([#4018](https://github.com/docker-mailserver/docker-mailserver/pull/4018))
+- When using `ENABLE_GETMAIL=1` the undocumented internal location `/var/lib/getmail/` usage has been removed. Only the config volume `/tmp/docker-mailserver/getmail/` location is supported when Getmail has not been configured to deliver mail to Dovecot as advised in the DMS docs ([#4018](https://github.com/docker-mailserver/docker-mailserver/pull/4018))
 - Dovecot dummy accounts (_virtual alias workaround for dovecot feature `ENABLE_QUOTAS=1`_) now correctly matches the home location of the user for that alias ([#3997](https://github.com/docker-mailserver/docker-mailserver/pull/3997))
 
 ## [v13.3.1](https://github.com/docker-mailserver/docker-mailserver/releases/tag/v13.3.1)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -109,6 +109,7 @@ The most noteworthy change of this release is the update of the container's base
 - Rspamd configuration: Add a missing comma in `local_networks` so that all internal IP addresses are actually considered as internal ([#3862](https://github.com/docker-mailserver/docker-mailserver/pull/3862))
 - Ensure correct SELinux security context labels for files and directories moved to the mail-state volume during setup ([#3890](https://github.com/docker-mailserver/docker-mailserver/pull/3890))
 - Use correct environment variable for fetchmail ([#3901](https://github.com/docker-mailserver/docker-mailserver/pull/3901))
+- When `ENABLE_GETMAIL=1` the expected state directory `/var/lib/getmail/` is now created for the Getmail service ([#4018](https://github.com/docker-mailserver/docker-mailserver/pull/4018))
 - Dovecot dummy accounts (_virtual alias workaround for dovecot feature `ENABLE_QUOTAS=1`_) now correctly matches the home location of the user for that alias ([#3997](https://github.com/docker-mailserver/docker-mailserver/pull/3997))
 
 ## [v13.3.1](https://github.com/docker-mailserver/docker-mailserver/releases/tag/v13.3.1)
--- a/target/bin/debug-fetchmail
+++ b/target/bin/debug-fetchmail
@@ -5,7 +5,13 @@ source /usr/local/bin/helpers/log.sh
 # shellcheck source=../scripts/startup/setup.d/fetchmail.sh
 source /usr/local/bin/setup.d/fetchmail.sh
 
-_setup_fetchmail
+# TODO: This should probably not implicitly enable the feature.
+# The setup method will feature gate and output a debug log if
+# the feature is not enabled.
+#
+# Dropping the ENV here will require updating legacy test:
+# test/tests/parallel/set3/scripts/setup_cli.bats
+ENABLE_FETCHMAIL=1 _setup_fetchmail
 
 su -s /bin/sh -c "/usr/bin/fetchmail \
   --verbose \
--- a/target/bin/debug-fetchmail
+++ b/target/bin/debug-fetchmail
@@ -5,7 +5,7 @@ source /usr/local/bin/helpers/log.sh
 # shellcheck source=../scripts/startup/setup.d/fetchmail.sh
 source /usr/local/bin/setup.d/fetchmail.sh
 
-ENABLE_FETCHMAIL=1 _setup_fetchmail
+_setup_fetchmail
 
 su -s /bin/sh -c "/usr/bin/fetchmail \
   --verbose \
--- a/target/bin/debug-getmail
+++ b/target/bin/debug-getmail
@@ -7,13 +7,7 @@ source /usr/local/bin/setup.d/getmail.sh
 
 _setup_getmail
 
-if [[ -d /var/lib/getmail ]]; then
-  GETMAILDIR=/var/lib/getmail
-else
-  mkdir -p /tmp/docker-mailserver/getmail
-  GETMAILDIR=/tmp/docker-mailserver/getmail
-fi
-
+GETMAILDIR=/tmp/docker-mailserver/getmail
 for FILE in /etc/getmailrc.d/getmailrc*; do
   getmail --getmaildir "${GETMAILDIR}" --rcfile "${FILE}" --dump | tail -n +6
 done
--- a/target/bin/getmail-cron
+++ b/target/bin/getmail-cron
@@ -1,7 +1,8 @@
 #! /bin/bash
 
+GETMAILDIR=/tmp/docker-mailserver/getmail
 for FILE in /etc/getmailrc.d/getmailrc*; do
   if ! pgrep -f "${FILE}$" &>/dev/null; then
-    getmail --getmaildir /var/lib/getmail --rcfile "${FILE}"
+    getmail --getmaildir "${GETMAILDIR}" --rcfile "${FILE}"
   fi
 done
--- a/target/scripts/startup/setup.d/getmail.sh
+++ b/target/scripts/startup/setup.d/getmail.sh
@@ -31,6 +31,10 @@ function _setup_getmail() {
 EOF
       chmod -R 600 "${GETMAILRC}"
     fi
+
+    # Both the debug command and cron job (that runs getmail) for getmail
+    # expect this location to exist. As does the `/var/mail-state` startup script.
+    mkdir -p /var/lib/getmail
   else
     _log 'debug' 'Getmail is disabled'
   fi
--- a/target/scripts/startup/setup.d/getmail.sh
+++ b/target/scripts/startup/setup.d/getmail.sh
@@ -33,8 +33,9 @@ EOF
     fi
 
     # Both the debug command and cron job (that runs getmail) for getmail
-    # expect this location to exist. As does the `/var/mail-state` startup script.
-    mkdir -p /var/lib/getmail
+    # expect this location to exist.
+    GETMAILDIR=/tmp/docker-mailserver/getmail
+    mkdir -p "${GETMAILDIR}"
   else
     _log 'debug' 'Getmail is disabled'
   fi
--- a/target/scripts/startup/setup.d/mail_state.sh
+++ b/target/scripts/startup/setup.d/mail_state.sh
@@ -23,7 +23,6 @@ function _setup_save_states() {
     [[ ${ENABLE_CLAMAV}       -eq 1 ]] && SERVICEDIRS+=('lib/clamav')
     [[ ${ENABLE_FAIL2BAN}     -eq 1 ]] && SERVICEDIRS+=('lib/fail2ban')
     [[ ${ENABLE_FETCHMAIL}    -eq 1 ]] && SERVICEDIRS+=('lib/fetchmail')
-    [[ ${ENABLE_GETMAIL}      -eq 1 ]] && SERVICEDIRS+=('lib/getmail')
     [[ ${ENABLE_MTA_STS}      -eq 1 ]] && SERVICEDIRS+=('lib/mta-sts')
     [[ ${ENABLE_POSTGREY}     -eq 1 ]] && SERVICEDIRS+=('lib/postgrey')
     [[ ${ENABLE_RSPAMD}       -eq 1 ]] && SERVICEDIRS+=('lib/rspamd')
--- a/target/scripts/startup/setup.d/mail_state.sh
+++ b/target/scripts/startup/setup.d/mail_state.sh
@@ -70,7 +70,10 @@ function _setup_save_states() {
         rm -rf "${SERVICEDIR}"
       elif [[ -d ${SERVICEDIR} ]]; then
         _log 'trace' "Moving contents of ${SERVICEDIR} to ${DEST}"
-        # Empty volume was mounted, or new content from enabling a feature ENV:
+        # An empty volume was mounted, or new content exists from enabling a feature ENV:
+        # Ensure the original directory exists before mv, otherwise with nothing to move
+        # the symlink created afterwards is invalid.
+        mkdir -p "${SERVICEDIR}"
         mv "${SERVICEDIR}" "${DEST}"
         # Apply SELinux security context to match the state directory, so access
         # is not restricted to the current running container:
--- a/target/scripts/startup/setup.d/mail_state.sh
+++ b/target/scripts/startup/setup.d/mail_state.sh
@@ -70,14 +70,13 @@ function _setup_save_states() {
         rm -rf "${SERVICEDIR}"
       elif [[ -d ${SERVICEDIR} ]]; then
         _log 'trace' "Moving contents of ${SERVICEDIR} to ${DEST}"
-        # An empty volume was mounted, or new content exists from enabling a feature ENV:
-        # Ensure the original directory exists before mv, otherwise with nothing to move
-        # the symlink created afterwards is invalid.
-        mkdir -p "${SERVICEDIR}"
+        # An empty volume was mounted, or new content dir now exists from enabling a feature ENV:
         mv "${SERVICEDIR}" "${DEST}"
         # Apply SELinux security context to match the state directory, so access
         # is not restricted to the current running container:
         chcon -R --reference="${STATEDIR}" "${DEST}" 2>/dev/null || true
+      else
+        _log 'error' "${SERVICEDIR} should exist but is missing"
       fi
 
       # Symlink the original path in the container ($SERVICEDIR) to be


georglauterbach:
Existing tests are failing:

```txt
not ok 30 [Getmail] debug-getmail works as expected in 565ms
# (from function `assert_line' in file test/test_helper/bats-assert/src/assert_line.bash, line 219,
#  in test file test/tests/parallel/set1/getmail.bats, line 58)
#   `assert_line --regexp 'retriever:.*SimpleIMAPSSLRetriever\(ca_certs="None", certfile="None", getmaildir="\/tmp\/docker-mailserver\/getmail", imap_on_delete="None", imap_search="None", keyfile="None", mailboxes="\(.*INBOX.*\)", move_on_delete="None", ***, password_command="\(\)", port="993", record_mailbox="True", server="imap.remote-service.test", ssl_cert_hostname="None", ssl_ciphers="None", ssl_fingerprints="\(\)", ssl_version="None", timeout="180", use_cram_md5="False", use_kerberos="False", use_peek="True", use_xoauth2="False", username="user3"\)'' failed
#
# -- no output line matches regular expression --
# regexp : retriever:.*SimpleIMAPSSLRetriever\(ca_certs="None", certfile="None", getmaildir="\/tmp\/docker-mailserver\/getmail", imap_on_delete="None", imap_search="None", keyfile="None", mailboxes="\(.*INBOX.*\)", move_on_delete="None", ***, password_command="\(\)", port="993", record_mailbox="True", server="imap.remote-service.test", ssl_cert_hostname="None", ssl_ciphers="None", ssl_fingerprints="\(\)", ssl_version="None", timeout="180", use_cram_md5="False", use_kerberos="False", use_peek="True", use_xoauth2="False", username="user3"\)
# output (22 lines):
#     retriever:  SimpleIMAPSSLRetriever(ca_certs="None", certfile="None", getmaildir="/var/lib/getmail", imap_on_delete="None", imap_search="None", keyfile="None", mailboxes="('INBOX',)", move_on_delete="None", ***, password_command="()", port="993", record_mailbox="True", server="imap.remote-service.test", ssl_cert_hostname="None", ssl_ciphers="None", ssl_fingerprints="()", ssl_version="None", timeout="180", use_cram_md5="False", use_kerberos="False", use_peek="True", use_xoauth2="False", username="user3")
#     destination:  MDA_external(allow_root_commands="True", arguments="('-d', 'user3@example.test')", command="deliver", group="None", ignore_stderr="False", path="/usr/lib/dovecot/deliver", pipe_stdout="True", unixfrom="False", user="None")
#     options:
#       delete : False
#       delete_after : 0
#       delete_bigger_than : 0
#       delivered_to : False
#       fingerprint : False
#       logfile : logfile(filename="/var/log/mail/getmail-user3.log")
#       max_bytes_per_session : 0
#       max_message_size : 0
#       max_messages_per_session : 500
#       message_log : /var/log/mail/getmail-user3.log
#       message_log_syslog : False
#       message_log_verbose : False
#       netrc_file : None
#       read_all : False
#       received : False
#       skip_imap_fetch_size : False
#       to_oldmail_on_each_mail : False
#       use_netrc : False
#       verbose : 0
# --
#
```

I have no idea why the EC lint is failing, though..

polarathene:
**TL;DR:** Seems that `getmail` should not be configured for `/var/mail-state` and this got missed during review as it was likened to `fetchmail` which is setup differently as a service (_although it may technically not need a state dir either for all I know_ 🤷‍♂️ )

---

> Existing tests are failing

This is due to the logic here:

https://github.com/docker-mailserver/docker-mailserver/blob/006f442cd0bdedb386acf328146b630887a88f84/target/bin/debug-getmail#L10-L15

Previously since the internal directory never existed in the first place (_except I guess as a symlink to nothing_) it would fallback to the config volume and configure for that instead...

I guess we missed that during review 🤷‍♂️ It shouldn't store data at locations for different concerns? (_runtime state that's usually discardable vs config volume that is generally used to persist anything else_)



---

The cronjob (_for runtime usage_) has the `/var/lib/getmail` location hard coded (_whereas the debug `setup` command above offers the alternative `/tmp/docker-mailserver` path_):

https://github.com/docker-mailserver/docker-mailserver/blob/006f442cd0bdedb386acf328146b630887a88f84/target/bin/getmail-cron#L3-L7

---

## More context

https://getmail6.org/configuration.html#running-commandline-options

> To use getmail, simply run the script getmail, which is typically installed in `/usr/local/bin/` by default.
> getmail will read the default getmail rc file (`getmailrc`) from the default configuration/data directory (`~/.getmail/`) and begin operating.

> `--getmaildir=DIR` or `-gDIR` — use `DIR` for configuration and data files

> `--rcfile=FILE` or `-rFILE` — read getmail rc file `FILE` instead of the default.
> - The file path is assumed to be relative to the getmaildir directory unless this value starts with a slash (`/`).
> - This option can be given multiple times to have getmail retrieve mail from multiple accounts.

Unlike fetchmail, there is no daemon service or polling, thus the cronjob:
- https://getmail6.org/faq.html#faq-how-daemon
- https://wiki.archlinux.org/title/Getmail#Fetching_mail_automatically_with_systemd

RC files are provided via the config volume and copied over to internal location in `/etc`:

https://github.com/docker-mailserver/docker-mailserver/blob/006f442cd0bdedb386acf328146b630887a88f84/target/scripts/startup/setup.d/getmail.sh#L14-L16

These all have a base config (_which our docs advise mounting directly over instead of supporting via the config volume_) but to be practical with DMS needs to configure to deliver over to Dovecot:

https://github.com/docker-mailserver/docker-mailserver/blob/006f442cd0bdedb386acf328146b630887a88f84/test/config/getmail/getmail-user3.cf#L1-L11

I'm not sure why this even needs to be in DMS beyond convenience, there isn't anything specific about it to integrate from the looks of it. Although that's probably the same case with fetchmail. Although [there's this interesting FAQ entry](https://getmail6.org/faq.html#faq-about-why) from the original getmail author criticizing fetchmail.

---

`getmaildir` in this case is a location to dump the retrieved remote mail if you haven't configured it to deliver to Dovecot (opt-in). We don't have anything that seems to suggest using `/tmp/docker-mailserver/getmail` other than in the debug script itself (_and test case matching the debug output_).

So any existing users of the feature either have had mail delivered to Dovecot as per guidance in the docs, or it's been stored in `/var/mail-state/lib-getmail` (_but we don't document this expectation, and the associated volume isn't really intended for it as it's not runtime state related_). The `getmail` command is run as root UID/GID AFAIK? There's really no value for it being handled in `/var/mail-state` then.

georglauterbach (approved):

mazzz1y:
Hi @polarathene,

I'm attempting to switch from fetchmail to getmail and have a question:

Why use `/tmp/docker-mailserver/getmail`? 

This directory is used by getmail to store information about emails that have already been downloaded.  This is because it retrieves messages based on the server's message ID rather than the 'unread' flag of the message as fetchmail doing. So, this directory serves a "data" purpose, and I believe it should not be located inside the config directory, which is often stored in a Git repository.

https://getmail6.org/faq.html#faq-about-oldmail

polarathene:
You are asking if this location should instead be in `/var/mail-state`? We tend to treat that more for run-time throwaway data, although I think @georglauterbach has redis use it for rspamd now too.

Most data there is intended for `/var/run` IIRC which has data like the Postfix mail queue, you rarely need to persist that as data you'd migrate to another system, but it is useful for crash recovery so that the queue remains available if you had a power loss event or kernel panic for example.

I can understand that the Config volume is probably not appropriate either. DMS hasn't introduced any other volumes for a long time, this one is usually used for persisting data that isn't logs or `/var/mail`. Most of the time that's config (hence the name), but the location `/tmp/docker-mailserver` isn't really reflective of it being config only 😅 

I've been thinking an overhaul will need to be considered in future. In the meantime, you are able to volume mount a separate host location over the `/tmp/docker-mailserver/getmail` path and that should work just fine?

---

Perhaps @casperklein or @georglauterbach can weigh in with their opinion. From what @mazzz1y describes the getmail data here is state, but it's state you'd want to persist I don't think it's intended for `/var/run` like other services.

So like the redis database, it might be better if it were stored at a separate location intended for persisting data. Presently that purpose overlaps with the `/tmp/docker-mailserver` container path, we just refer to this as the Config Volume.

Open to suggestions, but such changes would need to be queued for v15 as it'd be a breaking change.

georglauterbach:
I think it'd be best to store it in `/var/mail-state`, but we probably need a separate discussion what should go where.

mazzz1y:
> You are asking if this location should instead be in /var/mail-state?
> I think it'd be best to store it in /var/mail-state, but we probably need a separate discussion what should go where.

Yes! This is definitely what is needed. Getmail stores its own **state** there, and I think /var/mail-**state** is intended for exactly this purpose (at least by name).

>  In the meantime, you are able to volume mount a separate host location over the /tmp/docker-mailserver/getmail path and that should work just fine?

I'm already doing this, but it doesn't look nice :)

```yaml
    volumes:
      - ./data/maildata:/var/mail
      - ./data/getmail:/tmp/docker-mailserver/getmail
      - ./config/docker-mailserver:/tmp/docker-mailserver
```

Also, this method results in Docker creating an empty ./config/docker-mailserver/getmail directory on the host. This happens because the mounting action inside the Docker container requires the creation of an empty directory at /tmp/docker-mailserver/getmail before mounting. It doesn't look good, but it works, so the issue is not a significant one and is more of an aesthetic concern

I will create a separate issue

casperklein:
> I've been thinking an overhaul will need to be considered in future.

I might migrate from fetchmail to getmail in 6 weeks. I will then take a closer look at the getmail implementation and possibly improve it, e.g. making it a service that is handled by supervisord instead of the current cron approach.

Regarding `/var/mail-state`, I think that makes sense.

polarathene:
> e.g. making it a service that is handled by supervisord instead of the current cron approach.

Discussion should probably continue at the new issue for this, so I'll be brief.

I recall `getmail` lacking a daemon, it functions differently from `fetchmail`, hence the reliance on a cron timer to mimic the polling behaviour of fetchmail. (_**EDIT:** I covered this in my PR description under "More Context" section_)

PR #4035, opened 2024-05-25
scripts: update log format

georglauterbach:
# Description

Over time, I really came to despise the log format I implemented. I recently updated the log for another project of mine, and really wanted to get this into v14.0.0 (as potentially breaking).

The new format is more akin to Postfix's current format: `<ISO 8601 TIMESTAMP>  <LOG LEVEL>  <WHO LOGGED>: <MESSAGE>`. Having a timestamp and information about who actually logged is important; hence I included it.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,6 +70,8 @@ The most noteworthy change of this release is the update of the container's base
   - `supervisor-app.conf` renamed to `dms-services.conf`
 - **Rspamd**:
   - the Redis history key has been changed in order to not incorporate the hostname of the container (which is desirable in Kubernetes environments) ([#3927](https://github.com/docker-mailserver/docker-mailserver/pull/3927))
+- **Log**:
+  - the log level format was updated; the new format is `<ISO 8601 TIMESTAMP>  <LOG LEVEL>  <WHO LOGGED>: <MESSAGE>`
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,7 +44,7 @@ The most noteworthy change of this release is the update of the container's base
     - This was a community contributed feature for FTS (Full Text Search), the docs advise using an image that has not been maintained for over 2 years and lacks ARM64 support. Based on user engagement over the years this feature has very niche value to continue to support, thus is being removed.
     - If you use Solr, support can be restored if you're willing to contribute docs for the feature that resolves the concerns raised
 - **Log**:
-  - the log level format was updated; the new format is `<ISO 8601 TIMESTAMP> <LOG LEVEL> <WHO LOGGED>: <MESSAGE>`
+  - The format of DMS specific logs (_from our scripts, not running services_) has been changed. The new format is `<RFC 3339 TIMESTAMP> <LOG LEVEL> <LOG EVENT SRC>: <MESSAGE>` ([#4035](https://github.com/docker-mailserver/docker-mailserver/pull/4035))
 - **rsyslog:**
   - Debian 12 adjusted the `rsyslog` configuration for the default file template from `RSYSLOG_TraditionalFileFormat` to `RSYSLOG_FileFormat` (_upstream default since 2012_). This change may affect you if you have any monitoring / analysis of log output (_eg: `mail.log` / `docker logs`_).
     - The two formats are roughly equivalent to [RFC 3164](https://www.rfc-editor.org/rfc/rfc3164)) and [RFC 5424](https://datatracker.ietf.org/doc/html/rfc5424#section-1) respectively.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,8 @@ The most noteworthy change of this release is the update of the container's base
   - **Removed support for Solr integration:** ([#4025](https://github.com/docker-mailserver/docker-mailserver/pull/4025))
     - This was a community contributed feature for FTS (Full Text Search), the docs advise using an image that has not been maintained for over 2 years and lacks ARM64 support. Based on user engagement over the years this feature has very niche value to continue to support, thus is being removed.
     - If you use Solr, support can be restored if you're willing to contribute docs for the feature that resolves the concerns raised
+- **Log**:
+  - the log level format was updated; the new format is `<ISO 8601 TIMESTAMP> <LOG LEVEL> <WHO LOGGED>: <MESSAGE>`
 - **rsyslog:**
   - Debian 12 adjusted the `rsyslog` configuration for the default file template from `RSYSLOG_TraditionalFileFormat` to `RSYSLOG_FileFormat` (_upstream default since 2012_). This change may affect you if you have any monitoring / analysis of log output (_eg: `mail.log` / `docker logs`_).
     - The two formats are roughly equivalent to [RFC 3164](https://www.rfc-editor.org/rfc/rfc3164)) and [RFC 5424](https://datatracker.ietf.org/doc/html/rfc5424#section-1) respectively.
@@ -70,8 +72,6 @@ The most noteworthy change of this release is the update of the container's base
   - `supervisor-app.conf` renamed to `dms-services.conf`
 - **Rspamd**:
   - the Redis history key has been changed in order to not incorporate the hostname of the container (which is desirable in Kubernetes environments) ([#3927](https://github.com/docker-mailserver/docker-mailserver/pull/3927))
-- **Log**:
-  - the log level format was updated; the new format is `<ISO 8601 TIMESTAMP>  <LOG LEVEL>  <WHO LOGGED>: <MESSAGE>`
 - **Account Management**
   - addresses (accounts) are now normalized to lowercase automatically and a warning is logged in case uppercase letters are supplied
 
--- a/target/scripts/helpers/log.sh
+++ b/target/scripts/helpers/log.sh
@@ -100,7 +100,7 @@ function _log() {
       ;;
   esac
 
-  MESSAGE="$(date --iso-8601='seconds') ${!LOG_COLOR}${LOG_LEVEL_NAME}${RESET} $(basename "${0}"): ${2}"
+  MESSAGE="$(date --rfc-3339='seconds') ${!LOG_COLOR}${LOG_LEVEL_NAME}${RESET} $(basename "${0}"): ${2}"
 
   if [[ ${1} =~ ^(warn|error)$ ]]; then
     echo -e "${MESSAGE}" >&2
--- a/target/scripts/helpers/log.sh
+++ b/target/scripts/helpers/log.sh
@@ -79,12 +79,14 @@ function _log() {
     ( 'info' )
       [[ ${LEVEL_AS_INT} -ge 3 ]] || return 0
       LOG_COLOR='BLUE'
+      # the whitespace is intentional (for alignment purposes)
       LOG_LEVEL_NAME='INFO '
       ;;
 
     ( 'warn' )
       [[ ${LEVEL_AS_INT} -ge 2 ]] || return 0
       LOG_COLOR='LYELLOW'
+      # the whitespace is intentional (for alignment purposes)
       LOG_LEVEL_NAME='WARN '
       ;;
 
--- a/target/scripts/helpers/log.sh
+++ b/target/scripts/helpers/log.sh
@@ -53,8 +53,7 @@ function _log() {
     return 1
   fi
 
-  local LEVEL_AS_INT
-  local MESSAGE="${RESET}["
+  local LEVEL_AS_INT LOG_COLOR LOG_LEVEL_NAME MESSAGE
 
   case "$(_get_log_level_or_default)" in
     ( 'trace' ) LEVEL_AS_INT=5 ;;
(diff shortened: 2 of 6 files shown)


polarathene:
> Having a timestamp and information about who actually logged is important; hence I included it.

This information should be available with syslog IIRC, I forget the CLI command, but it's possible to call that to log a message that `rsyslog` should pick up and handle like it does other syslog compatible services like Postfix. That's where the actual timestamp and context should be coming from AFAIK?

Likewise, for stdout, Docker (and I assume k8s) should have their own logging drivers at least for timestamp tracking. Perhaps next month I'll finally get back to the Vector PR work upstream which will make the process a bit nicer for us 😅

target/scripts/helpers/log.sh
```diff
@@ -67,35 +66,42 @@ function _log() {
...
 
     ( * )
       _log 'error' "Call to '_log' with invalid log level argument '${1}'"
       return 1
       ;;
   esac
 
-  MESSAGE+="${RESET}]  ${2}"
+  MESSAGE="$(date --iso-8601='seconds')  ${!LOG_COLOR}${LOG_LEVEL_NAME}${RESET}  "
+  MESSAGE+="$(basename "${0}" || echo 'unknown'): ${2}"
```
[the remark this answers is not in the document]
polarathene: Technically the date should be `date --rfc-3339='seconds'` (_`ns` would be more correct, but I'm not sure how you restrict the precision to 6 digits which is expected, ns precision is optional though_)

polarathene (dismissed):
`logger` is the command, [here is a nice overview](https://www.cyberciti.biz/tips/howto-linux-unix-write-to-syslog.html) (_with CLI options shown at the bottom_).

Syslog has two widespread formats, RFC 3164 (DMS prior to v14), and RFC 5424 (DMS v14 onwards default), where most should be using the proper standard RFC 5424. The priority value represents the combined `facility` (_most services in DMS should be using `mail` IIRC_) **and** the `severity` (_your log levels_).

---

Here is some stripped down reference from my Vector PR:

<details>
<summary>Snippets (click to expand)</summary>

```rs
enum Severity {
    Emergency = 0,
    Alert = 1,
    Critical = 2,
    Error = 3,
    Warning = 4,
    Notice = 5,
    #[default]
    Informational = 6,
    Debug = 7,
}
```

```rs
enum Facility {
    Kern = 0,
    #[default]
    User = 1,
    Mail = 2,
    Daemon = 3,
    Auth = 4,
    Syslog = 5,
    LPR = 6,
    News = 7,
    UUCP = 8,
    Cron = 9,
    AuthPriv = 10,
    FTP = 11,
    NTP = 12,
    Security = 13,
    Console = 14,
    SolarisCron = 15,
    Local0 = 16,
    Local1 = 17,
    Local2 = 18,
    Local3 = 19,
    Local4 = 20,
    Local5 = 21,
    Local6 = 22,
    Local7 = 23,
}
```

The `Local*` facilities are for custom/local usage if desired. The syslog service (_`rsyslog` for us presently_) can parse that to handle differently like it does for filtering mail logs to files, or forwarding to another log service if configured to do so.

Calculating the priority value looks like this:
```rs
// Example (Priority value for `Facility::Mail` + `Severity::Error` is `19`):
// `(2 * 8) + 3` or `(2 << 3) + 3` (bitwise shift instead of multiplication)
let priority = (self.facility as u8 * 8) + self.severity as u8;
```

</details>

Nevermind that, `logger --priority mail.error` is supported as a more friendly way for you to use, no math needed 😅 (_use the relevant supported log levels from the collapsed snippet though_)

**Examples:**
```console
# RFC 5424 format (notq omits the structured data regarding time, msgid is rfc5424 only):
logger --rfc5424=notq --priority mail.error --id=1337 --msgid=hello --tag dms-changedetector "This is an example error message!"
<19>1 2024-05-26T11:45:28.305480+00:00 mail.example.test dms-changedetector 1337 hello - This is an example error message!

# Old RFC 3164 format (lacks a year, assumes present year):
logger --rfc3164 --priority mail.error --id=1337 --tag dms-changedetector "This is an example error message!"
<19>May 26 11:45:27 mail dms-changedetector[1337]: This is an example error message!
```

- Use `--stderr` with the above `logger` command examples to get the output to stderr.
- If you don't use `notq`, you'll get some metadata for a log ingester to have about quality of the timestamps (`[timeQuality tzKnown="1" isSynced="1" syncAccuracy="157"]`), this is Structured Data from RFC5424, doesn't really exist for prior RFC 3164.
- `--tag` is the service name like `amavis` or `postfix`. Despite the linked resource examples, do not use spaces, that makes the log ambiguous to parse. The RFC document has a breakdown for supported tokens and expected sequence format.
- `--id` isn't really needed, ~~it's the PID by default~~ `logger` omits it, PID of the `logger` command as default requires `--id`. Note that `--id 42` is not valid, that's `--id` then `42` as part of the message rather than as part of `--id`, make sure assignments for the option have `=` 😑 (_`-i` seems to be equivalent to `--id` too?_)
- `--rfc5424` is not being parsed properly by `rsyslog` in `:edge`, stick to `--rfc3164`

---

**UPDATE:** Nevermind. I thought `logger` would be appropriate, but regardless of syslog format used, `rsyslog` is not stripping away some of the data like expected:

```
# RFC 3164 followed by RFC 5424

# v13.3:
May 26 11:51:14 mail mail dms-changedetector[1337]: This is an example error message!
May 26 11:55:07 mail 1 2024-05-26T11:55:07.509942+00:00 mail.example.test dms-changedetector 1337 - - This is an example error message!

# v14
2024-05-26T12:05:58.700336+00:00 mail mail dms-changedetector[1337]: This is an example error message!
2024-05-26T12:17:07.824094+00:00 mail 1 2024-05-26T12:17:07.824035+00:00 mail.example.test dms-changedetector 1337 hello - This is an example error message!
```

Pretty sure on a related PR I discussed this and documented the `rsyslog` config. Perhaps we should address that before pushing out v14 to use the syslog format? (_I was under the impression it was effectively the same as the two but omitted the priority value, and for RFC 5424 the syslog version, eg `<19>1`_)

Seems like I made a note of this on the `rsyslog` changelog entry actually 🤔 

---

So just to clarify, your change here won't exactly line up to `rsyslog` format
- You are using a two space delimiter.
- You skip the host.
- Your timestamp won't match.
- You add your own log level (but so do other services with their message, so.._ 🤷‍♂️ )

Personally I prefer the current DMS log output which is more easier to distinguish if it's not going to align with `rsyslog` for log parsing (_at least with the current log output, you know that if the line starts with `[` it's probably emitted by DMS, otherwise if a valid timestamp is the starting portion it's more than likely from `rsyslog` from one of the services we run. Anything else is usually stderr or similar and could be filtered separately.

Approving this change if you still feel you want to go through with it as-is, but you should wait for @casperklein to provide input too.

target/scripts/helpers/log.sh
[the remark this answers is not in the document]
polarathene: > Do you have an example of what this looks like? (_yes I'm that lazy_ 😄 )

image

Not so bad, if someone was trying to parse these logs and separate them, they'll need to go based on two columns to match as DMS logs now.

Still the added verbosity of the timestamp (_and a bit to the script name, but that's a little helpful_), I'm not sure about. Adjusting the `rsyslog` config won't adjust the timestamp format here since this isn't being managed via `rsyslog`. If we had some consistency one could better parse the logs and filter/format them as they see fit.

georglauterbach:
@polarathene I adjusted the format to now look like this: `<ISO 8601 TIMESTAMP> <LOG LEVEL> <WHO LOGGED>: <MESSAGE>` (see my recent commits). The double-space has disappeared. I actually never tried to be rsyslog-compatible, but if you and @casperklein think it'd be a good idea, I can make that happen.

I would still like to go through with this, but I will wait for @casperklein's input on this as well, just as you requested 👍🏼

target/scripts/helpers/log.sh
```diff
@@ -67,35 +66,41 @@ function _log() {
...
+      ;;
 
     ( * )
       _log 'error' "Call to '_log' with invalid log level argument '${1}'"
       return 1
       ;;
   esac
 
-  MESSAGE+="${RESET}]  ${2}"
+  MESSAGE="$(date --iso-8601='seconds') ${!LOG_COLOR}${LOG_LEVEL_NAME}${RESET} $(basename "${0}"): ${2}"
```
polarathene, line 103:
```suggestion
  MESSAGE="$(date --rfc-3339='seconds') ${!LOG_COLOR}${LOG_LEVEL_NAME}${RESET} $(basename "${0}"): ${2}"
```

target/scripts/helpers/log.sh
```diff
@@ -67,35 +66,41 @@ function _log() {
...
-      MESSAGE+="   ${BLUE}INF   "
+      LOG_COLOR='BLUE'
+      LOG_LEVEL_NAME='INFO '
       ;;
 
     ( 'warn' )
       [[ ${LEVEL_AS_INT} -ge 2 ]] || return 0
-      MESSAGE+=" ${LYELLOW}WARNING "
+      LOG_COLOR='LYELLOW'
+      LOG_LEVEL_NAME='WARN '
```
polarathene, line 88:
```suggestion
      LOG_LEVEL_NAME='WARN'
```

target/scripts/helpers/log.sh
```diff
@@ -67,35 +66,41 @@ function _log() {
...
-      MESSAGE+="  ${PURPLE}DEBUG  "
+      LOG_COLOR='PURPLE'
+      LOG_LEVEL_NAME='DEBUG'
       ;;
 
     ( 'info' )
       [[ ${LEVEL_AS_INT} -ge 3 ]] || return 0
-      MESSAGE+="   ${BLUE}INF   "
+      LOG_COLOR='BLUE'
+      LOG_LEVEL_NAME='INFO '
```
polarathene, line 82:
```suggestion
      LOG_LEVEL_NAME='INFO'
```

CHANGELOG.md
```diff
@@ -43,6 +43,8 @@ The most noteworthy change of this release is the update of the container's base
   - **Removed support for Solr integration:** ([#4025](https://github.com/docker-mailserver/docker-mailserver/pull/4025))
     - This was a community contributed feature for FTS (Full Text Search), the docs advise using an image that has not been maintained for over 2 years and lacks ARM64 support. Based on user engagement over the years this feature has very niche value to continue to support, thus is being removed.
     - If you use Solr, support can be restored if you're willing to contribute docs for the feature that resolves the concerns raised
+- **Log**:
+  - the log level format was updated; the new format is `<ISO 8601 TIMESTAMP> <LOG LEVEL> <WHO LOGGED>: <MESSAGE>`
```
polarathene, line 47:
```suggestion
  - The format of DMS specific logs (_from our scripts, not running services_) has been changed. The new format is `<RFC 3339 TIMESTAMP> <LOG LEVEL> <LOG EVENT SRC>: <MESSAGE>` ([#4035](https://github.com/docker-mailserver/docker-mailserver/pull/4035))
```

polarathene (changes requested):
> I actually never tried to be rsyslog-compatible, but if you and @casperklein think it'd be a good idea, I can make that happen.

I think it could be nice, but proper approach is to use `logger` as described above, which will require messing with rsyslog config to properly ingest.

As I want to replace `rsyslog` with Vector in future I don't know if it's worth your time to integrate with `rsyslog`, but if you can make it work that'd be cool 👍

target/scripts/helpers/log.sh
casperklein:
```
# date --iso-8601='seconds'
2024-05-27T22:00:56+02:00

# date --rfc-3339='seconds'
2024-05-27 22:01:18+02:00
```

casperklein:
After @polarathene changes are applied, LGTM 👍

target/scripts/helpers/log.sh
polarathene: > ```console
> $ date --rfc-3339='seconds'
> 2024-05-27 22:01:18+02:00
> ```

Oh good catch! It should be able to support a `T` instead of space delimiter, but I guess `date` doesn't do that 🤷‍♂️ (_`logger` does as shown earlier, along with fractional secs at 6 digit precision_)

I just looked up RFC 5424 and it's actually requiring a specific format from RFC 3339 timestamps 😅

It seems RFC 3339 is usually chosen over ISO 8601 for [when software isn't intended to support the full range of ISO 8601 formats](https://www.reddit.com/r/ISO8601/comments/p572xy/rfc_3339_versus_iso_8601/), since most are only interested in the [intersection of the two standards](https://www.reddit.com/r/ISO8601/comments/v0903s/when_people_refer_to_the_iso8601_date_format_they/).

---

https://www.rfc-editor.org/rfc/rfc5424.html#section-6.2.3

> image

> image

---

The [two standards differ a bit](https://ijmacd.github.io/rfc3339-iso8601/) but they both have the preferred format as `%Y-%M-%DT%h:%m:%s.%u%Z:%z` (_not a compatible `date` format_), the `date` command doesn't seem to support the fractional sec though.

The visual doesn't show the `:%z` suffix variant for UTC offset, but it's roughly equivalent to the highlighted one:

image

casperklein: > Oh good catch! It should be able to support a T instead of space delimiter

I liked the change, BECAUSE of the missing T 😆

target/scripts/helpers/log.sh
georglauterbach: The space is intentional to have alignment after the log level.

target/scripts/helpers/log.sh
georglauterbach: The space is intentional to have alignment after the log level.

georglauterbach:
> After @polarathene suggestions are applied, LGTM 👍

I applied two of the 4; the ones I did not apply are related to the space; I want the space there to have the log level always aligned to five characters.

georglauterbach:
@polarathene when we transition to Vector, all efforts of integrating our logs into Rsyslog would be wasted, hence I'd like to stick to what's currently on the branch and then convert to Vector once we're ready.

polarathene:
> I want the space there to have the log level always aligned to five characters.

Maybe add a comment so others know that context? My first thought was there was trailing white-space, especially for diff in a review that didn't occur to me 😓

polarathene (dismissed):

polarathene (approved):

PR #4039, opened 2024-05-27
docs(k8s): Advise `externalTrafficPolicy: Local` if no PROXY protocol configured

Amphaal:
If not using `externalTrafficPolicy: Local` in service declaration, all incoming requests will most likely be branded as proxied by the `externalIp` by which the mail services are accessed; this is undesired. As a side effect, any `externalIp` will eventually get banned by Fail2Ban, which helped my diagnosis this misconfiguration.

I have no idea how well this enforced parameter plays along with "PROXY protocol" configuration.

### DIFF
--- a/docs/content/config/advanced/kubernetes.md
+++ b/docs/content/config/advanced/kubernetes.md
@@ -140,6 +140,7 @@ If using our Helm chart is not viable for you, here is some guidance to start wi
                 app: mailserver
 
             spec:
+              externalTrafficPolicy: Local # most likely required in all scenarii, else every incoming request would be identified as the External IP, which will get banned by Fail2Ban eventually 
               type: LoadBalancer
 
               selector:
--- a/docs/content/config/advanced/kubernetes.md
+++ b/docs/content/config/advanced/kubernetes.md
@@ -140,7 +140,9 @@ If using our Helm chart is not viable for you, here is some guidance to start wi
                 app: mailserver
 
             spec:
-              externalTrafficPolicy: Local # most likely required in all scenarii, else every incoming request would be identified as the External IP, which will get banned by Fail2Ban eventually 
+              # `Local` is most likely required, otherwise every incoming request would be identified by the external IP,
+              # which will get banned by Fail2Ban when monitored services are not configured for PROXY protocol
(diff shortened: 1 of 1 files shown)


docs/content/config/advanced/kubernetes.md
```diff
@@ -140,6 +140,7 @@ If using our Helm chart is not viable for you, here is some guidance to start wi
                 app: mailserver
 
             spec:
+              externalTrafficPolicy: Local # most likely required in all scenarii, else every incoming request would be identified as the External IP, which will get banned by Fail2Ban eventually
```
casperklein, line 143:
```suggestion
              externalTrafficPolicy: Local # most likely required in all scenarii, else every incoming request would be identified as the external IP address, which will get banned by Fail2Ban eventually 
```

polarathene:
```suggestion
              # `Local` is most likely required, otherwise every incoming request would be identified by the external IP,
              # which will get banned by Fail2Ban when monitored services are not configured for PROXY protocol
              externalTrafficPolicy: Local
```

- Shift comment to above field, split to two lines.
- Drop `scenarii` (_that doesn't seem to be a common use for what I assume is plural equivalent of scenarios?_)
- Added additional context/hint for the concern being related to lack of PROXY protocol usage.

I am not familiar with k8s to review further as I don't know how `Local` fixes the concern here. Or if "external IP" is referring to ingress/lb or real client IP.

Since PROXY protocol is usually required for k8s from what @georglauterbach has said in the past IIRC, I assume `Local` doesn't solve the concern across the board with external traffic?🤷‍♂️

polarathene (changes requested):

georglauterbach:
This is a welcome contribution, and something I missed. [`externalTrafficPolicy`](https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/#preserving-the-client-source-ip) with `Local` is correct; with PROXY protocol it does not matter though. I think the comment from @polarathene makes that clear. I am fine with the change here.

georglauterbach (approved):

Amphaal:
LGTM, thanks guys !

Issue #2819, opened 2022-10-09
[TODO]: `SPOOF_PROTECTION=1` feature and Alias support improvements


polarathene:
Adding this here in case anyone else has time to tackle any of it.

I am not likely to have the time until at least December, but I expect I'll have to prioritize my time on other commitments before I can come back to addressing these tasks.

## Task Overview

I've provided plenty of helpful details below, but a summary of tasks to complete is:

**`SPOOF_PROTECTION=1` feature:**

- [ ] [Only bother with the `unionmap` and drop the conditional](https://github.com/docker-mailserver/docker-mailserver/blob/ff969509f8b9c49e399327887c9a29f6fd664b80/target/scripts/startup/setup-stack.sh#L471-L476), as the file always exists, there is no known reason to need the alternative without unionmap or additional regex table.
- [ ] In addition, add support for another table to that `unionmap`, which is not a regex table. It will allow explicitly allowing additional senders for an account to send as.
  - [ ] Once supported, update the docs workaround example covered in `@gmail` related issue section below.
- [ ] Consider if `unionmap` would also be relevant for the `ldap` equivalent value in that same linked method.
- [ ] Consider improving user docs (or eventual maintainer docs) to better present the relevant information for this feature covered below. Potentially provide an example with the `noreply@external-domain` from referenced send-only related issue below. Likewise with the LDAP specific config ENV that was added to support the feature specifically there.
- [ ] Review the mentioned/linked concerns below for enabling the feature by default in a future release, and decide if it should be default once better docs and config support is merged. The ENV name may also need reconsideration (also discussed below).

**Related alias support:**
- [ ]  `setup alias add` section below notes a bug to resolve, and CLI help message to update. It also has poor error handling feedback if some `setup` commands do not receive the expected number of inputs IIRC (_may have been `setup alias del`_).
- [ ] In the final section below _"Alias domains being added to `/etc/postfix/vhost`"_, it has been identified that we could improve how we determine what domains are managed.
  - Alias domains should be handled better and this support should be easier to exclude domains via using Postfix `main.cf` settings more appropriately.
  - Keep in mind `/etc/postfix/vhost` is presently sourced elsewhere, such as the DKIM generation support in `setup` command.

---

## `SPOOF_PROTECTION=1` handling for `smtpd_sender_login_maps`

[...]

Issue #3824, opened 2024-01-25
question: Why do I get the error `Sender address rejected: not owned by user`? (`reject_authenticated_sender_login_mismatch`)


cfis:
### 📝 Preliminary Checks

- [X] I tried searching for an existing issue and followed the [debugging docs](https://docker-mailserver.github.io/docker-mailserver/latest/config/debugging/) advice, but still need assistance.

### 👀 What Happened?

This is more of a question about relay server setup than a bug.

I am testing docker-mailserver on a home machine. It uses a postfix server on aws (mail2.example.com) as a relay. Since port 25 is blocked by Comcast, communication is done over port 587. In particular:

mail2@example.com ------ 587  --- > mail@example.com

`mail2` forwards to `mail` using the user `relay@example.com`  which has been added to postfix-accounts.cf (and the password is known to mail2 via sasl_password).

Assume an email is received by `mail2`, say for myself at `cfis@example.com`, and is forwarded to `mail`. That results in this error in the log:

```
docker-mailserver-59cc9899d4-lfnbn postfix/submission/smtpd[3078]: NOQUEUE: reject: RCPT from mail2.example.com[x.x.x.x]: 553 5.7.1 <root@mail2.example.com>: Sender address rejected: not owned by user relay@example.com; from=<root@mail2.example.com> to=<cfis@example.com> proto=ESMTP helo=<mail2.example.com>
```
This is caused by the inclusion of `reject_authenticated_sender_login_mismatch` in `mua_sender_restrictions`.

If I override `mua_sender_restrictions` to remove `reject_authenticated_sender_login_mismatch`, then the relay user delivers the email to the correct account without error.

However, is there a better way to set this up? I want ports 465 and 587 to require authentication, so `mail2` has to login as some user. And port 25 is blocked by comcast. I figured this would be a common issue, but an internet search revealed people with the same issue but without clear guidance on the best solution.

Thanks!

### 🐋 DMS Version

v13.3.1

### 💻 Operating System and Architecture

Fedora

### ⚙️ Container configuration files

```yml
This is using Kubernetes, see https://github.com/cfis/docker-mailserver-helm
```

polarathene:
[...]

PR #4044, opened 2024-06-02
feat: `SPOOF_PROTECTION=1` support config to relax sender restrictions by login

### LINKED ISSUE #4042: feature request: allow sender spoofing for certain accounts
### Context

There are many cases where `SPOOF_PROTECTION` should be enabled for some accounts, and not for others. For example, you may have standard user accounts where you want it enabled, however, there may be a system/application account which you'd like to be able to send as any address (for example, a single application which you want to allow sending as `no-reply@foo.com` and `no-reply@bar.com`, or other addresses like `orders@foo.com`, `invoices@foo.com`, etc.).

### Description

I would like a way to selectively enable/disable spoofing per account. I understand that this is not likely something easily accomplished via environment variables, and adding dedicated setup scripts/etc would be a huge undertaking, but I would at least like to document a solution so others can easily implement this (here on GitHub or preferably on the web docs).

https://github.com/Mailu/Mailu/issues/1096#issuecomment-518931372 seems like a similar issue.

### Alternatives

- Creating accounts for each user and having the application log into each seperately
  - This is going to be an increasingly painful maintenance burden and result in a mess as an application grows, and would be totally unsuited for cases with dynamic email addresses

### Applicable Users

Administrators of software which sends emails on behalf of multiple system accounts.

I would appreciate some guidance, but hopefully a solution that can be documented for others running into the same issue.

ncovercash:
# Description

Currently, the only way to allow a user to send mail as another user (for example, to allow `admin@example.com` to send as `user@example.com`) would be to:
- disable `SPOOF_PROTECTION` (undesirable as it would allow any user to spoof any other user)
- alias `user@example.com` to `admin@example.com` (undesirable as `user@example.com`'s mail would be sent to the `admin` mailbox)
- manually edit configurations/map files into the container (not very user-friendly)

To improve this experience, this PR adds a new configuration file, `postfix-regexp-send-only.cf`, which allows configuring send-only aliases. These aliases differ from regular aliases in that they do not affect incoming mail, only outgoing, enabling admin/system accounts to send as other users, and with all the granular access control which regex can offer.

Some use cases could include:
- Allowing an administrator to send as other user accounts, particularly to aid debugging of mail delivery issues for a certain user/domain
- Permitting system/service accounts to send as a user
- ...probably some more that I cannot think of off the top of my head.

Fixes #4042, relates to #2819

## Type of change

- [x] This change requires a documentation update (also included here)

## Checklist

## Questions

- [ ] Which set inside `test/tests/parallel` should my tests go in? I did not see any that related to postfix/auth, so I created a folder inside `set2` as it seemed to be the smallest test set; please let me know how this should be organized to best fit in with the rest of the tests.
  - As part of these changes, I was able to consolidate the current spoofing tests from the main `tests.bats` into this separate parallel file, since they cover the same cases.


polarathene:
> * alias `user@example.com` to `admin@example.com` (undesirable as `user@example.com`'s mail would be sent to the `admin` mailbox)

We forbid aliasing an actual account. So that is presently not possible. Sharing an alias to multiple recipients is though (_not officially endorsed/supported by DMS however_).

> * manually edit configurations/map files into the container (not very user-friendly)

Isn't that the same as what this PR proposes by introducing a new regexp config to manage which sender addresses a user login (DMS account / address) can use?

Or you are referring to only the additional integration in `main.cf`? Which could be handled via `user-patches.sh`:

```bash
#!/bin/bash

# Include `regexp-send-only` to relax `SPOOF_PROTECTION=1` sender address restrictions for specific logins:
cp -f /tmp/docker-mailserver/postfix-regexp-send-only.cf /etc/postfix/regexp-send-only
postconf 'smtpd_sender_login_maps = unionmap:{ texthash:/etc/postfix/virtual, hash:/etc/aliases, pcre:/etc/postfix/maps/sender_login_maps.pcre, pcre:/etc/postfix/regexp-send-only }'
```

---

## Initial Review / Concerns

I'm a little confused, while there is value in this PR, ~~are you not already repeating the functionality I made you aware of in https://github.com/docker-mailserver/docker-mailserver/issues/4042#issuecomment-2143687135 ?~~ (_**EDIT:** PR adds separate config for send-only support_)

- [`sender_login_maps.pcre`](https://github.com/docker-mailserver/docker-mailserver/blob/ff969509f8b9c49e399327887c9a29f6fd664b80/target/postfix/sender_login_maps.pcre) is effectively the `SPOOF_PROTECTION=1` feature (_by using a regex lookup that restricts sender address to that of the DMS account_).
- Prepending regex patterns above that line would give you similar functionality?
- This is not only for aliases, but any address you allow will be permitted as a sender address, including ones DMS has no authority over (`support@github.com`).
- A [separate regexp config `postfix-regexp.cf`](https://github.com/docker-mailserver/docker-mailserver/issues/3824#issuecomment-1909796346) has been used / advised in the past (_with the caveat that it is shared with virtual aliases lookup_), it is also [used in our test-suite for `SPOOF_PROTECTION=1`](https://github.com/docker-mailserver/docker-mailserver/blob/ff969509f8b9c49e399327887c9a29f6fd664b80/test/config/postfix-regexp.cf):
  > image
- Using a separate regex table should be ok, but it was noted that avoiding regex was the preferred choice to support (_**you made this same mistake** in your config example_):
  > image
- It may be appropriate to continue using the existing `postfix-regexp.cf` file for this, and drop the support for virtual aliases with it, as the actual need for users to have regexp virtual aliases may be low? It was only added AFAIK to workaround `SPOOF_PROTECTION=1` issue before that was fixed via using `unionmap` instead? (_although I am not opposed to keeping regex support for both via separate files_):
  > image

docs/content/config/user-management.md
```diff
@@ -82,5 +82,19 @@ Postfix supports so-called address tags, in the form of plus (+) tags - i.e. `ad
...
+Sometimes, it may be useful to allow certain accounts to send as other accounts, even when [the `SPOOF_PROTECTION` environment variable is enabled][spoof-protection]. This may be used to allow services to send accounts as other addresses/users without needing to disable spoof protection entirely, and without affecting incoming mail (which traditional [aliases](#aliases) would do).
+
+To configure these aliases, add them to `docker-data/dms/config/postfix-regexp-send-only.cf` in the same format as [the other regexp aliases](#configuring-regexp-aliases). For example:
+
+```cf
+/^.*@example.com$/ admin@example.com
+/^.*$/ superadmin@example.com
+```
+
+In this example, `admin@example.com` would be able to send as any address at `example.com` and `superadmin@example.com` would be able to send as any address at any domain.
```
polarathene, line 96: `superadmin@example.com` will not be able to send as `user@example.com` AFAIK, this is a caveat with regex lookup tables that you need to be aware of. The 1st match from top-down on the sender address will then check to the right-side value(s) for permitted logins (_which are often email addresses in DMS, but they could just be user names_).

You did not escape `.` for `.com` in the regex, in this example it shouldn't be a real issue, since alternative characters aren't going to produce a valid sender address, but it's still a caveat to be careful of:

`jan.doe@example.com` would also grant `janedoe@example.com` sender address to the login account, which might be a security problem when using the feature for accounts that aren't admins.

target/scripts/helpers/aliases.sh
```diff
@@ -30,6 +30,17 @@ function _handle_postfix_regexp_config() {
   fi
 }
 
+function _handle_postfix_regexp_send_only_config() {
+  : >/etc/postfix/regexp-send-only
```
polarathene, line 34: I'd prefer not creating a file explicitly when it's not used.

target/scripts/startup/setup.d/security/spoofing.sh
```diff
@@ -14,7 +14,11 @@ function _setup_spoof_protection() {
       # NOTE: This file is always created at startup, it potentially has content added.
       # TODO: From section: "SPOOF_PROTECTION=1 handling for smtpd_sender_login_maps"
       # https://github.com/docker-mailserver/docker-mailserver/issues/2819#issue-1402114383
-      if [[ -f /etc/postfix/regexp ]]; then
+      if [[ -f /etc/postfix/regexp && -f /etc/postfix/regexp-send-only ]]; then
+        postconf 'smtpd_sender_login_maps = unionmap:{ texthash:/etc/postfix/virtual, hash:/etc/aliases, pcre:/etc/postfix/maps/sender_login_maps.pcre, pcre:/etc/postfix/regexp, pcre:/etc/postfix/regexp-send-only }'
+      elif [[ -f /etc/postfix/regexp-send-only ]]; then
+        postconf 'smtpd_sender_login_maps = unionmap:{ texthash:/etc/postfix/virtual, hash:/etc/aliases, pcre:/etc/postfix/maps/sender_login_maps.pcre, pcre:/etc/postfix/regexp-send-only }'
+      elif [[ -f /etc/postfix/regexp ]]; then
```
polarathene, line 21: Not acceptable. See the related `SPOOF_PROTECTION=1` issue.

This section needs to be refactored to unify with LDAP. You'll need to wait on that being resolved. I am adding friction here due to blockers I introduce external to this PR, I'll need to address those first.

polarathene (changes requested):
- I need to push out a rewrite docs that affects `user-management.md`. Your change will need to be adjusted to that afterwards. I'll try get that done for you by this weekend.
- Likewise the `CHANGELOG.md` items will need to be rebased after the v14 release. This is too late to land in that and will have to be delayed to 14.1.
- I've not reviewed the test contribution yet, I have seen your questions but will address that closer to resolving this PR as I'm a bit short on time and this is touching an area of functionality that has known issues already.
- `unionmap` logic in `spoofing.sh` won't be accepted. I think my existing PR has concerns related to this area with LDAP, so that may need to be addressed prior. All you're really doing here is appending a string of additional files, which could be handled via variable.

A good contribution overall, I was only expecting docs so I apologize about any blockers I'm responsible for. I'll see what I can do to get those out of the way.

ncovercash:
> Isn't that the same as what this PR proposes by introducing a new regexp config to manage which sender addresses a user login (DMS account / address) can use?

Yes; the primary purpose of this was to have a more "DMS-friendly" way that would 1) be supported in the documentation and 2) not require the user directly patching the main.cf.  Sure, this could all be accomplished by a user adding their patches, but theoretically [so is most of the config](https://pullrequest-4044--dms-doc-previews.netlify.app/config/advanced/optional-config/#files), so it seems like it'd be good for DMS to support/expose what it can?

> Prepending regex patterns above that line would give you similar functionality?

Sure; I can edit the commands to prepend the user-provided configuration above this and use the same file; I personally felt it was a little cleaner to have separate files for this and the user-provided ones. Your judgement there, just let me know what would be best.

> This is not only for aliases, but any address you allow will be permitted as a sender address

Yes; I can rewrite the added documentation if there's a better way to describe it (maybe "permitted sender addresses"?). Let me know if there's a name you would prefer.

> A [separate regexp config postfix-regexp.cf](https://github.com/docker-mailserver/docker-mailserver/issues/3824#issuecomment-1909796346) has been used / advised in the past (with the caveat that it is shared with virtual aliases lookup), it is also [used in our test-suite for SPOOF_PROTECTION=1](https://github.com/docker-mailserver/docker-mailserver/blob/ff969509f8b9c49e399327887c9a29f6fd664b80/test/config/postfix-regexp.cf)

This feature does not overlap, since `postfix-regexp.cf` will also redirect incoming mail. This specifically is for outgoing mail only. And yes, I saw the test — I moved it over [into my test suite](https://github.com/docker-mailserver/docker-mailserver/pull/4044/files#diff-c892b7fc84cd6b15efe36e51b079d426ae4fe1e99666f8d3c59f417c39a8c4dbR44-R70).

> Using a separate regex table should be ok, but it was noted that avoiding regex was the preferred choice to support (you made this same mistake in your config example)

Good catch; I can change to a non-regex version, I just wanted to align closely with the existing `postfix-regexp.cf`

> It may be appropriate to continue using the existing postfix-regexp.cf file for this, and drop the support for virtual aliases with it

I personally use `postfix-regexp.cf` for the virtual alias support ☹️. I use this to redirect `postmaster`, `webmaster`, and `hostmaster` to one centralized inbox (woo, wearing many hats), as well as a few other things, so it'd be an unfortunate feature to lose.

target/scripts/helpers/aliases.sh
ncovercash: I was just copying the regular postfix-regexp logic here — would I be OK to also remove the same thing in the `_handle_postfix_regexp_config()` above (and corresponding `# TODO: Investigate why this file is always created, nothing seems to append only the cp below?` comment)?

docs/content/config/user-management.md
ncovercash: Good catch on both counts; I'll clarify that in the documentation and warn about the caveats.

It looks like the following would allow `superadmin@example.com` to get the desired effect:
```suggestion
```cf
/^.*@example.com$/ admin@example.com superadmin@example.com
/^.*$/ superadmin@example.com
```

In this example, `admin@example.com` would be able to send as any address at `example.com` and `superadmin@example.com` would be able to send as any address at any domain.

...some notes about needing to order/etc
```

ncovercash:
> A good contribution overall, I was only expecting docs so I apologize about any blockers I'm responsible for. I'll see what I can do to get those out of the way.

I'm happy to hold off, I did not realize there were blockers. I have this temporarily worked around on my local mailserver, but I wanted to make a better experience for those who came after me. And no worries, I thought this would be a doc-only thing and then I realized it was (relatively) easy to add first-class support and figured may as well.

No rush at all, let me know when I need to update my changes to align with yours. Thanks for the review!

target/scripts/helpers/aliases.sh
polarathene: There's probably a related issue already open about details for that.

Without looking for that, I can't quite say. I've not had as much time for DMS as I'd like, but if the change was delayed I may have detailed why. I'd leave it alone for now, unless you have time to sink 😅 

Proper actionable feedback from me is going to need to wait until I have the time to refresh on related context like this. I'll try to get that sorted for you by this weekend if possible, sorry about the friction that adds.

polarathene:
> I personally felt it was a little cleaner to have separate files for this and the user-provided ones. Your judgement there, just let me know what would be best.

AFAIK `unionmap` will query all config files for the sender address, and return any results of valid logins that can use it.

It's probably best that we keep [`sender_login_maps.pcre`](https://github.com/docker-mailserver/docker-mailserver/blob/ff969509f8b9c49e399327887c9a29f6fd664b80/target/postfix/sender_login_maps.pcre) as-is and don't modify it then. It can remain as an internal config as you suggest, using separate user config.

---

> Good catch; I can change to a non-regex version, I just wanted to align closely with the existing `postfix-regexp.cf`

I think it would be simpler to document and support a non-regex table. If some users request regex support, then we could add that as another file.

> > It may be appropriate to continue using the existing postfix-regexp.cf file for this, and drop the support for virtual aliases with it
> 
> I personally use `postfix-regexp.cf` for the virtual alias support ☹️. I use this to redirect `postmaster`, `webmaster`, and `hostmaster` to one centralized inbox (woo, wearing many hats), as well as a few other things, so it'd be an unfortunate feature to lose.

👍 No worries, we can have both then :)

---

> I'm happy to hold off, I did not realize there were blockers. 
> No rush at all, let me know when I need to update my changes to align with yours. Thanks for the review!

Thanks ❤️ I really want to resolve the LDAP PR and related docs rewrite this month.

The LDAP test also touches on `SPOOF_PROTECTION=1` but there's a bug IIRC that was being resolved, which is probably why the test cases in `tests.bats` hadn't yet been migrated as you've tackled in this PR. I was probably meaning to handle that afterwards.

After that I can better go over the related `SPOOF_PROTECTION=1` issue again, and help push this forward. If the LDAP PR does become an actual blocker this PR would be delayed until v15.

So uhh, if you're comfortable with it, perhaps best to just wait until later this month before taking any further action.

ncovercash:
Sure thing, no worries! @ me when you're ready to proceed!

eadoking:
Hi @ncovercash, 

I just have some security concerns about the admin@example.com/ superadmin@example.com send email as user@example.com 

- what if for example it was a targeted attack by the admin? not for good purpose for testing email !!
- will the send message appears on the sent folder on the user@example.com ?
- can the admin/superadmin by any means delete the sent message from user@example.com ?
- can the admin/superadmin by any means read the respond for the user@example.com ?

polarathene:
> * what if for example it was a targeted attack by the admin? not for good purpose for testing email !!

They're an admin, it would not be surprising for them to have elevated permissions to do things that could be abused. That concern is unrelated to this feature.

---

> * will the send message appears on the sent folder on the `user@example.com`?

No. Without `SPOOF_PROTECTION=1`, you can send mail as any sender address you like. You could even try `ceo@github.com`, it will be rejected if sent outbound to any actual MTA with standard security checks since you don't actually own `github.com`.

The PR here is just to allow relaxing what additional sender addresses can be used per user. A sent folder is something your mail client keeps on it's end AFAIK (_unless your mailbox in `/var/mail` says otherwise?_).

---

> * can the admin/superadmin by any means delete the sent message from `user@example.com`?

Not with this feature. If the admin has control of the DMS deployment they can do whatever they want with the container really.

AFAIK, your mail client submits mail to DMS when you want to send mail out, that will temporarily be stored in a mail queue, and then is sent outbound to an external mail server or delivered to a mailbox within DMS. So if the sent copy is only kept with your mail client storage (like Thunderbird), it can't be touched.

If it's in `/var/mail`, this feature won't enable deletion. This feature is for Postfix service, whereas mailbox storage is handled by the Dovecot service. A related feature is Dovecot Master Users, which allows an admin to login as any user without needing that users password. If that admin had access to the DMS container, they could also read the mail content directly at `/var/mail` without any user login needed.

---

> * can the admin/superadmin by any means read the respond for the `user@example.com`?

Technically yes. If your response is submitted to DMS, Postfix will receive it and it'll end up in the queue. It could be monitored and read there if that is the only time it is available in transit. That requires actual access to the DMS container, not just a DMS user account (Postfix/Dovecot login).

Otherwise same caveats regarding deleting a sent copy as for being able to read one.

---

There is an on-disk encryption feature for `/var/mail` content that I will eventually contribute that with support for per-user keys. Then none of the concerns above really apply for access to mail content. Unless they have the ability to read memory or something else to read unencrypted content arriving inbound (delivery) or your credentials after TLS is terminated... mail queue concern would still be valid AFAIK as it's only for encrypting mail stored to `/var/mail`.

You can have your mail client use PGP encryption so long as the recipient also has this configuration to decrypt the mail, and then the mail is always encrypted between systems, DMS won't know the contents or be able to access it then.

georglauterbach:
I have to chime in here. I am currently cleaning up the repo from a bunch of backlog. What is the current state of this PR? Can it be rebased and feedback be applied? If not, I don't mind either way - I'd just like to see work being continued or the PR being closed.

Issue #4007, opened 2024-05-07
Rspamd rejects `asciinema` e-mails


jiriks74:
### 📝 Preliminary Checks

- [X] I tried searching for an existing issue and followed the [debugging docs](https://docker-mailserver.github.io/docker-mailserver/latest/config/debugging/) advice, but still need assistance.

### 👀 What Happened?

I wanted to log into [asciinema](https://asciinema.org/) but didn't get the login email. When I used another one it worked fine.

When looking into logs I see that the message was outrights rejected instead of being moved to the spam folder:

<details>
<summary>Logs</summary>

```
May  7 23:04:43 mx postfix/smtpd[717267]: 0437429007C5: client=unknown[64.147.123.154]

May  7 23:04:43 mx postfix/cleanup[717285]: 0437429007C5: message-id=<>

May  7 23:04:43 mx postfix/cleanup[717285]: 0437429007C5: milter-reject: END-OF-MESSAGE from unknown[64.147.123.154]: 5.7.1 Spam message rejected; from=<hello@hello.asciinema.org> to=<my@email> proto=ESMTP helo=<wfhigh3-smtp.messagingengine.com>

May  7 23:04:43 mx postfix/smtpd[717267]: disconnect from unknown[64.147.123.154] ehlo=2 starttls=1 mail=1 rcpt=1 data=0/1 quit=1 commands=6/7
```

</details>

### 👟 Reproduction Steps

1. Use rspamd
2. Try to log into [asciinema](https://asciinema.org/)
3. See no email in inbox or junk
4. See that the server rejected it outright in the logs

### 🐋 DMS Version

v13.3.1

### 💻 Operating System and Architecture

Debian GNU/Linux 11 (bullseye) aarch64

### ⚙️ Container configuration files

```yml
services:
  mailserver:
    image: ghcr.io/docker-mailserver/docker-mailserver:latest
    container_name: mailserver
    hostname: mx.example.com
    env_file: stack.env
    environment:
      SSL_TYPE: letsencrypt
      SSL_DOMAIN: mx.example.com
    ports:
      - "25:25"     # SMTP  (explicit TLS => STARTTLS, Authentication is DISABLED => use port 465/587 instead)
      - "143:143"   # IMAP4 (explicit TLS => STARTTLS)
      - "465:465"   # ESMTP (implicit TLS)
      - "587:587"   # ESMTP (explicit TLS => STARTTLS)
[...]

PR #4064, opened 2024-06-13
add sasldb authentication as another way to secure mailserver usage

edgd1er:
# Description

Add authentification to smtp when not using ldap, shadow or rimap.
sasldb is a plugin for cyrus, so i'm afraid not compatible with dovecot.

there is many operations to execute to have sasldb authentication. The idea is to implement more easily this feature.
The use case is a postfix as relay available on local lan, not limited to a docker network.
an env var was added as remote postfix will have rules that detect as a spammer, the container sending to quickly mails.
default value is the current one: 0s (no throttle)

setup.sh was not changed.
SMTP_USERNAMES and SMTP_PASSWORDS env vars are required as list of users /password to create.
sasldb entries and email accounts will be provisioned.

this PR add a feature and does not fix a bug.
all lints tests are ok.
I failed to implement and run bats tests.
I'm not an postfix expert so side effects may occur.
I will be adding  a new section to the docs: SASLDB authentication if you think it's needed

## Type of change

## Checklist

### DIFF
--- a/docs/content/config/environment.md
+++ b/docs/content/config/environment.md
@@ -921,6 +921,7 @@ Note: This postgrey setting needs `ENABLE_POSTGREY=1`
 - `shadow` => authenticate against local user db
 - `mysql` => authenticate against mysql db
 - `rimap` => authenticate against imap server
+- `cyrus` => authenticate aganinst sasl2db
 - NOTE: can be a list of mechanisms like pam ldap shadow
 
 ##### SASLAUTHD_MECH_OPTIONS
--- a/mailserver.env
+++ b/mailserver.env
@@ -536,6 +536,7 @@ ENABLE_SASLAUTHD=0
 # `shadow` => authenticate against local user db
 # `mysql` => authenticate against mysql db
 # `rimap` => authenticate against imap server
+# `cyrus` => authenticate aganinst sasl2db
 # Note: can be a list of mechanisms like pam ldap shadow
 SASLAUTHD_MECHANISMS=
 
--- a/target/scripts/startup/daemons-stack.sh
+++ b/target/scripts/startup/daemons-stack.sh
@@ -45,7 +45,10 @@ function _start_daemon_rsyslog        { _default_start_daemon 'rsyslog'        ;
 function _start_daemon_update_check   { _default_start_daemon 'update-check'   ; }
 
 function _start_daemon_saslauthd() {
-  _default_start_daemon "saslauthd_${SASLAUTHD_MECHANISMS}"
+  #if one of ldap|shadow|mysql|rimap , start the daemon
+  if [[ ${SASLAUTHD_MECHANISMS} =~ (ldap|shadow|mysql|rimap) ]]; then
+    _default_start_daemon "saslauthd_${SASLAUTHD_MECHANISMS}"
+  fi
 }
 
 function _start_daemon_postfix() {
--- a/target/scripts/startup/setup.d/postfix.sh
+++ b/target/scripts/startup/setup.d/postfix.sh
@@ -19,12 +19,9 @@ function _setup_postfix_early() {
     postconf "inet_protocols = ${POSTFIX_INET_PROTOCOLS}"
   fi
 
-  __postfix__log 'trace' "Configuring SASLauthd"
-  if [[ ${ENABLE_SASLAUTHD} -eq 1 ]] && [[ ! -f /etc/postfix/sasl/smtpd.conf ]]; then
-    cat >/etc/postfix/sasl/smtpd.conf << EOF
-pwcheck_method: saslauthd
-mech_list: plain login
-EOF
+  if [[ -n ${DEFAULT_DESTINATION_RATE_DELAY} ]]; then
+    __postfix__log 'trace' "Configure destination rate to ${DEFAULT_DESTINATION_RATE_DELAY}"
+    postconf "default_destination_rate_delay = ${DEFAULT_DESTINATION_RATE_DELAY}"
   fi
 
   # User has explicitly requested to disable SASL auth:
--- a/target/scripts/startup/setup.d/saslauthd.sh
+++ b/target/scripts/startup/setup.d/saslauthd.sh
@@ -1,17 +1,63 @@
 #!/bin/bash
 
 function _setup_saslauthd() {
-  _log 'debug' 'Setting up SASLAUTHD'
 
-  # NOTE: It's unlikely this file would already exist,
-  # Unlike Dovecot/Postfix LDAP support, this file has no ENV replacement
-  # nor does it copy from the DMS config volume to this internal location.
-  if [[ ${ACCOUNT_PROVISIONER} == 'LDAP' ]] \
-  && [[ ! -f /etc/saslauthd.conf ]]; then
-    _log 'trace' 'Creating /etc/saslauthd.conf'
+  __postfix__log 'trace' "Configuring SASLauthd cyrus if requested"
+  gpasswd -a postfix sasl >/dev/null
+  # cyrus uses a plugin not a saslauth service.
+  if [[ "${SASLAUTHD_MECHANISMS:-''}" == cyrus ]]; then
+    _log 'debug' 'Setting up SASLAUTHD for cyrus'
+    CPATH="/etc/postfix/sasl"
+    CFILE="smtpd"
 
-    # Create a config based on ENV
-    sed '/^.*: $/d'> /etc/saslauthd.conf << EOF
+    #smtpd_sasl_auth_enable / smtpd_sasl_type / smtpd_sasl_path = private/auth smtpd_sasl_security_options = noanonymous, noplaintext
+    #smtpd_sasl_tls_security_options = noanonymous
+    __postfix__log 'trace' 'Setting up cyrus smtp auth'
+    postconf "smtpd_sasl_auth_enable = yes"
+    # Cyrus SASL configuration file name: smtpd.conf
+    postconf "smtpd_sasl_path = ${CFILE}"
+    #cyrus/dovecot auth for smtp.
+    postconf "smtpd_sasl_type = cyrus"
+    #change master.cf as postconf -e is not editing it.
+    sed -i -n "s[^  -o smtpd_sasl_type=.*|^  -o smtpd_sasl_type=\${smtpd_sasl_type}/g" /etc/postfix/master.cf
+
+    # location where Cyrus SASL searches
+    postconf "cyrus_sasl_config_path = ${CPATH}"
+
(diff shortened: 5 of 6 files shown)


polarathene:
> The use case is a postfix as relay available on local lan, not limited to a docker network.

I would like a better justification and example for why you need this support upstreamed into DMS.

You should have raised a feature request to discuss first before tackling a PR that may get rejected. As a maintenance policy, we are reluctant to accept niche contributions without strong value add or user interest.

---

- You are not only configuring `cyrus`, but with SASLAuthD. Is there a reason you chose to use that too?
- You are relying on users to be created by two new ENV? Why?

What is wrong with the default support? You can create users with `setup email add user@example.com password_here`, and then rely on the config volume mount to persist the `postfix-accounts.cf` file that provides the same functionality via Dovecot.

edgd1er:
Thanks for taking the time to evaluate my proposition. 

> > The use case is a postfix as relay available on local lan, not limited to a docker network.
> 
> I would like a better justification and example for why you need this support upstreamed into DMS.

use case:
* I dont have a domain, nor a mx record for the moment, so all mails will be relayed.
* I need to send emails from different servers, all running containers representing 3 different environnements. If needed containers are using a bridge network. containers have an ip and a port on my local network. PERMIT_DOCKER has values to allows connections from host interface but without authentication. 
* I'm afraid to host an open relay which would cause a serious problem to the destination address as all mails are relayed. I wish to have emails relayed with local authentication (dovecot or cyrus, no requirement at the moment): PERMIT_DOCKER=none => Explicitly force authentication
* shadow, rimap, ldap and mysql authentication are available as far as I understood, but I would like to keep it as simple as possible.

> 
> You should have raised a feature request to discuss first before tackling a PR that may get rejected. As a maintenance policy, we are reluctant to accept niche contributions without strong value add or user interest.
I'm sorry I didn't start a discussion first, if you think this feature need to be discussed and more precisely defined, feel free to close the PR.

> 
>     * You are not only configuring `cyrus`, but with SASLAuthD. Is there a reason you chose to use that too?
> 
>     * You are relying on users to be created by two new ENV? Why?
> 
> 
> What is wrong with the default support? You can create users with `setup email add user@example.com password_here`, and then rely on the config volume mount to persist the `postfix-accounts.cf` file that provides the same functionality via Dovecot.

As for the changes suggested, they are the result of try/error process.
Authentication with dovecot should be possible but I stumble on a virtual error message. As said, i'm not an postfix expert and may be there is a better solution with dovecot. Is dovecot more 
Authentication with mysql, rimap, shadow is done through SASLAuthD daemon. Cyrus has a plugin to handle sasldb, no daemon is required.  
I though implementing cyrus sasldb preparation at the same stage would limit required change to scripts.

Creating two new env might be a decision to review. At the moment, when the container starts, users are added to the accounts cf file and to sasldb. it's needed once, not all the time. That part could be moved to the setup.sh I guess.

polarathene:
> I dont have a domain, nor a mx record for the moment, so all mails will be relayed.

Neither has any relevance to DMS accounts.

Although the username is `user@example.com`, this is actually login username that can use any sender address (_except for `SPOOF_PROTECTION=1`, which restricts it this username email address_).

---

> I need to send emails from different servers, all running containers representing 3 different environments.
> If needed containers are using a bridge network. containers have an ip and a port on my local network.
> `PERMIT_DOCKER` has values to allows connections from host interface but without authentication.

You should ideally always use authentication.

Mail should be submitted on ports 465 or 587 with your DMS account credentials. This will allow you to submit mail for DMS to send locally or outbound (directly or via a relay service).

It should be rare that you need to rely on `PERMIT_DOCKER` to send mail. You should especially be cautious with that setting if you trust a subnet, since some misconfigurations will allow third-parties to appear in the same subnet via the Docker network gateway IP. Not requiring authentication will enable spammers to abuse your DMS (_for example, they connect via IPv6 to your IPv4 only DMS container with Docker defaults, which now lets them be trusted via `PERMIT_DOCKER` on a private subnet_).

---

> I'm afraid to host an open relay which would cause a serious problem to the destination address as all mails are relayed.
> I wish to have emails relayed with local authentication (dovecot or cyrus, no requirement at the moment):
> `PERMIT_DOCKER=none` => Explicitly force authentication

Yes, that is better. `PERMIT_DOCKER=none` is the default for this reason.

---

> shadow, rimap, ldap and mysql authentication are available as far as I understood, but I would like to keep it as simple as possible.

SASLAuthd is an alternative authentication service than what we provide by default (_Postfix talks straight to Dovecot, Dovecot manages the accounts_). This was a feature contributed by a community member interested in LDAP without Dovecot active. They had a separate service outside of DMS for their account management.

`rimap` was added as part of that, allowing for remote/external Dovecot with SASLAuthd. But the primary focus of the SASLAuthd was `ldap` mechanism. Others have no official support in DMS.

Keeping it simple would be our default with `postfix-accounts.cf`. You just use the `setup email add user@example.com my_password` command to the container to create an account. I think this is where you had a misunderstanding?

---

> Authentication with dovecot should be possible but I stumble on a virtual error message.

I think this is your main issue and motivation for PR as you tried to investigate and fix it with a different approach?

DMS can be run completely locally and offline, the accounts work without needing to own a domain or have any MX record configured.

MX records in DNS are only for domains to say where mail should be delivered to for that domain. This would be a server you run DMS on, or any other service that you'd have receive mail instead. DMS can still send mail or relay it through another service (like SendGrid).

When you configure DMS with `hostname`, this would be what MX records would point to if mail is to be delivered to DMS. It is also the FQDN that should have a TLS certificate, not individual mail domains DMS manages.

If DMS only needs to send mail outbound, then you need DNS records like SPF for a domain to permit the DMS FQDN / IP to send on that domains behalf. If you instead use a relay service like SendGrid, DMS will authenticate with that service, and that relay service must instead have authority to send as that domain.

polarathene:
Closing this PR as no response from author and the PR itself seems misguided?

If you disagree, feel free to respond to my previous message, but it seems that this PR was an attempted workaround due to a misunderstanding with configuration (_it would be better to clarify that so that our docs could be improved instead_).

edgd1er:
I've been trying to use authentication with dovecot without success for the moment.
Problems are not related with dms, but with certificates and rights on ntfs mounted filesystems.
 but I understand the closure of the PR as this feature should already possible without any modification of the project.

polarathene:
> but with certificates and rights on ntfs mounted filesystems.

We do not support NTFS. Use Docker named volumes if you are not able to mount a linux filesystem via bind volume mounts.

It doesn't make much sense to try mount on a Windows disk, is this only for local testing on a Windows desktop machine? Consider WSL2, it works well and you can still browse the filesystem from Windows Explorer.

I don't think NTFS is compatible with preserving filesystem attributes, let alone UID + GID values? These are important.

---

As for the certificate concern, ensure the certificate is for your DMS `hostname`, your actual DMS email accounts have no relevance to the certificates for this part.

If you are having trouble provisioning certificates, assuming this is a local Windows desktop, that could be for a variety of reasons. LetsEncrypt will try to access your system via ports 80 and 443, if you have not forwarded those, or have some other network layer in the way (Firewall, NAT, CG-NAT, VPN, etc), it'll be a bit problematic.

You can try a Wildcard certificate which depends on your DNS provider. Cloudflare can be a good choice, but your ISP may not provide a stable static IP address which may complicate that a bit (_this may also cause you additional problems with sending mail btw, it's often blocked_). A wildcard cert will otherwise be easier to provision.

Alternatively use a cheap VPS and create a VPN connection to it (eg with TailScale) so that both systems are on the same network / subnet, and then route the external public traffic from the VPS to your local system (_this is a bit of a hassle_).

If you are provisioning your own local certificates (`openssl` / StepCA, etc), that should work fine too. You should sign them with your own root CA, and ensure any clients that need to connect to DMS have that root CA public key in their trust stores. Configuring this in DMS is a bit awkward, but it is supported.

Local self-hosting is a bit more troublesome, especially for email with the DNS requirements. You'll probably want to relay mail through a service like SendGrid for better deliverability to external mail services. Perhaps deploy to a cheap VPS first to get a working deployment, then adopt to a local deployment? You'll still want to prefer a Linux environment for Docker (such as WSL2) and avoid mounting to NTFS or other filesystems that are lacking (FAT32, exFAT I think?).

edgd1er:
thanks for the detailed comments.
you are right, ntfs is not a good idea when it comes to linux. 
For development I mainly use linux but I sometimes boot to windows. I had to keep a partition of my drive that can be read by both OSes, hence the NTFS.
I had previously on other docker projects problems with mounted volumes, due to ntfs.
the certificate I use is self-signed with SAN and wildcards and tests with sasldb cyrus feature enabled are successful.

Errors I have, come from certificate/tls related to the mail client or dms:
* mailserver  | 2024-06-18T00:11:03.588398+02:00 mail postfix/submissions/smtpd[5876]: warning: key at index 1 in /etc/dms/tls/mail.mission.lan-key.pem does not match next certificate
mailserver  | 2024-06-18T00:11:03.588410+02:00 mail postfix/submissions/smtpd[5876]: warning: TLS library problem: error:0A000120:SSL routines::private key mismatch:../ssl/ssl_rsa.c:977:
mailserver  | 2024-06-18T00:11:03.588417+02:00 mail postfix/submissions/smtpd[5876]: warning: error loading private keys and certificates from: /etc/dms/tls/mail.mission.lan-key.pem /etc/dms/tls/mail.mission.lan-cert.pem: disabling TLS support

msmtp logs:
Handshake failure. The TLS connection was non-properly terminated.' exitcode=EX_PROTOCOL

My problems are not related to dms and I'll have to fix these problems to be able to use dovecot with authentication and ssl.
Thanks for maintaining DMS and your dedication.

polarathene:
> Errors I have, come from certificate/tls related to the mail client or dms:

Make sure you have provisioned your certificate correctly.

I have documented how we generate certificate for our test suite [here](https://github.com/docker-mailserver/docker-mailserver/tree/master/test/files/ssl/example.test). At that link you will also find that I have provisioned self-signed certs along with CA root signed, both RSA and ECDSA types, along with wildcard.

They will not expire until 2030 IIRC, so you could try those files if your DNS / `hostname` can route for `mail.example.test`, or provision your own by following the information there.

See also our TLS docs page:
- Verify TLS cert is working correctly: https://docker-mailserver.github.io/docker-mailserver/latest/config/security/ssl/#testing-a-certificate-is-valid
- Self-signed guidance (alternative to my test suite link): https://docker-mailserver.github.io/docker-mailserver/latest/config/security/ssl/#self-signed-certificates

For guidance for local hosting with custom DNS (CoreDNS), I have some examples I reference here: https://github.com/docker-mailserver/docker-mailserver/issues/4007#issuecomment-2103639320

---

> Thanks for maintaining DMS and your dedication.

You are welcome! Best of luck resolving those issues 💪

PR #4118, opened 2024-07-17
tests: Update `curl` note for XOAUTH2 support

polarathene:
# Description

The test case will not be able to run until Debian provides a release of curl from 2024. Better visibility and references documented for maintenance.

## Type of change

### DIFF
--- a/test/tests/serial/mail_with_oauth2.bats
+++ b/test/tests/serial/mail_with_oauth2.bats
@@ -59,9 +59,14 @@ function teardown_file() {
 }
 
 @test "should authenticate with XOAUTH2" {
-  # curl packaged in Debian 12 (and the latest release as of Jan 2024) broke XOAUTH2 support
+  # curl 7.80.0 (Nov 2021) broke XOAUTH2 support (DMS v14 release with Debian 12 packages curl 7.88.1)
   # https://github.com/docker-mailserver/docker-mailserver/pull/3403#issuecomment-1907100624
-  skip 'unable to test XOAUTH mechanism due to bug since curl 7.80'
+  #
+  # Fixed in curl 8.6.0 (Jan 31 2024):
+  # - https://github.com/curl/curl/issues/10259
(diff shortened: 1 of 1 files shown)


polarathene:
Unexpected test failure:

```
not ok 74 [setup.sh] show usage when no arguments provided in 107ms
#  in test file test/tests/parallel/set3/scripts/setup_cli.bats, line 30)
#   `assert_success' failed
#
# -- command failed --
# status : 1
# output : Error response from daemon: No such container: dms-test_postconf-helper
# --
#
```

https://github.com/docker-mailserver/docker-mailserver/blob/34423c2f665b8c7839c40fbc7d1fe8cda6342456/test/tests/parallel/set3/scripts/setup_cli.bats#L6

https://github.com/docker-mailserver/docker-mailserver/blob/34423c2f665b8c7839c40fbc7d1fe8cda6342456/test/tests/parallel/set3/scripts/setup_cli.bats#L28-L32

---

It seems the `setup.sh` script being called found a DMS container running concurrently to the setup-sh tests? (_they're both in the same parallel test group_):

https://github.com/docker-mailserver/docker-mailserver/blob/34423c2f665b8c7839c40fbc7d1fe8cda6342456/test/tests/parallel/set3/scripts/postconf-helper.bats#L5

Probably due to `CONTAINER_NAME` usage in `setup.sh`:

https://github.com/docker-mailserver/docker-mailserver/blob/34423c2f665b8c7839c40fbc7d1fe8cda6342456/setup.sh#L172C11-L172C25

https://github.com/docker-mailserver/docker-mailserver/blob/34423c2f665b8c7839c40fbc7d1fe8cda6342456/setup.sh#L187

https://github.com/docker-mailserver/docker-mailserver/blob/34423c2f665b8c7839c40fbc7d1fe8cda6342456/setup.sh#L89-L96

https://github.com/docker-mailserver/docker-mailserver/blob/34423c2f665b8c7839c40fbc7d1fe8cda6342456/setup.sh#L189-L193

---

I could raise an issue about that. It's not likely to ever be an issue for users though. The test should probably be run in the serial group instead of parallel, that way the test shouldn't fail due to race condition.

casperklein (approved):

PR #4124, opened 2024-07-20
docs: Remove `restart: always` from `compose.yaml`

polarathene:
# Description

This line has been in the `compose.yaml` example [since March 2017](https://github.com/docker-mailserver/docker-mailserver/pull/550). As our docs actively discourage container restarts, and to ensure a fresh container start as an initial troubleshooting step, we should probably drop this line?

DMS is a bit more resilient since v14 release now that it skips initial setup scripts:

https://github.com/docker-mailserver/docker-mailserver/blob/c5f125c973d65d0f51000061952ec378b7449c6e/target/scripts/start-mailserver.sh#L172-L174

Although I'm not sure if that has completely addressed the concern? Should keep this line advising `restart: always` as ok? Or is it better to drop it?

## Type of change

## Checklist

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -19,7 +19,6 @@ services:
       - ./docker-data/dms/mail-logs/:/var/log/mail/
       - ./docker-data/dms/config/:/tmp/docker-mailserver/
       - /etc/localtime:/etc/localtime:ro
-    restart: always
     stop_grace_period: 1m
     # Uncomment if using `ENABLE_FAIL2BAN=1`:
     # cap_add:


georglauterbach:
I'm fine with dropping it, as there is no difference between a spinning-up-and-failing mail server and a not running mail server from the user's perspective, but resource-wise, it's better to not restart.

georglauterbach (approved):

casperklein:
> as there is no difference between a spinning-up-and-failing mail server

I cannot confirm this. In my tests, besides some warnings, DMS works just fine (in my setup) after a container restart.

In case of an (unexpected) host reboot, a maybe not perfectly started DMS (which accept mails) is better than nothing and the probability of losing mails.

> Although I'm not sure if that has completely addressed the concern? 

The [PR](https://github.com/docker-mailserver/docker-mailserver/pull/3929) just aimed to improve the situation, but not to fix it completely.

As we learned, there are still scripts running on every startup, that still modify things on container restarts (because they are part of ` _register_functions`).

I would keep the restart policy as it is. But feel free to merge/dismiss as you want.

polarathene:
> I would keep the restart policy as it is. But feel free to merge/dismiss as you want.

I'll close the PR for now. Given the improvement landed with v14 we can wait to see how many issues in future are reported and resolved from container restarts (_which implies the user didn't follow our troubleshooting docs as instructed_).

PR #4138, opened 2024-07-27
docs: Add minimal `compose.yaml` examples that demonstrate specific features

casperklein:
# Description

**NOTE:** This PR description below was added by @polarathene 

Add some of the self-contained `compose.yaml` examples that @polarathene has provided in past issues to support troubleshooting and demonstrate a feature working completely locally.

These are not intended to mirror production deployments, they sometimes take a different approach not suited for real deployments to better support quick setup
- Using non-persistent configs instead of volumes.
- Skipping proper DNS records, postfix security checks, and TLS when it can be avoided.
- Often oriented to using `swaks` as the mail client within a container and staying within the same docker network, rather than interacting with a host mail client. Hence the original issue context for related CLI commands/information.

Our users often aren't aware that they can run DMS like this as easily. They can use multiple instances of DMS to learn how features work or troubleshoot without needing to worry about external third-party services, LetsEncrypt + real domain DNS and VPS, etc.

Reference: https://github.com/docker-mailserver/docker-mailserver/issues/4136#issuecomment-2254115610

## Type of change

## Checklist

### DIFF
--- a/demo-setups/fetchmail-compose.yaml
+++ b/demo-setups/fetchmail-compose.yaml
@@ -18,7 +18,7 @@ services:
     #ports:
     #  - "465:465"  # ESMTP (implicit TLS)
     #  - "993:993"  # IMAP4 (implicit TLS)
-    # You'd normally use volumes here, for simplicity of the example, all config is contained within `compose.yaml`:
+    # You'd normally use `volumes` here but for simplicity of the example, all config is contained within `compose.yaml`:
     configs:
       - source: dms-accounts-fetch
         target: /tmp/docker-mailserver/postfix-accounts.cf
--- a/demo-setups/fetchmail-compose.yaml
+++ b/demo-setups/fetchmail-compose.yaml
@@ -26,7 +26,7 @@ services:
         target: /tmp/docker-mailserver/fetchmail.cf
 
   dms-remote:
-    image: ghcr.io/docker-mailserver/docker-mailserver:latest
+    image: ghcr.io/docker-mailserver/docker-mailserver:latest # :14.0
     hostname: mail.remote.test
     environment:
       # Allows for us send a test mail easily by trusting any mail client run within this container (`swaks`):
--- a/demo-setups/fetchmail-compose.yaml
+++ b/demo-setups/fetchmail-compose.yaml
@@ -1,0 +1,57 @@
+More information: https://github.com/orgs/docker-mailserver/discussions/3994#discussioncomment-9290570
+services:
+  dms-fetch:
+    image: ghcr.io/docker-mailserver/docker-mailserver:latest
+    hostname: mail.example.test
+    environment:
+      ENABLE_FETCHMAIL: 1
+      # We change this setting to 10 for quicker testing:
+      FETCHMAIL_POLL: 10
+    # Link the DNS lookup `remote.test` to resolve to the `dms-remote` container IP (for `@remote.test` address):
+    # This is only for this example, since no real DNS service is configured, this is a Docker internal DNS feature:
+    links:
+      - "dms-remote:remote.test"
+    # NOTE: Optional, You only need to publish ports if you want to verify via your own mail client.
+    #ports:
+    #  - "465:465"  # ESMTP (implicit TLS)
+    #  - "993:993"  # IMAP4 (implicit TLS)
+    # You'd normally use volumes here, for simplicity of the example, all config is contained within `compose.yaml`:
+    configs:
+      - source: dms-accounts-fetch
+        target: /tmp/docker-mailserver/postfix-accounts.cf
+      - source: fetchmail
+        target: /tmp/docker-mailserver/fetchmail.cf
+
+  dms-remote:
+    image: ghcr.io/docker-mailserver/docker-mailserver:latest
+    hostname: mail.remote.test
+    environment:
+      # Allows for us send a test mail easily by trusting any mail client run within this container (`swaks`):
+      PERMIT_DOCKER: container
+      # Alternatively, trust and accept any mail received from clients in same subnet of dms-fetch:
+      #PERMIT_DOCKER: connected-networks
+    configs:
+      - source: dms-accounts-remote
+        target: /tmp/docker-mailserver/postfix-accounts.cf
+
+# Using the Docker Compose `configs.content` feature instead of volume mounting separate files.
+# NOTE: This feature requires Docker Compose v2.23.1 (Nov 2023) or newer:
+# https://github.com/compose-spec/compose-spec/pull/446
+configs:
+  fetchmail:
+    content: |
+      poll 'mail.remote.test' proto imap
+        user 'jane.doe@remote.test'
+        pass 'secret'
+        is 'john.doe@example.test'
+        no sslcertck
+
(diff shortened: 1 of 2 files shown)


demo-setups/fetchmail-compose.yaml
```diff
@@ -0,0 +1,58 @@
...
+    # NOTE: Optional, You only need to publish ports if you want to verify via your own mail client.
+    #ports:
+    #  - "465:465"  # ESMTP (implicit TLS)
+    #  - "993:993"  # IMAP4 (implicit TLS)
+    # You'd normally use volumes here, for simplicity of the example, all config is contained within `compose.yaml`:
+    configs:
+      - source: dms-accounts-fetch
+        target: /tmp/docker-mailserver/postfix-accounts.cf
+      - source: fetchmail
+        target: /tmp/docker-mailserver/fetchmail.cf
```
polarathene, line 24: For reference, if anyone does expand on this in future IIRC, the following info might be worth being aware of.

---

It's possible to split these out to separate `compose.yaml` files and use the Docker Compose feature [`include`](https://docs.docker.com/compose/compose-file/14-include/) or the [merge/override feature](https://docs.docker.com/compose/compose-file/13-merge/#sequence) to share common configs.

Notably, you will have trouble merging YAML lists/sequences, which I think is only supported by multiple `-f` overrides (_requires the same service name_), whereas for YAML maps you can use [Fragments / Extensions](https://docs.docker.com/compose/compose-file/10-fragments/) (_YAML anchors/aliases + YAML merge `<<` syntax_).

- That can make `environment` more suited to map style instead of list, since that supports both forms.
- One notable difference over `volumes` with `configs` is that a bind mount volume will persist changes to disk whereas the `configs` approach here is useful for immutable init state when creating the container, but it can be modified after at runtime.
  - Re-creating the container will start with the initial config in `compose.yaml` again. This is the main value besides self-containing the example into a single file.
  - It might be possible to get the same with `volumes` if some of the volume driver has a setting to emulate one-way sync (_`:ro` / read-only does but it prevents writes in the container_)..

If there are separate YAML snippets to be composed, you just run [`docker compose config`](https://docs.docker.com/reference/cli/docker/compose/config/) instead of the equivalent `docker compose up` to output the full `compose.yaml`. If examples grow it may be useful to know about that :)

demo-setups/fetchmail-compose.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+  dms-remote:
+    image: ghcr.io/docker-mailserver/docker-mailserver:latest
```
polarathene, line 27:
```suggestion
    image: ghcr.io/docker-mailserver/docker-mailserver:latest # :14.0
```

demo-setups/relay-compose.yaml
```diff
@@ -0,0 +1,143 @@
+# More information: https://github.com/docker-mailserver/docker-mailserver/issues/4136#issuecomment-2253693490
```
polarathene, line 1:
```suggestion
# Docs: https://docker-mailserver.github.io/docker-mailserver/v14.0/config/advanced/mail-forwarding/relay-hosts/
# Additional context, with CLI commands for verification:
# https://github.com/docker-mailserver/docker-mailserver/issues/4136#issuecomment-2253693490
```

demo-setups/relay-compose.yaml
```diff
@@ -0,0 +1,143 @@
...
+
+services:
+  # This would represent your actual DMS container:
+  dms-sender:
+    image: mailserver/docker-mailserver:14.0
```
polarathene, line 6:
```suggestion
    image: mailserver/docker-mailserver:latest # :14.0
```

demo-setups/relay-compose.yaml
```diff
@@ -0,0 +1,143 @@
...
+      - source: tls-key
+        target: /tmp/tls/key.pem
+      # This is only needed if you want to verify the TLS cert chain with swaks
+      # (normally with public CA providers like LetsEncrypt this file is already available to a mail client)
+      - source: tls-ca-cert
+        target: /tmp/tls/ca-cert.pem
+
+  # Pretend this is your third-party relay service:
+  dms-relay:
+    image: mailserver/docker-mailserver:14.0
```
polarathene, line 38:
```suggestion
    image: mailserver/docker-mailserver:latest # :14.0
```

demo-setups/relay-compose.yaml
```diff
@@ -0,0 +1,143 @@
...
+      - source: dms-accounts-relay
+        target: /tmp/docker-mailserver/postfix-accounts.cf
+      - source: tls-cert
+        target: /tmp/tls/cert.pem
+      - source: tls-key
+        target: /tmp/tls/key.pem
+
+  # Pretend this is another mail server that your target recipient belongs to (like Gmail):
+  dms-destination:
+    image: mailserver/docker-mailserver:14.0
```
polarathene, line 63:
```suggestion
    image: mailserver/docker-mailserver:latest # :14.0
```

polarathene (changes requested):
May want to decide on being consistent with the registry? `ghcr.io/docker-mailserver/docker-mailserver` / `mailserver/docker-mailserver` (_implicit `docker.io/` prefix_)

Otherwise this is great, thanks for taking the time to follow-up so quickly with the discussion to make these easier to discover! 🎉 

I've left a feedback comment for any future maintainers that might land here, should the examples grow or test suite migrate to compose 👍

demo-setups/fetchmail-compose.yaml
```diff
@@ -0,0 +1,58 @@
+# More information: https://github.com/orgs/docker-mailserver/discussions/3994#discussioncomment-9290570
+
+services:
+  dms-fetch:
+    image: ghcr.io/docker-mailserver/docker-mailserver:latest
```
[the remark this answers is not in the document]
casperklein: Yes. The samples might be broken in the future, so it's good to know at least one version that definitely worked.

demo-setups/fetchmail-compose.yaml
```diff
@@ -0,0 +1,58 @@
...
+      FETCHMAIL_POLL: 10
+    # Link the DNS lookup `remote.test` to resolve to the `dms-remote` container IP (for `@remote.test` address):
+    # This is only for this example, since no real DNS service is configured, this is a Docker internal DNS feature:
+    links:
+      - "dms-remote:remote.test"
+    # NOTE: Optional, You only need to publish ports if you want to verify via your own mail client.
+    #ports:
+    #  - "465:465"  # ESMTP (implicit TLS)
+    #  - "993:993"  # IMAP4 (implicit TLS)
+    # You'd normally use volumes here, for simplicity of the example, all config is contained within `compose.yaml`:
```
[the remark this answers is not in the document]
casperklein: Did you remove the comma by intention?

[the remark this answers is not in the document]
polarathene: Yes, I think that is valid when introducing `but` as a replacement? Seemed to read better?

polarathene (approved):

georglauterbach (approved):
LGTM 👍🏼 Nice addition 🚀

PR #4146, opened 2024-07-31
fix: Ensure main log file is tailed from the start

polarathene:
# Description

I noticed a difference in log output when troubleshooting a recent issue. Depending on if `ENABLE_AMAVIS` was on/off, earlier logs in this file did not appear. The `touch` doesn't seem relevant as logs from starting up daemons were already present, but in this case only slightly delayed amavis logs were displayed.

Syntax of the `tail` command was slightly off, seems to require a `+` in front of the number given to `-n` 👍 

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,7 @@ All notable changes to this project will be documented in this file. The format 
   - Update logwatch `ignore.conf` to exclude Xapian messages about pending documents ([#4060](https://github.com/docker-mailserver/docker-mailserver/pull/4060))
   - `dovecot-fts-xapian` plugin was updated to `1.7.13`, fixing a regression with indexing ([#4095](https://github.com/docker-mailserver/docker-mailserver/pull/4095))
 - **Internal:**
-  - The main `mail.log` followed to stdout now correctly begins from the first log line ([#4146](https://github.com/docker-mailserver/docker-mailserver/pull/4146))
+  - The main `mail.log` which is piped to stdout via `tail` now correctly begins from the first log line of the active container run. Previously some daemon logs and potential warnings/errors were omitted. ([#4146](https://github.com/docker-mailserver/docker-mailserver/pull/4146))
   - Fixed a regression introduced in v14 where `postfix-main.cf` appended `stderr` output into `/etc/postfix/main.cf`, causing Postfix startup to fail ([#4147](https://github.com/docker-mailserver/docker-mailserver/pull/4147))
 
 ### CI
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,8 @@ All notable changes to this project will be documented in this file. The format 
 - **Dovecot:**
   - Update logwatch `ignore.conf` to exclude Xapian messages about pending documents ([#4060](https://github.com/docker-mailserver/docker-mailserver/pull/4060))
   - `dovecot-fts-xapian` plugin was updated to `1.7.13`, fixing a regression with indexing ([#4095](https://github.com/docker-mailserver/docker-mailserver/pull/4095))
+- **Internal:**
+  - The main `mail.log` followed to stdout now correctly begins from the first log line ([#4146](https://github.com/docker-mailserver/docker-mailserver/pull/4146))
 
 ### CI
 
--- a/target/scripts/start-mailserver.sh
+++ b/target/scripts/start-mailserver.sh
@@ -199,5 +199,8 @@ _start_daemons
 
 _log 'info' "${HOSTNAME} is up and running"
 
+# NOTE: This file should already exist with log output from rsyslogd:
 touch /var/log/mail/mail.log
-exec tail -Fn 0 /var/log/mail/mail.log
+# Container logs will receive updates to this log file,
+# `-n +0` ensures we output to stdout from the first line of the file.
+exec tail -Fn +0 /var/log/mail/mail.log
--- a/target/scripts/start-mailserver.sh
+++ b/target/scripts/start-mailserver.sh
@@ -194,7 +194,7 @@ fi
 # marker to check if container was restarted
 date >/CONTAINER_START
 
-# Container logs will receive updates to this log file:
+# Container logs will receive updates from this log file:
 MAIN_LOGFILE=/var/log/mail/mail.log
 # NOTE: rsyslogd would usually create this later during `_start_daemons`, however it would already exist if the container was restarted.
 touch "${MAIN_LOGFILE}"
--- a/target/scripts/start-mailserver.sh
+++ b/target/scripts/start-mailserver.sh
@@ -194,13 +194,16 @@ fi
 # marker to check if container was restarted
 date >/CONTAINER_START
 
+# Container logs will receive updates to this log file:
+MAIN_LOGFILE=/var/log/mail/mail.log
+# NOTE: rsyslogd would usually create this later during `_start_daemons`, however it would already exist if the container was restarted.
+touch "${MAIN_LOGFILE}"
+# Ensure `tail` follows the correct position of the log file for this container start (new logs begin once `_start_daemons` is called)
+TAIL_START=$(wc -l < "${MAIN_LOGFILE}")
+
 [[ ${LOG_LEVEL} =~ (debug|trace) ]] && print-environment
 _start_daemons
 
+# Container start-up scripts completed. `tail` will now pipe the log updates to stdout:
 _log 'info' "${HOSTNAME} is up and running"
-
-# NOTE: This file should already exist with log output from rsyslogd:
-touch /var/log/mail/mail.log
-# Container logs will receive updates to this log file,
-# `-n +0` ensures we output to stdout from the first line of the file.
-exec tail -Fn +0 /var/log/mail/mail.log
+exec tail -Fn "+${TAIL_START}" "${MAIN_LOGFILE}"
--- a/target/scripts/start-mailserver.sh
+++ b/target/scripts/start-mailserver.sh
@@ -199,7 +199,7 @@ MAIN_LOGFILE=/var/log/mail/mail.log
 # NOTE: rsyslogd would usually create this later during `_start_daemons`, however it would already exist if the container was restarted.
 touch "${MAIN_LOGFILE}"
 # Ensure `tail` follows the correct position of the log file for this container start (new logs begin once `_start_daemons` is called)
-TAIL_START=$(wc -l < "${MAIN_LOGFILE}")
+TAIL_START=$(( $(wc -l < "${MAIN_LOGFILE}") + 1 ))
 
 [[ ${LOG_LEVEL} =~ (debug|trace) ]] && print-environment
 _start_daemons


georglauterbach (approved):

casperklein:
> Syntax of the tail command was slightly off, seems to require a + in front of the number given to -n 👍

I think you misunderstood the intention of "-F -n 0". It means, follow/display the file, start from the very end. So essentially, only display log lines, that are for the current session.

`-n +0` / `-n +1` means, display the file from line 0/1. IMO, the latter is more logical.

Your change will display the whole log file on each DMS startup. Depending on logrotation interval and amount of log file lines, this can be a lot and may take some time to process.

With our default `LOGROTATE_INTERVAL=weekly`, up to one week of mail.log entries would be displayed on startup.

I see no value in displaying older log entrys than the current DMS session.

> but in this case only slightly delayed amavis logs were displayed.

That sounds like a race condition, when new log lines from the current sessions are written, while `tail` haven't been executed yet.

We could check, what the line count of mail.log is, before starting the daemons. And then using that value for `tail`:

```bash
TAIL_START=$(wc -l < mail.log)

tail -F -n +$TAIL_START
```

(not tested).

polarathene:
> Your change will display the whole log file on each DMS startup. Depending on logrotation interval and amount of log file lines, this can be a lot and may take some time to process.

Thanks for pointing this out. My main concern was that log lines were not being output (an error) as they were racey from daemons starting prior to the tail.

> > but in this case only slightly delayed amavis logs were displayed.
> 
> That sounds like a race condition, when new log lines from the current sessions are written, while `tail` haven't been executed yet.
> 
> We could check, what the line count of mail.log is, before starting the daemons. And then using that value for `tail`

Yeah, the missing log lines were about 0.5-1 second offset from the displayed Amavis ones. Some would show when Amavis was disabled but not the OpenDKIM/OpenDMARC IIRC (_these were the earliest log lines_).

Your suggestion sounds fine 👍 I'll verify and update the PR tomorrow with that.

georglauterbach (changes requested):
With @casperklein's explanation, I now understand the issue, and I agree with him. I'll block on this via review so that it does not get merged by accident.

polarathene:
I'll open a separate issue for the bug below.

<details>
<summary>Resolved - Cause is not from this PR</summary>

After verifying @casperklein advice as working well, I noticed that if I force stop the container (`docker compose up` for the container restart, followed by `CTRL+C` twice after container restart startup was successful_), there was a loop of error spam to `/var/log/mail.log` and `/var/log/syslog`, with lines like:

```
2024-08-02T00:13:59.113778+00:00 mail dovecot: master: Fatal: Dovecot is already running with PID 133 (read from /run/dovecot/master.pid)
```

~~I'm not sure exactly why the changes I've done here make that happen?~~ (_**EDIT:** Not due to this PR, see update at tend of message_)

Without them I only get that line logged once (_internally since prior to this PR it's too early to output_). It's a bit racey, and I know why it's a problem, but I just don't understand why the change here was introducing the crash loop 🤔 

After the standard Amavis log lines, it's just this over and over:

```
dms-1  | 2024-08-02 00:14:06,365 WARN exited: dovecot (exit status 89; not expected)
dms-1  | 2024-08-02 00:14:06,387 WARN exited: dovecot (exit status 89; not expected)
dms-1  | 2024-08-02 00:14:06,409 WARN exited: dovecot (exit status 89; not expected)
dms-1  | 2024-08-02 00:14:06,433 WARN exited: dovecot (exit status 89; not expected)
dms-1  | 2024-08-02T00:14:06.129730+00:00 mail dovecot: master: Fatal: Dovecot is already running with PID 133 (read from /run/dovecot/master.pid)
dms-1  | 2024-08-02T00:14:06.151205+00:00 mail dovecot: master: Fatal: Dovecot is already running with PID 133 (read from /run/dovecot/master.pid)
dms-1  | 2024-08-02T00:14:06.171097+00:00 mail dovecot: master: Fatal: Dovecot is already running with PID 133 (read from /run/dovecot/master.pid)
dms-1  | 2024-08-02T00:14:06.192423+00:00 mail dovecot: master: Fatal: Dovecot is already running with PID 133 (read from /run/dovecot/master.pid)
dms-1  | 2024-08-02T00:14:06.213524+00:00 mail dovecot: master: Fatal: Dovecot is already running with PID 133 (read from /run/dovecot/master.pid)
dms-1  | 2024-08-02T00:14:06.234588+00:00 mail dovecot: master: Fatal: Dovecot is already running with PID 133 (read from /run/dovecot/master.pid)
dms-1  | 2024-08-02T00:14:06.255106+00:00 mail dovecot: master: Fatal: Dovecot is already running with PID 133 (read from /run/dovecot/master.pid)
dms-1  | 2024-08-02T00:14:06.277583+00:00 mail dovecot: master: Fatal: Dovecot is already running with PID 133 (read from /run/dovecot/master.pid)
dms-1  | 2024-08-02T00:14:06.299474+00:00 mail dovecot: master: Fatal: Dovecot is already running with PID 133 (read from /run/dovecot/master.pid)
dms-1  | 2024-08-02T00:14:06.322426+00:00 mail dovecot: master: Fatal: Dovecot is already running with PID 133 (read from /run/dovecot/master.pid)
dms-1  | 2024-08-02T00:14:06.343959+00:00 mail dovecot: master: Fatal: Dovecot is already running with PID 133 (read from /run/dovecot/master.pid)
dms-1  | 2024-08-02T00:14:06.365593+00:00 mail dovecot: master: Fatal: Dovecot is already running with PID 133 (read from /run/dovecot/master.pid)
dms-1  | 2024-08-02T00:14:06.386614+00:00 mail dovecot: master: Fatal: Dovecot is already running with PID 133 (read from /run/dovecot/master.pid)
dms-1  | 2024-08-02T00:14:06.408952+00:00 mail dovecot: master: Fatal: Dovecot is already running with PID 133 (read from /run/dovecot/master.pid)
dms-1  | 2024-08-02T00:14:06.432847+00:00 mail dovecot: master: Fatal: Dovecot is already running with PID 133 (read from /run/dovecot/master.pid)
dms-1  | 2024-08-02 00:14:07,133 WARN exited: dovecot (exit status 89; not expected)
dms-1  | 2024-08-02 00:14:07,160 WARN exited: dovecot (exit status 89; not expected)
dms-1  | 2024-08-02 00:14:07,184 WARN exited: dovecot (exit status 89; not expected)
dms-1  | 2024-08-02 00:14:07,210 WARN exited: dovecot (exit status 89; not expected)
dms-1  | 2024-08-02 00:14:07,237 WARN exited: dovecot (exit status 89; not expected)
dms-1  | 2024-08-02T00:14:07.132721+00:00 mail dovecot: master: Fatal: Dovecot is already running with PID 133 (read from /run/dovecot/master.pid)
dms-1  | 2024-08-02T00:14:07.160243+00:00 mail dovecot: master: Fatal: Dovecot is already running with PID 133 (read from /run/dovecot/master.pid)
dms-1  | 2024-08-02T00:14:07.183814+00:00 mail dovecot: master: Fatal: Dovecot is already running with PID 133 (read from /run/dovecot/master.pid)
dms-1  | 2024-08-02T00:14:07.210406+00:00 mail dovecot: master: Fatal: Dovecot is already running with PID 133 (read from /run/dovecot/master.pid)
dms-1  | 2024-08-02T00:14:07.237414+00:00 mail dovecot: master: Fatal: Dovecot is already running with PID 133 (read from /run/dovecot/master.pid)
dms-1  | 2024-08-02 00:14:08,132 WARN exited: dovecot (exit status 89; not expected)
dms-1  | 2024-08-02 00:14:08,152 WARN exited: dovecot (exit status 89; not expected)
dms-1  | 2024-08-02 00:14:08,174 WARN exited: dovecot (exit status 89; not expected)
dms-1  | 2024-08-02 00:14:08,198 WARN exited: dovecot (exit status 89; not expected)
dms-1  | 2024-08-02 00:14:08,220 WARN exited: dovecot (exit status 89; not expected)
dms-1  | 2024-08-02 00:14:08,244 WARN exited: dovecot (exit status 89; not expected)
dms-1  | 2024-08-02 00:14:08,266 WARN exited: dovecot (exit status 89; not expected)
```

---

That'd be fixed if we ensure PID is cleaned up for container restarts too:

https://github.com/docker-mailserver/docker-mailserver/blob/4f1d075d43348a15dd2fec840f2427cb91760a69/target/scripts/startup/setup-stack.sh#L85-L104

https://github.com/docker-mailserver/docker-mailserver/blob/4f1d075d43348a15dd2fec840f2427cb91760a69/target/scripts/start-mailserver.sh#L121

It's not really a blocker for this PR, some change here just better surfaced the regression introduced with v14 for the container restart skip feature. I'd still like to know what this PR does to cause the loop though 🤔 

**EDIT:** My `:edge` image was built from https://github.com/docker-mailserver/docker-mailserver/commit/9175424d0f3c67622c030311cc720d7477f1c2b6 (_about 4 weeks old_), I was testing these changes with a volume mount to replace the file with a master copy modified, so that might have something to do with it..

**UPDATE:** Not this PR 😅 I could reproduce without the volume mount with both the old `:edge` and current `:edge` images, so definitely a race condition from unclean container exit that is not handled properly from v14 with https://github.com/docker-mailserver/docker-mailserver/pull/3929

For reference these are the SHA256 image digests:

```
# Either DockerHub or GHCR registries are fine, same digest for both.
# GHCR is easier to browse by digest, DockerHub only seems to list current digest for the published tag.
# https://github.com/docker-mailserver/docker-mailserver/pkgs/container/docker-mailserver/versions

# Current `:edge`
# Commit: https://github.com/docker-mailserver/docker-mailserver/commit/2f8ad142eccfe3d6291373629c5b1dbf89b71f52
ghcr.io/docker-mailserver/docker-mailserver@sha256:22bdbd1288a42ed7fe9e304cc0c99b022933ece73658d6b5b7f183387b58e226

# The older `:edge` image I had:
# Commit: https://github.com/docker-mailserver/docker-mailserver/commit/9175424d0f3c67622c030311cc720d7477f1c2b6
ghcr.io/docker-mailserver/docker-mailserver@sha256:e9c54948daed0f1d9d08872ea37e9cbd2a4446da094778b8a4e096c73db869a5
```

</details>

target/scripts/start-mailserver.sh
```diff
@@ -194,10 +194,16 @@ fi
 # marker to check if container was restarted
 date >/CONTAINER_START
 
+# Container logs will receive updates to this log file:
+MAIN_LOGFILE=/var/log/mail/mail.log
+# NOTE: rsyslogd would usually create this later during `_start_daemons`, however it would already exist if the container was restarted.
+touch "${MAIN_LOGFILE}"
+# Ensure `tail` follows the correct position of the log file for this container start (new logs begin once `_start_daemons` is called)
+TAIL_START=$(wc -l < "${MAIN_LOGFILE}")
```
polarathene, line 202: Increment by 1 via [arithimitic expansion](https://stackoverflow.com/questions/6348902/how-can-i-add-numbers-in-a-bash-script/6348945#6348945) bash syntax:

```suggestion
TAIL_START=$(( $(wc -l < "${MAIN_LOGFILE}") + 1 ))
```

Container restarts were otherwise displaying the last log line from the prior container run.

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ All notable changes to this project will be documented in this file. The format
   - Update logwatch `ignore.conf` to exclude Xapian messages about pending documents ([#4060](https://github.com/docker-mailserver/docker-mailserver/pull/4060))
   - `dovecot-fts-xapian` plugin was updated to `1.7.13`, fixing a regression with indexing ([#4095](https://github.com/docker-mailserver/docker-mailserver/pull/4095))
 - **Internal:**
+  - The main `mail.log` followed to stdout now correctly begins from the first log line ([#4146](https://github.com/docker-mailserver/docker-mailserver/pull/4146))
```
polarathene, line 35:
```suggestion
  - The main `mail.log` which is piped to stdout via `tail` now correctly begins from the first log line of the active container run. Previously some daemon logs and potential warnings/errors were omitted. ([#4146](https://github.com/docker-mailserver/docker-mailserver/pull/4146))
```

polarathene:
Verified locally. Should be sorted now 👍

georglauterbach (approved):

target/scripts/start-mailserver.sh
```diff
@@ -194,10 +194,16 @@ fi
 # marker to check if container was restarted
 date >/CONTAINER_START
 
+# Container logs will receive updates to this log file:
+MAIN_LOGFILE=/var/log/mail/mail.log
+# NOTE: rsyslogd would usually create this later during `_start_daemons`, however it would already exist if the container was restarted.
```
casperklein, line 199: I am confused about this comment. I don't get, what it tries to explain.

`touch` is used to avoid an error message from `wc`/`tail` if that file doesn't exist?

target/scripts/start-mailserver.sh
```diff
@@ -194,10 +194,16 @@ fi
 # marker to check if container was restarted
 date >/CONTAINER_START
 
+# Container logs will receive updates to this log file:
```
casperklein, line 197:
```suggestion
# Container logs will receive updates from this log file:
```

casperklein (dismissed):

target/scripts/start-mailserver.sh
polarathene: Perhaps you have a better way to phrase it?

I wanted to make it clear that the `touch` is only necessary on first startup here to support `wc`/`tail` before `_start_daemons`, otherwise on a container restart it would already exist even without `touch` here as `rsyslogd` would have created when started.

Main intent was clarification that `touch` isn't the origin of this file being created, since it's implicitly created by `rsyslogd`, but we don't have a custom config maintained for that (_we kinda do, in the sense of `sed` in the `Dockerfile`_).

---

The offset could be 0 or 1 by default, and only update in the restart condition branch before calling `_start_daemons` if that would be a better approach?

I went with what we have here since it seemed nicer to group related lines, rather than add a little more complexity via branched logic 😅

casperklein: > Main intent was clarification that touch isn't the origin of this file being created

👍

> Perhaps you have a better way to phrase it?

I am totally fine with your detailed explanation. Now I get it 😄 

> I went with what we have here

That's fine 👍

casperklein (approved):

Issue #1911, opened 2021-04-15
Mutiple hostnames with TLS


perry-amsterdam:
# Subject

I have questions about TLS/SSL/STARTTLS/OpenSSL

Is het possible to use multiple domain names with mail.domain1, mail.domain2, mail.domain3
Each with a letsencrypt certificate ?

This for postfix and dovecot ?

Perry

casperklein:
I don't know if multiple certificates are possible. However a SAN certificate (with multiple domain names) should work. Just give it a try 😉

reneploetz:
It should be possible to do this, tough you need to add your own configuration to either user-patches.sh for that or create a custom image:
- postfix: tls_server_sni_maps - http://www.postfix.org/postconf.5.html#tls_server_sni_maps
- dovecot: https://doc.dovecot.org/configuration_manual/dovecot_ssl_configuration/#with-client-tls-sni-server-name-indication-support

yash-ahir:
> I don't know if multiple certificates are possible. However a SAN certificate (with multiple domain names) should work. Just give it a try wink

Hello, sorry to re-open this issue but I'm currently in the same position. I've issued a certificate with 

```
docker run --rm -it -v $PWD/log/:/var/log/letsencrypt/ -v $PWD/etc/:/etc/letsencrypt/ -p 80:80 certbot/certbot certonly --standalone -d mail.yashahir.tech -d mail.syntx.dev
```

This is a SAN certificate yet it does not work as you mentioned.

Hostname is `mail`, the domainname is `yashahir.tech` in `docker-compose.yml`

---

> It should be possible to do this, tough you need to add your own configuration to either user-patches.sh for that or create a custom image:
> 
>     * postfix: tls_server_sni_maps - http://www.postfix.org/postconf.5.html#tls_server_sni_maps
> 
>     * dovecot: https://doc.dovecot.org/configuration_manual/dovecot_ssl_configuration/#with-client-tls-sni-server-name-indication-support
> 
> 
> See also: https://docker-mailserver.github.io/docker-mailserver/edge/config/advanced/override-defaults/user-patches/
> 
> An alternative would be to use a proxy in front of your mailserver / dovecot which terminates TLS, e.g. traefik.

So from what I understand here (and if this is applicable, hopefully to others with the same problem),

After putting the certificate and TLD information in both the `sni` file for postfix and `conf.d/10-ssl.conf`(?) for dovecot and restarting the container stack should work correctly?

[...]

Issue #1674, opened 2020-11-02
Add SNI support


martijnhiemstra:
## Is your feature request related to a problem?
Yes, some mail clients are having problems with the domain certificates

## Describe the solution you'd like
Implement SNI in postfix. Use the tls_server_sni_maps in postfix.

## Who will that feature be useful to?
Everybody implementing ssl/tls

## What are you going to contribute?
I can test the new implementation

aendeavor:
Hey, thanks for opening this issue and contributing this idea.

You will need to implement and test this feature yourself:) When you're ready, provide a PR.

PR #4155, opened 2024-08-08
feat: basic support for SNI via Traefik

vp1100:
# Description

My attempt to implement a support for SNI in Dovecot and Postfix. The configuration is intended for use with traefik

The idea is to provide a separate certificate for each domain trying to connect to the server (mail.domain2.com) instead of using a single main server (mail.domain.com)

TODO: add explanation in docs and a guide

## Type of change

## Checklist

### DIFF
--- a/docs/content/config/environment.md
+++ b/docs/content/config/environment.md
@@ -191,6 +191,12 @@ In the majority of cases, you want `letsencrypt` or `manual`.
 
 Please read [the SSL page in the documentation][docs-tls] for more information.
 
+##### SSL_SNI_DOMAINS
+
+Comma separated list of domains.
+
+Certificates will be exported for each domain from traefik `acme.json` and added to Postfix and Dovecot for SNI support.
+
 ##### TLS_LEVEL
 
 - **empty** => modern
--- a/mailserver.env
+++ b/mailserver.env
@@ -240,6 +240,9 @@ SMTP_ONLY=
 # self-signed => Enables self-signed certificates
 SSL_TYPE=
 
+# Provide a comma separated list of domains to additionally export the certificate from acme.json
+SSL_SNI_DOMAINS=
+
 # These are only supported with `SSL_TYPE=manual`.
 # Provide the path to your cert and key files that you've mounted access to within the container.
 SSL_CERT_PATH=
--- a/target/scripts/helpers/ssl.sh
+++ b/target/scripts/helpers/ssl.sh
@@ -127,6 +127,44 @@ function _setup_ssl() {
       fi
 
       _log 'trace' "letsencrypt (acme.json) extracted certificate using ${EXTRACTED_DOMAIN[0]}: '${EXTRACTED_DOMAIN[1]}'"
+
+      # Extracting certificates for SNI support
+      if [[ -n ${SSL_SNI_DOMAINS} ]] ; then
+        # add empty dovecot & postfix config
+        echo -n "" > /etc/dovecot/conf.d/20-sni.conf
+        echo -n "" > /etc/postfix/sni.map
+
+        # add tls_server_sni_maps yo main.cf if not exist
+        local SNI_MAPS="tls_server_sni_maps = hash:/etc/postfix/sni.map"
+        grep -qxF -- "${SNI_MAPS}" "/etc/postfix/main.cf" || echo "${SNI_MAPS}" >> /etc/postfix/main.cf
+
+        for SNI_DOMAIN in ${SSL_SNI_DOMAINS//,/ }
+        do
+          if _extract_certs_from_acme "${SNI_DOMAIN}"; then
+            local PRIVATE_KEY="/etc/letsencrypt/live/${SNI_DOMAIN}/key.pem"
+            local CERT_CHAIN="/etc/letsencrypt/live/${SNI_DOMAIN}/fullchain.pem"
+
+            # add certificate to postfix
+            echo "${SNI_DOMAIN} ${PRIVATE_KEY} ${CERT_CHAIN}" >> /etc/postfix/sni.map
+
+            # add certificate to dovecot
+            {
+              echo "local_name ${SNI_DOMAIN} {"
+              echo "  ssl_cert = <${CERT_CHAIN}"
+              echo "  ssl_key = <${PRIVATE_KEY}"
+              echo "}"
+            } >> /etc/dovecot/conf.d/20-sni.conf
+
+            _log 'trace' "SNI: extracted domain: ${SNI_DOMAIN}"
+          else
+            _log 'warn' "SNI: letsencrypt (acme.json) failed to extract domain: ${SNI_DOMAIN}"
+          fi
+        done
+
+        # create postfix SNI table
+        postmap -F hash:/etc/postfix/sni.map
+        _log 'trace' "SNI: creating postfix db (sni.map.db)"
+      fi
     fi
   }
 


polarathene:
I'm reluctant towards this PR, but it could be a misunderstanding on my end and I'll revisit when I have some more time to spare later.

There is no related issue or discussion proposing this feature right? You'll need maintainers to see value in this feature for it to be accepted, otherwise our policy is:
- Contribute a guide to our docs for other users. Often you can configure features with a few lines of our existing features + `user-patches.sh`.
- Sufficient user interest expressed / voted (via 👍 reactions) to justify bringing in new features.

It's not that they aren't appreciated as contributions, but that maintainers for the project are very limited in time and we often are not able to offer support going forward for each niche feature addition, but accepting the feature PR delegates the burden of maintenance onto us (_for example this area of scripts with managing certs needs to be completely refactored_).

---

So could you please explain a real problem that this feature solves? What is wrong with each mail domain (`@example.com`, @example.net`, etc) having their DNS configured for a single mail server? If you adopted professional mail services, that's usually expected to point to their mail servers and they manage the certs for it, without any need for your own domains in your control to have certs provisioned to use it.

You could also have multiple SAN assigned to a single certificate, you don't need to manage multiple with SNI like proposed here?

vp1100:
Related issues: [#1674](https://github.com/docker-mailserver/docker-mailserver/issues/1674), [#1911](https://github.com/docker-mailserver/docker-mailserver/issues/1911)

The problem in my case is that I have 5 clients, each have a separate domain with website and emails. I need each client to connect to the mail server by their own domain (mail.client1.com, mail.client2.com), so that they are not related with each other or with my main domain.

The usage of SAN in certificate is not an option for me, as the certificate will show all included domains, which will link these clients together.

The usage of multiple SAN have some other limitations:
1. It only supports 100 san names
2. Each time a new domain is added, the certificate should be recreated
3. Difficult to maintain, each san should be added to traefik labels on all websites.

polarathene:
> I need each client to connect to the mail server by their own domain (`mail.client1.com`, `mail.client2.com`), so that they are **not related** with each other or **with my main domain**.

Since this isn't a real technical barrier with DMS, I'm reluctant to approve.

In the past when this has been brought up (_like the issues you've linked to, but some others like a recent Github Discussion iirc_), maintainers have not expressed enthusiasm towards the use-case. @casperklein and @georglauterbach probably share the same reluctance I have for this PR.

I'm more than happy to review and collaborate with you for a contribution that is solely focused on a guide via docs. You can also raise a new feature request issue for that docs contribution to link to - and if user demand is high / convincing, this PR could be revisited.

In the meantime, I would suggest relying on `user-patches.sh` if my feedback above has not been convincing for you to avoid the feature you're seeking.

---

# Understanding your motivations for the feature

So that we're on the same page, I want to confirm that this isn't an [XY problem](https://en.wikipedia.org/wiki/XY_problem).

Below I've provided input regarding two common motivations for wanting this feature in DMS. Often it seems to be misguided with no technical merit, nor is likely to matter much to your clients or their users.

Please correct me if I'm mistaken or not identified what problem the feature proposed is intended to solve.

## Motivation - Mail domains with a vanity mail server FQDN 

Vanity FQDN used for DNS records, certificate, mail client config reasons.

> The usage of multiple SAN have some other limitations:
> 
> 1. It only supports 100 san names
> 2. Each time a new domain is added, the certificate should be recreated
> 3. Difficult to maintain, each san should be added to traefik labels on all websites.

Right, if you need to scale this out it becomes a bit more problematic but your alternative preference is to have over 100 individual vanity certs provisioned instead? Or are you using SAN/wildcard to share the mail cert with other services fronted by Traefik?

Are you able to point me to mail providers that support a custom mail domain certificate instead of using their mail server FQDN unrelated to your domain that they'll manage mail for?

- [ProtonMail](https://proton.me/support/custom-domain-google)
- [Gmail - MX](https://support.google.com/a/answer/174125)
- [Gmail - POP3 / IMAP](https://support.google.com/mail/answer/7104828)
- [Outlook](https://support.microsoft.com/en-us/office/configure-a-custom-email-domain-account-as-an-exchange-account-in-outlook-1402a3e9-4300-4e2f-a5fa-875d6f7a658e) (CNAME for autodiscover)
- [MXRoute](https://mxroutedocs.com/dns/dnsrecords/)

These all require you to configure for their mail servers, even when you have your own mail domains.

---

Now in your scenario with clients domains, you're using Traefik to manage certs and DMS for mail server. You are provisioning the certs by either having some semblance of control of the clients DNS service for their domain(s) or you have them configure prior to provisioning those certs.

The only motivations that come to mind to want this behaviour is either:
- You mistakenly think that each mail domain needs an equivalent certificate for the mail server to support (it doesn't).
- You or your clients expect DNS to be configured only with IPs to your mail server, no foreign FQDN (_like `mail.example.com` in a `MX` record_).
- Users of those clients during setup of their mail client being configured explicitly for a mail server FQDN that is not the one that belongs to your clients mail domain(s).

---

If the IP of your mailserver changes, then all your clients need to update their DNS for their vanity MX records to point to the new server IP? How are you managing SPF? Explicit IP for your mail server too? Or referring to that same vanity MX record? Anything else like that going on?

When certs are updated, they'll introduce some interruption to DMS. If this is frequent as your number of clients grows (such as the over 100 concern), that could also become more of an issue (_see the issue open about `check-for-changes.sh` at scale for over 1k users_). It would depend how often the certs renew and if they're all renewing roughly around the same time or staggered out (_based on different provision times as you acquire new clients_). For this concern with certs, it'd be more problematic when the updates are more closer together, but should otherwise be fine. Still it's better to just be a single cert (_without each client as an individual SAN_).

Avoiding the vanity DNS and certs is preferrable. Easier for you to manage and support. Mail clients can use **autodiscover** if configuration is a concern. I doubt the cert being for the mail server domain instead of the clients domain should be that big of a concern (_users aren't likely to care, while if the client does show them the links above I provided_).

---

## Motivation - Belief that it improves privacy or security

FWIW, if your issue with a cert with all client domains associated to it is viewing it as a security / privacy concern, separate certificates won't prevent discovery of these clients having your mail server in common:

```console
# Github Pages:
$ doggo A github.io

NAME            TYPE    CLASS   TTL     ADDRESS         NAMESERVER
github.io.      A       IN      0s      185.199.108.153 172.18.160.1:53
github.io.      A       IN      0s      185.199.111.153 172.18.160.1:53
github.io.      A       IN      0s      185.199.109.153 172.18.160.1:53
github.io.      A       IN      0s      185.199.110.153 172.18.160.1:53
```

Images below are from: https://search.censys.io/hosts/185.199.108.153/explore

> image

> image

> image

I ran out of queries apparently (requires registering an account), but in that last screenshot you can interact with the graph to reveal all known domains associated to that IP, and all certificates associated to the IP or domains. There's other sites like this like Shodan too which likewise requires an account for broader insights.

CA's like LetsEncrypt also keep track of a public log for certificates, so any subdomains you provision for are recorded in that log which gives away information that may otherwise not be as easily accessible for example.

Perhaps your reasoning was related to that, since services like the above are available to search DNS records, so that while your mailserver IP and a single cert provisioned for that may not identify your clients domains, their DNS records pointing to your mail server FQDN or IP would much in the same way. [Related discussion from May 2023](https://github.com/orgs/docker-mailserver/discussions/3312#discussioncomment-5826279).

vp1100:
Thank you for your detailed message. The tools you mentioned are great. This made me rethink my configuration and move towards a single domain for the mail server.

In my case I host clients websites and have control over DNS. The DNS is configured only with IPs which will require a change if server IP changes.

You are right about privacy concerns, if someone sets a goal, they will find a connection between sites by IP. Additionally all outgoing mail is coming from my main domain, where SPF is configured, which also reveals my domain. In the end I am thinking to just register another domain for mail server and websites hosting, place them on different IP, and to configure DNS via CNAME records. That way it will not be associated with my main domain.

As for the pr we can close it, unless there would be a high demand for this feature is future

polarathene:
> Thank you for your detailed message. The tools you mentioned are great. This made me rethink my configuration and move towards a single domain for the mail server.

You're welcome! Glad to hear my feedback made sense :)

> As for the pr we can close it, unless there would be a high demand for this feature is future

I will close it. I think anyone who discovers the past issues you linked to can find this PR and the discussion for context.

I may add an entry to our docs FAQ to link to my comment above, just to help avoid the usual XY problem.

PR #4183, opened 2024-09-18
docs: Podman - Document how to setup DMS with Quadlet

RoelSG:
Generate systemd is deprecated and Quadlets are now used. I have spent some time getting Quadlet to work, so I added some findings to the documentation. It is my first time contributing to documentation, so excuse my English if it comes across a bit rambly.

I have only added information, and have not removed any. However, generate-systemd is now deprecated and slirp4netns is out in favour of pasta. Pasta in particular is a lot easier to configure, but is still complicated to include into a docker-compose, which I do not use anymore.

Using podlet you can also translate a docker-compose to a Quadlet. Quadlets themselves behave a lot like individual  compose files, but are managed by systemd. 

## Checklist


docs/content/config/advanced/podman.md
```diff
@@ -96,6 +96,94 @@ docker compose up -d mailserver
...
+
+- $XDG_RUNTIME_DIR/containers/systemd/
+- $XDG_CONFIG_HOME/containers/systemd/ or ~/.config/containers/systemd/
+- /etc/containers/systemd/users/$(UID)
+- /etc/containers/systemd/users/
+
+#### Example Quadlet file
+
+We have to use the .container extension for the quadlet generator to pick up the service.
+Because dockermailserver uses multiple users inside the container, we will either have to use our own user as root, resulting in our e-mails being owned by a subuid. Alternatively, using UIDMap we can map our rootless user to UID 5000 in the container who owns our e-mails. Using UIDMap also maps root user 0 inside the container to an available sub-uid of our rootless user. Otherwise the container will not have permission to configure itself.
```
casperklein, line 125:
```suggestion
Because docker-mailserver uses multiple users inside the container, we will either have to use our own user as root, resulting in our e-mails being owned by a subuid. Alternatively, using UIDMap we can map our rootless user to UID 5000 in the container who owns our e-mails. Using UIDMap also maps root user 0 inside the container to an available sub-uid of our rootless user. Otherwise the container will not have permission to configure itself.
```

docs/content/config/advanced/podman.md
```diff
@@ -104,6 +192,9 @@ In rootless mode, podman resolves all incoming IPs as localhost, which results i
 
 The `PERMIT_DOCKER` variable in the `mailserver.env` file allows to specify trusted networks that do not need to authenticate. If the variable is left empty, only requests from localhost and the container IP are allowed, but in the case of rootless podman any IP will be resolved as localhost. Setting `PERMIT_DOCKER=none` enforces authentication also from localhost, which prevents sending unauthenticated emails.
 
+#### Use the pasta network driver
+As of podman 5.0 pasta is the default network driver of rootless containers. This will have the same functionality and caveats as the `slirp4netns` driver. You do not need to set and interface name.
```
casperklein, line 196:
```suggestion
As of podman 5.0 pasta is the default network driver of rootless containers. This will have the same functionality and caveats as the `slirp4netns` driver. You do not need to set an interface name.
```

docs/content/config/advanced/podman.md
```diff
@@ -96,6 +96,94 @@ docker compose up -d mailserver
...
+
+# Environment variables
+# General Settings
+
+Environment=HOSTNAME=example
+Environment=DOMAINNAME=example.com
+Environment=CONTAINER_NAME=dockermailserver
+...
+```
+Stopping the service with systemd will result in the container being removed. Restarting will use the existing container. You do not need to enable services with Quadlet.
```
casperklein, line 173:
```suggestion
Stopping the service with systemd will result in the container being removed. Restarting will use the existing container, which is however not recommended. You do not need to enable services with Quadlet.
```

casperklein (dismissed):

polarathene (changes requested):
My system crashed during this review, I am accessing via phone where thankfully github has persisted what I had.

It's incomplete but will have to do for now. I may be slow to respond due to this event. Please don't merge until I return as config section will need some revision too.

docs/content/config/advanced/podman.md
```diff
@@ -96,6 +96,94 @@ docker compose up -d mailserver
 docker compose ps
 ```
 
+### Dockermailserver as a rootless Quadlet
+
+Podman generate systemd is now deprecated and has been integrated into Quadlet. A quadlet is a ini-file with the same structure as a systemd unit file. Writing a Quadlet and placing it on your system will generate a systemd-service on your machine for containers, pods, volumes, networks, kubes, images and builds. Quadlets created as root will then auto-start.
```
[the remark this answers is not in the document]
RoelSG: Great addition, thank you

georglauterbach:
@RoelSG please also add a short entry in our `CHANGELOG.md` :)

polarathene:
I'll have time for returning to this within 48 hours. I need to spin up a VPS and verify some of the contribution to complete my review.

docs/content/config/advanced/podman.md
```diff
@@ -96,6 +96,107 @@ docker compose up -d mailserver
...
+Wants=network-online.target 
+After=network-online.target
+
+[Container]
+ContainerName=dms
+Image=docker.io/mailserver/docker-mailserver:latest 
+# DMS uses uid 5000 for mailstate, but creates other folders for different users, which will be mapped to different sub-uids
+UIDMap=5000:0:1
+UIDMap=0:1:5000
+UIDMap=5001:5001:60536
```
polarathene, line 161: Awaiting your feedback before I make a suggestion here. I'd still like to document `UIDMap` since it did take a while to grok and make sense of and I've seen others confused by it when I was looking into how the mapping works (_volumes also map in reverse which was a bit unexpected, so at one point my host user resolved to intermediate ID `0` which then mapped to container ID 1 (`daemon`), or in the case of config above 5000 (`docker`)_).

Also, I'm not sure of the exact cause but I noticed that `UIDMap=+0:0` and `UIDMap=+0:1` were quick to start, but when the intermediate ID was higher than that there was a 25-30 second delay to container startup, presumably some sort of `chown` (this is on SSD storage too)? Your suggested config here with mapping `5000` to `0` also triggers that IIRC (_not this specific mapping, but something about the other lines that come after_).

Seems to be related to range mapping, as you can do a single mapping for `0` intermediate ID just fine, except for a `supervisord` failure with it's socket (_but this seems to be resolved when `65534` for `nobody` / `nogroup` ID is mapped too_).

polarathene (changes requested):
I still need to wrap up some notes on the config section, then I'll provide some revision for the related docs. Sharing what I have so far in the meantime.

docs/content/config/advanced/podman.md
polarathene: **TL;DR:** You can ignore the bulk of this, it's mostly for me or anyone else that needs the additional context if the topic comes up again in future.

`UIDMap=+0:@%U` seems like the better choice to default to for rootless podman needs.

---

## `UIDMap` notes | Mapping behaviour + chown delay and storage accumulation

It only seems like the delay was a one off hit per mapping. Each one was copying the image (700MB+), presumably entirely remapped to `~/.local/share/containers/storage/overlay`.

Here is the initial pull without any `UIDMap` involved:

image

```
UIDMap=5000:0:1
UIDMap=0:1:5000
UIDMap=5001:5001:60536
```

Is equivalent to `UIDMap=+5000:@%U` in mapping and thus once one is used, switching config to the other won't have startup delay as the overlay copy is already handled.

```console
$ podman exec -it dms cat /proc/self/uid_map

         0          1       5000
      5000          0          1
      5001       5001      60536
```

---

### ClamAV (UID 200) example | When mapping triggers a larger chown copy

The overhead from that mapping results in a overlay squashed layer weighing `762.56 MiB` as opposed to `538.51 MiB` for `UIDMap=+0:@%U`, with the main difference in weight being the ClamAV layer (223 MiB) not being included in the squashed overlay diff.

```console
$ podman exec -it dms bash -c 'ls -l /var/lib/ | grep clamav && ls -l /var/lib/clamav'

drwxr-xr-x. 1 clamav       clamav       4096 Jun  3 01:07 clamav
total 228992
-rw-r--r--. 1 clamav clamav    289733 Jun  3 01:07 bytecode.cvd
-rw-r--r--. 1 clamav clamav  63700823 Jun  3 01:06 daily.cvd
-rw-r--r--. 1 clamav clamav 170479789 Jun  3 01:06 main.cvd
```

`clamav` user and group have UID/GID of `200` within the container.

```console
# An extra `chown` copy happens with `UIDMap=+0:201`:
$ podman exec -it dms cat /proc/self/uid_map

         0        201          1
         1          0          1
         2          1        200
       202        202      65335

# As opposed to `UIDMap=+0:200`:
$ podman exec -it dms cat /proc/self/uid_map
         0        200          1
         1          0          1
         2          1        199
       201        201      65336

# Or  `UIDMap=+0:199`:
$ podman exec -it dms cat /proc/self/uid_map
         0        199          1
         1          0          1
         2          1        198
       200        200      65337
```

- The separate `1 0 1` mapping I think is added when you map with `+` prefix and seems a bit redundant in this case. I think it's implemented to handle the container root ID separately to avoid this error `container ID 0 cannot be mapped to a host ID`, while also handling any before/after mapping.
- `UIDMap=+0:201` => The `2 1 200` line covers UID 200 (clamav) thus it is mapped with an offset requiring a `chown`.
- `UIDMap=+0:200` => AFAIK there is no mapping for container UID 200 to any host ID, so it is unaffected.
-  `UIDMap=+0:199` => There is a mapping now, but it's direct `200` => `200`, it will use the existing overlay layer (_created during original pull_) which has that ownership with `chown` already applied for that offset:

   ```console
   $ ls -l /home/linuxuser/.local/share/containers/storage/overlay/ccab93e94733e94b645eb852f9bfa7471955cb1d184ef6224fbe51980d30e767/diff/var/lib/clamav
   total 228980
   -rw-r--r--. 1 524487 524487    289733 Jun  3 01:07 bytecode.cvd
   -rw-r--r--. 1 524487 524487  63700823 Jun  3 01:06 daily.cvd
   -rw-r--r--. 1 524487 524487 170479789 Jun  3 01:06 main.cvd
   ```
   
   `524487` is the base `524288` + (200 -1). UID 0 technically maps to 1000 (`linuxuser` in this case), then from UID offset of 1 in `/etc/subuid` on the host we increment from `524288`, hence container UID `200` maps to host UID `524487`:
   
   ```console
   $ cat /etc/subuid
   
   linuxuser:524288:65536
   ```

---

### Default mapping vs `UIDMap=+0:@1000`

For reference this is the default without any `UIDMap` setting:

```console
$ podman exec -it dms cat /proc/self/uid_map

         0       1000          1
         1     524288      65536
```

That is valid for rootful where the 2nd column would align with the host UID, but in rootless the 2nd column is a 
"intermediate ID" that is mapped as an offset for the user host ID entry in `/etc/subuid`.

Hence the `@1000` syntax for `UIDMap` which would be the equivalent for rootless (mapping to the `0` offset), so `UIDMap=+0:@1000` would get us the desired rootless mapping:

```console
$ podman exec -it dms cat /proc/self/uid_map

         0          0          1
         1          1      65536
```

### `chown` storage copies cleanup

The extra copies of the image at `~/.local/share/containers/storage/overlay` are tied to the image, so when all references are removed via `podman image rm` or `podman image prune -a` / `podman system prune -a` (_assuming no container is running with the image_).

I'm not aware of how you can clear that more selectively than that (_`podman run --rm --uidmap ...` is the same with accumulation_).

docs/content/config/advanced/podman.md
```diff
@@ -96,6 +96,107 @@ docker compose up -d mailserver
...
+[podman::rootless-differences]: https://matduggan.com/replace-compose-with-quadlet/#rootless
+[network-driver::pasta]: https://passt.top/passt/about/#pasta
+[systemd-docs::loginctl::linger]: https://www.freedesktop.org/software/systemd/man/latest/loginctl.html#enable-linger%20USER%E2%80%A6
+[podman-docs::quadlet::config-search-path]: https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html#podman-rootful-unit-search-path
+
+
+#### Example Quadlet file
+
+We have to use the .container extension for the quadlet generator to pick up the service.
+Because docker-mailserver uses multiple users inside the container, we will either have to use our own user as root, resulting in our e-mails being owned by a subuid. Alternatively, using UIDMap we can map our rootless user to UID 5000 in the container who owns our e-mails. Using UIDMap also maps root user 0 inside the container to an available sub-uid of our rootless user. Otherwise the container will not have permission to configure itself.
```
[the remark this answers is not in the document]
RoelSG: The choice is arbitrary in the end. I wanted the mail-storage mount to be owned by the rootless user starting the container, and have root inside the container not directly mapped to the user starting the container.

I spent a long time too messing around with uid-maps, and it solves a lot of problems that rootless podman-containers seem to have for beginners. I seem to have learned even more from your findings, thanks!

docs/content/config/advanced/podman.md
```diff
@@ -96,6 +96,107 @@ docker compose up -d mailserver
...
+[Install]
+WantedBy=default.target
+
+[Unit]
+Wants=network-online.target 
+After=network-online.target
+
+[Container]
+ContainerName=dms
+Image=docker.io/mailserver/docker-mailserver:latest
```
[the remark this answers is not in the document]
RoelSG: Seems to be some legacy from the config I had before.

docs/content/config/advanced/podman.md
```diff
@@ -96,6 +96,107 @@ docker compose up -d mailserver
...
+
+# If you want to use podmans auto-update service:
+AutoUpdate=registry 
+
+# Environment variables
+# General Settings
+
+Environment=HOSTNAME=example
+Environment=DOMAINNAME=example.com
+Environment=CONTAINER_NAME=dockermailserver
```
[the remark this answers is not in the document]
RoelSG: I seem to remember this was an older way of doing it, got left over from before.

docs/content/config/advanced/podman.md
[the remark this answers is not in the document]
polarathene: > The choice is arbitrary in the end.

Yes, agreed. Just curious why you might have preferred this choice.

> I wanted the mail-storage mount to be owned by the rootless user starting the container, and have root inside the container not directly mapped to the user starting the container.

Ok, but no practical reason for that right? Since you'd rarely need access to this content outside of the container (_beyond persistent storage_) and you're more likely as a user to want to have ownership of `root` files for the Config Volume yeah? (_similar benefit for logs too_).

`idmap` option for `Volume` settings would be useful for such, but with rootless it's just not available unfortunately.

If you did have a legitimate use-case for preferring `5000` mapped to your host user that would be good to know. I'll otherwise switch to `UIDMap=+0:@%U` for the official advice, with some context added to the docs that I've touched on above and in my other review feedback, with `UIDMap=+5000:@%U` as an alternative example.

> I seem to have learned even more from your findings, thanks!

You're welcome!

---

I'll try allocate some time tomorrow for the final revision and we can then get this merged! 😁

polarathene:
Apologies for the delay, I had a stressful end of October that led to a burn out. I've committed the bulk of what I had managed prior to the burn out.

I'll go over the current state as I recall I had a few more parts I wanted to address and polish up, then we can get this approved and merged 🚀

polarathene:
Sorry about the lack of activity here, bit pre-occupied elsewhere but will return to this in the 2nd half of the month.

RoelSG:
Not a problem, this is not a bugfix or something, I'll check in periodically.

docs/content/config/advanced/podman.md
```diff
@@ -96,6 +96,181 @@ docker compose up -d mailserver
...
+#Environment=NETWORK_INTERFACE=tap0
+```
+
+!!! info "Systemd specifiers"
+
+    Systemd has a [variety of specifiers][systemd-docs::config-specifiers] (_prefixed with `%`_) that help manage configs.
+    
+    Here are the ones used from the above example:
+
+    - ***`%h`:** Location of the users home directory. Use this instead of `~` (_which would only work in a shell, not this config_).
```
polarathene, line 196:
```suggestion
    - **`%h`:** Location of the users home directory. Use this instead of `~` (_which would only work in a shell, not this config_).
```

docs/content/config/advanced/podman.md
```diff
@@ -96,6 +96,181 @@ docker compose up -d mailserver
...
+
+    The most common case is to map the containers root user (UID `0`) to your host user ID.
+
+    For a rootless user with the UID `1000` on the host, any of the following `UIDMap` values are equivalent:
+
+    - **`UIDMap=+0:0`:** The 1st `0` is the container root ID and the 2nd `0` refers to host mapping ID. For rootless the mapping ID is an indirect offset to their user entry in `/etc/subuid` where `0` maps to their host user ID.
+    - **`UIDMap=+0:@1000`:** A rootless Quadlet can also use `@` as a prefix which tells Podman to infer the mapping ID from `/etc/subuid`.
+    - **`UIDMap=+0:@%U`:** Instead of providing the explicit rootless UID, a better approach is to leverage `%U` (_a [systemd specifier][systemd-docs::config-specifiers]_) which will resolve to the UID of your rootless user that starts the Quadlet service.
+
+??? tip "What is the `+` syntax?"
```
polarathene, line 236:
```suggestion
??? tip "Why should I prefer `UIDMap=+0:@%U`? How does the `@` syntax work?"

    The most common case is to map the containers root user (UID `0`) to your host user ID.

    For a rootless user with the UID `1000` on the host, any of the following `UIDMap` values are equivalent:

    - **`UIDMap=+0:0`:** The 1st `0` is the container root ID and the 2nd `0` refers to host mapping ID. For rootless the mapping ID is an indirect offset to their user entry in `/etc/subuid` where `0` maps to their host user ID, while `1` or higher maps to the users subuid range.
    - **`UIDMap=+0:@1000`:** A rootless Quadlet can also use `@` as a prefix which Podman will then instead lookup as the host ID in `/etc/subuid` to get the offset value. If the host user ID was `1000`, then `@1000` would resolve that to `0`.
    - **`UIDMap=+0:@%U`:** Instead of providing the explicit rootless UID, a better approach is to leverage `%U` (_a [systemd specifier][systemd-docs::config-specifiers]_) which will resolve to the UID of your rootless user that starts the Quadlet service.

??? tip "What is the `+` syntax used with `UIDMap`?"
```

docs/content/config/advanced/podman.md
```diff
@@ -96,6 +96,181 @@ docker compose up -d mailserver
...
+# https://docker-mailserver.github.io/docker-mailserver/latest/config/environment/#network_interface
+#Environment=NETWORK_INTERFACE=enp1s0
+#Environment=NETWORK_INTERFACE=tap0
+```
+
+!!! info "Systemd specifiers"
+
+    Systemd has a [variety of specifiers][systemd-docs::config-specifiers] (_prefixed with `%`_) that help manage configs.
+    
+    Here are the ones used from the above example:
```
polarathene, line 194:
```suggestion
??? info "Systemd specifiers"

    Systemd has a [variety of specifiers][systemd-docs::config-specifiers] (_prefixed with `%`_) that help manage configs.
    
    Here are the ones used in the Quadlet config example:
```

docs/content/config/advanced/podman.md
```diff
@@ -96,6 +96,181 @@ docker compose up -d mailserver
...
+    Rootless containers [perform a copy of the image with `chown`][caveat::podman::rootless::image-chown] during the first pull/run of the image.
+
+    - The larger the image to copy, the longer the initial delay on first use.
+    - This process will be repeated if the `UIDMap` / `GIDMap` settings are changed to a value that has not been used previously (_accumulating more disk usage with additional image copies_).
+    - Only when the original image is removed will any of these associated `chown` image copies be purged from storage.
+
+    When you specify a `UIDMap` like shown in the prior tip for the `+` syntax with `UIDMap=+0:5000`, if the `/proc/self/uid_map` shows a row with the first two columns as equivalent then no excess `chown` should be applied.
+
+    - `UIDMap=+0:@%U` is equivalent from ID 2 onwards.
+    - `UIDMap=+5000:@%U` is equivalent from ID 5001 onwards. This is relevant with DMS as the container UID 200 is assigned to ClamAV, the offset introduced will now incur a `chown` copy of 230MB.
```
polarathene, line 272:
```suggestion
??? warning "Impact on disk usage of images with Rootless"

    **NOTE:** This should not usually be a concern, but has been documented here to explain the impact of creating new user namespaces (_such as by running a container with settings like `UIDMap` that differ between runs_).

    ---

    Rootless containers [perform a copy of the image with `chown`][caveat::podman::rootless::image-chown] during the first pull/run of the image.

    - The larger the image to copy, the longer the initial delay on first use.
    - This process will be repeated if the `UIDMap` / `GIDMap` settings are changed to a value that has not been used previously (_accumulating more disk usage with additional image layer copies_).
    - Only when the original image is removed will any of these associated `chown` image copies be purged from storage.

    When you specify a `UIDMap` like demonstrated in the earlier tip for the `+` syntax with `UIDMap=+0:5000`, if the `/proc/self/uid_map` shows a row with the first two columns as equivalent then no excess `chown` should be applied.

    - `UIDMap=+0:@%U` is equivalent from ID 2 onwards.
    - `UIDMap=+5000:@%U` is equivalent from ID 5001 onwards. This is relevant with DMS as the container UID 200 is assigned to ClamAV, the offset introduced will now incur a `chown` copy of 230MB.
```

docs/content/config/advanced/podman.md
```diff
@@ -96,6 +96,181 @@ docker compose up -d mailserver
...
+        ```bash
+        loginctl enable-linger user
+        ```
+    - [Config locations between rootful vs rootless][podman-docs::quadlet::config-search-path].
+
+#### Example Quadlet file
+
+1. Create your DMS Quadlet at `~/.config/containers/systemd/dms.container` with the example content shown below.
+2. Run [`systemctl --user daemon-reload`][systemd-docs::systemctl::daemon-reload], which will trigger the Quadlet service generator (_required whenever you adjust config in `dms.container`_).
+3. You should now be able to start the service with `systemctl --user start dms`.
```
polarathene, line 127:
```suggestion
???+ example

    1. Create your DMS Quadlet at `~/.config/containers/systemd/dms.container` with the example content shown below.
        - Adjust settings like `HostName` as needed. You may prefer a different convention for your `Volume` host paths.
        - Some syntax like systemd specifiers and Podman's `UIDMap` value are explained in detail after this example.
    2. Run [`systemctl --user daemon-reload`][systemd-docs::systemctl::daemon-reload], which will trigger the Quadlet service generator. This command is required whenever you adjust config in `dms.container`.
    3. You should now be able to start the service with `systemctl --user start dms`.
```

polarathene (changes requested):
Going through stuff I've had open, came across these changes which AFAIK were going towards a final revision until I had to put that on hold. I'll apply what I have here as at a glance it does seem to be an improvement overall.

I am relocating presently but still plan to investigate the concern with network drivers I had so that this can finally get wrapped up (_hopefully later this month_ 😓)

docs/content/config/advanced/podman.md
```diff
@@ -96,6 +96,185 @@ docker compose up -d mailserver
...
+# https://docs.podman.io/en/latest/markdown/podman-run.1.html#tz-timezone
+# NOTE: Should the host modify the system TZ, neither approach will sync the change to the `/etc/localtime` inside the running container.
+Timezone=local
+
+Environment=SSL_TYPE=letsencrypt
+# NOTE: You may need to adjust the default `NETWORK_INTERFACE`:
+# https://docker-mailserver.github.io/docker-mailserver/latest/config/environment/#network_interface
+#Environment=NETWORK_INTERFACE=enp1s0
+#Environment=NETWORK_INTERFACE=tap0
+```
```
polarathene, line 192: Just indentation change here. Committing separately.

```suggestion
???+ example

    1. Create your DMS Quadlet at `~/.config/containers/systemd/dms.container` with the example content shown below.
        - Adjust settings like `HostName` as needed. You may prefer a different convention for your `Volume` host paths.
        - Some syntax like systemd specifiers and Podman's `UIDMap` value are explained in detail after this example.
    2. Run [`systemctl --user daemon-reload`][systemd-docs::systemctl::daemon-reload], which will trigger the Quadlet service generator. This command is required whenever you adjust config in `dms.container`.
    3. You should now be able to start the service with `systemctl --user start dms`.

    ```ini title="dms.container"
    [Unit]
    Description="Docker Mail Server"
    Documentation=https://docker-mailserver.github.io/docker-mailserver/latest

    [Service]
    Restart=always
    # Optional - This will run before the container starts:
    # - It ensures all the DMS volumes have the host directories created for you.
    # - For `mkdir` command to leverage the shell brace expansion syntax, you need to run it via bash.
    ExecStartPre=/usr/bin/bash -c 'mkdir -p %h/volumes/%N/{mail-data,mail-state,mail-logs,config}'

    # This section enables the service at generation, avoids requiring `systemctl --user enable dms`:
    # - `multi-user.target` => root
    # - `default.target` => rootless
    [Install]
    WantedBy=default.target

    [Container]
    ContainerName=%N
    HostName=mail.example.com
    Image=docker.io/mailserver/docker-mailserver:latest

    PublishPort=25:25
    PublishPort=143:143
    PublishPort=587:587
    PublishPort=993:993

    # The container UID for root will be mapped to the host UID running this Quadlet service.
    # All other UIDs in the container are mapped via the sub-id range for that user from host configs `/etc/subuid` + `/etc/subgid`.
    UIDMap=+0:@%U

    # Volumes (Base location example: `%h/volumes/%N` => `~/volumes/dms`)
    # NOTE: If your host has SELinux enabled, avoid permission errors by appending the mount option `:Z`.
    Volume=%h/volumes/%N/mail-data:/var/mail
    Volume=%h/volumes/%N/mail-state:/var/mail-state
    Volume=%h/volumes/%N/mail-logs:/var/log/mail
    Volume=%h/volumes/%N/config:/tmp/docker-mailserver
    # Optional - Additional mounts:
    # NOTE: For SELinux, when using the `z` or `Z` mount options:
    #   Take caution if choosing a host location not belonging to your user. Consider using `SecurityLabelDisable=true` instead.
    #   https://docs.podman.io/en/latest/markdown/podman-run.1.html#volume-v-source-volume-host-dir-container-dir-options
    Volume=%h/volumes/certbot/certs:/etc/letsencrypt:ro

    # Podman can create a timer (defaults to daily at midnight) to check the `registry` or `local` storage for detecting if the
    # image tag points to a new digest, if so it updates the image and restarts the service (similar to `containrrr/watchtower`):
    # https://docs.podman.io/en/latest/markdown/podman-auto-update.1.html
    AutoUpdate=registry

    # Podman Quadlet has a better alternative instead of a volume directly bind mounting `/etc/localtime` to match the host TZ:
    # https://docs.podman.io/en/latest/markdown/podman-run.1.html#tz-timezone
    # NOTE: Should the host modify the system TZ, neither approach will sync the change to the `/etc/localtime` inside the running container.
    Timezone=local

    Environment=SSL_TYPE=letsencrypt
    # NOTE: You may need to adjust the default `NETWORK_INTERFACE`:
    # https://docker-mailserver.github.io/docker-mailserver/latest/config/environment/#network_interface
    #Environment=NETWORK_INTERFACE=enp1s0
    #Environment=NETWORK_INTERFACE=tap0
    ```
```

georglauterbach:
I resolved the merge conflicts for this PR, as I'd like to get it merged. @polarathene what's the status here - can we ship this update? I do not need everything to be polished right away - if the gist is good, we can build on that.

polarathene:
I am presently unable to context shift back to this PR, there may be some concerns with the current state of it where I wanted to revise the content or verify.

I recall some claims regarding the networking drivers that might be inaccurate in the scope of the two discussed or rootless vs rootful contexts. So while I'm a little iffy about merging an incomplete update, I still think this information is valuable to our users and we'll need to defer to future contributions to correct any oversights as I'm unsure when I'll have time to prioritize this on the backlog.

I will merge the PR in it's current state and add a reference to my TODO issue.

polarathene (approved):
Thanks for the contribution, I've refined it as best I could with the time I had to spare.

The information should be quite helpful to Podman users and for maintainers to grok when necessary.

Issue #3778, opened 2024-01-14
feat: use `selector_map` for DKIM with Rspamd


polarathene:
### Description

Referencing [my review comment here](https://github.com/docker-mailserver/docker-mailserver/pull/3702/files#r1428935477), [`selector_map`](https://rspamd.com/doc/modules/dkim_signing.html#using-maps) might be a good approach for automating generation of multiple DKIM selectors when necessary. It's unlikely our path convention would change, and as per the PR contribution, a standard path convention / template can use variables for the _selector_ (_with a default fallback_) and _domain_.

```
# local.d/dkim_signing.conf
try_fallback = true;
path = "/tmp/docker-mailserver/dkim/keys/$domain/$selector.private";
selector_map = "/etc/rspamd/dkim_selectors.map";
selector = "mail";
```

- `try_fallback = true;` AFAIK is to permit fallback to that default `selector` variable.
- `path` is the template used, where `$domain` is the domain being handled (_[defaults to MIME `From` (message header)](https://github.com/docker-mailserver/docker-mailserver/pull/3702/files#r1428929641), not SMTP `From` mail header (envelope sender) although you can configure for that_), and `$selector` will either be `mail` from above, or changed to a matching selector if found in `selector_map`.
- A `selector_map` is just a key/value pair per line delimiting key/values by white-space. If a domain matches, it'll use that selector:
  
  ```
  example.com dkim-example
  another-domain.com another-selector
  ```

---

## Additional Notes

I'm not 100% sure how that works for domains with multiple selectors available. Such as to support both RSA and ECC DKIM keys.

The `path` example uses the Docker config volume directly, but would probably be an internal location that the files are copied to. It's ideally unified with OpenDKIM config to share a common location for keys as these should technically be no different. It's a question raised by users several times about migrating to rspamd if they need to generate keys again or can use the same ones created for OpenDKIM without much friction.

- `$domain` is additionally derived from `use_domain` (_what part of the mail to source the domain from, or a fixed domain_)
- [`use_esld = true`](https://github.com/docker-mailserver/docker-mailserver/pull/3702#issuecomment-1863480339) (_normalize domains to [eSLD](https://en.wikipedia.org/wiki/Domain_name#Second-level_and_lower_level_domains) (effective second-level domain))
  - [Drops any subdomains](https://github.com/rspamd/rspamd/issues/1808#issuecomment-324560663) when matching DKIM configured domains to use by the sender domain received by mail to be conditionally signed.
[...]

PR #4201, opened 2024-09-28
scripts: improve DKIM path scanning in Rspamd setup

### LINKED ISSUE #4172: bug report: Rspamd DKIM private key file '/tmp/docker-mailserver/rspamd/dkim/$domain/$selector.private' is configured for usage, but does not appear to exist
### 📝 Preliminary Checks

### 👀 What Happened?

Well this is not really a bug as this does not impact DMS operation. The following warning message is emitted at startup when using a Rspamd `path` syntax using the `$domain` and `$selector` variables (see [Rspamd configuration](https://rspamd.com/doc/modules/dkim_signing.html#configuration) or PR #3702 that documents it):
```
Rspamd DKIM private key file '/tmp/docker-mailserver/rspamd/dkim/$domain/$selector.private' is configured for usage, but does not appear to exist
```
It could be nice not to emit such message.

Our `dkim_signing.conf` file contains the following line:
```
path = "/tmp/docker-mailserver/rspamd/dkim/$domain/$selector.private";
```

From what I found in the code, it seems to come from the [`__rspamd__check_dkim_permissions()`](https://github.com/docker-mailserver/docker-mailserver/blob/v14.0.0/target/scripts/startup/setup.d/security/rspamd.sh#L344-L347) function.

### 👟 Reproduction Steps

_No response_

### 🐋 DMS Version

v14.0.0

Debian 12

_No response_

### 📜 Relevant log output

_No response_

_No response_

georglauterbach:
# Description

This PR improves the algorithm to scan for DKIM paths in Rspamd'd configuration. Moreover, it fixes a small inconvenience: `cp` would output "Nothing to copy" because the directory contained no contents; this has been fixed as well.

Fixes #4172

## Type of change

## Checklist

### DIFF
--- a/target/scripts/startup/setup.d/security/rspamd.sh
+++ b/target/scripts/startup/setup.d/security/rspamd.sh
@@ -319,8 +319,7 @@ function __rspamd__setup_check_authenticated() {
   local MODULE_FILE="${RSPAMD_LOCAL_D}/settings.conf"
   readonly MODULE_FILE
   if _env_var_expect_zero_or_one 'RSPAMD_CHECK_AUTHENTICATED' \
-  && [[ ${RSPAMD_CHECK_AUTHENTICATED} -eq 0 ]]
-  then
+  && [[ ${RSPAMD_CHECK_AUTHENTICATED} -eq 0 ]]; then
     __rspamd__log 'debug' 'Content checks for authenticated users are disabled'
   else
     __rspamd__log 'debug' 'Enabling content checks for authenticated users'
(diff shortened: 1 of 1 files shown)


polarathene:
I'm happy to approve the change, but I do recall suggesting an alternative back in Dec 2023: https://github.com/docker-mailserver/docker-mailserver/pull/3702#discussion_r1428935477

I've not refreshed myself on the subject, but I did document a way forward here: https://github.com/docker-mailserver/docker-mailserver/issues/3778

If you have the time to have a quick look over that and let me know if that's potentially a better approach, and if it's acceptable for v15 or should be delayed, that'd be good. It may be a bit of a breaking change / inconvenience, but might be worth considering before switching rspamd to the default (_which I assume may also be blocked on adopting Valkey to minimize the wider impact of anyone that has been relying on the internal Redis service?_).

target/scripts/startup/setup.d/security/rspamd.sh
```diff
@@ -332,32 +332,22 @@ function __rspamd__setup_check_authenticated() {
...
+  local KEY_FILE
+  while read -r KEY_FILE; do
+    if [[ -f ${KEY_FILE} ]]; then
+      __rspamd__log 'trace' "Checking DKIM file '${KEY_FILE}'"
       # See https://serverfault.com/a/829314 for an explanation on `-exec false {} +`
       # We additionally resolve symbolic links to check the permissions of the actual files
-      if find "$(realpath -eL "${FILE}")" \( -user _rspamd -or -group _rspamd -or -perm -o=r \) -exec false {} +; then
-        __rspamd__log 'warn' "Rspamd DKIM private key file '${FILE}' does not appear to have correct permissions/ownership for Rspamd to use it"
+      if find "$(realpath -L "${KEY_FILE}")" \( -user _rspamd -or -group _rspamd -or -perm -o=r \) \
+        -exec false {} +; then
```
polarathene, line 344: Any reason for the line split? Was it just width?

The concern handled here would be a non-issue if the keys were copied to an internal path right?

target/scripts/startup/setup.d/security/rspamd.sh
```diff
@@ -332,32 +332,22 @@ function __rspamd__setup_check_authenticated() {
...
       else
-        __rspamd__log 'trace' "DKIM file '${FILE}' permissions and ownership appear correct"
+        __rspamd__log 'trace' "DKIM file '${KEY_FILE}' permissions and ownership appear correct"
       fi
     else
-      __rspamd__log 'warn' "Rspamd DKIM private key file '${FILE}' is configured for usage, but does not appear to exist"
+      __rspamd__log 'warn' "Rspamd DKIM private key file '${KEY_FILE}' is configured for usage, but does not appear to exist"
     fi
-  done
+  done < <(rspamadm configdump dkim_signing | grep 'path =' | grep -v -F '$' | awk '{print $3}' | tr -d ';"')
```
polarathene, line 352: So the config has multiple `path` values to extract and check. It seems with the mentioned `selector_map` config, none of that would be necessary. All keys would stored with predictable hierarchy:

```
selector = "mail";
# These paths should technically be internal locations instead,
# with data sourced from the users config volume (/tmp/docker-mailserver):
path = "/tmp/docker-mailserver/rspamd/dkim/keys/$domain/$selector.private";
selector_map = "/tmp/docker-mailserver/rspamd/dkim/selector_map";
```

The `selector_map` file then maps domain with a selector:

```
example.com dkim-example
another-domain.com another-selector
```

So I think the `selector_map` config approach would make all of this logic redundant 🤔

polarathene (approved):
LGTM. Go with this fix for now 🚀 

I'd still appreciate your opinion on adopting `selector_map` in future before rspamd becomes default. That'll probably need some doc changes and migration advice in changelog for existing users, which I know is a hassle 😅 (_but getting it in before the release that defaults to rspamd would be good for testing/feedback_)

target/scripts/startup/setup.d/security/rspamd.sh
georglauterbach: > Any reason for the line split? Was it just width?

Yes :)
 
> The concern handled here would be a non-issue if the keys were copied to an internal path right?

Indeed!

target/scripts/startup/setup.d/security/rspamd.sh
georglauterbach: I would need to have another look at your proposed, in-depth, but I guess you are right, your approach would simplify and unify our setup.

georglauterbach:
> LGTM. Go with this fix for now 🚀

Just a reminder: this PR is a fix for #4172 only, not a rewrite of the DKIM functionality :)

> I'd still appreciate your opinion on adopting `selector_map` in future before rspamd becomes default. That'll probably need some doc changes and migration advice in changelog for existing users, which I know is a hassle 😅 (_but getting it in before the release that defaults to rspamd would be good for testing/feedback_)

I think you are absolutely right; when I got some time in the future, I will tackle your proposals.

PR #2640, opened 2022-06-14
fix: Dovecot Quota dummy accounts for aliases should check for existing users with an exact user key lookup

tomav:
# Description

Fixes #2639

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [ ] If necessary I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #4222, opened 2024-10-10
fix: Avoid alias being used as regex during dovecot dummy account userdb detection

### LINKED ISSUE #4170: bug report: start-mailserver.sh says the alias will not be added twice
### 📝 Preliminary Checks

### 👀 What Happened?

Hi DMS maintainers,

I'm helping manage a DMS instance and for a while we've seen the following line in our log at startup:
```
WARN  start-mailserver.sh: Alias 'first.name@domain.example' will not be added to '/etc/dovecot/userdb' twice
```
After investigation, it appears this alias is not duplicated in the `/tmp/docker-mailserver/postfix-virtual.cf` file. However this file contains the following 2 lines (in this order):
```
first-name@domain.example account1@anotherdomain.example
first.name@domain.example account2@anotherdomain.example
```
Could it be that the grep check in [`_create_dovecot_alias_dummy_accounts()`](https://github.com/docker-mailserver/docker-mailserver/blob/v14.0.0/target/scripts/helpers/accounts.sh#L138) is missing the `-F` option? Would the dot `.` in `first.name` then be interpreted as any character matching the previous entry `first-name@domain.example`?

I haven't run any test since this is our production server and I haven't taken the time to install a clean DMS instance to try reproducing it.

### 👟 Reproduction Steps

_No response_

### 🐋 DMS Version

v14.0.0

Debian 12

_No response_

### 📜 Relevant log output

_No response_

_No response_

pitilux:
# Description

Fixes #4170 and test coverage of #2640.

This is my first PR, let me know if I missed something. And regarding the CHANGELOG, would the following line be ok?
```
Fixes
- Dovecot:
   - Fix dummy account creation (PR number)
```

## Type of change

## Checklist

### DIFF
--- a/target/scripts/helpers/accounts.sh
+++ b/target/scripts/helpers/accounts.sh
@@ -135,7 +135,7 @@ function _create_dovecot_alias_dummy_accounts() {
       fi
 
       DOVECOT_USERDB_LINE="${ALIAS}:${REAL_ACC[1]}:${DMS_VMAIL_UID}:${DMS_VMAIL_GID}::/var/mail/${REAL_DOMAINNAME}/${REAL_USERNAME}/home::${REAL_ACC[2]:-}"
-      if grep -qi "^${ALIAS}:" "${DOVECOT_USERDB_FILE}"; then
+      if grep -qxF "${DOVECOT_USERDB_LINE}" "${DOVECOT_USERDB_FILE}"; then
         _log 'warn' "Alias '${ALIAS}' will not be added to '${DOVECOT_USERDB_FILE}' twice"
       else
         echo "${DOVECOT_USERDB_LINE}" >>"${DOVECOT_USERDB_FILE}"
--- a/target/scripts/helpers/accounts.sh
+++ b/target/scripts/helpers/accounts.sh
@@ -135,7 +135,8 @@ function _create_dovecot_alias_dummy_accounts() {
       fi
 
       DOVECOT_USERDB_LINE="${ALIAS}:${REAL_ACC[1]}:${DMS_VMAIL_UID}:${DMS_VMAIL_GID}::/var/mail/${REAL_DOMAINNAME}/${REAL_USERNAME}/home::${REAL_ACC[2]:-}"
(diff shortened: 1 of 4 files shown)


target/scripts/helpers/accounts.sh
```diff
@@ -135,7 +135,7 @@ function _create_dovecot_alias_dummy_accounts() {
       fi
 
       DOVECOT_USERDB_LINE="${ALIAS}:${REAL_ACC[1]}:${DMS_VMAIL_UID}:${DMS_VMAIL_GID}::/var/mail/${REAL_DOMAINNAME}/${REAL_USERNAME}/home::${REAL_ACC[2]:-}"
-      if grep -qi "^${ALIAS}:" "${DOVECOT_USERDB_FILE}"; then
+      if grep -qxF "${DOVECOT_USERDB_LINE}" "${DOVECOT_USERDB_FILE}"; then
```
polarathene, line 138: I don't think there's any harm in using `-i` still. We don't want to introduce aliases that differ by casing as that tends to introduce bugs.

```suggestion
      # Match a full line with `-xF` to avoid regex patterns introducing false positives matching `ALIAS`:
      if grep -qixF "${DOVECOT_USERDB_LINE}" "${DOVECOT_USERDB_FILE}"; then
```

polarathene (changes requested):
Thanks for taking time to tackle this 😁

polarathene:
> regarding the CHANGELOG, would the following line be ok?
> 
> ```
> Fixes
> - Dovecot:
>    - Fix dummy account creation (PR number)
> ```

Something like this should be good 👍
```markdown
### Fixes

- **Dovecot:**
  - The Dovecot Quota support "dummy account" workaround no longer treats the alias as a regex when checking the Dovecot UserDB ([#4222](https://github.com/docker-mailserver/docker-mailserver/pull/4222))
```

The first few lines are already there:

https://github.com/docker-mailserver/docker-mailserver/blob/26a44995a9bba825b1348dbf3551687c2100f7a3/CHANGELOG.md?plain=1#L44-L46

pitilux:
Thanks for the review. I applied all your suggested changes.

polarathene (approved):
LGTM! 😎

PR #4243, opened 2024-10-29
doc: add an example for using the keytype ed25519 when using DKIM & Rspamd

### LINKED ISSUE #4180: bug report: built-in documentation does not give an example for rspamd-dkim's keytype
### 📝 Preliminary Checks

### 👀 What Happened?

Under rspamd-enable, the call chain for `docker exec -it <CONTAINER NAME> setup config dkim` (as recommended) goes like:
* https://github.com/docker-mailserver/docker-mailserver/blob/cace9c56d932bbd53a36f5bc36f21a676d92a708/target/bin/setup#L134
* https://github.com/docker-mailserver/docker-mailserver/blob/cace9c56d932bbd53a36f5bc36f21a676d92a708/target/bin/open-dkim#L10
* https://github.com/docker-mailserver/docker-mailserver/blob/cace9c56d932bbd53a36f5bc36f21a676d92a708/target/bin/rspamd-dkim

Each of these scripts have their help texts. Of these:

* The first does not mention `keytype` at all. Which is fine, really, since RSA is the main thing and we have the most important ones down.
* The second help text is not invoked on the rspamd path, so let's ignore it.
* The third describes the option, but has no example to go with it.

### 👟 Reproduction Steps

Click the links!

### 🐋 DMS Version

git cace9c56d932bbd53a36f5bc36f21a676d92a708

GitHub source browser

_No response_

### 📜 Relevant log output

_No response_

_No response_

georglauterbach:
# Description

Fixes #4180

## Type of change

## Checklist

No `CHANGELOG.md` entry required.

### DIFF
--- a/target/bin/rspamd-dkim
+++ b/target/bin/rspamd-dkim
@@ -55,6 +55,10 @@ ${ORANGE}EXAMPLES${RESET}
     ${LWHITE}setup config dkim domain example.com${RESET}
         Generate the DKIM key for a different domain (example.com).
 
+    ${LWHITE}setup config dkim keytype ed25519 domain edward.com selector elliptic-test${RESET}
+        Generate the DKIM key using the ED25519 elliptic curve for the domain
+        edward.com and the selector elliptic-test.
+
 ${ORANGE}EXIT STATUS${RESET}
     Exit status is 0 if command was successful. If wrong arguments are provided or arguments contain
     errors, the script will exit early with a non-zero exit status.


target/bin/rspamd-dkim
```diff
@@ -55,6 +55,10 @@ ${ORANGE}EXAMPLES${RESET}
     ${LWHITE}setup config dkim domain example.com${RESET}
         Generate the DKIM key for a different domain (example.com).
 
+    ${LWHITE}setup config dkim keytype ed25519 domain edward.com selector elliptic-test${RESET}
+        Generate the DKIM key using the ED25519 elliptic curve for the domain
+        edward.com and the selector elliptic-test.
```
polarathene, line 60: - Any reason to use `edward.com` instead of `example.com` placeholder?
- A selector example was already provided and probably isn't adding much here?

Revised suggestion if you agree:

```suggestion
    ${LWHITE}setup config dkim keytype ed25519 domain example.com${RESET}
        Generate a DKIM key for the domain 'example.com' but with ED25519 instead of RSA.
```

polarathene (approved):

target/bin/rspamd-dkim
georglauterbach: Please don't hesitate to adjust the code; the PR has been merged with auto-merge. I used "edward.com" because of ed25519 is based on Edward's Elliptical curve; but we can stick with `example.com` and remove the selector.

PR #4247, opened 2024-11-02
ci: Revise `docs-preview-deploy.yml`

pwntester:
# Description

Fixes workflow

### DIFF
--- a/.github/workflows/docs-preview-deploy.yml
+++ b/.github/workflows/docs-preview-deploy.yml
@@ -45,10 +45,10 @@ jobs:
           head_sha: ${{ github.event.workflow_run.head_sha }}
           pull_requests: ${{ tojson(github.event.workflow_run.pull_requests) }}
         run: |
-          PR_NUMBER=$(jq -r '[.[] | select(.head.sha == "${head_sha}")][0].number' <<< '${pull_requests}')
+          PR_NUMBER=$(jq -r '[.[] | select(.head.sha == "${head_sha}")][0].number' <<< "${pull_requests}")
           {
             echo "PR_NUMBER=${PR_NUMBER}"
-            echo 'PR_HEADSHA=${head_sha}'
+            echo "PR_HEADSHA=${head_sha}"
           } >> "${GITHUB_ENV}"
 
       # ==================== #
--- a/.github/workflows/docs-preview-deploy.yml
+++ b/.github/workflows/docs-preview-deploy.yml
@@ -52,7 +52,7 @@ jobs:
           PR_NUMBER=$(jq -r '[.[] | select(.head.sha == "${{ env.head_sha }}")][0].number' <<< "${pull_requests}")
           {
             echo "PR_NUMBER=${PR_NUMBER}"
-            echo "PR_HEADSHA=${head_sha}"
+            echo 'PR_HEADSHA=${{ env.head_sha }}'
           } >> "${GITHUB_ENV}"
 
       # ==================== #
--- a/.github/workflows/docs-preview-deploy.yml
+++ b/.github/workflows/docs-preview-deploy.yml
@@ -17,7 +17,7 @@ jobs:
     name: 'Deploy Preview'
     runs-on: ubuntu-22.04
     # Requires a PR event triggered `docs-preview-prepare.yml` workflow run that was successful + ensure the head SHA belongs to an associated PR:
-    if: |
+    if: >
       ${{
         github.event.workflow_run.conclusion == 'success'
         && github.event.workflow_run.event == 'pull_request'
@@ -47,10 +47,10 @@ jobs:
           head_sha: ${{ github.event.workflow_run.head_sha }}
           pull_requests: ${{ tojson(github.event.workflow_run.pull_requests) }}
         run: |
-          PR_NUMBER=$(jq -r '[.[] | select(.head.sha == "${{ env.head_sha }}")][0].number' <<< '${{ env.pull_requests }}')
+          PR_NUMBER=$(jq -r '[.[] | select(.head.sha == "${head_sha}")][0].number' <<< '${pull_requests}')
           {
             echo "PR_NUMBER=${PR_NUMBER}"
-            echo 'PR_HEADSHA=${{ env.head_sha }}'
+            echo 'PR_HEADSHA=${head_sha}'
           } >> "${GITHUB_ENV}"
 
       # ==================== #
--- a/.github/workflows/docs-preview-deploy.yml
+++ b/.github/workflows/docs-preview-deploy.yml
@@ -17,7 +17,7 @@ jobs:
     name: 'Deploy Preview'
     runs-on: ubuntu-22.04
     # Requires a PR event triggered `docs-preview-prepare.yml` workflow run that was successful + ensure the head SHA belongs to an associated PR:
-    # Multi-line `if` GHA expression should avoid wrapping with `${{ }}`:
+    # NOTE: A multi-line `if` GHA expression must avoid wrapping with `${{ }}`, otherwise it is unintentionally parsed as a string:
     # https://github.com/nikitastupin/pwnhub/blob/main/writings/if-condition.md
     if: |
       github.event.workflow_run.conclusion == 'success'
@@ -42,6 +42,8 @@ jobs:
 
       # The `workflow_run` metadata contains an array of `pull_requests`, get the `workflow_run` equivalent of `github.event.pull_request.number`.
       # There should only be one PR item in the array, verify that it shares the same `head_sha` (latest commit of PR).
+      # NOTE: Careful when using GHA context expressions that may have untrusted input here. The expressions are evaluated before the script content itself is run:
+      # https://github.com/docker-mailserver/docker-mailserver/pull/4247#discussion_r1827067475
       - name: 'Get PR number'
         env:
           head_sha: ${{ github.event.workflow_run.head_sha }}
--- a/.github/workflows/docs-preview-deploy.yml
+++ b/.github/workflows/docs-preview-deploy.yml
@@ -17,9 +17,10 @@ jobs:
     name: 'Deploy Preview'
     runs-on: ubuntu-22.04
     # Requires a PR event triggered `docs-preview-prepare.yml` workflow run that was successful + ensure the head SHA belongs to an associated PR:
-    if: github.event.workflow_run.conclusion == 'success'
-        && github.event.workflow_run.event == 'pull_request'
-        && contains(github.event.workflow_run.pull_requests.*.head.sha, github.event.workflow_run.head_sha)
+    if: >-
+      github.event.workflow_run.conclusion == 'success'
+      && github.event.workflow_run.event == 'pull_request'
+      && contains(github.event.workflow_run.pull_requests.*.head.sha, github.event.workflow_run.head_sha)
     steps:
 
       # ======================== #
--- a/.github/workflows/docs-preview-deploy.yml
+++ b/.github/workflows/docs-preview-deploy.yml
@@ -17,12 +17,9 @@ jobs:
     name: 'Deploy Preview'
     runs-on: ubuntu-22.04
     # Requires a PR event triggered `docs-preview-prepare.yml` workflow run that was successful + ensure the head SHA belongs to an associated PR:
-    if: >
-      ${{
-        github.event.workflow_run.conclusion == 'success'
+    if: github.event.workflow_run.conclusion == 'success'
         && github.event.workflow_run.event == 'pull_request'
         && contains(github.event.workflow_run.pull_requests.*.head.sha, github.event.workflow_run.head_sha)
-      }}
     steps:
 
       # ======================== #
--- a/.github/workflows/docs-preview-deploy.yml
+++ b/.github/workflows/docs-preview-deploy.yml
@@ -17,7 +17,9 @@ jobs:
     name: 'Deploy Preview'
     runs-on: ubuntu-22.04
     # Requires a PR event triggered `docs-preview-prepare.yml` workflow run that was successful + ensure the head SHA belongs to an associated PR:
-    if: >-
+    # Multi-line `if` GHA expression should avoid wrapping with `${{ }}`:
+    # https://github.com/nikitastupin/pwnhub/blob/main/writings/if-condition.md
+    if: |
       github.event.workflow_run.conclusion == 'success'
       && github.event.workflow_run.event == 'pull_request'
       && contains(github.event.workflow_run.pull_requests.*.head.sha, github.event.workflow_run.head_sha)
@@ -45,7 +47,7 @@ jobs:
           head_sha: ${{ github.event.workflow_run.head_sha }}
           pull_requests: ${{ tojson(github.event.workflow_run.pull_requests) }}
         run: |
-          PR_NUMBER=$(jq -r '[.[] | select(.head.sha == "${head_sha}")][0].number' <<< "${pull_requests}")
+          PR_NUMBER=$(jq -r '[.[] | select(.head.sha == "${{ env.head_sha }}")][0].number' <<< "${pull_requests}")
           {
             echo "PR_NUMBER=${PR_NUMBER}"
             echo "PR_HEADSHA=${head_sha}"


.github/workflows/docs-preview-deploy.yml
```diff
@@ -47,10 +45,10 @@ jobs:
           head_sha: ${{ github.event.workflow_run.head_sha }}
           pull_requests: ${{ tojson(github.event.workflow_run.pull_requests) }}
         run: |
-          PR_NUMBER=$(jq -r '[.[] | select(.head.sha == "${{ env.head_sha }}")][0].number' <<< '${{ env.pull_requests }}')
+          PR_NUMBER=$(jq -r '[.[] | select(.head.sha == "${head_sha}")][0].number' <<< '${pull_requests}')
```
polarathene, line 48: I'm not sure what the value of this would be, and I don't think your suggestion here would work as-is due to the use of `'` single-quotes wrapping this expression it would not interpolate the variables. Thus your suggestion would break the intended functionality.

The expression syntax to use [`env` context](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#example-usage-of-the-env-context) works as that is pre-processed before the `run` executes.

---

Both of these variables are referencing the steps own `env` vars set directly above, which use context that shouldn't be possible to tamper with? Could you please explain the value of this suggestion?

I think there is less risk for error for maintainers with the `env` context expressions used, since it is distinctively clear for us to grok vs the subtle error with quotes in shell scripts affecting interpolation which can introduce bugs as this PR has done unintentionally?

.github/workflows/docs-preview-deploy.yml
```diff
@@ -47,10 +45,10 @@ jobs:
...
           {
             echo "PR_NUMBER=${PR_NUMBER}"
-            echo 'PR_HEADSHA=${{ env.head_sha }}'
+            echo 'PR_HEADSHA=${head_sha}'
```
polarathene, line 51: Likewise, the content is single quote wrapped, this would not interpolate and instead the value of `PR_HEADSHA` as an ENV will now be the string `${head_sha}` instead of the actual fixed string of a the head SHA.

polarathene (changes requested):
Thank you for the contribution, but I am not sure if it make sense to resolve these unless there is a clear benefit over the current implementation?
- `if` can exclude `${{ }}` yes, but for maintainers I rather they don't have to think about that and would prefer to keep it simple for them to recognize and not question when it's valid to omit the expression syntax. These workflows are rarely modified beyond dependabot.
- Replacing the `env` context usage with explicit shell variables could be done, but unless there is a valid improvement, I would again prefer to keep it as-is since it'll be easier to grok and less error prone. It can be quite easy for these little mistakes to slip through.

.github/workflows/docs-preview-deploy.yml
```diff
@@ -17,12 +17,10 @@ jobs:
...
-    if: |
-      ${{
-        github.event.workflow_run.conclusion == 'success'
-        && github.event.workflow_run.event == 'pull_request'
-        && contains(github.event.workflow_run.pull_requests.*.head.sha, github.event.workflow_run.head_sha)
-      }}
+    if: >-
+      github.event.workflow_run.conclusion == 'success'
+      && github.event.workflow_run.event == 'pull_request'
+      && contains(github.event.workflow_run.pull_requests.*.head.sha, github.event.workflow_run.head_sha)
```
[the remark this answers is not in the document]
pwntester: The way it was defined before it was always evaluated to true since it was a string rather than a boolean (see https://github.com/nikitastupin/pwnhub/blob/main/writings/if-condition.md)

.github/workflows/docs-preview-deploy.yml
pwntester: The sha is totally safe since its only alphanumeric value. The `github.event.workflow_run.pull_requests` contains untrusted data from the triggering Pull Request though (eg: title and body). An attacker can add a body like ``foo'`id`'bar`` which will pollute the value of the `pull_requests` env var. If you use the `${{}}` expression interpolation, the workflow expressions will get interpolated into the final bash script that will be then passed to bash for exection. So if you use `${{ env.pull_requests }}` the attacker will be able to use the PR body to close the bash single quote and add new commands (eg: ``foo';`id`;bar``).

You are right that `'${pull_requests}'` needs to be replaced with `"${pull_requests}"` since otherwise it wont get expanded. Will fix that in the PR. More info [here](https://securitylab.github.com/resources/github-actions-untrusted-input/)

.github/workflows/docs-preview-deploy.yml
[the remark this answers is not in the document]
polarathene: Ok thank you :)


```suggestion
    # Multi-line `if` GHA expression should avoid wrapping with `${{ }}`:
    # https://github.com/nikitastupin/pwnhub/blob/main/writings/if-condition.md
    if: |
      github.event.workflow_run.conclusion == 'success'
      && github.event.workflow_run.event == 'pull_request'
      && contains(github.event.workflow_run.pull_requests.*.head.sha, github.event.workflow_run.head_sha)
```

Seems like official docs don't touch on this caveat?:

- https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions
- https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/using-conditions-to-control-job-execution

.github/workflows/docs-preview-deploy.yml
polarathene: > So if you use `${{ env.pull_requests }}` the attacker will be able to use the PR body to close the bash single quote and add new commands (eg: `` foo';`id`;bar ``).

~~While I get what you're saying here, how is it different from `${pull_requests}`? Both would be a string of JSON key/value pairs?~~

Ah ok, templating to generate script with content vs runtime variable.

> The `github.event.workflow_run.pull_requests` contains untrusted data from the triggering Pull Request though (eg: title and body).

I don't think it does, `pull_request` context has specific metadata like that but not for this `workflow_run` event:

```bash
echo "PRs: ${{ tojson(github.event.workflow_run.pull_requests) }}"
```

```
PRs: [
  {
    base: {
      ref: main,
      repo: {
        id: 506839796,
        name: actions-example,
        url: https://api.github.com/repos/polarathene/actions-example
      },
      sha: e2e0c6555bec8ca661b59057443cfa5a54b8da75
    },
    head: {
      ref: test-branch,
      repo: {
        id: 506839796,
        name: actions-example,
        url: https://api.github.com/repos/polarathene/actions-example
      },
      sha: 113f63be4028584a1528e63ff33f821d990c28d2
    },
    id: 2157169857,
    number: 7,
    url: https://api.github.com/repos/polarathene/actions-example/pulls/7
  }
]
```

No PR title in this data?

.github/workflows/docs-preview-deploy.yml
```diff
@@ -47,10 +45,10 @@ jobs:
           head_sha: ${{ github.event.workflow_run.head_sha }}
           pull_requests: ${{ tojson(github.event.workflow_run.pull_requests) }}
         run: |
-          PR_NUMBER=$(jq -r '[.[] | select(.head.sha == "${{ env.head_sha }}")][0].number' <<< '${{ env.pull_requests }}')
+          PR_NUMBER=$(jq -r '[.[] | select(.head.sha == "${head_sha}")][0].number' <<< "${pull_requests}")
```
polarathene, line 48: This will still be invalid:

```bash
jq -r '[.[] | select(.head.sha == "${head_sha}")][0].number'
```

`jq` is splatting (`.[]`) all array items from `pull_requests` variable, and selecting only the one that has `.head.sha` value that matches the expected checksum, but due to single quote wrapping this will not work. Flipping the quotes alone will not work either, you'd need to escape the inner `"` AFAIK.

The `env` input appears fairly trustworthy though? I don't think the branch name can be used as an attack vector with quotes?

```suggestion
          PR_NUMBER=$(jq -r "[.[] | select(.head.sha == \"${head_sha}\")][0].number" <<< "${pull_requests}")
```

vs

```suggestion
          PR_NUMBER=$(jq -r '[.[] | select(.head.sha == "${{ env.head_sha }}")][0].number' <<< '${{ env.pull_requests }}')
```

The latter seems simpler to grok, so long as the input is not exploitable that seems less prone to human error with maintenance?

.github/workflows/docs-preview-deploy.yml
pwntester: Good to know, the [docs](https://docs.github.com/en/webhooks/webhook-events-and-payloads#workflow_run) just mention that its an array of pull request objects so I assumed they contained the title and body. I would be in the safe side though, perhaps untrusted data is added to these objects in the future

.github/workflows/docs-preview-deploy.yml
pwntester: > I don't think the branch name can be used as an attack vector with quotes?

The branch name can, but the SHA is safe to use, so in this case its ok to use the env workflow expression

.github/workflows/docs-preview-deploy.yml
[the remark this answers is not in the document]
pwntester: Unfortunately not

.github/workflows/docs-preview-deploy.yml
pwntester, line 48:
```suggestion
          PR_NUMBER=$(jq -r '[.[] | select(.head.sha == "${{ env.head_sha }}")][0].number' <<< "${pull_requests}")
```

.github/workflows/docs-preview-deploy.yml
```diff
@@ -42,15 +42,17 @@ jobs:
...
         env:
           head_sha: ${{ github.event.workflow_run.head_sha }}
           pull_requests: ${{ tojson(github.event.workflow_run.pull_requests) }}
         run: |
-          PR_NUMBER=$(jq -r '[.[] | select(.head.sha == "${{ env.head_sha }}")][0].number' <<< '${{ env.pull_requests }}')
+          PR_NUMBER=$(jq -r '[.[] | select(.head.sha == "${{ env.head_sha }}")][0].number' <<< "${pull_requests}")
           {
             echo "PR_NUMBER=${PR_NUMBER}"
-            echo 'PR_HEADSHA=${{ env.head_sha }}'
+            echo "PR_HEADSHA=${head_sha}"
```
polarathene, line 55: Consistency with revert back to `env.head_sha`:

```suggestion
            echo 'PR_HEADSHA=${{ env.head_sha }}'
```

polarathene (changes requested):
I'll apply this. LGTM, thank you for contributing the fix and improvements! ❤️

.github/workflows/docs-preview-deploy.yml
```diff
@@ -17,12 +17,12 @@ jobs:
...
+    # https://github.com/nikitastupin/pwnhub/blob/main/writings/if-condition.md
     if: |
-      ${{
-        github.event.workflow_run.conclusion == 'success'
-        && github.event.workflow_run.event == 'pull_request'
-        && contains(github.event.workflow_run.pull_requests.*.head.sha, github.event.workflow_run.head_sha)
-      }}
+      github.event.workflow_run.conclusion == 'success'
+      && github.event.workflow_run.event == 'pull_request'
+      && contains(github.event.workflow_run.pull_requests.*.head.sha, github.event.workflow_run.head_sha)
```
polarathene, line 25: This condition appears to be [triggering a skip](https://github.com/docker-mailserver/docker-mailserver/actions/runs/11760837244/job/32761886447) for some reason.

---

`docs-preview-prepare` ran successfully on the [PR pre-merge commit](https://github.com/docker-mailserver/docker-mailserver/actions/runs/11760832296/job/32761878540#step:2:59):

```console
$ git -c protocol.version=2 fetch --no-tags --prune --no-recurse-submodules --depth=1 origin +a721146229a6adfb6f18cd28f14b8396c54ec6f7:refs/remotes/pull/4183/merge

a721146229a6adfb6f18cd28f14b8396c54ec6f7 -> pull/4183/merge
```

When cloning this repo and checkout that PR pre-merge commit, we can run `git log`:

```console
$ git log

commit a721146229a6adfb6f18cd28f14b8396c54ec6f7 (grafted, HEAD, pull/4183/merge)
Author: RoelSG <email-address-here>
Date:   Sun Nov 10 00:56:27 2024 +0000

    Merge a5682e7c805397493aec6e5f333b236fb13c7cb3 into 0ff9c0132a8914d6756739a7a3b085e47870b93d
```

`docs-preview-deploy` then [triggers for commit `0ff9c01`](https://github.com/docker-mailserver/docker-mailserver/actions/runs/11760837244), and the `if` condition above evaluates to false, skipping the workflow 🤔 

---

The `doc-preview-*` workflows were triggered by a [merge commit from `master` into the PR](https://github.com/docker-mailserver/docker-mailserver/pull/4183#issuecomment-2466529988) (_that I did via Github's Web UI button to update the branch_).

[Commit `a5682e7`](https://github.com/docker-mailserver/docker-mailserver/pull/4183/commits/a5682e7c805397493aec6e5f333b236fb13c7cb3) reflects that, while `a721146` would be the pre-merge commit Github created for the PR sync event, and [commit `0ff9c01`](https://github.com/docker-mailserver/docker-mailserver/commit/0ff9c0132a8914d6756739a7a3b085e47870b93d) is the current latest commit on `master` branch (_the squash merge commit from this very merged PR updating the workflow_).

Looking at an [older workflow run](https://github.com/docker-mailserver/docker-mailserver/actions/runs/11091863613) that did pass, [this commit](https://github.com/docker-mailserver/docker-mailserver/commit/3937e1e719f4058e0a4a846bff39383071f8391d) was the trigger, even though it shouldn't have met the [`docs-preview-prepare` path requirement to trigger](https://github.com/docker-mailserver/docker-mailserver/blob/3937e1e719f4058e0a4a846bff39383071f8391d/.github/workflows/docs-preview-prepare.yml#L3-L8), it was the [PR merge to master branch commit](https://github.com/docker-mailserver/docker-mailserver/pull/4201#event-14444896433), rather than a pre-merge commit, or as mentioned in this case a commit merging `master` into the PR branch (_which has been a [valid trigger in the past too](https://github.com/docker-mailserver/docker-mailserver/pull/4183#issuecomment-2361414208), unrelated to the commit content_).

---

The previous working `if` condition was:

```yaml
if: ${{ github.event.workflow_run.event == 'pull_request' && github.event.workflow_run.conclusion == 'success' }}
```

Presumably the `contains` addition is not compatible here. If that's the case then the context may not be valid below either 🤔 

I suppose it depends on if `a721146` (github generated pre-merge commit) or the actual latest commit on the PR branch (`a5682e7`?) is treated as the `head_sha`. Perhaps in this case it's a compatibility issue with the `master` to PR merge commit only, I'll verify that shortly 😕

polarathene: **UPDATE:**
- The condition did not skip on my own test repo when I created a PR and sync'd to changes of the primary branch `main`.
- [Opened a PR](https://github.com/docker-mailserver/docker-mailserver/pull/4258) with a small change to trigger the workflow, that [worked fine](https://github.com/docker-mailserver/docker-mailserver/actions/runs/11761492789).

Perhaps it has something to do with the the failing PR having been opened prior to this CI workflow update, or it's something related to a third-party contributor/branch 🤷‍♂️ 

---

**UPDATE 2:**

[Opened another PR](https://github.com/docker-mailserver/docker-mailserver/pull/4259), this time from my own fork of DMS. This triggered the same problem with [the `if` condition skipping the workflow](https://github.com/docker-mailserver/docker-mailserver/actions/runs/11761829712), and that was just the PR commits itself, no new pushes/updates.

polarathene: I removed the third condition:

```yaml
&& contains(github.event.workflow_run.pull_requests.*.head.sha, github.event.workflow_run.head_sha)
```

The workflow runs but [shows that the context we want is missing completely](https://github.com/docker-mailserver/docker-mailserver/actions/runs/11762692806/job/32765833419#step:4:1):

```
  PR_NUMBER=$(jq -r '[.[] | select(.head.sha == "45f63b686fb61d85fec2ddab7ac4504f8b191555")][0].number' <<< "${pull_requests}")
  {
    echo "PR_NUMBER=${PR_NUMBER}"
    echo 'PR_HEADSHA=45f63b686fb61d85fec2ddab7ac4504f8b191555'
  } >> "${GITHUB_ENV}"
  shell: /usr/bin/bash -e {0}
  env:
    head_sha: 45f63b686fb61d85fec2ddab7ac4504f8b191555
    pull_requests: []
```

Whereas for the working PR (branch on the same repo) [we have that extra context](https://github.com/docker-mailserver/docker-mailserver/actions/runs/11761492789/job/32763286908#step:4:2) available:

```
  PR_NUMBER=$(jq -r '[.[] | select(.head.sha == "8bef84e53657709eda5d35729399817035362efd")][0].number' <<< "${pull_requests}")
  {
    echo "PR_NUMBER=${PR_NUMBER}"
    echo 'PR_HEADSHA=8bef84e53657709eda5d35729399817035362efd'
  } >> "${GITHUB_ENV}"
  shell: /usr/bin/bash -e {0}
  env:
    head_sha: 8bef84e53657709eda5d35729399817035362efd
    pull_requests: [
    {
      "base": {
        "ref": "master",
        "repo": {
          "id": 33037215,
          "name": "docker-mailserver",
          "url": "https://api.github.com/repos/docker-mailserver/docker-mailserver"
        },
        "sha": "0ff9c0132a8914d6756739a7a3b085e47870b93d"
      },
      "head": {
        "ref": "docs/smtp-bind-fix-snippet-titles",
        "repo": {
          "id": 33037215,
          "name": "docker-mailserver",
          "url": "https://api.github.com/repos/docker-mailserver/docker-mailserver"
        },
        "sha": "8bef84e53657709eda5d35729399817035362efd"
      },
      "id": 2171261491,
      "number": 4258,
      "url": "https://api.github.com/repos/docker-mailserver/docker-mailserver/pulls/4258"
    }
  ]
```

So perhaps some metadata does need to be passed through, but then anyone could adjust the PR/issue provided as input 🤷‍♂️

pwntester: Sorry for the late response, I totally missed the notification.
The third condition can be safely removed.
I dont know why `github.event.workflow_run.pull_requests` is empty for that case, but I dont see how passing it via an env var rather than interpolating it directly could change the value it receives from the runner.
I found this answer that says that `pull_requests` is not filled for PRs coming from a fork and they had to iterate through all open PRs to find the right one: https://stackoverflow.com/a/79017997

polarathene: Here we go, from the [GHA workflow event trigger docs](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request)

> _The `pull_request` webhook event payload is empty for merged pull requests and pull requests that come from forked repositories._

---

There has also been an ongoing [discussion about this particular issue for years](https://github.com/orgs/community/discussions/25220#discussioncomment-3246921):
1. [GH blog advises with passing PR number via artifact as ENV from untrusted to trusted workflows](https://github.com/orgs/community/discussions/25220#discussioncomment-7856118) (_which led to the security concern in the first place_ 😅 )
2. [GH CLI or REST API via JS script](https://github.com/orgs/community/discussions/25220#discussioncomment-8697399) (rate limit risk)
3. [`pull_request_target` + re-usable workflows with restricted permissions for untrusted builds](https://github.com/orgs/community/discussions/25220#discussioncomment-9164547)
4. [`gh run view`](https://github.com/orgs/community/discussions/25220#discussioncomment-7532132) provides the PR number in an awkward manner, but is only compatible with forks apparently 🤷‍♂️ 

Perhaps this is something GH can properly address, as community discussions like that with the confusion and various solutions probably isn't benefiting anyone 😅 

---

The [blog post referenced](https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/) by solution 1 above, does touch on `pull_request_target` caveats:

> You may ask yourself: if the `pull_request_target` workflow only checks out and builds the PR, i.e. runs untrusted code but doesn’t reference any secrets, is it still vulnerable?
> 
> Yes it is, because a workflow triggered on `pull_request_target` still has the read/write repository token in memory that is potentially available to any running program.
> If the workflow uses `actions/checkout` and does not pass the optional parameter `persist-credentials` as false, it makes it even worse. The default for the parameter is `true`.
> 
>  It means that in any subsequent steps any running code can simply read the stored repository token from the disk.
>  If you don’t need a repository write access or secrets, just stick to the `pull_request` trigger.

The `persist-credentials` setting appears to have that related logic [here](https://github.com/actions/checkout/blob/3b9b8c884f6b4bb4d5be2779c26374abadae0871/src/git-auth-helper.ts#L151-L189).

The blog post only refers to `steps` for a job, while [solution 3](https://github.com/orgs/community/discussions/25220#discussioncomment-9170862) from above shows a workflow with two jobs, the `build` job of the 1st workflow provides inputs into the 2nd via workflow call trigger to perform a git checkout of the PR without secrets or permissions, which should avoid the `persist-credentials` concern for their `preview` job that runs next?

---

**EDIT:** I've been looking into solution 3 (`pull_request_target`) and it seems like it might be the right way to go, I'll get a PR up for this and ping you for a glance over? :)

polarathene: > The third condition can be safely removed.

Yeah, it was more of a safe-guard and in this case it did correctly skip the workflow because the expected metadata was missing. I dropped it to verify that it was in fact lacking the metadata needed, so a different solution is required.

> I dont know why `github.event.workflow_run.pull_requests` is empty for that case, but I dont see how passing it via an env var rather than interpolating it directly could change the value it receives from the runner.

The previous solution that did work was using the untrusted `pull_request` workflow (prepare) to get the pull request number and related info, then send that over to the trusted `workflow_run` workflow (deploy) which appended the lines to the filepath of `$GITHUB_ENV`.

While that did work, there was a security concern with `LD_PRELOAD` you mentioned and there was various concerns I had with how to approach that properly that I tried the alternative approach that we have in place currently.

I'd rather not go back to the previous method since the PR contributor can manipulate what those values would be.

> I found this answer that says that `pull_requests` is not filled for PRs coming from a fork and they had to iterate through all open PRs to find the right one: https://stackoverflow.com/a/79017997

Yes, as my prior message posted at roughly the same time as your comment notes, that appears to be the case. I'll push ahead with the `pull_request_target` solution 👍

PR #4258, opened 2024-11-10
docs(fix): SMTP bind example show snippet titles

polarathene:
# Description

Fixes a syntax issue that broke rendering of these snippets.

### DIFF
--- a/docs/content/examples/use-cases/bind-smtp-network-interface.md
+++ b/docs/content/examples/use-cases/bind-smtp-network-interface.md
@@ -29,7 +29,7 @@ to the respective IP-address on the server you want to use.
 
     === "Contributed solution"
 
-        ```title="postfix-main.cf"
+        ``` title="postfix-main.cf"
         smtp_bind_address = 198.51.100.42
         smtp_bind_address6 = 2001:DB8::42
         ```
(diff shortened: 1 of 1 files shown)


casperklein (approved):

polarathene:
> > that broke rendering of these snippets.
> 
> What exactly is broken in the snippets? They are rendering fine for me.

I've only noticed it on `:edge` a month or so ago, yet the docs page and docker docs builder image haven't been modified since 10 months ago with v14. So unless it's a regression of the `9.5` point releases of mkdocs-material, perhaps it was introduced somewhere else 🤷‍♂️ 

> image

I could try ~~bumping the docs builder image~~ (_**EDIT:** Nope, no new minor, it's at `9.5.44` now, so probably a regression_).

---

> Not related, but I think that could be useful additions:

Yeah, we could add the copy button. Some time it's been annoying when I've seen it in docs due to the button overlay covering up text visibility when you just want to read the actual content (_that was a bit too long/wide_). Feel free to PR it :)

The line selection feature you linked we cannot use for now as it's insiders members only.

casperklein:
> I've only noticed it on :edge

Confirmed 👍

polarathene:
This won't be necessary, turns out we've not had a build since Sep 29 which was just bad timing with the regression which has since been fixed: https://github.com/squidfunk/mkdocs-material/discussions/7687#discussioncomment-11219842

polarathene:
Closing as verified this is not necessary.

PR #4259, opened 2024-11-10
fix: SASLAuth - Drop services for `mysql`, `shadow`, `pam` auth mechanisms

### LINKED ISSUE #4251: [TODO]: Remove `mysql` from valid `SASLAUTHD_MECHANISMS`
### Subject

I would like to contribute to the project

### Description

Currently, the bundled saslauthd does not support mysql authentication mechanism, but is included in [master/mailserver.env](https://github.com/docker-mailserver/docker-mailserver/blob/master/mailserver.env) which is very misleading.

```
usage: saslauthd [options]

option information:
  -a <authmech>  Selects the authentication mechanism to use.
  -c             Enable credential caching.
  -d             Debugging (don't detach from tty, implies -V)
  -r             Combine the realm with the login before passing to authentication mechanism
                 Ex. login: "foo" realm: "bar" will get passed as login: "foo@bar"
                 The realm name is passed untouched.
  -O <option>    Optional argument to pass to the authentication
                 mechanism.
  -l             Disable accept() locking. Increases performance, but
                 may not be compatible with some operating systems.
  -m <path>      Alternate path for the saslauthd working directory,
                 must be absolute.
  -n <procs>     Number of worker processes to create.
  -s <kilobytes> Size of the credential cache (in kilobytes)
  -t <seconds>   Timeout for items in the credential cache (in seconds)
  -v             Display version information and available mechs
  -V             Enable verbose logging
  -h             Display this message.

saslauthd 2.1.28
authentication mechanisms: sasldb getpwent kerberos5 pam rimap shadow ldap
```

polarathene:
# Description

`SASLAUTHD_MECHANISMS` only has test coverage for `ldap` and `rimap`, which were discussed during the original implementation of the feature.

The other mechanisms were added for the sake of it, but likely don't make much sense with DMS.
- Potentially `pam` would as it appears to be the way to leverage databases like `mysql` for authentication.
- `mysql` has not been a valid value for well over a decade. It's likely never worked in DMS.

Generally `SASLAUTHD` related support is not necessary for users. It was contributed as an alternative SASL auth method for Postfix to use when DMS has Dovecot disabled that would normally be delegated for SASL authentication.

Users interested in MySQL or similar DB access could configure for such directly with Postfix and Dovecot (or just Dovecot due to existing SASL auth config).

There may be potential breakage with `SMTP_ONLY=1` users. I've not considered this as our support/docs for such are lacking. As this is for a breaking release, it should be fine and we can easily bring these back should users raise issues about it.

Docs wise, I don't see how the note about multiple values for mechanisms being supported was valid. Perhaps it was for direct usage of the command but not with how we start the supervisord services.

Fixes #4251

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file. The format 
 
 ### Breaking
 
+- **saslauthd** mechanism support via ENV `SASLAUTHD_MECHANISMS` with `pam`, `shadow`, `mysql` values has been removed. Only `ldap` and `rimap` remain supported.
 - **getmail6** has been refactored: ([#4156](https://github.com/docker-mailserver/docker-mailserver/pull/4156))
   - The [DMS config volume](https://docker-mailserver.github.io/docker-mailserver/v15.0/config/advanced/optional-config/#volumes) now has support for `getmailrc_general.cf` for overriding [common default settings](https://docker-mailserver.github.io/docker-mailserver/v15.0/config/advanced/mail-getmail/#common-options). If you previously mounted this config file directly to `/etc/getmailrc_general` you should switch to our config volume support.
   - IMAP/POP3 example configs added to our [`config-examples`](https://github.com/docker-mailserver/docker-mailserver/tree/v15.0.0/config-examples/getmail).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ All notable changes to this project will be documented in this file. The format 
 
 ### Breaking
 
-- **saslauthd** mechanism support via ENV `SASLAUTHD_MECHANISMS` with `pam`, `shadow`, `mysql` values has been removed. Only `ldap` and `rimap` remain supported.
+- **saslauthd** mechanism support via ENV `SASLAUTHD_MECHANISMS` with `pam`, `shadow`, `mysql` values has been removed. Only `ldap` and `rimap` remain supported ([#4259](https://github.com/docker-mailserver/docker-mailserver/pull/4259))
 - **getmail6** has been refactored: ([#4156](https://github.com/docker-mailserver/docker-mailserver/pull/4156))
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file. The format
 
 ### Breaking
 
+- **saslauthd** mechanism support via ENV `SASLAUTHD_MECHANISMS` with `pam`, `shadow`, `mysql` values has been removed. Only `ldap` and `rimap` remain supported.
```
polarathene, line 11:
```suggestion
- **saslauthd** mechanism support via ENV `SASLAUTHD_MECHANISMS` with `pam`, `shadow`, `mysql` values has been removed. Only `ldap` and `rimap` remain supported ([#4259](https://github.com/docker-mailserver/docker-mailserver/pull/4259))
```

target/scripts/startup/variables-stack.sh
```diff
@@ -182,8 +182,9 @@ function _environment_variables_ldap() {
 function _environment_variables_saslauthd() {
   _log 'debug' 'Setting SASLAUTHD-related environment variables now'
 
-  # Only used by the supervisor service command (upstream default: `/etc/default/saslauthd`)
-  VARS[SASLAUTHD_MECHANISMS]="${SASLAUTHD_MECHANISMS:=pam}"
+  # This ENV is only used by the supervisor service command:
```
polarathene, line 185:
```suggestion
  # This ENV is only used by the supervisor service config `saslauth.conf`:
```

casperklein (approved):

polarathene:
Heads-up, **do not merge this PR.**

It's waiting on @pwntester for advice (if any), relating to our CI docs preview no longer working for PRs coming from forks: https://github.com/docker-mailserver/docker-mailserver/pull/4247#discussion_r1835596427

PR #4262, opened 2024-11-14
ci(docs-preview): Refactor workflows

polarathene:
# Description

Refactored and better documented our CI workflows for generating documentation previews.

This PR needs to be merged before a follow-up which proposes to move away from the current `pull_request` + `workflow_run` approach that presently is not compatible with PRs from forks.

- If you would like to block and provide a review but lack time to do so within the next 24 hours, let me know otherwise I'm not expecting much feedback here and will merge without approval.
- I have ensured this change works as intended by testing via a separate repo of my own.

---

**TL;DR:** Overall, it's mostly shuffling stuff around, added comments and minor improvements. I wanted to keep the noise of these changes separate from the follow-up PR which will get a proper review process.

**Overview of changes:**
- Runner bumped from Ubuntu 22.04 => 24.04
- Revised inline documentation for maintainers.
- `build-docs.sh` output is grouped in action step log + quiets info about pulling the image.
- Since no additional files are bundled into the archive, the separate `tar` steps with ZSTD aren't needed. The `upload` + `download` actions have since been improved over the past 3-4 years and are sufficient by themselves.
- Deploy workflow run extracted the PR context restore to a separate job.
- `actions-netlify` is effectively the same functionality. `github-token` is no longer configured as it doesn't appear needed with the functions disabled. Now opts-out of the GH deployments feature which we have no use for.

**NOTE:** The current `workflow_run` (`master`) is expected to fail due to the `pull_request` (prepare) workflow running from this PR instead of the `master` branch version of the workflow. Failure is due to change in artifact upload/download.

## Type of change

## Checklist

### DIFF
--- a/.github/workflows/docs-preview-deploy.yml
+++ b/.github/workflows/docs-preview-deploy.yml
@@ -1,133 +1,164 @@
 name: 'Documentation (run)'
 
 on:
+  # This workflow runs off the primary branch which provides access to the `secrets` context:
   workflow_run:
     workflows: ['Documentation (PR)']
     types:
       - completed
 
-# Note: If limiting concurrency is required for this workflow:
-# 1. Add an additional job prior to `preview` to get the PR number make it an output.
-# 2. Assign that new job as a `needs` dependency for the `preview` job.
-# It is still required for `preview` job to download the artifact so that it can access the preview build files.
+permissions:
+  # Required by `actions/download-artifact`:
+  actions: read
+  # Required by `marocchino/sticky-pull-request-comment`:
+  pull-requests: write
+  # Required by `myrotvorets/set-commit-status-action`:
+  statuses: write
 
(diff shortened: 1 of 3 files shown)


.github/workflows/docs-preview-deploy.yml
```diff
@@ -1,133 +1,164 @@
...
       # ==================== #
       # Deploy preview build #
       # ==================== #
 
-      # Manage workflow deployment status. `enable-commit-status` from `nwtgck/actions-netlify` would handle this,
-      # but presently does not work correctly via split workflow. It is useful in a split workflow as the 1st stage
-      # no longer indicates if the entire workflow/deployment was successful.
-      - name: 'Commit Status: Set Workflow Status as Pending'
+      # Manage workflow deployment status (Part 1/2):
+      # NOTE:
```
polarathene, line 85:
```suggestion
      # NOTE:
```

PR #4264, opened 2024-11-15
ci(docs-preview): Refactor to use `pull_request_target`

polarathene:
# Description

This should restore our docs preview workflow to be compatible with PRs from forks.

- Follow-up PR from https://github.com/docker-mailserver/docker-mailserver/pull/4262
- Proposes solution 3 (`pull_request_target`) discussed here: https://github.com/docker-mailserver/docker-mailserver/pull/4247#discussion_r1837228792

### Context

Originally we had compatibility with PRs from forks with the `pull_request` event workflow, but relied on passing the PR metadata via ENV in a file that the PR author could manipulate which was a security risk.
- That was [changed recently](https://github.com/docker-mailserver/docker-mailserver/pull/4244), but in doing so we found it [broke the needed context from `workflow_run` when PRs were from forks](https://github.com/docker-mailserver/docker-mailserver/pull/4247#discussion_r1835559791).
- We could revert back to that with some changes to prevent the security issue it presented, or consider other solutions like [this one](https://stackoverflow.com/questions/59077079/how-to-get-pull-request-number-within-github-actions-workflow/79017997#79017997) (but may risk failure due to rate limits).
- `pull_request_target` seems like the most appropriate choice, provided it's handled carefully by maintainers.

---

## Change overview

Changes from `pull_request` (untrusted) + `workflow_run` (trusted) workflows, to use a `pull_request_target` workflow with separate `prepare` (untrusted) + `deploy` (trusted) re-usable workflows managed. 

- `docs-preview.yml` is now the entrypoint (_via `pull_request_target` event trigger_).
- `prepare` + `deploy` receive any configurables and secrets via inputs (JSON) from `docs-preview.yml`.
- Care has been taken to ensure that `pull_request_target` can safely be used to build the PR in a restricted trust environment. The only secret available is `GITHUB_TOKEN` with minimal permissions, but this should not be accessible to the contributor as their PR cannot modify this workflow like they could with `pull_request` event based workflows, and the `actions/checkout` step uses `persist-credentials: false` to ensure it's accidentally stored on disk.
- With this approach we no longer have two separate workflow runs to provide the docs preview functionality. This simplifies the `actions/downloads` step and removes the need for commit status management to have a check suite status for the `workflow_run` workflow previously used.

`docs-preview-prepare.yml` + `docs-preview-deploy.yml` provide re-usable workflows (`workflow_call` trigger) instead of inline jobs in `docs-preview` to minimize `permissions` and `secrets` to each job accordingly. It also hopefully communicates the distinction of untrusted vs trusted contexts for maintainers.

## Type of change

## Checklist


polarathene:
Pinging @pwntester for review 🙏

.github/workflows/docs-preview-deploy.yml
```diff
@@ -1,100 +1,45 @@
...
+      NETLIFY_SITE_ID:
+        required: true
+
+env:
+  BUILD_DIR:  ${{ fromJSON( inputs.preview-context ).build_dir }}
+  # PR head SHA (latest commit):
+  PR_HEADSHA: ${{ fromJSON( inputs.preview-context ).pull_request.head_sha }}
+  PR_NUMBER:  ${{ fromJSON( inputs.preview-context ).pull_request.number }}
+  # Deploy URL preview prefix (the site name for the prefix is managed at Netlify):
+  PREVIEW_SITE_PREFIX: ${{ fromJSON( inputs.preview-context ).netlify.deploy_prefix }}
```
polarathene, line 22: Both workflows should be fairly easy to grok now with `inputs` / `env` hoisted up to the top of the workflow.

Let me know if you disagree with the JSON input approach. I could make each one a separate input instead, but that seemed redundant and would just add more noise?

.github/workflows/docs-preview.yml
```diff
@@ -0,0 +1,89 @@
...
+        "number":    "${{ github.event.pull_request.number }}"
+      }
+    }
+
+# Grant `secrets.GITHUB_TOKEN` only the minimum permissions needed by actions to function:
+# https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token
+# NOTE: See the associated `preview` and `deploy` workflows called for when these permissions are needed.
+permissions:
+  contents: read
+  pull-requests: write
```
polarathene, line 47: These are the only "dynamic" config for the workflow. Not sure if the `repo.full_name` context could be used here for manipulation?

`permissions` has been limited accordingly. The `prepare` workflow that later runs the PR branch to build the docs only has `contents: read` available, but AFAIK should not have access to `GITHUB_TOKEN` on the PR author's side of things and they no longer have influence over the workflow itself.

That only leaves `build-docs.sh` as a concern, but with the restricted permissions for `GITHUB_TOKEN` and preventing any secrets from being in scope via `workflow_call` trigger, I think it's very close to the restricted environment of `pull_request` event triggered workflows? (_provided any future additions don't introduce risks in `docs-preview-prepare.yml`_)

.github/workflows/docs-preview.yml
```diff
@@ -0,0 +1,89 @@
...
+  # The `deploy` job is for running the remaining steps in a trusted context after building the PR branch:
+  # CAUTION: Do not execute any content from untrusted sources (the PR branch or the retrieved artifact from the `prepare` job)
+  deploy:
+    needs: [create-context, prepare]
+    uses: docker-mailserver/docker-mailserver/.github/workflows/docs-preview-deploy.yml@main
+    secrets:
+      NETLIFY_AUTH_TOKEN: ${{ secrets.NETLIFY_AUTH_TOKEN }}
+      NETLIFY_SITE_ID:    ${{ secrets.NETLIFY_SITE_ID    }}
+    with:
+      preview-context: ${{ needs.create-context.outputs.preview-context }}
```
polarathene, line 89: Presently the `prepare` + `deploy` jobs here have `uses` with an explicit `<org>/<repo> ... @<branch>` value.

I could relax that to omit those. PRs for branches from forks would not run changes on those branches AFAIK so that'd be ok. Might be better for any future maintenance to run the PR change, but I'm not sure how that may affect PRs from forks, especially with `pull_request_target`, it'd be the workflow from the PR base branch (which most of the time would be `master`.

---

Which reminds me I forgot to change the branch here 😓 

```suggestion
  # The `prepare` job is for running steps in an untrusted context (necessary to build the docs):
  # CAUTION: This runs a build script which the PR could modify for malicious purposes.
  prepare:
    needs: [create-context]
    uses: docker-mailserver/docker-mailserver/.github/workflows/docs-preview-prepare.yml@master
    with:
      preview-context: ${{ needs.create-context.outputs.preview-context }}

  # The `deploy` job is for running the remaining steps in a trusted context after building the PR branch:
  # CAUTION: Do not execute any content from untrusted sources (the PR branch or the retrieved artifact from the `prepare` job)
  deploy:
    needs: [create-context, prepare]
    uses: docker-mailserver/docker-mailserver/.github/workflows/docs-preview-deploy.yml@master
    secrets:
      NETLIFY_AUTH_TOKEN: ${{ secrets.NETLIFY_AUTH_TOKEN }}
      NETLIFY_SITE_ID:    ${{ secrets.NETLIFY_SITE_ID    }}
    with:
      preview-context: ${{ needs.create-context.outputs.preview-context }}
```

.github/workflows/docs-preview-prepare.yml
```diff
@@ -33,11 +28,21 @@ jobs:
...
     steps:
       - uses: actions/checkout@v4
+        with:
+          # As `on: pull_request_target` runs this workflow from the default branch of our repo,
+          # Adjust to checkout the correct PR branch (including repo to support PRs from forks):
+          ref: ${{ env.PR_REF }}
+          repository: ${{ env.PR_REPO }}
+          # Prevent `secrets.GITHUB_TOKEN` from being stored in a `.git/config` file, which could otherwise
+          # be compromised when executing untrusted code from a PR (as is done in the build step below).
+          persist-credentials: false
```
polarathene, line 38: Without this, my understanding is the equivalent of `secrets.GITHUB_TOKEN` would be stored in `.git/config`, so whatever `permissions` grant, the PR author could leverage for malicious intent should they gain access to the token.

I'm not sure how many other actions may cache the token, or other secrets by default like `actions/checkout` does implicitly. I don't think we would have many scenarios for workflows requiring to execute untrusted third-party content + access to secrets... presumably this is mostly a concern for PR oriented workflows when `pull_request` isn't sufficient?

.github/workflows/docs-preview-prepare.yml
```diff
@@ -33,11 +28,21 @@ jobs:
...
 
       # ================== #
       # Build docs preview #
       # ================== #
 
+      # CAUTION: Security risk due to executing `build-docs.sh` (which a PR could modify to be malicious):
+      # Care must be taken to run this workflow as if it were in an untrusted context (like the `on: pull_request` event trigger would).
```
polarathene, line 45: The only reason we allow this is so that an update to the Docker image tag via PR would trigger the workflow to detect any breakage to docs generation.

I could make another `workflow_call` workflow file for this instead (_[`docs-production-deploy.yml` requires calling `build-docs.sh`](https://github.com/docker-mailserver/docker-mailserver/blob/6b4627ceab44c65d04ae25926aab05956ca2bafb/.github/workflows/docs-production-deploy.yml#L45) too_), and perhaps with `uses:` excluding the repo and branch pin, we'd still be able to verify that version bump PR works correctly 🤔

polarathene (commented):
Context if it helps work through the review faster :)

polarathene:
## Alternative

I recently came across how another project chose to resolve the PR metadata concern: https://github.com/woodruffw/zizmor/issues/156#issuecomment-2480316590

`$GITHUB_ENV` in the `workflow_run` could be changed to `$GITHuB_OUTPUT` and use outputs for the step instead of `env` context.
- As an additional precaution would be to store values per variable into individual files, and validate the values (_eg only digits for a PR number_).
- Minor issue that valid values could still be manipulated by the PR author, but I've been made aware that we can now [enforce requiring approval to run workflows for external contributors](https://github.com/woodruffw/zizmor/issues/156#issuecomment-2480349409)

It may be worth considering as an alternative to this PR.
- We only need the PR number and head SHA, and `BUILD_DIR` is static.
- The approval for PRs from forks is [only for the `pull_request` event](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/approving-workflow-runs-from-public-forks#about-workflow-runs-from-public-forks), while `pull_request_target` would allow the PR author to always run `build-docs.sh` to execute code without our approval, unless we [address that](https://github.com/docker-mailserver/docker-mailserver/pull/4264#discussion_r1843092320).

georglauterbach:
I had a somewhat superficial review because this is rather intricate, and I lack the time to perform an in-depth investigation. The changes look good to me, and when you resolve some of the discussions you started yourself (branch names, etc.), I think we can ship this.

I cannot give an opinion on https://github.com/docker-mailserver/docker-mailserver/pull/4264#issuecomment-2480376111. Unless you implement it because you know it's better, I'd go with the current set of changes.

polarathene:
> I cannot give an opinion on [#4264 (comment)](https://github.com/docker-mailserver/docker-mailserver/pull/4264#issuecomment-2480376111). Unless you implement it because you know it's better, I'd go with the current set of changes.

That alternative is to revert back to our original `pull_request` + `workflow_run` approach passing the equivalent `workflow_call.inputs` / JSON via artifact upload/download again.
- It'd just use `$GITHUB_OUTPUT` instead of `$GITHUB_ENV`, which avoids any concern for `LD_PRELOAD`. But we could also be a be a bit more specific and add validation.
- The PR author could still manipulate the PR number or SHA via modifying the `pull_request` workflow. It won't have any known vulnerability to exploit anymore, but we'd probably want to consider always requiring approval to run the workflow as a added precaution 🤔 

I'm personally not a fan of the old approach we had.
- I find the approach I've proposed here with `pull_request_target` + `workflow_call` is more intuitive to follow (_`pull_request_target` directly calls the `preview` + `deploy` workflows, rather than `workflow_run` implicitly triggered by `pull_request` workflow_).
- The new approach doesn't need to use additional actions to manage the check suite status to report success/failure of the deployment.
- The required inputs are clearer to the maintainers, with the `workflow_call` workflows (`prepare` + `deploy`) defining these right at the top, while the shared context for those inputs are managed at the parent workflow `pull_request_target` (`docs-preview.yml`).
- PR authors cannot manipulate these workflows.

However this is not an approach I'd contribute to other projects as there is some risk if maintainers make changes to the workflow without care. The `pull_request` + `workflow_run` approach can be a little inconvenient, but is technically safer.

To my knowledge `docs-preview-prepare.yml` should be roughly equivalent in safety to `pull_request`, provided `permissions` are locked down (_defaults AFAIK is `read` for everything, even for PRs from forks from non-collaborators_) and `secrets` remains at `none` (_unfortunately it's not possible to enforce that expectation_). So long as `docs-preview-prepare.yml` is treated as if it were the equivalent `pull_request` workflow, all is well 😅

.github/workflows/docs-preview.yml
```diff
@@ -0,0 +1,89 @@
...
+        # The output value must be a single-line; use `jq` to convert the JSON to minified:
+        # NOTE: YAML `>-` does not help fold this multi-line content due to use of indentation:
+        # https://github.com/orgs/community/discussions/26105#discussioncomment-3250413
+        run: echo "preview-context=$(jq --compact-output <<< "${PREVIEW_CONTEXT}")" >> "${GITHUB_OUTPUT}"
+
+  # The `prepare` job is for running steps in an untrusted context (necessary to build the docs):
+  # CAUTION: This runs a build script which the PR could modify for malicious purposes.
+  prepare:
+    needs: [create-context]
+    uses: docker-mailserver/docker-mailserver/.github/workflows/docs-preview-prepare.yml@master
```
pwntester, line 76: This job is running untrusted code in the context of pull_request_target. There are no secrets shared with the job and only Pull Request write permissions so the risk is low, however there is still some risk. Cant this be moved to a `pull_request` workflow?

.github/workflows/docs-preview-prepare.yml
pwntester: The token is available to an attacker by dumping the runner memory so you need to assume that the attacker will be able to get Pull Request write permissions 
https://karimrahal.com/2023/01/05/github-actions-leaking-secrets/

The safest approach is to run all untrusted code in a `pull_request` triggered workflow and pass any needed information to a `workflow_run` privileged workflow but verifying this information to make sure its the expected one.

polarathene (commented):
**TL:DR:**
- The restricted workflow (`docs-preview-prepare.yml`) only has `contents: read` granted. As there is no secrets, it should roughly mirror the `pull_request` workflow environment security wise?
- Confirmed memory dump for accessing the GH token, provided PR author can run shell commands on the host.

---

While I'm fine to take a different approach than this PR proposes, I would greatly appreciate confirmation on `pull_request_target` + `workflow_call` (_as has been implemented here_) as being secure, akin to a `pull_request` event triggered for PRs from forks.

AFAIK with the proposed approach, the PR author has even less flexibility for an attack than they would with `pull_request`.

polarathene:
As another alternative ([Solution 4](https://github.com/docker-mailserver/docker-mailserver/pull/4247#discussion_r1837228792)). I have [recently documented `gh pr view`](https://github.com/orgs/community/discussions/25220#discussioncomment-11285971) to acquire the PR number and head SHA, and verified that it should not pose any rate limit risk.

This would be for `pull_request` + `workflow_run`, with this step added to the `workflow_run`:

```yml
# Minimum read permissions needed for `gh pr view`:
permissions:
  pull-requests: read
  contents: read

jobs:
# ...
  steps:
      - name: 'Get PR context'
        id: pr-context
        env:
          # Token required for GH CLI:
          GH_TOKEN: ${{ github.token }}
          # Best practice for scripts is to reference via ENV at runtime. Avoid using the expression syntax in the script content directly:
          PR_TARGET_REPO: ${{ github.repository }}
          # If the PR is from a fork, prefix it with `<owner-login>:`, otherwise only the PR branch name is relevant:
          PR_BRANCH: |-
            ${{
              (github.event.workflow_run.head_repository.owner.login != github.event.workflow_run.repository.owner.login)
                && format('{0}:{1}', github.event.workflow_run.head_repository.owner.login, github.event.workflow_run.head_branch)
                || github.event.workflow_run.head_branch
            }}
        # Use GH CLI to query the PR branch for the PR number and assign it to the steps output:
        # (`--jq` formats JSON to `key=value` pairs and renames `headRefOid` to `head-sha`)
        run: |
          gh pr view --repo "${PR_TARGET_REPO}" "${PR_BRANCH}" \
            --json 'number,headRefOid' \
            --jq '"number=\(.number)\nhead-sha=\(.headRefOid)"' \
            >> $GITHUB_OUTPUT
```

This then provides the PR context via outputs `${{ steps.pr-context.outputs.number }}` and `${{ steps.pr-context.outputs.head-sha }}` for later steps/jobs to use.

polarathene:
> Unfortunately using `pull_request_target` is very dangerous and even if you set `permissions: {}`, executing untrusted code still opens the door to [Cache Poisoning attacks](https://adnanthekhan.com/2024/05/06/the-monsters-in-your-build-cache-github-actions-cache-poisoning/) since the code is executed in the context of the default branch and the Cache Token available has write permissions.

Thanks for pointing this out! I had read similar caution from the [GHA docs on `pull_request_target` event workflow trigger](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_target), but didn't realize that the cache could still be uploaded by an attacker. It makes sense though, given it's probably not unlike the situation with uploading artifacts 😅 

A little unfortunate perhaps that we cannot opt-out of these attack vectors when we do not need the workflow to support that functionality.

Technically if my concern with `build-docs.sh` was addressed to limit the untrusted code to the Docker container, I don't think the attacker would have an injection point to leverage?

---

> Using an artifact as you were doing at the beginning is also ok as long as the contents of the artifact are unzipped to `/tmp` and contents verified to be a number or sha.

Any specific reason for the `/tmp` extraction location?

Validating a PR number or SHA hash value are simple enough (_I [documented those here](https://github.com/docker-mailserver/docker-mailserver/pull/4267#issuecomment-2484565209) under "Solution A"_), but gets a bit more complicated when the values aren't as simple to validate, and that validation alone does not ensure they're the expected values 🤔 

Since these specific values can be reliably sourced via `gh pr view` without the PR author manipulating them I'll favor that.

---

I'd really like to express my gratitude for your feedback and insights shared here. The articles you've linked are fantastic at demonstrating the dangers of `pull_request_target` with untrusted code that are far less apparent to maintainers 😬 

While I do prefer this `pull_request_target` + `workflow_call` approach, and isolating untrusted code to only run within the Docker container should be reasonably safe... I am no longer comfortable with the risks from unknowns, especially when that could affect our users.

I will move forward with [Solution B (`gh pr view`)](https://github.com/docker-mailserver/docker-mailserver/pull/4267#issuecomment-2484565209). I've updated the linked reference to heavily discourage Solution C (`pull_request_target`), citing the resources you've provided.

polarathene:
Closing in favor of https://github.com/docker-mailserver/docker-mailserver/pull/4267

Thank you so much @pwntester for your feedback! ❤️ I almost made an even bigger mistake here 😬

PR #4330, opened 2025-01-27
docs: `bind-smtp-network-interface.md` - Add bridge network config advice

zveljkovic:
# Description

Added info how to setup docker to use the same interface for outgoing traffic. Problem was that I was using single VPS for sites and mailserver. I got the additional IP address but all outgoing traffic from mailserver was going through the original IP. This network config makes the container use specified IP for outgoing traffic too. 

Problem manifested as the mail-tester.com saw the original VPS IP address and failed on rDNS checks even when setup with suggestions from this part of page.

Fixes #

## Type of change

- [ ] Documentation update

## Checklist

- [] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [ ] If necessary, I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] **I have added information about changes made in this PR to `CHANGELOG.md`**

Issue #4206, opened 2024-09-30
question: How to ensure outbound SMTP traffic is sent through a specific host interface when using a bridge network?


MarekPikula:
### 📝 Preliminary Checks

- [X] I tried searching for an existing issue and followed the [debugging docs](https://docker-mailserver.github.io/docker-mailserver/latest/config/debugging/) advice, but still need assistance.

### 👀 What Happened?

I followed [the guide](https://docker-mailserver.github.io/docker-mailserver/edge/examples/use-cases/bind-smtp-network-interface/) to bind SMTP to a specific network address. When I try to send an e-mail, there is the following message in logs:

```
postfix/smtp[1264]: warning: smtp_connect_addr: bind aaa.aaa.aaa.aaa: Cannot assign requested address
```

What is the solution to this? Do I need to run DMS with host network, or is there any other way with bridge network?

### 🐋 DMS Version

v14.0.0

### 💻 Operating System and Architecture

Fedora 39

### ⚙️ Container configuration files

```yml
version: '2'

services:
  mail:
    image: mailserver/docker-mailserver:latest
    hostname: mail
    domainname: xxx
    container_name: xxx_mail
    ports:
      - "aaa.aaa.aaa.aaa:25:25"
      - "aaa.aaa.aaa.aaa:143:143"
      - "aaa.aaa.aaa.aaa:465:465"
      - "aaa.aaa.aaa.aaa:587:587"
      - "aaa.aaa.aaa.aaa:993:993"
      - "aaa.aaa.aaa.aaa:995:995"
    volumes:
      - xxx_maildata:/var/mail
      - xxx_mailstate:/var/mail-state
      - xxx_maillog:/var/log/mail
      - ./config/:/tmp/docker-mailserver/
      - /store/config/sys_certs/certs/xxx.crt:/tmp/cert.pem
      - /store/config/sys_certs/private/xxx.key:/tmp/key.pem
    environment:
      ENABLE_SPAMASSASSIN: 1
      ENABLE_CLAMAV: 1
      ENABLE_FAIL2BAN: 0
      ENABLE_POSTGREY: 0
      ONE_DIR: 1
      LOG_LEVEL: debug

      ACCOUNT_PROVISIONER: 'LDAP'
      LDAP_SERVER_HOST: 'ldap://ldap'
      LDAP_BIND_DN: 'cn=xxx,dc=xxx,dc=xxx'
      LDAP_BIND_PW: 'xxxxxx'
[...]

PR #4345, opened 2025-02-04
docs: Revise `bind-smtp-network-interface.md`

polarathene:
# Description

This is a follow-up for the [recent update to this page](https://github.com/docker-mailserver/docker-mailserver/pull/4330), to rewrite the page for clearer distinction for config based on network type for the container.

Related past discussions on the topic for reference:
- https://github.com/docker-mailserver/docker-mailserver/pull/3465
- https://github.com/docker-mailserver/docker-mailserver/issues/4206

---

Presently draft status until I verify the host/postfix config advice actually handles the outbound IP concern.


georglauterbach:
I am currently going over open PRs. I will be merging this now because it's already been open for too long and has become stale.

> Presently draft status until I verify the host/postfix config advice actually handles the outbound IP concern.

If this ever becomes a concern again, please simply provide a follow-up PR. I am fine with it as-is for now and shipping it is better that having nothing.

georglauterbach (approved):

Issue #3769, opened 2024-01-10
[TODO]: Avoid being inconsistent with `POSTMASTER_ADDRESS`


polarathene:
### Subject

Something else that requires developers attention

### Description

This defaults to the [assumption that a hostname `mail.example.com` will use mail accounts `@example.com`](https://github.com/docker-mailserver/docker-mailserver/issues/3602#issuecomment-1783911289) and thus creates an implicit postmaster address to use within our scripts for that mail domain.

We've had several users in the past report deployments where this was not the case, thus the postmaster address was invalid.

They also need to add postmaster address for other mail domains they manage (_some have used a regexp alias on the local-part to handle all `postmaster@` redirects to a preferred address, but this appears to also affect foreign / remote `postmaster@` recipient domains too which would not be desired_).

---

### Reference topic

I recently responded to a discussion on this topic: https://github.com/orgs/docker-mailserver/discussions/3758#discussioncomment-8075486

It's a bit messy / disorganized, but is a public source of insights / notes from me 😅 This issue is to document the concern as it's not just the `POSTMASTER_ADDRESS` ENV default to consider, and our docs perhaps could also better cover this too.

> `postfix "warning: required alias not found: postmaster"`

---

### `mydestination`

Partly related to `mydestination` investigation from 2023:

https://github.com/docker-mailserver/docker-mailserver/blob/aba218e6d7f30fbf4717178581c3e04537bea178/target/postfix/main.cf#L15-L18

Postfix does not require a `local` delivery agent configured. The `mail.example.com` hostname was intended to distinguish the MTA itself from the hosted mail domains DMS manages, along with system related mail. It still is technically reachable to send mail to, or for it to send mail externally from, and itself [would be expected to have a postmaster address](https://en.wikipedia.org/wiki/Postmaster_(computing)).

It is not a change that we've been too eager to make, as my prior input expressed 😓 

- https://github.com/docker-mailserver/docker-mailserver/pull/3264#pullrequestreview-1396816109
- https://github.com/docker-mailserver/docker-mailserver/pull/3264#issuecomment-1518993555

I don't have time to dig through all that again to gain the confidence in adjusting that part of config.

Related to such a change is [concerns with aliasing](https://github.com/orgs/docker-mailserver/discussions/3765#discussioncomment-8075768) (_catchall address and it's gotchas_), which can be partly related to postmaster advice. Like this regexp alias in our tests which would not be wise for production deployments AFAIK:

[...]

PR #4350, opened 2025-02-10
chore: Migrate dovecot config from Dockerfile

polarathene:
# Description

I noticed the `postmaster_address` line looked odd in the Dockerfile, and that we actually run it properly in our startup scripts (_it's been in scripts [for over 5 years](https://github.com/docker-mailserver/docker-mailserver/blame/189e5376cc91af4f8a6b1e0db7b8f20e53a932a1/target/scripts/start-mailserver.sh#L542)_). That motivated this PR.

The protocol line was likewise migrated to be grouped with the related script lines with some minor revision.

The other 3 lines from the Dockerfile were all LDA specific which we haven't used since prior to DMS v2 (2016?), actually no I think we switched from Courier to Dovecot for v2, and then there was a switch from LDA to LMTP when around when LDAP support was contributed (_2016H2 or 2017 IIRC, **EDIT:** [LDA to LMTP in Oct 2016](https://github.com/docker-mailserver/docker-mailserver/pull/360)_). Anyway LDA is still technically relevant for supporting Getmail at least I think, and anything else that might call `/usr/lib/dovecot/deliver` to give mail to Dovecot directly.

I've included some extensive commentary inline since these settings at least for `hostname` and `postmaster_address` don't seem to be specific to LDA, that's just where they were placed in the Dovecot example configs we've copied. Dovecot 2.4 has removed those, so this will at least keep awareness of the original source until a maintainer/contributor addresses the TODO and we've migrated to 2.4 in a future base image update.

No changelog entry as this did not seem like a change that is user facing, it's effectively just shifting some commands around. I have altered the sed expressions into what should be easier to grok, along with only doing a single write/update to the file. I've manually tested the sed command to ensure it still works as intended :+1:

### DIFF
--- a/target/scripts/startup/setup.d/dovecot.sh
+++ b/target/scripts/startup/setup.d/dovecot.sh
@@ -37,12 +37,19 @@ function _setup_dovecot() {
   # Initial config files committed to DMS in April 2016:
   # TODO: Consider housekeeping on config to only represent relevant changes/support by scripts
   # https://github.com/docker-mailserver/docker-mailserver/commit/ee0d0853dd672488238eecb0ec2d26719ff45d7d
+  #
+  # TODO: `mail_plugins` appending `sieve` should probably be done for both `15-lda.conf` and `20-lmtp.conf`
+  # Presently DMS replaces the `20-lmtp.conf` from `dovecot-lmtpd` package with our own modified copy from 2016.
+  # The DMS variant only makes this one change to that file, thus we could adjust it as we do below for `15-lda.conf`
+  # Reference: https://github.com/docker-mailserver/docker-mailserver/pull/4350#issuecomment-2646736328
+
+  # shellcheck disable=SC2016
   sedfile -i -r \
     -e 's|^(\s*)#?(mail_plugins =).*|\1\2 $mail_plugins sieve|' \
     -e 's|^#?(lda_mailbox_autocreate =).*|\1 yes|' \
     -e 's|^#?(lda_mailbox_autosubscribe =).*|\1 yes|' \
     -e "s|^#?(postmaster_address =).*|\1 ${POSTMASTER_ADDRESS}|" \
-    -e "s|^#?(hostname =).*|\1 '${HOSTNAME}'|"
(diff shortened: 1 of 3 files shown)


polarathene:
For reference, this is technically related to https://github.com/docker-mailserver/docker-mailserver/issues/3769

georglauterbach:
This was opened against the v15.0.0 milestone. Do you think you can cram this in? It is probably fine because we are waiting on the ClamAV CVE fix, right?

target/scripts/startup/setup.d/dovecot.sh
```diff
@@ -3,13 +3,54 @@
...
+  # The DMS variant only makes this one change to that file, thus we could adjust it as we do below for `15-lda.conf`
+  # Reference: https://github.com/docker-mailserver/docker-mailserver/pull/4350#issuecomment-2646736328
+
+  # shellcheck disable=SC2016
+  sedfile -i -r \
+    -e 's|^(\s*)#?(mail_plugins =).*|\1\2 $mail_plugins sieve|' \
+    -e 's|^#?(lda_mailbox_autocreate =).*|\1 yes|' \
+    -e 's|^#?(lda_mailbox_autosubscribe =).*|\1 yes|' \
+    -e "s|^#?(postmaster_address =).*|\1 ${POSTMASTER_ADDRESS}|" \
+    -e "s|^#?(hostname =).*|\1 ${HOSTNAME}|"
```
polarathene, line 52: Last commit was done purely via editor without testing, hence failures. Forgot a trailing `\`:

```suggestion
    -e "s|^#?(hostname =).*|\1 ${HOSTNAME}|" \
```

polarathene:
> Do you think you can cram this in? It is probably fine because we are waiting on the ClamAV CVE fix, right?

Yeah this change should be fairly minor. There's no rush to get this change in, just a drive-by improvement while I was looking into something else.

I can add a changelog entry if you like, but logically nothing should have really changed, just shifting around the same logic and improving the sed expression. Although `hostname` will no longer quote the value, but I don't think it should have been to begin with.

georglauterbach (approved):

PR #4359, opened 2025-02-13
feat: Enable reading env vars from files

### LINKED ISSUE #3457: feature request: Adding `LDAP_BIND_PW_FILE` to support docker secrets
### Context

I want to connect LDAP but want to use docker secrets.

### Description

As the title states I want to use *docker secrets* to inject passwords.
I have looked trough the supported environment variables but couldn't find anything.
Since my credentials are already stored inside a file for docker secrets I would like to have one source of truth, besides the other security benefits.

Possbile candidates would be:

- `LDAP_BIND_PW_FILE`
- `SASLAUTHD_LDAP_PASSWORD_FILE`

### Alternatives

I can write the passwords into a `.env` file :-1: 

### Applicable Users

For anyone using **openldap** and docker compose.

I would love to contribute and implement it my self, at least I can try.

aartoni:
# Description

Fixes #3457.

Allows setting any environment variable via the content of a file, this is especially useful to carry Docker secrets in. This fix takes inspiration from a few similar approaches, namely Bitnami container images (e.g., [OpenLDAP](https://github.com/bitnami/containers/blob/02bcf4b54c8af61c18d5cb557b21dba317e36b48/bitnami/openldap/2.6/debian-12/rootfs/opt/bitnami/scripts/libopenldap.sh#L99-L117)) and [Grafana](https://github.com/grafana/grafana/blob/be426858529d9742808cc8ad4ed226379fb770da/packaging/docker/run.sh#L49-L61).

I have marked this as a breaking change since it would change the behavior of setups providing `<VAR>__FILE` env vars where `<VAR>` is an existing configuration variable. However, you may consider that to be negligible.

I've noticed a few Fail2ban tests breaking, but I'm not sure if that's due to my local setup or what. I've ignored them as they seemed unrelated.

## Type of change

## Checklist

### DIFF
--- a/target/scripts/startup/variables-stack.sh
+++ b/target/scripts/startup/variables-stack.sh
@@ -261,7 +261,7 @@ function __environment_variables_from_files() {
 
     # Skip if the target ENV is already set:
     if [[ -v TARGET_ENV ]]; then
-      _log 'warn' "Ignoring '${TARGET_ENV_NAME}' since '${ENV_WITH_FILE_REF}' is also set"
+      _log 'warn' "ENV value will not be sourced from '${ENV_WITH_FILE_REF}' since '${TARGET_ENV_NAME}' is already set"
       continue
     fi
 
--- a/target/scripts/startup/variables-stack.sh
+++ b/target/scripts/startup/variables-stack.sh
@@ -249,24 +249,27 @@ function __environment_variables_export() {
 # This function sets any environment variable with a value from a referenced file
 # when an equivalent ENV with a `__FILE` suffix exists with a valid file path as the value.
 function __environment_variables_from_files() {
-  for ENV_WITH_FILE_REF in $(env | grep -Po '^.+?__FILE'); do
+  while read -r ENV_WITH_FILE_REF; do
     # Store the ENV name without the `__FILE` suffix:
-    local TARGET_ENV="${ENV_WITH_FILE_REF/__FILE/}"
+    local TARGET_ENV_NAME="${ENV_WITH_FILE_REF/__FILE/}"
+    local -n TARGET_ENV="${ENV_WITH_FILE_REF/__FILE/}"
     # Store the value of the `__FILE` ENV:
     local FILE_PATH="${!ENV_WITH_FILE_REF}"
 
-    # Skip sourcing form `__FILE` if ENV is already set:
-    if [[ -n "${!TARGET_ENV}" ]]; then
-      _log 'warn' "ENV value will not be sourced from '${ENV_WITH_FILE_REF}' since '${TARGET_ENV}' is already set"
+    # Skip sourcing from `__FILE` if ENV is already set:
+    if [[ -v TARGET_ENV ]]; then
+      _log 'warn' "Ignoring '${TARGET_ENV_NAME}' since '${ENV_WITH_FILE_REF}' is also set"
       continue
     fi
 
-    # Otherwise retrieve the value from file and set the ENV or fail if invalid reference:
-    if [[ -f "${FILE_PATH}" ]]; then
-      _log 'info' "Getting secret ${TARGET_ENV} from ${FILE_PATH}"
-      printf -v "${TARGET_ENV}" '%s' "$(< "${FILE_PATH}")"
-    else
-      _log 'error' "File ${FILE_PATH} does not exist, defined in ${ENV_WITH_FILE_REF}"
+    # Otherwise, retrieve the value from a file and set
+    # the ENV, or fail if the reference is invalid:
+    if [[ ! -f ${FILE_PATH} ]]; then
+      _log 'warn' "File defined for secret '${TARGET_ENV_NAME}' with path '${FILE_PATH}' does not exist"
+      continue
     fi
-  done
+
+    _log 'info' "Getting secret '${TARGET_ENV_NAME}' from '${FILE_PATH}'"
+    TARGET_ENV="$(< "${FILE_PATH}")"
+  done < <(env | grep -Po '^.+?__FILE')
 }
--- a/target/scripts/startup/variables-stack.sh
+++ b/target/scripts/startup/variables-stack.sh
@@ -260,7 +260,7 @@ function __environment_variables_from_files() {
 
     # Skip sourcing form `__FILE` if ENV is already set:
     if [[ -n "${!TARGET_ENV}" ]]; then
-      _log 'warn' "Ignoring ${TARGET_ENV} since ${ENV_WITH_FILE_REF} is also set"
+      _log 'warn' "ENV value will not be sourced from '${ENV_WITH_FILE_REF}' since '${TARGET_ENV}' is already set"
       continue
     fi
 
--- a/target/scripts/startup/variables-stack.sh
+++ b/target/scripts/startup/variables-stack.sh
@@ -258,7 +258,7 @@ function __environment_variables_from_files() {
 
     # Skip sourcing from `__FILE` if ENV is already set:
     if [[ -v TARGET_ENV ]]; then
-      _log 'warn' "Ignoring '${TARGET_ENV_NAME}' since '${ENV_WITH_FILE_REF}' is also set"
+      _log 'warn' "ENV value will not be sourced from '${ENV_WITH_FILE_REF}' since '${TARGET_ENV_NAME}' is already set"
       continue
     fi
 
--- a/target/scripts/startup/variables-stack.sh
+++ b/target/scripts/startup/variables-stack.sh
@@ -249,23 +249,27 @@ function __environment_variables_export() {
   sort -o /etc/dms-settings /etc/dms-settings
 }
 
-# This function reads any environment variable ending with `__FILE` from its
-# referenced file, then makes it available under the same name without `__FILE`.
+# This function sets any environment variable with a value from a referenced file
+# when an equivalent ENV with a `__FILE` suffix exists with a valid file path as the value.
 function __environment_variables_from_files() {
-  for file_env_var in $(env | grep -Po '^.+?__FILE'); do
-    local env_var="${file_env_var/__FILE/}"
-    local file_path="${!file_env_var}"
+  for ENV_WITH_FILE_REF in $(env | grep -Po '^.+?__FILE'); do
+    # Store the ENV name without the `__FILE` suffix:
+    local TARGET_ENV="${ENV_WITH_FILE_REF/__FILE/}"
+    # Store the value of the `__FILE` ENV:
+    local FILE_PATH="${!ENV_WITH_FILE_REF}"
 
-    if [[ -n "${!env_var}" ]]; then
-      _log 'warn' "Ignoring ${env_var} since ${file_env_var} is also set"
+    # Skip sourcing form `__FILE` if ENV is already set:
+    if [[ -n "${!TARGET_ENV}" ]]; then
+      _log 'warn' "Ignoring ${TARGET_ENV} since ${ENV_WITH_FILE_REF} is also set"
       continue
     fi
 
-    if [[ -f "${file_path}" ]]; then
-      _log 'info' "Getting secret ${env_var} from ${file_path}"
-      printf -v "${env_var}" '%s' "$(< "${file_path}")"
+    # Otherwise retrieve the value from file and set the ENV or fail if invalid reference:
+    if [[ -f "${FILE_PATH}" ]]; then
+      _log 'info' "Getting secret ${TARGET_ENV} from ${FILE_PATH}"
+      printf -v "${TARGET_ENV}" '%s' "$(< "${FILE_PATH}")"
     else
-      _log 'error' "File ${file_path} does not exist, defined in ${file_env_var}"
+      _log 'error' "File ${FILE_PATH} does not exist, defined in ${ENV_WITH_FILE_REF}"
     fi
   done
 }
--- a/target/scripts/startup/variables-stack.sh
+++ b/target/scripts/startup/variables-stack.sh
@@ -249,26 +249,29 @@ function __environment_variables_export() {
 # This function sets any environment variable with a value from a referenced file
 # when an equivalent ENV with a `__FILE` suffix exists with a valid file path as the value.
 function __environment_variables_from_files() {
+  # Iterate through all ENV found with a `__FILE` suffix:
   while read -r ENV_WITH_FILE_REF; do
-    # Store the ENV name without the `__FILE` suffix:
-    local TARGET_ENV_NAME="${ENV_WITH_FILE_REF/__FILE/}"
-    local -n TARGET_ENV="${ENV_WITH_FILE_REF/__FILE/}"
     # Store the value of the `__FILE` ENV:
     local FILE_PATH="${!ENV_WITH_FILE_REF}"
+    # Store the ENV name without the `__FILE` suffix:
+    local TARGET_ENV_NAME="${ENV_WITH_FILE_REF/__FILE/}"
+    # Assign a value representing a variable name,
+    # `-n` will alias `TARGET_ENV` so that it is treated as if it were the referenced variable:
+    local -n TARGET_ENV="${TARGET_ENV_NAME}"
 
-    # Skip sourcing from `__FILE` if ENV is already set:
+    # Skip if the target ENV is already set:
     if [[ -v TARGET_ENV ]]; then
-      _log 'warn' "ENV value will not be sourced from '${ENV_WITH_FILE_REF}' since '${TARGET_ENV_NAME}' is already set"
+      _log 'warn' "Ignoring '${TARGET_ENV_NAME}' since '${ENV_WITH_FILE_REF}' is also set"
       continue
     fi
 
-    # Otherwise, retrieve the value from a file and set
-    # the ENV, or fail if the reference is invalid:
+    # Skip if the file path provided is invalid:
     if [[ ! -f ${FILE_PATH} ]]; then
       _log 'warn' "File defined for secret '${TARGET_ENV_NAME}' with path '${FILE_PATH}' does not exist"
       continue
     fi
 
+    # Read the value from a file and assign it to the intended ENV:
     _log 'info' "Getting secret '${TARGET_ENV_NAME}' from '${FILE_PATH}'"
     TARGET_ENV="$(< "${FILE_PATH}")"
   done < <(env | grep -Po '^.+?__FILE')
--- a/target/scripts/startup/variables-stack.sh
+++ b/target/scripts/startup/variables-stack.sh
@@ -263,7 +263,7 @@ function __environment_variables_from_files() {
 
     if [[ -f "${file_path}" ]]; then
       _log 'info' "Getting secret ${env_var} from ${file_path}"
-      export "${env_var}"="$(< "${file_path}")"
+      printf -v "${env_var}" '%s' "$(< "${file_path}")"
     else
       _log 'error' "File ${file_path} does not exist, defined in ${file_env_var}"
     fi
--- a/target/scripts/startup/variables-stack.sh
+++ b/target/scripts/startup/variables-stack.sh
@@ -5,6 +5,7 @@ declare -A VARS
 
 function _early_variables_setup() {
   __environment_variables_log_level
+  __environment_variables_from_files
   _obtain_hostname_and_domainname
   __environment_variables_backwards_compatibility
   __environment_variables_general_setup
@@ -246,4 +247,25 @@ function __environment_variables_export() {
 
   sort -o /root/.bashrc     /root/.bashrc
   sort -o /etc/dms-settings /etc/dms-settings
+}
+
+# This function reads any environment variable ending with `__FILE` from its
+# referenced file, then makes it available under the same name without `__FILE`.
+function __environment_variables_from_files() {
+  for file_env_var in $(env | grep -Po '^.+?__FILE'); do
+    local env_var="${file_env_var/__FILE/}"
+    local file_path="${!file_env_var}"
+
+    if [[ -n "${!env_var}" ]]; then
+      _log 'warn' "Ignoring ${env_var} since ${file_env_var} is also set"
+      continue
+    fi
+
+    if [[ -f "${file_path}" ]]; then
+      _log 'info' "Getting secret ${env_var} from ${file_path}"
+      export "${env_var}"="$(< "${file_path}")"
+    else
+      _log 'error' "File ${file_path} does not exist, defined in ${file_env_var}"
+    fi
+  done
 }
--- a/test/tests/parallel/set3/container_configuration/env_vars_from_files.bats
+++ b/test/tests/parallel/set3/container_configuration/env_vars_from_files.bats
@@ -1,66 +1,79 @@
 load "${REPOSITORY_ROOT}/test/helper/setup"
 load "${REPOSITORY_ROOT}/test/helper/common"
 
-export CONTAINER1_NAME='dms-test_env-files_success'
-export CONTAINER2_NAME='dms-test_env-files_warning'
-export CONTAINER3_NAME='dms-test_env-files_error'
+# Feature (ENV value sourced from file):
+# - An ENV with a `__FILE` suffix will read a value from a referenced file path to set the actual ENV (assuming it is empty)
+# - Feature implemented at: `variables-stack.sh:__environment_variables_from_files()`
+# - Feature PR: https://github.com/docker-mailserver/docker-mailserver/pull/4359
 
-setup_file() {
+BATS_TEST_NAME_PREFIX='[Configuration] (ENV __FILE support) '
+CONTAINER1_NAME='dms-test_env-files_success'
+CONTAINER2_NAME='dms-test_env-files_warning'
+CONTAINER3_NAME='dms-test_env-files_error'
+
+function setup_file() {
   export CONTAINER_NAME
-  export TEST__FILE
-  export TEST__FILE_SUCCESS
-  export NON_EXISTENT__FILE
+  export FILEPATH_VALID='/tmp/file-with-value'
+  export FILEPATH_INVALID='/path/to/non-existent-file'
 
+  export FILE_WITH_VALUE="${TEST_TMP_CONFIG}/test_secret"
+  echo 1 > "${FILE_WITH_VALUE}"
+
+  # ENV is set via file content (valid file path):
   CONTAINER_NAME=${CONTAINER1_NAME}
   _init_with_defaults
-  TEST__FILE=${TEST_TMP_CONFIG}/test_secret
-  echo 1 > "${TEST__FILE}"
-  TEST__FILE_SUCCESS=${TEST__FILE}
   local CUSTOM_SETUP_ARGUMENTS=(
-    --env ENABLE_POP3__FILE="${TEST__FILE}"
-    -v "${TEST__FILE}:${TEST__FILE}"
+    --env ENABLE_POP3__FILE="${FILEPATH_VALID}"
+    -v "${FILE_WITH_VALUE}:${FILEPATH_VALID}"
   )
   _common_container_setup 'CUSTOM_SETUP_ARGUMENTS'
 
+  # ENV is already set explicitly, logs warning:
   CONTAINER_NAME=${CONTAINER2_NAME}
   _init_with_defaults
-  TEST__FILE=${TEST_TMP_CONFIG}/test_secret
   local CUSTOM_SETUP_ARGUMENTS=(
-    --env TEST="manual-secret"
-    --env TEST__FILE="${TEST__FILE}"
+    --env TEST_ENV="manual-secret"
+    --env TEST_ENV__FILE="${FILEPATH_VALID}"
   )
   _common_container_setup 'CUSTOM_SETUP_ARGUMENTS'
 
+  # ENV is not set by file content (invalid file path):
   CONTAINER_NAME=${CONTAINER3_NAME}
   _init_with_defaults
-  NON_EXISTENT__FILE="/tmp/non_existent_secret"
   local CUSTOM_SETUP_ARGUMENTS=(
-    --env TEST__FILE="${NON_EXISTENT__FILE}"
+    --env TEST__FILE="${FILEPATH_INVALID}"
   )
   _common_container_setup 'CUSTOM_SETUP_ARGUMENTS'
 }
 
-teardown_file() {
-  rm -f "${TEST__FILE_SUCCESS}" "${TEST__FILE_WARNING}"
+function teardown_file() {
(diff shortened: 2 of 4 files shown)


polarathene:
I'll review when DMS v15 is released. Presently trying to tackle various docs PRs within that window until then.

target/scripts/startup/variables-stack.sh
```diff
@@ -247,3 +248,24 @@ function __environment_variables_export() {
...
+    local file_path="${!file_env_var}"
+
+    if [[ -n "${!env_var}" ]]; then
+      _log 'warn' "Ignoring ${env_var} since ${file_env_var} is also set"
+      continue
+    fi
+
+    if [[ -f "${file_path}" ]]; then
+      _log 'info' "Getting secret ${env_var} from ${file_path}"
+      export "${env_var}"="$(< "${file_path}")"
```
casperklein, line 266: Is `export` needed, what's the purpose?

test/tests/parallel/set3/container_configuration/env_vars_from_files.bats
```diff
@@ -0,0 +1,62 @@
+load "${REPOSITORY_ROOT}/test/helper/setup"
+load "${REPOSITORY_ROOT}/test/helper/common"
+
+export CONTAINER1_NAME='dms-test_env-files_success'
+export CONTAINER2_NAME='dms-test_env-files_warning'
+export CONTAINER3_NAME='dms-test_env-files_error'
+
+setup_file() {
+  export TEST__FILE
+  TEST__FILE=$(mktemp)
```
casperklein, line 10: Afaik we store temporary files inside the test directory:

https://github.com/docker-mailserver/docker-mailserver/blob/8ca2bd212cc9e582b7d28e21663c547dddc3f9b0/.gitignore#L17-L23

These files will also be deleted by `make clean` (in case the tests are interrupted and `teardown_file()` was not run)

target/scripts/startup/variables-stack.sh
aartoni: Nice catch! Exporting the variables is only needed if we need programs called by the startup script to read them. I expected that to be the case, I'll remove the `export` if that's not the case!

test/tests/parallel/set3/container_configuration/env_vars_from_files.bats
aartoni: Fixing now!

polarathene (commented):
My review feedback is mostly about the feature itself rather than implementation here (_I'll leave that to @casperklein and @georglauterbach as they're both better versed with bash than I am_).

One concern is the tests don't seem to be verifying the value read from file to ENV is correct.
- I have seen other projects implement file loaded ENV where a trailing `\n` or similar was accidentally part of the assigned value.
- **EDIT:** While the output to stdout in a shell would include white-space, assignment from the subshell seems to trim the trailing white-space 🤔 (_so not affected by this caveat, but should probably still have a test for expected vs actual value_)

---

Could we confirm what the expectation of this feature is security wise?

- Is it to avoid the container metadata about ENV the container was configured with? (_`docker inspect <container-name> | jq -r .[].Config.Env` / Docker socket API query_) Both operations should technically be privileged, but it's not uncommon to provide docker socket access to a container with more access than it needs (_even just for `labels` monitoring I think this would intersect the same API call for `environment`/`env_file` on a container - **confirmed**_)
- Or is the concern with ENV visibility within the container from a non-root user/process? This information is presently publicly accessible at `/etc/dms-settings` (_although I don't think it needs to be `644`?_), thus not a major improvement.

This PR would only be addressing the ENV metadata on the container from exposing secrets. If internal file access permissions is a concern that'll probably need to be tackled via a separate PR.

---

## Reference

Linking my feedback from the FR issue:
- https://github.com/docker-mailserver/docker-mailserver/issues/3457#issuecomment-1709430690 (_FR approved for minor convenience, suggested to use filepath as ENV value instead of `__FILE` suffix, although the suffix is more useful when we don't have an explicit list_)
- https://github.com/docker-mailserver/docker-mailserver/issues/3457#issuecomment-2644041130 (_may leak to ENV + written copy to disk_)

**Verified:**
- Both `/root/.bashrc` and `/etc/dms-settings` are written with `644` permissions, but `/root` itself is `700`, thus non-root users can only access `/etc/dms-settings`.
- `env` will have all ENV loaded from `/etc/dms-settings` when using a `bash` shell specifically to trigger the `.bashrc` (_`docker exec -it bash`, or if `sh` then running a login shell via `bash -lc 'env'`_).
- Running bash scripts via `bash` (without `-l`) or relying on the standard shebang we use, neither will load in the ENV, hence why we source it from `/etc/dms-settings` manually when needed.

My earlier comments in the FR also expressed a bit of uncertainty to the value for this feature in DMS, discouraging it given the intention for security benefit seemed minimal?:
- https://github.com/docker-mailserver/docker-mailserver/issues/3457#issuecomment-1666642956
- https://github.com/docker-mailserver/docker-mailserver/issues/3457#issuecomment-1666981189

**NOTE:** In the original FR, I cited Compose `secrets` as not reliable. There has been recent activity to actually get proper [`secrets` support in Docker Compose with `file` working with `uid` / `gid` / `mode`](https://github.com/docker/compose/issues/9648#issuecomment-2731264556) (_still WIP, nothing landed yet_), but that won't really help with the concerns raised for secrets exposure in DMS.
- https://docs.docker.com/reference/compose-file/secrets/
- https://docs.docker.com/reference/compose-file/services/#secrets

casperklein:
I am not sure if I understand you correctly, but using `export` makes a variable available to child processes. `check-for-changes.sh` is started by `supervisord`. So it's not a child of the start script.

polarathene:
> `check-for-changes.sh` is started by `supervisord`. So it's not a child of the start script.

Ok, but the `source /etc/dms-settings` will be run after the start script has finished updating that file yes?

---

Anything within the container that can read `/etc/dms-settings` (`0644` permissions) will have access to the file-based secret values centralized there.

This feature AFAIK is only to prevent secrets visibility from anyone with access to the Docker socket by sending a `GET` HTTP request, similar to how the Docker CLI works:

```bash
# Read the ENV associated to the running container:
docker inspect <container-name> | jq -r .[].Config.Env

# Shell into DMS as with an interactive TTY as root (implicitly starts a login shell `-l`):
# NOTE: Secrets would be visible in ENV loaded from `/etc/dms-settings` via `/root/.bashrc`
docker exec -it <container-name> bash
```

---

I think this feature makes more sense when the secret is read temporarily into memory to use it, but due to how DMS is managing the secrets currently, this is to effectively only defend against a separately compromised container inspecting metadata through the Docker socket.

I just want to clarify with the author (_and any others interested in the feature_) if that meets their actual expectations of implementing this feature, or if it falls short of the expected security benefit.

If @aartoni still wants to proceed with that security benefit clarified, then they just need to ensure the PR has a test-case that verifies the target ENV has the expected value resolved by the `__FILE` variant.

aartoni:
@polarathene I'm willing to proceed. I'll get in touch in a few days in case I get stuck.

aartoni:
Hello @polarathene! I got stuck while trying to address [this comment](https://github.com/docker-mailserver/docker-mailserver/pull/4359#discussion_r2002119943).

It is not clear to me how to obtain the path for the copy of the configuration that I should be using in this case, here's what I have tried:

```sh
TEST_TMP_CONFIG=$(_duplicate_config_for_container . 'env_vars_from_files')
```

polarathene:
We have a variety of tests you can reference that do this :)

`TEST_TMP_CONFIG` should be [already managed for you](https://github.com/docker-mailserver/docker-mailserver/blob/ef66dd5d12c4f0efa2b2e345ffdce796e3a7df03/test/helper/setup.bash#L93-L112) (_duplicated for each container configured in `setup_file()`_), so like the examples below, you can use `mv` or `cp` to place files where you need 👍

If you need to create files instead of `cp`/`mv`, you should be able to create files at the `TEST_TMP_CONFIG` location too. Just make sure you do this after calling `_init_with_defaults` first (_configures `TEST_TMP_CONFIG` for the test_) and before calling `_common_container_setup` (_starts the container with `TEST_TMP_CONFIG` volume mounted_).

<details>
<summary>Reference of tests manipulating configs from `TEST_TMP_CONFIG` (click to view)</summary>

https://github.com/docker-mailserver/docker-mailserver/blob/ef66dd5d12c4f0efa2b2e345ffdce796e3a7df03/test/tests/parallel/set1/dovecot/dovecot_sieve.bats#L13-L14

https://github.com/docker-mailserver/docker-mailserver/blob/ef66dd5d12c4f0efa2b2e345ffdce796e3a7df03/test/tests/parallel/set1/config_overrides.bats#L10-L11

https://github.com/docker-mailserver/docker-mailserver/blob/ef66dd5d12c4f0efa2b2e345ffdce796e3a7df03/test/tests/parallel/set1/fetchmail.bats#L11-L17

https://github.com/docker-mailserver/docker-mailserver/blob/ef66dd5d12c4f0efa2b2e345ffdce796e3a7df03/test/tests/parallel/set3/mta/lmtp_ip.bats#L19-L29

https://github.com/docker-mailserver/docker-mailserver/blob/ef66dd5d12c4f0efa2b2e345ffdce796e3a7df03/test/tests/parallel/set3/mta/dsn.bats#L20-L26

https://github.com/docker-mailserver/docker-mailserver/blob/ef66dd5d12c4f0efa2b2e345ffdce796e3a7df03/test/tests/parallel/set3/mta/smtp_delivery.bats#L44-L48

https://github.com/docker-mailserver/docker-mailserver/blob/ef66dd5d12c4f0efa2b2e345ffdce796e3a7df03/test/tests/parallel/set3/scripts/setup_cli.bats#L12-L14

https://github.com/docker-mailserver/docker-mailserver/blob/ef66dd5d12c4f0efa2b2e345ffdce796e3a7df03/test/tests/parallel/set3/scripts/replace_by_env_in_file.bats#L7-L12

</details>

---

## Extra notes

- If you need to have files provided already, you can commit them to [this directory](https://github.com/docker-mailserver/docker-mailserver/tree/v15.0.0/test/config) which is the contents of `TEST_TMP_CONFIG` (_duplicated per container setup_).

- `_init_with_defaults()` duplicates the test config directory for you setting it to `TEST_TMP_CONFIG`:

  https://github.com/docker-mailserver/docker-mailserver/blob/ef66dd5d12c4f0efa2b2e345ffdce796e3a7df03/test/helper/setup.bash#L96-L97

  `TEST_TMP_CONFIG` is then added as a volume mount for the container to use as `/tmp/docker-mailserver` (DMS Config Volume):

  https://github.com/docker-mailserver/docker-mailserver/blob/ef66dd5d12c4f0efa2b2e345ffdce796e3a7df03/test/helper/setup.bash#L108

  https://github.com/docker-mailserver/docker-mailserver/blob/ef66dd5d12c4f0efa2b2e345ffdce796e3a7df03/test/helper/setup.bash#L158

- We also document here how you can create a separate `TEST_TMP_CONFIG` on demand when needed (see the example portion):

  https://github.com/docker-mailserver/docker-mailserver/blob/ef66dd5d12c4f0efa2b2e345ffdce796e3a7df03/test/helper/setup.bash#L77-L92

- Automated removal happens here:

  https://github.com/docker-mailserver/docker-mailserver/blob/ef66dd5d12c4f0efa2b2e345ffdce796e3a7df03/Makefile#L27-L33

  Which will remove all the duplicated config folders tests created at `test/duplicate_configs/`:

  https://github.com/docker-mailserver/docker-mailserver/blob/ef66dd5d12c4f0efa2b2e345ffdce796e3a7df03/.gitignore#L21

aartoni:
Thank you @polarathene for referencing the relevant code. I have implemented the tests the way you pointed out, I'd say that this is ready for the final review!

target/scripts/startup/variables-stack.sh
```diff
@@ -247,3 +248,24 @@ function __environment_variables_export() {
...
+    fi
+
+    if [[ -f "${file_path}" ]]; then
+      _log 'info' "Getting secret ${env_var} from ${file_path}"
+      printf -v "${env_var}" '%s' "$(< "${file_path}")"
+    else
+      _log 'error' "File ${file_path} does not exist, defined in ${file_env_var}"
+    fi
+  done
+}
```
polarathene, line 271:
```suggestion
# This function sets any environment variable with a value from a referenced file
# when an equivalent ENV with a `__FILE` suffix exists with a valid file path as the value.
function __environment_variables_from_files() {
  for ENV_WITH_FILE_REF in $(env | grep -Po '^.+?__FILE'); do
    local TARGET_ENV="${ENV_WITH_FILE_REF/__FILE/}"
    local FILE_PATH="${!ENV_WITH_FILE_REF}"

    if [[ -n "${!TARGET_ENV}" ]]; then
      _log 'warn' "Ignoring ${TARGET_ENV} since ${ENV_WITH_FILE_REF} is also set"
      continue
    fi

    if [[ -f "${FILE_PATH}" ]]; then
      _log 'info' "Getting secret ${TARGET_ENV} from ${FILE_PATH}"
      printf -v "${TARGET_ENV}" '%s' "$(< "${FILE_PATH}")"
    else
      _log 'error' "File ${FILE_PATH} does not exist, defined in ${ENV_WITH_FILE_REF}"
    fi
  done
}
```

test/tests/parallel/set3/container_configuration/env_vars_from_files.bats
```diff
@@ -0,0 +1,79 @@
...
+  rm -f "${FILE_WITH_VALUE}"
+  docker rm -f "${CONTAINER1_NAME}" "${CONTAINER2_NAME}" "${CONTAINER3_NAME}"
+}
+
+@test "ENV can be set from a file" {
+  export CONTAINER_NAME=${CONTAINER1_NAME}
+
+  run docker logs "${CONTAINER_NAME}"
+  assert_success
+  assert_line --partial "Getting secret ENABLE_POP3 from ${FILEPATH_VALID}"
```
polarathene, line 59: Each of these `run docker logs` lines should be able to use [one of our helper methods](https://github.com/docker-mailserver/docker-mailserver/blob/c9aac24a7256352f537aa32b149fcf5a4e4c7706/test/helper/log_and_filtering.bash#L21-L33) for asserting logs.

```suggestion
  _service_log_should_contain_string 'mail' "Getting secret ENABLE_POP3 from ${FILEPATH_VALID}"
```

That will use grep with the string you used for `assert_line --partial`, and apply `assert_success` internally. It's a bit more clear that we expect this log line in `mail.log`.

test/tests/parallel/set3/container_configuration/env_vars_from_files.bats
```diff
@@ -0,0 +1,79 @@
...
+
+  _run_in_container [ -f /etc/dovecot/protocols.d/pop3d.protocol ]
+  assert_success
```
polarathene, line 62:
```suggestion
  # Verify ENABLE_POP3 was enabled (disabled by default), by checking this file path is valid:
  _run_in_container [ -f /etc/dovecot/protocols.d/pop3d.protocol ]
  assert_success
```

test/tests/parallel/set3/container_configuration/env_vars_from_files.bats
```diff
@@ -0,0 +1,79 @@
...
+  local CUSTOM_SETUP_ARGUMENTS=(
+    --env TEST__FILE="${FILEPATH_INVALID}"
+  )
+  _common_container_setup 'CUSTOM_SETUP_ARGUMENTS'
+}
+
+function teardown_file() {
+  rm -f "${FILE_WITH_VALUE}"
+  docker rm -f "${CONTAINER1_NAME}" "${CONTAINER2_NAME}" "${CONTAINER3_NAME}"
+}
```
polarathene, line 52: Technically, there shouldn't be a need to remove this file. It's in the containers temp config directory so it should be cleared when using [`make clean`](https://github.com/docker-mailserver/docker-mailserver/blob/c9aac24a7256352f537aa32b149fcf5a4e4c7706/Makefile#L29).

There's no mutation or sharing concern either so it shouldn't make a difference regardless for the tests.

```suggestion
function teardown_file() {
  docker rm -f "${CONTAINER1_NAME}" "${CONTAINER2_NAME}" "${CONTAINER3_NAME}"
}
```

test/tests/parallel/set3/container_configuration/env_vars_from_files.bats
```diff
@@ -0,0 +1,79 @@
...
+  _common_container_setup 'CUSTOM_SETUP_ARGUMENTS'
+
+  # ENV is already set explicitly, logs warning:
+  CONTAINER_NAME=${CONTAINER2_NAME}
+  _init_with_defaults
+  local CUSTOM_SETUP_ARGUMENTS=(
+    --env TEST_ENV="manual-secret"
+    --env TEST_ENV__FILE="${FILEPATH_VALID}"
+  )
+  _common_container_setup 'CUSTOM_SETUP_ARGUMENTS'
```
polarathene, line 38: Should the file still be mounted? I understand it shouldn't be relevant for the current logic, but for the path to be valid we could better match the first test case?

```suggestion
  # ENV is already set explicitly, a warning should be logged:
  CONTAINER_NAME=${CONTAINER2_NAME}
  _init_with_defaults
  local CUSTOM_SETUP_ARGUMENTS=(
    --env ENABLE_POP3="0"
    --env ENABLE_POP3__FILE="${FILEPATH_VALID}"
    -v "${FILE_WITH_VALUE}:${FILEPATH_VALID}"
  )
  _common_container_setup 'CUSTOM_SETUP_ARGUMENTS'
```

test/tests/parallel/set3/container_configuration/env_vars_from_files.bats
```diff
@@ -0,0 +1,79 @@
...
+CONTAINER2_NAME='dms-test_env-files_warning'
+CONTAINER3_NAME='dms-test_env-files_error'
+
+function setup_file() {
+  export CONTAINER_NAME
+  export FILEPATH_VALID='/tmp/file-with-value'
+  export FILEPATH_INVALID='/path/to/non-existent-file'
+
+  export FILE_WITH_VALUE="${TEST_TMP_CONFIG}/test_secret"
+  echo 1 > "${FILE_WITH_VALUE}"
```
polarathene, line 20: I tidied these up from what you previously had, since the container feature and test cases are more focused on the file path being valid or not, this made more sense to me.

polarathene (changes requested):
Almost there 💪

test/tests/parallel/set3/container_configuration/env_vars_from_files.bats
aartoni: This seems the be the same as [`_should_fail_to_extract_for_wildcard_env`](https://github.com/docker-mailserver/docker-mailserver/blob/master/test/tests/parallel/set1/tls/letsencrypt.bats#L191-L202) from `letsencrypt.bats` where no log file contains the relevant information.

Here's an excerpt:

```bash
# /var/log/mail/mail.log is not equivalent to stdout content,
# Relevant log content only available via docker logs:
run docker logs "${CONTAINER_NAME}"
assert_output --partial "_extract_certs_from_acme | Unable to find key and/or cert for '${DOMAIN_WILDCARD}' in '/etc/letsencrypt/acme.json'"
assert_output --partial "_extract_certs_from_acme | Certificate successfully extracted for '${DOMAIN_MAIL}'"
```

aartoni: I've added the same comment I found in the above snippet to make it clear

test/tests/parallel/set3/container_configuration/env_vars_from_files.bats
```diff
@@ -0,0 +1,88 @@
...
+BATS_TEST_NAME_PREFIX='[Configuration] (ENV __FILE support) '
+CONTAINER1_NAME='dms-test_env-files_success'
+CONTAINER2_NAME='dms-test_env-files_warning'
+CONTAINER3_NAME='dms-test_env-files_error'
+
+function setup_file() {
+  export CONTAINER_NAME
+  export FILEPATH_VALID='/tmp/file-with-value'
+  export FILEPATH_INVALID='/path/to/non-existent-file'
+  export FILE_WITH_VALUE
```
polarathene, line 18:
```suggestion
  # Each `_init_with_defaults` call updates the `TEST_TMP_CONFIG` location to create a container specific file:
  local FILE_WITH_VALUE
```

polarathene (approved):
LGTM 👍 

Just waiting on @casperklein or @georglauterbach for their opinion on https://github.com/docker-mailserver/docker-mailserver/pull/4359/files#r2085827754

target/scripts/startup/variables-stack.sh
```diff
@@ -244,3 +245,28 @@ function __environment_variables_export() {
...
+    fi
+
+    # Otherwise retrieve the value from file and set the ENV or fail if invalid reference:
+    if [[ -f "${FILE_PATH}" ]]; then
+      _log 'info' "Getting secret ${TARGET_ENV} from ${FILE_PATH}"
+      printf -v "${TARGET_ENV}" '%s' "$(< "${FILE_PATH}")"
+    else
+      _log 'error' "File ${FILE_PATH} does not exist, defined in ${ENV_WITH_FILE_REF}"
+    fi
+  done
```
georglauterbach, line 271: 1. This `for` loop should become a `while read` construct. The reason is that the current approach relies on a single string being split at whitespaces, which is quite unstable. The following approach is a little neater:

```suggestion
  while read -r ENV_WITH_FILE_REF; do
    # Store the ENV name without the `__FILE` suffix:
    local TARGET_ENV_NAME="${ENV_WITH_FILE_REF/__FILE/}"
    local -n TARGET_ENV="${ENV_WITH_FILE_REF/__FILE/}"
    # Store the value of the `__FILE` ENV:
    local FILE_PATH="${!ENV_WITH_FILE_REF}"

    # Skip sourcing form `__FILE` if ENV is already set:
    if [[ -v TARGET_ENV ]]; then
      _log 'warn' "Ignoring '${TARGET_ENV_NAME}' since '${ENV_WITH_FILE_REF}' is also set"
      continue
    fi

    # Otherwise, retrieve the value from a file and set
    # the ENV, or fail if the reference is invalid:
    if [[ ! -f ${FILE_PATH} ]]; then
      _log 'warn' "File defined for secret '${TARGET_ENV_NAME}' with path '${FILE_PATH}' does not exist"
      continue
    fi

    _log 'info' "Getting secret '${TARGET_ENV_NAME}' from '${FILE_PATH}'"
    TARGET_ENV="$(< "${FILE_PATH}")"
  done < <(env | grep -Po '^.+?__FILE')
```

2. I have also removed the `"` where not necessary.
3. The `[[ -n ${!VAR} ]]` check should actually be `[[ -v ${VAR} ]]`, as the `-v` flag was introduced with Bash 5 for checking whether a variable is set.
4. Then I realized we can use `local -n`, which makes things easier to understand. `local -n` saves name like a "reference":

  ```bash
  function x() { local -n X="${1:?VAR required}" ; [[ -v X ]] && echo "yes: $X" ; X='lol' ; return 0; }
  SOME_ENV_VAR=abc
  x SOME_ENV_VAR # prints 'yes: abc'
  echo "${SOME_ENV_VAR}" # prints 'lol'
  ```
6. Hence, the `printf -v` can go a well.

I hope I did not accidentally changed something for the worse.. please double-check my logic before applying this suggestion.

georglauterbach (changes requested):
I prefer the `local -n` approach outlined in my suggestion. I hope you agree with my idea :) Please double-check it, though :) Other than that: LGTM, nice addition 👍🏼

target/scripts/startup/variables-stack.sh
```diff
@@ -247,3 +248,28 @@ function __environment_variables_export() {
...
+    # Skip sourcing form `__FILE` if ENV is already set:
+    if [[ -n "${!TARGET_ENV}" ]]; then
+      _log 'warn' "Ignoring ${TARGET_ENV} since ${ENV_WITH_FILE_REF} is also set"
+      continue
+    fi
+
+    # Otherwise retrieve the value from file and set the ENV or fail if invalid reference:
+    if [[ -f "${FILE_PATH}" ]]; then
+      _log 'info' "Getting secret ${TARGET_ENV} from ${FILE_PATH}"
+      printf -v "${TARGET_ENV}" '%s' "$(< "${FILE_PATH}")"
```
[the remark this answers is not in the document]
georglauterbach: Resolving this in favor of <https://github.com/docker-mailserver/docker-mailserver/pull/4359#discussion_r2089297086>.

target/scripts/startup/variables-stack.sh
polarathene: I'm fine with `local -n` as we are using that in the main test helper to create containers. We may want to document it here for the less familiar though so they don't have to try search it up? 😅 (_same with `-v` which I assume will be similarly niche to see in typical shell scripts? The comment for that condition already implies that so we're good there_)

You introduced a new ENV `TARGET_ENV_NAME` but set the same value to `TARGET_ENV` (_except this one due to `-n` became a reference_). It would still be equivalent if we took a reference of `TARGET_ENV_NAME` right?

```
# Assign a value representing a variable name,
# `-n` will alias `TARGET_ENV` so that it is treated as if it were the referenced variable:
local -n TARGET_ENV="${TARGET_ENV_NAME}"
```

---

> I hope I did not accidentally change something for the worse.. please double-check my logic before applying this suggestion.

Looks good to me, just the commentary at the end where you changed the if/else to a continue condition for the failure/error, minor adjustment needed there.

polarathene: Revised with my feedback from prior comment:

```bash
  while read -r ENV_WITH_FILE_REF; do
    # Store the value of the `__FILE` ENV:
    local FILE_PATH="${!ENV_WITH_FILE_REF}"
    # Store the ENV name without the `__FILE` suffix:
    local TARGET_ENV_NAME="${ENV_WITH_FILE_REF/__FILE/}"
    # Assign a value representing a variable name,
    # `-n` will alias `TARGET_ENV` so that it is treated as if it were the referenced variable:
    local -n TARGET_ENV="${TARGET_ENV_NAME}"

    # Skip if the target ENV is already set:
    if [[ -v TARGET_ENV ]]; then
      _log 'warn' "Ignoring '${TARGET_ENV_NAME}' since '${ENV_WITH_FILE_REF}' is also set"
      continue
    fi

    # Skip if the file path provided is invalid:
    if [[ ! -f ${FILE_PATH} ]]; then
      _log 'warn' "File defined for secret '${TARGET_ENV_NAME}' with path '${FILE_PATH}' does not exist"
      continue
    fi

    # Read the value from a file and assign it to the intended ENV:
    _log 'info' "Getting secret '${TARGET_ENV_NAME}' from '${FILE_PATH}'"
    TARGET_ENV="$(< "${FILE_PATH}")"
  done < <(env | grep -Po '^.+?__FILE')
```

casperklein:
I'll review this on the weekend.

target/scripts/startup/variables-stack.sh
```diff
@@ -244,3 +245,31 @@ function __environment_variables_export() {
...
+    # Otherwise, retrieve the value from a file and set
+    # the ENV, or fail if the reference is invalid:
+    if [[ ! -f ${FILE_PATH} ]]; then
+      _log 'warn' "File defined for secret '${TARGET_ENV_NAME}' with path '${FILE_PATH}' does not exist"
+      continue
+    fi
+
+    _log 'info' "Getting secret '${TARGET_ENV_NAME}' from '${FILE_PATH}'"
+    TARGET_ENV="$(< "${FILE_PATH}")"
+  done < <(env | grep -Po '^.+?__FILE')
```
polarathene, line 274:
```suggestion
  # Iterate through all ENV found with a `__FILE` suffix:
  while read -r ENV_WITH_FILE_REF; do
    # Store the value of the `__FILE` ENV:
    local FILE_PATH="${!ENV_WITH_FILE_REF}"
    # Store the ENV name without the `__FILE` suffix:
    local TARGET_ENV_NAME="${ENV_WITH_FILE_REF/__FILE/}"
    # Assign a value representing a variable name,
    # `-n` will alias `TARGET_ENV` so that it is treated as if it were the referenced variable:
    local -n TARGET_ENV="${TARGET_ENV_NAME}"

    # Skip if the target ENV is already set:
    if [[ -v TARGET_ENV ]]; then
      _log 'warn' "Ignoring '${TARGET_ENV_NAME}' since '${ENV_WITH_FILE_REF}' is also set"
      continue
    fi

    # Skip if the file path provided is invalid:
    if [[ ! -f ${FILE_PATH} ]]; then
      _log 'warn' "File defined for secret '${TARGET_ENV_NAME}' with path '${FILE_PATH}' does not exist"
      continue
    fi

    # Read the value from a file and assign it to the intended ENV:
    _log 'info' "Getting secret '${TARGET_ENV_NAME}' from '${FILE_PATH}'"
    TARGET_ENV="$(< "${FILE_PATH}")"
  done < <(env | grep -Po '^.+?__FILE')
```

test/tests/parallel/set3/container_configuration/env_vars_from_files.bats
```diff
@@ -0,0 +1,89 @@
...
+    --env ENABLE_POP3__FILE="${FILEPATH_VALID}"
+    -v "${FILE_WITH_VALUE}:${FILEPATH_VALID}"
+  )
+  _common_container_setup 'CUSTOM_SETUP_ARGUMENTS'
+
+  # ENV is not set by file content (invalid file path):
+  CONTAINER_NAME=${CONTAINER3_NAME}
+  _init_with_defaults
+  local CUSTOM_SETUP_ARGUMENTS=(
+    --env TEST__FILE="${FILEPATH_INVALID}"
```
polarathene, line 48:
```suggestion
    --env ENABLE_POP3__FILE="${FILEPATH_INVALID}"
```

test/tests/parallel/set3/container_configuration/env_vars_from_files.bats
```diff
@@ -0,0 +1,89 @@
...
+}
+
+@test "Referencing a non-existent file logs an error" {
+  export CONTAINER_NAME=${CONTAINER3_NAME}
+
+  # /var/log/mail/mail.log is not equivalent to stdout content,
+  # Relevant log content only available via docker logs:
+  run docker logs "${CONTAINER_NAME}"
+  assert_success
+  assert_line --partial "File defined for secret 'TEST' with path '${FILEPATH_INVALID}' does not exist"
```
polarathene, line 88:
```suggestion
  assert_line --partial "File defined for secret 'ENABLE_POP3' with path '${FILEPATH_INVALID}' does not exist"
```

polarathene (changes requested):

polarathene (approved):

georglauterbach (approved):

casperklein (approved):

PR #4360, opened 2025-02-14
docs: Revise Rspamd page

polarathene:
# Description

I initially wanted to just fix a typo on the ARC section, then rewrote that section and got carried away tidying up the docs for this page 😅 

[Spotted a possible issue](https://github.com/rspamd/rspamd/issues/5341) for upstream to resolve in the process 🎉 (_nothing concerning, just unexpected_)

**NOTE:** I'm now referring to our existing anti-spam services as "Legacy", I'm not sure if that's accurate/fitting, let me know if there's a better alternative if that's not ok.

---

Apologies again for the verbose diff and single commit. I'm trying to get quite a bit of docs updates through for DMS v15 release and keep getting distracted 😮‍💨

### DIFF
--- a/docs/content/config/security/rspamd.md
+++ b/docs/content/config/security/rspamd.md
@@ -175,7 +175,7 @@ You can choose to enable ClamAV, and Rspamd will then use it to check for viruse
 
 #### RBLs (Real-time Blacklists) / DNSBLs (DNS-based Blacklists)
 
-The [RBL module][rspamd-docs::modules::rbl] is enabled by default. As a consequence, Rspamd will perform DNS lookups to various blacklists. Whether an RBL or a DNSBL is queried depends on where the domain name was obtained: RBL servers are queried with IP addresses extracted from message headers, DNSBL server are queried with domains and IP addresses extracted from the message body \[[source][www::rbl-vs-dnsbl]\].
+The [RBL module][rspamd-docs::modules::rbl] is enabled by default. As a consequence, Rspamd will perform DNS lookups to various blacklists. Whether an RBL or a DNSBL is queried depends on where the domain name was obtained: RBL servers are queried with IP addresses extracted from message headers, DNSBL server are queried with domains and IP addresses extracted from the message body ([source][www::rbl-vs-dnsbl]).
 
 !!! danger "Rspamd and DNS Block Lists"
 
--- a/docs/content/config/security/rspamd.md
+++ b/docs/content/config/security/rspamd.md
@@ -105,12 +105,15 @@ DMS does not set a default password for the controller worker. You may want to d
 When Rspamd is enabled, we implicitly also start an instance of Redis in the container:
 
 - Redis is configured to persist its data via RDB snapshots to disk in the directory `/var/lib/redis` (_or the [`/var/mail-state/`][docs::dms-volumes-state] volume when present_).
-- With the volume mount, the snapshot will restore the Redis data across container restarts, and provide a way to keep backup.
+- With the volume mount, the snapshot will restore the Redis data across container updates, and provide a way to keep a backup.
+- Without a volume mount a containers internal state will persist across restarts until the container is recreated due to changes like ENV or upgrading the image for the container.
 
 Redis uses `/etc/redis/redis.conf` for configuration:
 
 - We adjust this file when enabling the internal Redis service.
 - If you have an external instance of Redis to use, the internal Redis service can be opt-out via setting the ENV [`ENABLE_RSPAMD_REDIS=0`][docs::env::enable-redis] (_link also details required changes to the DMS Rspamd config_).
+
+If you are interested in using Valkey instead of Redis, please refer to [this guidance][gh-dms::guide::valkey].
 
 ### Web Interface
 
@@ -122,7 +125,7 @@ To use the web interface you will need to configure a password, [otherwise you w
 
 ??? example "Set a custom password"
 
-    Add this line to [your rspamd `custom-commands.conf` config](#with-the-help-of-a-custom-file) which sets the `password` option of the _controller worker_:
+    Add this line to [your Rspamd `custom-commands.conf` config](#with-the-help-of-a-custom-file) which sets the `password` option of the _controller worker_:
 
     ```
     set-option-for-controller password "your hashed password here"
@@ -133,6 +136,10 @@ To use the web interface you will need to configure a password, [otherwise you w
     ```bash
     docker exec -it <CONTAINER_NAME> rspamadm pw
     ```
+
+    ---
+
+    **Related:** A minimal Rspamd `compose.yaml` [example with a reverse-proxy for web access][gh-dms:guide::rspamd-web].
 
 ### DNS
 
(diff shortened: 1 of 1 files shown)


polarathene:
I have a question for @georglauterbach are the repo commited `local.d/` configs copy/paste from the Rspamd package/repo at the time?

For example `rspamd/local.d/worker-proxy.inc` was committed in Jan 2023:

https://github.com/docker-mailserver/docker-mailserver/blob/v14.0.0/target/rspamd/local.d/worker-proxy.inc

Rspamd was initially introduced with DMS v12 (Debian 11 Bullseye) released in April 2024 with Rspamd 3.5 (_[upstream `worker-proxy.inc`](https://github.com/rspamd/rspamd/blob/3.5.0/conf/worker-proxy.inc) for 3.5_) (_Debian package [would have been 3.4](https://packages.debian.org/search?keywords=rspamd)_). Looking at the upstream `git blame` the config added the `reject_message` line with the other options, so it's unclear why your committed copy differs from upstream by omitting that one.

For our image the equivalent upstream file lives at `/etc/rspamd/worker-proxy.inc`, with a roughly similar config from us at `/etc/rspamd/local.d/worker-proxy.inc` to override that:

https://github.com/docker-mailserver/docker-mailserver/blob/f2fedff251a91da3bfc1f780a17b30e1346a53a3/Dockerfile#L96

<details>
<summary>Diff output</summary>

```console
$ diff /etc/rspamd/worker-proxy.inc /etc/rspamd/local.d/worker-proxy.inc

1,13c1,2
< # Proxy worker setup
< # Please don't modify this file as your changes might be overwritten with
< # the next update.
< #
< # You can modify 'local.d/worker-proxy.inc' to add and merge
< # parameters defined inside this section
< #
< # You can modify 'override.d/worker-proxy.inc' to strictly override all
< # parameters defined inside this section
< #
< # See https://rspamd.com/doc/faq.html#what-are-the-locald-and-overrided-directories
< # for details
< # Module documentation: https://rspamd.com/doc/workers/rspamd_proxy.html
---
> # documentation: https://rspamd.com/doc/workers/rspamd_proxy.html
> # see also: https://rspamd.com/doc/quickstart.html#using-of-milter-protocol-for-rspamd--16
15c4
< milter = yes; # Enable milter mode
---
> bind_socket = "127.0.0.1:11332";
17,23c6,7
< # This timeout is mostly specific to the milter mode.
< # Please also bear in mind that if this timeout is longer than `task_timeout`,
< # your messages might be processed for much longer due to this timeout (this is
< # true for self-scan mode).
< # If this behaviour is not desired, then it is recommended to reduce and adjust this
< # value accordingly
< timeout = 60s;
---
> milter = yes;
> timeout = 120s;
27c11
<   hosts = "localhost";
---
>   self_scan = yes;
30,35c14,18
< count = 1; # Do not spawn too many processes of this type
< max_retries = 5; # How many times master is queried in case of failure
< discard_on_reject = false; # Discard message instead of rejection
< quarantine_on_reject = false; # Tell MTA to quarantine rejected messages
< spam_header = "X-Spam"; # Use the specific spam header
< reject_message = "Spam message rejected"; # Use custom rejection message
---
> count = 2;
> max_retries = 5;
> discard_on_reject = false;
> quarantine_on_reject = false;
> spam_header = "X-Spam";
```

</details>

---

If possible, I'd like to avoid this becoming an issue over time, similar to the Dovecot config situation. I've largely kept my distance from rspamd in the project but have become a bit more familiar with it lately as we get closer to switching over to it.

I'm a bit iffy about the `custom-commands.conf` feature, but I hope the revised docs for that in this PR communicate it more clearly. I found the existing docs a bit sporadic to grok and needed to double check by going over your script implementation for it to understand properly. Didn't help that that rspamd docs were another layer of friction, I've better conveyed those insights in the example comments.

polarathene:
I'll do another pass on the page in future when I have more time.

Shifts the config example to switch to Rspamd to the very top of the docs, with the related ENV in a separate section below that. I think that'd be a bit more helpful to the reader for the first time?


# Visual diff

Bulk of the changes are covered below, bit easier if you want to compare 😅 

## Early guidance for how to get going with Rspamd

### Before

**NOTE:** "Other Anti-Spam Services" section seemed a bit redundant with the revision so was removed:

> image

Near the end of the docs page:

> image

### After

> image

---

## `rspamd/override.d/` + `custom-commands.conf`

### Before

I wasn't a fan of the admonition usage and found these two sections a bit more effort to grok as a new reader to them than I felt they needed to be.

> image

> image

> image

### After

Seeing this rendered it could probably be improved a bit, but hopefully it's a step in the right direction 😅 

> image

> image

> image

> image

## DKIM / ARC sections

Slight improvement here, I'll tackle the unified DKIM feature which will simplify the config further 👍 

### Before

> image

### After

> image

casperklein (dismissed):

georglauterbach:
I'm a big fan of admonitions, but the extend to which they are used in our docs leads me to the conclusion that you use them to structure content, instead of using them as additional hints.

Admonitions are similar to GitHub's "Alerts", on which [GitHub advises](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts):

> Use alerts only when they are crucial for user success and limit them to one or two per article to prevent overloading the reader.

I agree with this, and hence, I feel like the content on these docs would be better represented without this amount of admonitions.

Would you hence be okay with removing a few admonitions and replacing them with headings again?

polarathene:
> I'm a big fan of admonitions, but the extend to which they are used in our docs leads me to the conclusion that you use them to structure content, instead of using them as additional hints.

You are correct, I do over use them while polishing up the docs as personally it makes the content easier to navigate and digest for me than without them.

This concern though is specific to our current docs generator poor typography.

While I would be more familiar with Meta's Docusaurus project due to it's usage of React... I have noticed these two sites docs are using VuePress and I quite like the user experience, both the sidebar for navigation and typography layout. The admonition style less so (_see the nushell link for that_), but tbh that's probably better since it only further promotes it to be used more minimally 😛 

- https://doc.dovecot.org/main/
- https://www.nushell.sh/book/quick_tour.html#acting-on-data

Once the docs have had a full revision pass and been re-organized (_I plan to shift pages around by DMS v16_), should I have the time I'll look into alternatives to mkdocs-material. I'd prefer to minimize our CSS workarounds for issues upstream won't fix, and better typography/nav would be fantastic too.

Markdown will still be used, non-standard features like admonitions will look a tad different. Admonitions differ in name, for VuePress they're called [Custom Containers](https://v1.vuepress.vuejs.org/guide/markdown.html#custom-containers) instead (_that is v1, their v2 is in RC stage and [seems to be lacking the feature](https://v2.vuepress.vuejs.org/guide/markdown.html), seems [to be moved to a plugin](https://ecosystem.vuejs.press/themes/default/markdown.html#custom-containers)_), and they use the same syntax as mkdocs-material plans to migrate to IIRC (_we could look into opt-in for that if they're resolved some issues I recall it having last time I investigated_).

I don't know if I'm onboard with VuePress itself, but those two references using it at least demonstrate the improvements I'd like our docs to have.

---

> Would you hence be okay with removing a few admonitions and replacing them with headings again?

As mentioned, I'd rather defer that until I've done a full pass over our docs.

No users have been complaining about it so presently it's just a preference/bias difference between us, but we'll be in alignment once content has better structure from improved typography in future 👍

polarathene:
FWIW, regarding admonitions I strongly dislike the use of them as shown in the before screenshots for questions where it wraps the title with no supplementary content. That was done again for an example link.

We have collapsed admonitions which can hide excess related information which is useful for UX, while the other usage you had on this page felt like an anti-pattern to me and conflicted with that UX that implies (at a glance) expandable admonitions 😅 

I personally don't see comparisons like the ARC section to be better without admonitions.

---

Just noticed my `custom-commands.conf` section has a broken admonition rendered, I'll fix that.

georglauterbach (approved):

Issue #3561, opened 2023-10-02
feature request: Vector for log management


polarathene:
### Context

[`Vector`](https://vector.dev/) for easier and better documented configuration, offering more flexible log management than the current `rsyslog` package supports (_syslog focused only_):
- Forwarding logs to external instance of Fail2Ban: https://github.com/docker-mailserver/docker-mailserver/issues/3178#issuecomment-1477101386
- Similar brief mention of adopting Vector for better log management with Redis and Rspamd (_also briefly touches on syslog support, and some other redis config to consider unrelated to logging_):
  - https://github.com/docker-mailserver/docker-mailserver/issues/3138#issuecomment-1453519599
  - https://github.com/docker-mailserver/docker-mailserver/issues/3138#issuecomment-1454251202
  - https://github.com/docker-mailserver/docker-mailserver/pull/3143#pullrequestreview-1324908603
- Process to approach the change discussed here: https://github.com/docker-mailserver/docker-mailserver/pull/3131#issuecomment-1452768189
- Overview / Comparision to `rsyslog` for why to adopt Vector detailed here: https://github.com/docker-mailserver/docker-mailserver/pull/3134#issuecomment-1450935046

Another maintainer approving interest in adopting Vector: https://github.com/docker-mailserver/docker-mailserver/pull/3134#issuecomment-1451031308

### Description

This has been considered for a while. I just haven't properly documented / discussed an issue for going forward.

It's not likely to be a quick and easy change to implement, tests will likely need to be updated and current logging configuration reviewd/revised. Docs should also be added to better communicate the support to users that want to leverage it.

### Alternatives

I think there was some improvements that could be done with `rsyslog` config, but beyond that there were not many other appealing options within a container that I came across. 

Vector looks very nice but needs further investigation to verify as a suitable fit for the project that actually provides real value to ship with the image without increasing burden to maintainers.

### Applicable Users

Anyone that needs better control of log management than we currently have. I am particularly interested in it as our current docs are lacking in this area, our log support with DMS is rather limited AFAIK.

[...]

Issue #1300, opened 2019-10-19
clamd fails to start due to missing logfile permissions


KampfCaspar:
Upon starting a :latest from today, I get an infinite startup loop for clamd. 

Trying to start clamd manually, I see it cannot initialize its logger as its missing rights to /var/log/mail/clamav.log

After 'touching' /var/log/mail/clamav.log with rights for clamav:clamav, the issue is (at least temporarily) resolved.

erik-wramner:
Are you using a volume for the logs? If so, mounted from file system or using a docker volume? And have you changed your settings, or did you just do the upgrade?

We could add a touch step, but I wonder why it is needed. I certainly don't get this problem.

KampfCaspar:
It was a fresh install. Yes, I keep /var/mail/log in a filesystem directory, not a docker volume. The default permissions to the log directory were root write only; mail.(log|info|warn) are created as root and were therefore ok. Only clamd seems to want to create its logfile as non-root user and therefore fails.

The 'most correct' solution would probably be to check/correct permissions on the log directory upon startup. I didn't check what permissions clamd keeps up until what point.

erik-wramner:
Indeed. This just works with volumes, but you are the second person running into this thing with a mounted directory, so I guess we should fix it.

erik-wramner:
See #1305

erik-wramner:
This should be fixed in latest now, please verify and close if it works.

KampfCaspar:
Seems fine. thx

PR #4370, opened 2025-02-16
fix(log): su syslog root

MichaelSp:
# Description

This addresses an issue raised 
https://github.com/docker-mailserver/docker-mailserver-helm/issues/137

Line 10 enforces permissions of the `/var/log/mail` folder to be `syslog:root`. As described in the issue, in my helm-based setup this causes the following warning

```
/etc/cron.daily/logrotate:
error: skipping "/var/log/mail/mail.log" because parent directory has insecure permissions (It's world writable or writable by group which is not "root") Set "su" directive in config file to tell logrotate which user/group should be used for rotation.
error: skipping "/var/log/mail/rspamd.log" because parent directory has insecure permissions (It's world writable or writable by group which is not "root") Set "su" directive in config file to tell logrotate which user/group should be used for rotation.
```

The proposed change will implement what the warning suggests.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,7 @@ All notable changes to this project will be documented in this file. The format 
   - Fixed a regression introduced in DMS v14 where `postfix-main.cf` appended `stderr` output into `/etc/postfix/main.cf`, causing Postfix startup to fail ([#4147](https://github.com/docker-mailserver/docker-mailserver/pull/4147))
   - Fixed a regression introduced in DMS v14 to better support running `start-mailserver.sh` with container restarts, which now only skip calling `_setup()` ([#4323](https://github.com/docker-mailserver/docker-mailserver/pull/4323#issuecomment-2629559254))
   - The command `swaks --help` is now functional ([#4282](https://github.com/docker-mailserver/docker-mailserver/pull/4282))
+  - fix(log): su syslog root ([#4370](https://github.com/docker-mailserver/docker-mailserver/pull/4370))
 - **Rspamd:**
   - DKIM private key path checking is now performed only on paths that do not contain `$` ([#4201](https://github.com/docker-mailserver/docker-mailserver/pull/4201))
 
--- a/target/scripts/startup/setup.d/log.sh
+++ b/target/scripts/startup/setup.d/log.sh
@@ -33,6 +33,7 @@ function _setup_logrotate() {
   delaycompress
   rotate ${LOGROTATE_COUNT}
   ${LOGROTATE_INTERVAL}
+  su syslog root
 }
 EOF
 }


georglauterbach:
@casperklein @polarathene should we cram this into v15.0.0?

georglauterbach:
Please verify that we do not accidentally include a potential regression with this change right before the v15.0.0 release. I do not know enough about the permissions at the moment, hence I am not sure whether the actual permissions need a change or whether this PR's script change is accurate.

polarathene:
> I wonder, what the initial motivation for `syslog:root` was

Unclear, it was [committed to the Dockerfile in April 2016](https://github.com/docker-mailserver/docker-mailserver/commit/c63af947886574c1d95f91117e760bfd17d90d38), the only context is a reference to the [DMS v2 tracking issue](https://github.com/docker-mailserver/docker-mailserver/issues/109) but nothing in that issue thread or linked commit provides any context.

---

**EDIT:** Assumption is rsyslog runs as the `syslog` user (_unexpected if we explicitly need to create a `syslog` user for it to function though_), which if true answers the ownership requirement for `/var/log/mail` which is our own centralized log dir for DMS (_prior to introducing supervisor, but both are required atm AFAIK_). `root` group ownership is likely to infer that nothing else needs access via group permissions.

Additionally when deploying a container as rootless there is a common pattern to only change the runtime user while keeping a root group (`docker run --user 1000:0`), but that isn't too relevant here, definitely wasn't when this was implemented.

You could also use SGID bit IIRC to have any service with write permission to a directory to enforce files created are assigned the common parent directory group.

---

> It's neither world writeable nor writable by group which is not "root" 🤔

That'd depend on the context. `/var/log/mail` is a location DMS has volume mounted, which like `/var/mail` means a bind mount replaces all ownership/permissions assigned at the image (_hence why these runtime fixes came about, and why this type of thing belongs in ~~`mail-state.sh`~~ (`setup-stack.sh`, see below) as we'd skip these kind of fixes during container restarts otherwise_).

For `/var/mail` this removes the SGID bit that Debian otherwise has on the folder, I don't think we restore that since I identified it (_I have an open issue with all the details regarding this concern and related issues to resolve_).

As the linked bug report shows, their bind mount volume has modified permissions to `777`. That or it's a k8s / helm related config difference for how the volume mount is handled. Perhaps that's **why the bug report was raised at the helm repo** and not here?

## Related history / context

We do have this other related change in the `Dockerfile`:

https://github.com/docker-mailserver/docker-mailserver/blob/bcee78e2c1c310c4815b3e03c5e9278ad8ef1343/Dockerfile#L147-L148

That was introduced in [Jan 2018 during the base image swap](https://github.com/docker-mailserver/docker-mailserver/commit/49b3867c1b39e9832f460a344573c1673b964b1e) from Ubuntu 16.04 to Debian 9 Stretch ([PR](https://github.com/docker-mailserver/docker-mailserver/pull/784)):

> _`rsyslog` does not add `syslog` user and has different conf-structure_

While this runtime snippet:

https://github.com/docker-mailserver/docker-mailserver/blame/1a938dfb15e94f43f1af5ddaf7b8276aba92a825/target/scripts/startup/setup.d/log.sh#L3-L11

Originates from [Oct 2019](https://github.com/docker-mailserver/docker-mailserver/commit/37e0082cd7998924564c542cc24e5ac72c7ef6c2) (_associated [issue](https://github.com/docker-mailserver/docker-mailserver/issues/1300) + [PR](https://github.com/docker-mailserver/docker-mailserver/pull/1305)_). Basically belongs in ~~`mail-state.sh` AFAIK~~ (`setup-stack.sh`, see below).

The Oct 2019 issue was specifically about `/var/log/mail/clamav.log` failing, and expecting `clamav:clamav` permissions, although the PR/commit is `clamav:adm` for the `clamav.log` and `freshclam.log` files, along with `syslog:root` for `/var/log/mail`.

Not sure if we're actually leveraging the supervisor ClamAV logs for anything:

https://github.com/docker-mailserver/docker-mailserver/blob/1a938dfb15e94f43f1af5ddaf7b8276aba92a825/target/supervisor/conf.d/dms-services.conf#L72-L79

The two separate ClamAV logs are configured only via the `Dockerfile` at build (?), probably should be extracted out to a script 🤷‍♂️ (_even if that's only run during build-time_)

https://github.com/docker-mailserver/docker-mailserver/blob/1a938dfb15e94f43f1af5ddaf7b8276aba92a825/Dockerfile#L246-L259

Even with syslog, we have rsyslog run via supervisord. I haven't looked for config for this, presumably something has it run as the `syslog` user to enforce that requirement, `root` ownership is likely a side-effect of just not implying a need access by anything else via group:

https://github.com/docker-mailserver/docker-mailserver/blob/1a938dfb15e94f43f1af5ddaf7b8276aba92a825/target/supervisor/conf.d/dms-services.conf#L27-L34

---

## Probably should address for DMS v15

DMS v12 release [split the runtime fix for `/var/log/mail` + separate ClamAV log files](https://github.com/docker-mailserver/docker-mailserver/commit/8ec5dbec74839e0460252bcc2115035310f49327), so now we have as of DMS v15 (_unless changed before release_),

https://github.com/docker-mailserver/docker-mailserver/blob/bcee78e2c1c310c4815b3e03c5e9278ad8ef1343/target/scripts/startup/setup.d/security/misc.sh#L158-L163

https://github.com/docker-mailserver/docker-mailserver/blob/bcee78e2c1c310c4815b3e03c5e9278ad8ef1343/target/scripts/startup/setup.d/log.sh#L3-L11

Until we have a better location, the current location for such fixes is grouped into this method:

https://github.com/docker-mailserver/docker-mailserver/blob/bcee78e2c1c310c4815b3e03c5e9278ad8ef1343/target/scripts/startup/setup-stack.sh#L85-L104

DMS v15 correctly runs that even for container restarts so we could move these changes into that 👍 

---

## Will eventually improve

I've covered extensive info on [our current logging situation in the past](https://github.com/docker-mailserver/docker-mailserver/issues/3561) (_see the "What are you going to contribute?" section_)
- April 2016 (DMS v2) centralized services logs to live at `/var/log/mail` to pair with a volume mount for persistence.
- [August 2017 introduced Supervisord](https://github.com/docker-mailserver/docker-mailserver/pull/676), and over time more services were properly configured to manage logs through stdout to collect service logs at a different location, while remaining syslog output is collected in our central mail log that the container outputs to stdout (_for Docker's own log drivers to leverage_), and our tests rely on this mail log file.

That'll get tackled once I've found time to return to the Vector syslog PR. With Vector we can better manage ingesting log files, syslog, etc and output to centralized logs for individual services or other formats for structured logs, etc.

polarathene:
> should we cram this into v15.0.0?

I'm not sure if it makes sense to merge this PR, like @casperklein pointed out it doesn't look like this would be reproduced unless the user has modified permissions for their volume bind mount.

I don't know how likely that should be needed to support or expect. If this concern doesn't occur with our defaults, then we could just make the changes I mention in **"Probably should address for DMS v15"** section in my previous comment, which can add another line to reset permissions for `/var/mail/log` to the expected 755 (rwx rx rx)?

---

https://github.com/docker-mailserver/docker-mailserver-helm/issues/137#issuecomment-2440579271 mentions the helm / k8s config (_sorry not too familiar with that_) is a limitation for managing this, I'm not sure how compatible the fix I propose would be with the helm project, if someone can confirm it's compatible that'd be great 👍

casperklein:
> EDIT: Assumption is rsyslog runs as the syslog user

As you stated, we create a 'syslog' user ourself. It's afaik not needed normally (otherwise the setup routine of rsyslog would create that user). The fact, that rsyslog can creates files as other users implies, that it (must) runs as root?

casperklein:
> EDIT: Assumption is rsyslog runs as the syslog user

rsyslogd runs as root:

```
root@mail:/# ps aux|grep rsyslogd
root      2332  0.0  0.0 152120  2936 ?        Sl   Feb16   0:01 /usr/sbin/rsyslogd -n
```

polarathene:
**TL;DR:** Ok I think I got the gist of it now 💪 
  - `syslog` user and related `chown` appears redundant, should be safe to remove 👍 
  - ClamAV related commands also should be fine to remove, given the fix at image build to ensure a fixed UID/GID (_avoiding mapping mismatch when order of packages installed changes the incremental ID that'd otherwise have been assigned_).

Am I confident in avoiding a breaking change? **No**. Technically we should also ensure ownership is changed away from `syslog` once removing that too, not that it should matter in practice. I'd need more time to go over everything I've documented here and verify, so perhaps best to delay that.

Earlier suggestion as an **interim fix is still valid for DMS v15**.

---

**Summary:**
- `syslog` ownership appears to be inherited from the Ubuntu to Debian base image switch.
  - This seems redundant and not actually required anymore. It's inclusion at runtime was with resolving the ClamAV ownership errors and not questioned/explained during review.
  - Base image upgrade from Ubuntu to Debian adding the user `syslog` as a fix would only have been to support the later `chown` command from DMS v2, not questioned during review.
  - DMS v2 would have likely added the `chown` for `syslog` for similar reason as explained below, a compatibility fix for volume mounts where the friendly user/group names map to different UID/GID between distro. Ubuntu potentially had rsyslog configured to use the `syslog` user, unlike with Debian.
- Ownership fixes were for ClamAV from what I can tell shouldn't have been needed
  - Likely occurred due to volume mount and an upgrade as this would have been before I introduced a fixed UID/GID for clamav, hence a mismatch during image upgrade.
  - ClamAV configs have the user `clamav` configured, and only a `clamav` group for the ClamAV socket. Thus correcting the owner of the log files at container startup was a fix that should no longer be necessary.

---

> rsyslogd runs as root

It may still be configured somewhere to write as syslog?

Initial `/var/log/mail/mail.log` setup (_`Dockerfile` adjusts `rsyslog` config to write syslog mail filtered logs to this file, our test suite checks this too_):

https://github.com/docker-mailserver/docker-mailserver/blob/02942947559d3c0048bb3ebe62ccd1ec2272758a/target/scripts/start-mailserver.sh#L193-L205

Symlink created to `/var/log/mail.log`:

https://github.com/docker-mailserver/docker-mailserver/blob/02942947559d3c0048bb3ebe62ccd1ec2272758a/Dockerfile#L170-L179

---

## rsyslog package

Installing the package in `debian:12-slim` image creates `/etc/rsyslog.conf` with these config snippets:

```
###########################
#### GLOBAL DIRECTIVES ####
###########################

#
# Set the default permissions for all log files.
#
$FileOwner root
$FileGroup adm
$FileCreateMode 0640
$DirCreateMode 0755
$Umask 0022

#
# Where to place spool and state files
#
$WorkDirectory /var/spool/rsyslog

#
# Include all config files in /etc/rsyslog.d/
#
$IncludeConfig /etc/rsyslog.d/*.conf
```

```
###############
#### RULES ####
###############

#
# Log anything besides private authentication messages to a single log file
#
*.*;auth,authpriv.none          -/var/log/syslog

#
# Log commonly used facilities to their own log file
#
auth,authpriv.*                 /var/log/auth.log
cron.*                          -/var/log/cron.log
kern.*                          -/var/log/kern.log
mail.*                          -/var/log/mail.log
user.*                          -/var/log/user.log

#
# Emergencies are sent to everybody logged in.
#
*.emerg                         :omusrmsg:*
```

```
#################
#### MODULES ####
#################

module(load="imuxsock") # provides support for local system logging
module(load="imklog")   # provides kernel logging support
#module(load="immark")  # provides --MARK-- message capability

# provides UDP syslog reception
#module(load="imudp")
#input(type="imudp" port="514")

# provides TCP syslog reception
#module(load="imtcp")
#input(type="imtcp" port="514")
```

DMS `Dockerfile` changes to that config:

https://github.com/docker-mailserver/docker-mailserver/blob/02942947559d3c0048bb3ebe62ccd1ec2272758a/Dockerfile#L240-L241

https://github.com/docker-mailserver/docker-mailserver/blob/02942947559d3c0048bb3ebe62ccd1ec2272758a/Dockerfile#L254-L255

https://github.com/docker-mailserver/docker-mailserver/blob/02942947559d3c0048bb3ebe62ccd1ec2272758a/Dockerfile#L234-L235

## Observations

- [`mail-state.sh` doesn't have any handling](https://github.com/docker-mailserver/docker-mailserver/blob/02942947559d3c0048bb3ebe62ccd1ec2272758a/target/scripts/startup/setup.d/mail_state.sh) for persisting rsyslog state (`$WorkDirectory /var/spool/rsyslog`)
- We have `rsyslog` run as `root` via supervisord service, and the original rsyslog config `/etc/rsyslog.conf` shown above runs the process as `root:adm` with permissions for files/dirs configured fine, while `/etc/rsyslog.d/` is empty.
- We modify rsyslog config to:
  - route syslog entries received marked with the `mail` facility **from `/var/log/mail.log`** (_DMS makes this a symlink instead_) **to** `/var/log/mail/mail.log`
  - Alter the filter for `/var/log/syslog` to prepend `mail.none;mail.error` (_I forget this syntax but I've documented it somewhere in our issues_ 😅)
  - Disable the kernel logging since we're running in a container, not a host/system syslog process where that'd be relevant.
  -

polarathene:
I have applied the interim fix I suggested: https://github.com/docker-mailserver/docker-mailserver/pull/4374

I do not know if that'll resolve the issue with k8s / helm but I don't think the proposed solution from this PR is the right fix? (_especially since we'd be going back to `root` ownership in future_)

PR #4375, opened 2025-02-17
refactor: setup CLI `open-dkim`

polarathene:
# Description

Up to you if you'd like to include this in DMS `15.0` or DMS `15.1`. As this is only a change for a setup CLI command, it's not something I'd consider breaking, and functionality wise it should be roughly the same. Tests also seem to indicate all is well 👍 

---

I recently did some refactoring on this command while going over unifying our DKIM feature support.
- Adds a `main()` method similar to other setup CLI commands.
- Help text more aligned with equivalent rspamd DKIM setup CLI command.
- DRY some repetition such as hard-coded paths to use variables.
- OpenDKIM config files are created / initialized early on now with `_create_opendkim_configs()`. `while` loop only needs to append entries, so is easier to grok.
- `_create_dkim_key()` to scope just the logic (_and additional notes_) to key generation via `opendkim-genkey`
- Now overall logic with the `while` loop of the script occurs in `_generate_dkim_keys()`:
  - Ownership fixes are now applied after the `while` loop as that seems more appropriate than per iteration.
  - Temporary VHOST config is now removed since it's no longer useful after running.
- Tests adjusted for one new log for adding of default trusted hosts content.

Overall this should be nicer to grok/maintain. Some of this logic will be reused for the unified DKIM generation command in future, which is more likely to shift towards all domains using the same keypair by default with rspamd/opendkim config generated at runtime rather than reliant upon DMS config volume to provide that (_still expected for private key_).

## Type of change

## Checklist

### DIFF
--- a/target/bin/open-dkim
+++ b/target/bin/open-dkim
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # shellcheck source=../scripts/helpers/index.sh
 source /usr/local/bin/helpers/index.sh
@@ -12,9 +12,17 @@ if [[ -f /etc/dms-settings ]] && [[ $(_get_dms_env_value 'ENABLE_RSPAMD') -eq 1 
   fi
 fi
 
-KEYSIZE=2048
-SELECTOR=mail
-DOMAINS=
+function _main() {
+  # Default parameters (updated by `_parse_arguments()`):
+  local KEYSIZE=2048
+  local SELECTOR=mail
+  local DMS_DOMAINS=
+
+  _require_n_parameters_or_print_usage 0 "${@}"
+  _parse_arguments "${@}"
+
+  _generate_dkim_keys
+}
 
 function __usage() {
   printf '%s' "${PURPLE}OPEN-DKIM${RED}(${YELLOW}8${RED})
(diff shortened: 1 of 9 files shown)


georglauterbach:
I'd rather not have this in v15.0.0. I marked it appropriately.

georglauterbach:
I'll take a look at this soon

target/bin/open-dkim
```diff
@@ -57,122 +65,171 @@ ${ORANGE}EXAMPLES${RESET}
...
 DATABASE_VHOST='/tmp/vhost.dkim'
-# Prepare a file with one domain per line:
 function _generate_domains_config() {
   local TMP_VHOST='/tmp/vhost.dkim.tmp'
 
   # Generate the default vhost (equivalent to /etc/postfix/vhost),
-  # unless CLI arg DOMAINS provided an alternative list to use instead:
-  if [[ -z ${DOMAINS} ]]; then
+  # unless CLI arg DMS_DOMAINS provided an alternative list to use instead:
+  if [[ -z ${DMS_DOMAINS} ]]; then
```
casperklein, line 167:
```suggestion
  if [[ -z ${DMS_DOMAINS:-} ]]; then
```

casperklein (approved):
Code reviewed, but not tested. LGTM 👍

target/bin/open-dkim
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
polarathene: Just habit, originally it was `/usr/bin/env sh`, until I realized that while that usually points to `bash` or `ash` on most distros, Debian symlinks `sh` to `dash` which is even more limited than `ash` IIRC 😓 

Open to reverting 👍

target/bin/open-dkim
```diff
@@ -57,122 +65,171 @@ ${ORANGE}EXAMPLES${RESET}
...
+        ;;
+
+      ( 'help' )
+          __usage
+          exit 0
+          ;;
+
+      ( * )
+        __usage
+        _exit_with_error "Unknown option(s) '${1}' ${2:+"and '${2}'"}"
```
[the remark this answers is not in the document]
polarathene: Similar to other review feedback, I think I was just making this `open-dkim` script to be more consistent with the rspamd one before a future refactor to be DKIM tooling agnostic via `step` CLI command.

target/bin/open-dkim
polarathene: For my benefit, what exactly is `:-` doing here without a fallback value? Not the friendliest thing for me to look up online [mostly I get examples of it with an actual value](https://devhints.io/bash).

I'm assuming it's related to an interest with that setting to fail if a variable is unbound? (_`set -u` or something like that_)

This would be clearer with a more modern alternative: https://www.nushell.sh/book/environment.html#reading-environment-variables

```nu
# Equivalent for `if [[ -z ${FOO:-} ]]; then ... fi`

# `-z` => `is-empty`
# `-n` => `is-not-empty`
if ( $env.FOO? | is-empty ) { ... }
```

```nu
# Is the ENV set (NOTE: doesn't account for empty value):
if "FOO" in $env {
  echo $env.FOO
}

# Fails by default when invalid:
$env.FOO | describe
# => Error: nu::shell::column_not_found

# `?` allows for unset/null values, avoiding failure (describe reports type "nothing"):
# Use `?` when no separate error handling needs to be handled via try/catch on errors.
$env.FOO? | describe
# => nothing

# Return a default value when ENV isn't set:
$env.FOO? | default "BAR"
# => BAR
```

Nushell in my experiments so far isn't always nice, but there's plenty of times it's friendlier to grok than bash syntax when you're not using arcane syntax on the regular 😝

casperklein: > For my benefit, what exactly is :- doing here without a fallback value?

> I'm assuming it's related to an interest with that setting to fail if a variable is unbound? (set -u or something like that)

That's what I had in mind. Just best practice, to ensure, we are not using an unbound var + streamline with our other code.

target/bin/open-dkim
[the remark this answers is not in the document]
casperklein: I prefer a clear `/bin/bash` 👍 But there’s no blocker from my side. That applies to all my suggestions.

target/bin/open-dkim
```diff
@@ -57,122 +65,171 @@ ${ORANGE}EXAMPLES${RESET}
...
-
-  esac
-done
+function _parse_arguments() {
+  # Parse the command args through iteration:
+  while [[ ${#} -gt 0 ]]; do
+    case "${1}" in
+
+      ( 'keysize' )
+        if [[ -n ${2+set} ]]; then
```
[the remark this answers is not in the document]
casperklein: Yes, they are functionally equivalent.

>  do you want the same done to the rspamd script then?

At the best, all of our scripts use more or less the same styles/approaches. No rush to adjust everything right away - this was more of a 'while we're at it' note.

target/bin/open-dkim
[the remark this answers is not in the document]
casperklein: I understand that 😄 But while we at it, this could be fixed. The old variant makes no sense (to me).

target/bin/open-dkim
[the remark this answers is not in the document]
georglauterbach: We use `/bin/bash` in all the other scripts too, hence I agree with @casperklein here.

target/bin/open-dkim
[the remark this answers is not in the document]
georglauterbach: While I like `+set` better because it carries the intent, I am fine with using `:-`.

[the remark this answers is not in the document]
polarathene: > because it carries the intent

That is still arcane magic syntax to me compared to nushell 😝 (_which also has `match` btw_):

```nu
if ( $keysize? | is-not-empty ) { ... }
```

polarathene:
Bundled in the `+set` => `:-` suggestion to other files unrelated to the focus of this PR (😬) at https://github.com/docker-mailserver/docker-mailserver/pull/4375/commits/7eed5d1db4fe0abca166e3ab2f57576fd3e01075

test/helper/setup.bash
```diff
@@ -17,7 +17,7 @@
 # This function is internal and should not be used in tests.
 function __initialize_variables() {
   function __check_if_set() {
-    if [[ ${!1+set} != 'set' ]]; then
+    if [[ -z ${1:-} ]]; then
```
casperklein, line 20:
```suggestion
    if [[ -z ${!1:-} ]]; then
```

Without the exclamation mark, you just check, if $1 is not empty.

${!1} is indirect expansion, meaning it retrieves the value of the variable whose name is stored in $1

casperklein (changes requested):

test/helper/setup.bash
polarathene: > ${!1} is indirect expansion, meaning it retrieves the value of the variable whose name is stored in $1





I figured it was something I was going to trip up on 😂 Thanks for spotting that! 🙏 

This is the sorta thing that adds friction to attracting more third-party contributions though I think? 🤔

casperklein: > This is the sorta thing that adds friction to attracting more third-party contributions though I think? 🤔

Or it encourage people, to learn something new (in bash). I learned a lot while working on DMS 🚀

casperklein (approved):

test/helper/setup.bash
polarathene: I learned plenty too, but given the syntax `!` / `:-`, they're not exactly search friendly for a newbie that doesn't know what related keywords to search up.

If we were to swap bash for something else, you've previously expressed reluctance due to the fact you'd have to learn something new... so I don't think that's a particular fair stance 😝 

The nushell examples I've shared several snippets of across various discussions is often far more friendly/familiar (_or obvious in syntax alone_) with equivalent features.

---

The more friction there is to contribute, the less likely it'll happen. It's often a concern that'll discourage me from contributing to projects for example as I've got enough on my plate to suss out already 😂

georglauterbach: I think we had this discussion already, but it does not lead anywhere. I am currently not willing to switch to NuShell. This is not because I think NuShell is inferior (on the contrary, I think it's probably better). But because no one has the time to rewrite this project not can we maintain it efficiently afterward, etc.

Given we went over this with Rust - which I deem to be superior to NuShell for DMS as a whole - we do not need to extend it here.

polarathene: > But because no one has the time to rewrite this project not can we maintain it efficiently afterward, etc.

If someone wants to port DMS and maintainers were open to it, that's up to the contributor to handle the rewrite. Bigger issue isn't that contributor time it's the review process, one thing to support a port, another to confidently review something unfamiliar 👍 

After that point, I don't think it'd have a negative impact on maintenance tbh.

> Given we went over this with Rust - which I deem to be superior to NuShell for DMS as a whole - we do not need to extend it here.

One of the issues with Rust was that it was compile time, rather than the flexibility of an interpreter running arbitrary scripts like bash and python. Nushell was identified as alternative that had perks of rust, but more similar to bash for development without the issues raised with python 😅 

Don't mind me though, just an occasional rant about gripes with bash. End of the day there's still two maintainers that are quite comfortable with bash and I am very familiar with our past discussions on the topic, I am just concerned about the bus factor there vs reducing maintenance complexity in a way that more contributors could easily understand what they're reading.

georglauterbach: > > Given we went over this with Rust - which I deem to be superior to NuShell for DMS as a whole - we do not need to extend it here.
>
> One of the issues with Rust was that it was compile time, rather than the flexibility of an interpreter running arbitrary scripts like bash and python. Nushell was identified as alternative that had perks of rust, but more similar to bash for development without the issues raised with python 😅

Agreed, that is indeed an upside! I am sure we can find many upsides, and we could even compare all approaches. But I, too, think that no one will rewrite this any time soon. I am still open, if anyone wants to try :D

georglauterbach (dismissed):
LGTM 👍🏼 

Needs a rebase, but other than that - nice!

georglauterbach (approved):

PR #4376, opened 2025-02-17
chore: Revise utility install scripts + add Smallstep `step` CLI

polarathene:
# Description

`step` isn't used for anything yet, so no rush for this PR, but it won't introduce any breakage into DMS v15 😅 

---

Changes:
- `jaq` should probably live in `/usr/local/bin` with other third-party sourced binaries.
- `swaks` install properly with just `tar`, no `mv` + `rm` needed.
- Added Smallstep `step` CLI. This serves similar purpose to `openssl` commands, but is generally nicer for usage with generation and inspection of certs/keys. I've talked up using in DMS a few times in the past for our TLS helper and unifying DKIM support (_instead of separate OpenDKIM/Rspamd generators_).
- Including `step` for both AMD64 / ARM64 archs needs the alternate naming convention that it's published to GH releases with.
- Added commentary about the `tar` usage. The ownership is a common one, technically a non-issue when running as `root` 🤷‍♂️ Let me know if it's overkill and you prefer the `mv` + `rm` approach with a additional `chown`.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,12 +6,22 @@ All notable changes to this project will be documented in this file. The format 
 
 > **Note**: Changes and additions listed here are contained in the `:edge` image tag. These changes may not be as stable as released changes.
 
+### Added
+
+- **Internal:**
+  - Added the Smallstep `step` CLI command for future internal usage ([#4376](https://github.com/docker-mailserver/docker-mailserver/pull/4376))
+
 ### Fixes
 
 - **Postfix:**
   - `setup email restrict` generated configs now only prepend to `dms_smtpd_sender_restrictions` ([#4379](https://github.com/docker-mailserver/docker-mailserver/pull/4379))
 - **Internal:**
   - A permissions fix for `/var/log/mail` that was [added in DMS v15]((https://github.com/docker-mailserver/docker-mailserver/pull/4374)) no longer encounters an error when no log files are present during a container restart, such as with a `tmpfs` volume mount ([#4391](https://github.com/docker-mailserver/docker-mailserver/pull/4391))
+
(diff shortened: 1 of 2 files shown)


georglauterbach:
> step isn't used for anything yet

No need to cram this into v15.0.0.

polarathene:
> I also lack background for this change, it seems. What's the rationale here?

For `step`? I explained this in the "changes" list regarding it.

I think it's a more user friendly `openssl` alternative that we can use for both TLS/X.509 certs and [public-key cryptography](https://en.wikipedia.org/wiki/Public-key_cryptography) (DKIM/ARC). It'd be useful for docs, troubleshooting advice (_nicer output vs openssl, can output results as JSON too_), self-signed cert and key generation.

As such it can replace the separate opendkim/rspamd DKIM key generators. Your rspamd script for example appears to start rspamd in the background just to generate the keys.

We could technically not add `step` into the image, but given the image already weighs over 700MB, 13MB isn't a big ask for the convenience to users and our docs.

---

Or were you asking about the other changes unrelated to `step`? Those should also be explained under the "changes" list I provided?

georglauterbach:
Sorry, I was blind I guess... I somehow missed it in the PR description.

I really like these modern tools! And I also appreciate this replacing the horrible `openssl` command ❤️👍🏼

casperklein:
LGTM 👍 If there are more tools in the future, we should put them in own function.

casperklein (approved):

georglauterbach:
IMO we can put this into v15.0.1 too @polarathene if you update `CHANGELOG.md`.

georglauterbach (approved):

PR #4391, opened 2025-03-02
fix: Support `chmod` on `/var/log/mail/*` when dir is empty

mazzz1y:
# Description

To prevent SSD wear from logs, I mount `/var/log/mail/` to tmpfs in my setup. I collect logs from standard output using Loki.

As a result, the `/var/log/mail` directory is empty whenever the container restarts. Attempting to change permissions on this directory using shell expansion leads to errors like:
```
chmod: missing operand after ‘640’
Try 'chmod --help' for more information.
```

Using `find` avoids running `chmod` if there is nothing to modify, preventing these errors. It's also more convenient than shell expansion, which can cause errors like "too many arguments".

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ All notable changes to this project will be documented in this file. The format 
 
 > **Note**: Changes and additions listed here are contained in the `:edge` image tag. These changes may not be as stable as released changes.
 
+### Updates
+
+**Internal:**
+  - Improved permission updating command to prevent errors when no logs are present ([#4391](https://github.com/docker-mailserver/docker-mailserver/pull/4391))
+
 ## [v15.0.0](https://github.com/docker-mailserver/docker-mailserver/releases/tag/v15.0.0)
 
 ### Breaking
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,12 @@ All notable changes to this project will be documented in this file. The format 
 
 > **Note**: Changes and additions listed here are contained in the `:edge` image tag. These changes may not be as stable as released changes.
 
-### Updates
+### Fixes
 
-**Internal:**
-  - Improved permission updating command to prevent errors when no logs are present ([#4391](https://github.com/docker-mailserver/docker-mailserver/pull/4391))
+- **Postfix:**
+  - `setup email restrict` generated configs now only prepend to `dms_smtpd_sender_restrictions` ([#4379](https://github.com/docker-mailserver/docker-mailserver/pull/4379))
+- **Internal:**
+  - A permissions fix for `/var/log/mail` that was [added in DMS v15]((https://github.com/docker-mailserver/docker-mailserver/pull/4374)) no longer encounters an error when no log files are present during a container restart, such as with a `tmpfs` volume mount ([#4391](https://github.com/docker-mailserver/docker-mailserver/pull/4391))
 
 ## [v15.0.0](https://github.com/docker-mailserver/docker-mailserver/releases/tag/v15.0.0)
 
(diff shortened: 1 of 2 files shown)


casperklein:
LGTM.

However, I am wondering about the error message. Since we don't use `shopt -s nullglob`, `/var/log/mail/*`  should be "expand" to `/var/log/mail/*` if there are no files in that directory:

```
# mkdir empty
# chmod 640 empty/*
chmod: cannot access 'empty/*': No such file or directory

# shopt -s nullglob
# chmod 640 empty/*
chmod: missing operand after ‘640’
Try 'chmod --help' for more information.
```

polarathene:
> Since we don't use `shopt -s nullglob`

We use it a [few times in scripts](https://github.com/search?q=repo%3Adocker-mailserver%2Fdocker-mailserver%20nullglob&type=code). The setup call for getmail during initial container setup does set it which might continue to affect usage afterwards that follows outside the function? (_I haven't verified, just a guess_):

https://github.com/docker-mailserver/docker-mailserver/blob/5686a4097ae5719d41afda33d3523d47b5a36b4d/target/scripts/startup/setup.d/getmail.sh#L23-L25

casperklein:
Thanks for pointing that out. You are right, I've overlooked that.

casperklein (dismissed):

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ All notable changes to this project will be documented in this file. The format
 
 > **Note**: Changes and additions listed here are contained in the `:edge` image tag. These changes may not be as stable as released changes.
 
+### Updates
+
+**Internal:**
+  - Improved permission updating command to prevent errors when no logs are present ([#4391](https://github.com/docker-mailserver/docker-mailserver/pull/4391))
```
polarathene, line 12: Revised the changelog entry and added another one that was merged without the log since this probably warrants pushing a patch release.

```suggestion
### Fixes

- **Postfix:**
  - `setup email restrict` generated configs now only prepend to `dms_smtpd_sender_restrictions` ([#4379](https://github.com/docker-mailserver/docker-mailserver/pull/4379))
- **Internal:**
  - A permissions fix for `/var/log/mail` that was [added in DMS v15]((https://github.com/docker-mailserver/docker-mailserver/pull/4374)) no longer encounters an error when no log files are present during a container restart, such as with a `tmpfs` volume mount ([#4391](https://github.com/docker-mailserver/docker-mailserver/pull/4391))
```

polarathene (changes requested):

polarathene (approved):
Good to merge.

Might wait a day or two for any other v15 bug reports, then push this out as a bug fix release as `v15.0.1`? @casperklein @georglauterbach 

I'm reluctant to delay until a 15.1 given what our release cadence typically is like.

georglauterbach:
I'll review this today, please do not merge yet.

polarathene:
@georglauterbach the fix seems fine, it properly only applies to files, not that I think they'd be any directories in `/var/log/mail/` itself.

The directory would be empty when given a fresh volume mount at container startup, or like the author during container restarts with `tmpfs` as well. Hence the error.

`shopt -s nullglob` should only accidentally be left enabled when the Getmail feature is enabled AFAIK. I assume that is a separate issue that should undo that to avoid surprises?

georglauterbach:
I just wanted to verify that there are not too many files, and no directories inside `/var/log/mail`. LGMT 👍🏼 

The `nullglob` can be handled separately, but I do not see it as an issue.

casperklein:
> The nullglob can be handled separately, but I do not see it as an issue, really.

It's an issue, but not a big one. OPs error message would be different, if getmail was not enabled. See my first post.

I think that should be streamlined. We could move the nullglob command to the top of start-mailserver.sh, making it consequently available to all code that is processed by `start-mailserver.sh`.

PR #4420, opened 2025-03-17
fix: DMS state volume must ensure `o+x` permission

### LINKED ISSUE #4419: 'INBOX': Timeout (180s) while waiting for lock for transaction log file
### Subject

### Description

Hi,

We are using docker mail server (version 14.0). After sometime, we are not able to receive emails for some of our mail boxes and seeing following logs.

Timeout (180s) while waiting for lock for transaction log file
failed to store into mailbox 'INBOX': Timeout (180s) while waiting for lock for transaction log file 

Any insights on this, how we can fix this issue ?

polarathene:
# Description

While we expect the volume to have `755` as the default for `/var/mail-state`, some storage services will differ with their default permissions.

Service users that need to access their state directory cannot do so when a parent directory has `root` ownership only and lacks the executable bit for the `other` permission set. This is the source of numerous bug reports in the past that was non-obvious to me for how to resolve due to the errors logged 😓 (_big thanks to @rixwan-sharif for identifying the correct fix_)

More details: https://github.com/docker-mailserver/docker-mailserver/issues/4419#issuecomment-2731196821

Fixes #4419

## Type of change

## Checklist

### DIFF
--- a/target/scripts/startup/setup.d/mail_state.sh
+++ b/target/scripts/startup/setup.d/mail_state.sh
@@ -95,6 +95,11 @@ function _setup_save_states() {
 function _setup_adjust_state_permissions() {
   [[ ! -d ${DMS_STATE_DIR} ]] && return 0
 
+  # Parent directories must have executable bit set to descend the file tree for access,
+  # as each service running as a non-root user requires this to access their state directory,
+  # `/var/mail-state` must allow all users `+x`:
+  chmod o+x "${DMS_STATE_DIR}"
+
   # This ensures the user and group of the files from the external mount have their
   # numeric ID values in sync. New releases where the installed packages order changes
   # can change the values in the Docker image, causing an ownership mismatch.


target/scripts/startup/setup.d/mail_state.sh
```diff
@@ -95,6 +95,11 @@ function _setup_save_states() {
 function _setup_adjust_state_permissions() {
   [[ ! -d ${DMS_STATE_DIR} ]] && return 0
 
+  # Parent directories must have executable bit set to descend the file tree for access,
+  # as each service running as a non-root user requires this to access their state directory,
+  # `/var/mail-state` must allow all users `+x`:
+  chmod o+x "${DMS_STATE_DIR}"
```
polarathene, line 101: Wondering if this should be `ugo+x`?

```suggestion
  chmod ugo+x "${DMS_STATE_DIR}"
```

rixwan-sharif: So ugo+x is just like `chmod +x` and I think should fine because the permissions inside /var/mail-state is still open than this 

```shell
drwxr-x--x  8 amavis  amavis  4096 Mar 18 04:36 lib-amavis
drwxr-xr-x  2 docker  docker  4096 Mar 18 04:56 lib-dovecot
drwxr-xr-x  2 root    root    4096 Dec 14  2022 lib-logrotate
drwxr-xr-x  2 postfix postfix 4096 Mar 17 22:09 lib-postfix
drwxr-xr-x 20 docker  docker  4096 Mar 17 22:00 qa.auto1.team
drwxr-xr-x 16 root    root    4096 Mar 18 04:56 spool-postfix
```

georglauterbach: I too think `+x` should be fine as it's the same as `ugo+x`.

casperklein (approved):

target/scripts/startup/setup.d/mail_state.sh
casperklein: I didn't know that either 🤯

georglauterbach (approved):

PR #4120, opened 2024-07-18
chore: Prevent Microsoft MUAs from sending reactions

mpldr:
# Description

Microsoft added a way to "react" to emails which is turned into
> [Image] <Name> liked this message.

No MUA except Microsoft's latest ones displays these properly and they are not providing significant value to an Email-exchange.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Improvement (non-breaking change that does improve existing functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [ ] If necessary, I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [x] **I have added information about changes made in this PR to `CHANGELOG.md`**

Issue #4263, opened 2024-11-14
bug report:  Outlook Attachment Icon


alessiodeia:
### 📝 Preliminary Checks

- [X] I tried searching for an existing issue and followed the [debugging docs](https://docker-mailserver.github.io/docker-mailserver/latest/config/debugging/) advice, but still need assistance.

### 👀 What Happened?

I installed DMS version 14.0.0 using Docker Compose. Everything works fine, but I've discovered a potential bug or compatibility issue with Outlook 2016/2021. When I send an email, it is saved in the "Sent" folder and is marked as having an attachment.

I traced the cause of the attachment icon appearing to my signature, which includes an embedded image. I attempted to modify the HTML of the signature to include an img tag that loads a remote URL, but the icon still appears. However, when I move the email to a local PST file or another IMAP server, the attachment icon disappears.

What steps can I take to resolve this issue?

### 👟 Reproduction Steps

- Install DMS and create a mailbox
- Add mailbox to Outlook 2016/2021 via IMAP config
- Setup a signature with an embedded image
- Try to send an email with just the signature
- Optional: try to move the Sent email inside another IMAP server or local PST folder

### 🐋 DMS Version

v14.0.0

### 💻 Operating System and Architecture

Rocky Linux 9.4 (Blue Onyx)

### ⚙️ Container configuration files

```yml
services:
  mailserver:
    image: ghcr.io/docker-mailserver/docker-mailserver:14.0.0
    container_name: mailserver
    hostname: example.com
    env_file: mailserver.env
    ports:
      - "25:25"    # SMTP  (explicit TLS => STARTTLS, Authentication is DISABLED => use port 465/587 instead)
      - "143:143"  # IMAP4 (explicit TLS => STARTTLS)
      - "465:465"  # ESMTP (implicit TLS)
      - "587:587"  # ESMTP (explicit TLS => STARTTLS)
      - "993:993"  # IMAP4 (implicit TLS)
    volumes:
      - /root/docker/mailserver/mail-data/:/var/mail/
      - /root/docker/mailserver/mail-state/:/var/mail-state/
      - /root/docker/mailserver/mail-logs/:/var/log/mail/
      - /root/docker/mailserver/config/:/tmp/docker-mailserver/
      - /root/docker/mailserver/config/15-mailboxes.conf:/etc/dovecot/conf.d/15-mailboxes.conf:ro
[...]

Issue #4224, opened 2024-10-11
bug report: List-Unsubscribe not working


sccsspnk:
### 📝 Preliminary Checks

- [X] I tried searching for an existing issue and followed the [debugging docs](https://docker-mailserver.github.io/docker-mailserver/latest/config/debugging/) advice, but still need assistance.

### 👀 What Happened?

![image](https://github.com/user-attachments/assets/0c0986aa-51f2-4d1a-90a3-12166515bba1)

### 👟 Reproduction Steps

I can't set `List-Unsubscribe`
See below

### 🐋 DMS Version

v14.0.0

### 💻 Operating System and Architecture

Ubuntu 24.04 LTS

### ⚙️ Container configuration files

```yml
services:
  mailserver:
    image: ghcr.io/docker-mailserver/docker-mailserver:latest
    container_name: mailserver-prod
    hostname: mail.example.com
    env_file: mailserver.env
    ports:
      - "25:25"
      - "143:143"
      - "465:465"
      - "587:587"
      - "993:993"
    volumes:
      - ./docker-data/dms/mail-data/:/var/mail/
      - ./docker-data/dms/mail-state/:/var/mail-state/
      - ./docker-data/dms/mail-logs/:/var/log/mail/
      - ./docker-data/dms/config/:/tmp/docker-mailserver/
      - ./docker-data/dms/config/postfix/list_unsub_header:/etc/postfix/list_unsub_header:ro
      - /etc/letsencrypt:/etc/letsencrypt/
      - /etc/localtime:/etc/localtime:ro
    restart: always
    stop_grace_period: 1m
    healthcheck:
      test: "ss --listening --tcp | grep -P 'LISTEN.+:smtp' || exit 1"
      timeout: 3s
      retries: 0
```

<details>
<summary>Click to view mailserver.env</summary>

```env
OVERRIDE_HOSTNAME=mail.example.com
DMS_DEBUG=1
LOG_LEVEL=debug
SUPERVISOR_LOGLEVEL=
DMS_VMAIL_UID=
DMS_VMAIL_GID=
ACCOUNT_PROVISIONER=
POSTMASTER_ADDRESS=postmaster@example.com
ENABLE_UPDATE_CHECK=1
UPDATE_CHECK_INTERVAL=1d
PERMIT_DOCKER=none
TZ=Europe/Moscow
NETWORK_INTERFACE=
TLS_LEVEL=
SPOOF_PROTECTION=
ENABLE_SRS=0
[...]

PR #4429, opened 2025-03-24
fix: ensure message content is not modified by header filter

mpldr:
Due to an oversight, one of the header filters has potentially modified the message body, leading to broken cryptographic signatures.

Switch to the Message-Id instead of the Content-Type header, which is usually only present in the global header.

Reported-by: Cole Young <cole@young.sh>
Fixes: 009237cc ("chore: Prevent Microsoft MUAs from sending reactions (#4120)")

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ All notable changes to this project will be documented in this file. The format 
 
 > **Note**: Changes and additions listed here are contained in the `:edge` image tag. These changes may not be as stable as released changes.
 
+### Fixes
+
+- **Postfix**
+  - Avoid modifying the message body when filtering sender headers. This regression was introduced from [#4120](https://github.com/docker-mailserver/docker-mailserver/pull/4120) as part of DMS v15.0.0 ([#4429](https://github.com/docker-mailserver/docker-mailserver/pull/4429))
+
 ## [v15.0.1](https://github.com/docker-mailserver/docker-mailserver/releases/tag/v15.0.1)
 
 ### Added
--- a/target/postfix/sender_header_filter.pcre
+++ b/target/postfix/sender_header_filter.pcre
@@ -8,4 +8,5 @@
 /^\s*X-Mailer/                      IGNORE
 /^\s*X-Originating-IP/              IGNORE
 /^\s*Received: from.*127.0.0.1/     IGNORE
-/^Content-Type:/i                   PREPEND X-MS-Reactions: disallow
+/^\s*X-MS-Reactions:/               IGNORE
+/^\s*Message-Id:/i                  PREPEND X-MS-Reactions: disallow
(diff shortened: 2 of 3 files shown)


georglauterbach (approved):

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ All notable changes to this project will be documented in this file. The format
 
 > **Note**: Changes and additions listed here are contained in the `:edge` image tag. These changes may not be as stable as released changes.
 
+### Fixes
+
+- **Postfix**
+  - ensure message content is not modified by header filter ([#4425](https://github.com/docker-mailserver/docker-mailserver/pull/4425))
```
polarathene, line 12: The PR referenced needs to be updated, I've revised the changelog entry 

- [SMTP `DATA`](https://learn.microsoft.com/en-us/previous-versions/office/developer/exchange-server-2010/aa493918(v=exchg.140)) ([RFC 822](https://datatracker.ietf.org/doc/html/rfc822)) - Message headers + body.
- [MIME](https://learn.microsoft.com/en-us/previous-versions/office/developer/exchange-server-2010/aa563350(v=exchg.140)) ([RFC 1341](https://datatracker.ietf.org/doc/html/rfc1341)) - Multi-part body each with their own `Content-Type` headers ([example](https://learn.microsoft.com/en-us/previous-versions/office/developer/exchange-server-2010/aa563375(v=exchg.140))).

```suggestion
- **Postfix**
  - Avoid modifying the message body when filtering sender headers. This regression was introduced from [#4120](https://github.com/docker-mailserver/docker-mailserver/pull/4120) as part of DMS v15.0.0 ([#4429](https://github.com/docker-mailserver/docker-mailserver/pull/4429))
```

polarathene (changes requested):
LGTM thanks, just an improvement to the changelog.

casperklein (approved):

mpldr:
Done as requested :)

casperklein (approved):

casperklein:
Not related to this PR: Does anyone knows, why there are these `\s*` in the regex? If never saw mail headers prefixed by space(s).

georglauterbach:
> Not related to this PR: Does anyone knows, why there are these `\s*` in the regex? If never saw mail headers prefixed by space(s).

I don't know either. I'll revise this after afterward anyway, the whitespaces are very likely not required.

georglauterbach (approved):

polarathene:
> I'll revise this after afterward anyway, the whitespaces are very likely not required.

FWIW, I did [investigate this feature in DMS](https://github.com/docker-mailserver/docker-mailserver/issues/4224#issuecomment-2408077643) and it looks like it had a poor implementation to begin with: https://github.com/docker-mailserver/docker-mailserver/issues/4263#issuecomment-2483889812

> _AFAIK, this was intended only for affecting mail sent from Postfix, but looks like it was misconfigured to also apply to mail arriving to Postfix to give to Dovecot._

polarathene (approved):

georglauterbach:
This probably warrants a v15.0.2. Do you all agree?

mpldr:
Not sure how frequently its encountered, but I would think those who are would very much appreciate a timely fix.

polarathene:
> Do we want to release v15.0.2 right away?

Given the regression that'd be wise to do. Anything else that props up over the next week can go into a `.3` release if needed before merging in changes for 15.1.

georglauterbach:
ref #4432

PR #4433, opened 2025-03-27
docs: Dovecot Solr - Add compatibility note

beertje44:
# Description

Added some notes for the current release of solr (9.8) and deployment using podman.

## Type of change

## Checklist

- [X] If necessary, I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/docs/content/examples/tutorials/dovecot-solr.md
+++ b/docs/content/examples/tutorials/dovecot-solr.md
@@ -163,7 +163,7 @@ services:
 
     From the Solr 10 release onwards, this opt-in ENV will no longer be available.
 
-    If Dovecot has not updated their example Solr config ([upstream PR][dovecot::pr::solr-config-lib]), you will need to manually modify the Solr XML config.
+    If Dovecot has not updated their example Solr config ([upstream PR][dovecot::pr::solr-config-lib]), you will need to manually modify the Solr XML config to remove the `<lib>` directives and replace the suggested ENV `SOLR_CONFIG_LIB_ENABLED=true` with `SOLR_MODULES=analysis-extras`.
 
 [docs::user-patches]: ../../config/advanced/override-defaults/user-patches.md
 [docs::dovecot::full-text-search]: ../../config/advanced/full-text-search.md
--- a/docs/content/examples/tutorials/dovecot-solr.md
+++ b/docs/content/examples/tutorials/dovecot-solr.md
@@ -146,18 +146,24 @@ docker compose exec mailserver doveadm fts rescan -A
 
     Usually within 15 minutes or so, you should be able to search your mail using the Dovecot FTS feature! :tada:
 
-### Notes
+### Compatibility
 
-1. As of writing of this guide, SOLR 9.8 is the current release. This release introduces a notable [change][solr-9.8-change]: it no longer automatically loads additional jars via <lib> directives. This behaviour is now deprecated and will be completely removed from SOLR 10 and onwards. The current dovecot scheme however does need this, so you need to add an extra environment variable to your `compose.yaml` if you want to use SOLR 9.8.x:
+Since Solr 9.8.0 was released (Jan 2025), a breaking change [deprecates support for `<lib>` directives][solr::9.8::lib-directive] which is presently used by the Dovecot supplied Solr config (`solr-config-9.xml`) to automatically load additional jars required.
+
+To enable support for `<lib>` directives, add the following ENV to your `solr` container:
 
 ```yaml
-...
+services:
+  solr:
     environment:
-      ...
-      - SOLR_OPTS: "-Dsolr.config.lib.enabled=true"
+      SOLR_CONFIG_LIB_ENABLED: true
 ```
 
-2. If you use [podman][podman] the composer build 'trick' only works with more recent versions of podman-compose, more recent than the version that currently ships with [EPEL][fedora-epel] for example. You should really be using [Quadlets][podman-quadlets] though if you use podman, which is a topic of its own and doesn't fit in this tutorial.
+!!! warning "Solr 10"
+
+    From the Solr 10 release onwards, this opt-in ENV will no longer be available.
+
+    If Dovecot has not updated their example Solr config ([upstream PR][dovecot::pr::solr-config-lib]), you will need to manually modify the Solr XML config.
 
 [docs::user-patches]: ../../config/advanced/override-defaults/user-patches.md
 [docs::dovecot::full-text-search]: ../../config/advanced/full-text-search.md
@@ -168,8 +174,5 @@ docker compose exec mailserver doveadm fts rescan -A
 [github-solr]: https://github.com/apache/solr
 [github-dovecot::core-docs]: https://github.com/dovecot/core/tree/main/doc
 
-[solr-9.8-change]: https://issues.apache.org/jira/browse/SOLR-16781
-
-[podman]: https://podman.io
-[fedora-epel]: https://docs.fedoraproject.org/en-US/epel/
-[podman-quadlets]: https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html
+[solr::9.8::lib-directive]: https://issues.apache.org/jira/browse/SOLR-16781
+[dovecot::pr::solr-config-lib]: https://github.com/dovecot/core/pull/238
--- a/docs/content/examples/tutorials/dovecot-solr.md
+++ b/docs/content/examples/tutorials/dovecot-solr.md
@@ -157,7 +157,7 @@ docker compose exec mailserver doveadm fts rescan -A
       - SOLR_OPTS: "-Dsolr.config.lib.enabled=true"
 ```
 
-2. If you use [podman][podman] the composer build 'trick' only works with more recent versions of podman-compose, more recent then the version that currently ships with [EPEL][fedora-epel] for example. You should really be using [Quadlets][podman-quadlets] though if you use podman, which is a topic of its own and doesn't fit in this tutorial.
+2. If you use [podman][podman] the composer build 'trick' only works with more recent versions of podman-compose, more recent than the version that currently ships with [EPEL][fedora-epel] for example. You should really be using [Quadlets][podman-quadlets] though if you use podman, which is a topic of its own and doesn't fit in this tutorial.
 
 [docs::user-patches]: ../../config/advanced/override-defaults/user-patches.md
 [docs::dovecot::full-text-search]: ../../config/advanced/full-text-search.md
--- a/docs/content/examples/tutorials/dovecot-solr.md
+++ b/docs/content/examples/tutorials/dovecot-solr.md
@@ -148,7 +148,7 @@ docker compose exec mailserver doveadm fts rescan -A
 
 ### Notes
 
-1. As of writing of this guide, SOLR 9.8 is the current release. This release introduces a notable [change][solr-9.8-change]: it no longer automaticly loads additional jars via <lib> directives. This behaviour is no deprecated and will be completely removed from SOLR 10 and onwards. The current dovecot scheme however does need this so you need to add an extra environment variable to your `compose.yaml` if you want to use SOLR 9.8.x:
+1. As of writing of this guide, SOLR 9.8 is the current release. This release introduces a notable [change][solr-9.8-change]: it no longer automatically loads additional jars via <lib> directives. This behaviour is now deprecated and will be completely removed from SOLR 10 and onwards. The current dovecot scheme however does need this, so you need to add an extra environment variable to your `compose.yaml` if you want to use SOLR 9.8.x:
 
 ```yaml
 ...
--- a/docs/content/examples/tutorials/dovecot-solr.md
+++ b/docs/content/examples/tutorials/dovecot-solr.md
@@ -146,6 +146,19 @@ docker compose exec mailserver doveadm fts rescan -A
 
     Usually within 15 minutes or so, you should be able to search your mail using the Dovecot FTS feature! :tada:
 
+### Notes
+
+1. As of writing of this guide, SOLR 9.8 is the current release. This release introduces a notable [change][solr-9.8-change]: it no longer automaticly loads additional jars via <lib> directives. This behaviour is no deprecated and will be completely removed from SOLR 10 and onwards. The current dovecot scheme however does need this so you need to add an extra environment variable to your `compose.yaml` if you want to use SOLR 9.8.x:
+
+```yaml
+...
+    environment:
+      ...
+      - SOLR_OPTS: "-Dsolr.config.lib.enabled=true"
+```
+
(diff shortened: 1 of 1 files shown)


docs/content/examples/tutorials/dovecot-solr.md
```diff
@@ -146,6 +146,19 @@ docker compose exec mailserver doveadm fts rescan -A
 
     Usually within 15 minutes or so, you should be able to search your mail using the Dovecot FTS feature! :tada:
 
+### Notes
+
+1. As of writing of this guide, SOLR 9.8 is the current release. This release introduces a notable [change][solr-9.8-change]: it no longer automaticly loads additional jars via <lib> directives. This behaviour is no deprecated and will be completely removed from SOLR 10 and onwards. The current dovecot scheme however does need this so you need to add an extra environment variable to your `compose.yaml` if you want to use SOLR 9.8.x:
```
casperklein, line 151:
```suggestion
1. As of writing of this guide, SOLR 9.8 is the current release. This release introduces a notable [change][solr-9.8-change]: it no longer automatically loads additional jars via <lib> directives. This behaviour is now deprecated and will be completely removed from SOLR 10 and onwards. The current dovecot scheme however does need this, so you need to add an extra environment variable to your `compose.yaml` if you want to use SOLR 9.8.x:
```

docs/content/examples/tutorials/dovecot-solr.md
```diff
@@ -146,6 +146,19 @@ docker compose exec mailserver doveadm fts rescan -A
...
+
+```yaml
+...
+    environment:
+      ...
+      - SOLR_OPTS: "-Dsolr.config.lib.enabled=true"
+```
+
+2. If you use [podman][podman] the composer build 'trick' only works with more recent versions of podman-compose, more recent then the version that currently ships with [EPEL][fedora-epel] for example. You should really be using [Quadlets][podman-quadlets] though if you use podman, which is a topic of its own and doesn't fit in this tutorial.
```
casperklein, line 160:
```suggestion
2. If you use [podman][podman] the composer build 'trick' only works with more recent versions of podman-compose, more recent than the version that currently ships with [EPEL][fedora-epel] for example. You should really be using [Quadlets][podman-quadlets] though if you use podman, which is a topic of its own and doesn't fit in this tutorial.
```

polarathene:
Will this environment workaround remain compatible with SOLR 10? Or is it only applicable to 9.8 and any other versions prior to 10?

beertje44:
Replied via email but don't see it here: stricly speaking as of now 9.8 is a requirement. However it is not uncommon that the latest and greatest SOLR has features that are incomtible with a specific scheme so most times there is a deley between major releases and specific use cases. Second dovecot as of now does support the most current SOLR, which is good.

What the exact policy for SOLR releases is I find hard to find, I did find this however:

https://solr.apache.org/docs/9_8_1/changes/Changes.html#older

8 was released in 2019 and the latest 8 version is from september 2024. 9 was released in 2022 and still is the current release,

docs/content/examples/tutorials/dovecot-solr.md
```diff
@@ -146,6 +146,19 @@ docker compose exec mailserver doveadm fts rescan -A
...
+1. As of writing of this guide, SOLR 9.8 is the current release. This release introduces a notable [change][solr-9.8-change]: it no longer automatically loads additional jars via <lib> directives. This behaviour is now deprecated and will be completely removed from SOLR 10 and onwards. The current dovecot scheme however does need this, so you need to add an extra environment variable to your `compose.yaml` if you want to use SOLR 9.8.x:
+
+```yaml
+...
+    environment:
+      ...
+      - SOLR_OPTS: "-Dsolr.config.lib.enabled=true"
+```
+
+2. If you use [podman][podman] the composer build 'trick' only works with more recent versions of podman-compose, more recent than the version that currently ships with [EPEL][fedora-epel] for example. You should really be using [Quadlets][podman-quadlets] though if you use podman, which is a topic of its own and doesn't fit in this tutorial.
```
polarathene, line 160: - `environment` is either object key/value pairs, or a list of `key=value` strings. The example you've provided looks like a mix of that which is invalid AFAIK? Use the object key syntax like already use in on this page.
- The ENV `SOLR_CONFIG_LIB_ENABLED=true` seems to be a better choice than `SOLR_OPTS`. Presumably that makes the Podman Compose caveat redundant thus I've dropped that as I could not identify relevant issue or release notes regarding the mentioned incompatibility and it being fixed.
- Project name is "Solr" not "SOLR" AFAIK?
- Your environment snippet doesn't indicate the service to append the ENV to (_earlier `compose.yaml` on this page has both `solr` and `mailserver`_), better to avoid ambiguity to the user.

I've rewritten the contribution with the above changes applied. I don't use Solr myself so if you could verify my assumptions, that would be appreciated.

````suggestion
### Compatibility

Since Solr 9.8.0 was released (Jan 2025), a breaking change [deprecates support for `<lib>` directives][solr::9.8::lib-directive] which is presently used by the Dovecot supplied Solr config (`solr-config-9.xml`) to automatically load additional jars required.

To enable support for `<lib>` directives, add the following ENV to your `solr` container:

```yaml
services:
  solr:
    environment:
      SOLR_CONFIG_LIB_ENABLED: true
```

!!! warning "Solr 10"

    From the Solr 10 release onwards, this opt-in ENV will no longer be available.

    If Dovecot has not updated their example Solr config ([upstream PR][dovecot::pr::solr-config-lib]), you will need to manually modify the Solr XML config.
````

---

## Context

I looked into this change (_thanks for the bug report link that had the information I needed_). The below are notes to support suggested changes.

The [associated upstream commit](https://github.com/apache/solr/commit/f492e24881c#diff-469f7b0e2e6843ecf2eb42ca9e5b867386fe140a061f0043af6838b7853414ebR74-R78) documents changelog:

> _`<lib/>` directives have been made into an "opt-in" feature in Solr 9.8 and must be explicitly enabled if desired (using a system property or environment variable like `-Dsolr.config.lib.enabled=true`)_
> - _If not explicitly enabled, Solr will log a warning but otherwise quietly ignore any `<lib/>` tags in configsets._
> - _This is a breaking change, and we apologize for the disruption, but it's a necessary response to a number of security gaps that have come to light around this feature._
> - _`<lib/>` directives should be considered deprecated and will be removed entirely in Solr 10.0._
> _Current users of `<lib/>` have a number of options to ensure their libraries remain on Solr's classpath, including: `SOLR_MODULES`, core "lib" directories, the `solr.xml` `<sharedLib>` tag, Solr's package manager, etc._

So I see the `-Dsolr.config.lib.enabled=true` advice there being used with `SOLR_OPTS` ENV for our advice here, which I assume is what is referred to as a sysprop / system property?

Another part of the associated commit is a [separate changelog](https://github.com/apache/solr/commit/f492e24881c#diff-cdd1ae1b418ea74ca689c6e68e61f4e5b4c6f77f6408a4e6cf76ed8421b24d09R145) from the earlier quoted upgrade notes changelog:

> _`SOLR-16781`: "<lib/>" tags in `solrconfig.xml` are now quietly ignored by default unless explicitly enabled with the `SOLR_CONFIG_LIB_ENABLED=true` enviroment variable (or corresponding sysprop).  These tags are now considered deprecated and will be removed in Solr 10._

Which suggests a more specific and simpler ENV `SOLR_CONFIG_LIB_ENABLED=true` may also work? (_**EDIT:** Yes it should work, see [ENV convention docs for mapping to JVM system properties](https://solr.apache.org/guide/solr/latest/configuration-guide/property-substitution.html#jvm-system-properties): `SOLR_` prefix + sysprop where `_` is substituted for `.`_)

Either way it's communicated clearly that this is an interim opt-in for 9.8+ to continue supporting `<lib>` directive until Solr 10 release. There are other alternatives available as noted that will remain compatible with Solr 10, will these docs be updated to provide guidance with that instead for Solr 10 compatibility once available?

**EDIT:** My bad, I see the `<lib>` directive is dependent upon the [upstream Dovecot referenced `solr-config-9.xml` config](https://github.com/dovecot/core/blob/60d431a0cc0890dbddcea121ddb3eee7ac001354/doc/solr-config-9.xml#L40-L86). There is an [open PR upstream to resolve this](https://github.com/dovecot/core/pull/238), which removes the `<lib>` directives and suggests adjusting the Solr command options to include `-Dsolr.modules=analysis-extras`, which I guess you could use `SOLR_OPTS` env for if that's not already set.

File: docs/content/examples/tutorials/dovecot-solr.md [hunk 4]
```diff
@@ -154,3 +167,9 @@ docker compose exec mailserver doveadm fts rescan -A
 [dockerfile-solr-uidgid]: https://github.com/apache/solr-docker/blob/9cd850b72309de05169544395c83a85b329d6b86/9.6/Dockerfile#L89-L92
 [github-solr]: https://github.com/apache/solr
 [github-dovecot::core-docs]: https://github.com/dovecot/core/tree/main/doc
+
+[solr-9.8-change]: https://issues.apache.org/jira/browse/SOLR-16781
+
+[podman]: https://podman.io
+[fedora-epel]: https://docs.fedoraproject.org/en-US/epel/
+[podman-quadlets]: https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html
```
polarathene, line 175:
```suggestion
[solr::9.8::lib-directive]: https://issues.apache.org/jira/browse/SOLR-16781
[dovecot::pr::solr-config-lib]: https://github.com/dovecot/core/pull/238
```

polarathene (changes requested):
Revised, thanks for the contribution ❤️

docs/content/examples/tutorials/dovecot-solr.md
beertje44: Thank you for your thorough review, apreaciate all the hard work you put into dms.

I can confrim your env variable works and imho it looks cleaner.

However the podman remark has nothing to do with solr but with the way we include dovecot-solr into dms, I noticed this before and even opened a bug report for this untill I noticed podman-compose would build using a Dockerfile which just happens to be in the same place as composer.yaml if you checkout the dms github repo ... and to make matters worse: in my case just happened to have docker-solr included also so I never noticed before :smiley: podman-compose however can be considered as legacy though, so hence my remark.

Will update again once Solr 10 comes out. 9 will be supported for at least a year though and requiring the latest Solr really isn't a nescessety in my experience.

polarathene: I don't quite follow the relevance to the Podman Compose mention then? If it was about building a custom image Podman doesn't support `SHELL` instruction in our `Dockerfile` unless you provide an option to use docker compatibility (_default is to build an OCI image, and `SHELL` is not part of the OCI spec_).

If you believe the compose concern is worth documenting, a reference link to the associated issue will be required as it's otherwise difficult to maintain docs when we can't ascertain if they're relevant at a future date.

We do have an active PR update to the Podman docs for Quadlets instead of podman compose once I can verify some networking config.

---

> Will update again once Solr 10 comes out. 9 will be supported for at least a year though and requiring the latest Solr really isn't a necessity in my experience.

I forgot to mention it but the ENV `SOLR_MODULES=analysis-extras` is the correct way forward to load the module. Dovecot upstream will likely update their config to drop the `<lib>` directives and that will be the suggested way to config the lib/module needed to load.

You should be able to use `SOLR_MODULES` already with 9.8 I believe, and Solr will just log a warning about the `<lib>` directive usage while ignoring it, but with the `SOLR_MODULES` ENV configured it should work all the same.

polarathene (dismissed):

docs/content/examples/tutorials/dovecot-solr.md
```diff
@@ -146,6 +146,25 @@ docker compose exec mailserver doveadm fts rescan -A
...
+  solr:
+    environment:
+      SOLR_CONFIG_LIB_ENABLED: true
+```
+
+!!! warning "Solr 10"
+
+    From the Solr 10 release onwards, this opt-in ENV will no longer be available.
+
+    If Dovecot has not updated their example Solr config ([upstream PR][dovecot::pr::solr-config-lib]), you will need to manually modify the Solr XML config.
```
polarathene, line 166:
```suggestion
!!! warning "Solr 10"

    From the Solr 10 release onwards, this opt-in ENV will no longer be available.

    If Dovecot has not updated their example Solr config ([upstream PR][dovecot::pr::solr-config-lib]), you will need to manually modify the Solr XML config to remove the `<lib>` directives and replace the suggested ENV `SOLR_CONFIG_LIB_ENABLED=true` with `SOLR_MODULES=analysis-extras`.
```

polarathene (approved):

docs/content/examples/tutorials/dovecot-solr.md
beertje44: > I don't quite follow the relevance to the Podman Compose mention then? If it was about building a custom image Podman doesn't support `SHELL` instruction in our `Dockerfile` unless you provide an option to use docker compatibility (_default is to build an OCI image, and `SHELL` is not part of the OCI spec_).
> 
> If you believe the compose concern is worth documenting, a reference link to the associated issue will be required as it's otherwise difficult to maintain docs when we can't ascertain if they're relevant at a future date.
> 
> We do have an active PR update to the Podman docs for Quadlets instead of podman compose once I can verify some networking config.
> 

This tutorial uses a docker-compose as an example. Perhaps mentioning podman is overkill, however "podman-compose" and "podman compose" both exists and it seems they only recently started to support the build construction we use here. That was all. But you are right, quadlets are the way forward.

Perhaps a line like this then:

"If you use podman you should be looking at quadlets to implement all of this. For docker-mailserver you probably want to use a small Dockerfile which adds dovecot-solr to the standard docker-mailserver image."

Used podman-compose and podman generate systemd a lot, but there are so many things we used to do :wink:

polarathene: We don't see Podman too often in reports nor do we officially support it, let alone either compose variant.

We do have a page to provide some advice for it like we do for kubernetes, but everything else stays focused around Docker Compose so I think we'll keep it like that. We make exceptions when it can be more helpful than _"using alternatives to Docker Compose may have compatibility issues"_.

So without anything more specific and helpful, I think we're good. Once the related podman PR is merged we'll have fairly good overview for Podman users to lean in to Quadlets if they're not already :)

beertje44: Perhaps I can do something in that matter, I don't have everyting converted to quadlets yet. but dms is running in a pod from quadlets and runs just fine. Do think podman is not far fetched for dms though, it's a rather serious container project suitable for small businesses and podman is imho better integrated with Linux server OSes then vanilla docker. But lets keep it ontopic: agreed and from my point of view ready to merge :smile:

polarathene: I understand, but as no maintainer has that much experience with podman AFAIK we're not able to engage that much with users for support so we keep it simple and focused on Docker Compose, deferring other container engine support to community.

I did take interest in learning more about Podman with Quadlets, but there is a section in our docs referring to the older rootless networking mode with it's caveats that I need to look into with Quadlets and compared to the current rootless driver `pasta`.

I think for rootful quadlets it may also be a slightly different situation with the network driver/behaviour, so I just wanted to make sure the docs are accurate for users, not accidentally interleaved with outdated information - especially for maintenance it can be quite confusing to try follow docs without realizing technical details in some sections of the page have fallen out of sync.

PR #4450, opened 2025-04-20
feat: Configurable poll rate for `check-for-changes.sh`

### LINKED ISSUE #4414: feature request: Idle load - ``check-for-changes.sh`` improve execution interval
### Context

I noticed that amongst other container docker-mailservers has an unusal high idle load on my machine.

While debugging the running process I saw that ``check-for-changes.sh`` causes roughly ~50% of the idle load

Note: ``ENABLE_RSPAMD`` is set to 1

### Description

``check-for-changes.sh`` checks every 2 seconds for changes:

https://github.com/docker-mailserver/docker-mailserver/blob/6b1a5664979528064672f056bf0dd8fe0c421952/target/scripts/check-for-changes.sh#L213-L214

As described above this causes a lot of idle load.

The check invokes 

https://github.com/docker-mailserver/docker-mailserver/blob/6b1a5664979528064672f056bf0dd8fe0c421952/target/scripts/check-for-changes.sh#L37-L60

and this executes

https://github.com/docker-mailserver/docker-mailserver/blob/6b1a5664979528064672f056bf0dd8fe0c421952/target/scripts/helpers/change-detection.sh#L24-L80

As far as I can tell these extremely frequent checks are likely only required when debugging/actively changing configuration files.

I therefore propose the following:
1. Make the change interval configurable (e.g. using a env variable like ``CHECK_FOR_CHANGES_INTERVAL_SEC``)
2. Change the default interval sec to something higher like ``5`` or ``10`` to lessen the default idle load
3. Document this somewhere 
    * Recommend lower values during active development/configuration and higher ones when not

### Alternatives

N/A

### Applicable Users

Everyone

I can provide a PR (changed code is likey ~1 line).

See also: https://github.com/docker-mailserver/docker-mailserver/issues/2348

litetex:
# Description

Makes the ``check-for-changes.sh`` execution interval configurable (see issue for details).

Picked a default value of ``20`` because
* it roughly halfs the idle load 
* it's still reactive enough when changes occur

Fixes #4414 

## Type of change

## Checklist

### DIFF
--- a/docs/content/config/environment.md
+++ b/docs/content/config/environment.md
@@ -360,11 +360,22 @@ Default: empty (no prefix will be added to e-mails)
 
     Add trailing white-space by quote wrapping the value: `SPAM_SUBJECT='[SPAM] '`
 
-##### CHECK_FOR_CHANGES_INTERVAL_SEC
+##### DMS_CONFIG_POLL
 
-Defines how often the background check for changes (e.g. if the configuration changed or a new certificate needs to be applied) should be executed.
+Defines how often DMS polls [monitored config files][gh::monitored-configs] for changes in the DMS Config Volume. This also includes TLS certificates and is often relied on for applying changes managed via `setup` CLI commands.
 
-Default: ``20``
+- **`2`** => How often (in seconds) [change detection][gh::check-for-changes] is performed.
+
+!!! note "Decreasing the frequency of polling for changes"
+
+    Raising the value will delay how soon a change is detected which may impact UX expectations for responsiveness, but reduces resource usage when changes are rare.
+
+!!! info
+
+    When using `ACCOUNT_PROVISIONER=LDAP`, the change detection feature is presently disabled.
+
+[gh::check-for-changes]: https://github.com/docker-mailserver/docker-mailserver/blob/v15.0.0/target/scripts/check-for-changes.sh#L37
+[gh::monitored-configs]: https://github.com/docker-mailserver/docker-mailserver/blob/v15.0.0/target/scripts/helpers/change-detection.sh#L30-L42
 
 #### Rspamd
 
--- a/docs/content/config/environment.md
+++ b/docs/content/config/environment.md
@@ -360,6 +360,12 @@ Default: empty (no prefix will be added to e-mails)
 
     Add trailing white-space by quote wrapping the value: `SPAM_SUBJECT='[SPAM] '`
 
+##### CHECK_FOR_CHANGES_INTERVAL_SEC
+
+Defines how often the background check for changes (e.g. if the configuration changed or a new certificate needs to be applied) should be executed.
+
+Default: ``20``
+
 #### Rspamd
 
 ##### ENABLE_RSPAMD
--- a/target/scripts/check-for-changes.sh
+++ b/target/scripts/check-for-changes.sh
@@ -211,7 +211,7 @@ function _rspamd_changes() {
 
 while true; do
   _check_for_changes
-  sleep ${CHECK_FOR_CHANGES_INTERVAL_SEC:-20}
+  sleep ${DMS_CONFIG_POLL:-2}
 done
 
 exit 0
--- a/target/scripts/check-for-changes.sh
+++ b/target/scripts/check-for-changes.sh
@@ -211,7 +211,7 @@ function _rspamd_changes() {
 
 while true; do
   _check_for_changes
-  sleep 2
+  sleep ${CHECK_FOR_CHANGES_INTERVAL_SEC:-20}
 done
 
 exit 0
--- a/target/scripts/check-for-changes.sh
+++ b/target/scripts/check-for-changes.sh
@@ -211,7 +211,7 @@ function _rspamd_changes() {
 
 while true; do
   _check_for_changes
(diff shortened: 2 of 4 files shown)


docs/content/config/environment.md
```diff
@@ -360,6 +360,12 @@ Default: empty (no prefix will be added to e-mails)
 
     Add trailing white-space by quote wrapping the value: `SPAM_SUBJECT='[SPAM] '`
 
+##### CHECK_FOR_CHANGES_INTERVAL_SEC
+
+Defines how often the background check for changes (e.g. if the configuration changed or a new certificate needs to be applied) should be executed.
+
+Default: ``20``
```
polarathene, line 367: - ENV var renamed
- Default line changed to match the format used with other poll ENV.
- Rewrote the description with reference links so users have better insight.

```suggestion
##### DMS_CONFIG_POLL

Defines how often DMS polls [monitored config files][gh::monitored-configs] for changes in the DMS Config Volume. This also includes TLS certificates and is often relied on for applying changes managed via `setup` CLI commands.

- **`2`** => How often (in seconds) [change detection][gh::check-for-changes] is performed.

!!! note "Decreasing the frequency of polling for changes"

    Raising the value will delay how soon a change is detected which may impact UX expectations for responsiveness, but reduces resource usage when changes are rare.

!!! info

    When using `ACCOUNT_PROVISIONER=LDAP`, the change detection feature is presently disabled.

[gh::check-for-changes]: https://github.com/docker-mailserver/docker-mailserver/blob/v15.0.0/target/scripts/check-for-changes.sh#L37
[gh::monitored-configs]: https://github.com/docker-mailserver/docker-mailserver/blob/v15.0.0/target/scripts/helpers/change-detection.sh#L30-L42
```

target/scripts/check-for-changes.sh
```diff
@@ -211,7 +211,7 @@ function _rspamd_changes() {
 
 while true; do
   _check_for_changes
-  sleep 2
+  sleep ${CHECK_FOR_CHANGES_INTERVAL_SEC:-20}
```
polarathene, line 214:
```suggestion
  sleep ${DMS_CONFIG_POLL:-2}
```

Pretty sure if the default is left at 20 it'll affect our CI tests without having the ENV updated there to reduce it.

While I understand wanting to have a more sane default than `2`, this would negatively impact new users trying to setup their DMS instance based on our docs with the `setup` CLI, having to wait up to 20 seconds until the update is actually applied (_the CLI will not wait, it updates the monitored config, then defers to `check-for-changes.sh`_). As such it'd also be breaking for any automation reliant upon that existing default (like our CI).

Your main concern would be completely avoided should someone contribute `watchexec`, which for the majority of users would remove any polling entirely.

---

The other maintainers may request that this ENV is also handled here:

https://github.com/docker-mailserver/docker-mailserver/blob/ef66dd5d12c4f0efa2b2e345ffdce796e3a7df03/target/scripts/startup/variables-stack.sh#L50-L53

That's where defaults are usually set (and written to `/etc/dms-settings`), but since this is specifically for `check-for-changes.sh` and not likely to be prone to breakage/conflict, I'm not going to require it.

polarathene (changes requested):
Thanks for the PR :)

We have `_POLL` for Getmail and Fetchmail services, I think we can be consistent with that suffix.

Likewise `DMS_` as a prefix to remove ambiguity. "Check for changes" is a legacy name, as the feature hasn't been officially documented yet we can just refer to it as "change detection". Retaining the script name reference in the changelog is fine though, added a link to the relevant docs page for more context (such as the default).

litetex:
@polarathene

Thank you for the feedback.

I applied you changes accordingly :)

target/scripts/check-for-changes.sh
casperklein: > The other maintainers may request that this ENV is also handled here:
> 
> https://github.com/docker-mailserver/docker-mailserver/blob/ef66dd5d12c4f0efa2b2e345ffdce796e3a7df03/target/scripts/startup/variables-stack.sh#L50-L53
> 
> That's where defaults are usually set (and written to `/etc/dms-settings`), but since this is specifically for `check-for-changes.sh` and not likely to be prone to breakage/conflict, I'm not going to require it.

For completeness, please add the new ENV to the list.

target/scripts/check-for-changes.sh
```diff
@@ -211,7 +211,7 @@ function _rspamd_changes() {
 
 while true; do
   _check_for_changes
-  sleep 2
+  sleep ${DMS_CONFIG_POLL:-2}
```
casperklein, line 214:
```suggestion
  sleep "${DMS_CONFIG_POLL:-2}"
```

polarathene:
Oh whoops, I applied the change suggestion from @casperklein before noticing his other feedback 😓 

@litetex feel free to force push your own branch or pull that change and then commit the other change request.

LGTM thus far though 👍

target/scripts/check-for-changes.sh
litetex: Done :)

polarathene (approved):

casperklein (approved):

PR #4453, opened 2025-04-23
fix: `setup email list` should only work with `ACCOUNT_PROVISIONER=FILE`

polarathene:
# Description

This addresses https://github.com/orgs/docker-mailserver/discussions/4452#discussioncomment-12918479

It should provide a better UX than a bunch of unhelpful errors like the user experienced.

## Type of change

## Checklist

### DIFF
--- a/target/bin/listmailuser
+++ b/target/bin/listmailuser
@@ -8,6 +8,10 @@ source /usr/local/bin/helpers/index.sh
 source /etc/dms-settings 2>/dev/null
 
 function _main() {
+  if [[ ${ACCOUNT_PROVISIONER} != 'FILE' ]]; then
+    _exit_with_error "This command is only compatible with 'ACCOUNT_PROVISIONER=FILE'"
+  fi
+
   local DATABASE_ACCOUNTS='/tmp/docker-mailserver/postfix-accounts.cf'
   local DATABASE_VIRTUAL='/tmp/docker-mailserver/postfix-virtual.cf'
 
--- a/target/scripts/startup/variables-stack.sh
+++ b/target/scripts/startup/variables-stack.sh
@@ -116,8 +116,8 @@ function __environment_variables_general_setup() {
 
   # The Dovecot Quotas feature is presently only supported with the default FILE account provisioner,
   # Enforce disabling the feature, unless it's been explicitly set via ENV (to avoid mismatch between explicit ENV and sourcing from /etc/dms-settings)
-  if [[ ${ACCOUNT_PROVISIONER} != 'FILE' ]] || [[ ${SMTP_ONLY} -eq 1 ]] || [[ ${ENABLE_QUOTAS} -ne 1 ]]; then
-    _log 'debug' "The 'ENABLE_QUOTAS' feature is enabled by default but is not compatible with your config. Disabling"
+  if [[ ${ACCOUNT_PROVISIONER} != 'FILE' || ${SMTP_ONLY} -eq 1 ]] && [[ ${ENABLE_QUOTAS:-1} -eq 1 ]]; then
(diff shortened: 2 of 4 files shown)


target/bin/listmailuser
```diff
@@ -8,6 +8,10 @@ source /usr/local/bin/helpers/index.sh
 source /etc/dms-settings 2>/dev/null
 
 function _main() {
+  if [[ ${ACCOUNT_PROVISIONER} != 'FILE' ]]; then
+    _exit_with_error "This command is only compatible with 'ACCOUNT_PROVISIONER=FILE'"
+  fi
```
polarathene, line 13: This could technically be dropped, but I don't think DMS is intended to support both file based accounts and LDAP at the same time, so I'm inclined to keep the early exit constraint here? 🤔

georglauterbach: Fail-fast is the right approach here, indeed.

georglauterbach (approved):

georglauterbach (changes requested):
See my comment <https://github.com/docker-mailserver/docker-mailserver/pull/4453/files#r2056882452>

target/scripts/startup/variables-stack.sh
```diff
@@ -115,6 +114,15 @@ function __environment_variables_general_setup() {
...
   VARS[ENABLE_UPDATE_CHECK]="${ENABLE_UPDATE_CHECK:=1}"
 
+  # The Dovecot Quotas feature is presently only supported with the default FILE account provisioner,
+  # Enforce disabling the feature, unless it's been explicitly set via ENV (to avoid mismatch between explicit ENV and sourcing from /etc/dms-settings)
+  if [[ ${ACCOUNT_PROVISIONER} != 'FILE' ]] || [[ ${SMTP_ONLY} -eq 1 ]] || [[ ${ENABLE_QUOTAS} -ne 1 ]]; then
+    _log 'debug' "The 'ENABLE_QUOTAS' feature is enabled by default but is not compatible with your config. Disabling"
+    VARS[ENABLE_QUOTAS]="${ENABLE_QUOTAS:=0}"
+  else
+    VARS[ENABLE_QUOTAS]="${ENABLE_QUOTAS:=1}"
+  fi
```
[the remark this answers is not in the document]
polarathene: > I think my suggestion should have fixed this and only disables quotas forcibly if
> 
> 1. `$ACCOUNT_PROVISIONER` is not `FILE` _or_ `$SMTP_ONLY` is `1`
> 2. and `ENABLE_QUOTAS` is `1`

The intention was that if either of the first two conditions were true, we want the default `ENABLE_QUOTAS` to be set to `0`, **unless** it's been explicitly set to `1`.

The reasoning for the unless condition is to avoid the mess of someone setting `ENABLE_QUOTAS=1` and any scripts running that might not source from `/etc/dms-settings` then having a mismatch 🤷‍♂️ 

It would be a bit buggy for someone that used these incompatible settings with an explicit `ENABLE_QUOTAS=1`, but that'd be more obvious to troubleshoot than thinking about the context of what `ENABLE_QUOTAS` might be throughout scripts 😅 

So basically, if the user explicitly set `ENABLE_QUOTAS=1` we'd leave it alone.

---

That said, I should just trust anything in our scripts is getting the value set here, or sourcing it from `/etc/dms-settings`. At least with debug output the user would get informed about the incompatibility.

Your suggestion is definitely better, handling `ENABLE_QUOTAS` as unset or when set to `1`, and only in combination with either of the other two conditions 😅 Thanks!

georglauterbach:
I pushed a fix for the still-occurring test failure. Hope the issue is now fixed.

PR #4456, opened 2025-04-24
Add LOGIN_TRUSTED_NETWORKS option for Dovecot configuration

shunkica:
# Description

Add LOGIN_TRUSTED_NETWORKS option for Dovecot configuration.

Currently I need to add this option to the dovecot local config manually after creating the volume, but I would prefer if this was something that can be setup using the environment variables like the rest of the settings.

Personally I use it to display the correct client IP when requests come from Roundcube.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file. The format 
 
 - **Internal:**
   - [`DMS_CONFIG_POLL`](https://docker-mailserver.github.io/docker-mailserver/v15.0/config/environment/#dms_config_poll) supports adjusting the polling rate (seconds) for the change detection service `check-for-changes.sh` ([#4450](https://github.com/docker-mailserver/docker-mailserver/pull/4450))
+  - [`LOGIN_TRUSTED_NETWORKS`](https://docker-mailserver.github.io/docker-mailserver/v15.0/config/environment/#login_trusted_networks) allows specifying trusted networks for Dovecot login connections, enabling the client to tell the server what the original client's IP address was
 
 ### Updates
 
--- a/docs/content/config/environment.md
+++ b/docs/content/config/environment.md
@@ -320,6 +320,13 @@ Note: More details at <http://www.postfix.org/postconf.5.html#inet_protocols>
 
 Note: More information at <https://dovecot.org/doc/dovecot-example.conf>
 
+##### LOGIN_TRUSTED_NETWORKS
+
+Specify trusted networks for Dovecot login connections. This setting allows the client connection to tell the server what the original client's IP address was.
+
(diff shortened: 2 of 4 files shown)


polarathene:
> Currently I need to add this option to the dovecot local config manually after creating the volume, but I would prefer if this was something that can be setup using the environment variables like the rest of the settings.

We'd prefer not to add ENV that map a value to a single line setting. This has happened in the past when maintainers were not as strict but it is a bit of an anti-pattern IMO, so I'm against accepting such contributions without sufficient justification to reason the benefit over our existing config override support.

Would I be mistaken if this is just about simplifying your config by using ENV instead of one more file? If so you can avoid a separate file by using a Docker Compose feature `configs` to inline/embed a file in `compose.yaml` instead.

```yaml
services:
  dms:
    configs:
      - source: dovecot-override
        target: /tmp/docker-mailserver/dovecot.cf

configs:
  dovecot-override:
    content: |
      login_trusted_networks = 10.0.0.0/8 192.168.0.0/16
```

If you actually have a need for this to rely on being configured via ENV, you can use Docker Compose variable interpolation:

```yaml
configs:
  dovecot-override:
    content: |
      login_trusted_networks = ${LOGIN_TRUSTED_NETWORKS:?Trusted networks must bet set}
```

Now the ENV `LOGIN_TRUSTED_NETWORKS` must be set for this `compose.yaml` to run. You can have this ENV already set in the environment of the `docker compose` command, or [via an `.env` file](https://docs.docker.com/compose/how-tos/environment-variables/variable-interpolation/#ways-to-set-variables-with-interpolation).

---

Given the above I really don't see the need for this PR (_as much as I appreciate the effort, this would have been pointed out if feature requested prior to save you the time_ 😅)

Does the above approach I suggested instead work for you?

polarathene:
> Personally I use it to display the correct client IP when requests come from Roundcube.

I believe that's been an issue raised in the past. It might be good to document it via a FAQ entry if you'd like to help others be aware of the setting.

Another alternative that I think we could consider is integration of this setting with `PERMIT_DOCKER`. That's a bit broader in scope though, trusting the subnets it covers (_which presently isn't implemented well or that configurable.._).

Maybe if the feature got reworked, or we just make a docs page for manual configuration of the various service settings that could be manually configured, and like shown in my previous comment an ENV (_not a DMS one, unless handled via `user-patches.sh` instead of Compose `configs`_) could make it easy to set a subnet(s) across the settings.

shunkica:
Well it wasn't an absolute waste of time because at least I am now aware of the configs key. I have not tested it yet but it looks like that should solve my problem.

georglauterbach:
I'll close this for now; it seems the concern is addressed. We recommend opening an issue before directly opening a PR to avoid this.

PR #4483, opened 2025-05-22
docs: PROXY protocol advice for Traefik should reference correct entrypoints

Blitz-Cloud:
When the entrypoints for traefik are defined there are named mail-*(smtp,...), but it the dms labels when it comes to using these entry point in the TLS section just smtp is used, and for TLS it is used smtp-submissions witch is an invalid entrypoint

# Description

I tried recreating the setup for DMS using traefik as a reverse proxy, but instead of writing the entrypoints in the docker compose file, i used the toml version to do this. Following the docs and translating the cli commands into the toml I got this:

```tml
...............
entryPoints:
  mail-smtp:
    address: ":25"
  mail-submission:
    address: ":587"
  mail-submissions:
    address: ":465"
  mail-iamp:
    address: ":143"
  mail-imaps:
    address: ":993"
  mail-pop3:
    address: ":110"
  mail-pop3s:
    address: ":995"
  mail-managesieve:
    address: ":4190"
...............
```
The code above is created based on this part of the  [docs](https://docker-mailserver.github.io/docker-mailserver/latest/examples/tutorials/mailserver-behind-proxy/#configuration)

The problem arises when in the DMS labels because in the name of the entrypoints used aren't matching what is defined in the traefik configuration.

All i did is to change the name of the entrypoints for TLS and SSL to match what was declared. 

## Type of change

## Checklist

### DIFF
--- a/docs/content/examples/tutorials/mailserver-behind-proxy.md
+++ b/docs/content/examples/tutorials/mailserver-behind-proxy.md
@@ -114,15 +114,15 @@ The below guidance is focused on configuring [Traefik][traefik-web], but the adv
           # These are examples, configure the equivalent for any additional ports you proxy.
           # Explicit TLS (STARTTLS):
           - traefik.tcp.routers.mail-smtp.rule=HostSNI(`*`)
-          - traefik.tcp.routers.mail-smtp.entrypoints=smtp
-          - traefik.tcp.routers.mail-smtp.service=smtp
+          - traefik.tcp.routers.mail-smtp.entrypoints=mail-smtp
+          - traefik.tcp.routers.mail-smtp.service=mail-smtp
           - traefik.tcp.services.mail-smtp.loadbalancer.server.port=25
           - traefik.tcp.services.mail-smtp.loadbalancer.proxyProtocol.version=2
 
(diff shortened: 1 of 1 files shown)


polarathene (approved):
Thanks for spotting this ❤️ 

I believe I noticed this too a few months back but got side-tracked before I PR'd a correction 😓

If you have any other feedback about that page please let me know.

---

For the most part many users have been mistakenly using that guide just because they have Traefik, but the focus is meant to be on PROXY protocol and solving a use-case there when that's actually required (_Kubernetes ingress or other proxy layer between client and DMS_).

This setup is quite a bit verbose, and only needed when routing traffic to DMS through a reverse proxy like Traefik. Since DMS should be the only mailserver instance using those ports, we tend to advise keeping it simple and just exposing the ports directly. DMS after all needs to terminate TLS itself so Traefik is rarely providing any added benefit like it would with your web services.

polarathene:
@georglauterbach @casperklein did either of you enable Copilot for the repo? Or is this something Github is rolling out itself? First time I've seen it engaging in our PRs.

casperklein:
Copilot was suggested as reviewer, so I gave it a try 👍

Issue #4485, opened 2025-05-22
bug report: rspamd cant load DKIM key


s0ftcorn:
### 📝 Preliminary Checks

- [x] I tried searching for an existing issue and followed the [debugging docs](https://docker-mailserver.github.io/docker-mailserver/latest/config/debugging/) advice, but still need assistance.

### 👀 What Happened?

from  /var/log/mail/rspamd.log:
`#370(rspamd_proxy) <6A4B3F>; proxy; dkim_module_load_key_format: cannot load dkim key /var/lib/rspamd/dkim/domain.tld.dkim.key: cannot stat key file: '/var/lib/rspamd/dkim/domain.tld.dkim.key' No such file or directory`

./config/ is mounted to /tmp/docker-mailserver/ and contains:
```
~/m/c/rspamd> tree
.
├── custom-commands.conf
├── dkim
│   ├── rsa-2048-mail-anotherdomain.tld.public.dns.txt
│   ├── rsa-2048-mail-anotherdomain.tld.public.txt
│   ├── rsa-2048-mail-one.domain.tld.private.txt
│   ├── rsa-2048-mail-one.domain.tld.public.txt
│   ├── rsa-2048-mail-other-domain.tld.private.txt
│   ├── rsa-2048-mail-other-domain.tld.txt
│   ├── rsa-2048-mail-domain.tld.private.txt
│   └── rsa-2048-mail-domain.tld.public.txt
└── override.d
    ├── dkim_signing.conf
    ├── known_senders.conf
    ├── multimap.conf
    ├── whitelist.sender.domain.map
    └── whitelist.sender.email.map

```

And yes, as you can see from anotherdomain.tld the `docker exec -it <CONTAINER NAME> setup config dkim domain <DOMAIN>` doesnt generate a private key, so i had to do this manually as explained in https://github.com/orgs/docker-mailserver/discussions/3803#discussioncomment-8204632
I get a warning when executing  `docker exec -it <CONTAINER NAME> setup config dkim domain <DOMAIN>`: `rspamd-dkim: The Rspamd user ('_rspamd') seems to be unable to list files in the keys directory ('/tmp/docker-mailserver/rspamd/dkim') - Rspamd may experience permission errors later`
Permissions of ./config/rspamd are:
```
~/m/c/rspamd> ls -lahn
total 20K
drwxr-xr-x 4  111  114 4.0K Oct 10  2024 .
[...]

PR #4487, opened 2025-05-22
fix: Ensure DMS config volume can be accessed by non-root users

polarathene:
# Description

This is similar to a [recent fix for the DMS state volume](https://github.com/docker-mailserver/docker-mailserver/pull/4420), where a non-root user needs access but the host directory mounted has mistakenly been too aggressive with permissions for the directory (_one known example is [AWS ECS (EC2) using EBS volumes](https://github.com/docker-mailserver/docker-mailserver/issues/4419#issuecomment-2731430310)_).

It appears that Rspamd support with DKIM at least [expects this due to running operations as the `_rspamd` user](https://github.com/docker-mailserver/docker-mailserver/issues/4485#issuecomment-2902642469) to read content from the DMS config volume. In future this concern shouldn't be the case, as the root user during setup could make internal copies that anything running as `_rspamd` should instead interact with.

However since this is [not always an obvious failure](https://github.com/docker-mailserver/docker-mailserver/pull/4420) (_various unresolved reports in the past were tripped up by this for the state volume_), the fix is probably worth adding? (_makes me wonder if similar is prone to our other volumes too then_ 🤔)

**NOTE:** While we only need `chmod o+x`, we can use `chmod +x` here as [unlike other permissions like `+w` it appears to apply `+x` to `ugo` by default](https://github.com/docker-mailserver/docker-mailserver/pull/4420#discussion_r1999878806).

Fixes: #4485

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,11 @@ All notable changes to this project will be documented in this file. The format 
 - **Internal:**
   - [`DMS_CONFIG_POLL`](https://docker-mailserver.github.io/docker-mailserver/v15.0/config/environment/#dms_config_poll) supports adjusting the polling rate (seconds) for the change detection service `check-for-changes.sh` ([#4450](https://github.com/docker-mailserver/docker-mailserver/pull/4450))
 
+### Fixes
+
+- **Internal:**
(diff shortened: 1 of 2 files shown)


polarathene:
I've added a separate `DMS_CONFIG_DIR` variable instead of referencing `/tmp/docker-mailserver` directly, I could remove that, but we may want to consider at some point having a global var for this location as at some point it probably shouldn't be `/tmp/docker-mailserver` 🤔

casperklein (approved):

PR #4497, opened 2025-06-01
chore: Avoid logging a warning when `LOG_LEVEL` is unset

polarathene:
# Description

When `LOG_LEVEL` is not explicitly set, DMS logs a warning early on:

```
WARN  start-mailserver.sh: Log level '' is invalid (falling back to default 'info')
```

This is redundant, so let's configure the fallback earlier and only log this when it has actually been set with an unexpected value.

I've additionally collapsed the series of conditionals against `LOG_LEVEL` to use a regex instead, then inversed the logic so we don't need a redundant branch.

**Reference:**

`LOG_LEVEL` defaults to `info`, where calls to `_log()` filter out by value comparison:

https://github.com/docker-mailserver/docker-mailserver/blob/ea03808c8f8d7ed67a0a9ed6d255e62338c02e9d/target/scripts/helpers/log.sh#L58-L64

For historical PR reference:
- https://github.com/docker-mailserver/docker-mailserver/pull/4323
- https://github.com/docker-mailserver/docker-mailserver/pull/2493

## Type of change

## Checklist

Minor internal change, seems unnecessary to document in the changelog.

### DIFF
--- a/target/scripts/startup/variables-stack.sh
+++ b/target/scripts/startup/variables-stack.sh
@@ -153,7 +153,6 @@ function __environment_variables_general_setup() {
   VARS[FETCHMAIL_PARALLEL]="${FETCHMAIL_PARALLEL:=0}"
   VARS[FETCHMAIL_POLL]="${FETCHMAIL_POLL:=300}"
   VARS[GETMAIL_POLL]="${GETMAIL_POLL:=5}"
-  VARS[LOG_LEVEL]="${LOG_LEVEL:=info}"
   VARS[LOGROTATE_INTERVAL]="${LOGROTATE_INTERVAL:=weekly}"
   VARS[LOGROTATE_COUNT]="${LOGROTATE_COUNT:=4}"
   VARS[LOGWATCH_INTERVAL]="${LOGWATCH_INTERVAL:=none}"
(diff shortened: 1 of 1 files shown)


target/scripts/startup/variables-stack.sh
```diff
@@ -181,21 +180,12 @@ function __environment_variables_general_setup() {
...
-  else
-    local DEFAULT_LOG_LEVEL='info'
-    _log 'warn' "Log level '${LOG_LEVEL}' is invalid (falling back to default '${DEFAULT_LOG_LEVEL}')"
+  VARS[LOG_LEVEL]="${LOG_LEVEL:=info}"
 
-    # shellcheck disable=SC2034
-    VARS[LOG_LEVEL]="${DEFAULT_LOG_LEVEL}"
-    LOG_LEVEL="${DEFAULT_LOG_LEVEL}"
+  if [[ ! ${LOG_LEVEL} =~ ^(trace|debug|info|warn|error)$ ]]; then
+    _log 'warn' "Log level '${LOG_LEVEL}' is invalid (falling back to default: 'info')"
```
casperklein, line 187: Isn't the actual fallback missing?
```suggestion
    _log 'warn' "Log level '${LOG_LEVEL}' is invalid (falling back to default: 'info')"
    VARS[LOG_LEVEL]="info"
    LOG_LEVEL="info"
```

polarathene: Nope, it is set a few lines earlier with the `VARS` assignment. Unless I misunderstood what `${LOG_LEVEL:=info}` means? It should be updating that `LOG_LEVEL` variable when it's unset.

Ohh.... rightttt. I didn't account for when it's an invalid value already assigned 😆 Thank you review process! 🙏

polarathene: I think I should restore the other `VARS[LOG_LEVEL]="${LOG_LEVEL:=info}"` and drop the one from this function then?

```bash
if [[ ! ${LOG_LEVEL:-info} =~ ^(trace|debug|info|warn|error)$ ]]; then
  _log 'warn' "Log level '${LOG_LEVEL}' is invalid (falling back to default: 'info')"
  LOG_LEVEL='info'
fi
```

casperklein: Function wise, that shouldn't make a difference. But if you restore the var, we would have all VARS[XXX]="" in one place.

georglauterbach (approved):

casperklein (dismissed):
One last thing: As the function serves only one purpose, we might rename it to something like: 

`__environment_variables_validate_log_level` or `__environment_variables_check_log_level_value`

Issue #4326, opened 2025-01-23
other: Multiple Domains with SSL via acme lego client


C0D3-M4513R:
### Subject

I would like to configure a not documented mail server use case

### Description

I would like to configure a mailserver for multiple different Domains (e.g. a.com, b.com and c.com), where each cert for mail.{a,b,c}.com is acquired externally (and separately) via acme's lego client.

Currently the only way there seems to be is to use the letsencrypt type and to then manually fake the directory layout. Even then I'm not sure, that multiple different domains are supported, because: 
> Internally, hostname -f will be used to retrieve the FQDN as configured in the below examples.
\- https://docker-mailserver.github.io/docker-mailserver/latest/config/security/ssl/

However this restriction seems to go against the FAQ: https://docker-mailserver.github.io/docker-mailserver/latest/faq/#do-you-support-multiple-domains
It almost seems to me, that you are assuming in that faq entry, that one certificate exists, which spans e.g. mail.example.com, mail.example.de and server.example.org simultaneously. (And I would like to be a little more discrete about which mail that one mailserver is handling)

polarathene:
Search existing issues and discussions on this topic please.

You'll find my responses cover this question multiple times. It's incredibly likely that you have the same misconceptions as those prior to you had.

I would link you, but my system just ran into a snag that I need to resolve (_stuck with only my phone atm, where getting the links via my phone is awkward_).

---

Roughly though, DMS should only operate at one FQDN, having multiple is vanity reasons for clients that wouldn't have that if setup with a proprietary mail service like gmail either.

If I'm mistaken and you can show me where Gmail or Outlook allow customers to provide their own certificates and vanity records to verify those certs, please do share a reference link.

What you want can be done, but DMS does not have any official support as no past request could justify the added complexity. DMS has one hostname (which is all the container permits), that is used for outbound greetings to other mail servers. How does it know which one you should use?

Mail domains DMS manages aren't tied to the mail server itself. That is Tls certs are for the DMS FQDN and you can have DMS receive and send mail for completely unrelated domains, there is no need for certs for that. So how would DMS know what hostname to report itself as if you expect it to present the same vanity FQDN?

Mail clients just need the DMS FQDN configured once, same with DNS. It's not user facing after that. Keep it simple with a single FQDN and cert.

polarathene:
Regarding your concerns for discretion/privacy, find the past issues, one involves shodan info as an example.

[...]

PR #4534, opened 2025-07-30
Update faq.md

GAS85:
Add Information how to enable TLS for Multiple domains based on a #1911

# Description

Fixes #

## Type of change

## Checklist

### DIFF
--- a/docs/content/faq.md
+++ b/docs/content/faq.md
@@ -95,11 +95,132 @@ All files are using the Unix format with `LF` line endings. Please do not use `C
 
 DMS supports multiple domains out of the box, so you can do this:
 
-``` BASH
+```bash
 ./setup.sh email add user1@example.com
 ./setup.sh email add user1@example.de
 ./setup.sh email add user1@server.example.org
 ```
+
+Please remember, you have to setup TLS for multiple domains. Ether you are using the same Certificate for all mail domains, or you can use following script to automate SNI setup with `letsencrypt`.
+
+??? example
+
+    You have to save this script as `user-patches.sh`. Please refer to [How to adjust settings with the `user-patches.sh` script](#how-to-adjust-settings-with-the-user-patchessh-script).
+
+    ```bash
+    #!/bin/bash
+    ##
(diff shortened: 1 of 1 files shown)


georglauterbach:
Thank you very much for the PR :) I think providing information about how to acquire certificates properly is outside the scope of DMS. Moreover, there are many methods to do so - using a shell script with Let'sEncrypt is only one of them.

I'd rather not maintain this in our documentation. But again, thank you very much for taking the time :)

GAS85:
Ok, but at least it shall be a link to the #1911, because it is beyond "yes we support", you have to setup TLS and here starts tricky part that took me some time to find out. I can write it in general without letsencrypt too.

GAS85:
@georglauterbach Just read your message again, and I see that this is misunderstanding. This PR **is not about how to get Certificate**. It is about hot to configure DMS to use multiple Certificates with multiple domains.

georglauterbach:
I see. _Could you please open an issue_ explaining the case first. I understand that your changes focus on SNI, but it would be nice for us maintainers to see what is actually required in terms of changes and what is Let'sEncrypt-specific (and hence, not on DMS). This is why we prefer issues first instead of direct MRs, because we need to maintain the docs and also work on issues that come up with the docs and we hence need to properly evaluate everything first.

GAS85:
The issue is #1911 from 2021, it was not clear how to configure DMS and I found an answer only in tickets discussion. I can create a new one to discuss it if needed.

polarathene:
I fully agree with @georglauterbach here.

While the issue you referenced may not have made it as clear, you will find other issues referencing it where I've commented about why multi-domain certs for DMS is discouraged and doesn't make much sense. We do not want to endorse that in our documentation without a valid use-case beyond vanity reasons, so far no one has provided one.

For reference please see:
- https://github.com/docker-mailserver/docker-mailserver/issues/4326#issuecomment-2611427089
- https://github.com/docker-mailserver/docker-mailserver/pull/4155#issuecomment-2278861770

I am happy to take a doc contribution that better communicates those reference links more tersely, since I have seen support requested / contributed multiple times, but it usually results in an agreement that it was better to configure for a single certificate.

PR #4547, opened 2025-08-10
release: `v15.1.0`

georglauterbach:
# Description

This PR marks the release of v15.1.0 which includes the fix for GPG encryption & signing.

@casperklein @polarathene how long should we wait before releasing this? Do you think a feature freeze is necessary again? I do not think so, but I'd not pressure on not doing it either.

## Checklist

### DIFF
--- a/.github/workflows/generic_vulnerability-scan.yml
+++ b/.github/workflows/generic_vulnerability-scan.yml
@@ -55,7 +55,7 @@ jobs:
           provenance: false
 
       - name: 'Run the Anchore Grype scan action'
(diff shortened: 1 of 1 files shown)


casperklein:
> Do you think a feature freeze is necessary again?

No. Ship it 🚀

polarathene (approved):
The only concern I have is the incomplete Podman Quadlet docs update, but as it's just docs for unofficial support guidance, since we have nothing else pressing to include in this release and nothing significant merged recently to warrant concern go ahead and release 15.1.0 👍

PR #4536, opened 2025-08-02
chore: update base image from Debian 12 to Debian 13

georglauterbach:
# Description

This PR updates our base image from Debian 12 to Debian 13. I have tried to keep the diff as small as possible and only perform necessary changes. Almost all major components are affected, though.

Closes #4447
Fixes #4466
Closes #4512

## Type of change

- [x] Update
- [x] Breaking change
- [x] This change requires a documentation update

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [x] If necessary, I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] **I have added information about changes made in this PR to `CHANGELOG.md`**

PR #4562, opened 2025-08-18
remove: Diffie-Hellman parameters for Postfix & Dovecot

### LINKED ISSUE #4538: update: drop support for DHE params
### Subject

Something else that requires developers attention

### Description

The latest Postfix (on Debian 13) will log the following:

```
postconf: warning: /etc/postfix/main.cf: support for parameter "smtpd_tls_dh1024_param_file" will be removed; instead, do not specify (leave at default)
```

[`smtpd_tls_dh1024_param_file`](https://www.postfix.org/postconf.5.html#smtpd_tls_dh1024_param_file) when not configured will then defer to negotiating standard TLS 1.3 FFDHE params (_presumably also works under TLS 1.2?_). This will be the same behaviour when Postfix drops support entirely.

While with Dovecot the equivalent setting [`ssl_server_dh_file`](https://doc.dovecot.org/2.4.1/core/summaries/settings.html#ssl_server_dh_file) has [been optional since Dovecot 2.3](https://github.com/docker-mailserver/docker-mailserver/pull/2192) so [it is not required either](https://doc.dovecot.org/2.4.1/core/config/ssl.html#ssl-security-settings).

- Presently in DMS we [default to 4096-bit FFDHE params](https://github.com/docker-mailserver/docker-mailserver/tree/e448b5e53e7741b9d4862bdfea2cb5aed72d7543/target/shared) (_standardized via RFC 7919_), which is added via our [`Dockerfile`](https://github.com/docker-mailserver/docker-mailserver/blob/e448b5e53e7741b9d4862bdfea2cb5aed72d7543/Dockerfile#L210-L214) to default [Postfix](https://github.com/docker-mailserver/docker-mailserver/blob/e448b5e53e7741b9d4862bdfea2cb5aed72d7543/target/postfix/main.cf#L48) + [Dovecot](https://github.com/docker-mailserver/docker-mailserver/blob/e448b5e53e7741b9d4862bdfea2cb5aed72d7543/target/dovecot/10-ssl.conf#L48-L51) configured file paths.
- We also have support for users to provide their own alternative custom DH params file to use instead, which our [Dovecot/Postfix setup helper](https://github.com/docker-mailserver/docker-mailserver/blob/e448b5e53e7741b9d4862bdfea2cb5aed72d7543/target/scripts/helpers/ssl.sh#L3-L18)  provides support for (_assumed rarely used?_).
- [Related docs/tests](https://github.com/docker-mailserver/docker-mailserver/pull/2192) for reference.

---

Since support will be dropped in future by Postfix, and ideally most connections to modern mail servers are using TLS 1.3 by now, it probably makes sense to consider dropping this default support?

We can still provide documentation for users that need this functionality, which they can add by providing their own DH params file (_[standard RFC 7919 files](https://github.com/internetstandards/dhe_groups) provides these with content matching the [associated entries from the RFC document](https://datatracker.ietf.org/doc/html/rfc7919#appendix-A.1) - these are the same DH params used by TLS 1.3 for DHE cipher-suites_) and our `postfix-main.cf` + `dovecot.cf` config overrides/additions support. That will work for them until software like Postfix drops support, then they'll need to pursue further alternatives beyond what DMS can assist with.

Might be beneficial to verify if DHE ciphersuites are still negotatied for TLS 1.2 (_as is expected for Postfix at least when paired with OpenSSL 3.x_) on [port 25](https://github.com/docker-mailserver/docker-mailserver/blob/e448b5e53e7741b9d4862bdfea2cb5aed72d7543/test/tests/parallel/set2/tls_cipherlists.bats#L227) and [any other port](https://github.com/docker-mailserver/docker-mailserver/blob/e448b5e53e7741b9d4862bdfea2cb5aed72d7543/test/tests/parallel/set2/tls_cipherlists.bats#L217) as we allow a wider range for port 25 with TLS 1.2.

---

**Actionables:**
- [ ] Remove related DH content in DMS (_bulk of which should be linked above_).
- [ ] Changes may require an update to [our TLS tests](https://github.com/docker-mailserver/docker-mailserver/blob/e448b5e53e7741b9d4862bdfea2cb5aed72d7543/test/tests/parallel/set2/tls_cipherlists.bats#L217), which may be complicated if Dovecot and Postfix diverge in behaviour for DHE support.
- [ ] Update documentation to provide guidance for manually adding back support for DHE (_should only be relevant to deployments with legacy clients, TLS 1.3 has been around since 2018 and TLS 1.2 from 2008 still has other ciphers available beyond DHE_). Ideally this is a simple `compose.yaml` snippet that would be fairly self-explanatory.
- [ ] Reference updated documentation in Changelog (_with versioned docs link that will become valid upon release_).

georglauterbach:
# Description

This patch removes all custom DH parameters from DMS.

Fixes #4538

## Type of change

## Checklist


polarathene:
**TL;DR UPDATE:** Dovecot `ssl_cipher_list` needs to be adjusted accordingly to drop the DHE cipher suites too (_`testssl.sh` appears to [misreport supported ciphersuites as a result](https://github.com/testssl/testssl.sh/issues/2883#issuecomment-3261828200)_), our tests caught the failure thankfully (_although it wasn't obvious as to why unexpected suites were dropped_)

https://github.com/docker-mailserver/docker-mailserver/blob/5f4e868c541b142c705a643315ba8828d1836660/target/scripts/helpers/ssl.sh#L145-L146

---

Test failure in `test/tests/parallel/set2/tls_cipherlists.bats` is expected due to the removal of the FFDHE params, but these two particular tests have failed a bit unexpectedly:

```
[Security] (TLS) (cipher lists) 'TLS_LEVEL=intermediate' + RSA

# -- output differs --
# expected : "ECDHE-RSA-CHACHA20-POLY1305 ECDHE-RSA-AES128-GCM-SHA256 ECDHE-RSA-AES256-GCM-SHA384 DHE-RSA-AES128-GCM-SHA256 DHE-RSA-AES256-GCM-SHA384 ECDHE-RSA-AES128-SHA256 ECDHE-RSA-AES256-SHA384 DHE-RSA-AES128-SHA256 DHE-RSA-AES256-SHA256"
# actual   : "ECDHE-RSA-CHACHA20-POLY1305 ECDHE-RSA-AES128-GCM-SHA256 ECDHE-RSA-AES256-GCM-SHA384"
# --


[Security] (TLS) (cipher lists) 'TLS_LEVEL=intermediate' + ECDSA with RSA fallback
# -- output differs --
# expected : "ECDHE-ECDSA-CHACHA20-POLY1305 ECDHE-RSA-CHACHA20-POLY1305 ECDHE-ECDSA-AES128-GCM-SHA256 ECDHE-RSA-AES128-GCM-SHA256 ECDHE-ECDSA-AES256-GCM-SHA384 ECDHE-RSA-AES256-GCM-SHA384 DHE-RSA-AES128-GCM-SHA256 DHE-RSA-AES256-GCM-SHA384 ECDHE-ECDSA-AES128-SHA256 ECDHE-RSA-AES128-SHA256 ECDHE-RSA-AES256-SHA384 ECDHE-ECDSA-AES256-SHA384 DHE-RSA-AES128-SHA256 DHE-RSA-AES256-SHA256"
# actual   : "ECDHE-ECDSA-CHACHA20-POLY1305 ECDHE-RSA-CHACHA20-POLY1305 ECDHE-ECDSA-AES128-GCM-SHA256 ECDHE-RSA-AES128-GCM-SHA256 ECDHE-ECDSA-AES256-GCM-SHA384 ECDHE-RSA-AES256-GCM-SHA384"
# --
```

The `actual` has additionally excluded some CBC suites:

```
ECDHE-ECDSA-AES128-SHA256 ECDHE-RSA-AES128-SHA256 ECDHE-RSA-AES256-SHA384 ECDHE-ECDSA-AES256-SHA384
```

All dropped DHE ciphers were `DHE-RSA` (_RSA cert for authentication_), expected as there is none for ECDSA certs.

AES-CBC cipher is not an AEAD cipher, so only AEAD ciphers remain after this change. Apart from slight priority order/sort of the cipher suites between `modern` and `intermediate` `TLS_LEVEL` setting, these two profiles are now equivalent (**EDIT:** Only for Dovecot).

---

All related test failures were occurring at `_verify_cipherlists()` for the IMAP port 143, so this is Dovecot specific, confirming the difference to Postfix (_which is leveraging the OpenSSL 3.0 support to continue supporting the DHE cipher suites_).

https://github.com/docker-mailserver/docker-mailserver/blob/5f4e868c541b142c705a643315ba8828d1836660/test/tests/parallel/set2/tls_cipherlists.bats#L132-L133

Since Postfix and Dovecot now diverge, we will need to either:
- **A:** Update the tests with separate expected ciphersuite expectations between the two.
- **B:** Configure Postfix to exclude the extra ciphers (shouldn't be necessary)
- **C:** Retain the DH config for Dovecot. It's less likely that these ciphers are needed for Dovecot clients however but it would retain parity between Postfix and Dovecot 😅

For option **A** we would only have the following comparisons differing:

https://github.com/docker-mailserver/docker-mailserver/blob/5f4e868c541b142c705a643315ba8828d1836660/test/tests/parallel/set2/tls_cipherlists.bats#L208-L210

https://github.com/docker-mailserver/docker-mailserver/blob/5f4e868c541b142c705a643315ba8828d1836660/test/tests/parallel/set2/tls_cipherlists.bats#L216-L218

Which needs to be used in `_verify_cipherlists()` only here:

https://github.com/docker-mailserver/docker-mailserver/blob/5f4e868c541b142c705a643315ba8828d1836660/test/tests/parallel/set2/tls_cipherlists.bats#L132-L138

This could be supported similar to how the port 25 variants are already done, using a suffix as a 2nd option for the lookup key:

https://github.com/docker-mailserver/docker-mailserver/blob/5f4e868c541b142c705a643315ba8828d1836660/test/tests/parallel/set2/tls_cipherlists.bats#L174-L181

Doing so with a suffix like `_dovecot` would work but will need a similar workaround for the the ECDSA variants that have no difference:

https://github.com/docker-mailserver/docker-mailserver/blob/5f4e868c541b142c705a643315ba8828d1836660/test/tests/parallel/set2/tls_cipherlists.bats#L212-L214

https://github.com/docker-mailserver/docker-mailserver/blob/5f4e868c541b142c705a643315ba8828d1836660/test/tests/parallel/set2/tls_cipherlists.bats#L229-L232

---

**UPDATE:** I've since found that the `ECDHE` mentioned cipher suites still negotiate successfully with `openssl` if choosing to use that ciphersuite directly:

```bash
timeout 1 openssl s_client -tls1_2 -cipher ECDHE-RSA-AES256-SHA384 -connect mail.example.test:143 -starttls imap -CAfile /tmp/tls/ca-cert.pem
```

The failure appears to be [specific to `testssl.sh` itself and how it was handling server preference selection](https://github.com/testssl/testssl.sh/issues/2883#issuecomment-3261828200), where it bailed as soon as it came across one of the `DHE` cipher suites we still have listed for `ssl_cipher_list` in `10-ssl.conf`, which Dovecot mistakenly communicated / implied that it supported negotiating those ciphersuites with the connecting client 😓 (_technically not a bug per se, as it reflects what any client would do, [like OpenSSL when negotiating a compatible cipher against server preference](https://github.com/testssl/testssl.sh/issues/2883#issuecomment-3263397756)_)

https://github.com/docker-mailserver/docker-mailserver/blob/5f4e868c541b142c705a643315ba8828d1836660/target/scripts/helpers/ssl.sh#L145-L146

https://github.com/docker-mailserver/docker-mailserver/blob/5f4e868c541b142c705a643315ba8828d1836660/target/dovecot/10-ssl.conf#L60-L61

I've been thinking we should disable server preference for some time now (_at least with `TLS_LEVEL=modern` which uses all AEAD ciphers, but `TLS_LEVEL=intermediate` and Postfix port 25 does not.._). **EDIT:** As noted in next paragraph, client preference is still affected too.

I think we may have seen some bug reports potentially related to this [selection order "bug"](https://github.com/testssl/testssl.sh/issues/2883#issuecomment-3263397756) in the past but the errors reported weren't helpful at identifying that 😓 (_the problem of misconfigured server service like Dovecot here [also affects negotiation with client preference](https://github.com/testssl/testssl.sh/issues/2883#issuecomment-3264297477)_)

---

Beyond the adjustment to the tests, we still need to consider docs. Which I'll address when I return to this (_I'll apply any needed changes at that point too_). Spent too much troubleshooting the above today 😝

georglauterbach:
> Since Postfix and Dovecot now diverge, we will need to either:

Additional question before: Should we try to somehow update to Debian 13 first and then decide on this, if possible? We are currently seeing the Debian 12 / Dovecot 2.3 behavior - maybe we'd be having double work if we change it now, update, and then have to change it again.

polarathene:
> Should we try to somehow update to Debian 13 first and then decide on this, if possible?

That's a good point. I don't think much will change though.

Given that Postfix will continue to provide DHE cipher suite support via the OpenSSL 3.0 fallback they have (_which seems to be Postfix specific_), and Dovecot only works by ensuring `ssl_cipher_list` excludes the DHE cipher suites, we could just retain configuration for DHE support in Dovecot?

I think the ability to customize can still be dropped since there shouldn't be a need to keep that, and we can update the docs for anyone that would want that.

georglauterbach:
> Given that Postfix will continue to provide DHE cipher suite support via the OpenSSL 3.0 fallback they have (_which seems to be Postfix specific_), and Dovecot only works by ensuring `ssl_cipher_list` excludes the DHE cipher suites, we could just retain configuration for DHE support in Dovecot?

Sounds good 👍🏼 
 
> I think the ability to customize can still be dropped since there shouldn't be a need to keep that, and we can update the docs for anyone that would want that.

Agreed 👍🏼

I will be tackler some smaller issues that are still to be done before we have a more thorough version of #4536. If you can tackle this PR, I'd be very grateful.

polarathene:
> If you can tackle this PR, I'd be very grateful.

Yes no worries, I'll take care of it tomorrow if I can spare the time 👍

**UPDATE:** Will be a bit longer as I need to take some time to recover from an injury 😅

---

**NOTE:** Latest push is still WIP.

georglauterbach:
@polarathene I'd like to work on the Debian 13 changes again. If you cannot pick this up, I can give it a try. Just tell me what you'd like to see implemented.

polarathene:
Unfortunately past me didn't leave a checklist when I got interrupted by an injury, so what I intended to tackle on the docs front is unclear (_I lost any local work associated to this PR when my system crashed again_).

I did make a mention about the test failure at least, we need to exclude the DHE cipher defaults from Dovecot, otherwise even without a configured FFDHE params file it will still accept negotiating these cipher suites which will fail. The alternative is to keep DHE params, which may be useful for legacy IMAP/POP3 clients, but I did update the docs for users that would require that support.

I'll give this a glance over soon so the PR can unblock you.

georglauterbach:
I'll have some days off that I'd like to dedicate to DMS soon. @polarathene can we get this PR ready for review and merge it to unblock me? :)

polarathene:
> can we get this PR ready for review and merge it to unblock me? :)

I can try to squeeze in some time for this over the next few days. Juggling quite a bit atm.

Just to be clear I don't believe this PR should be blocking you? IIRC we were looking at dropping DH params which this PR does focus on but Dovecot 2.4 was fine to keep DH params present.

Was there something else concerning as a blocker?

georglauterbach:
> Just to be clear I don't believe this PR should be blocking you? IIRC we were looking at dropping DH params which this PR does focus on but Dovecot 2.4 was fine to keep DH params present.
> 
> Was there something else concerning as a blocker?

Sorry for the late response; I was way busier than I thought I'd be. I was under the impression this PR was blocking the Debian 13 PR. Never mind me then, I'll tend to it in the next days.

PS: I wish you a happy new year @casperklein @polarathene! :)

polarathene:
Just for future reference, [this OpenSSL PR](https://github.com/openssl/openssl/pull/24551) has since been merged upstream and may be relevant for a future OpenSSL release.

georglauterbach:
@polarathene I have revisited this PR, rebased the branch and applied a bit of feedback. Can you proceed from here or should I take over? I'd be very nice to have your support here.

PR #4565, opened 2025-09-03
Delete user without prompt but keep mailbox data

dano19:
# Description

Added option to allow to delete email but do not delete email data without prompt using "-n", default and "-y" remains as is.
Its required for UI I made and I want to allow admin to delete email without deleting data.
https://gitlab.com/div-solutions/docker-mailserver-ui

## Type of change

- [X ] Improvement (non-breaking change that does improve existing functionality)

## Checklist

### DIFF
--- a/target/bin/delmailuser
+++ b/target/bin/delmailuser
@@ -21,6 +21,8 @@ function _main() {
 
     if [[ ${MAILDEL} -eq 1 ]]; then
       _remove_maildir "${MAIL_ACCOUNT}"
+    elif [[ ${MAILDEL} -eq 2 ]]; then
+      _log 'info' "Mailbox data explicitly kept (using -n)."
     else
       _log 'info' "The mailbox data will not be deleted."
     fi
@@ -48,6 +50,8 @@ ${ORANGE}USAGE${RESET}
 ${ORANGE}OPTIONS${RESET}
     -y
         Skip prompt by approving to ${LWHITE}delete all mail data${RESET} for the account(s).
+    -n
+        Skip prompt by declining to ${LWHITE}delete all mail data${RESET} for the account(s).
 
     ${BLUE}Generic Program Information${RESET}
         help       Print the usage information.
@@ -65,6 +69,10 @@ ${ORANGE}EXAMPLES${RESET}
         Delete the two mail accounts requested, their associated data and
         delete the mailbox data for both accounts without asking.
 
+    ${LWHITE}./setup.sh email del -n user@example.com extra-user@example.com${RESET}
+        Delete the two mail accounts requested, their associated data and DO NOT
+        delete the mailbox data for both accounts without asking.
+
 ${ORANGE}EXIT STATUS${RESET}
     Exit status is 0 if command was successful. If wrong arguments are provided
     or arguments contain errors, the script will exit early with exit status 1.
@@ -73,12 +81,14 @@ ${ORANGE}EXIT STATUS${RESET}
 }
 
 function _parse_options() {
-  while getopts ":yY" OPT; do
+  while getopts ":yYnN" OPT; do
     case "${OPT}" in
       ( 'y' | 'Y' )
         MAILDEL=1
         ;;
-
+      ( 'n' | 'N' )
+        MAILDEL=2
+        ;;
       ( * )
         __usage
         _exit_with_error "The option '${OPT}' is unknown"


polarathene:
This doesn't seem to be necessary?

I think it would be better to avoid the issue entirely by ensuring you run commands non-interactively, without needing to patch support in per command as-needed 😅

---

You want the default logic of this command, but your concern is the interactive prompt when you call the command non-interactively?:

```console
$ setup email del john.doe@example.test
Do you want to delete the mailbox data as well (removing all mails)? [y/N]
```

This script handles the prompt by reading from stdin, so you can just pipe in the `n` input into the commands stdin and it'll be processed:

```console
echo 'n' | setup email del john.doe@example.test
2025-09-04 00:30:46+00:00 INFO  delmailuser: The mailbox data will not be deleted.
2025-09-04 00:30:46+00:00 INFO  delmailuser: 'john.doe@example.test' and associated data (aliases, quotas) deleted
```

Or from the docker host when calling into the container you can wrap that like this:

```console
docker exec dms bash -c 'echo "n" | setup email del john.doe@example.test'
```

Since I omitted `-it` from the docker command, the `echo` workaround isn't actually necessary either, but you will find the equivalent for `docker compose exec` does need it unless you add the required opt-out flags when using that to get non-interactive shell.

## Sending a command from outside of the container via the docker host

This is for a quick reference, if you're interested in the details to better understand this advice, I've covered that in the next section.

```console
$ docker run --rm \
  --env ACCOUNT_PROVISIONER=FILE \
  --volume ./postfix-accounts.cf:/tmp/docker-mailserver/postfix-accounts.cf:ro \
  ghcr.io/docker-mailserver/docker-mailserver:15.1.0 \
    setup email del john.doe@example.test

2025-09-04 01:06:43+00:00 INFO  delmailuser: The mailbox data will not be deleted.
sed: cannot rename /tmp/docker-mailserver/sedauD0Xw: Device or resource busy
2025-09-04 01:06:43+00:00 ERROR delmailuser: 'john.doe@example.test' could not be deleted
2025-09-04 01:06:43+00:00 ERROR delmailuser: Aborting
```

`docker compose` equivalent (for both `run` and `exec`) are opt-out instead of opt-in for interactive terminal via `--no-TTY --interactive=false`:

```yaml
services:
  dms:
    image: ghcr.io/docker-mailserver/docker-mailserver:15.1.0
    volumes:
      - ./postfix-accounts.cf:/tmp/docker-mailserver/postfix-accounts.cf:ro
```

```console
# Only `--no-TTY` => Prompt is hidden but otherwise remains interactive
# Only `--interactive=false` => Prompt appears and hangs as no input can actually be received
# Both opt-out flags => No prompt shown and no stdin to receive input, proper non-interactive shell

$ docker compose run --rm --no-TTY --interactive=false \
  --env ACCOUNT_PROVISIONER=FILE \
  dms \
    setup email del john.doe@example.test

2025-09-04 01:06:43+00:00 INFO  delmailuser: The mailbox data will not be deleted.
sed: cannot rename /tmp/docker-mailserver/sedauD0Xw: Device or resource busy
2025-09-04 01:06:43+00:00 ERROR delmailuser: 'john.doe@example.test' could not be deleted
2025-09-04 01:06:43+00:00 ERROR delmailuser: Aborting
```

---

## `docker compose exec` vs `docker exec` default flags

When executing a command in a container from the host however, you'd typically use one of these two CLI commands with the flags `-it` to get an interactive tty. The two commands however differ in defaults for `--interactive` and `--tty`.

To better demonstrate this I've provided a bash script below and reproduction commands with output to compare/try. This is a difference worth being aware about when it comes to calling interactive commands in containers :)

### Relevance of flags to the CLI prompt logic

Using `-i` will enable interactive prompting, but `-t` is required to actually see that prompt (_due to how the `read` command used for the input + prompt logic works_). When both are disabled, only then will the [prompt (_via `read` command_)](https://github.com/docker-mailserver/docker-mailserver/blob/5f4e868c541b142c705a643315ba8828d1836660/target/bin/delmailuser#L94) fallback to the implicit default.
- **Without `--tty`**, the [`-p` of the `read` command will be ignored](https://github.com/bminor/bash/blob/a8a1c2fac029404d3f42cd39f5a20f24b6e4fe4b/builtins/read.def#L487-L489) so no prompt text appears.. while anything from our logs via `echo` continues to output_).
- While **only `--tty` without `--interactive`**, the prompt is output but your containers stdin is not connected into the container, so anything you input appears unresponsive/ignored.
- **With `--interactive` only**, you'd instead trigger that same without `--tty` condition hiding the prompt, but as stdin was kept open through the pipe, `read` is still waiting on input to continue. While normally in a piped command this would make sense such as `echo 'hello' | read -p 'Showing this would be redundant'`, when it's an interactive pipe you'd want that TTY to see the prompt text 😅

## Reference - Reproduction Example

You can run the following bash script mounted into a container (_eg: `-v ./detect.sh:/usr/local/bin/detect:ro`_) with both `docker` and `docker compose` using either subcommand `exec` or `run`:

```bash
#!/bin/bash

# Detected output:
# https://stackoverflow.com/questions/911168/how-can-i-detect-if-my-shell-script-is-running-through-a-pipe/30811869#30811869
function detect_output() {
  [[ -t 1 ]] && echo 'STDOUT is attached to TTY'
  [[ -p /dev/stdout ]] && echo 'STDOUT is attached to a pipe'
  [[ ! -t 1 && ! -p /dev/stdout ]] && echo 'STDOUT is attached to a redirection'
}

# Detect input (stdin vs `bash -i`):
# https://stackoverflow.com/questions/65265637/what-does-tty-s-have-to-do-with-detecting-whether-a-shell-is-interactive/65265888#65265888
function detect_input() {
  # Can input actually be received?:
  if [[ -t 0 ]]; then
    echo 'STDIN is attached to TTY'
  elif [[ -p /dev/stdin ]]; then
    echo 'STDIN is attached to a pipe'
  else
    echo 'Missing STDIN, cannot receive input?'
  fi

  # Bash (only relevant if the script was run with `bash -i` flag):
  case $- in *i*) echo "Shell is INTERACTIVE";; *) echo "Shell is NON-INTERACTIVE";; esac
}

detect_output
detect_input
```

### `docker exec`

The defaults for the Docker CLI `run` / `exec` commands require opt-in to both `--interactive / -i` + `--tty / -t`:
- https://docs.docker.com/reference/cli/docker/container/run/#interactive
- https://docs.docker.com/reference/cli/docker/container/run/#tty

```console
$ docker run --rm -d -v ./detect.sh:/usr/local/bin/detect:ro --name shell-test fedora:42 sleep 60 \
  && echo -e '\nNeither:' && docker exec shell-test detect \
  && echo -e '\nBoth:' && docker exec -it shell-test detect \
  && echo -e '\nOnly -i:' && docker exec -i shell-test detect \
  && echo -e '\nOnly -t:' && docker exec -t shell-test detect

Neither:
STDOUT is attached to a pipe
Missing STDIN, cannot receive input?
Shell is NON-INTERACTIVE

Both:
STDOUT is attached to TTY
STDIN is attached to TTY
Shell is NON-INTERACTIVE

Only -i:
STDOUT is attached to a pipe
STDIN is attached to a pipe
Shell is NON-INTERACTIVE

Only -t:
STDOUT is attached to TTY
STDIN is attached to TTY
Shell is NON-INTERACTIVE
```

So `read -p` in this scenario only becomes non-interactive in the "Neither" variant.

STDIN via pipe or TTY present will make it handled interactively.

### `docker compose exec`

Docker Compose equivalent commands [`exec`](https://docs.docker.com/reference/cli/docker/compose/exec/) + [`run`](https://docs.docker.com/reference/cli/docker/compose/run/) unlike the regular Docker CLI equivalent commands differ in interactive defaults (_presumably for legacy reasons_).

As such you need to opt-out with `docker compose run` with flags `--no-TTY --interactive=false` / `-T -i=false`, ~~but `docker compose exec` lacks the ability to control interactive like `docker exec`, only the pseudo-TTY flag is supported~~ (_which can also be configured to a different default with the `tty` key in `compose.yaml`](https://docs.docker.com/reference/compose-file/services/#tty)_).

**UPDATE:** Despite missing from the [official documentation `docker compose exec` page](https://docs.docker.com/reference/cli/docker/compose/exec/), the `-i` / `--interactive` flag is actually [implemented for `docker compose exec`](https://github.com/docker/compose/issues/10418#issuecomment-1491426721) as can be seen in the below output.

<details>
<summary>Equivalent Docker Compose exec example (click to view)</summary>

Try this basic `compose.yaml` like this with the earlier `detect.sh` bash script file and you'll find the following commands and output.

```yaml
services:
  shell-test:
    image: fedora:42
    command: sleep 5
    volumes:
      - ./detect.sh:/usr/local/bin/detect:ro
```

```console
# Every result is the same since neither `-i` or `-t` is opt-in, but instead opt-out:
$ docker compose up -d \
  && echo -e '\nNeither:' && docker compose exec shell-test detect \
  && echo -e '\nBoth:' && docker compose exec -it shell-test detect \
  && echo -e '\nOnly -i:' && docker compose exec -i shell-test detect \
  && echo -e '\nOnly -t:' && docker compose exec -t shell-test detect \
  && docker compose down

Neither:
STDOUT is attached to TTY
STDIN is attached to TTY
Shell is NON-INTERACTIVE

Both:
STDOUT is attached to TTY
STDIN is attached to TTY
Shell is NON-INTERACTIVE

Only -i:
STDOUT is attached to TTY
STDIN is attached to TTY
Shell is NON-INTERACTIVE

Only -t:
STDOUT is attached to TTY
STDIN is attached to TTY
Shell is NON-INTERACTIVE
```

```console
# Actual equivalent to `docker exec` results via opt-out flags `-T` + `-i=false`:
$ docker compose up -d \
  && echo -e '\nNeither enabled:' && docker compose exec -T -i=false shell-test detect \
  && echo -e '\nBoth (defaults):' && docker compose exec shell-test detect \
  && echo -e '\nOnly STDIN:' && docker compose exec -T shell-test detect \
  && echo -e '\nOnly TTY:' && docker compose exec -i=false shell-test detect \
  && docker compose down

Neither enabled:
STDOUT is attached to a pipe
Missing STDIN, cannot receive input?
Shell is NON-INTERACTIVE

Both (defaults):
STDOUT is attached to TTY
STDIN is attached to TTY
Shell is NON-INTERACTIVE

Only STDIN:
STDOUT is attached to a pipe
STDIN is attached to a pipe
Shell is NON-INTERACTIVE

Only TTY:
STDOUT is attached to TTY
STDIN is attached to TTY
Shell is NON-INTERACTIVE
```

</details>

polarathene:
Oh, I just saw https://gitlab.com/div-solutions/docker-mailserver-ui is referring to k8s.

Kubernetes is rather slow at resolving issues with landing capabilities that Docker has, so if it lacks the equivalent ability to run a command non-interactively perhaps the support in DMS is necessary? @cfis or @georglauterbach can probably chime in on that?

We could technically infer lack of stdin via `[[ ! -t 0 && ! -p /dev/stdin ]]`? Could be a helper method for implicit opt-out of interactive prompts?

I'll wait on feedback from other maintainers.

cfis:
I believe you want:

```
kubectl exec <pod-name> -- <command>
```

Add `-it` for an interactive console.

dano19:
Thanks for your reply @polarathene. 

I can use 'echo "n" | setup email del john.doe@example.test'

My process of thinking was, because my code to execute setup though k8s client in .net looks like this:
```
await Config.ClusterClient
    .WebSocketNamespacedPodExecAsync(Config.MailServerPod.Metadata.Name, Config.MailServerPodNamespace, new List<string>()
        {
            "sh", "-c", $"/usr/local/bin/setup {command}"
        },
    Config.MailServerPod.Spec.Containers[0].Name)
```
so if there is -y, why not -n, doesn't impact feature in any way except adding new parameter for promptless deletion 😄 

If this is too much of an issue, I will just modify my execution function to insert the echo n.

Thanks

polarathene:
I'm not against the change, I was just curious if you could run the command non-interactively without needing that support whenever a command has an interactive prompt.

I'll leave the PR up to @casperklein or @georglauterbach to decide if we should include flags for skipping prompts to use defaults. It's probably still a valid use-case, you're just the first user I've seen bring it up :)

georglauterbach:
I am fine with this change. @casperklein what do you make out of this?

casperklein:
`echo 'n' | setup email del john.doe@example.test` works, but an extra parameter is the cleaner solution.

PR #4568, opened 2025-09-09
fix: Postfix: `disable_dns_lookups` => `smtp_dns_support_level`

georglauterbach:
see <https://www.postfix.org/postconf.5.html#disable_dns_lookups>

# Description

Replaces the deprected option `disable_dns_lookups` with `smtp_dns_support_level`. This option was deprecated since Postfix 2.11.

## Type of change

## Checklist

### DIFF
--- a/target/amavis/postfix-amavis.cf
+++ b/target/amavis/postfix-amavis.cf
@@ -7,7 +7,7 @@ smtp-amavis     unix    -       -       n       -       2       smtp
   -o syslog_name=postfix/$service_name
   -o smtp_data_done_timeout=1200
   -o smtp_send_xforward_command=yes
-  -o disable_dns_lookups=yes
+  -o smtp_dns_support_level=disabled
   -o max_use=20
   -o smtp_tls_security_level=none
   -o smtp_tls_wrappermode=no
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ All notable changes to this project will be documented in this file. The format 
     - `step` CLI from `0.28.2` to [`0.28.7`](https://github.com/smallstep/cli/releases/tag/v0.28.7))
 - **Dovecot**
   - Updated the FTS plugin Xapian from `1.9` to [`1.9.1`](https://github.com/grosjo/fts-xapian/releases/tag/1.9.1) which adds Dovecot 2.4 compatibility ([#4557](https://github.com/docker-mailserver/docker-mailserver/pull/4557))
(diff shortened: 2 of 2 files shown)


casperklein (approved):

target/amavis/postfix-amavis.cf
```diff
@@ -7,7 +7,7 @@ smtp-amavis     unix    -       -       n       -       2       smtp
   -o syslog_name=postfix/$service_name
   -o smtp_data_done_timeout=1200
   -o smtp_send_xforward_command=yes
-  -o disable_dns_lookups=yes
+  -o smtp_dns_support_level=disabled
```
polarathene, line 10: Just to reference the Postfix docs on [`smtp_dns_support_level=disabled`](https://www.postfix.org/postconf.5.html#smtp_dns_support_level):

> _Disable DNS lookups. No MX lookups are performed and hostname to address lookups are unconditionally "native"._
> _This setting is not appropriate for hosts that deliver mail to the public Internet. Some obsolete how-to documents recommend disabling DNS lookups in some configurations with `content_filters`._
> _This is no longer required and strongly discouraged._

and the prior (now deprecated) [`disable_dns_lookups=yes`](https://www.postfix.org/postconf.5.html#disable_dns_lookups):

> Disable DNS lookups in the Postfix SMTP and LMTP clients.
> - When disabled, hosts are looked up with the `getaddrinfo()` system library routine which normally also looks in `/etc/hosts`.
> - As of Postfix 2.11, this parameter is deprecated; use `smtp_dns_support_level` instead.

This setting would be for outbound SMTP traffic, so for when Amavis is sending back to Postfix, where DNS would be redundant 👍 

---

Within our project, I did add some commentary about this for an [`SMTP_ONLY=1` feature test](https://github.com/docker-mailserver/docker-mailserver/pull/3018#discussion_r1083416192), but the test itself seemed borked, so we don't bother with testing it in the current state:

https://github.com/docker-mailserver/docker-mailserver/blob/f343f7d960d2c09ab428c34c108ea768919676c8/test/tests/parallel/set3/mta/smtponly.bats#L28-L30

In that case, it seems it was to avoid external DNS record lookups, and was about a different/related Postfix DNS setting.

polarathene (commented):
No need to pay much attention to this review comment. It's only to provide some extra context for traceability.

Originally part of a larger PR with request to extract to separate one:
https://github.com/docker-mailserver/docker-mailserver/pull/4536#pullrequestreview-3082614980

PR #4577, opened 2025-09-22
docs: Update dkim_dmarc_spf.md

audioscavenger:
page update: dkim_dmarc_spf
* added comments to the dkim_signing.conf example
* updated file paths for other dkim_signing.conf examples with what the `setup config dkim domain`  command actually use
* added a quick DKIM rotation 3-steps example

# Description

Fixes #

## Type of change

## Checklist

### DIFF
--- a/docs/content/config/best-practices/dkim_dmarc_spf.md
+++ b/docs/content/config/best-practices/dkim_dmarc_spf.md
@@ -48,6 +48,12 @@ DKIM requires a public/private key pair to enable **signing (_via private key_)*
 
     DMS does not provide any automation or support for key rotation, [nor is it likely to provide a notable security benefit][gh-discussion::dkim-key-rotation-expiry] to the typical small scale DMS deployment.
 
+    Here is a quick DKIM Rotation example with minimal downtime (due to restart of dms):
+    1. generate new keys with new selector: `dmss config dkim domain ${domain} selector new_selector [keytype ${keytype} keysize ${keysize}]`
+    2. create new TXT entries 'new_selector._domainkey' with the content of the generated `rspamd/dkim/*-public.dns.txt keys`
+    3. modify the selector name in `rspamd/override.d/dkim_signing.conf` and restart dms
+
+
 ### Generating Keys
 
 You'll need to repeat this process if you add any new domains.
@@ -161,24 +167,34 @@ DKIM is currently supported by either OpenDKIM or Rspamd:
 
         enabled = true;
 
+        # If false, messages from authenticated users are not selected for signing
         sign_authenticated = true;
+        # If false, messages from local networks are not selected for signing
         sign_local = true;
 
+        # Domain to use for ARC signing: can be "header" (MIME From), "envelope" (SMTP From), "recipient" (SMTP To), "auth" (SMTP username) or directly specified domain name
         use_domain = "header";
-        use_redis = false; # don't change unless Redis also provides the DKIM keys
+        # don't change unless Redis also provides the DKIM keys
+        use_redis = false;
+        # Whether to normalise domains to eSLD
         use_esld = true;
-        check_pubkey = true; # you want to use this in the beginning
+        # you want to use this in the beginning
+        check_pubkey = true;
 
+        # global DKIM-selector: this is critical and must match a TXT entry called selector._domainkey in ALL of your domains
         selector = "mail";
+        
         # The path location is searched for a DKIM key with these variables:
         # - `$domain` is sourced from the MIME mail message `From` header
         # - `$selector` is configured for `mail` (as a default fallback)
-        path = "/tmp/docker-mailserver/dkim/keys/$domain/$selector.private";
+        # Update the keytype=rsa and keysize=2048 to the values you use for your keys
+        #path = "/tmp/docker-mailserver/dkim/keys/$domain/$selector.private";  # eSLD
+        path = "/tmp/docker-mailserver/rspamd/dkim/rsa-2048-$selector-$domain.private.txt";
 
         # domain specific configurations can be provided below:
         domain {
             example.com {
-                path = "/tmp/docker-mailserver/rspamd/dkim/mail.private";
+                path = "/tmp/docker-mailserver/rspamd/dkim/rsa-2048-$selector-example.com.private.txt";
                 selector = "mail";
             }
         }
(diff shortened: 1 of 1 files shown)


docs/content/config/best-practices/dkim_dmarc_spf.md
```diff
@@ -48,6 +48,12 @@ DKIM requires a public/private key pair to enable **signing (_via private key_)*
 
     DMS does not provide any automation or support for key rotation, [nor is it likely to provide a notable security benefit][gh-discussion::dkim-key-rotation-expiry] to the typical small scale DMS deployment.
 
+    Here is a quick DKIM Rotation example with minimal downtime (due to restart of dms):
+    1. generate new keys with new selector: `dmss config dkim domain ${domain} selector new_selector [keytype ${keytype} keysize ${keysize}]`
+    2. create new TXT entries 'new_selector._domainkey' with the content of the generated `rspamd/dkim/*-public.dns.txt keys`
+    3. modify the selector name in `rspamd/override.d/dkim_signing.conf` and restart dms
```
polarathene, line 54: `dmss` is not something we offer, did you mean `setup`? (_if run outside of a container you could use `setup.sh`, or just `docker compose exec <service name> setup config dkim ...`_)

This example is Rspamd specific focused, but is mostly focused on the same steps to produce a new DKIM key, with instructions to update the DNS and related OpenDKIM/Rspamd config file.

There was a related GH discussion about DKIM key rotation you chimed in recently and I responded to you there (_and this area of the docs actually links to_) that for most of our users, rotating keys shouldn't be needed.

Is this still something you see value in documenting as a user? If so what about it in particular?

docs/content/config/best-practices/dkim_dmarc_spf.md
```diff
@@ -161,24 +167,34 @@ DKIM is currently supported by either OpenDKIM or Rspamd:
 
         enabled = true;
 
+        # If false, messages from authenticated users are not selected for signing
         sign_authenticated = true;
+        # If false, messages from local networks are not selected for signing
         sign_local = true;
 
+        # Domain to use for ARC signing: can be "header" (MIME From), "envelope" (SMTP From), "recipient" (SMTP To), "auth" (SMTP username) or directly specified domain name
```
polarathene, line 175: This config section is regarding DKIM signing not ARC.

Some of these are taken from the official Rspamd docs, I don't mind the additional context in our docs if it's of value, but we can probably just direct users to Rspamd docs for information that's already there (_and less likely to become outdated_).

We do direct users to this official docs for the config already however. So my main concern here is similar to the one you experienced with not being aware of `try_fallback = true;` for global config. If there is too much to read, users will gloss over the docs (_this is why some sections are collapsed on our docs as a UX principle known as "progressive disclosure", minimizing overwhelming users with information overload_).

I totally get wanting to understand more context about what settings do, but it's a balancing act 😅

docs/content/config/best-practices/dkim_dmarc_spf.md
```diff
@@ -161,24 +167,34 @@ DKIM is currently supported by either OpenDKIM or Rspamd:
...
-        use_redis = false; # don't change unless Redis also provides the DKIM keys
+        # don't change unless Redis also provides the DKIM keys
+        use_redis = false;
+        # Whether to normalise domains to eSLD
         use_esld = true;
-        check_pubkey = true; # you want to use this in the beginning
+        # you want to use this in the beginning
+        check_pubkey = true;
 
+        # global DKIM-selector: this is critical and must match a TXT entry called selector._domainkey in ALL of your domains
```
polarathene, line 184: Actually, this is the global default/fallback, as per my responses to your GH discussions on the topic.
- https://github.com/orgs/docker-mailserver/discussions/4572#discussioncomment-14471581
- https://github.com/orgs/docker-mailserver/discussions/4575#discussioncomment-14482755 (_came after your PR and documents the importance of `try_fallback = true;`_)

Perhaps our existing docs could be improved on this concern? (_although [they're already quite verbose](https://docker-mailserver.github.io/docker-mailserver/v15.1/config/best-practices/dkim_dmarc_spf/#dkim-dns)_)

Our docs should be rather clear on the step that you need a DNS TXT record with the selector that was generated. Perhaps we could streamline this process better in some way or communicate more clearly/quickly without tiring out the reader that just wants to setup the keys/selector and have it working, I'm just not sure what that'd look like.

docs/content/config/best-practices/dkim_dmarc_spf.md
```diff
@@ -161,24 +167,34 @@ DKIM is currently supported by either OpenDKIM or Rspamd:
...
         selector = "mail";
+        
         # The path location is searched for a DKIM key with these variables:
         # - `$domain` is sourced from the MIME mail message `From` header
         # - `$selector` is configured for `mail` (as a default fallback)
-        path = "/tmp/docker-mailserver/dkim/keys/$domain/$selector.private";
+        # Update the keytype=rsa and keysize=2048 to the values you use for your keys
+        #path = "/tmp/docker-mailserver/dkim/keys/$domain/$selector.private";  # eSLD
+        path = "/tmp/docker-mailserver/rspamd/dkim/rsa-2048-$selector-$domain.private.txt";
```
polarathene, line 192: I'm a little on the fence here. `path` will work in this example config unlike in the default `setup config dkim` config we generate, since there is no `try_fallback = false;` set.

I think when these docs were originally written with the example, the `setup config dkim` command did not generate a DKIM config when the user had not provided one, but now that is not required and our docs do try make the user aware of such as this config example section is collapsed by default, with an earlier section pointing out the DKIM config will be generated for you.

> image

---

For custom needs, the config can be useful as a reference and `path` here can be more useful.

I think I avoided the key type/size here since it wasn't particularly relevant. You could bring your own keys from elsewhere, the filename did not matter so long as it met the dynamic search path defined for `path` here.

With the planned rework to our DKIM key generation utility, we probably won't have the key type/size in the filename as it's not that useful or relevant to most users with single key (_which can be used for multiple domains, meaning you don't even need `$domain` technically_).

The `# eSLD` comment also doesn't make sense to me, I don't see anything that warrants a difference related to that. `use_esld = true;` will normalize the domain extracted from `use_domain`, should it have a matching suffix in the public suffix list (PSL).

---

So with all that in mind, do you see value in this change?

Issue #4178, opened 2024-09-12
bug report: Dovecot compression not working


alex9434:
### 📝 Preliminary Checks

- [X] I tried searching for an existing issue and followed the [debugging docs](https://docker-mailserver.github.io/docker-mailserver/latest/config/debugging/) advice, but still need assistance.

### 👀 What Happened?

I have the following setting in `dovecot.cf`

```
protocol imap {
	mail_plugins = $mail_plugins zlib imap_zlib
}

# Enable these only if you want compression while saving:
plugin {
	zlib_save = gz
	zlib_save_level = 9
}

plugin {
	## NOTE! Setting was called imap_zlib_compression_level before 2.3.15
	## imap_compress_<algorithm>_level = value
	##
	imap_compress_deflate_level = 6
}
```

### 👟 Reproduction Steps

New mail that arrives in maildir is does not get compressed. These settings should save all new email compressed with gz.

### 🐋 DMS Version

v14.0.0

### 💻 Operating System and Architecture

Debian 11 x86_64

### ⚙️ Container configuration files

```yml
services:
  mailserver:
    image: ghcr.io/docker-mailserver/docker-mailserver:latest
    read_only: false
    container_name: mailserver
    # Provide the FQDN of your mail server here (Your DNS MX record should point to this value)
    hostname: mail.emimx.com
    environment:
      - SSL_TYPE=manual
      - SSL_CERT_PATH=/tmp/dms/custom-certs/fullchain.cer
      - SSL_KEY_PATH=/tmp/dms/custom-certs/cert.key   
    env_file: mailserver.env
    # More information about the mail-server ports:
    # https://docker-mailserver.github.io/docker-mailserver/latest/config/security/understanding-the-ports/
    ports:
      - "25:25"    # SMTP  (explicit TLS => STARTTLS, Authentication is DISABLED => use port 465/587 instead)
      - "143:143"  # IMAP4 (explicit TLS => STARTTLS)
      - "465:465"  # ESMTP (implicit TLS)
      - "587:587"  # ESMTP (explicit TLS => STARTTLS)
      - "993:993"  # IMAP4 (implicit TLS)
    volumes:
      - ./mail/:/var/mail/
      - ./mail-state/:/var/mail-state/
      - ./mail-logs/:/var/log/mail/
      - ./config/:/tmp/docker-mailserver/
      - ./letsencrypt:/etc/letsencrypt
      - /etc/localtime:/etc/localtime:ro
[...]

PR #4578, opened 2025-09-22
docs: How to enable compression in Dovecot

audioscavenger:
# Description
New advanced page to enable zlib compression plugin for dovecot. Based off fts-xapian page. Tested and in production.

Fixes #

## Type of change

## Checklist

### DIFF
--- a/docs/content/examples/tutorials/mailbox-compression.md
+++ b/docs/content/examples/tutorials/mailbox-compression.md
@@ -1,0 +1,105 @@
+---
+title: 'Tutorials | Mailbox Compression'
+---
+
+## Overview
+
+!!! warning "This is a community contributed guide"
+
+    This content is entirely community supported. If you find errors, please open an issue and provide a PR.
+
+Dovecot provides a plugin for _mailbox compression_. This plugin enables support to read messages stored in a compressed format and to write new messages in a compressed format.
+
+This guide applies to DMS 16 and later, which uses Dovecot 2.4 and the [`mail-compress` plugin][dovecot::plugin::mail-compress].
+
+!!! tip "Compression algorithm"
+
+    The plugin supports several compression algorithms. For a good balance between compression ratio and performance, this guide uses `zstd`. See the [Dovecot documentation][dovecot::plugin::mail-compress] for the other supported algorithms.
+
+!!! info "Raising the memory limit to avoid errors"
+
+    Depending on usage, compression may exceed Dovecot's [default process memory limit][dovecot::config::default-vsz-limit] of 256 MiB. The configuration below shows how to raise that limit globally if necessary; tune it for your environment.
+
+[dovecot::plugin::mail-compress]: https://doc.dovecot.org/2.4.1/core/plugins/mail_compress.html
+[dovecot::config::default-vsz-limit]: https://doc.dovecot.org/2.4.1/core/summaries/settings.html#default_vsz_limit
+
+## Setup
+
+Create `docker-data/dms/config/dovecot/90-mail-compress.conf` with the following configuration:
+
+```conf
+# Enable the compression plugin globally for reading and writing:
+mail_plugins {
+  mail_compress = yes
+}
+
+# Compress new messages with zstd when they are saved:
+mail_compress_write_method = zstd
+
+# Uncomment and tune this only if Dovecot's default limit is too low:
+# default_vsz_limit = 1G
+```
+
+Alternatively, add the same configuration to `docker-data/dms/config/dovecot.cf`. DMS copies that file to Dovecot's local configuration.
+
+Add the configuration file to the `mailserver` service in your `compose.yaml`:
+
+```yaml
+services:
+  mailserver:
+    volumes:
+      - ./docker-data/dms/config/dovecot/90-mail-compress.conf:/etc/dovecot/conf.d/90-mail-compress.conf:ro
+```
+
+Restart DMS so it loads the new configuration:
+
+```console
+$ docker compose up -d --force-recreate
+```
+
+### Verify your configuration
+
+After restarting the DMS container, check that `mail_compress` is enabled for the services that read and write mail:
+
+```console
+$ docker compose exec mailserver doveconf -f protocol=lmtp mail_plugins
+mail_plugins {
+  mail_compress = yes
+  ...
+}
+
+$ docker compose exec mailserver doveconf -f protocol=imap mail_plugins
+mail_plugins {
+  mail_compress = yes
+  ...
+}
+
+$ docker compose exec mailserver doveconf -f protocol=pop3 mail_plugins
+mail_plugins {
+  mail_compress = yes
+  ...
+}
+```
+
+To verify that new mail is actually stored compressed, send a test message to an existing account and inspect the resulting Maildir file:
+
+```console
+$ docker compose exec mailserver swaks \
+    --server 0.0.0.0 \
+    --from hello@not-relevant.test \
+    --to john.doe@example.com
+$ file docker-data/dms/mail-data/example.com/john.doe/new/*
+...: Zstandard compressed data ...
+```
+
+## Troubleshooting
+
+### Cached message size larger than expected
+
+> ```
+> dms dovecot: indexer-worker(john.doe@example.com)<1662><PLy3NgU3z2h8BgAAPDOPJQ:vWooFAY3z2h>: Error: Mailbox INBOX: UID=3: read(/var/mail/example.com/john.doe/cur/1758395999.M371909P1818.mail.example.com,S=30343,W=30821:2,S) failed: Cached message size larger than expected (30343 > 16567, box=INBOX, UID=3) (read reason=mail stream)
+> 
+> dms dovecot: indexer-worker(john.doe@example.com)<1662><PLy3NgU3z2h8BgAAPDOPJQ:vWooFAY3z2h>: Error: Mailbox INBOX: Deleting corrupted cache record uid=3: UID 3: Broken physical size in mailbox INBOX: read(/var/mail/example.com/john.doe/cur/1758395999.M371909P1818.mail.example.com,S=30343,W=30821:2,S) failed: Cached message size larger than expected (30343 > 16567, box=INBOX, UID=3)
+> ```
+
+This can occur when importing mailboxes from another Dovecot system that were already compressed before the `mail-compress` plugin was enabled. Ensure the plugin is loaded and restart DMS with `docker compose up -d --force-recreate`.
--- a/docs/content/examples/tutorials/mailbox-compression.md
+++ b/docs/content/examples/tutorials/mailbox-compression.md
@@ -98,7 +98,7 @@ $ file docker-data/dms/mail-data/example.com/john.doe/new/*
 
 > ```
 > dms dovecot: indexer-worker(john.doe@example.com)<1662><PLy3NgU3z2h8BgAAPDOPJQ:vWooFAY3z2h>: Error: Mailbox INBOX: UID=3: read(/var/mail/example.com/john.doe/cur/1758395999.M371909P1818.mail.example.com,S=30343,W=30821:2,S) failed: Cached message size larger than expected (30343 > 16567, box=INBOX, UID=3) (read reason=mail stream)
-> 
+>
 > dms dovecot: indexer-worker(john.doe@example.com)<1662><PLy3NgU3z2h8BgAAPDOPJQ:vWooFAY3z2h>: Error: Mailbox INBOX: Deleting corrupted cache record uid=3: UID 3: Broken physical size in mailbox INBOX: read(/var/mail/example.com/john.doe/cur/1758395999.M371909P1818.mail.example.com,S=30343,W=30821:2,S) failed: Cached message size larger than expected (30343 > 16567, box=INBOX, UID=3)
 > ```
 
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -181,6 +181,7 @@ nav:
   - 'Examples':
     - 'Tutorials':
       - 'Basic Installation': examples/tutorials/basic-installation.md
+      - 'Mailbox Compression': examples/tutorials/mailbox-compression.md
       - 'Mailserver behind Proxy': examples/tutorials/mailserver-behind-proxy.md
       - 'Crowdsec': examples/tutorials/crowdsec.md
       - 'Building your own Docker image': examples/tutorials/docker-build.md
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@ All notable changes to this project will be documented in this file. The format 
 - **Dovecot**
   - FTS solr config: pinned to solr:10.0, explictly start solr in user managed mode and remove the attachment text extraction example. 
 
+### Documentation
+
+- Added a community guide for enabling Dovecot 2.4 mailbox compression ([#4578](https://github.com/docker-mailserver/docker-mailserver/pull/4578))
+
 ## [v16.0.1](https://github.com/docker-mailserver/docker-mailserver/releases/tag/v16.0.1)
 
 ### Added


docs/content/examples/tutorials/mailbox-compression.md
```diff
@@ -0,0 +1,139 @@
...
+| Algorithm | Size (GB) | Compression | Time (MM:SS) |
+| --------- | --------- | ----------- | ------------ |
+| None | 7.99 | 0% | 0:21 |
+| `bz2` | 3.41 | 57% | 7:08 |
+| `gz`  | 3.44 | 57% | 2:30 |
+| `deflate` | 3.44 | 57% | 2:34 |
+| `lz4` | 4.76 | 40% | 0:23 |
+| `zstd` | 3.41 | 57% | 0:34 |
+
+The clear winner is `zstd` but your results may vary.
```
polarathene: I don't see too much value in copy/pasting the referenced resource? Probably drop this section?

---

This benchmark is only covering compression time, LZ4 tends to fair better at decompression. Reads are more common than writes typically, so you may want to keep that in mind when conveying information (_optimizing for disk storage vs user perf_).

There's also a related Dovecot plugin for IMAP clients that support the extension which allows sending the compressed mail for reduced bandwidth (_costs the user less on data plans, and on slow connections they'll get it faster so win-win when supported_). However that would depend upon what the clients can then support decompressing.

docs/content/examples/tutorials/mailbox-compression.md
```diff
@@ -0,0 +1,139 @@
...
+
+    # Enable Zlib for imap
+    protocol imap {
+      mail_plugins = $mail_plugins zlib
+    }
+
+    # Enable Zlib for pop3
+    protocol pop3 {
+      mail_plugins = $mail_plugins zlib
+    }
```
polarathene: These enable reading from decompressed mail, but don't you need to also configure to store compressed mail? Have you tested that?

There was an issue raised last year about this feature and it seemed to require enabling for `lmtp` (_which we use to hand mail over from Postfix to Dovecot for storage_): https://github.com/docker-mailserver/docker-mailserver/issues/4178#issuecomment-2378091730

docs/content/examples/tutorials/mailbox-compression.md
```diff
@@ -0,0 +1,139 @@
...
+
+    Adjust the settings to tune for your desired memory limits and CPU usage.
+
+    The following compression levels are supported for `zlib_save_level`:
+    | Name   |   Minimum          |   Default | Maximum |
+    |--------|--------------------|-----------|-------- |
+    | `bz2`  | 1                  | 9         | 9       |
+    | `gz`   | 0 (no compression) | 6         | 9       |
+    | `lz4`  | 1                  | 1         | 9       |
+    | `zstd` | 1                  | 3         | 22      |
```
polarathene: I'm a bit reluctant to merge this as it would become part of our DMS v16 docs, and this table is copied from [Dovecot 2.3 `zlib_plugin` docs](https://doc.dovecot.org/2.3/settings/plugin/zlib-plugin/#plugin-zlib), but DMS v16 will be upgraded to Dovecot 2.4 where [the equivalent settings have been renamed](https://doc.dovecot.org/2.4.1/core/plugins/mail_compress.html) (_`zlib_save_level` is now split into multiple settings, one per codec_).

Your [related Github docs link](https://github.com/dovecot/documentation/blob/main/docs/core/plugins/mail_compress.md#example-configuration) is also the Dovecot 2.4 docs.

docs/content/examples/tutorials/mailbox-compression.md
```diff
@@ -0,0 +1,139 @@
...
+            - ./docker-data/dms/mail-logs/:/var/log/mail/
+            - ./docker-data/dms/config/:/tmp/docker-mailserver/
+    
+            - ./docker-data/dms/config/dovecot/12-compress.conf:/etc/dovecot/conf.d/12-compress.conf:ro
+          restart: always
+          stop_grace_period: 1m
+          # Uncomment if using `ENABLE_FAIL2BAN=1`:
+          # cap_add:
+            # - NET_ADMIN
+    ```
```
polarathene: The bulk of this is noise and can be discarded.

All that really matters here to the user is:

```yaml
services:
  mailserver:
    volumes:
      - ./docker-data/dms/config/dovecot/12-compress.conf:/etc/dovecot/conf.d/12-compress.conf:ro
```

They are interested in the specific change to make, not to identify the minor addition from the example to apply to their own `compose.yaml` 😅 

Sometimes full reproductions are useful (_like I [documented here](https://github.com/docker-mailserver/docker-mailserver/issues/4178#issuecomment-2378091730) when someone was having trouble with this specific Dovecot plugin_), but for simple stuff like this a single focused snippet should be sufficient.

docs/content/examples/tutorials/mailbox-compression.md
```diff
@@ -0,0 +1,139 @@
...
+
+4. Recreate containers:
+
+    ```
+    docker compose down
+    docker compose up -d --force-recreate
+    ```
```
polarathene: `down` is redundant when using `--force-recreate` AFAIK.

docs/content/examples/tutorials/mailbox-compression.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Debugging
+
+#### Cached message size larger than expected
+```
+mailu dovecot: indexer-worker(user@mydomain.com)<1662><PLy3NgU3z2h8BgAAPDOPJQ:vWooFAY3z2h>: Error: Mailbox INBOX: UID=3: read(/var/mail/mydomain.com/user/cur/1758395999.M371909P1818.mx.mydomain.com,S=30343,W=30821:2,S) failed: Cached message size larger than expected (30343 > 16567, box=INBOX, UID=3) (read reason=mail stream)
+mailu dovecot: indexer-worker(user@mydomain.com)<1662><PLy3NgU3z2h8BgAAPDOPJQ:vWooFAY3z2h>: Error: Mailbox INBOX: Deleting corrupted cache record uid=3: UID 3: Broken physical size in mailbox INBOX: read(/var/mail/mydomain.com/user/cur/1758395999.M371909P1818.mx.mydomain.com,S=30343,W=30821:2,S) failed: Cached message size larger than expected (30343 > 16567, box=INBOX, UID=3)
+```
```
polarathene: `mailu` should not be present in our docs, trim that out.

`mydomain.com` should be `example.com` to remain consistent across usage in our docs.

docs/content/examples/tutorials/mailbox-compression.md
```diff
@@ -0,0 +1,139 @@
...
+
+Common error after importing devocot mailboxes from another system where they are compressed, but you did not enable zlib yet.
```
polarathene:
```suggestion
This is a common error after importing dovecot mailboxes from another system that were already compressed, but you did not enable zlib yet.
```

I'm not sure how relevant that is to the reader of the docs here though?

If someone came across the docs page contributed here, and followed the configuration steps, zlib should be enabled?

I understand that in your case, without such docs this was a problem you encountered along the way and that it can happen with DMS if the config wasn't properly applied (_such as requiring a fresh container with `--force-recreate`_).

A useful tip here would be to verify that the plugin has been configured/detected by Dovecot at least, as I show at the bottom of [this comment](https://github.com/docker-mailserver/docker-mailserver/issues/4178#issuecomment-2378091730) where I document commands to help verify the feature is working correctly.

```console
# Verify Dovecot config is correct for `mail_plugins` per protocol:
$ doveconf -f protocol=lmtp mail_plugins
mail_plugins =  quota sieve zlib
$ doveconf -f protocol=imap mail_plugins
mail_plugins =  quota imap_quota imap_zlib
```

docs/content/examples/tutorials/mailbox-compression.md
```diff
@@ -0,0 +1,139 @@
...
+
+Also could be caused by incorrect loading of the plugin: wrong path in `compose.yaml`, wrong order, syntax error in 12-compress.conf ...
```
polarathene: These are valid, but more generic and I think our standard debugging/troubleshooting docs page has that mostly covered. To avoid repeating that on all our config guides, we prefer to exclude such and only include when enough users repeatedly encounter these mistakes (_which likely means our docs are failing them and that should be addressed instead_).

docs/content/examples/tutorials/mailbox-compression.md
```diff
@@ -0,0 +1,139 @@
...
+| `gz`  | [zlib (gzip)](https://www.zlib.net/) | v2.0+ |
+| `deflate` | [zlib (gzip)](https://www.zlib.net/) | v2.0+ |
+| `lz4` | [liblz4](https://www.lz4.org/) | v2.2.11+ |
+| `zstd` | [Zstandard](https://facebook.github.io/zstd/) | v2.3.12+ |
+
+Please be aware that compression increases the CPU usage of the container just a little. However, when dealing with large attachements, users may experience delays and the container CPU may spike.
+
+### Zlib
+
+The [dovecot-zlib](https://doc.dovecot.org/2.3/configuration_manual/zlib_plugin/) plugin can be used to read compressed mbox, maildir or dbox files. It can also be used to write (via IMAP, LDA and/or LMTP Server) compressed messages to dbox or Maildir mailboxes.
```
polarathene: This is taken straight from the Dovecot docs.

- DMS uses `maildir` and we'll likely drop support for other formats in future, one is already discouraged/deprecated with Dovecot 2.4 IIRC.
- DMS uses LMTP only. [LDA was dropped](https://github.com/docker-mailserver/docker-mailserver/issues/4178#issuecomment-2375489302) in favor of LMTP a while back, although some may customize DMS to use LDA or the equivalent command (_for fetchmail/getmail to deliver to Dovecot directly IIRC_), so there may be relevance there (_it'd then need the equivalent plugin config like I mentioned for `lmtp`_).

This line and the section header directly above it can probably be removed?

docs/content/examples/tutorials/mailbox-compression.md
```diff
@@ -0,0 +1,139 @@
...
+
+While compression can be memory intensive for large messages, you can configure the plugin to limit the amount of memory consumed by the workers. Using `zstd` seems to be the best and faster algorith to date.
```
polarathene: Switched to a tip admonition, and fixed typo "algorith".

```suggestion
!!! tip "Adjusting the memory limit"

    While compression can be memory intensive for large messages, you can configure the plugin to limit the amount of memory consumed by the workers. An example is shown in the below Dovecot config example with the `vsz_limit` setting to limit to 1GB.

    Using `zstd` seems to be the best and faster algorithm to date.
```

docs/content/examples/tutorials/mailbox-compression.md
```diff
@@ -0,0 +1,139 @@
...
+
+Compression allows all messages to be compressed on save, and compressed messages to be read, regardless of the Algorithm chosen. Dovecot **zlib** plugin supports a variety of algorithms:
+
+| Name | Library (algorithm) | Dovecot Support |
+| ---- | ------------------- | --------------- |
+| `bz2` | [libbzip2 (bzip2)](https://sourceware.org/bzip2/) | v2.0+ |
+| `gz`  | [zlib (gzip)](https://www.zlib.net/) | v2.0+ |
+| `deflate` | [zlib (gzip)](https://www.zlib.net/) | v2.0+ |
+| `lz4` | [liblz4](https://www.lz4.org/) | v2.2.11+ |
+| `zstd` | [Zstandard](https://facebook.github.io/zstd/) | v2.3.12+ |
```
polarathene: Better to link to the Dovecot docs perhaps and just keep it simple for users happy with the advised `zstd`? Those that care about the compression codec and level can reference the upstream docs if they need to deviate?

The simpler the docs are, the easier they'll be to ingest by the reader. I tend to only add verbosity when the technical information can be more useful and harder to source/grok.

polarathene (changes requested):
Thanks for the contribution (_almost exactly one year later since [another user offered to contribute docs on this feature](https://github.com/docker-mailserver/docker-mailserver/issues/4178#issuecomment-2378091730)_ 😄) 

There's quite a bit that I don't think our docs need to carry. The feature should be fairly simple to document for DMS. Some small typos to correct and a bit of restructuring to do but otherwise looks good 👍

The docs will be part of DMS v16 however, and this will require changing the config to be compatible with Dovecot 2.4. I would suggest just focusing on instructing users how to configure for `zstd` and should anyone have a valid need for the others, we can update the docs at that point.

---

This page might better live at our Examples nav tab for now? Under Tutorials perhaps?

To actually have a new docs page be part of our nav, we do need to add it into the nav hierarchy like shown here in `mkdocs.yml`:

https://github.com/docker-mailserver/docker-mailserver/blob/369f3573510cdcfda911b28d9d51846abb48617b/docs/mkdocs.yml#L181-L183

So you'll need to update that file too.

docs/content/examples/tutorials/mailbox-compression.md
audioscavenger: the full snippet includes the compression on save, i am not sure what is confusing?
```
    mail_plugins = $mail_plugins zlib

    # Enable these only if you want compression while saving: <<<<<<<<<<<<<<<<<<<<<<
    plugin {
      # zlib_save = gz
      # zlib_save_level = 1
      zlib_save = zstd
    }

    # Enable Zlib for imap
    protocol imap {
      mail_plugins = $mail_plugins zlib
    }

    # Enable Zlib for pop3
    protocol pop3 {
      mail_plugins = $mail_plugins zlib
    }

    # Increase memory allowed for imap as it costs more to read compressed files
    service imap {
      vsz_limit = 1GB
    }
```

docs/content/examples/tutorials/mailbox-compression.md
audioscavenger: we must include some sort of summary for compression options, what harm is it? besides, the values have not changed aside from the keyword names. 

So i have rewritten this section to separate between dovecot 2.3-/2.4+ and DMS 15-/16+

docs/content/examples/tutorials/mailbox-compression.md
audioscavenger: yup, removed the fluff, agreed

docs/content/examples/tutorials/mailbox-compression.md
audioscavenger: okay right, removed the extra line

docs/content/examples/tutorials/mailbox-compression.md
audioscavenger: okay i did just that and will keep the whole original md file for my blog, got it

docs/content/examples/tutorials/mailbox-compression.md
audioscavenger: I used your fix and moved all into #overview, and removed the fluff

docs/content/examples/tutorials/mailbox-compression.md
audioscavenger: I removed all mentions of LDA and LMTP as this document is for mail read/save compression only

docs/content/examples/tutorials/mailbox-compression.md
audioscavenger: I have encountered this mistake when I imported mailu into dms. Also I got it again when I did not load 12-compress,conf in the right order. I want to keep this in the document.

docs/content/examples/tutorials/mailbox-compression.md
audioscavenger: got it, fixed

docs/content/examples/tutorials/mailbox-compression.md
audioscavenger: I added a ### Verify your configuration section, thanks

docs/content/examples/tutorials/mailbox-compression.md
audioscavenger: anyways, I have now included 2 sections for DMS15- and DMS16+

docs/content/examples/tutorials/mailbox-compression.md
audioscavenger: i have removed all this and put a link to  the recent benchmark in the ## Overview section

audioscavenger:
@polarathene I have updated the file to reflect all your suggestions, modified mkdocs.yml and moved the file under examples

docs/content/examples/tutorials/mailbox-compression.md
```diff
@@ -0,0 +1,142 @@
...
+## Overview
+
+Compression allows all messages compressed or not to be READ, and/or compressed on save, regardless of the Algorithm chosen for write. Dovecot message compression plugin supports a variety of algorithms, please review them and the options available at:
+
+* Dovecot 2.3- / DMS 15-: [dovecot-zlib](https://doc.dovecot.org/2.3/configuration_manual/zlib_plugin/)
+* Dovecot 2.4+ / DMS 16+: [mail-compress](https://doc.dovecot.org/2.4.1/core/plugins/mail_compress.html)
+
+Using `zstd` method seems to be the best and faster algorithm to date, according to a recent [mail_compress-benchmark](https://github.com/dovecot/documentation/edit/main/docs/core/plugins/mail_compress.md).
+
+While compression can be memory intensive for large messages, you can configure the plugin to limit the amount of memory consumed by the workers. Examples shown mention the `vsz_limit` setting to limit to 1GB RAM for instance.
```
polarathene: **Revised:**
- Rewrote introduction paragraph to better clarify read/write behaviour.
- Switched to markdown reflinks, as is standard in our docs. Bullet points also use `-` rather than `*`
- Final two paragraphs converted to admonitions (tip and info types), with revised content.


```suggestion
Dovecot provides a plugin for _mailbox compression_. This plugin enables support to read messages from a mailbox that have been stored in a compressed format, as well as to write new incoming messages into a compressed format for storage.

Configuration differs by Dovecot version, which is dependent upon your DMS version:

- Dovecot 2.3 (DMS <=15): [`zlib`][dovecot::plugin::compression-2.3]
- Dovecot 2.4 (DMS >=16): [`mail-compress`][dovecot::plugin::compression-2.4]

!!! tip "Compression Algorithm"

    The plugin supports a variety of compression algorithms (see [this May 2025 benchmark][dovecot::plugin::compression-benchmark]), you'll likely want to choose `zstd`.

!!! info "Raising the memory limit to avoid errors"

    The memory requirements to support this plugin (depending on usage) may exceed the [default configured process memory limit][dovecot::config::default-vsz-limit] (_`default_vsz_limit`, 256 MiB_), or if configured a [service specific memory limit](https://doc.dovecot.org/2.4.1/core/config/service.html#service_vsz_limit). Examples shown mention the `vsz_limit` setting to limit to 1GB RAM for instance.

[dovecot::plugin::compression-2.3]: https://doc.dovecot.org/2.3/configuration_manual/zlib_plugin
[dovecot::plugin::compression-2.4]: https://doc.dovecot.org/2.4.1/core/plugins/mail_compress.html
[dovecot::plugin::compression-benchmark]: https://github.com/dovecot/documentation/blob/main/docs/core/plugins/mail_compress.md#benchmarking
[dovecot::config::default-vsz-limit]: https://doc.dovecot.org/2.4.1/core/summaries/settings.html#default_vsz_limit
[dovecot::config::service-vsz-limit]: https://doc.dovecot.org/2.4.1/core/config/service.html#service_vsz_limit
```

docs/content/examples/tutorials/mailbox-compression.md
```diff
@@ -0,0 +1,142 @@
...
+
+### Cached message size larger than expected
+```
+dms dovecot: indexer-worker(user@example.com)<1662><PLy3NgU3z2h8BgAAPDOPJQ:vWooFAY3z2h>: Error: Mailbox INBOX: UID=3: read(/var/mail/example.com/user/cur/1758395999.M371909P1818.mx.example.com,S=30343,W=30821:2,S) failed: Cached message size larger than expected (30343 > 16567, box=INBOX, UID=3) (read reason=mail stream)
+dms dovecot: indexer-worker(user@example.com)<1662><PLy3NgU3z2h8BgAAPDOPJQ:vWooFAY3z2h>: Error: Mailbox INBOX: Deleting corrupted cache record uid=3: UID 3: Broken physical size in mailbox INBOX: read(/var/mail/example.com/user/cur/1758395999.M371909P1818.mx.example.com,S=30343,W=30821:2,S) failed: Cached message size larger than expected (30343 > 16567, box=INBOX, UID=3)
+```
+
+This is a common error after importing dovecot mailboxes from another system that were already compressed, but you did not enable zlib yet and dovecot is failing to read the messages.
+
+Also could be caused by incorrect loading of the plugin: wrong path in `compose.yaml`, wrong order, syntax error in `12-compress.conf` ...
```
polarathene: Rename section title, minor revision to formatting and revised paragraph clarifying the cause.

````suggestion
## Troubleshooting

### Cached message size larger than expected

> ```
> dms dovecot: indexer-worker(john.doe@example.com)<1662><PLy3NgU3z2h8BgAAPDOPJQ:vWooFAY3z2h>: Error: Mailbox INBOX: UID=3: read(/var/mail/example.com/john.doe/cur/1758395999.M371909P1818.mail.example.com,S=30343,W=30821:2,S) failed: Cached message size larger than expected (30343 > 16567, box=INBOX, UID=3) (read reason=mail stream)
> 
> dms dovecot: indexer-worker(john.doe@example.com)<1662><PLy3NgU3z2h8BgAAPDOPJQ:vWooFAY3z2h>: Error: Mailbox INBOX: Deleting corrupted cache record uid=3: UID 3: Broken physical size in mailbox INBOX: read(/var/mail/example.com/john.doe/cur/1758395999.M371909P1818.mail.example.com,S=30343,W=30821:2,S) failed: Cached message size larger than expected (30343 > 16567, box=INBOX, UID=3)
> ```

This can occur from importing mail from Dovecot mailboxes from another system that were already compressed, but you haven't yet enabled the Dovecot plugin (_which is required to read them_).
````

The last line I've removed as that seems more appropriate for directing the user to a common troubleshooting docs page (_that we have_), for which that kind of generic problem solving would be suitable for.

If it were to be kept it'd at least be revised to something like this:

```
Other potential causes could be incorrect loading of the plugin due to:
- An invalid path in the `compose.yaml`.
- Loading order was too late in `/etc/dovecot/conf.d/` due to filename.
- Syntax error in `12-compress.conf`.
```

File: docs/mkdocs.yml [hunk 14]
```diff
@@ -181,6 +181,7 @@ nav:
   - 'Examples':
     - 'Tutorials':
       - 'Basic Installation': examples/tutorials/basic-installation.md
+      - 'Enable messages compression': examples/tutorials/mailbox-compression.md
```
polarathene:
```suggestion
      - 'Mailbox Compression': examples/tutorials/mailbox-compression.md
```

docs/content/examples/tutorials/mailbox-compression.md
```diff
@@ -0,0 +1,142 @@
...
+Adjust the settings to tune for your desired memory limits and CPU usage.
+
+The following compression levels are supported, and compression level are now controlled by one keyword per method:
+
+| mail_compress_write_method |        Dovecot 2.4+          |       Minimum      |   Default | Maximum |
+|----------------------------|------------------------------|--------------------|-----------|-------- |
+| `bz2`                      | compress_bz2_block_size_100k | 1                  | 9         | 9       |
+| `gz`                       | compress_gz_level            | 0 (no compression) | 6         | 9       |
+| `deflate`                  | compress_deflate_level       | 0 (no compression) | 6         | 9       |
+| `lz4`                      | ?                            | 1                  | 1         | 9       |
```
polarathene: - There is no equivalent Dovecot setting (_[like for zstd](https://github.com/dovecot/core/blob/0962ed2104800eae2ea00899f106638f8654c8e2/src/lib-compression/ostream-zstd.c#L33-L36) or others_) for [LZ4 supported by Dovecot](https://github.com/dovecot/core/blob/2.4.2/src/lib-compression/ostream-lz4.c), just a [call to the default LZ4 compression level](https://github.com/dovecot/core/blob/0962ed2104800eae2ea00899f106638f8654c8e2/src/lib-compression/ostream-lz4.c#L88).
- For reference the [LZ4 CLI args](https://github.com/lz4/lz4/blob/v1.10.0/programs/lz4.1.md#operation-modifiers) explain default settings (_and the related options to favor faster compression or higher compression levels_).

```suggestion
| `lz4`                      | (not configurable)           | 1                  | 1         | 9       |
```

---

I'm still of the opinion that this isn't that relevant for us to document though? This is a feature where Dovecot is compressing/decompressing from storage, none of it relevant to the client compatibility wise, so the choice is really down to zstd vs lz4 (lacks config flexibility), where LZ4's main advantage is decompression speed AFAIK but without that being tunable while zstd is, it may make zstd the sane choice.

More interested users that care about the difference could do so via browsing upstream docs if they care enough to make a decision there? Majority likely don't need to think much about this and would be fine with the happy path :)

docs/content/examples/tutorials/mailbox-compression.md
```diff
@@ -0,0 +1,142 @@
...
+        - ./docker-data/dms/config/dovecot/12-compress-2.3.conf:/etc/dovecot/conf.d/12-compress-2.3.conf:ro
+        # for DMS 16+
+        - ./docker-data/dms/config/dovecot/12-compress-2.4.conf:/etc/dovecot/conf.d/12-compress-2.4.conf:ro
+```
+
+Finally, restart compose:
+
+```
+docker compose up -d --force-recreate
+```
```
polarathene: There really shouldn't be a need to document the config filename by Dovecot version with the DMS comment. Additionally the yaml snippet was indented which was unnecessary.

````suggestion
In your `compose.yaml`, add a volume bind mount for the Dovecot plugin config to the container location `/etc/dovecot/conf.d/` (_the main Dovecot config `/etc/dovecot/dovecot.conf` will load all configs from that directory_).

```yaml
services:
  mailserver:
    volumes:
      - ./docker-data/dms/config/dovecot/12-compress.conf:/etc/dovecot/conf.d/12-compress.conf:ro
```

Ensure the new config is applied correctly by restarting DMS with `docker compose up -d --force-recreate`.
````

I am a bit iffy with that choice for a `12-` prefix, which seems arbitrary?

---

## Reference - Why the numeric prefix (`12-`) in the config file name?

This PR doesn't document or mention the reasoning for the `12-` prefix you chose for the config file (_it was however previously explained [here](https://github.com/docker-mailserver/docker-mailserver/issues/4570#issuecomment-3316268904)_).

Without the prefix a warning would be logged like this:

> ```
> dovecot: doveconf: Warning: /etc/dovecot/conf.d/compression.conf line 1: Global setting mail_plugins won't change the setting inside an earlier filter at /etc/dovecot/conf.d/15-lda.conf line 47 (if this is intentional, avoid this warning by moving the global setting before /etc/dovecot/conf.d/15-lda.conf line 47)
> ```

`15-lda.conf` is a Dovecot supplied config, DMS is not involved in that. It's related snippet triggering that warning is this:

```conf
protocol lda {
  # Space separated list of plugins to load (default is global mail_plugins).
  mail_plugins = $mail_plugins sieve
}
```

`/etc/dovecot/dovecot.conf` loads all configs in lexicographical order via this snippet:

```conf
# Most of the actual configuration gets included below. The filenames are
# first sorted by their ASCII value and parsed in that order. The 00-prefixes
# in filenames are intended to make it easier to understand the ordering.
!include conf.d/*.conf
```

In this case `15-lda.conf` was processed earlier than any non-numeric prefixed config files, where it referenced the global setting `$mail_plugins` before the later parsed config (`compression.conf`) was loaded with it's own modification to the `mail_plugins` global setting. As such, `15-lda.conf` (_and any other configs using `$mail_plugins`, since Dovecot only logs the first instance encountered to trigger that warning_) the expected change wasn't applied (_unless the more explicit protocol override was present which also included the change_).

So to avoid that warning, `12-` as a prefix does work... but it could still introduce the problem should a custom config (or Dovecot update) introduce similar earlier than that `12-` prefix. Alternatively, as per the affected Dovecot snippets, one could just reference the services themselves to avoid the otherwise necessary prefix and warnings concerns.

---

DMS also supplies custom `20-lmtp.conf` + `20-imap.conf` configs (_there's additionally `20-pop3.conf` included internally from Dovecot, which DMS does not provide an override for, where `mail_plugins` for that protocol/service is unset, and thus only uses whatever was set globally_):

https://github.com/docker-mailserver/docker-mailserver/blob/12bb27ed8238d92445d3b981585553ec87d44bc3/target/dovecot/20-lmtp.conf#L16-L19

https://github.com/docker-mailserver/docker-mailserver/blob/12bb27ed8238d92445d3b981585553ec87d44bc3/target/dovecot/20-imap.conf#L1-L4

Along with DMS supplied `10-mail.conf` (_which sets the `mail_plugins` global setting_):

https://github.com/docker-mailserver/docker-mailserver/blob/12bb27ed8238d92445d3b981585553ec87d44bc3/target/dovecot/10-mail.conf#L207-L209

Avoiding the global setting, the warning is avoided, but as a result services not explicitly configured would lack the plugin being enabled (_which was the concern [I raised in my previous review feedback](https://github.com/docker-mailserver/docker-mailserver/pull/4578#discussion_r2373814962) and referenced [my comment from an earlier `zlib` config issue](https://github.com/docker-mailserver/docker-mailserver/issues/4178#issuecomment-2378091730)_):

```yaml
services:
  dms:
    image: ghcr.io/docker-mailserver/docker-mailserver:15.1
    hostname: mail.example.test
    configs:
      - source: dms-accounts
        target: /tmp/docker-mailserver/postfix-accounts.cf
      - source: dovecot-compression-2.3
        target: /etc/dovecot/conf.d/compression.conf

# Docker Compose `configs` feature
# NOTE: `$` literals must be escaped as `$$` to avoid being mistaken for ENV interpolation.
configs:
  dovecot-compression-2.3:
    content: |
      # Configure the `zlib` plugin to use `zstd`:
      plugin {
        zlib_save = zstd
      }

      # Globally enable:
      #mail_plugins = $$mail_plugins zlib

      # Enable per protocol (redundant if globally enabled)
      protocol imap {
        mail_plugins = $$mail_plugins zlib
      }

      protocol pop3 {
        mail_plugins = $$mail_plugins zlib
      }

  dms-accounts:
    content: |
      john.doe@example.test|{SHA512-CRYPT}$$6$$sbgFRCmQ.KWS5ryb$$EsWrlYosiadgdUOxCBHY0DQ3qFbeudDhNMqHs6jZt.8gmxUwiLVy738knqkHD4zj4amkb296HFqQ3yDq4UXt8.
```

```console
# Global setting:
$ doveconf mail_plugins
mail_plugins =  quota

# LMTP missing support as it wasn't explicitly configured,
# without being set globally nor was it inherited.
$ doveconf -f protocol=lmtp mail_plugins
mail_plugins =  quota sieve

# IMAP added support:
$ doveconf -f protocol=imap mail_plugins
mail_plugins =  quota imap_quota zlib

# POP3 originally only used global config:
$ doveconf -f protocol=pop3 mail_plugins
mail_plugins =  quota zlib
```

So this would enable IMAP/POP3 **read** support, where Dovecot would decompress such mail upon retrieval, but when mail is delivered (**write**) to Dovecot it will be via LDA or LMTP, so these are relevant too.

Test mails to the Dovecot LMTP and LDA services can be tried via:

```console
# LMTP:
# `--socket` + `--protocol` options to send via LMTP to Dovecot directly (instead of SMTP via Postfix)
$ swaks --silent --socket /var/run/dovecot/lmtp --protocol lmtp \
  --from hello@not-relevant.test \
  --to john.doe@example.test


# LDA:
# Have swaks dump it's test mail to a file:
$ swaks --from hello@not-relevant.test --to john.doe@example.test --dump-mail > /tmp/example.eml
# Then reference that via `dovecot-lda` with the `-p` option, `-d` must be given for where to deliver:
$ /usr/lib/dovecot/dovecot-lda -d john.doe@example.test -p /tmp/example.eml


# Both delivered as text (no compression):
# Types identified differ as `dovecot-lda` excludes additional mail headers that `file` recognizes
$ file /var/mail/example.test/john.doe/new/*
/var/mail/example.test/john.doe/new/1762066677.M530062P7803.mail.example.test,S=549,W=566: SMTP mail, ASCII text
/var/mail/example.test/john.doe/new/1762066687.M352940P7825.mail.example.test,S=276,W=285: news or mail, ASCII text


# For reference, this would submit mail to Postfix that delivers (via LMTP) to the Dovecot managed mailbox:
swaks --silent --server 0.0.0.0 --port 465 \
  --auth-user john.doe@example.test --auth-pass secret \
  --from hello@not-relevant.test \
  --to john.doe@example.test
```

Normally mail will go through Postfix which will deliver mail to Dovecot via that LMTP socket, but it's also possible that Getmail/Fetchmail services for example are using `dovecot-lda` to deliver mail to Dovecot directly without going through LMTP (_see [this example documented by me](https://github.com/orgs/docker-mailserver/discussions/4348#discussioncomment-12106141) a while back_). Thus LDA config should be considered when configuring this feature too.

Therefore, just as the [Dovecot docs config example shows](https://doc.dovecot.org/2.3/configuration_manual/zlib_plugin/#sample-configuration), we should just keep it simple with a global `mail_plugins` setting only? (_which would require a numeric prefix below `15`_) This way the user doesn't need to think about each protocol used and their separate syntax to support?

docs/content/examples/tutorials/mailbox-compression.md
```diff
@@ -0,0 +1,142 @@
...
+
+Using `zstd` method seems to be the best and faster algorithm to date, according to a recent [mail_compress-benchmark](https://github.com/dovecot/documentation/edit/main/docs/core/plugins/mail_compress.md).
+
+While compression can be memory intensive for large messages, you can configure the plugin to limit the amount of memory consumed by the workers. Examples shown mention the `vsz_limit` setting to limit to 1GB RAM for instance.
+
+## Setup
+
+### Dovecot < 2.4 / DMS < 16
+
+To configure  as a dovecot plugin, create a file at `docker-data/dms/config/dovecot/12-compress-2.3.conf` and place the following in it:
```
polarathene:
```suggestion
To configure as a dovecot plugin, create a file at `docker-data/dms/config/dovecot/12-compress.conf` and place the following in it:
```

docs/content/examples/tutorials/mailbox-compression.md
```diff
@@ -0,0 +1,142 @@
...
+
+    The following compression levels are supported for `zlib_save_level`:
+    
+    | Name   |   Minimum          |   Default | Maximum |
+    |--------|--------------------|-----------|-------- |
+    | `bz2`  | 1                  | 9         | 9       |
+    | `gz`   | 0 (no compression) | 6         | 9       |
+    | `lz4`  | 1                  | 1         | 9       |
+    | `zstd` | 1                  | 3         | 22      |
+
```
polarathene: Correct the indentation, although I'm also in favor of removing these tables, I'm more concerned about the rest of this docs page providing quality advice.

```suggestion
Adjust the settings to tune for your desired memory limits and CPU usage.

The following compression levels are supported for `zlib_save_level`:

| Name   |   Minimum          |   Default | Maximum |
|--------|--------------------|-----------|-------- |
| `bz2`  | 1                  | 9         | 9       |
| `gz`   | 0 (no compression) | 6         | 9       |
| `lz4`  | 1                  | 1         | 9       |
| `zstd` | 1                  | 3         | 22      |
```

docs/content/examples/tutorials/mailbox-compression.md
```diff
@@ -0,0 +1,142 @@
...
+  # mail_plugins =  quota fts fts_xapian zlib sieve
+
+docker exec -it dms doveconf -f protocol=imap mail_plugins
+  # mail_plugins =  quota fts fts_xapian zlib zlib imap_quota imap_sieve
+
+docker exec -it dms doveconf -f protocol=pop3 mail_plugins
+  # mail_plugins =  quota fts fts_xapian zlib zlib
+```
+
+The output above show that zlib is activated for lmtp/imap/pop3 with DMS 15-/Dovecot 2.3-. This output will change a little with Dovecot 2.4+ as the compression plugin name has changed.
```
polarathene: The docs are standardized around `mailserver` compose service name. We should use the `docker compose exec` command instead of `docker exec -it` as well.

- Fixed typo on compression.
- Cleaned up the introductory line.
- Revised the terminal snippet (_which may not be accurate, I need to reproduce_), removing the FTS plugins from your own setup output, and duplicated `zlib`.
- Revised the final paragraph, although ideally the final sentence wouldn't be necessary and an example for that was provided too.

````suggestion
After restarting the DMS container, check that the compression plugin was loaded properly:

```console
$ docker compose exec mailserver bash

$ doveconf -f protocol=lmtp mail_plugins
mail_plugins =  quota zlib sieve

$ doveconf -f protocol=imap mail_plugins
mail_plugins =  quota zlib imap_quota imap_sieve

$ doveconf -f protocol=pop3 mail_plugins
mail_plugins =  quota zlib
```

The output above show that compression plugin (`zlib` for Dovecot 2.3) is activated for the protocols queried (LMTP, IMAP, POP3). This output will change a little with Dovecot 2.4+ as the compression plugin name has changed.
````

File: docs/content/examples/tutorials/mailbox-compression.md [hunk 20]
```diff
@@ -0,0 +1,142 @@
...
+    |--------|--------------------|-----------|-------- |
+    | `bz2`  | 1                  | 9         | 9       |
+    | `gz`   | 0 (no compression) | 6         | 9       |
+    | `lz4`  | 1                  | 1         | 9       |
+    | `zstd` | 1                  | 3         | 22      |
+
+
+### Dovecot 2.4+ / DMS 16+
+
+From Dovecot 2.4, the compression plugin name and setup have changed: [mail-compress](https://doc.dovecot.org/2.4.1/core/plugins/mail_compress.html). Create a file at `docker-data/dms/config/dovecot/12-compress.conf` and place the following in it:
```
polarathene: Align with the same Dovecot 2.3 sentence:

```suggestion
To configure as a dovecot plugin, create a file at `docker-data/dms/config/dovecot/12-compress.conf` and place the following in it:
```

docs/content/examples/tutorials/mailbox-compression.md
```diff
@@ -0,0 +1,142 @@
...
+  mail_compress = yes
+}
+
+# Enable WRITE compression on saving messages:
+mail_compress_write_method = zstd
+
+# Increase memory allowed for imap as it costs more to read compressed files
+service imap {
+  vsz_limit = 1GB
+}
```
polarathene: This is only considering imap, as noted in other feedback that may be inaccurate. If memory is an issue it could be relevant to consider the cost for compression for inbound mail large enough to warrant this, in addition to the outbound read to clients over IMAP and potentially POP3.

A global upgrade to that setting may be better?

docs/content/examples/tutorials/mailbox-compression.md
```diff
@@ -0,0 +1,142 @@
...
+# Enable Zlib for imap
+protocol imap {
+  mail_plugins = $mail_plugins zlib
+}
+
+# Enable READ compression for POP3:
+protocol pop3 {
+  mail_plugins = $mail_plugins zlib
+}
+
```
polarathene: As noted, these are redundant, they duplicate the `zlib` addition inherited from the global `mail_plugins` setting (_which is the reason for needing the `12-` prefix for the filename in the first place_).

```suggestion
```

docs/content/examples/tutorials/mailbox-compression.md
```diff
@@ -0,0 +1,142 @@
...
+```
+mail_plugins = $mail_plugins zlib
+
+# Enable WRITE compression on saving messages:
+plugin {
+  # zlib_save = gz
+  # zlib_save_level = 1
+  zlib_save = zstd
+}
```
polarathene: Perhaps this is more appropriate? I don't think the prior default codec and it's specific default level needs to be included?

```suggestion
# Enable to support reading compressed messages:
mail_plugins = $mail_plugins zlib

# Enable compression with `zstd` when storing new messages
# (requires the `zlib` plugin enabled in `mail_plugins`):
plugin {
  zlib_save = zstd
}
```

docs/content/examples/tutorials/mailbox-compression.md
```diff
@@ -0,0 +1,142 @@
+---
+title: 'Tutorials | Mailbox Compression'
+---
+
+## Overview
```
polarathene, line 5: Including a disclaimer as we've done with other examples.

```suggestion
## Overview

!!! warning "This is a community contributed guide"

    This content is entirely community supported. If you find errors, please open an issue and provide a PR.
```

polarathene (changes requested):
This review is mostly complete, I would apply the changes myself and continue on to another iteration but I'm a bit short on time atm. I'm submitting it now as my system is acting up and I'd rather not lose the review unexpectedly like I just did with some other processes 😓 

The Dovecot 2.3/2.4 differences will not be that relevant to retain the 2.3 going forward, but I'm fine with the comparison of the two for the v16 docs. A future release can discard 2.3 section after that.

I would like to verify Dovecot 2.4. I mostly verified with Dovecot 2.3, and planned to tackle the concerns raised regarding 2.3 config and config filename on the 2nd pass. Verification section probably include additional snippets like [I showed for testing mail and verifying with `file`](https://github.com/docker-mailserver/docker-mailserver/pull/4578#discussion_r2484793244) (_which I didn't seem to get around to showing for compressed zstd mail in my review since I last worked on it_).

---

I'll apply some of the suggestions made but need to sort out some other tasks before I can return back to this.

docs/content/examples/tutorials/mailbox-compression.md
polarathene: Unresolved for visibility.

georglauterbach:
@audioscavenger It's been quite tough for us to review PRs, but I'm currently going through the list. If you could give the current feedback a chance, I'd be willing to review and merge :)

georglauterbach:
Also please rebase and adjust to Dovecot 2.4.

georglauterbach:
Implemented the requested review changes and rebased `patch-2` onto `origin/master`.

- Rewrote the guide for DMS 16 / Dovecot 2.4 using the global `mail_compress` named list and `zstd`.
- Removed the obsolete Dovecot 2.3 configuration and codec tables.
- Covered LMTP/IMAP/POP3 verification, stored-message `file` verification, and a global optional `default_vsz_limit`.
- Added the Unreleased documentation changelog entry.
- Squashed the branch into `a23bd596` and force-pushed it to the PR fork.

The strict MkDocs build passes. I left the review threads unresolved for maintainer confirmation.

georglauterbach (approved):

PR #4581, opened 2025-10-02
fix: `accounts.sh` - Support first valid local account

polarathene:
# Description

- Ensure catch-all alias entries (eg: `@example.test`) are also skipped. Adding a dummy account for catch-all will not match a `quota-status` query to Dovecot by Postfix for a recipient.
- When there are multiple addresses provided, they will now be iterated through by the `,` delimiter, instead of as a single value that fails. This way the first valid aliased address to a local Dovecot mailbox account will now be used for the dummy alias account.
- The common logic for extracting the quota user attribute is now split out to a common function call.

## Type of change

## Checklist

### DIFF
--- a/target/scripts/helpers/accounts.sh
+++ b/target/scripts/helpers/accounts.sh
@@ -36,16 +36,7 @@ function _create_accounts() {
       USER=$(echo "${LOGIN}" | cut -d @ -f1)
       DOMAIN=$(echo "${LOGIN}" | cut -d @ -f2)
 
-      # test if user has a defined quota
-      if [[ -f /tmp/docker-mailserver/dovecot-quotas.cf ]]; then
-        declare -a USER_QUOTA
-        IFS=':' read -r -a USER_QUOTA < <(grep "${USER}@${DOMAIN}:" -i /tmp/docker-mailserver/dovecot-quotas.cf)
-
-        if [[ ${#USER_QUOTA[@]} -eq 2 ]]; then
-          USER_ATTRIBUTES="${USER_ATTRIBUTES:+${USER_ATTRIBUTES} }userdb_quota_rule=*:bytes=${USER_QUOTA[1]}"
-        fi
-      fi
-
+      USER_ATTRIBUTES="$(_add_attribute_dovecot_quota "${LOGIN}" "${USER_ATTRIBUTES}")"
       if [[ -z ${USER_ATTRIBUTES} ]]; then
         _log 'debug' "Creating user '${USER}' for domain '${DOMAIN}'"
       else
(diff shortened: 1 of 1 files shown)


target/scripts/helpers/accounts.sh
```diff
@@ -82,41 +73,53 @@ function _create_accounts() {
...
+      [[ ! ${ALIAS} =~ .+@.+ ]] && continue
+
+      # ${REAL_FQUN} is a user's fully-qualified username
+      unset REAL_FQUN
+      local REAL_FQUN REAL_USERNAME REAL_DOMAINNAME
+
+      # Support checking multiple aliased addresses (split by `,` delimiter):
+      # - The first local account matched will be associated to the alias
+      # - Does not support resolving FQUN if it were also an alias
+      for FQUN in $(echo "${ALIASED}" | sed "s/,/ /g")
```
polarathene:
```suggestion
      for FQUN in ${ALIASED//,/ }
```

target/scripts/helpers/accounts.sh
```diff
@@ -82,41 +73,53 @@ function _create_accounts() {
...
-# `quota-status` policy service and remain at risk of backscatter.
+# Required when using Dovecot Quotas to avoid blacklisting risk from backscatter.
+# Note: This is a workaround only suitable for basic aliases that map to a single real address,
+# not multiple addresses (real accounts or additional aliases), those will not work with
+# the Postfix `quota-status` policy service and remain at risk of backscatter.
 #
-# see https://github.com/docker-mailserver/docker-mailserver/pull/2248#issuecomment-953313852
-# for more details on this method
+# for more details on this method, see:
+# https://github.com/docker-mailserver/docker-mailserver/pull/2248#issuecomment-953313852
```
polarathene:
```suggestion
# https://github.com/docker-mailserver/docker-mailserver/pull/2248#issuecomment-953313852
```

polarathene:
Tests will fail until a separate PR addresses replacing the LDAP image: https://github.com/docker-mailserver/docker-mailserver/issues/4582 ([PR](https://github.com/docker-mailserver/docker-mailserver/pull/4583))

test/tests/parallel/set3/mta/account_management.bats
```diff
@@ -27,19 +27,16 @@ function teardown_file() { _default_teardown ; }
...
-  assert_line --index 7 'prefixtest@localhost.localdomain'
-  assert_line --index 8 'test@localhost.localdomain'
-  assert_line --index 9 'first-name@localhost.localdomain'
-  assert_line --index 10 'first.name@localhost.localdomain'
-  _should_output_number_of_lines 11
+  assert_line --index 6 'prefixtest@localhost.localdomain'
+  assert_line --index 7 'test@localhost.localdomain'
+  assert_line --index 8 'first-name@localhost.localdomain'
+  assert_line --index 9 'first.name@localhost.localdomain'
+  _should_output_number_of_lines 10
```
polarathene, line 35:
```suggestion
  _should_output_number_of_lines 10
  
  refute_line --partial '@localdomain2.com'
```

target/scripts/helpers/accounts.sh
```diff
@@ -82,41 +73,53 @@ function _create_accounts() {
...
-      unset REAL_ACC USER_QUOTA
-      declare -a REAL_ACC USER_QUOTA
-
-      local REAL_USERNAME REAL_DOMAINNAME
-      REAL_USERNAME=$(cut -d '@' -f 1 <<< "${REAL_FQUN}")
-      REAL_DOMAINNAME=$(cut -d '@' -f 2 <<< "${REAL_FQUN}")
+    local ALIAS ALIASED DOVECOT_USERDB_LINE
+    while read -r ALIAS ALIASED; do
+      # Skip any alias lacking `@` (aka local-part only) or is a catch-all domain (starts with `@`):
+      [[ ! ${ALIAS} =~ .+@.+ ]] && continue
```
georglauterbach, line 93: Easier to read would be

```suggestion
      [[ ${ALIAS} =~ .+@.+ ]] || continue
```

IMO

target/scripts/helpers/accounts.sh
```diff
@@ -82,41 +73,53 @@ function _create_accounts() {
...
+
+      # ${REAL_FQUN} is a user's fully-qualified username
+      unset REAL_FQUN
```
georglauterbach, line 96: Why the `unset`? The `local` later should be enough.

target/scripts/helpers/accounts.sh
```diff
@@ -82,41 +73,53 @@ function _create_accounts() {
...
+      if [[ -z ${REAL_FQUN} ]] || ! grep -q "${REAL_FQUN}" "${DATABASE_ACCOUNTS}"; then
         _log 'debug' "Alias '${ALIAS}' is non-local (or mapped to a non-existing account) and will not be added to Dovecot's userdb"
         continue
       fi
 
       _log 'debug' "Adding alias '${ALIAS}' for user '${REAL_FQUN}' to Dovecot's userdb"
 
+      # Clear possibly already filled arrays (do not remove the following line of code)
+      unset REAL_ACC USER_QUOTA
+      declare -a REAL_ACC USER_QUOTA
```
georglauterbach, line 121: Same here. The `unset` is not required, I'd say, and the `declare -a` is neither. You could use `local` instead. In Bash, you do not need to declare arrays with `declare -a`, only maps (dictionaries) with `declare -A`.

target/scripts/helpers/accounts.sh
casperklein: Normally I also try to avoid negated statements. But in this case my brain processes the first one faster - dunno why 😆
Your variant saves 2 bytes 👍

target/scripts/helpers/accounts.sh
casperklein: I am wondering about the comment, to not remove the unset line.

target/scripts/helpers/accounts.sh
polarathene: - If alias does not match pattern skip adding it
- If alias does match pattern continue processing, otherwise skip it

I'm definitely inclined that the former is easier to read 😅 I think you wrote this statement originally and all I've done is changed the pattern really to avoid the catch-all domains.

target/scripts/helpers/accounts.sh
polarathene: I would have thought so too, but IIRC when I ran it locally it persisted the previous assigned value in earlier loop iterations (if any). `local` scopes the variable assignment within the function body, but doesn't seem to reset it's value within the loop iterations, so it's more like it's a hoisted var AFAIK.

target/scripts/helpers/accounts.sh
polarathene: Given [my feedback about what I had observed with needing `unset REAL_FQUN`](https://github.com/docker-mailserver/docker-mailserver/pull/4581#discussion_r2408989229) when putting this PR together, I wouldn't be surprised if it's for the same reason.

I'd have to trawl through git blame history about the `declare -a` usage, but you're probably right. These were the original lines already so I just left that as-is.

polarathene:
Beyond potentially double checking the `unset` requirement, and potentially adding a test case to ensure the problem is caught should someone remove it and introduce a regression bug, and perhaps a more clearer comment for the reader as to why these lines require caution... I think the PR itself has nothing else blocking?

---

Concerns like these are also incentives towards migrating some of the scripts away from bash (_which I'd look into if we didn't have so much of a backlog as-is elsewhere_). I definitely would like to pursue migrating our `setup` CLI to Rust at some point (perhaps next year), or at the very least the accounts/alias management.

target/scripts/helpers/accounts.sh
georglauterbach: Completely fine with me 👍🏼

georglauterbach:
Will take a look soon and merge if everything is fine

georglauterbach:
> Will take a look soon and merge if everything is fine

Didn't yet get around to it in time it seems; any pressure merging this?

polarathene:
Oh sorry I just pressed the update branch button, didn't notice the auto merge I had enabled 😓

No there was no pressure to merge it. If you have anything you wanted to go over let me know and I can setup another PR.

georglauterbach:
No worries, I'lll go over it soon and if anything comes up, I'll let you know. I do not anticipate any issues, though.

target/scripts/helpers/accounts.sh
georglauterbach: I'm still not completely sure why, but really it's nothing to worry too much about 👍🏼

georglauterbach:
Resolved all conversations, LGTM 👍🏼

PR #4586, opened 2025-10-05
fix: Prevent logs leaking into config files (`FETCHMAIL_PARALLEL=1`)

lkwg82:
# Description

fixes fetchmail in parallel mode, when logging is added as first line

```
root@mail:/etc/fetchmailrc.d# head fetchmail-1.rc 
2025-10-05 15:16:55+02:00 INFO  start-mailserver.sh: File '/etc/fetchmailrc' was missing a final newline - this has been fixed
set daemon 300
set syslog
poll posteo.de with proto IMAP
...
```
results in 

```
mailserver_1  | 2025-10-05 15:16:55+02:00 INFO  start-mailserver.sh: Starting daemons
mailserver_1  | 2025-10-05 15:16:56,859 WARN exited: fetchmail-1 (exit status 5; not expected)
mailserver_1  | 2025-10-05 15:16:57,066 WARN exited: fetchmail-2 (exit status 5; not expected)
mailserver_1  | 2025-10-05 15:16:57+02:00 INFO  start-mailserver.sh: mail.example.com is up and running
```

# explanation:

this reads from STDOUT and adds the log as well to the rc file 

`target/scripts/startup/setup.d/fetchmail.sh` (line >54)
```bash
   while read -r LINE; do
        if [[ ${LINE} =~ poll ]]; then
         ...
          # Just the server settings that need to be added to the specific rc.d file
          echo "${LINE}" >>"${FETCHMAILRCD}/fetchmail-${COUNTER}.rc"
        fi
      done < <(_get_valid_lines_from_file "${FETCHMAILRC}")
```

`target/scripts/helpers/utils.sh`
```bash
# Returns input after filtering out lines that are:
# empty, white-space, comments (`#` as the first non-whitespace character)
function _get_valid_lines_from_file() {
  _convert_crlf_to_lf_if_necessary "${1}"
  _append_final_newline_if_missing "${1}"

  grep --extended-regexp --invert-match "^\s*$|^\s*#" "${1}" || true
}
```

below

```bash
# This is to sanitize configs from users that unknowingly removed the end-of-file LF:
function _append_final_newline_if_missing() {
   ....
      _log 'info' "File '${1}' was missing a final newline - this has been fixed"
  ....

}
```

because of

`target/scripts/helpers/logs.sh`
```bash
function _log() {
  ...
  if [[ ${1} =~ ^(warn|error)$ ]]; then
    echo -e "${MESSAGE}" >&2
  else
    echo -e "${MESSAGE}"  # <----- 
  fi
}
```

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ All notable changes to this project will be documented in this file. The format 
   - Third-party sourced CLI tools updated ([#4557](https://github.com/docker-mailserver/docker-mailserver/pull/4557)):
     - `jaq` from `2.1.0` to [`2.3.0`](https://github.com/01mf02/jaq/releases/tag/v2.3.0)
     - `step` CLI from `0.28.2` to [`0.28.7`](https://github.com/smallstep/cli/releases/tag/v0.28.7))
-  - all logs prints to STDERR (formerly only warning/error) 
+  - DMS logs now all output to STDERR (formerly only warning/error logs) (#[4586](https://github.com/docker-mailserver/docker-mailserver/pull/4586))
 - **Dovecot**
   - Updated the FTS plugin Xapian from `1.9` to [`1.9.1`](https://github.com/grosjo/fts-xapian/releases/tag/1.9.1) which adds Dovecot 2.4 compatibility ([#4557](https://github.com/docker-mailserver/docker-mailserver/pull/4557))
 - **Postfix**
--- a/target/scripts/helpers/log.sh
+++ b/target/scripts/helpers/log.sh
@@ -104,7 +104,9 @@ function _log() {
 
   MESSAGE="$(date --rfc-3339='seconds') ${!LOG_COLOR}${LOG_LEVEL_NAME}${RESET} $(basename "${0}"): ${2}"
 
-  echo -e "${MESSAGE}" >&2  # printing STDOUT would break things :/
+  # All logs should go through to STDERR,
+  # STDOUT is only appropriate for expected program output
+  echo -e "${MESSAGE}" >&2
 }
 
 # Get the value of the environment variable LOG_LEVEL if
--- a/target/scripts/startup/setup.d/fetchmail.sh
+++ b/target/scripts/startup/setup.d/fetchmail.sh
@@ -86,7 +86,7 @@ autorestart=true
 stdout_logfile=/var/log/supervisor/%(program_name)s.log
 stderr_logfile=/var/log/supervisor/%(program_name)s.log
 user=fetchmail
-command=/usr/bin/fetchmail -f ${RC} -v --nodetach --daemon %(ENV_FETCHMAIL_POLL)s -i /var/lib/fetchmail/.fetchmail-${COUNTER}-UIDL-cache --pidfile /var/run/fetchmail/%(program_name)s.pid
+command=/usr/bin/fetchmail -f ${RC} -v --nodetach --daemon %(ENV_FETCHMAIL_POLL)s -i /var/lib/fetchmail/.fetchmail-UIDL-cache --pidfile /var/run/fetchmail/%(program_name)s.pid
 EOF
       chmod 700 "${RC}"
       chown fetchmail:root "${RC}"
--- a/test/tests/parallel/set1/fetchmail.bats
+++ b/test/tests/parallel/set1/fetchmail.bats
@@ -60,11 +60,11 @@ function teardown_file() {
   _should_not_have_in_config 'pop3.third-party.test. '   /etc/fetchmailrc.d/fetchmail-2.rc
 }
 
-@test "(ENV FETCHMAIL_PARALLEL=1) should detect info line in fetchmail-x.rc" {
+@test "(ENV FETCHMAIL_PARALLEL=1) should not have logs emitted into fetchmail-x.rc" {
   export CONTAINER_NAME=${CONTAINER2_NAME}
 
   _should_not_have_in_config      'INFO'                      /etc/fetchmailrc.d/fetchmail-1.rc
-  _should_note_have_in_config     'INFO'                      /etc/fetchmailrc.d/fetchmail-2.rc
+  _should_not_have_in_config      'INFO'                      /etc/fetchmailrc.d/fetchmail-2.rc
 }
 
 function _should_have_in_config() {


lkwg82:
I tried to run the tests locally, but I use podman and even after adjusting `Dockerfile` the testsuite fails because of sth with `eth0` (I do have `eno1`)

target/scripts/helpers/log.sh
```diff
@@ -104,11 +104,7 @@ function _log() {
 
   MESSAGE="$(date --rfc-3339='seconds') ${!LOG_COLOR}${LOG_LEVEL_NAME}${RESET} $(basename "${0}"): ${2}"
 
-  if [[ ${1} =~ ^(warn|error)$ ]]; then
-    echo -e "${MESSAGE}" >&2
-  else
-    echo -e "${MESSAGE}"
-  fi
```
polarathene, line 111: I don't think this is the solution we want to accept 😓 

It's a little tricky with how we're relying on `echo` to STDOUT here for both the bash script helper processing/filtering the lines, and the logger.

I think we could resolve this particular concern by having our log messages go through a file, that'd work fine for our actual logs to `/var/log/mail.log` that we `tail` to STDOUT when running a container and that would avoid the concern with the logs being leaked into config. However some `setup` CLI tools are using our same logging support for user facing feedback which would be lost 🤷‍♂️

The helper utility uses `grep => sed => printf` commands and could be adjusted to use a different file-descriptor instead of `/dev/stdout` (default FD) if needed.

However as your change here notes, when you look at the actual scripts involved all other log lines are correctly `warn`. `info` probably made more sense for that line, but it wouldn't be logged either if the user fixes it on their end. We should just switch the `info` to `warn` 👍

https://github.com/docker-mailserver/docker-mailserver/blob/d0a915bd65eb7af6256a40f08b05a1d426f9ed60/target/scripts/helpers/utils.sh#L17-L56

target/scripts/startup/setup.d/fetchmail.sh
```diff
@@ -77,16 +77,16 @@ function _setup_fetchmail_parallel() {
...
+      cat >"/etc/supervisor/conf.d/fetchmail-${COUNTER}.conf" << EOF
 [program:fetchmail-${COUNTER}]
 startsecs=0
 autostart=false
 autorestart=true
 stdout_logfile=/var/log/supervisor/%(program_name)s.log
 stderr_logfile=/var/log/supervisor/%(program_name)s.log
 user=fetchmail
-command=/usr/bin/fetchmail -f ${RC} -v --nodetach --daemon %(ENV_FETCHMAIL_POLL)s -i /var/lib/fetchmail/.fetchmail-UIDL-cache --pidfile /var/run/fetchmail/%(program_name)s.pid
+command=/usr/bin/fetchmail -f ${RC} -v --nodetach --daemon %(ENV_FETCHMAIL_POLL)s -i /var/lib/fetchmail/.fetchmail-${COUNTER}-UIDL-cache --pidfile /var/run/fetchmail/%(program_name)s.pid
```
polarathene: This is an unrelated change, can you please explain the need to set a unique fetchmail UIDL-cache file per instance?

From what I understand a shared cache would work fine given the parallel feature is splitting by `poll` entries? The cache is for client-side tracking of messages already pulled from a server with POP3 to avoid redownloading.

You could add separate cache like this, but I would be curious about the need for that. Should you update your fetchmail config to adjust the `poll` order at any point, this would modify the `COUNTER` variable used here, and thus lose the benefit the cache file was providing no?

test/tests/parallel/set1/fetchmail.bats
```diff
@@ -60,6 +60,13 @@ function teardown_file() {
   _should_not_have_in_config 'pop3.third-party.test. '   /etc/fetchmailrc.d/fetchmail-2.rc
 }
 
+@test "(ENV FETCHMAIL_PARALLEL=1) should detect info line in fetchmail-x.rc" {
+  export CONTAINER_NAME=${CONTAINER2_NAME}
+
+  _should_not_have_in_config      'INFO'                      /etc/fetchmailrc.d/fetchmail-1.rc
+  _should_note_have_in_config     'INFO'                      /etc/fetchmailrc.d/fetchmail-2.rc
+}
```
polarathene: - There is a typo on the 2nd `_should` line, `note` => `not`.
- The test name implies it should detect an info line, while the assertions are that it should not find such a line 😅

```suggestion
@test "(ENV FETCHMAIL_PARALLEL=1) should not have logs emitted into fetchmail-x.rc" {
  export CONTAINER_NAME=${CONTAINER2_NAME}

  _should_not_have_in_config      'INFO'                      /etc/fetchmailrc.d/fetchmail-1.rc
  _should_not_have_in_config      'INFO'                      /etc/fetchmailrc.d/fetchmail-2.rc
}
```

target/scripts/helpers/log.sh
polarathene: I think we should consider adopting my suggestion to change `info` to `warn` here:

https://github.com/docker-mailserver/docker-mailserver/blob/d0a915bd65eb7af6256a40f08b05a1d426f9ed60/target/scripts/helpers/utils.sh#L51

---

However, apparently [logs should all be output to STDERR](https://stackoverflow.com/questions/4919093/should-i-log-messages-to-stderr-or-stdout) regardless? 😅 (_`git` does this for example_).

Although there are plenty of others that have an expectation of [only sending logs worth the users attention to STDERR instead](https://stackoverflow.com/questions/2302315/how-can-info-and-debug-logging-message-be-sent-to-stdout-and-higher-level-messag?noredirect=1&lq=1#comment136770103_2302336), I think it makes sense for all our logs to go to STDERR.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ All notable changes to this project will be documented in this file. The format
   - Third-party sourced CLI tools updated ([#4557](https://github.com/docker-mailserver/docker-mailserver/pull/4557)):
     - `jaq` from `2.1.0` to [`2.3.0`](https://github.com/01mf02/jaq/releases/tag/v2.3.0)
     - `step` CLI from `0.28.2` to [`0.28.7`](https://github.com/smallstep/cli/releases/tag/v0.28.7))
+  - all logs prints to STDERR (formerly only warning/error)
```
polarathene:
```suggestion
  - DMS logs now all output to STDERR (formerly only warning/error logs) (#[4586](https://github.com/docker-mailserver/docker-mailserver/pull/4586))
```

target/scripts/helpers/log.sh
```diff
@@ -104,11 +104,7 @@ function _log() {
...
+  echo -e "${MESSAGE}" >&2  # printing STDOUT would break things :/
```
polarathene:
```suggestion
  # All logs should go through to STDERR,
  # STDOUT is only appropriate for expected program output
  echo -e "${MESSAGE}" >&2
```

target/scripts/startup/setup.d/fetchmail.sh
polarathene:
```suggestion
command=/usr/bin/fetchmail -f ${RC} -v --nodetach --daemon %(ENV_FETCHMAIL_POLL)s -i /var/lib/fetchmail/.fetchmail-UIDL-cache --pidfile /var/run/fetchmail/%(program_name)s.pid
```

polarathene (changes requested):
I am in support of DMS emitting all logs it emits to STDERR 👍

This particular issue could be resolved however by ensuring your config file you supply has a final new line present.

The scripts could also resolve it with:
- Calling the helper prior to the while loop, then reading the file contents in afterwards.
- Adjusting the single `info` log from the helper to be a `warn` level log.

The proposed change is sufficient, so as long as tests pass I'm good with that :)

---

I'll apply the feedback suggestions so we can check all is good in the test suite.

## Reference - Reproduction

```yaml
services:
  dms:
    image: ghcr.io/docker-mailserver/docker-mailserver:15.1
    hostname: mail.example.test
    environment:
      ENABLE_FETCHMAIL: 1
      FETCHMAIL_PARALLEL: 1
    configs:
      - source: dms-accounts
        target: /tmp/docker-mailserver/postfix-accounts.cf
      - source: fetchmail
        target: /tmp/docker-mailserver/fetchmail.cf

configs:
  fetchmail:
    # `|-` instead of `|` will prevent appending a new line to the file content:
    content: |-
      poll 'mail.remote.test' proto imap
        user 'jane.doe@remote.test'
        pass 'secret'
        is 'john.doe@example.test'
        no sslcertck

  dms-accounts:
    content: |
      john.doe@example.test|{SHA512-CRYPT}$$6$$sbgFRCmQ.KWS5ryb$$EsWrlYosiadgdUOxCBHY0DQ3qFbeudDhNMqHs6jZt.8gmxUwiLVy738knqkHD4zj4amkb296HFqQ3yDq4UXt8.
```

```console
# No final new line at user provided config:
$ docker compose exec dms bash -c 'tail -c1 /tmp/docker-mailserver/fetchmail.cf | wc -l'
0

# Internal copy differs (includes globals/defaults and final new line):
$ docker compose exec dms bash -c 'tail -c1 /etc/fetchmail | wc -l'
1

# Log line stored as first line of each split config file generated:
$ docker compose exec dms bash -c 'cat /etc/fetchmailrc.d/*'

2025-10-06 03:32:59+00:00 INFO  start-mailserver.sh: File '/etc/fetchmailrc' was missing a final newline - this has been fixed
set daemon 300
set syslog
poll 'mail.remote.test' proto imap
user 'jane.doe@remote.test'
pass 'secret'
is 'john.doe@example.test'
no sslcertck

# Original unused config:
$ docker compose exec dms bash -c 'cat /etc/fetchmailrc'
# General options

set daemon 300
set syslog

# Fetch rules

poll 'mail.remote.test' proto imap
  user 'jane.doe@remote.test'
  pass 'secret'
  is 'john.doe@example.test'
  no sslcertck
```

---

The internal copy is fine as we don't do a direct copy, this adds a new line for us implicitly:

https://github.com/docker-mailserver/docker-mailserver/blob/d0a915bd65eb7af6256a40f08b05a1d426f9ed60/target/scripts/startup/setup.d/fetchmail.sh#L9-L16

While Parallel feature does something similar, but only after we've processed their user provided config (_**NOTE:** The commentary is a bit inaccurate for the `DEFAULT_FILE` var as well mind you_):

https://github.com/docker-mailserver/docker-mailserver/blob/d0a915bd65eb7af6256a40f08b05a1d426f9ed60/target/scripts/startup/setup.d/fetchmail.sh#L30-L44

https://github.com/docker-mailserver/docker-mailserver/blob/d0a915bd65eb7af6256a40f08b05a1d426f9ed60/target/scripts/startup/setup.d/fetchmail.sh#L55-L62

polarathene:
**NOTE:** I have reverted your [unrelated change for a separate fetchmail cache per instance](https://github.com/docker-mailserver/docker-mailserver/pull/4586#discussion_r2404831645), unless you can provide more information to justify that being necessary it seems like it'd not provide much good?

---

> the testsuite fails because of sth with `eth0` (I do have `eno1`)

[`NETWORK_INTERFACE=eno1`](https://docker-mailserver.github.io/docker-mailserver/v15.1/config/environment/#network_interface) is probably what you'd want, although I don't think our test suite allows for adjusting that 😓

casperklein (approved):

casperklein:
Not related to this PR:

From the naming, I wouldn't expect a `_get_*()` function to alter anything.

lkwg82:
@casperklein it is BASH, the language which is making  use almost only of sideeffects 😁

PR #4492, opened 2025-05-26
docs: Revise docs for the `OVERRIDE_HOSTNAME` ENV

s0ftcorn:
Changed the description of OVERRIDE_HOSTNAME environment variable, to raise awareness of unforeseen side effects.

# Description

The current documentation of `OVERRIDE_HOSTNAME` in `mailserver.env` doesnt reflect the fact that setting this can lead to unforeseen problems. See explanation at the bottom of this comment https://github.com/docker-mailserver/docker-mailserver/issues/4484#issuecomment-2908101121

When setting up DMS the first time i thought it would be better to set it to make sure all internal services use the correct hostname. When in reality its better not to set this unless you need to.

While at it i reformated the text so it looks more like the rest of the comments in the file. Now with no trailing whitespaces.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [X] Improvement (non-breaking change that does improve existing functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my code
- [X] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
I dont think this is necessary since the documentation already tells to only set this when you cant change the hostname.
- [ ] If necessary, I have added tests that prove my fix is effective or that my feature works
No tests necessary, since this only changes a comment.
- [ ] New and existing unit tests pass locally with my changes
No tests necessary, since this only changes a comment.
- [X] I have added information about changes made in this PR to `CHANGELOG.md`

PR #4598, opened 2025-10-29
docs: mention that SPF should be on both EHLO and sender domains

neagix:
# Description

Small documentation update: as far as best practices go, it is currently not mentioned on which domains to put an SPF record.

## Type of change

## Checklist

### DIFF
--- a/docs/content/config/best-practices/dkim_dmarc_spf.md
+++ b/docs/content/config/best-practices/dkim_dmarc_spf.md
@@ -321,6 +321,8 @@ The DMARC status may not be displayed instantly due to delays in DNS (caches). D
 
 ### Adding an SPF Record
 
+It is a best practice to have an SPF record for both your EHLO domain and your sender email address domain.
+
 To add a SPF record in your DNS, insert the following line in your DNS zone:
 
 ```txt


polarathene:
Citation please?

When an MTA receives mail, the sender address domain will be queried for an SPF record to verify that it permits the connected client (MTA/MUA) to send mail on behalf of that domain.

Your DMS FQDN AFAIK should not be relevant beyond that when it comes to SPF checks? What is the reasoning for including an SPF record there if that FQDN is not used in a mails sender address?

neagix:
> Citation please?

I think it's explained here: https://datatracker.ietf.org/doc/html/rfc7208#section-2.3

>   It is RECOMMENDED that SPF verifiers not only check the "MAIL FROM"
>   identity but also separately check the "HELO" identity by applying
>   the check_host() function ([Section 4](https://datatracker.ietf.org/doc/html/rfc7208#section-4)) to the "HELO" identity as the
>   <sender>.  Checking "HELO" promotes consistency of results and can
>   reduce DNS resource usage.  If a conclusive determination about the
>   message can be made based on a check of "HELO", then the use of DNS
>   resources to process the typically more complex "MAIL FROM" can be
>   avoided.  Additionally, since SPF records published for "HELO"
>   identities refer to a single host, when available, they are a very
>   reliable source of host authorization status.  Checking "HELO" before
>   "MAIL FROM" is the RECOMMENDED sequence if both are checked.
>
>   Note that requirements for the domain presented in the EHLO or HELO
>   command are not always clear to the sending party, and SPF verifiers
>   have to be prepared for the identity to be an IP address literal (see
>   [[RFC5321], Section 4.1.3](https://datatracker.ietf.org/doc/html/rfc5321#section-4.1.3)) or simply be malformed.  This SPF check can
>   only be performed when the "HELO" string is a valid, multi-label
>   domain name.
>

polarathene:
You are mistaken.

See section 4, this is about implementing SPF checks (_which DMS defers to [`policyd-spf`](https://github.com/docker-mailserver/docker-mailserver/blob/5865f546ec55d2cc15b3b88e801d7cca724034a2/target/scripts/startup/setup.d/dmarc_dkim_spf.sh#L85-L104) or [rspamd](https://github.com/docker-mailserver/docker-mailserver/blob/5865f546ec55d2cc15b3b88e801d7cca724034a2/target/scripts/startup/setup.d/security/rspamd.sh#L96-L98)_) of the sender domain SPF record where the pseudo-function `check_host()` is defined for implementers.

No separate SPF record is required to be available the HELO FQDN, it is used during verification of the SPF record, to assert the MTA is authorized to send on behalf of that sender domain.

neagix:
> No separate SPF record is required to be available the HELO FQDN, it is used during verification of the SPF record, to assert the MTA is authorized to send on behalf of that sender domain.

This also corresponds to my knowledge until recently: I was receiving DMARC reports about the HELO FQDN and concluded that there was no outgoing email from my server with the HELO FQDN hostname, leaving as only possibility that those mailservers were checking the HELO FQDN's SPF, for some reason.

I need to backtrack the actual reason for those DMARC reports.

polarathene:
> I need to backtrack the actual reason for those DMARC reports.

They may be performing rDNS check. DMS (or Postfix rather) greets the mail server it's sending mail to _"Hi I'm `mail.example.com`!"_, and for an rDNS check the mail server can look at the IP from the DMS connection associated to that greeting, and see what FQDN that responds with.

The rDNS result might differ from the DMS FQDN, but unless the mail server is very strict that shouldn't matter AFAIK. The FQDN DMS greeted with can then be queried via DNS to get an IP, and that should match the IP that the rDNS also resolves back to. Something like that 😅

You can look over [this information](https://github.com/docker-mailserver/docker-mailserver/issues/3520#issuecomment-1700191973) which shows how to test different scenarios and some caveats to be aware of. There are some scenarios where things may get iffy.

neagix:
My first strategy was to somehow log headers for **ALL** mail leaving my server, this way I hoped to track down if any email was leaving with the mailserver hostname (`mail.example.com`) instead of the configured mailboxes hostname (`example.com`). This however turned out to be more complex than I thought and I gave up; another lead I found today is here: https://github.com/docker-mailserver/docker-mailserver/pull/4492

Since I have been using `hostname:` (not a FQDN) and `domainname:` in `compose.yaml`, and `OVERRIDE_HOSTNAME` (the actual FQDN for the mailserver), I thought that perhaps in some cases this setup is letting mail (for example, SRS forwaded mail) out with a wrong sender mailbox hostname.

E.g. old setup:
```yaml
version: '3'
services:
  mailserver:
    # OVERRIDE_HOSTNAME set to mail.example.com
    hostname: mail
    domainname: example.com
```

New setup:
```yaml
version: '3'
services:
  mailserver:
    # OVERRIDE_HOSTNAME not set
    hostname: mail.example.com
```

I will check if this leads to any improvement on the issue of SPF checks on the mail server's hostname (`mail.example.com`).

polarathene:
> Since I have been using `hostname:` (not a FQDN) and `domainname:` in `compose.yaml`, and `OVERRIDE_HOSTNAME` (the actual FQDN for the mailserver), I thought that perhaps in some cases this setup is letting mail (for example, SRS forwaded mail) out with a wrong sender mailbox hostname.

We specifically cover in our docs and `compose.yaml` example to only use `hostname` not `domainname` (_this is only meant for NIS domain_). Do you have a reason you're using `domainname`, or like many others accidentally misunderstood it?

Some distro differ in opinion for a hostname being a single label (valid for mDNS with Avahi advertising) or multi-label as a full FQDN. I had done some experiments there and pretty sure I identified that FQDN was more appropriate/compatible.

Additionally if you do not configure `hostname` properly, this may cause bugs with expectations from DMS and it's services.
- This setting will affect `/etc/hostname`, `/etc/hosts`, and `/etc/resolv.conf` (_this will point to `127.0.0.11:53` when using user-defined networks like is default with Docker Compose, but not Docker CLI, important for internal DNS resolution_). This affects the kernel namespace hostname, which some software uses via glibc to get information IIRC, without that it'll get the random hexadecimal container ID instead which is problematic.
- `OVERRIDE_HOSTNAME` is a workaround for container engines that lack the ability to set the containers hostname. It's not meant to be paired with `hostname`, and only allows us to control some configs with DMS having awareness of the intended DMS FQDN, but presently not making efforts to add an entry into `/etc/hosts` (_which NSS uses via glibc for some resolution_) and likewise the kernel namespace would have the container ID (_which gets matched to `/etc/hosts` IIRC and without the `/etc/hosts` modified by us can make some services unhappy when inferring the FQDN for the HELO/EHLO, such as with fetchmail and swaks_).

---

> New setup:
> 
> ```yaml
> version: '3'
> services:
>   mailserver:
>     # OVERRIDE_HOSTNAME not set
>     hostname: mail.example.com
> ```

It should. You also should not need `version: '3'` if you are running Docker Compose v2 (_which has been available for years, so you're most likely on the 2.x series_).

Can you clarify where you got such config? Our docs should not be advising such, typically it seems to come from AI/LLMs generating config for users 😕

neagix:
> We specifically cover in our docs and `compose.yaml` example to only use `hostname` not `domainname` (_this is only meant for NIS domain_). Do you have a reason you're using `domainname`, or like many others accidentally misunderstood it?

I made this configuration in 2023, I was following the docs back then.

> Can you clarify where you got such config? Our docs should not be advising such, typically it seems to come from AI/LLMs generating config for users 😕

No, I did this myself; I remember I started with hostname + domainname and then at some point added `OVERRIDE_HOSTNAME`.

In any case, I changed it now so I will be monitoring if the problem of forwarded mail with wrong hostname disappears.

PR #4606, opened 2025-11-13
tests: fix _get_container_ip for docker 29.0

ap-wtioit:
# Description

Docker 29 removes NetworkSettings.IPAddress
* https://github.com/moby/moby/pull/50846
* https://github.com/moby/moby/pull/50848

<details>

<summary>Docker 28.4.0 inspect of container using bridge</summary>

`docker inspect ...`
```yaml
[
    {
        "Id": "2a426fc9f069d3849efbca89c661d41f4ae06687efd35883c32053590538cd91",
        # ...
        "NetworkSettings": {
            "Bridge": "",
            "SandboxID": "9528a9bf4462ab058779d206bc3072263b4351cb7870305a4d6cac504acff6cb",
            "SandboxKey": "/var/run/docker/netns/9528a9bf4462",
            "Ports": {},
            "HairpinMode": false,
            "LinkLocalIPv6Address": "",
            "LinkLocalIPv6PrefixLen": 0,
            "SecondaryIPAddresses": null,
            "SecondaryIPv6Addresses": null,
            "EndpointID": "b915d5a4fd3bdb5fa0c7144dc19dc32c8714ff6be44e9e45f2b1537dc7da7baa",
            "Gateway": "172.17.0.1",
            "GlobalIPv6Address": "",
            "GlobalIPv6PrefixLen": 0,
            "IPAddress": "172.17.0.2",
            "IPPrefixLen": 16,
            "IPv6Gateway": "",
            "MacAddress": "76:74:ff:bd:af:0c",
            "Networks": {
                "bridge": {
                    "IPAMConfig": null,
                    "Links": null,
                    "Aliases": null,
                    "MacAddress": "76:74:ff:bd:af:0c",
                    "DriverOpts": null,
                    "GwPriority": 0,
                    "NetworkID": "2bcc680edb64388426905c0ad51dd2d3d6bd4da0974afad5a0924fdb0ac99367",
                    "EndpointID": "b915d5a4fd3bdb5fa0c7144dc19dc32c8714ff6be44e9e45f2b1537dc7da7baa",
                    "Gateway": "172.17.0.1",
                    "IPAddress": "172.17.0.2",
                    "IPPrefixLen": 16,
                    "IPv6Gateway": "",
                    "GlobalIPv6Address": "",
                    "GlobalIPv6PrefixLen": 0,
                    "DNSNames": null
                }
            }
        }
    }
]

```
</details>

<details>

<summary>Docker 29.0.0 inspect of container using bridge</summary>

`docker inspect ...`
```yaml
[
    {
        "Id": "bf1d685dda76fd70336dcfb03eadd84b6ef9df84b5091c836c48124c7ebdd4ac",
        # ...
        "NetworkSettings": {
            "SandboxID": "abf13e6108cc7eb63b18ae1543522d1c323eb0f67b0f298dae1885cd58ae161d",
            "SandboxKey": "/var/run/docker/netns/abf13e6108cc",
            "Ports": {},
            "Networks": {
                "bridge": {
                    "IPAMConfig": null,
                    "Links": null,
                    "Aliases": null,
                    "DriverOpts": null,
                    "GwPriority": 0,
                    "NetworkID": "a236919fbac09468275c1fa155752a5065208ed35eba0735c21f48438aa06892",
                    "EndpointID": "f51e9b253ee4f00cda4a6834c55cd1c7df506b7d7e772ec9fa0929f6998a6419",
                    "Gateway": "172.17.0.1",
                    "IPAddress": "172.17.0.2",
                    "MacAddress": "ee:96:61:cd:b0:ff",
                    "IPPrefixLen": 16,
                    "IPv6Gateway": "",
                    "GlobalIPv6Address": "",
                    "GlobalIPv6PrefixLen": 0,
                    "DNSNames": null
                }
            }
        }
    }
]

```
</details>

Test error on our CI when having a docker 29.0.0 runner:
```
...
ok 54 [Fail2Ban] fail2ban-jail.cf overrides in 997ms
not ok 55 [Fail2Ban] ban ip on multiple failed login in 26ms
# (in test file test/tests/parallel/set1/spam_virus/fail2ban.bats, line 73)
#   `CONTAINER1_IP=$(_get_container_ip "${CONTAINER1_NAME}")' failed
# template parsing error: template: :1:19: executing "" at <.NetworkSettings.IPAddress>: map has no entry for key "IPAddress"
not ok 56 [Fail2Ban] unban ip works in 27ms
# (in test file test/tests/parallel/set1/spam_virus/fail2ban.bats, line 101)
#   `CONTAINER2_IP=$(_get_container_ip "${CONTAINER2_NAME}")' failed
# template parsing error: template: :1:19: executing "" at <.NetworkSettings.IPAddress>: map has no entry for key "IPAddress"
ok 57 [Fail2Ban] bans work properly (single IP) in 1231ms
...
ok 66 [Postgrey] (enabled) should whitelist recipient 'user2@otherdomain.tld' in 161ms
not ok 67 [Postscreen] should fail send when talking out of turn in 33ms
# (from function `setup' in test file test/tests/parallel/set1/spam_virus/postscreen.bats, line 9)
#   `CONTAINER1_IP=$(_get_container_ip "${CONTAINER1_NAME}")' failed
# template parsing error: template: :1:19: executing "" at <.NetworkSettings.IPAddress>: map has no entry for key "IPAddress"
not ok 68 [Postscreen] should successfully pass postscreen and get postfix greeting message (respecting postscreen_greet_wait time) in 27ms
# (from function `setup' in test file test/tests/parallel/set1/spam_virus/postscreen.bats, line 9)
#   `CONTAINER1_IP=$(_get_container_ip "${CONTAINER1_NAME}")' failed
# template parsing error: template: :1:19: executing "" at <.NetworkSettings.IPAddress>: map has no entry for key "IPAddress"
ok 69 [Rspamd] (DKIM) log level is applied correctly in 573ms
...
```

## Type of change

## Checklist

Info @wt-io-it

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ All notable changes to this project will be documented in this file. The format 
 - **Internal:**
   - `ENABLE_QUOTAS=1` - When an alias has multiple addresses, the first local mailbox address found will be used for the Dovecot dummy account workaround ([#4581](https://github.com/docker-mailserver/docker-mailserver/pull/4581))
 - **Tests:**
-  - `_get_container_ip` - Add compatibility with docker 29
+  - Make the helper method `_get_container_ip()` compatible with Docker 29 ([#4606](https://github.com/docker-mailserver/docker-mailserver/pull/4606))
 
 ### Removed
 


test/helper/common.bash
```diff
@@ -364,11 +364,16 @@ function _reload_postfix() {
...
 #
 # @param ${1} = container name [OPTIONAL]
 function _get_container_ip() {
   local TARGET_CONTAINER_NAME=$(__handle_container_name "${1:-}")
-  docker inspect --format '{{ .NetworkSettings.IPAddress }}' "${TARGET_CONTAINER_NAME}"
+  # use .NetworkSettings.IPAddress when available (docker < 29 and container uses bridge)
+  # else use .NetworkSettings.Networks.bridge.IPAddress (docker > 29 and container uses bride)
+  # else empty string
+  # with newline at the end
+  docker inspect --format '{{with (index .NetworkSettings "IPAddress")}}{{.}}{{else}}{{with (index .NetworkSettings.Networks "bridge")}}{{.IPAddress}}{{end}}{{end}}' "${TARGET_CONTAINER_NAME}"
```
polarathene: We could do it with the `--format` template, but since it's JSON we might be better off with just using `jq`?

```suggestion
  docker inspect --format '{{with (index .NetworkSettings.Networks "bridge")}}{{.IPAddress}}{{end}}' "${TARGET_CONTAINER_NAME}"
```

vs via `jq`:

```suggestion
  docker inspect "${TARGET_CONTAINER_NAME}" | jq '.[].NetworkSettings.Networks["bridge"].IPAddress'
```

Latter looks simpler to grok to me 🤔

ap-wtioit: well, `jq` could work, the comment tries actually explaining of the fallbacks:
* jq needs `-r` and `// ""`
   ``` bash
   docker inspect container_with_bridge | jq '.[].NetworkSettings.Networks["bridge"].IPAddress'
   # "172.17.0.2"
   docker inspect container_with_bridge | jq -r '.[].NetworkSettings.Networks["bridge"].IPAddress'
   # 172.17.0.2
   docker inspect container_without_bridge | jq -r '.[].NetworkSettings.Networks["bridge"].IPAddress'
   # null
   docker inspect container_without_bridge | jq -r '.[].NetworkSettings.Networks["bridge"].IPAddress // ""'
   # 
   ```
   and that would need the adequate comments for explaining the fallback (as i can understand that most people would find that hard to read). i just documented the cases i could think of were the new codes output is the same as the old code with docker < 29.
   
Should i also change _container_is_running to `jq`?

ap-wtioit: And also introducing a pipe would mean we should check the error handling:
```bash
docker inspect unknown_container | jq -r '.[].NetworkSettings.Networks["bridge"].IPAddress // ""' || echo "FAIL"
#
#error: no such object: unknown_container
docker inspect --format '{{with (index .NetworkSettings "IPAddress")}}{{.}}{{else}}{{with (index .NetworkSettings.Networks "bridge")}}{{.IPAddress}}{{end}}{{end}}' unknown_container || echo "FAIL"
#
#error: no such object: unknown_container
#FAIL
```

polarathene: Regarding the `jq` fallback to empty with `// ""`, while the reasoning is good our tests don't seem to use such a fallback when encountering a failure.

```console
# With blank line included (stdout):
$ docker inspect --format '{{ .NetworkSettings.IPAddress }}'

Error: No such object: unknown_container
```

~~I'm not sure why your output example differs there with all lowercase~~ 🤔 (_Oh, just compared to v29+ where it is lowercase, prior to that it was as my output shows_)

> that would need the adequate comments for explaining the fallback (as i can understand that most people would find that hard to read).

I'd agree if we had the `jq` fallback you suggest it would need clarification.

From the looks of it the blank value to stdout probably isn't meaningful to our tests (_`fail2ban.bats` + `postscreen.bats` that use this utility method_), they'd still fail as your [shared error output](https://github.com/docker-mailserver/docker-mailserver/pull/4606#issue-3620086538) demonstrated with stderr being reported instead.

---

> And also introducing a pipe would mean we should check the error handling:

`set -o pipefail` will handle that:

```console
$ set -o pipefail
$ docker inspect unknown_container | jq -r '.[].NetworkSettings.Networks["bridge"].IPAddress // ""' || echo "FAIL"
error: no such object: unknown_container

FAIL
$ echo $?
1
```

That said there's quite a bit of discouragement online for using pipefail broadly/globally, so that's a bit of a bummer.

`jq` does have an option to output an exit code for invalid input though:

> `-e` / `--exit-status`: Sets the exit status of `jq` to
> - `0` if the last output values was neither `false` nor `null`
> - `1` if the last output value was either `false` or `null`
> - `4` if no valid result was ever produced.

```console
$ docker inspect unknown_container | jq -r --exit-status '.[].NetworkSettings.Networks["bridge"].IPAddress'
error: no such object: unknown_container
$ echo $?
4
```

---

> Should i also change `_container_is_running()` to `jq`?

I don't think that's necessary other than for consistency. My main gripe was the template syntax for `docker inspect --format` is less intuitive vs the JQ equivalent:

```bash
docker inspect --format '{{with (index .NetworkSettings.Networks "bridge")}}{{.IPAddress}}{{end}}' "${TARGET_CONTAINER_NAME}"
```

vs

```bash
docker inspect "${TARGET_CONTAINER_NAME}" | jq -r --exit-status '.[].NetworkSettings.Networks["bridge"].IPAddress'
```

At least regarding the response is an array of objects, with some nested field access. In this case with `jq` I've added a bit of extra complexity for the `bridge` key since it won't work well if there's a hypen in the network name should this change in future (_I assume that might be a similar case with your template equivalent syntax having a simpler form?_), the `jq` query could just as easily be `.[].NetworkSettings.Networks.bridge.IPAddress`.

`_container_is_running()` is simple enough to grok:

https://github.com/docker-mailserver/docker-mailserver/blob/b410e56aa922d32c3ff5289cbfa761af4ccc4238/test/helper/common.bash#L377-L380

vs

```bash
[[ $(docker inspect "${TARGET_CONTAINER_NAME}" | jq -re '.[].State.Running') == 'true' ]]
```

polarathene (changes requested):
I personally prefer the `jq` approach with:

```bash
docker inspect "${TARGET_CONTAINER_NAME}" | jq -r --exit-status '.[].NetworkSettings.Networks["bridge"].IPAddress'
```

I do appreciate the concerns and effort towards replicating the existing behaviour prior to Docker v29 though 👍

I think for the most part it's largely myself and @georglauterbach that work on the test suite, so he may want to weigh in here too, otherwise I think `jq` approach is preferred unless I've missed addressing any other concern?

georglauterbach:
Thank you for spottend this @ap-wtioit! It seems like the latest changes with Docker cause havoc wherever I look.. I'll also review this today, please wait with changes (jq vs no jq) until I took a loo :)

test/helper/common.bash
georglauterbach: After reading through the changes and this discussion, I am in favor of not using `jq` but keeping the changes as they are. They are mostly similar in quality IMHO but what `jq` gains in terms of lower complexity for its query it loses IMO in terms of error handling and cases with `""` etc.

I think the query that this PR introduces is easy enough to understand.

polarathene: > what `jq` gains in terms of lower complexity for its query it loses IMO in terms of error handling and cases with `""` etc.

There is no need to emit a blank output, the default `null` is fine since the error still can be output via `--exit-status` (_causing the test to fail, while the earlier `docker inspect` stderr should be captured for the error message_).

> I am in favor of not using `jq` but keeping the changes as they are.

I disagree with "changes as they are", but I'll compromise.

We do not need the much longer `docker inspect` proposed here with two branches based on field detection when the latter is sufficient.

polarathene: So I was correct with the assumption that the syntax proposed was for if the network name had an unsupported character such as a hypen present, we can simplify it down to this direct look up for `bridge`:

```suggestion
  docker inspect --format '{{ .NetworkSettings.Networks.bridge.IPAddress }}' "${TARGET_CONTAINER_NAME}"
```

Otherwise if this network name were to differ (_which would require the helper function to support configuration of the network name_), this is the equivalent to what I proposed with `jq`:

```
docker inspect --format '{{ (index .NetworkSettings.Networks "example-net").IPAddress }}' "${TARGET_CONTAINER_NAME}"
```

I still personally feel `jq` was more intuitive for maintainers vs Go template syntax when we need to tackle concerns like that, but I could be mistaken 🤔 As @georglauterbach preferred the approach without `jq` (_which the test suite already depends on elsewhere btw_), this is my compromise unless there's any valid reason to carry more complexity that all the extra commentary for context in understanding the syntax is necessary.

polarathene (changes requested):

test/helper/common.bash
georglauterbach: By all means, I am completely fine with `jq` here; you seem to be certain that it's an improvement, hence I have zero objection to using `jq`! I was just fine with the initial patch set too, that's all.

polarathene: The last suggestion will do for now, but if the test suite is ever migrated to Docker Compose instead of just the Docker CLI, we'll probably need to refactor to support customizing the network name, it's not too bad without `jq` as shown above 👍

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ All notable changes to this project will be documented in this file. The format
 
 - **Internal:**
   - `ENABLE_QUOTAS=1` - When an alias has multiple addresses, the first local mailbox address found will be used for the Dovecot dummy account workaround ([#4581](https://github.com/docker-mailserver/docker-mailserver/pull/4581))
+- **Tests:**
+  - `_get_container_ip` - Add compatibility with docker 29
```
polarathene:
```suggestion
  - Make the helper method `_get_container_ip()` compatible with Docker 29 ([#4606](https://github.com/docker-mailserver/docker-mailserver/pull/4606))
```

Issue #4579, opened 2025-09-23
bug report: Rspamd crashing in loop from `setup config dkim`


polarathene:
### 📝 Preliminary Checks

- [x] I tried searching for an existing issue and followed the [debugging docs](https://docker-mailserver.github.io/docker-mailserver/latest/config/debugging/) advice, but still need assistance.

### 👀 What Happened?

When running `setup config dkim` using Rspamd, I noticed sometimes it was failing like this:

```console
$ setup config dkim selector dkim

2025-09-23 00:22:27+00:00 INFO  rspamd-dkim: Creating DKIM keys of type 'rsa' and length '2048' with selector 'dkim' for domain 'example.test'
2025-09-23 00:22:27+00:00 INFO  rspamd-dkim: Successfully created DKIM keys
2025-09-23 00:22:27+00:00 INFO  rspamd-dkim: Supplying a default configuration (to '/tmp/docker-mailserver/rspamd/override.d/dkim_signing.conf')
rspamd: ERROR (not running)
rspamd: ERROR (abnormal termination)
2025-09-23 00:22:28+00:00 ERROR rspamd-dkim: Unexpected error occured :: script = /usr/local/bin/rspamd-dkim  | function = _setup_default_signing_conf | command = supervisorctl restart rspamd | line = 260 | exit code = 7
```

Similarly I could sometimes reproduce this with just:

```console
$ supervisorctl restart rspamd
rspamd: ERROR (not running)
rspamd: ERROR (abnormal termination)

$ echo $?
7
```

If you then checked if the process was running via `ps -aux` or `supervisorctl status rspamd` it would often appear to be running, but the PID was constantly changing (_along with the uptime not incrementing_), so this was stuck in a loop crashing...

~~There is no log files for rspamd being captured from the looks of it~~ 🤔 (_**EDIT:** Nevermind, see below update, they were configured for a different location_)

Here we can see the crash loop occurring:

```console
$ cat /var/log/supervisor/supervisord.log

2025-09-23 00:11:44,401 WARN process 'rspamd' already running
2025-09-23 00:18:08,747 WARN stopped: rspamd (terminated by SIGTERM)
2025-09-23 00:18:16,472 WARN stopped: rspamd (terminated by SIGTERM)
2025-09-23 00:22:12,557 WARN stopped: rspamd (terminated by SIGTERM)
2025-09-23 00:22:28,265 WARN process 'rspamd' already running
[...]

PR #4627, opened 2025-12-20
fix: LDAP - Enable `check-for-changes.sh` support

FDHoho007:
# Description

Enables the change detector for account provisioner ldap. This is required to watch for changes in the SSL certificates and reload postfix when needed. Therefore the change detector needs to be aware whether he should also watch the file databases for changes.

Fixes #4528 

## Type of change

- Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (README.md or the documentation under `docs/`)
- [x] If necessary, I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] **I have added information about changes made in this PR to `CHANGELOG.md`**

PR #4632, opened 2026-01-03
fix: Reload the `rspamd` service instead of restarting it

scristian71:
# Description

Context: https://github.com/docker-mailserver/docker-mailserver/issues/4579#issuecomment-3322117837

Fixes the failure by requesting Rspamd reload configuration via sending a `HUP` signal to the main `rspamd` process. `supervisorctl restart rspamd` was prone to a race condition causing tests to fail and sometimes putting the rspamd service into a bad state.

As per review below, the decision was to switch send the signal via `supervisorctl signal <signal> <service name>`.

## Type of change

## Checklist

- [] My code follows the style guidelines of this project

### DIFF
--- a/target/scripts/helpers/utils.sh
+++ b/target/scripts/helpers/utils.sh
@@ -190,6 +190,6 @@ function _env_var_expect_integer() {
 }
 
 function _reload_rspamd() {
-  _log 'debug' 'Reloading configuration for Rspamd via sending `SIGHUP`'
+  _log 'debug' "Reloading configuration for Rspamd via sending 'SIGHUP'"
   supervisorctl signal SIGHUP rspamd
 }


polarathene:
You should provide more context with the PR, appears to be as identified here: https://github.com/docker-mailserver/docker-mailserver/issues/4579#issuecomment-3322117837

End of that comment I am unsure of the current PR approach resolving the problem properly due to change detection still being racey (_although the issue should be less frequently encountered by dropping the redundant restart_). I proposed for change detection to be refactored with adopting `watchexec` in future that would resolve the concern.

scristian71:
In rspamd is restarted twice, once in _setup_default_signing_conf and the second time in _final_steps as it is mentioned in the referenced issue.
In my case I have failed tests:
https://github.com/scristian71/docker-mailserver/actions/runs/20618435610
And also if I run locally "[Rspamd] (DKIM)" I had random failures.

polarathene:
**TL;DR:**
- Please try and verify the following change suggestions (_changing from restart to [reload via `SIGHUP`](https://github.com/rspamd/rspamd/issues/2959#issuecomment-3712505897)_):
  - Change:
    https://github.com/docker-mailserver/docker-mailserver/blob/0f5763ce48eba83f7065f09de246a09b80d377b2/target/bin/rspamd-dkim#L277-L284
    to:
    ```bash
    function _final_steps() {
      # Request Rspamd to reload it's config immediately (instead of waiting on change detection):
      pkill --signal SIGHUP $(pgrep --oldest rspamd)

      _log 'trace' 'Finished DKIM key creation'
    }
    ```
  - Change:
    https://github.com/docker-mailserver/docker-mailserver/blob/0f5763ce48eba83f7065f09de246a09b80d377b2/target/scripts/check-for-changes.sh#L209-L210
    to:
    ```bash
    _log 'debug' 'Reloading Rspamd configuration'
    pkill --signal SIGHUP $(pgrep --oldest rspamd)
    ```

---

> In my case I have failed tests:
> https://github.com/scristian71/docker-mailserver/actions/runs/20618435610
> And also if I run locally "[Rspamd] (DKIM)" I had random failures.

~~We [recently landed a change](https://github.com/docker-mailserver/docker-mailserver/pull/4627#issuecomment-3691994402) that improved the rspamd test to avoid a failure that started appearing.~~

**EDIT:** I see your test run was from the merge commit of that PR, so the same spurious failure detailed in #4579 and referenced in https://github.com/docker-mailserver/docker-mailserver/pull/4627#issuecomment-3692722708

Change detection is presently at a default 2s poll rate, Rspamd can take a few seconds to complete a restart IIRC. So depending on how close the two restarts are from change detection and the `setup config dkim` script, any of the relevant [`rspamd_dkim.bats` test cases](https://github.com/docker-mailserver/docker-mailserver/blob/0f5763ce48eba83f7065f09de246a09b80d377b2/test/tests/parallel/set1/spam_virus/rspamd_dkim.bats#L98) may still fail.

**EDIT:** Sending a `SIGHUP` signal to Rspamd for it to reload config might avoid the concern.

---

> In rspamd is restarted twice, once in _setup_default_signing_conf and the second time in _final_steps as it is mentioned in the referenced issue.

Yes, but it will still be possible that it is triggered and causes this failure even with your PR here.

Here is our change detection logic for rspamd support, it will detect the DKIM files created, then at the end of the snippet restart the rspamd service:

https://github.com/docker-mailserver/docker-mailserver/blob/0f5763ce48eba83f7065f09de246a09b80d377b2/target/scripts/check-for-changes.sh#L187-L212

The DKIM keys generated into the DMS Config Volume `rspamd/dkim/` dir will trigger that:

https://github.com/docker-mailserver/docker-mailserver/blob/0f5763ce48eba83f7065f09de246a09b80d377b2/target/bin/rspamd-dkim#L163-L176

Just depends on the timing 🤔 

The logic for both were contributed for DMS v13 back in 2023, where I raised some concerns:
- Sending a `SIGHUP` signal would be better than a restart to reload Rspamd instead of a full service restart. There was a [concern expressed about `SIGHUP` not being supported](https://github.com/docker-mailserver/docker-mailserver/pull/3599#discussion_r1370940923) and it's [not entirely clear if that's still valid in 2026](https://github.com/rspamd/rspamd/issues/2959#issuecomment-3712505897).
- The [change detection concern was also raised](https://github.com/docker-mailserver/docker-mailserver/pull/3599#discussion_r1371197797) at the time. There [was a reply](https://github.com/docker-mailserver/docker-mailserver/pull/3599#discussion_r1371492343) but it was focused on the `dkim_signing.conf` file only, the DKIM keys still result in a trigger.

---

So technically both restarts in this Rspamd DKIM script could be avoided I think? (_LDAP is now enabled for change detection support, so I think that'll go smoothly_). Doing so would make the following commentary a bit redundant?:

https://github.com/docker-mailserver/docker-mailserver/blob/0f5763ce48eba83f7065f09de246a09b80d377b2/target/bin/rspamd-dkim#L253-L257

We may still want to go ahead with the immediate copy though, but instead of a restart a `SIGHUP` could be sent to request rspamd to reload it's configs via `pkill --signal SIGHUP $(pgrep --oldest rspamd)` (_**Side-note:** We don't have `rspamd` covered in our [service restart tests](https://github.com/docker-mailserver/docker-mailserver/blob/0f5763ce48eba83f7065f09de246a09b80d377b2/test/tests/parallel/set3/container_configuration/process_check_restart.bats#L165)_)

target/scripts/check-for-changes.sh
```diff
@@ -206,8 +206,8 @@ function _rspamd_changes() {
       _log 'trace' 'Rspamd - DKIM files updated'
     fi
 
-    _log 'debug' 'Rspamd configuration has changed - restarting service'
-    supervisorctl restart rspamd
+    _log 'debug' 'Reloading Rspamd configuration'
+    pkill --signal SIGHUP $(pgrep --oldest rspamd)
```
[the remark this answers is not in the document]
scristian71: All the explanation makes sense as the service restarting should be synncronized in some way.
I did the requested changes in the branch but now there is another issue:
https://github.com/scristian71/docker-mailserver/actions/runs/20763050368/job/59622717076
Will try to analyze in the next days.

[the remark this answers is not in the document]
ap-wtioit: The new issue looks to be:
* RSpamd has a wrong config and pkill "reloads" the config but fails to check the status (rspamd is dead)
* User fixes the config
* now a pkill cannot send a signal to a not existing process, and we need to probably start rspamd with `supervisorctl start rspamd`

(We are also constantly failing the "[Rspamd] (DKIM) argument 'keysize' is applied correctly for RSA keys"  test on our CI)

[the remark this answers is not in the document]
ap-wtioit: Also pkill does not accept pids, when using pgrep to get the pid use kill:

image

target/bin/rspamd-dkim
```diff
@@ -275,10 +272,8 @@ function _transform_public_key_file_to_dns_record_contents() {
 }
 
 function _final_steps() {
```
ap-wtioit, line 274: This is what seems to work for us:
```
function _final_steps() {
  local rspamd_pid
  rspamd_pid=$(pgrep --oldest rspamd) || rspamd_pid=""
  if [[ -z ${rspamd_pid} ]]; then
    supervisorctl start rspamd
  elif ! kill -s SIGHUP "${rspamd_pid}"; then
    _log 'error' "Failed to send SIGHUP to Rspamd (PID: ${rspamd_pid})"
  fi
  # check status of rspamd after a possible reload (that could have killed rspamd in case of a breaking config)
  supervisorctl status rspamd
  _log 'trace' 'Finished DKIM key creation'
}
```

target/scripts/check-for-changes.sh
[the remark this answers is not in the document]
ap-wtioit: And here i changed it to:
```
    local rspamd_pid
    rspamd_pid=$(pgrep --oldest rspamd) || rspamd_pid=""
    if [[ -z ${rspamd_pid} ]]; then
      _log 'debug' 'Starting Rspamd after configuration change'
      supervisorctl start rspamd
    else
      _log 'debug' 'Reloading Rspamd configuration'
      if ! kill -s SIGHUP "${rspamd_pid}"; then
        _log 'error' "Failed to send SIGHUP to Rspamd (PID: ${rspamd_pid})"
      fi
    fi
```

target/bin/rspamd-dkim
```diff
@@ -275,11 +272,16 @@ function _transform_public_key_file_to_dns_record_contents() {
...
-  if ! supervisorctl restart rspamd; then
-    _log 'warn' 'Could not restart Rspamd via Supervisord'
+  # Request Rspamd to reload it's config immediately (instead of waiting on change detection):
+  local rspamd_pid
+  rspamd_pid=$(pgrep --oldest rspamd) || rspamd_pid=""
+  if [[ -z ${rspamd_pid} ]]; then
+    supervisorctl start rspamd
+  elif ! kill -s SIGHUP "${rspamd_pid}"; then
+    _log 'error' "Failed to send SIGHUP to Rspamd (PID: ${rspamd_pid})"
   fi
```
polarathene: This whole if / else and variable should be replaced with a function call:

```suggestion
  _reload_rspamd
```

target/scripts/check-for-changes.sh
```diff
@@ -206,8 +206,17 @@ function _rspamd_changes() {
...
+    rspamd_pid=$(pgrep --oldest rspamd) || rspamd_pid=""
+    if [[ -z ${rspamd_pid} ]]; then
+      _log 'debug' 'Starting Rspamd after configuration change'
+      supervisorctl start rspamd
+    else
+      _log 'debug' 'Reloading Rspamd configuration'
+      if ! kill -s SIGHUP "${rspamd_pid}"; then
+        _log 'error' "Failed to send SIGHUP to Rspamd (PID: ${rspamd_pid})"
+      fi
+    fi
```
polarathene: Likewise:

```suggestion
    _reload_rspamd
```

scristian71:
I have added the changes suggested by @ap-wtioit and the tests are passing now:

https://github.com/scristian71/docker-mailserver/actions/runs/20793106916

target/bin/rspamd-dkim
```diff
@@ -275,11 +272,16 @@ function _transform_public_key_file_to_dns_record_contents() {
 }
 
 function _final_steps() {
-  # We need to restart Rspamd so the changes take effect immediately.
-  if ! supervisorctl restart rspamd; then
-    _log 'warn' 'Could not restart Rspamd via Supervisord'
+  # Request Rspamd to reload it's config immediately (instead of waiting on change detection):
+  local rspamd_pid
+  rspamd_pid=$(pgrep --oldest rspamd) || rspamd_pid=""
+  if [[ -z ${rspamd_pid} ]]; then
```
casperklein:
```suggestion
  rspamd_pid=$(supervisorctl pid rspamd)
  if (( rspamd_pid == 0 )); then
```

casperklein: I think this is a bit cleaner. If rspamd is running, the PID is returned, otherwise 0.

target/bin/rspamd-dkim
```diff
@@ -275,11 +272,16 @@ function _transform_public_key_file_to_dns_record_contents() {
...
+    supervisorctl start rspamd
+  elif ! kill -s SIGHUP "${rspamd_pid}"; then
+    _log 'error' "Failed to send SIGHUP to Rspamd (PID: ${rspamd_pid})"
   fi
-
+  # check status of rspamd after a possible reload (that could have killed rspamd in case of a breaking config)
+  supervisorctl status rspamd
```
casperklein: This doesn't work as you think. If rspamd fails to start due an invalid configuration, supervisor will immediately start rspam again (in a loop).
The output will almost always be "running".

target/bin/rspamd-dkim
polarathene: **EDIT:** Not sure why this is listed under my posted review, it's a [reply to earlier feedback](https://github.com/docker-mailserver/docker-mailserver/pull/4632#discussion_r2670173050).

---

I'm not sure if it's better since a condition is still required? `supervisorctl pid <service name>` is neat, but the main difference is instead of an empty value on failure it returns a `0` 🤔

Then again, it's clearer for intent than the `pgrep` command, and might actually be better for our older test case on process restarts where I noted a few services that were less than obvious when piecing that together that extra context was required for maintenance.

Why the `(( rspamd_pid == 0 ))` condition vs `[[ rspamd_pid -eq 0 ]]`? (_which is what we typically use elsewhere in our scripts AFAIK?_)

polarathene:
This is in response to [@ap-wtioit review input](https://github.com/docker-mailserver/docker-mailserver/pull/4632#discussion_r2667906630), but on the main thread to keep the added information below visible.

> Also pkill does not accept pids, when using pgrep to get the pid use kill

Whoops that's my bad sorry 😓

I was referencing [some test logic I wrote a while back](https://github.com/docker-mailserver/docker-mailserver/blob/0f5763ce48eba83f7065f09de246a09b80d377b2/test/tests/parallel/set3/container_configuration/process_check_restart.bats#L160-L167) but didn't pay attention to how the PID was used.

---

This is additional context for my benefit on the topic. No need to read below this unless it interests you 👍 

## Notes

Using the `rspamd` process name with `--oldest` should still work well, just `pgrep` itself isn't necessary:

```console
$ pgrep --oldest rspamd
1360

$ pgrep --list-full rspamd
1360 rspamd: main process
1639 rspamd: rspamd_proxy process (127.0.0.1:11332)
1640 rspamd: rspamd_proxy process (127.0.0.1:11332)
1641 rspamd: controller process (0.0.0.0:11334)
1642 rspamd: hs_helper process

# Reload (NOTE: `--echo` produces the output lines instead of being silent):
$ pkill --echo --signal SIGHUP rspamd
rspamd killed (pid 1360)
rspamd killed (pid 1639)
rspamd killed (pid 1640)
rspamd killed (pid 1641)
rspamd killed (pid 1642)

# Only sends signal to the main rspamd process:
# (NOTE: This still respawns the child processes)
$ pkill --oldest --echo --signal SIGHUP rspamd
rspamd killed (pid 1360)

# Original PID 1360 wasn't actually "killed":
# (Child processes were respawned)
$ pgrep --full --list-full rspamd
1360 rspamd: main process
2115 rspamd: rspamd_proxy process (127.0.0.1:11332)
2116 rspamd: rspamd_proxy process (127.0.0.1:11332)
2117 rspamd: controller process (0.0.0.0:11334)
2118 rspamd: hs_helper process

# Process tree (`pstree` via `psmisc` package):
$ pstree -p -s "$(pgrep --oldest rspamd)"
dumb-init(1)---supervisord(7)---rspamd(1360)-+-rspamd(2115)
                                             |-rspamd(2116)
                                             |-rspamd(2117)
                                             `-rspamd(2118)
```

**NOTE:** The increasingly larger PID of the child processes appears to be from `check-for-changes.sh` calling `sleep 2` in a loop with each invocation incrementing the PID (_along with any other commands I use via CLI_).

Related logs (_different PIDs from another run_), completes within 1 second:

<details>
<summary>Rspamd SIGHUP reload event logs</summary>

```
2026-01-07 20:59:33 #635(main) <e96fe3>; main; rspamd_hup_handler: rspamd 3.12.1 is requested to reload configuration
# ...
2026-01-07 20:59:34 #635(main) <e96fe3>; main; reread_config: config has been reread successfully
2026-01-07 20:59:34 #635(main) <e96fe3>; main; rspamd_hup_handler: spawn workers with a new config
2026-01-07 20:59:34 #635(main) <e96fe3>; main; spawn_workers: worker of type fuzzy(localhost:11335) is disabled in the config, skip spawning
2026-01-07 20:59:34 #635(main) <e96fe3>; main; rspamd_fork_worker: prepare to fork process rspamd_proxy (0); listen on: 127.0.0.1:11332
2026-01-07 20:59:34 #635(main) <e96fe3>; main; rspamd_fork_worker: prepare to fork process rspamd_proxy (1); listen on: 127.0.0.1:11332
2026-01-07 20:59:34 #1261(rspamd_proxy) <e96fe3>; main; rspamd_worker_set_limits: use system max file descriptors limit: 1024KiB cur and 1024KiB max
2026-01-07 20:59:34 #1261(rspamd_proxy) <e96fe3>; main; rspamd_worker_set_limits: use system max core size limit: -1B cur and -1B max
2026-01-07 20:59:34 #635(main) <e96fe3>; main; rspamd_fork_worker: prepare to fork process controller (0); listen on: 0.0.0.0:11334
2026-01-07 20:59:34 #1262(rspamd_proxy) <e96fe3>; main; rspamd_worker_set_limits: use system max file descriptors limit: 1024KiB cur and 1024KiB max
2026-01-07 20:59:34 #1262(rspamd_proxy) <e96fe3>; main; rspamd_worker_set_limits: use system max core size limit: -1B cur and -1B max
2026-01-07 20:59:34 #635(main) <e96fe3>; main; spawn_workers: worker of type normal(127.0.0.1:11333) is disabled in the config, skip spawning
2026-01-07 20:59:34 #635(main) <e96fe3>; main; rspamd_fork_worker: prepare to fork process hs_helper (0), no bind socket
2026-01-07 20:59:34 #1263(controller) <e96fe3>; main; rspamd_worker_set_limits: use system max file descriptors limit: 1024KiB cur and 1024KiB max
2026-01-07 20:59:34 #1263(controller) <e96fe3>; main; rspamd_worker_set_limits: use system max core size limit: -1B cur and -1B max
2026-01-07 20:59:34 #1263(controller) <19ntq8>; controller; rspamd_controller_password_sane: your normal password is not encrypted, we strongly recommend to replace it with the encrypted one
2026-01-07 20:59:34 #1263(controller) <19ntq8>; controller; rspamd_controller_password_sane: enable password is not set, so you should filter controller availability by using of firewall or `secure_ip` option
2026-01-07 20:59:34 #635(main) <e96fe3>; main; rspamd_hup_handler: workers spawning has been finished
2026-01-07 20:59:34 #635(main) <e96fe3>; main; rspamd_hup_handler: kill old workers
2026-01-07 20:59:34 #635(main) <e96fe3>; main; kill_old_workers: send signal to worker 806
2026-01-07 20:59:34 #635(main) <e96fe3>; main; kill_old_workers: send signal to worker 805
2026-01-07 20:59:34 #635(main) <e96fe3>; main; kill_old_workers: send signal to worker 804
2026-01-07 20:59:34 #635(main) <e96fe3>; main; kill_old_workers: send signal to worker 807
2026-01-07 20:59:34 #1264(hs_helper) <e96fe3>; main; rspamd_worker_set_limits: use system max file descriptors limit: 1024KiB cur and 1024KiB max
2026-01-07 20:59:34 #1264(hs_helper) <e96fe3>; main; rspamd_worker_set_limits: use system max core size limit: -1B cur and -1B max
2026-01-07 20:59:34 #635(main) <e96fe3>; main; rspamd_check_termination_clause: hs_helper process 807 terminated normally
2026-01-07 20:59:34 #635(main) <e96fe3>; main; rspamd_cld_handler: do not respawn process hs_helper after found terminated process with pid 807
2026-01-07 20:59:34 #635(main) <e96fe3>; main; rspamd_check_termination_clause: rspamd_proxy process 804 terminated normally
2026-01-07 20:59:34 #635(main) <e96fe3>; main; rspamd_cld_handler: do not respawn process rspamd_proxy after found terminated process with pid 804
2026-01-07 20:59:34 #635(main) <e96fe3>; main; rspamd_check_termination_clause: controller process 806 terminated normally
2026-01-07 20:59:34 #635(main) <e96fe3>; main; rspamd_cld_handler: do not respawn process controller after found terminated process with pid 806
2026-01-07 20:59:34 #635(main) <e96fe3>; main; rspamd_check_termination_clause: rspamd_proxy process 805 terminated normally
2026-01-07 20:59:34 #635(main) <e96fe3>; main; rspamd_cld_handler: do not respawn process rspamd_proxy after found terminated process with pid 805
```

</details>

---

### Rspamd service management and failure handling

If Rspamd service itself is not running, I don't think we need to start the service.

Supervisord handles restarting the main process when the service was started during container init, and it will handle restarting if the rspamd process is killed abruptly too (_which is what the linked test case covers, albeit it is missing rspamd coverage it should still be handled the same_).

> - RSpamd has a wrong config and pkill "reloads" the config but fails to check the status (rspamd is dead)

Rspamd shouldn't be dead?

The actions run failure linked by @scristian71 has test cases fail with:

```
ERROR rspamd-dkim: Unexpected error occurred :: script = /usr/local/bin/rspamd-dkim  | function = _final_steps | command = pkill --signal SIGHUP $(pgrep --oldest rspamd) | line = 276 | exit code = 1
```

Not too different from the error at the start of [my associated report](https://github.com/docker-mailserver/docker-mailserver/issues/4579#issue-3443334238) (_minus the earlier log lines related to a `supervisorctl restart rspamd`_):

```
rspamd: ERROR (not running)
rspamd: ERROR (abnormal termination)
ERROR rspamd-dkim: Unexpected error occured :: script = /usr/local/bin/rspamd-dkim  | function = _setup_default_signing_conf | command = supervisorctl restart rspamd | line = 260 | exit code = 7
```

However the current failure of (`exit code = 1`) is occurring because (as you mentioned) the failure for `pkill` to kill by PID input:

```console
$ pkill --signal SIGHUP 123
$ echo $?
1
```

I also tried to trigger a bad config failure, but this didn't affect the `rspamd` process in anyway. Even with a reload invoked.. when using `rspamadm configdump` it showed the bad config but no failure (_possibly because the tool just dumps config_), nor was anything showing up in `/var/log/mail/rspamd.log` (_it would probably log errors if I triggered the functionality for the `/etc/rspamd/override.d/multimap.conf` invalid config I added_).

I'm lacking further context on where bad config is causing rspamd to die in our test suite 😅

---

### Side-Note - Accidental `supervisord` reload impact (`SIGHUP` to PID 1)

If we ran into a scenario that caused `pkill --signal SIGHUP ''` (_by getting an empty `pgrep` subshell response with the quote wrapping advised from AI suggestions_), this would actually restart everything in the container as `supervisord` receives the `SIGHUP` signal. That in turn brings down Rspamd temporarily.

```console
$ docker compose up -d --force-recreate
$ docker compose exec dms bash

# When no PID is given, the empty string input sends the signal to every process:
# (thus the `supervisord` process, as an empty string matches all processes)
$ pkill --signal SIGHUP ""

# The previous command also booted you out of the container (bash process reloaded):
$ docker compose exec dms bash

# `supervisord` reloads due to `SIGHUP`, this restarts the DMS container init script
# which results in rspamd and other services being started again...
# - New rspamd process running for 17s
# - NOTE: If you ran this command too early, it'll still have the existing rspamd service time
$ ps -p "$(pgrep --oldest rspamd)" -o etimes=
     17
```

Inspecting `/var/log/supervisor/supervisord.log` we can see this is the case:

```
2026-01-07 21:16:27,106 WARN exited: mailserver (terminated by SIGHUP; not expected)
2026-01-07 21:16:27,164 WARN exited: update-check (terminated by SIGHUP; not expected)
2026-01-07 21:16:27,164 WARN exited: changedetector (terminated by SIGHUP; not expected)
2026-01-07 21:16:27,165 WARN received SIGHUP indicating restart request
2026-01-07 21:16:54,648 WARN stopped: cron (terminated by SIGTERM)
```

After the `pkill` command boots you out of container, we can go back in and see how this affects the `rspamd` service that `supervisord` manages:

```console
$ supervisorctl status rspamd
error: <class 'xmlrpc.client.Fault'>, <Fault 6: 'SHUTDOWN_STATE'>: file: /usr/lib/python3.11/xmlrpc/client.py line: 668

# Try again:
$ supervisorctl status rspamd
rspamd                           STOPPED   Not started

# Try again:
$ supervisorctl status rspamd
rspamd                           RUNNING   pid 6271, uptime 0:00:01
```

**EDIT:** Turns out it wasn't just PID 1 and forwarding to `supervisord` that our `ENTRYPOINT` + `CMD` instructions in the `Dockerfile` configure, but all processes are matched and received the `SIGHUP` 😬

<details>
<summary>Console output for context</summary>

```console
$ pkill --echo --signal SIGHUP ''

dumb-init killed (pid 1)
supervisord killed (pid 6)
tail killed (pid 7768)
cron killed (pid 7908)
rsyslogd killed (pid 7912)
dovecot killed (pid 7918)
anvil killed (pid 7923)
log killed (pid 7924)
config killed (pid 7925)
update-check.sh killed (pid 7927)
redis-server killed (pid 7953)
sleep killed (pid 7966)
rspamd killed (pid 7967)
pidproxy killed (pid 7973)
postfix-script killed (pid 7977)
check-for-chang killed (pid 8004)
rspamd killed (pid 8006)
rspamd killed (pid 8007)
rspamd killed (pid 8008)
rspamd killed (pid 8011)
master killed (pid 8093)
pickup killed (pid 8094)
qmgr killed (pid 8095)
bash killed (pid 8096)
sleep killed (pid 9126)
```

</details>

I could see how the `supervisord` reload would result in encountering rspamd in a bad state, or the process not being visible briefly (or appearing dead).

`pkill --signal SIGHUP $(pgrep --oldest rspamd)` still results in a exit code of `1` just the same though.. it's only a bigger concern if it was given an empty input with via quotes:

```console
# Error (exit code 2):
$ pkill --signal SIGHUP $(pgrep --oldest rspamdx)
pkill: no matching criteria specified
Try `pkill --help' for more information.

# The `pgrep` subshell returns an empty return value,
# all PID are matched and given the `SIGHUP` signal:
$ pkill --signal SIGHUP "$(pgrep --oldest rspamdx)"
```

target/bin/rspamd-dkim
ap-wtioit: Ok than my local tests were not doing what i thought was going on. I was on the wrong track there. I mistook the pkill not working with pids (that i discovered later) for one of the negative tests leaving the config in a dirty state.

ap-wtioit:
I was about to write a comment for the _reload_rspamd utils method. When i read https://github.com/Supervisor/supervisor/issues/53 once again and came to https://github.com/Supervisor/supervisor/issues/53#issuecomment-162905711. It seems there is a `supervisorctl signal SIGHUP rspamd` that should do what we need to reload Rspamd.

Edit: confirmed that for us the tests are working with `supervisorctl signal SIGHUP rspamd` in both target/bin/rspamd-dkim and target/scripts/check-for-changes.sh

scristian71:
I have applied the suggested changes. Locally the tests are passing.

target/bin/rspamd-dkim
casperklein: The reason was not to avoid the condition. "grepping" for something that we (supervisord) already know seemed odd to me.

> Why the `(( rspamd_pid == 0 ))` condition vs `[[ rspamd_pid -eq 0 ]]`? (which is what we typically use elsewhere in our scripts AFAIK?)

That was just a mistake. Personally I use `(( ))` for pure arithmetic stuff. You are right, in DMS, we should stick to the style we typically use elsewhere 👍

georglauterbach (approved):
Didn't read the whole conversation, but the changes look sensible.

target/bin/rspamd-dkim
```diff
@@ -275,11 +272,10 @@ function _transform_public_key_file_to_dns_record_contents() {
...
 function _final_steps() {
-  # We need to restart Rspamd so the changes take effect immediately.
-  if ! supervisorctl restart rspamd; then
-    _log 'warn' 'Could not restart Rspamd via Supervisord'
-  fi
-
+  # Request Rspamd to reload it's config immediately (instead of waiting on change detection):
+  _reload_rspamd
+  # check status of rspamd after a possible reload (that could have killed rspamd in case of a breaking config)
+  supervisorctl status rspamd
```
polarathene: This should no longer be necessary, the scenario of rspamd being dead due to failed config I need an example of for reproduction, especially if that is somehow due to config generation from this DKIM script.

```suggestion

```

target/scripts/helpers/utils.sh
```diff
@@ -188,3 +188,8 @@ function _env_var_expect_integer() {
   _log 'warn' "The value of '${ENV_VAR_NAME}' is not an integer ('${!ENV_VAR_NAME}'), but was expected to be"
   return 1
 }
+
+function _reload_rspamd() {
+  _log 'debug' 'Reloading Rspamd configuration via SIGHUP.'
```
polarathene:
```suggestion
  _log 'debug' 'Reloading configuration for Rspamd via sending `SIGHUP`'
```

polarathene (changes requested):
~~I'll apply these~~ (nevermind, not given permission to do so)

In addition to these final changes, please add a Changelog entry just above this section:

https://github.com/docker-mailserver/docker-mailserver/blob/0f5763ce48eba83f7065f09de246a09b80d377b2/CHANGELOG.md?plain=1#L11-L13

Copy/paste this:

```markdown
- **Rspamd:**
  - Configuration changes now trigger a service reload instead of a restart ([#4632](https://github.com/docker-mailserver/docker-mailserver/pull/4632))
```

target/scripts/helpers/utils.sh
```diff
@@ -188,3 +188,8 @@ function _env_var_expect_integer() {
   _log 'warn' "The value of '${ENV_VAR_NAME}' is not an integer ('${!ENV_VAR_NAME}'), but was expected to be"
   return 1
 }
+
+function _reload_rspamd() {
+  _log 'debug' 'Reloading configuration for Rspamd via sending `SIGHUP`'
```
polarathene: Linter rule is mistaking the backticks as an attempt to use an expression 🙄 We could disable that rule here or just adjust quotes like this:

```suggestion
  _log 'debug' "Reloading configuration for Rspamd via sending 'SIGHUP'"
```

casperklein (approved):

polarathene (approved):

Issue #1024, opened 2018-08-21
FreeIPA Authentication


hijak:
Hi

Has any one had any luck with authing with freeipa? is there a custom schema im missing ?

hijak:
ive kinda cracked it ill have a guide somewhere

sspreitzer:
FreeIPA is either its totally own API, which then enables also resolving AD trust users, etc.. Or it can be used as a normal LDAP provider. 

Solving the FreeIPA API could be done via a sssd running and serving PAM and NSS to the runnning contiainer daemons.

See my #1021 for Kerberos authentication feature request.

aendeavor:
This issue was closed due to one or more of the following reasons:

1. Age
2. Contributor inactivity
3. The issue seems to be resolved

If you think this happened by accident, or feel like this issue was not actually resolved, please feel free to re-open it. If there is an issue you could resolve in the meantime, please open a PR based on the current master branch so we can review it.

padrino121:
@hijak Did you end up sorting this out, any feedback? Configuring LDAP now but having issues getting the filters right.

PR #4633, opened 2026-01-03
feat: Add Kerberos (GSSAPI) Login support via SASL

MatthieuCoder:
# Description

This is a PR adding Kerberos login support to docker-mailserver - using a keytab

## Type of change

## Checklist

### DIFF
--- a/mailserver.env
+++ b/mailserver.env
@@ -451,16 +451,21 @@ OAUTH2_INTROSPECTION_URL=
 # --- Kerberos Section --------------------------
 # -----------------------------------------------
 
+# empty => Kerberos authentication is disabled
+# 1 => Kerberos authentication is enabled
 ENABLE_KERBEROS=
 
-KERBEROS_AUTH_MECHANISMS=
-
+# the keytab to be used for Kerberos authentication
+# empty => /etc/krb5.keytab
 KERBEROS_KRB5_KEYTAB=
 
+# the default krb5 realm
 KERBEROS_AUTH_DEFAULT_REALM=
 
+# the allowed kerberos realms
 KERBEROS_AUTH_REALMS=
 
+# the hostname to be used for kerberos authentication
 KERBEROS_AUTH_GSSAPI_HOSTNAME=
 
 # -----------------------------------------------
--- a/mailserver.env
+++ b/mailserver.env
@@ -457,7 +457,7 @@ ENABLE_KERBEROS=
 
 # the keytab to be used for Kerberos authentication
 # empty => /etc/krb5.keytab
-KERBEROS_KRB5_KEYTAB=
+KERBEROS_AUTH_KRB5_KEYTAB=
 
 # the default krb5 realm
 KERBEROS_AUTH_DEFAULT_REALM=
--- a/mailserver.env
+++ b/mailserver.env
@@ -448,6 +448,22 @@ ENABLE_OAUTH2=
 OAUTH2_INTROSPECTION_URL=
 
 # -----------------------------------------------
+# --- Kerberos Section --------------------------
+# -----------------------------------------------
+
+ENABLE_KERBEROS=
+
+KERBEROS_AUTH_MECHANISMS=
+
+KERBEROS_KRB5_KEYTAB=
+
+KERBEROS_AUTH_DEFAULT_REALM=
+
+KERBEROS_AUTH_REALMS=
+
+KERBEROS_AUTH_GSSAPI_HOSTNAME=
+
+# -----------------------------------------------
 # --- LDAP Section ------------------------------
 # -----------------------------------------------
 
--- a/target/dovecot/10-auth.conf
+++ b/target/dovecot/10-auth.conf
@@ -68,7 +68,7 @@
 # Host name to use in GSSAPI principal names. The default is to use the
 # name returned by gethostname(). Use "$ALL" (with quotes) to allow all keytab
 # entries.
-#auth_gssapi_hostname =
+auth_gssapi_hostname = "$ALL"
 
 # Kerberos keytab to use for the GSSAPI mechanism. Will use the system
 # default (usually /etc/krb5.keytab) if not specified. You may need to change
@@ -127,5 +127,6 @@ auth_mechanisms = plain login
 #!include auth-checkpassword.conf.ext
 #!include auth-vpopmail.conf.ext
 #!include auth-static.conf.ext
+#!include auth-kerberos.inc
 
 #!include auth-master.inc
--- a/target/dovecot/auth-ldap.conf.ext
+++ b/target/dovecot/auth-ldap.conf.ext
@@ -6,6 +6,13 @@
 
 passdb {
   driver = ldap
+  args = /etc/dovecot/dovecot-ldap.conf.ext
+  mechanisms = gssapi
+  default_fields = nopassword=y
+}
+
+passdb {
+  driver = ldap
   mechanisms = plain login
 
   # Path for LDAP configuration file, see example-config/dovecot-ldap.conf.ext
--- a/target/dovecot/auth-ldap.conf.ext
+++ b/target/dovecot/auth-ldap.conf.ext
@@ -6,13 +6,6 @@
 
 passdb {
   driver = ldap
-  args = /etc/dovecot/dovecot-ldap.conf.ext
-  mechanisms = gssapi
-  default_fields = nopassword=y
-}
-
-passdb {
-  driver = ldap
   mechanisms = plain login
 
   # Path for LDAP configuration file, see example-config/dovecot-ldap.conf.ext
--- a/target/dovecot/dovecot-ldap.conf.ext
+++ b/target/dovecot/dovecot-ldap.conf.ext
@@ -6,7 +6,7 @@ uris                = ldap://mail.example.com
 tls                 = no
 ldap_version        = 3
 pass_attrs          = uniqueIdentifier=user,userPassword=password
-pass_filter         = (&(objectClass=PostfixBookMailAccount)(uniqueIdentifier=%n))
+pass_filter         = (&(objectClass=PostfixBookMailAccount)(|(uniqueIdentifier=%u)(uniqueIdentifier=%n)(krbPrincipalName=%u)))
 user_attrs          = mailHomeDirectory=home,mailUidNumber=uid,mailGidNumber=gid,mailStorageDirectory=mail
-user_filter         = (&(objectClass=PostfixBookMailAccount)(uniqueIdentifier=%n))
+user_filter         = (&(objectClass=PostfixBookMailAccount)(|(uniqueIdentifier=%u)(uniqueIdentifier=%n)(krbPrincipalName=%u)))
 auth_bind           = no
--- a/target/dovecot/dovecot-ldap.conf.ext
+++ b/target/dovecot/dovecot-ldap.conf.ext
@@ -5,7 +5,7 @@ dnpass              = admin
 uris                = ldap://mail.example.com
 tls                 = no
 ldap_version        = 3
-pass_attrs          = uniqueIdentifier=user,userPassword=password
+pass_attrs          = uniqueIdentifier=user,userPassword=password,krbPrincipalName=k5principals
 pass_filter         = (&(objectClass=PostfixBookMailAccount)(uniqueIdentifier=%n))
 user_attrs          = mailHomeDirectory=home,mailUidNumber=uid,mailGidNumber=gid,mailStorageDirectory=mail
 user_filter         = (&(objectClass=PostfixBookMailAccount)(uniqueIdentifier=%n))
--- a/target/dovecot/dovecot-ldap.conf.ext
+++ b/target/dovecot/dovecot-ldap.conf.ext
@@ -6,7 +6,7 @@ uris                = ldap://mail.example.com
 tls                 = no
 ldap_version        = 3
 pass_attrs          = uniqueIdentifier=user,userPassword=password
-pass_filter         = (&(objectClass=PostfixBookMailAccount)(|(uniqueIdentifier=%u)(uniqueIdentifier=%n)(krbPrincipalName=%u)))
+pass_filter         = (&(objectClass=PostfixBookMailAccount)(uniqueIdentifier=%n))
 user_attrs          = mailHomeDirectory=home,mailUidNumber=uid,mailGidNumber=gid,mailStorageDirectory=mail
-user_filter         = (&(objectClass=PostfixBookMailAccount)(|(uniqueIdentifier=%u)(uniqueIdentifier=%n)(krbPrincipalName=%u)))
+user_filter         = (&(objectClass=PostfixBookMailAccount)(uniqueIdentifier=%n))
 auth_bind           = no
--- a/target/dovecot/auth-kerberos.inc
+++ b/target/dovecot/auth-kerberos.inc
@@ -2,7 +2,7 @@
 auth_mechanisms = $auth_mechanisms gssapi
 
 userdb {
-  driver = static
+  driver = ldap
   mechanisms = gssapi
   default_fields = uid=docker gid=docker home=/var/mail/%d/%u/home/
 }
--- a/target/dovecot/auth-kerberos.inc
+++ b/target/dovecot/auth-kerberos.inc
@@ -1,0 +1,8 @@
+# Allow clients to use these additional mechanisms:
+auth_mechanisms = $auth_mechanisms gssapi
+
+userdb {
+  driver = static
+  mechanisms = gssapi
+  default_fields = uid=docker gid=docker home=/var/mail/%d/%u/home/
+}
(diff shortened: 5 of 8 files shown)


mailserver.env
```diff
@@ -447,6 +447,27 @@ ENABLE_OAUTH2=
...
+
+# the default krb5 realm
+KERBEROS_AUTH_DEFAULT_REALM=
+
+# the allowed kerberos realms
+KERBEROS_AUTH_REALMS=
+
+# the hostname to be used for kerberos authentication
+KERBEROS_AUTH_GSSAPI_HOSTNAME=
+
```
polarathene, line 470: Only `ENABLE_KERBEROS` is a relevant ENV here. The others would belong in a dedicated docs section on this auth method.

Additionally your equivalent ENV here is lacking the `AUTH_` component to match the associated `auth_krb5_keytab` setting in the config.

target/dovecot/10-auth.conf
```diff
@@ -68,7 +68,7 @@
 # Host name to use in GSSAPI principal names. The default is to use the
 # name returned by gethostname(). Use "$ALL" (with quotes) to allow all keytab
 # entries.
-#auth_gssapi_hostname =
+auth_gssapi_hostname = "$ALL"
```
polarathene, line 71: I'm not familiar with this, when is `gethostname()` a no go?

target/dovecot/auth-ldap.conf.ext
```diff
@@ -4,6 +4,13 @@
 # This prevents unnecessary auth failure logs triggering Fail2Ban when
 # additional passdb are enabled (OAuth2).
 
+passdb {
+  driver = ldap
+  args = /etc/dovecot/dovecot-ldap.conf.ext
+  mechanisms = gssapi
+  default_fields = nopassword=y
+}
+
```
polarathene: This is only relevant when opting into the feature no?

It could be handled via a separate config that is toggled just like with LDAP:

https://github.com/docker-mailserver/docker-mailserver/blob/0f5763ce48eba83f7065f09de246a09b80d377b2/target/scripts/startup/setup.d/ldap.sh#L54-L55

Is this form of auth only relevant in the LDAP context?

target/dovecot/dovecot-ldap.conf.ext
```diff
@@ -5,7 +5,7 @@ dnpass              = admin
 uris                = ldap://mail.example.com
 tls                 = no
 ldap_version        = 3
-pass_attrs          = uniqueIdentifier=user,userPassword=password
+pass_attrs          = uniqueIdentifier=user,userPassword=password,krbPrincipalName=k5principals
```
polarathene, line 8: Is `krbPrincipalName` set to a fixed value here for some reason? Is that not related to `auth_gssapi_hostname`? (_which has some relevance to "GSSAPI principal names"_)

polarathene (commented):
This project tends to be reluctant towards new features without prior discussion/approval, especially when there is low demand from users.

Could you please add context to why Kerberos would be beneficial? (_I'm not familiar with it beyond name, nor do I recall any users requesting it over the past 5 years_) I am interested to know why it may be preferred over existing auth support.

If it were to be approved documentation would need to be added, along with tests. Often that is more effort than just adding a docs example for the benefits of other interested users.

---

The LDAP auth has tests and docs but lacks proper maintainer support over the years, yet has much more demand/adoption from users than Kerberos. So while I appreciate the PR I'm not sure if we want to take this feature on. It'll likely have some breaking changes that'd need to be changed for the Dovecot 2.4 upgrade too 😅 

I've provided an example `compose.yaml` below that might work well enough for you?

polarathene:
## Alternative (_DIY without landing the PR_)

Usually when I demonstrate (like below) how to approach without upstreaming via PR a user is happy with that as it's much less burden for both parties involved 😅

You would have your typical DMS `compose.yaml`:

```yaml
services:
  dms:
    image: ghcr.io/docker-mailserver/docker-mailserver:latest # :15.0
    hostname: mail.example.test
    # ... Any other DMS related config ...
```

You can then append/merge this additional config, or layer it on with:
- CLI flags: `docker compose -f compose.yaml -f <filename> up`
- Implicitly via `compose.override.yaml`
- Docker Compose `includes` feature.

Here is roughly the same functionality as your PR, but as a single file you can carry/reference that doesn't require upstreaming into DMS with docs + tests.

```yaml
services:
  dms:
    # You'd normally use `volumes` here but for simplicity of the example, all config is contained within `compose.yaml`:
    configs:
      - source: dovecot-custom
        target: /tmp/docker-mailserver/dovecot.cf
      - source: dovecot-gssapi-auth
        target: /etc/dovecot/conf.d/auth-kerberos.inc
      - source: dovecot-gssapi-passdb
        target: /etc/dovecot/conf.d/auth-kerberos.conf.ext

# Using the Docker Compose `configs.content` feature instead of volume mounting separate files.
# NOTE: This feature requires Docker Compose v2.23.1 (Nov 2023) or newer:
# https://github.com/compose-spec/compose-spec/pull/446
configs:
  # Enable Kerberos PassDB (Authentication):
  dovecot-custom:
    content: |
      auth_gssapi_hostname = "$ALL"
      !include auth-kerberos.inc
      !include auth-kerberos.conf.ext

  dovecot-gssapi-auth:
    content: |
      auth_gssapi_hostname = mail.example.com
      auth_krb5_keytab = /etc/dovecot/krb5.keytab
      auth_realms = example.com
      auth_default_realm = example.com

  dovecot-gssapi-passdb:
    content: |
      passdb {
        driver = ldap
        args = /etc/dovecot/dovecot-ldap.conf.ext
        mechanisms = gssapi
        default_fields = nopassword=y
      }

  dms-user-patches:
    content: |
      #!/bin/bash
      
      # Ensure custom CA certificates are recognized:
      update-ca-certificates

      # Add GSSAPI support:
      apt-get update && apt-get install dovecot-gssapi

      # Append `,krbPrincipalName=k5principals` to end of `pass_attrs`
      # if the line appears unmodified:
      sed -i -r -e \
        '|^(pass_attrs.*userPassword=password)$|\1,krbPrincipalName=k5principals|' \
        /etc/dovecot/dovecot-ldap-conf.ext
```

If you still needed to use this helper function for ENV based overrides, that would still be possible in `user-patches.sh`, you'd just need to import (source) our `utils.sh` scripts first like this:

```console

_replace_by_env_in_file 'KERBEROS_' '/etc/dovecot/conf.d/auth-kerberos.inc'
```

**NOTE:** The package install in `user-patches.sh` will add a startup delay. We [document similar advice for users interested in Apache SOLR](https://docker-mailserver.github.io/docker-mailserver/latest/examples/tutorials/dovecot-solr/#composeyaml) for FTS functionality. If you need to minimize the delay you can easily extend DMS locally instead like this:

```yaml
services:
  dms:
    hostname: mail.example.com
    # The `image` setting now represents the tag for the local build configured below:
    image: localhost/dms:${DMS_TAG?Must set DMS image tag}
    # Local build (no need to try pull `image` remotely):
    pull_policy: build
    # `Dockerfile` to build is embedded into the Docker Compose config:
    build:
      dockerfile_inline: |
        FROM docker.io/mailserver/docker-mailserver:${DMS_TAG?Must set DMS image tag}
        RUN apt-get update && apt-get install dovecot-gssapi
```

Likewise the `configs` section for Kerberos additions could be added into that image as separate files via `COPY` if you prefer that instead of additional `volumes`/`configs` entries. The related `user-patches.sh` could be a separate script that you have a simpler `user-patches.sh` call too.

The main caveat apart from carrying all that yourself instead of offloading it to the project is potential maintenance burden when upgrading if anything changes internally within DMS that'd affect those configs.

target/dovecot/10-auth.conf
MatthieuCoder: From my understanding, a user could use different dns entries for smtp, imap and pop. These could be included in the keytab but *not* be the actual hostname of the machine, a keytab could include keys to multiple Kerberos principals, hance the $ALL usage

target/dovecot/dovecot-ldap.conf.ext
MatthieuCoder: `krbPrincipalName` is an ldap value containing the principals of the user,
the "k5principals" dovecot setting is used by dovecot to match the gssapi user principal to the actual ldap user (see: https://doc.dovecot.org/2.4.2/core/config/auth/passdb.html#k5principals)

target/dovecot/auth-ldap.conf.ext
MatthieuCoder: This auth is only relevant in the ldap context in order to match principals names to users; It's complementary to the auth; I used the same configuration files because kerberos only requires nopassword (there are no passwords because the user is already authenticated)

MatthieuCoder:
> This project tends to be reluctant towards new features without prior discussion/approval, especially when there is low demand from users.

Yes. I saw that but wanted to contribute either way, I already have a working setup with kerberos.

> Could you please add context to why Kerberos would be beneficial? (_I'm not familiar with it beyond name, nor do I recall any users requesting it over the past 5 years_) I am interested to know why it may be preferred over existing auth support.

Kerberos enables no-password logins to imap, pop and smtp, by using the user's sessions TGT (Ticket Granting Ticket), It's more of a complement to regular ldap login that is pretty well deployed in the enterprise space (ex. Windows Single Sign On).

There have been some issues regarding this, namely #1024 and #1021

> If it were to be approved documentation would need to be added, along with tests. Often that is more effort than just adding a docs example for the benefits of other interested users.

Yes, of course, I sent the PR to gather feedback regarding my changes, and will be documenting everything accordingly.

> The LDAP auth has tests and docs but lacks proper maintainer support over the years, yet has much more demand/adoption from users than Kerberos. So while I appreciate the PR I'm not sure if we want to take this feature on. It'll likely have some breaking changes that'd need to be changed for the Dovecot 2.4 upgrade too 😅

The upgrade of these changes to 2.4 shouldn't really be such a burden given that Kerberos auth *didn't* change 

> I've provided an example `compose.yaml` below that might work well enough for you?

Thanks! But I already have a setup of my own ^^"

mailserver.env
MatthieuCoder: Yes indeed, I had updated my testing configuration but forgot to change it here

target/dovecot/auth-ldap.conf.ext
MatthieuCoder: I will be moving the gssapi passdb to `auth-kerberos.inc`

MatthieuCoder:
I still have several things in my todo for this PR:

- [x] Add SASLAUTHD support
- [x] Enable the use of the keytab for ldap - not possible due to dovecot limitations
- [ ] Figure out a test plan (by changing the date-time during tests ? Kerberos is time sensitive)

georglauterbach:
> > This project tends to be reluctant towards new features without prior discussion/approval, especially when there is low demand from users.
> 
> Yes. I saw that but wanted to contribute either way, I already have a working setup with kerberos.

That's nice, but not the point. You opening the issue is not for you, but for us. I am not against adding this functionality, but we need to make sure we can maintain it and we need to make sure there is enough need to justify maintenance.

MatthieuCoder:
> > > This project tends to be reluctant towards new features without prior discussion/approval, especially when there is low demand from users.
> > 
> > 
> > Yes. I saw that but wanted to contribute either way, I already have a working setup with kerberos.
> 
> That's nice, but not the point. You opening the issue is not for you, but for us.

I have a working setup. This is not for me, I want my work to benefit the community - hence my contribution.

> I am not against adding this functionality, but we need to make sure we can maintain it and we need to make sure there is enough need to justify maintenance.

Maintaining this shouldn't *be* and issue given that the configuration in dovecot do not seems to have went through major changes in the last few *years*.

Regarding usage, I do agree that Kerberos is not *well-known* and is rather niche. His usage is mainly in Microsoft Windows Active Directory and FreeIPA (the one I tested with) and allows the single sign-on using clients like Thunderbird [ref](https://wiki.mozilla.org/Thunderbird:Supported_authentication_methods).

This is mostly used in enterprise environments, but some people (like me) like to run it for their own infrastructure and such.

I pointed out an issue requesting Kerberos / GSSAPI authentication in my earlier messages.

I do not mind my PR being rejected, I offer my contribution, and it's not *my* project at the end.

polarathene:
> I want my work to benefit the community - hence my contribution.

Appreciated, but there has been no one else engaging in this community requesting GSSAPI support for a long time (_your references are from 2018, with one comment in 2020, nothing else for 5+ years_).

Hence maintainers are understandably reluctant to have the burden of looking after the feature as it can add friction for us to support when there's only 1 (or few) users interested in the feature.

This is why [we have a process to defer to a docs contribution first](https://github.com/docker-mailserver/docker-mailserver/issues/4637#issuecomment-3747157031) when viable (_I [already demonstrated how this feature could be supported easily that way](https://github.com/docker-mailserver/docker-mailserver/pull/4633#issuecomment-3707588891)_).

---

> Maintaining this shouldn't _be_ and issue given that the configuration in dovecot do not seems to have went through major changes in the last few _years_.

There's been a notable change between Dovecot 2.3 and Dovecot 2.4.

The burden is usually on support from users seeking help, or if we do run into any issues like we have with LDAP.

Beyond support requests and docs not being in good shape for LDAP, we've not been able to as easily offer feature parity as it's slightly out of sync with the file based provisioning and some differences. Additionally the tests could fail (_we've gone through refactoring in the past where the lack of LDAP knowledge added friction, and the external deps like an LDAP container has required maintenance due to upstreams dropping maintenance support of their images_).

We've also had issues with other features that had been contributed previously such as Dovecot FTS with Xapian (_where we ended up needing to compile the plugin support ourselves_) and Solr (we carried a package with docs that had become stale/unmaintained, the package lacked ARM64 support requiring extra support on our end_).

Your PR already adds to the burden of LDAP support with the changes this PR proposes as it adds additional complexity (not for you but for us). Once we officially support a feature, user support and maintenance is offloaded onto us.

As we will require tests like we do with LDAP, this does add extra burden when maintainers need to interact with the related feature support. A similar feature for [SQL DB support](https://github.com/docker-mailserver/docker-mailserver/issues/4637) offers a bit more value but would also have similar costs for maintainers, along with adding friction to refactoring to some degree in my experience 😓

---

> I do not mind my PR being rejected, I offer my contribution, and it's not _my_ project at the end.

It's small enough that I might be ok with it, but I'm also wanting to step down from my maintainer status at some point, thus the other maintainers would need to approve.

You'd still have docs (_for which I tend to be a bit strict on for quality_), and tests (_that you've already raised is a bit tricky to approach_)... Are you sure you want to push for upstreaming this into DMS straight away vs the much simpler documentation contribution? (_I'm more lax on review when it's an unofficial guide_)

Docs would still be discoverable and benefit any interested users. It can also be as simple as a small FAQ entry (or similar location) to reference a link to the PR here (_or my `compose.yaml` override comment_).

A reference `compose.override.yaml` (_similar to [what I already demonstrated](https://github.com/docker-mailserver/docker-mailserver/pull/4633#issuecomment-3707588891)_), should make it rather simple to add that support?

You could almost avoid the `user-patches.sh`:
- The `dovecot-gssapi` package could potentially be included in our build by default (_on the basis it causes us no burden, otherwise we're free to drop it_).
- `update-ca-certificates` could rather easily get in through a separate PR (_it would still require some mention in docs + test to verify_).

georglauterbach:
> I have a working setup. This is not for me, I want my work to benefit the community - hence my contribution.

Definitely appreciated!
 
> > I am not against adding this functionality, but we need to make sure we can maintain it and we need to make sure there is enough need to justify maintenance.
> 
> Maintaining this shouldn't _be_ and issue given that the configuration in dovecot do not seems to have went through major changes in the last few _years_.

Whether this is the case is up to us - not to you. This project mandates holding the discussion _beforehand_ in an issue.
 
> I do not mind my PR being rejected, I offer my contribution, and it's not _my_ project at the end.

I'd like to see this become an additional page in the documentation, as @polarathene has pointed out. Can you do that @MatthieuCoder?

MatthieuCoder:
> > I have a working setup. This is not for me, I want my work to benefit the community - hence my contribution.
> 
> Definitely appreciated!
> 
> > > I am not against adding this functionality, but we need to make sure we can maintain it and we need to make sure there is enough need to justify maintenance.
> > 
> > 
> > Maintaining this shouldn't _be_ and issue given that the configuration in dovecot do not seems to have went through major changes in the last few _years_.
> 
> Whether this is the case is up to us - not to you. This project mandates holding the discussion _beforehand_ in an issue.
> 
> > I do not mind my PR being rejected, I offer my contribution, and it's not _my_ project at the end.
> 
> I'd like to see this become an additional page in the documentation, as @polarathene has pointed out. Can you do that @MatthieuCoder?

Yes I can do that. Sorry I'm a bit busy currently, I'll get to it in a few weeks

polarathene:
@MatthieuCoder is this still something you want to tackle and will find time for at some point?

MatthieuCoder:
> @MatthieuCoder is this still something you want to tackle and will find time for at some point?

Yes. Might be better to move this to an issue

georglauterbach:
This needs rebasing now and adjustments for Dovecot 2.4. I will then review it

PR #4661, opened 2026-02-08
docs(rspamd): Revise DNSBL advice with example + minor tweaks

### LINKED ISSUE #4654: docs: How setup caching to avoid DNSBL rate limits?
### Subject

I would like to contribute to the project

### Description

Our mailserver constantly hits DNSBL (DNS Block List) rate limits. We only send/receive about 500 emails per day. As anyone encountered the same? Any solutions? Internal DNS caching maybe?
Currently our spam filtering is very ineffective.
```
RCVD_IN_VALIDITY_RPBL_BLOCKED - bl.score.senderscore.com
RCVD_IN_VALIDITY_CERTIFIED_BLOCKED - sa-trusted.bondedsender.org
RCVD_IN_DNSWL_BLOCKED - list.dnswl.org
RCVD_IN_ZEN_BLOCKED_OPENDNS - zen.spamhaus.org
```

georglauterbach:
# Description

Fixes #4654

## Type of change

## Checklist

### DIFF
--- a/docs/content/config/advanced/kubernetes.md
+++ b/docs/content/config/advanced/kubernetes.md
@@ -266,19 +266,20 @@ If using our Helm chart is not viable for you, here is some guidance to start wi
                         privileged: false
                         capabilities:
                           add:
-                            # file permission capabilities
+                            # file permission
                             - CHOWN
                             - FOWNER
                             - MKNOD
                             - SETGID
                             - SETUID
                             - DAC_OVERRIDE
-                            # network capabilities
+                            # network
                             - NET_ADMIN  # needed for F2B
                             - NET_RAW    # needed for F2B
                             - NET_BIND_SERVICE
-                            # miscellaneous  capabilities
+                            # miscellaneous
                             - SYS_CHROOT
+                            - MAC_OVERRIDE
                             - KILL
                           drop: [ALL]
                         seccompProfile:
--- a/docs/content/config/security/rspamd.md
+++ b/docs/content/config/security/rspamd.md
@@ -177,11 +177,17 @@ You can choose to enable ClamAV, and Rspamd will then use it to check for viruse
 
 The [RBL module][rspamd-docs::modules::rbl] is enabled by default. As a consequence, Rspamd will perform DNS lookups to various blacklists. Whether an RBL or a DNSBL is queried depends on where the domain name was obtained: RBL servers are queried with IP addresses extracted from message headers, DNSBL server are queried with domains and IP addresses extracted from the message body ([source][www::rbl-vs-dnsbl]).
 
-!!! danger "Rspamd and DNS Block Lists"
+??? warning "Rspamd & DNS Blocklists"
 
     When the RBL module is enabled, Rspamd will do a variety of DNS requests to (amongst other things) DNSBLs. There are a variety of issues involved when using DNSBLs. Rspamd will try to mitigate some of them by properly evaluating all return codes. This evaluation is a best effort though, so if the DNSBL operators change or add return codes, it may take a while for Rspamd to adjust as well.
 
-    If you want to use DNSBLs, **try to use your own DNS resolver** and make sure it is set up correctly, i.e. it should be a non-public & **recursive** resolver. Otherwise, you might not be able ([see this Spamhaus post][spamhaus::faq::dnsbl-usage]) to make use of the block lists.
+!!! danger "Properly Querying Blocklists"
+
+    If you want to use DNS Blocklists (DNSBLs) properly, **you have to use your own _non-public, recursive_ DNS resolver!**
+
+    Running your own recursive DNS resolver is feasible, but **you need to ensure that it is not available from the internet!** Hosting a public (i.e., available to the internet, also called "open") DNS resolver has serious security implications.
+
+    DNS blocklists will not answer properly if you use public DNS resolvers (like Cloudflare' `1.1.1.1` or Google's `8.8.8.8`) because they are rate-limited. Hence, the issue is not about caching, but about making a direct (recursive) request. Rspamd makes an efforet to handle rate-limiting properly, but you will not have any benefit from using DNS blocklists if you are not using a recursive resolver.
 
 ## Providing Custom Settings & Overriding Settings
 
--- a/docs/content/config/security/rspamd.md
+++ b/docs/content/config/security/rspamd.md
@@ -155,7 +155,7 @@ DMS does not supply custom values for DNS servers (to Rspamd). If you need to us
 
     This setting is enabled to not allow spam to proceed just because DNS requests did not succeed. It could deny legitimate e-mails to pass though too in case your DNS setup is incorrect or not functioning properly.
 
-???+ example "Setup a recursive DNS resolver for DMS to use"
+??? example "Setup a recursive DNS resolver for DMS to use"
 
     This example is specifically focused on how to run a local DNS service capable of recursive resolution to [properly support DNSBL services](#rspamd-module-rbl) such as [SpamHaus][spamhaus::faq::what-is-a-dnsbl].
 
--- a/docs/content/config/security/rspamd.md
+++ b/docs/content/config/security/rspamd.md
@@ -141,7 +141,7 @@ To use the web interface you will need to configure a password, [otherwise you w
 
 ### DNS
 
-DMS does not supply custom values for DNS servers (to Rspamd). If you need to use custom DNS servers, which could be required when using [DNS-based deny/allowlists](#rbls-real-time-blacklists-dnsbls-dns-based-blacklists), you need to adjust [`options.inc`][rspamd-docs::config::global] yourself. Make sure to also read our [FAQ page on DNS servers][docs::faq::dns-servers].
+DMS does not supply custom values for DNS servers (to Rspamd). If you need to use custom DNS servers, which could be required when using [DNS-based deny/allowlists](#rspamd-module-rbl), you need to adjust [`options.inc`][rspamd-docs::config::global] yourself. Make sure to also read our [FAQ page on DNS servers][docs::faq::dns-servers].
 
 !!! warning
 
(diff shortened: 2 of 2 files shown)


docs/content/config/security/rspamd.md
```diff
@@ -4,9 +4,7 @@ title: 'Security | Rspamd'
 
 ## About
 
-Rspamd is a ["fast, free and open-source spam filtering system"][rspamd-web]. DMS integrates Rspamd like any other service. We provide a basic but easy to maintain setup of Rspamd.
-
-If you want to take a look at the default configuration files for Rspamd that DMS adds, navigate to [`target/rspamd/` inside the repository][dms-repo::default-rspamd-configuration]. Please consult the [section "The Default Configuration"](#the-default-configuration) section down below for a written overview.
```
polarathene, line 9: **EDIT:** Associated commit communicates the change (_just lacks context for why it was missing_): https://github.com/docker-mailserver/docker-mailserver/pull/4661/changes/020cd8f8d32eaf49406d4d970cb0c8e1adcb4abe

For context, `#the-default-configuration` referred to the docs page starting from the "Mode of Operation" section, while the referenced section to link to was dropped by me in https://github.com/docker-mailserver/docker-mailserver/commit/bcee78e2c1c310c4815b3e03c5e9278ad8ef1343 (_Feb 2025, part of DMS v15.0_) where I missed dropping the link (or correcting it to the renamed section title "Overview of Rspamd support" / `#overview-of-rspamd-support`.

docs/content/config/security/rspamd.md
```diff
@@ -177,11 +175,17 @@ You can choose to enable ClamAV, and Rspamd will then use it to check for viruse
...
 
-!!! danger "Rspamd and DNS Block Lists"
+??? warning "Rspamd & DNS Blocklists"
 
     When the RBL module is enabled, Rspamd will do a variety of DNS requests to (amongst other things) DNSBLs. There are a variety of issues involved when using DNSBLs. Rspamd will try to mitigate some of them by properly evaluating all return codes. This evaluation is a best effort though, so if the DNSBL operators change or add return codes, it may take a while for Rspamd to adjust as well.
 
-    If you want to use DNSBLs, **try to use your own DNS resolver** and make sure it is set up correctly, i.e. it should be a non-public & **recursive** resolver. Otherwise, you might not be able ([see this Spamhaus post][spamhaus::faq::dnsbl-usage]) to make use of the block lists.
+!!! danger "Properly Querying Blocklists"
+
+    If you want to use DNS Blocklists (DNSBLs) properly, **you have to use your own _non-public, recursive_ DNS resolver!**
```
polarathene: I know we have different writing styles (_yours is more like a conversation and mine tends to be more technical/verbose_), this and the other suggestion are an alternative suggestion if you agree with the adaption.

```suggestion
    To use DNS Blocklists (DNSBLs) properly, DMS must use a **private and recursive** DNS resolver.
```

docs/content/config/security/rspamd.md
```diff
@@ -177,11 +175,17 @@ You can choose to enable ClamAV, and Rspamd will then use it to check for viruse
...
+
+    Running your own recursive DNS resolver is feasible, but **you need to ensure that it is not available from the internet!** Hosting a public (i.e., available to the internet, also called "open") DNS resolver has serious security implications.
+
+    DNS blocklists will not answer properly if you use public DNS resolvers (like Cloudflare' `1.1.1.1` or Google's `8.8.8.8`) because they are rate-limited. Hence, the issue is not about caching, but about making a direct (recursive) request. Rspamd makes an efforet to handle rate-limiting properly, but you will not have any benefit from using DNS blocklists if you are not using a recursive resolver.
```
polarathene: Revised this a fair bit. Please verify it still makes sense?

**Changes:**
- Prior suggestion already indicates the need to use a private recursive resolver, dropped (_serious security implications hasn't provided additional context to the reader_).
- Immediately focuses on rate-limit issue (with typo fix for cloudflare), with clarification on caching a response from a forwarded query is ineffective.
- Final paragraph reiterates to avoid forwarding and instead resolve recursively via the private service. Trailing Rspamd context lacks a meaningful reference for the reader, thus dropped if no additional value can be inferred.

```suggestion
    DNSBL services are rate-limited, thus if your DNS queries are forwarded through a public resolver (_like Cloudflare's `1.1.1.1` or Google's `8.8.8.8`_) caching the DNSBL service responses received from a public DNS resolver will not be reliable when public load has triggered a rate limit.

    Instead of relying on forwarding DNS queries, they must be resolved recursively (directly) via running your own private recursive DNS service.
```

docs/content/config/security/rspamd.md
polarathene: Is there a reason why the old spamhaus link was dropped? (_I haven't checked if it's used elsewhere in the page, otherwise it's entry should be removed from the bottom of the docs too_):

> ```markdown
> Otherwise, you might not be able ([see this Spamhaus post][spamhaus::faq::dnsbl-usage]) to make use of the block lists.
> ```

I know the link itself isn't as meaningful today, it's rather vague but the FAQ resource itself there might be more beneficial than no extra reference for those wanting more insights 😅

docs/content/config/advanced/kubernetes.md
```diff
@@ -266,19 +266,20 @@ If using our Helm chart is not viable for you, here is some guidance to start wi
...
                             - DAC_OVERRIDE
-                            # network capabilities
+                            # network
                             - NET_ADMIN  # needed for F2B
                             - NET_RAW    # needed for F2B
                             - NET_BIND_SERVICE
-                            # miscellaneous  capabilities
+                            # miscellaneous
                             - SYS_CHROOT
+                            - MAC_OVERRIDE
```
polarathene, line 282: Although this is unrelated, since you're adding a capability could you provide a bit more context for it's inclusion? (_should anyone `git blame` in future_)

**EDIT:** Nevermind, context is in scoped commit: https://github.com/docker-mailserver/docker-mailserver/pull/4661/changes/666628cdac617674cbdc66b78298edb498035db1

polarathene (changes requested):
I'm not sure if it resolves the referenced issue exactly.

The docs are definitely improved, but it would be good if we could provide a minimal example to the reader for guidance, even if that's just steering them to an OSS recursive DNS resolver they can run locally if a minimal `compose.yaml` is difficult to express in our docs due to time constraints.

docs/content/config/security/rspamd.md
georglauterbach: I think the old link is now useless, as it points to a page _somewhere_ within Spamhaus. You're far better off directly searching for it in your favorite search machine.

georglauterbach:
@polarathene proposed to add an example; I am not against such an example. Should we go for a compose file? CC @casperklein

casperklein:
Here is a simple example that works:

```yaml
services:
  mailserver:
    container_name: mailserver
    hostname: mail.example.com
    env_file: mailserver.env
    # https://docker-mailserver.github.io/docker-mailserver/latest/config/security/understanding-the-ports/
    ports:
      - "25:25"    # SMTP  (explicit TLS => STARTTLS, Authentication is DISABLED => use port 465/587 instead)
    volumes:
      - ./docker-data/dms/mail-data/:/var/mail/
      - ./docker-data/dms/mail-state/:/var/mail-state/
      - ./docker-data/dms/mail-logs/:/var/log/mail/
      - ./docker-data/dms/config/:/tmp/docker-mailserver/
    restart: always
    stop_grace_period: 1s
    # Uncomment if using `ENABLE_FAIL2BAN=1`:
    # cap_add:
    #   - NET_ADMIN
    healthcheck:
      test: "ss --listening --ipv4 --tcp | grep --silent ':smtp' || exit 1"
      timeout: 3s
      retries: 0

# EVERYTHING BELOW IS NEW

    dns:
      - "10.10.10.10"
    depends_on:
      - powerdns-recursor

  powerdns-recursor:
    image: powerdns/pdns-recursor-master:latest
    container_name: powerdns-recursor
    volumes:
    ports:
      - "53:53"
      - "53:53/udp"
    command:
      # https://doc.powerdns.com/recursor/settings.html
      # Uncomment to enable logging
      # --quiet=no
    restart: always
    stop_grace_period: 0s
    networks:
      default:
        ipv4_address: 10.10.10.10

networks:
  default:
    ipam:
      driver: default
      config:
        - subnet: "10.10.10.0/24"
```

This is our default `compose.yaml`. My additions are below the comment "# EVERYTHING BELOW IS NEW"

PS: Instead of PowerDNS, any other resolver should also fit this setup.

docs/content/config/security/rspamd.md
polarathene: Oh no like I agree with you that the current context from the docs is misleading, but the current link does bring you to a FAQ on the topic itself IIRC, where there was plenty of FAQ entries that could be useful to the reader? (_at least that's what I recall_).

I haven't tried looking at search engine results for comparison, so you might be right. It will matter less though if we provide a good enough example of _"run this to use DNSBLs properly"_, I doubt many users care as much about understanding these topics like I do, they just want solutions 😅

polarathene:
Revised for docs submission.
- No need to show the big `mailserver` snippet, our docs tend to omit the stuff that isn't relevant for the user to pay attention to, it should be implied.
- Technically it could also be used as a `compose.override.yaml` or `compose.dns.yaml` (_with `docker compose -f compose.yaml -f compose.dns.yaml up -d --force-recreate`_).

````markdown
???+ example "Setup a recursive DNS resolver for DMS to use"

    Configure your DMS container (`mailserver`) to forward DNS queries through to the added `dns-recursor` container via adding the `dns` service setting as shown below.

    This `dns` setting requires an explicit IP address. The [implicit `default` network][docker-docs::compose::default-network] is explicitly configured with a subnet, so that a specific IP address can then be assigned to the `dns-recursor` container.

    In this example PowerDNS was chosen, however you can use any DNS container service that's capable of functioning as a recursive resolver.

    ```yaml title="compose.yaml"
    services:
      # Append these settings to your real `compose.yaml`
      mailserver:
        dns:
          - "10.10.10.10"
        depends_on:
          - dns-recursor

      dns-recursor:
        # NOTE: `-master:latest` is the equivalent of DMS `:edge`,
        # PowerDNS stable releases have a naming convention like: `powerdns/pdns-recursor-53:5.3.5`
        # To track the latest stable release, follow their changelog:
        # https://doc.powerdns.com/recursor/changelog/index.html
        # https://github.com/PowerDNS/pdns/blob/master/Docker-README.md
        image: powerdns/pdns-recursor-master:latest
        # Uncomment to enable logging:
        # https://doc.powerdns.com/recursor/settings.html#quiet
        #command: '--quiet=no'
        restart: always
        stop_grace_period: 0s
        volumes:
        networks:
          default:
            ipv4_address: 10.10.10.10

    networks:
      default:
        # Advised if your container host can be reached via IPv6:
        enable_ipv6: true
        ipam:
          driver: default
          config:
            - subnet: "10.10.10.0/24"
    ```

    !!! info "Docker Compose includes embedded DNS"

        Docker Compose with user-defined networks (default) first route DNS queries internally to resolve IPs to containers, or perform rDNS on container IPs.

        If there is no match by the embedded DNS service (_`127.0.0.11:53`, only reachable within the container_), the DNS query will be forwarded to the configured `dns` service.

    !!! warning "Ensure IPv6 support if your container host routes IPv6"

        `enable_ipv6: true` will [prevent a security risk][docs::ipv6::security-risks] for published ports that are reachable via IPv6 connections to the container host. This concern isn't specific to the `dns-recursor` service itself, but rather the standard DMS container when publishing ports to it's internal network and the default `0.0.0.0` binding (all interfaces).

        If your host does not have an IPv6 enabled interface or you have `"userland-proxy": false` configured in `/etc/docker/daemon.json`, this is additional setting is not required.

[docs::ipv6::security-risks]: ../advanced/ipv6.md#what-can-go-wrong
[docker-docs::compose::default-network]: https://docs.docker.com/reference/compose-file/networks/#the-default-network
````

Those official images maintenance looks a bit sloppy though 😓 

- https://github.com/PowerDNS/pdns/blob/master/Docker-README.md (_still mentions Debian 11 Bullseye as the base_)
- `Dockerfile` upgraded to Debian 13 Trixie, but [retained the additional `bookworm-backports` apt repo](https://github.com/PowerDNS/pdns/blob/28dde5167a67530ff7f1810efb1ba00032b7fd67/Dockerfile-recursor#L8-L16).
- As noted in my example commentary the image publishing convention is unconventional with how they're publishing releases (_split into separate images by major+minor, rather than just using tags properly_).
- I haven't looked to see if there's much benefit in relying on that image vs a custom inline Dockerfile that uses a distro like Fedora ([`pdns-recursor`](https://packages.fedoraproject.org/pkgs/pdns-recursor/pdns-recursor/)) or Alpine with the package.

I am not too familiar with PowerDNS, so if it's default config is sufficient without any concerns and that resolves the DNSBL issue well with caching the queries to avoid rate limits, awesome 👍

---

> ```yaml
> ports:
>   - "53:53/tcp"
>   - "53:53/udp"
> ```

For the Compose example, this should not be necessary?

A container can reach another containers ports within the same bridged network, no need to publish ports unless something else were to use this service, thus I think there's no need to document this?

> ```yaml
> container_name: powerdns-recursor
> volumes:
>   - /etc/localtime:/etc/localtime:ro
> ```

This is just QoL?:
- Friendlier container name to use commands on? (_for Docker Compose you'll probably be fine with just `docker compose <cmd> dns-recursor`_)
- `/etc/localtime` mount for logs? Or is there another purpose that's helping with? It wasn't in the upstream `docker-compose.yml` example or related docs.

**Comment by casperklein** [comment]
>Technically it could also be used as a compose.override.yaml

Absolutely 👍 

>For the Compose example, this should not be necessary?

You are right. That can be removed.

> This is just QoL?:

Yes

For your other PDNS concerns: This should just be a quick POC. Any other DNS recursor should work fine too. I've only chosen PowerDNS here, because I used it in the past, so I could quickly provide an example.

**Comment by georglauterbach** [comment]
Appreciate the feedback! Would it be okay @polarathene / @casperklein  if you insert the changes into this MR? I could copy-paste it, but I think it's more efficient in case you want to adjust the surroundings as well.

File: docs/content/config/security/rspamd.md [hunk 5]
```diff
@@ -157,6 +155,65 @@ DMS does not supply custom values for DNS servers (to Rspamd). If you need to us
 
     This setting is enabled to not allow spam to proceed just because DNS requests did not succeed. It could deny legitimate e-mails to pass though too in case your DNS setup is incorrect or not functioning properly.
 
+???+ example "Setup a recursive DNS resolver for DMS to use"
```
polarathene: My bad this should have been collapsed by default:

```suggestion
??? example "Setup a recursive DNS resolver for DMS to use"
```

polarathene (approved):
LGTM 👍

Thanks everyone 😁 (_this was quite a collaborative PR between us all_ 😎)

slydetector:
Thanks for adding this documentation. I had been ignoring all the DNS failure warnings for ages!

PR #4673, opened 2026-03-03
docs: getmail - Update getmail xoauth example links

Nailik:
# Description

I noticed the links were outdated and was struggling with the setup until i found https://github.com/getmail6/getmail6/issues/203 and noticed that i was on an outdated documentation page.

### DIFF
--- a/docs/content/config/advanced/mail-getmail.md
+++ b/docs/content/config/advanced/mail-getmail.md
@@ -106,8 +106,8 @@ It is possible to utilize the `getmail-gmail-xoauth-tokens` helper to provide au
 
 [getmail-website]: https://www.getmail6.org
 [getmail-docs]: https://getmail6.org/configuration.html
-[getmail-docs-xoauth-12]: https://github.com/getmail6/getmail6/blob/master/docs/getmailrc-examples#L286
-[getmail-docs-xoauth-13]: https://github.com/getmail6/getmail6/blob/master/docs/getmailrc-examples#L351
+[getmail-docs-xoauth-12]: https://github.com/getmail6/getmail6/blob/v6.19.10/docs/getmailrc-examples#L286
+[getmail-docs-xoauth-13]: https://github.com/getmail6/getmail6/blob/v6.19.10/docs/getmailrc-examples#L351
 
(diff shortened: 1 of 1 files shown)


docs/content/config/advanced/mail-getmail.md
```diff
@@ -106,8 +106,8 @@ It is possible to utilize the `getmail-gmail-xoauth-tokens` helper to provide au
 
 [getmail-website]: https://www.getmail6.org
 [getmail-docs]: https://getmail6.org/configuration.html
-[getmail-docs-xoauth-12]: https://github.com/getmail6/getmail6/blob/1f95606156231f1e074ba62a9baa64f892a92ef8/docs/getmailrc-examples#L286
-[getmail-docs-xoauth-13]: https://github.com/getmail6/getmail6/blob/1f95606156231f1e074ba62a9baa64f892a92ef8/docs/getmailrc-examples#L351
+[getmail-docs-xoauth-12]: https://github.com/getmail6/getmail6/blob/master/docs/getmailrc-examples#L286
+[getmail-docs-xoauth-13]: https://github.com/getmail6/getmail6/blob/master/docs/getmailrc-examples#L351
```
polarathene: The reason we don't use `master` is because a related issue to what you experienced can happen, where the lines referenced at the time of our docs update may no longer point to the same expected content/lines at the upstream location (_depending on changes to that file since the commit in our docs_).

What we can do instead is use the current release tag of their project, which will communicate more clearly what released version of `getmail6` we referenced at the time in our docs 👍

```suggestion
[getmail-docs-xoauth-12]: https://github.com/getmail6/getmail6/blob/v6.19.10/docs/getmailrc-examples#L286
[getmail-docs-xoauth-13]: https://github.com/getmail6/getmail6/blob/v6.19.10/docs/getmailrc-examples#L351
```

Another concern can also be with the getmail version shipped within DMS may not align with the latest upstream version 😓 We're on Debian and have a slower release cycle, so we're often behind upstream, `master` branch would risk confusing users.

polarathene (changes requested):

PR #1730, opened 2020-12-23
Solve Fetchmail imap idle issue

brainkiller:
Initial pull request to implement the functionality requested in #1728 
Only a few smoke tests have been done so far.

I'm not sure if the fetchmail idfile can be share across all separate fetchmail instances. If needed/preferred we can split this also.

Issue #1728, opened 2020-12-23
Solve Fetchmail imap idle issue


brainkiller:
## Is your feature request related to a problem?

The known issue with fetchmail imap idle is already solved a few times.
You can also see the example on how to solve it + detailed problem description here: https://marek.dopiera.pl/2018/10/03/fetchmail-imap-idle-systemd-generators/

I suggest to adapt the solution a little bit to split the fetchmail conf in multiple entries per server and add those as new supervisor entries at startup.

aendeavor:
@brainkiller Thank you for your tip. (I would have loved to see the feature request template being used.)

Nevertheless, I will keep this issue open since I think this is worthwhile. You'd need to provide a PR for this since capabilities are currently completely exhausted. Do you think that's possible?:)

brainkiller:
If I find some time in the next couple of days I can probably make the changes and submit a pull request.

aendeavor:
Nice. I'm sure that we can merge this after reviews are in.

brainkiller:
FYI - I'm considering the pull request as complete unless if you request some additional enhancements.

aendeavor:
Closed due to migration to [docker-mailserver/docker-mailserver](https://github.com/docker-mailserver/docker-mailserver). Please re-open over there :D

PR #4675, opened 2026-03-06
feat: add IMAP IDLE support for getmail

brainkiller:
Similar to the FETCHMAIL_PARALLEL option which enables the use of IMAP IDLE, this PR provides the same functionality for getmail.
It builds upon the existing code as it doesn't touch the config file generation. Instead it only focusses on the start of getmail.
This should provide the functionality requested in discussion #4667

## Type of change

- [x]  This change requires a documentation update

## Checklist


casperklein:
Thank you for your contribution. Just wanted to let you know, that we are quiet busy and review will take same time.

georglauterbach:
I will also take a look, but all PRs are currently blocked for me because of #4536

polarathene:
Oh I thought this was scheduled for v16 😓 

I've had a glance over the PR diff but I would like to try take a simpler approach which I've outlined here: https://github.com/docker-mailserver/docker-mailserver/issues/4713#issuecomment-4572260658

That's dependent upon upstream and DMS switching back to using the PyPi package instead of the Debian one to benefit from that.

Taking the approach I've outlined would roughly reduce down to this:

```ini
[program:getmail-parallel]
numprocs = 2
numprocs_start = 1
process_name = %(program_name)s-%(process_num)s
autorestart = true
stopasgroup = true
command = bash -c 'getmail --getmaildir /var/lib/getmail --rcfile getmailrc-account-%(process_num)s && sleep %(ENV_GETMAIL_POLL)sm'
```

---

**NOTE:** Until such a PR is feasible, any DMS user could add the above as a config mount into `/etc/supervisor/conf.d/` and with a complimentary [`user-patches.sh` file](https://docker-mailserver.github.io/docker-mailserver/v15.1/config/advanced/override-defaults/user-patches/) (docs) that places the `--rcfile` referenced entries into the `--getmaildir` location (_using a naming convention that can utilize `%(process_num)s`_), should be simple enough to iterate over and rename/symlink such file names.

For the IDLE support, add the `--idle=INBOX` option. If you need something more flexible, a script should work. Instead of an ENV to split up, you could use the service `process_num` expression (_perhaps via the supervisord `environment` setting, similar to what you already did for `SERVICE_NAME`_) and take a list of files and sort them (`$(ls -1 /var/lib/getmail | sort)`?) then use the index on that to select the appropriate file-name. You could then have a regex for the filename suffix `.idle-INBOX` to detect when to add `--idle` and the value to assign it? (_seemed simpler than the ENV or a JSON/YAML/TOML mapping_)

Or adapt the approach you've got here. It's just more of a hassle to merge a feature when it could be greatly simplified by getting support for the idle setting in config files upstream 😅

polarathene:
I spent more time on investigating this integration with upstream.
- As it stands the `--idle` option has only been relevant for what mailbox to monitor over IMAP for notification of when to restart Getmail's processing loop.
- `mailboxes` must be configured as this is what actually is used to retrieve over IMAP, the `--idle` is just a trigger event. As such the default `mailboxes = ("INBOX",)` would result in whatever mailbox `--idle` subscribes to as pulling from the primary INBOX instead. Something this PR with it's test implementation was not covering, thus not caught.

The latest unreleased Getmail commits have added the ability to use `--idle=` (no value assigned) to exclude a mailbox value and infer it from the first item of `mailboxes`. I'd still rather push for a breaking change to treat `--idle` as a boolean toggle like Fetchmail and have that supported in configs, but upstream has been a bit reluctant on this (_the project being named `getmail6` doesn't lend too well to semver and breaking changes.._).

Arguably the feature itself as I have described it (_and as it is vaguely documented at the Getmail6 project_) is in itself broken and not working as implied/documented.

---

Furthermore, I'd like to know the benefit of enabling/supporting IDLE, given the single mailbox limitation?
- What advantages are being sought out with long-lived connections over multiple Getmail instances, instead of polling at a shorter interval?
- Or rather... Is there an actual problem that IMAP IDLE is solving vs using Polling? How many mailboxes are you configuring, is it more than the standard INBOX per account?

I assume it's very few remote accounts and quite possibly only the default INBOX being monitored. I'm just curious how important the feature actually is to you, or if it's more of just an ick with polling often instead of receiving a notification from the remote server?

Related history for Fetchmail idle support:
- https://github.com/docker-mailserver/docker-mailserver/issues/1728 (_Feature request_)
- https://github.com/docker-mailserver/docker-mailserver/pull/1730 (_merged via [this commit](https://github.com/docker-mailserver/docker-mailserver/commit/061fe12aa7bab6ed4d1e69932b9884e3153da9fd)_)

brainkiller:
Personally I don't consider the single folder a limitation.
As you say, it will mostly only monitor the inbox, except in some odd use cases someone might want to change this (to only get mails filtered by sieve).
If we can get this feature implemented I plan to configure it for 7 accounts to download mails.

The issue I find with polling is that you need to get the interval down to a minute maximum.
Due to all these security featues on various sites where they send you a mail to confirm it's you but these links are only valid for a couple  of minutes. If your interval is too long you have to make several attempts before you can successfully login (hoping you don't get locked out after the 5th attempt).

polarathene:
> The issue I find with polling is that you need to get the interval down to a minute maximum.

Fair enough, thank you :)

> As you say, it will mostly only monitor the inbox, except in some odd use cases someone might want to change this (to only get mails filtered by sieve).

I've yet to get around to filing a bug report properly upstream, but I have pointed it out to the maintainer already that the `--idle` flag is implemented poorly.

It is designed to specify a mailbox to receive a notification about updates from, and yet there's no documentation to note that configuring that alone is insufficient if your mailbox is not `INBOX`. Fetchmail implemented the feature far better.

As this PR lacks a test case for catching that, users are just as likely to make that mistake unaware of it (_I'm kind of surprised that it hasn't been raised as a bug report upstream thus far, but perhaps nobody is bothering to use Getmail with mailboxes other than INBOX_ 🤷‍♂️) 

---

I'll return to this PR after I've approved the Dovecot 2.4 / Debian 13 upgrade PR. I'd much prefer convincing upstream to improve so integration in DMS has reduced complexity.

Apologies for the delay that causes here, but thus far I think you're the only user that's requested support for the feature in DMS 😅 I don't mind supporting integrations that can justify the convenience for users well enough, but I also would very much like it done right so that it's minimal burden to maintainers to carry.

georglauterbach:
@brainkiller It took me a while, but we are now ready to look at this PR again. Please rebase onto the latest `master` to resolve the conflicts. I will then be reviewing this PR so we can go forward with it.

Note: In general, a discussion is not sufficient to get a new feature in like that because it's likely to be unnoticed by maintainers like me. But looking at the discussion at hand, I think it's fair. Please open an issue next time, though :)

brainkiller:
Damn looks like I closed this PR accidently while trying to re-base. I'm re-opening it with my proposed changes

georglauterbach:
No worries, I'll be looking into it shortly.

target/getmail/getmail-service.sh
```diff
@@ -1,47 +1,149 @@
...
+  if [[ ! -f "${RC_FILE}" ]]; then
+    _syslog_error "Specified RC file does not exist: ${RC_FILE}"
     _stop_service
   fi
 
-  sleep "${GETMAIL_POLL}m"
-done
+  GETMAIL_OPTS=()
+  # If the getmailrc file contains IMAP configuration and the GETMAIL_IDLE variable is set to "auto" or contains the specific getmailrc file, enable IMAP IDLE for this getmailrc file.
+  if grep -q 'IMAP' "${RC_FILE}" && [[ ${GETMAIL_IDLE} == "auto" || ${GETMAIL_IDLE} == *$(basename "${RC_FILE}")* ]]; then
```
georglauterbach, line 119: `GETMAIL_IDLE` is matched with unquoted substring globs (`*$(basename "${RC_FILE}")*`), including the inner check on line 124. A value of `user4` also matches configs named `user`, `user4-backup`, or `myuser4`, and can attach the wrong `--idle=` mailbox. Compare the token before `:` with an exact equality against the rc basename (the `.cf` stem in `/etc/getmailrc.d/`).

test/helper/log_and_filtering.bash
```diff
@@ -28,7 +28,7 @@ function _filter_service_log() {
 
   local FILE="/var/log/supervisor/${SERVICE}.log"
   # Alternative log location fallback:
-  [[ -f ${FILE} ]] || FILE="/var/log/mail/${SERVICE}.log"
+  _exec_in_container test -f ${FILE} || FILE="/var/log/mail/${SERVICE}.log"
```
georglauterbach, line 31: This used to run `[[ -f ... ]]` on the **host**, so `/var/log/supervisor/${SERVICE}.log` was never found and every caller fell back to `/var/log/mail/`. Checking inside the container makes any supervised program (notably `rspamd`) prefer an often-empty or stdout-only supervisor log. Getmail’s `_log` messages do belong in supervisor logs; assert that in `getmail.bats` (or a dedicated helper) instead of changing this shared helper.

target/scripts/startup/variables-stack.sh
```diff
@@ -152,6 +152,8 @@ function __environment_variables_general_setup() {
   VARS[FETCHMAIL_PARALLEL]="${FETCHMAIL_PARALLEL:=0}"
   VARS[FETCHMAIL_POLL]="${FETCHMAIL_POLL:=300}"
   VARS[GETMAIL_POLL]="${GETMAIL_POLL:=5}"
+  VARS[GETMAIL_PARALLEL]="${GETMAIL_PARALLEL:=0}"
+  VARS[GETMAIL_IDLE]="${GETMAIL_IDLE:=auto}"
```
georglauterbach, line 156: Defaulting `GETMAIL_IDLE` to `auto` means `GETMAIL_PARALLEL=1` turns on `--idle` for every IMAP rc, with no documented off switch (`mailserver.env` is also not updated). Default it empty/off, and treat `auto` as an explicit opt-in.

docs/content/config/environment.md
```diff
@@ -780,6 +780,17 @@ Enable or disable `getmail`.
...
+##### GETMAIL_PARALLEL
+
+Create a service for each getmail configuration so they can all run independently.
+
+- **0** => Disabled
+- 1 => Enabled
+
+##### GETMAIL_IDLE
+
+- **auto** => `getmail` Specify which getmail configs should be started with the --idle flag. By default it shall look for the folder INBOX. This can be changed by specifying the folder with a colon. If auto is set, it will be enable for ever IMAP config. Default: auto.
```
georglauterbach, line 792: This describes the selector as `account1`, and `__usage` in `getmail-service.sh` uses `getmail-1.rc`, but matching uses the config filename without `.cf` (e.g. `user3`). Please document that token, and that `--idle=FOLDER` only watches that folder for notifications — getmail still fetches `mailboxes` (default `INBOX`).

test/tests/parallel/set1/getmail.bats
```diff
@@ -2,16 +2,39 @@ load "${REPOSITORY_ROOT}/test/helper/setup"
...
+
+  CONTAINER_NAME=${CONTAINER2_NAME}
+  local CUSTOM_SETUP_ARGUMENTS=(
+    --env ENABLE_GETMAIL=1
+    --env GETMAIL_PARALLEL=1
+  )
+  _init_with_defaults
+  _common_container_setup 'CUSTOM_SETUP_ARGUMENTS'
+
+  CONTAINER_NAME=${CONTAINER3_NAME}
```
georglauterbach, line 25: `setup_file` leaves `CONTAINER_NAME` set to `CONTAINER3_NAME`, so the original sequential tests run against the parallel+IDLE container. Export `CONTAINER_NAME=${CONTAINER1_NAME}` in those tests. The new cases only grep debug log strings; a `pgrep --full --idle=` check would catch argv mistakes the logs can miss.

georglauterbach (changes requested):
Requesting changes for two blocking issues: `GETMAIL_IDLE` account matching is a substring glob (wrong configs/mailboxes can get `--idle`), and `_filter_service_log` now looks inside the container, which redirects other test suites away from `/var/log/mail`. Inline comments have the details.

test/helper/log_and_filtering.bash
brainkiller: Indeed this is what I also noticed that the test was always failing. This is why I considered it as a bug. I'll move this to a dedicated helper instead

PR #4696, opened 2026-04-19
fix: update permissions check to invoke `chown` directly on returned paths of `find`

### LINKED ISSUE #4112: bug report: Scaling issues with `check-for-changes.sh` when managing mailboxes / aliases
### 📝 Preliminary Checks

### 👀 What Happened?

Hi,

We host a medium-sized docker-mailserver instance with the following characteristics:
- 40 hosted domain names (vhosts)
- 317 mailboxes (using `maildir` dovecot format)
- 1500 aliases
- 27GB of used mailbox data storage (each mailbox is strictly limited to 1GB).

According to pflogsumm, the server sends and receives between 3000 and 4000 mails per week.

The VPS is essentially only running docker-mailserver, though it does have little resources:
- 2 vCPU (Intel Xeon CPU E5-2680 v4 @ 2.40GHz)
- 2GB RAM
- 250GB HDD

Every time an alias or mailbox is added or deleted, the `check-for-changes.sh` script proceeds to add it to the postfix/dovecot database through the `_postfix_dovecot_changes()` function… and takes **around 1m30s to complete**.

There is usually a high number of IMAP sessions running, which gets killed when reloading the daemon. Most of them aren’t killed properly (imap, imap-login, config). See below for the log output.

In our setup, mailboxes and aliases are managed by users themselves. By repeatedly creating and removing mailboxes/aliases, any user can easily achieve a denial of service that way. A rate-limit is in place and we monitor such operations, though this is still a risky situation and the `check-for-changes.sh` execution time keeps growing with the number of mailboxes and aliases we host.

The HDD and low-end CPU doesn’t help, though I wonder if we can achieve performance improvements on the change detector scripts:
1. is it due to change detector running chown on all folders, if it is still a thing?
    - In that case, may it help if we use Dovecot’s `sdbox` format instead to reduce the number of files ?
    - Though since imap sessions usually gets killed, I’m afraid that it may lead to file corruption, which would be more severe if it affects the whole mailbox instead of a single message.
2. is it possible to reduce resource-consuming operations by selectively recreating the mailbox/aliases list? e.g. only reconstruct the alias list for a new alias, or the mailbox list for a new mailbox (I suppose the default behavior is "recreate everything")
3. are there any new upstream features, or configuration options I may not know of, that we can use to avoid reloading the daemon *or* avoid killing dozens of active IMAP sessions when adding a new mailbox/alias?

### 👟 Reproduction Steps

Our server supports both IPv4 and IPv6, here is its configuration:

```env
HOSTNAME=mail
DOMAINNAME=…
CONTAINER_NAME=mail
TZ=Europe/Paris
ONE_DIR=1
ENABLE_SRS=1
ENABLE_POP3=1
ENABLE_OPENDKIM=0
ENABLE_OPENDMARC=0
ENABLE_DNSBL=0
ENABLE_CLAMAV=0
ENABLE_POLICYD_SPF=0
ENABLE_RSPAMD=1
RSPAMD_LEARN=0
RSPAMD_GREYLISTING=1
RSPAMD_HFILTER=1
SPOOF_PROTECTION=1
ENABLE_FAIL2BAN=0
ENABLE_AMAVIS=0
ENABLE_MANAGESIEVE=1
SSL_TYPE=manual
SSL_CERT_PATH=/etc/certs/fullchain.pem
SSL_KEY_PATH=/etc/certs/privkey.pem
POSTFIX_MAILBOX_SIZE_LIMIT=1024000000
POSTFIX_MESSAGE_SIZE_LIMIT=20480000
ENABLE_SPAMASSASSIN=0
MOVE_SPAM_TO_JUNK=1
SRS_SENDER_CLASSES=envelope_sender,header_sender
REPORT_INTERVAL=weekly
TLS_LEVEL=modern
POSTMASTER_ADDRESS=…
REPORT_SENDER=…
SRS_SECRET=…
PFLOGSUMM_TRIGGER=logrotate
LOG_LEVEL=info
ENABLE_MTA_STS=1
```
We also set the following in `user-patches.sh`:
```sh
# increase limits for dovecot IMAP connections
sed -i 's/^.*default_process_limit.*/default_process_limit = 300/g' /etc/dovecot/conf.d/10-master.conf
sed -i 's/^.*default_client_limit.*/default_client_limit = 1800/g' /etc/dovecot/conf.d/10-master.conf
sed -i 's/^.*default_vsz_limit.*/default_vsz_limit = 512M/g' /etc/dovecot/conf.d/10-master.conf
```
And we have a [custom Dockerfile](https://git.lacontrevoie.fr/lacontrevoie/core/src/branch/root/services/mail/Dockerfile) to add an SSH deamon (not the best idea, I know) so we can manage alias/mailboxes programmatically from a web interface.

1. run a server with the above configuration and the aforementioned hardware specifications, then create the specified number of mailboxes and aliases.
2. attempt to run create or remove a mailbox or alias with the `setup.sh` script while measuring the time between the command execution and the daemon reload

### 🐋 DMS Version

v14.0.0

Debian 12 (Bookworm), Linux 6.1.0-18-amd64, Docker 27.0.3

```yml
https://git.lacontrevoie.fr/lacontrevoie/core/src/branch/root/services/mail/docker-compose.yml
```

### 📜 Relevant log output

```Text
2024-07-14T15:55:36.487416+02:00: (addmailuser executes)
...
2024-07-14T15:56:52.964846+02:00 mail postfix/postfix-script[352320]: refreshing the Postfix mail system
2024-07-14T15:56:53.077823+02:00 mail postfix/master[23975]: reload -- version 3.7.10, configuration /etc/postfix
2024-07-14T15:56:53.439657+02:00 mail dovecot: master: Warning: SIGHUP received - reloading configuration
...
2024-07-14T15:57:00.966536+02:00 mail dovecot: master: Warning: Sent SIGTERM to 10 imap processes
2024-07-14T15:57:00.966644+02:00 mail dovecot: master: Warning: Sent SIGTERM to 10 imap-login processes
2024-07-14T15:57:00.966666+02:00 mail dovecot: master: Warning: Sent SIGTERM to 1 config processes
2024-07-14T15:57:00.966679+02:00 mail dovecot: master: Warning: Processes aren't dying after reload, sent SIGTERM to 21 processes.
2024-07-14T15:57:00.979377+02:00 mail dovecot: imap(314083): Warning: Killed with signal 15 (by pid=23803 uid=0 code=kill)
2024-07-14T15:57:00.979405+02:00 mail dovecot: imap(314081): Warning: Killed with signal 15 (by pid=23803 uid=0 code=kill)
2024-07-14T15:57:00.979416+02:00 mail dovecot: imap(314082): Warning: Killed with signal 15 (by pid=23803 uid=0 code=kill)
2024-07-14T15:57:00.979427+02:00 mail dovecot: imap(314080): Warning: Killed with signal 15 (by pid=23803 uid=0 code=kill)
2024-07-14T15:57:00.988644+02:00 mail dovecot: imap-login: Warning: Killed with signal 15 (by pid=23803 uid=0 code=kill)
2024-07-14T15:57:00.988725+02:00 mail dovecot: imap-login: Warning: Killed with signal 15 (by pid=23803 uid=0 code=kill)
2024-07-14T15:57:00.988744+02:00 mail dovecot: imap-login: Warning: Killed with signal 15 (by pid=23803 uid=0 code=kill)
2024-07-14T15:57:00.988769+02:00 mail dovecot: imap-login: Warning: Killed with signal 15 (by pid=23803 uid=0 code=kill)
2024-07-14T15:57:00.988801+02:00 mail dovecot: imap-login: Warning: Killed with signal 15 (by pid=23803 uid=0 code=kill)
2024-07-14T15:57:01.001929+02:00 mail dovecot: config: Warning: Killed with signal 15 (by pid=23803 uid=0 code=kill)
```

_No response_

neil-lcv-cs:
# Description

As stated in #4112, this avoids running chown on a huge number of files and folders by directly using it on the results of the `find` command.

I renamed the function to omit the `_if_necessary` parts as it is not possible anymore to check when find triggers a chown or not. I am not sure if it would be useful to know this information anyway.

Note: The function does not return 1 in case of error anymore. I guess we should reimplement that before merging.

## Type of change

## Checklist

### DIFF
--- a/target/scripts/check-for-changes.sh
+++ b/target/scripts/check-for-changes.sh
@@ -144,7 +144,7 @@ function _postfix_dovecot_changes() {
 
   # Legacy workaround handled here, only seems necessary for _create_accounts:
   # - `helpers/accounts.sh` logic creates folders/files with wrong ownership.
-  _chown_var_mail_if_necessary
+  _chown_var_mail
 }
 
 function _ssl_changes() {
--- a/target/scripts/check-for-changes.sh
+++ b/target/scripts/check-for-changes.sh
@@ -141,10 +141,6 @@ function _postfix_dovecot_changes() {
   [[ ${CHANGED} =~ ${DMS_DIR}/postfix-virtual.cf ]] && _handle_postfix_virtual_config
   [[ ${CHANGED} =~ ${DMS_DIR}/postfix-regexp.cf  ]] && _handle_postfix_regexp_config
   [[ ${CHANGED} =~ ${DMS_DIR}/postfix-aliases.cf ]] && _handle_postfix_aliases_config
-
-  # Legacy workaround handled here, only seems necessary for _create_accounts:
-  # - `helpers/accounts.sh` logic creates folders/files with wrong ownership.
-  _chown_var_mail
 }
 
 function _ssl_changes() {
--- a/target/scripts/helpers/utils.sh
+++ b/target/scripts/helpers/utils.sh
@@ -72,12 +72,10 @@ function _get_dms_env_value() {
 #
 # `helpers/accounts.sh:_create_accounts` (mkdir, cp) appears to be the only writer to
 # /var/mail folders (used during startup and change detection handling).
-function _chown_var_mail_if_necessary() {
-  # fix permissions, but skip this if 3 levels deep the user id is already set
-  if find /var/mail -maxdepth 3 -a \( \! -user "${DMS_VMAIL_UID}" -o \! -group "${DMS_VMAIL_GID}" \) | read -r; then
-    _log 'trace' 'Fixing /var/mail permissions'
-    chown -R "${DMS_VMAIL_UID}:${DMS_VMAIL_GID}" /var/mail || return 1
-  fi
+function _chown_var_mail() {
+  # if needed, fix permissions for all files and folders 3 levels deep /var/mail
+  log 'trace' 'Fixing /var/mail permissions'
+  find /var/mail -maxdepth 3 \( \! -user "${DMS_VMAIL_UID}" -o \! -group "${DMS_VMAIL_GID}" \) -exec chown "${DMS_VMAIL_UID}:${DMS_VMAIL_GID}" {} +
 }
 
 function _require_n_parameters_or_print_usage() {
--- a/target/scripts/helpers/utils.sh
+++ b/target/scripts/helpers/utils.sh
@@ -73,9 +73,9 @@ function _get_dms_env_value() {
 # `helpers/accounts.sh:_create_accounts` (mkdir, cp) appears to be the only writer to
 # /var/mail folders (used during startup and change detection handling).
 function _chown_var_mail() {
-  # if needed, fix permissions for all files and folders 3 levels deep /var/mail
-  log 'trace' 'Fixing /var/mail permissions'
-  find /var/mail -maxdepth 3 \( \! -user "${DMS_VMAIL_UID}" -o \! -group "${DMS_VMAIL_GID}" \) -exec chown "${DMS_VMAIL_UID}:${DMS_VMAIL_GID}" {} +
+  # Fix ownership of mismatched paths within 3 levels of /var/mail, recursively.
+  _log 'trace' 'Fixing /var/mail permissions'
+  find /var/mail -maxdepth 3 \( \! -user "${DMS_VMAIL_UID}" -o \! -group "${DMS_VMAIL_GID}" \) -exec chown -R -- "${DMS_VMAIL_UID}:${DMS_VMAIL_GID}" {} +
 }
 
 function _require_n_parameters_or_print_usage() {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,13 +8,14 @@ All notable changes to this project will be documented in this file. The format 
 
 ### Breaking
 
-- **Rspamd:**
-  - `setup config dkim` now writes keys as `<domain>-<selector>.private` (previously `<keytype>-<keysize>-<selector>-<domain>.private.txt`). The default `dkim_signing.conf` uses a `$domain-$selector` path template with `try_fallback = true`, so multiple domains share one config. Existing `dkim_signing.conf` files are not overwritten. Rename keys to the new layout if you regenerate them (OpenDKIM: `opendkim/keys/<domain>/<selector>.private` → `rspamd/dkim/<domain>-<selector>.private`). ([#4653](https://github.com/docker-mailserver/docker-mailserver/pull/4653))
-
-### Updated
-
 - **Rspamd**
+  - `setup config dkim` now writes keys as `<domain>-<selector>.private` (previously `<keytype>-<keysize>-<selector>-<domain>.private.txt`). The default `dkim_signing.conf` uses a `$domain-$selector` path template with `try_fallback = true`, so multiple domains share one config. Existing `dkim_signing.conf` files are not overwritten. Rename keys to the new layout if you regenerate them (OpenDKIM: `opendkim/keys/<domain>/<selector>.private` → `rspamd/dkim/<domain>-<selector>.private`). ([#4653](https://github.com/docker-mailserver/docker-mailserver/pull/4653))
   - `setup config dkim` writes a persisted `dkim_selectors.map` for custom selectors. Documentation covers multi-domain setup, OpenDKIM migration, and Ed25519 + RSA fallback (distinct selectors plus a `selectors` array). ([#4653](https://github.com/docker-mailserver/docker-mailserver/pull/4653))
+
+### Fixed
+
+- **Internal**
+  - `/var/mail` permission repair now runs `chown -R` only on paths `find` reports as mismatched (within `-maxdepth 3`), instead of `chown -R` on the entire tree. Account creation sets ownership on new mailbox `home/` and copied sieve files so change detection no longer needs to scan `/var/mail` ([#4696](https://github.com/docker-mailserver/docker-mailserver/pull/4696), [#4112](https://github.com/docker-mailserver/docker-mailserver/issues/4112))
 
 ## [v16.0.0](https://github.com/docker-mailserver/docker-mailserver/releases/tag/v16.0.0)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ All notable changes to this project will be documented in this file. The format 
 ### Fixed
 
 - **Internal**
-  - `/var/mail` permission repair now runs `chown -R` only on paths `find` reports as mismatched (within `-maxdepth 3`), instead of `chown -R` on the entire tree. Account creation sets ownership on new mailbox `home/` and copied sieve files so change detection no longer needs to scan `/var/mail` ([#4696](https://github.com/docker-mailserver/docker-mailserver/pull/4696), [#4112](https://github.com/docker-mailserver/docker-mailserver/issues/4112))
+  - `/var/mail` permission repair now runs `chown -R` only on paths `find` reports as mismatched (within `-maxdepth 3`), instead of `chown -R` on the entire tree. Account creation sets ownership on the new mailbox directory (`mail_path`) so change detection no longer needs to scan `/var/mail` ([#4696](https://github.com/docker-mailserver/docker-mailserver/pull/4696), [#4112](https://github.com/docker-mailserver/docker-mailserver/issues/4112))
 
 ## [v16.0.0](https://github.com/docker-mailserver/docker-mailserver/releases/tag/v16.0.0)
 
--- a/target/scripts/helpers/accounts.sh
+++ b/target/scripts/helpers/accounts.sh
@@ -61,15 +61,17 @@ function _create_accounts() {
         echo "${DOVECOT_USERDB_LINE}" >>"${DOVECOT_USERDB_FILE}"
       fi
 
-      local MAILBOX_HOME="/var/mail/${DOMAIN}/${USER}/home"
+      # Dovecot mail_path is /var/mail/<domain>/<user> (mail_home is .../home).
+      local MAILBOX="/var/mail/${DOMAIN}/${USER}"
+      local MAILBOX_HOME="${MAILBOX}/home"
       mkdir -p "${MAILBOX_HOME}"
-      chown -- "${DMS_VMAIL_UID}:${DMS_VMAIL_GID}" "${MAILBOX_HOME}"
 
       # copy user provided sieve file, if present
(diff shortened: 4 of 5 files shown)


casperklein:
LGTM.

> Note: The function does not return 1 in case of error anymore. I guess we should reimplement that before merging.

Any idea, what the purpose of `return 1` was? The function returns the exit code of the last command by default.

neil-lcv-cs:
> Any idea, what the purpose of return 1 was?

Its only usage seems to be here:

https://github.com/docker-mailserver/docker-mailserver/blob/005ea8fa066c2f117a9b8e3dff8e53045994a692/target/scripts/startup/setup-stack.sh#L92-L95

There was that part `|| return 1` in the previous code. I guess it was there to interrupt the function in case of failure.

target/scripts/helpers/utils.sh
```diff
@@ -72,12 +72,10 @@ function _get_dms_env_value() {
...
-function _chown_var_mail_if_necessary() {
-  # fix permissions, but skip this if 3 levels deep the user id is already set
-  if find /var/mail -maxdepth 3 -a \( \! -user "${DMS_VMAIL_UID}" -o \! -group "${DMS_VMAIL_GID}" \) | read -r; then
-    _log 'trace' 'Fixing /var/mail permissions'
-    chown -R "${DMS_VMAIL_UID}:${DMS_VMAIL_GID}" /var/mail || return 1
-  fi
+function _chown_var_mail() {
+  # if needed, fix permissions for all files and folders 3 levels deep /var/mail
+  log 'trace' 'Fixing /var/mail permissions'
+  find /var/mail -maxdepth 3 \( \! -user "${DMS_VMAIL_UID}" -o \! -group "${DMS_VMAIL_GID}" \) -exec chown "${DMS_VMAIL_UID}:${DMS_VMAIL_GID}" {} +
```
polarathene: While I understand why you'd like to drop the `chown -R` to only run `chown`, the `-maxdepth 3` will prevent recursively applying `chown` when necessary.

`chown -R 5000:5000 /var/mail` was introduced into the source since the first pushed commit IIRC. There was no related context for that intention provided at the time unfortunately.

That was [later changed to the conditional check with `-maxdepth 3`](https://github.com/docker-mailserver/docker-mailserver/pull/2256#discussion_r732316498) because of the overhead of `chown -R` on larger `/var/mail` contents with each container start.

---

I'm not concerned about container setup time where a user deploys DMS with `/var/mail` contents that they've migrated Dovecot mail storage from another source than DMS, or even from a DMS backup where the user copied the contents but didn't preserve ownership.

We do copy over some user config however like this:

https://github.com/docker-mailserver/docker-mailserver/blob/e532b37a689a1a047b17afacdd5f689ff1928492/target/scripts/helpers/accounts.sh#L64-L70

There is no ownership correction applied, that appears to be deferred to this utility command. [In your case where you add a user account](https://github.com/docker-mailserver/docker-mailserver/issues/4112#issuecomment-4191982087), that `mkdir` alone would be triggering the condition at runtime to cause the `chown -R` over all of `/var/mail`.

Likewise we probably don't want to drop the `-maxdepth 3` limit, that is an adequate check AFAIK and covers the expectations for DMS at least (`/var/mail/<domain-part>/<local-part>/*`). No need to iterate over all files/folders beyond that AFAIK.

Some configurations [may deviate from that convention](https://github.com/docker-mailserver/docker-mailserver/issues/2238#issuecomment-1712250577) (LDAP), but I don't recall anyone having reported performance concerns that has done such.
- AFAIK, an existing domain-part (1st depth level) won't be affected, while new accounts will add their own `local-part` path (2nd depth level) and I guess anything in that respective folder (third depth level) which would be redundant for `chown -R` (_but as you're using `{} +`, provided all related paths are bundled it probably makes no difference and the child content would be coalesced with the parent folder path?_).
- Any content changes that add new files like the sieve scripts, when all other content has correct ownership however may fail to get detected by `find`. Originally we didn't have a `home/` subdirectory in the users mailbox directory, so that should probably get addressed separately in it's own related script logic.

---

There is also a bug with the current `find` -> `chown` affecting `/var/mail/*` content that [breaks Postfix from delivering internal mail for system users at `/var/mail/<user>`](https://github.com/orgs/docker-mailserver/discussions/4695#discussioncomment-16681214) via it's `local` delivery agent (_as opposed to `virtual` which routes to Dovecot to handle_), where ownership is expected to match the UID/GID of the user as per their `/etc/passwd` entry, otherwise Postfix fails to deliver new mail to that user file in `/var/mail/` after ownership is modified.

Technically this shouldn't be a concern for DMS users, especially if they've got `/etc/aliases` setup correctly via our `POSTMASTER_ADDRESS` ENV + `postfix-aliases.cf`, DMS could improve on that concern though 😓

It is however a [known issue for users preferring LDAP](https://github.com/docker-mailserver/docker-mailserver/issues/2238), with a [workaround to not have per-user ownership](https://github.com/docker-mailserver/docker-mailserver/issues/2238#issuecomment-1662970885) (_by configuring Dovecot to use static vmail UID/GID like the default file account provisioner does_)

polarathene: This should be valid?:

```suggestion
  find /var/mail -maxdepth 3 \( \! -user "${DMS_VMAIL_UID}" -o \! -group "${DMS_VMAIL_GID}" \) -exec chown -R "${DMS_VMAIL_UID}:${DMS_VMAIL_GID}" {} +
```

Otherwise we need to be aware of more scenarios where the lack of `-R` flag for `chown` can go awry.

target/scripts/check-for-changes.sh
```diff
@@ -144,7 +144,7 @@ function _postfix_dovecot_changes() {
 
   # Legacy workaround handled here, only seems necessary for _create_accounts:
   # - `helpers/accounts.sh` logic creates folders/files with wrong ownership.
-  _chown_var_mail_if_necessary
+  _chown_var_mail
```
polarathene: If we are certain that this is only needed for `helpers/accounts.sh`, we can probably fix ownership in that script itself?

Past involvement from me also suggests it could be valid solution:
- https://github.com/docker-mailserver/docker-mailserver/issues/2238#issuecomment-1754011129 (_suggests to provide a convenience via `setup.sh` if users need to manually invoke this only upon startup, they could likewise use `user-patches.sh` for this now_)
- https://github.com/docker-mailserver/docker-mailserver/issues/3557#issuecomment-1741708401 (detailed history on the functionality for reference)
- https://github.com/docker-mailserver/docker-mailserver/issues/3557#issuecomment-2407043943
  - A user mentions that `find -maxdepth 3` results in 30k paths but that is actually [due to LDAP use with shorter mail-storage paths](https://github.com/docker-mailserver/docker-mailserver/issues/3557#issuecomment-2410307019).
  - Avoiding the 30k overhead would be preferable but after adjusting the storage path to have another folder level to align with `find -maxdepth 3` [they reported no overhead concerns](https://github.com/docker-mailserver/docker-mailserver/issues/3557#issuecomment-2423146488) (_container startup, [LDAP users aren't affected by the helper at runtime](https://github.com/docker-mailserver/docker-mailserver/blob/e532b37a689a1a047b17afacdd5f689ff1928492/target/scripts/check-for-changes.sh#L73-L76)_).
- https://github.com/docker-mailserver/docker-mailserver/pull/2256#issuecomment-947240208 (_various test failures will be raised to catch issues from dropping this helper call, pretty much all due to a [Dovecot vmail (`docker` user/group) ownership mismatch](https://github.com/docker-mailserver/docker-mailserver/pull/2256#issuecomment-1665955418) as mailbox dirs would be `root` IIRC_)

polarathene (changes requested):
I think there might be some minor concerns to address so this doesn't cause any subtle unexpected breakages (_we might be lacking test coverage in some cases?_)

target/scripts/check-for-changes.sh
polarathene: If interested in pursuing this, I think we can remove the `_chown_var_mail` line here in favour of adding to this `helpers/accounts.sh` logic:

https://github.com/docker-mailserver/docker-mailserver/blob/e532b37a689a1a047b17afacdd5f689ff1928492/target/scripts/helpers/accounts.sh#L64-L69

```diff
+     local MAILBOX_HOME="/var/mail/${DOMAIN}/${USER}/home"
-     mkdir -p "/var/mail/${DOMAIN}/${USER}/home"
+     mkdir -p "${MAILBOX_HOME}"
+     chown "${DMS_VMAIL_UID}:${DMS_VMAIL_GID}" "${MAILBOX_HOME}"

-     # copy user provided sieve file, if present
+     # Copy user provided `.sieve` file:
      if [[ -e "/tmp/docker-mailserver/${LOGIN}.dovecot.sieve" ]]; then
-       cp "/tmp/docker-mailserver/${LOGIN}.dovecot.sieve" "/var/mail/${DOMAIN}/${USER}/home/.dovecot.sieve"
+       cp "/tmp/docker-mailserver/${LOGIN}.dovecot.sieve" "${MAILBOX_HOME}/.dovecot.sieve"
+       chown "${DMS_VMAIL_UID}:${DMS_VMAIL_GID}" "${MAILBOX_HOME}/.dovecot.sieve"
      fi
```

That way we skip the `find` operation, I'm just not sure if we should do that for DMS v16 given we might be overlooking something and not have adequate test coverage for it 😓 Probably better for now to just make the change the PR presently proposes.

polarathene:
> > Any idea, what the purpose of `|| return 1` was?
> 
> Its only usage seems to be here:
> 
> https://github.com/docker-mailserver/docker-mailserver/blob/005ea8fa066c2f117a9b8e3dff8e53045994a692/target/scripts/startup/setup-stack.sh#L92-L95
> 
> There was that part `|| return 1` in the previous code. I guess it was there to interrupt the function in case of failure.

Yeah, it was expected that if the `chown` was unsuccessful that DMS would not run well so the container would exit as a result. I think due to the `chown` being in a conditional, to bubble up that failure status it had to use `return`?

I assume that's a non-issue now with the conditional dropped and just executing via `find`? Assuming if any of the `chown` commands it executes fails it'd return the non-zero status?

target/scripts/check-for-changes.sh
georglauterbach: Resolved in [9b69863](https://github.com/docker-mailserver/docker-mailserver/pull/4696/commits/9b6986354a42aafd21cd2fa96d3c17247bf45f01)

target/scripts/helpers/utils.sh
georglauterbach: Resolved in [9b69863](https://github.com/docker-mailserver/docker-mailserver/pull/4696/commits/9b6986354a42aafd21cd2fa96d3c17247bf45f01)

georglauterbach (dismissed):

casperklein (dismissed):

georglauterbach (approved):

Issue #4431, opened 2025-03-26
other: Missing DKIM-Signature Header in Outgoing Emails


xehrad:
### Subject

I would like to configure a not documented mail server use case

### Description

I have encountered an issue where outgoing emails from the server do not include the `DKIM-Signature` header. This is problematic because emails sent from our domain are not being signed with DKIM, which can lead to deliverability issues and reduced trust in our emails.

### Steps to Reproduce:
1. Set up the email server (e.g., using Docker-Mailserver or any mail server implementation).
2. Configure the server to send outgoing emails. (setup  config dkim keysize 1024 domain 'example.com')
3. Set DNS
4. Send an email to an external email address.
5. Check the headers of the sent email (e.g., using Gmail).
6. Verify that the `DKIM-Signature` header is absent from the email.

Please let me know if you need additional information or logs to help resolve this issue.

xehrad:
I found the problem: the default [docker-compose](https://github.com/docker-mailserver/docker-mailserver/blob/master/compose.yaml) does not include a volume for OpenDKIM configuration (`/etc/opendkim/`). Without this volume, outgoing emails are not signed with a `DKIM-Signature` header, leading to potential deliverability issues.  



### **Solution**:  
Adding the following volume to the `docker-compose.yaml` resolves the issue:  

```yaml
volumes:
  - ./docker-data/dms/config/opendkim/:/etc/opendkim/
```

Would it be possible to update the default `docker-compose.yaml` to include this volume mapping by default? This would prevent DKIM signing issues for new users.  

Thanks! 🚀

polarathene:
> Without this volume, outgoing emails are not signed with a `DKIM-Signature` header, leading to potential deliverability issues.

That doesn't make much sense.

> Would it be possible to update the default `docker-compose.yaml` to include this volume mapping by default? This would prevent DKIM signing issues for new users.

How did the volume resolve the issue? What needed to be persisted that wasn't created implicitly at that location internally?

Config changes like these can require a fresh container start to ensure the startup scripts are run. OpenDKIM is likely one of these AFAIK. Prior to DMS v14 it would not have been an issue as we always ran startup scripts upon a container restart, but this was error prone (_as our troubleshooting docs page should explain_), and from DMS v15 this should be properly handled.

[...]

PR #4698, opened 2026-04-21
docs: DKIM - Include command to restart DMS properly

fawazahmed0:
# Description

see https://github.com/docker-mailserver/docker-mailserver/issues/4431#issuecomment-2762353370

## Type of change

## Checklist

### DIFF
--- a/docs/content/config/best-practices/dkim_dmarc_spf.md
+++ b/docs/content/config/best-practices/dkim_dmarc_spf.md
@@ -122,7 +122,7 @@ DKIM is currently supported by either OpenDKIM or Rspamd:
 
     !!! info "Restart required"
 
-        After restarting DMS(docker compose down && docker compose up), outgoing mail will now be signed with your new DKIM key(s) :tada:
+        After restarting DMS (`docker compose up --force-recreate`), outgoing mail will now be signed with your new DKIM key(s) :tada:
 
 === "Rspamd"
 
--- a/docs/content/config/best-practices/dkim_dmarc_spf.md
+++ b/docs/content/config/best-practices/dkim_dmarc_spf.md
@@ -122,7 +122,7 @@ DKIM is currently supported by either OpenDKIM or Rspamd:
 
     !!! info "Restart required"
 
-        After restarting DMS, outgoing mail will now be signed with your new DKIM key(s) :tada:
+        After restarting DMS(docker compose down && docker compose up), outgoing mail will now be signed with your new DKIM key(s) :tada:
 
 === "Rspamd"
 


docs/content/config/best-practices/dkim_dmarc_spf.md
```diff
@@ -122,7 +122,7 @@ DKIM is currently supported by either OpenDKIM or Rspamd:
 
     !!! info "Restart required"
 
-        After restarting DMS, outgoing mail will now be signed with your new DKIM key(s) :tada:
+        After restarting DMS(docker compose down && docker compose up), outgoing mail will now be signed with your new DKIM key(s) :tada:
```
polarathene: What were you doing prior? `docker compose restart`?

---

[`docker compose up --force-recreate`](https://docker-mailserver.github.io/docker-mailserver/latest/config/debugging/#preliminary-checks) should be sufficient (_although `docker compose down` then up is a bit more thorough IIRC_). We also document this command [here](https://docker-mailserver.github.io/docker-mailserver/latest/usage/#get-up-and-running).

```suggestion
        After restarting DMS (`docker compose up --force-recreate`), outgoing mail will now be signed with your new DKIM key(s) :tada:
```

polarathene (changes requested):

docs/content/config/best-practices/dkim_dmarc_spf.md
fawazahmed0: > What were you doing prior? `docker compose restart`?

Yes, `docker compose restart` . The files were never copied to `/etc/opendkim/` on `restart`. Anyways using `docker compose down && docker compose up` fixed this issue. 

> `docker compose up --force-recreate` should be sufficient (although docker compose down then up is a bit more thorough IIRC).

I did not try [--force-recreate](https://docs.docker.com/reference/cli/docker/compose/up/#:~:text=If%20you%20want%20to%20force%20Compose%20to%20stop%20and%20recreate%20all%20containers%2C%20use%20the%20%2D%2Dforce%2Drecreate%20flag.), only difference is it does not recreate `networks` defined in compose file. [down](https://docs.docker.com/reference/cli/docker/compose/down/#:~:text=By%20default%2C%20the,one%20is%20used.) `&& up` also recreates `networks`.  Also there is an old issue for [--force-recreate](https://github.com/docker/compose/issues/4273), though I assume it's already fixed.

polarathene (approved):

docs/content/config/best-practices/dkim_dmarc_spf.md
polarathene: > Yes, `docker compose restart` . The files were never copied to `/etc/opendkim/` on `restart`.

Yeah that's because it presently relies upon scripts that only run on container init (fresh container), we have detection for restarted containers to prevent most of the startup scripts from running as the changes from scripts often assume they're applied from a fresh state, not applying the same change again (_which may not apply cleanly depending on the approach_).

> I did not try `--force-recreate`, only difference is it does not recreate `networks` defined in compose file.

Yeah, which should be fine in this case. There shouldn't be a change to the networks 😅

PR #4706, opened 2026-05-23
feat: add `dms-healthcheck` script for container health checks

radicand:
## Summary

Adds a new `dms-healthcheck` script to `target/bin/` that performs an intelligent health check by querying `supervisorctl` only for services that are actually enabled in the current DMS installation.

The script sources `/etc/dms-settings` (written during container startup) to determine which optional features are enabled, then checks that all corresponding supervisor-managed services are in the `RUNNING` state. `supervisorctl status` returns exit code `0` only when all named services are running — any non-running service results in a non-zero exit, which container runtimes (Docker, Kubernetes) correctly interpret as unhealthy.

If `/etc/dms-settings` does not yet exist, the container is still initializing and the check exits non-zero immediately to avoid false positives during startup.

## Covered Services

| Service | Condition |
|---|---|
| `cron`, `rsyslog`, `postfix`, `changedetector` | Always |
| `dovecot` | `SMTP_ONLY != 1` |
| `opendkim` | `ENABLE_OPENDKIM=1` |
| `opendmarc` | `ENABLE_OPENDMARC=1` |
| `amavis` | `ENABLE_AMAVIS=1` |
| `clamav` | `ENABLE_CLAMAV=1` |
| `fail2ban` | `ENABLE_FAIL2BAN=1` |
| `postgrey` | `ENABLE_POSTGREY=1` |
| `postsrsd` | `ENABLE_SRS=1` |
| `rspamd-redis` | `ENABLE_RSPAMD_REDIS=1` |
| `rspamd` | `ENABLE_RSPAMD=1` |
| `getmail` | `ENABLE_GETMAIL=1` |
| `mta-sts-daemon` | `ENABLE_MTA_STS=1` |
| `saslauthd_ldap` / `saslauthd_rimap` | `ENABLE_SASLAUTHD=1` |
| `fetchmail` / `fetchmail-N` | `ENABLE_FETCHMAIL=1` (parallel-aware) |

## Changes

- **`target/bin/dms-healthcheck`** — new script (installed to `/usr/local/bin/` via existing `COPY target/bin/*` in Dockerfile)
- **`Dockerfile`** — adds `HEALTHCHECK` directive using the new script
- **`compose.yaml`** — updates `healthcheck` from the basic SMTP port check to use `dms-healthcheck`

## Motivation

Discussed in [docker-mailserver-helm #135](https://github.com/docker-mailserver/docker-mailserver-helm/pull/135#discussion_r1748971741) by @polarathene — the Helm chart currently embeds probe logic that needs to know which services are enabled, but this belongs upstream in DMS itself. With this script, external orchestrators can simply call `dms-healthcheck` as their probe command without needing to replicate any of this logic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ All notable changes to this project will be documented in this file. The format 
   - `ENABLE_QUOTAS=1` - When an alias has multiple addresses, the first local mailbox address found will be used for the Dovecot dummy account workaround ([#4581](https://github.com/docker-mailserver/docker-mailserver/pull/4581))
   - Change Detection service - Added support for responding to updated DMS config (_Rspamd and TLS certificates_) to `ACCOUNT_PROVISIONER=LDAP` ([#4627](https://github.com/docker-mailserver/docker-mailserver/pull/4627))
   - The `Dockerfile` has changed the source of `dovecot-fts-xapian` package from the [upstream Github repo source](https://github.com/grosjo/fts-xapian) (_which is presently unaccessible_) to instead use the [Debian package source files](https://deb.debian.org/debian/pool/main/d/dovecot-fts-xapian/) (#4700)
+  - Improved service healthcheck capabilities to validate against the services the running instance has actually enabled. Pre-requiste to resolve an outstanding edge case from [docker-mailserver-helm](https://github.com/docker-mailserver/docker-mailserver-helm/pull/135) ([#4706](https://github.com/docker-mailserver/docker-mailserver/pull/4706)) 
 - **Tests:**
   - Make the helper method `_get_container_ip()` compatible with Docker 29 ([#4606](https://github.com/docker-mailserver/docker-mailserver/pull/4606))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,9 @@ All notable changes to this project will be documented in this file. The format 
   - `ENABLE_QUOTAS=1` - When an alias has multiple addresses, the first local mailbox address found will be used for the Dovecot dummy account workaround ([#4581](https://github.com/docker-mailserver/docker-mailserver/pull/4581))
   - Change Detection service - Added support for responding to updated DMS config (_Rspamd and TLS certificates_) to `ACCOUNT_PROVISIONER=LDAP` ([#4627](https://github.com/docker-mailserver/docker-mailserver/pull/4627))
   - The `Dockerfile` has changed the source of `dovecot-fts-xapian` package from the [upstream Github repo source](https://github.com/grosjo/fts-xapian) (_which is presently unaccessible_) to instead use the [Debian package source files](https://deb.debian.org/debian/pool/main/d/dovecot-fts-xapian/) (#4700)
-  - Improved service healthcheck capabilities to validate against the services the running instance has actually enabled. Pre-requiste to resolve an outstanding edge case from [docker-mailserver-helm](https://github.com/docker-mailserver/docker-mailserver-helm/pull/135) ([#4706](https://github.com/docker-mailserver/docker-mailserver/pull/4706)) 
+  - Added the command `dms-healthcheck` to ease implementing a healthcheck for services enabled in DMS. ([#4706](https://github.com/docker-mailserver/docker-mailserver/pull/4706))
+    - Includes a default [`HEALTHCHECK`](https://docs.docker.com/reference/dockerfile/#healthcheck) in the `Dockerfile` for container runtimes that support it.
+    - Runtimes like k8s or podman must configure a healthcheck explicitly. The DMS Helm chart (`docker-mailserver-helm`) will be [updated to use `dms-healthcheck`](https://github.com/docker-mailserver/docker-mailserver-helm/pull/135).
 - **Tests:**
   - Make the helper method `_get_container_ip()` compatible with Docker 29 ([#4606](https://github.com/docker-mailserver/docker-mailserver/pull/4606))
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -296,6 +296,10 @@ ENV POSTGREY_MAX_AGE=35
 ENV POSTGREY_TEXT="Delayed by Postgrey"
 ENV SASLAUTHD_MECH_OPTIONS=""
 
+# NOTE: This is not part of the OCI image spec (limited compatibility at runtime)
+HEALTHCHECK --interval=30s --timeout=5s --retries=3 \
+  CMD dms-healthcheck
+
 # Add metadata to image:
 LABEL org.opencontainers.image.title="docker-mailserver"
 LABEL org.opencontainers.image.vendor="The Docker Mailserver Organization"
--- a/Dockerfile
+++ b/Dockerfile
@@ -298,8 +298,7 @@ ENV SASLAUTHD_MECH_OPTIONS=""
 
 # NOTE: HEALTHCHECK is not part of the OCI image spec and should not be relied on.
 # Ensure it is either supported by your runtime or use this as an example for your deployment scenario (e.g., kubernetes livenessProbe etc)
-HEALTHCHECK --interval=30s --timeout=5s --retries=3 \
-  CMD dms-healthcheck
+HEALTHCHECK --start-period=30s CMD dms-healthcheck
 
 # Add metadata to image:
 LABEL org.opencontainers.image.title="docker-mailserver"
--- a/Dockerfile
+++ b/Dockerfile
@@ -296,7 +296,8 @@ ENV POSTGREY_MAX_AGE=35
 ENV POSTGREY_TEXT="Delayed by Postgrey"
 ENV SASLAUTHD_MECH_OPTIONS=""
 
-# NOTE: This is not part of the OCI image spec (limited compatibility at runtime)
+# NOTE: HEALTHCHECK is not part of the OCI image spec and should not be relied on.
+# Ensure it is either supported by your runtime or use this as an example for your deployment scenario (e.g., kubernetes livenessProbe etc)
 HEALTHCHECK --interval=30s --timeout=5s --retries=3 \
   CMD dms-healthcheck
 
--- a/compose.yaml
+++ b/compose.yaml
@@ -24,6 +24,7 @@ services:
     # Uncomment if using `ENABLE_FAIL2BAN=1`:
     # cap_add:
     #   - NET_ADMIN
-    healthcheck:
-      # Note that the docker image contains a HEALTHCHECK directive in the Dockerfile; the below is in case your runtime does not support it.
-      test: ["CMD", "dms-healthcheck"]
+    # NOTE: The container image configures a `HEALTHCHECK` internally.
+    # Docker Compose should default to using `dms-healthcheck`.
+    # healthcheck:
+    #   test: "ss --listening --ipv4 --tcp | grep --silent ':smtp' || exit 1"
--- a/compose.yaml
+++ b/compose.yaml
@@ -25,6 +25,5 @@ services:
     # cap_add:
     #   - NET_ADMIN
     healthcheck:
-      test: "ss --listening --ipv4 --tcp | grep --silent ':smtp' || exit 1"
-      timeout: 3s
-      retries: 0
+      # Note that the docker image contains a HEALTHCHECK directive in the Dockerfile; the below is in case your runtime does not support it.
+      test: ["CMD", "dms-healthcheck"]
--- a/target/bin/dms-healthcheck
+++ b/target/bin/dms-healthcheck
@@ -3,13 +3,12 @@
 # Verifies that all supervisor-managed services that should be running are in
 # the `RUNNING` state. Returns exit code `0` if all are healthy, otherwise `1`.
 
-# NOTE: There is an improvement opportunity, if identified in the future, to verify
-# service ports are also listening and operational. An example of this might look like:
-# `ss --listening --ipv4 --tcp | grep --silent ':smtp' || exit 1` or similar.
+# TODO: If beneficial, the script could additionally verify service ports have
+# listeners bound (`ss --listening --ipv4 --tcp | grep --silent ':smtp' || exit 1`)
 
 readonly DMS_SETTINGS='/etc/dms-settings'
 
-# Required to proceed (ensures awareness default enabled services):
+# Required to proceed (ensures awareness of default enabled services):
 [[ ! -f ${DMS_SETTINGS} ]] && exit 1
 
 # shellcheck source=/dev/null
--- a/target/bin/dms-healthcheck
+++ b/target/bin/dms-healthcheck
@@ -53,9 +53,12 @@ declare -a SERVICES=(
 [[ ${ENABLE_MTA_STS}      -eq 1 ]] && SERVICES+=('mta-sts-daemon')
 [[ ${ENABLE_SASLAUTHD}    -eq 1 ]] && SERVICES+=("saslauthd_${SASLAUTHD_MECHANISMS}")
 
+# shellcheck disable=SC2207
 [[ ${ENABLE_FETCHMAIL}    -eq 1 ]] && SERVICES+=(
   $(_get_services_for 'fetchmail' "${FETCHMAIL_PARALLEL}")
 )
+
+# shellcheck disable=SC2207
 [[ ${ENABLE_GETMAIL}      -eq 1 ]] && SERVICES+=(
   $(_get_services_for 'getmail' "${GETMAIL_PARALLEL}")
 )
--- a/target/bin/dms-healthcheck
+++ b/target/bin/dms-healthcheck
@@ -57,8 +57,6 @@ declare -a SERVICES=(
   $(_get_services_for 'getmail' "${GETMAIL_PARALLEL}")
 )
 
-# --- Verify all expected services are RUNNING ---
-
 # `supervisorctl status` reports RUNNING for STARTING/BACKOFF states too.
 # Extract the actual state of each service and verify they all reduce to
 # a single unique value: RUNNING.
--- a/target/bin/dms-healthcheck
+++ b/target/bin/dms-healthcheck
@@ -57,7 +57,7 @@ declare -a SERVICES=(
   $(_get_services_for 'getmail' "${GETMAIL_PARALLEL}")
 )
 
-# `supervisorctl status` reports RUNNING for STARTING/BACKOFF states too.
-# Extract the actual state of each service and verify they all reduce to
-# a single unique value: RUNNING.
+# Ensure all enabled services report a 'RUNNING' status.
+# - We cannot rely upon the exit status of `supervisorctl status` (due to `STARTING`/`BACKOFF`).
+# - Returns an exit status of `0` only when the 2nd column reduces to `RUNNING`.
 test 'RUNNING' = "$(supervisorctl status "${SERVICES[@]}" | awk '{print $2}' | sort -u)"
--- a/target/bin/dms-healthcheck
+++ b/target/bin/dms-healthcheck
@@ -1,0 +1,70 @@
+#!/bin/bash
+
+# Health check for docker-mailserver.
+# Verifies that all supervisor-managed services that should be running are in
+# the `RUNNING` state. Returns exit code `0` if all are healthy, otherwise `1`.
+
+readonly DMS_SETTINGS='/etc/dms-settings'
+
+# The settings file is written near the end of container startup.
+# Until it exists the container is still initializing.
+if [[ ! -f ${DMS_SETTINGS} ]]; then
+  exit 1
+fi
+
+# shellcheck source=/dev/null
+source "${DMS_SETTINGS}"
+
+# When FETCHMAIL_PARALLEL or GETMAIL_PARALLEL is enabled, numbered service
+# instances (e.g. fetchmail-1, fetchmail-2) are registered with supervisor
+# at runtime. Resolve a service name to the actual instance name(s).
+SERVICE_LIST=$(supervisorctl status | awk '{print $1}')
+
+function _get_services_for() {
+  local SERVICE_NAME=${1:?Service name required}
+  local IS_PARALLEL="${2:-0}"
+
+  if [[ ${IS_PARALLEL} -eq 1 ]]; then
+    grep -E "^${SERVICE_NAME}-[0-9]+$" <<< "${SERVICE_LIST}"
+  else
+    echo "${SERVICE_NAME}"
+  fi
+}
+
+# --- Build expected service list ---
+
+# Core services that are always expected to be running:
+declare -a SERVICES=(
+  'cron'
+  'rsyslog'
+  'postfix'
+  'changedetector'
+)
+
+# Conditional services based on enabled features:
+[[ ${SMTP_ONLY}           -ne 1 ]] && SERVICES+=('dovecot')
+[[ ${ENABLE_OPENDKIM}     -eq 1 ]] && SERVICES+=('opendkim')
+[[ ${ENABLE_OPENDMARC}    -eq 1 ]] && SERVICES+=('opendmarc')
+[[ ${ENABLE_AMAVIS}       -eq 1 ]] && SERVICES+=('amavis')
+[[ ${ENABLE_CLAMAV}       -eq 1 ]] && SERVICES+=('clamav')
+[[ ${ENABLE_FAIL2BAN}     -eq 1 ]] && SERVICES+=('fail2ban')
+[[ ${ENABLE_POSTGREY}     -eq 1 ]] && SERVICES+=('postgrey')
+[[ ${ENABLE_SRS}          -eq 1 ]] && SERVICES+=('postsrsd')
+[[ ${ENABLE_RSPAMD_REDIS} -eq 1 ]] && SERVICES+=('rspamd-redis')
+[[ ${ENABLE_RSPAMD}       -eq 1 ]] && SERVICES+=('rspamd')
+[[ ${ENABLE_MTA_STS}      -eq 1 ]] && SERVICES+=('mta-sts-daemon')
+[[ ${ENABLE_SASLAUTHD}    -eq 1 ]] && SERVICES+=("saslauthd_${SASLAUTHD_MECHANISMS}")
+
+[[ ${ENABLE_FETCHMAIL}    -eq 1 ]] && SERVICES+=(
+  $(_get_services_for 'fetchmail' "${FETCHMAIL_PARALLEL}")
+)
+[[ ${ENABLE_GETMAIL}      -eq 1 ]] && SERVICES+=(
+  $(_get_services_for 'getmail' "${GETMAIL_PARALLEL:-0}")
+)
+
+# --- Verify all expected services are RUNNING ---
+
+# `supervisorctl status` reports RUNNING for STARTING/BACKOFF states too.
+# Extract the actual state of each service and verify they all reduce to
+# a single unique value: RUNNING.
+test 'RUNNING' = "$(supervisorctl status "${SERVICES[@]}" | awk '{print $2}' | sort -u)"
--- a/target/bin/dms-healthcheck
+++ b/target/bin/dms-healthcheck
@@ -6,11 +6,8 @@
 
 readonly DMS_SETTINGS='/etc/dms-settings'
 
-# The settings file is written near the end of container startup.
-# Until it exists the container is still initializing.
-if [[ ! -f ${DMS_SETTINGS} ]]; then
-  exit 1
-fi
+# Required to proceed (ensures awareness default enabled services):
+[[ ! -f ${DMS_SETTINGS} ]] && exit 1
 
 # shellcheck source=/dev/null
 source "${DMS_SETTINGS}"
--- a/target/bin/dms-healthcheck
+++ b/target/bin/dms-healthcheck
@@ -28,8 +28,6 @@ function _get_services_for() {
   fi
 }
 
-# --- Build expected service list ---
-
 # Core services that are always expected to be running:
 declare -a SERVICES=(
   'cron'
--- a/target/bin/dms-healthcheck
+++ b/target/bin/dms-healthcheck
@@ -59,7 +59,7 @@ declare -a SERVICES=(
   $(_get_services_for 'fetchmail' "${FETCHMAIL_PARALLEL}")
 )
 [[ ${ENABLE_GETMAIL}      -eq 1 ]] && SERVICES+=(
-  $(_get_services_for 'getmail' "${GETMAIL_PARALLEL:-0}")
+  $(_get_services_for 'getmail' "${GETMAIL_PARALLEL}")
 )
 
 # --- Verify all expected services are RUNNING ---
--- a/target/bin/dms-healthcheck
+++ b/target/bin/dms-healthcheck
@@ -12,11 +12,11 @@ readonly DMS_SETTINGS='/etc/dms-settings'
 # shellcheck source=/dev/null
 source "${DMS_SETTINGS}"
 
-# When FETCHMAIL_PARALLEL or GETMAIL_PARALLEL is enabled, numbered service
-# instances (e.g. fetchmail-1, fetchmail-2) are registered with supervisor
-# at runtime. Resolve a service name to the actual instance name(s).
 SERVICE_LIST=$(supervisorctl status | awk '{print $1}')
 
+# Support for `IMAP IDLE` (via ENV `FETCHMAIL_PARALLEL` / `GETMAIL_PARALLEL`):
+# This feature requires multiple instances of a service managed by supervisord.
+# Filter the service list for a numbered suffix of the service name.
 function _get_services_for() {
   local SERVICE_NAME=${1:?Service name required}
   local IS_PARALLEL="${2:-0}"
--- a/target/bin/dms-healthcheck
+++ b/target/bin/dms-healthcheck
@@ -3,8 +3,9 @@
 # Verifies that all supervisor-managed services that should be running are in
 # the `RUNNING` state. Returns exit code `0` if all are healthy, otherwise `1`.
 
-# TODO: If beneficial, the script could additionally verify service ports have
-# listeners bound (`ss --listening --ipv4 --tcp | grep --silent ':smtp' || exit 1`)
+# TODO: The script could also verify that service ports have listeners bound (if justified).
+# Implementation approach was discussed with two examples for checking port 25 here:
+# https://github.com/docker-mailserver/docker-mailserver/pull/4706#discussion_r3344258001
 
 readonly DMS_SETTINGS='/etc/dms-settings'
 
--- a/target/bin/dms-healthcheck
+++ b/target/bin/dms-healthcheck
@@ -1,8 +1,11 @@
 #!/bin/bash
 
-# Health check for docker-mailserver.
 # Verifies that all supervisor-managed services that should be running are in
 # the `RUNNING` state. Returns exit code `0` if all are healthy, otherwise `1`.
+
+# NOTE: There is an improvement opportunity, if identified in the future, to verify
+# service ports are also listening and operational. An example of this might look like:
+# `ss --listening --ipv4 --tcp | grep --silent ':smtp' || exit 1` or similar.
 
 readonly DMS_SETTINGS='/etc/dms-settings'
 


target/bin/dms-healthcheck
```diff
@@ -0,0 +1,59 @@
...
+[[ ${ENABLE_FAIL2BAN}     -eq 1 ]] && SERVICES+=('fail2ban')
+[[ ${ENABLE_POSTGREY}     -eq 1 ]] && SERVICES+=('postgrey')
+[[ ${ENABLE_SRS}          -eq 1 ]] && SERVICES+=('postsrsd')
+[[ ${ENABLE_RSPAMD_REDIS} -eq 1 ]] && SERVICES+=('rspamd-redis')
+[[ ${ENABLE_RSPAMD}       -eq 1 ]] && SERVICES+=('rspamd')
+[[ ${ENABLE_GETMAIL}      -eq 1 ]] && SERVICES+=('getmail')
+
+# ENABLE_MTA_STS is not exported to /etc/dms-settings; fall back to container ENV:
+[[ ${ENABLE_MTA_STS:-0}   -eq 1 ]] && SERVICES+=('mta-sts-daemon')
+
```
polarathene: Thanks for spotting this, I raised a [PR to address that](https://github.com/docker-mailserver/docker-mailserver/pull/4707).

```suggestion
[[ ${ENABLE_MTA_STS}      -eq 1 ]] && SERVICES+=('mta-sts-daemon')
```

target/bin/dms-healthcheck
```diff
@@ -0,0 +1,59 @@
...
+if [[ ${ENABLE_SASLAUTHD} -eq 1 ]]; then
+  SERVICES+=("saslauthd_${SASLAUTHD_MECHANISMS}")
+fi
```
polarathene: Any reason this one doesn't match the style of prior lines? (_folded length is 85 chars_)

```suggestion
[[ ${ENABLE_SASLAUTHD}    -eq 1 ]] && SERVICES+=("saslauthd_${SASLAUTHD_MECHANISMS}")
```

target/bin/dms-healthcheck
```diff
@@ -0,0 +1,59 @@
...
+  if [[ ${FETCHMAIL_PARALLEL} -eq 1 ]]; then
+    COUNTER=0
+    for _ in /etc/fetchmailrc.d/fetchmail-*.rc; do
+      COUNTER=$(( COUNTER + 1 ))
+      SERVICES+=("fetchmail-${COUNTER}")
+    done
+  else
+    SERVICES+=('fetchmail')
+  fi
+fi
```
polarathene: This logic from the related startup script to generate the `fetchmail` parallel services doesn't seem necessary to include here.

Instead with a helper method defined earlier like this:

```bash
SERVICE_LIST=$(supervisorctl status | awk '{print $1}')

function get_services_for() {
    local SERVICE_NAME=${1:?Service name required}
    local IS_PARALLEL="${2:-0}"

    if [[ ${IS_PARALLEL} -eq 1 ]]; then
        grep -E "^${SERVICE_NAME}-[0-9]+$" <<< "${SERVICE_LIST}"
    else
        echo "${SERVICE_NAME}"
    fi
}
```

We avoid the noise in the assignment here:

```suggestion
[[ ${ENABLE_FETCHMAIL} -eq 1 ]] && SERVICES+=(
    $(get_services_for 'fetchmail' "${FETCHMAIL_PARALLEL}")
)
```

There's another PR that wants to introduce the [parallel service instances for the alternative Getmail service](https://github.com/docker-mailserver/docker-mailserver/pull/4675), which was another motivator to extract a common helper method so that after that PR lands we can make the following change:

```bash
[[ ${ENABLE_GETMAIL} -eq 1 ]] && SERVICES+=(
    $(get_services_for 'getmail' "${GETMAIL_PARALLEL}")
)
```

target/bin/dms-healthcheck
```diff
@@ -0,0 +1,59 @@
...
+
+supervisorctl status "${SERVICES[@]}"
```
polarathene: Apart from this sending the command output to stdout (do we want that?), the exit status code is still `0` for `STARTING` and `BACKOFF` states.

We have the services we want to filter our check on at this point, we can take the status of each (2nd column of each line) and then do a unique sort such that a healthy status for the container would mean the value was reduced down to `RUNNING`. Even if it were empty this would fail (_should something else go wrong in the process at some point in future_).

Thus we just return the exit status of a string comparison via `test` command and it'll be the expected `0` (true) / `1` (false):

```suggestion
test 'RUNNING' = "$(supervisorctl status "${SERVICES[@]}" | awk '{print $2}' | sort -u)"
```

polarathene: [Docker's docs for `HEALTHCHECK`](https://docs.docker.com/reference/dockerfile/#healthcheck) have an expectation of `0` (success) / `1` (unhealthy), not just any non-zero code (_they specifically state `2` is reserved_). I'm not sure what Kubernetes or Podman may expect on their ends, but at least with the above suggestion we ensure that the result is only `0`/`1`.

---

Those docs also state that stdout + stderr are captured into a 4096 byte buffer to assist with debugging unhealthy services.

This is 1,396 bytes and is not filtered, so we could optionally capture the output and log it:

```console
$ supervisorctl status

amavis                           RUNNING   pid 719, uptime 1:38:50
changedetector                   RUNNING   pid 723, uptime 1:38:49
clamav                           STOPPED   Not started
cron                             RUNNING   pid 576, uptime 1:38:51
dovecot                          RUNNING   pid 586, uptime 1:38:51
fail2ban                         STOPPED   Not started
fetchmail                        STOPPED   Not started
fetchmail-1                      RUNNING   pid 646, uptime 1:38:50
fetchmail-2                      RUNNING   pid 10208, uptime 0:00:16
getmail                          STOPPED   Not started
mailserver                       RUNNING   pid 8, uptime 1:38:53
mta-sts-daemon                   STOPPED   Not started
opendkim                         RUNNING   pid 595, uptime 1:38:50
opendmarc                        RUNNING   pid 605, uptime 1:38:50
postfix                          RUNNING   pid 616, uptime 1:38:50
postgrey                         STOPPED   Not started
postsrsd                         STOPPED   Not started
rspamd                           STOPPED   Not started
rspamd-redis                     STOPPED   Not started
rsyslog                          RUNNING   pid 580, uptime 1:38:51
saslauthd_ldap                   STOPPED   Not started
saslauthd_rimap                  STOPPED   Not started
update-check                     STOPPED   Not started
```

I don't personally see a need to do that, so I'm fine without it (_users can just run that command themselves manually_).

compose.yaml
```diff
@@ -25,6 +25,6 @@ services:
     # cap_add:
     #   - NET_ADMIN
     healthcheck:
-      test: "ss --listening --ipv4 --tcp | grep --silent ':smtp' || exit 1"
-      timeout: 3s
-      retries: 0
+      test: ["CMD", "dms-healthcheck"]
```
polarathene: Since you're adding the check into the image, we could probably remove the check entirely from the `compose.yaml`? (_or leave a comment that that image has one by default?_)

**UPDATE:** Healthcheck implementations aren't standardized with OCI images, [it's a reserved field](https://github.com/opencontainers/image-spec/issues/749#issuecomment-2289913422) (_but not looking like it'll be accepted into the actual spec_).
- I have not verified but AFAIK it may be present in the OCI image metadata still? It's future support isn't clear and since support varies across the ecosystem I guess we should keep an example in the compose config after all?
- [One user notes multi-platform image builds via a custom builder are not preserving the `HEALTHCHECK` instruction as metadata](https://github.com/containers/podman/issues/25454#issuecomment-2737372742) into the exported images for them, which may be due to a difference of exporting docker format images or OCI images (_in addition to this Docker apparently has an extension to the OCI image spec, where they're persisting such info into their OCI image builds as metadata for compatible clients to use_).

---

The previous healthcheck might be useful if the `supervisorctl status` of `RUNNING` wasn't sufficient, but `postfix` seems to start and bind to the port pretty quickly. We can always add improvements to the healthcheck script in future, if reports of scenarios appear where the `RUNNING` status is misleading (_due to failures that don't cause the service to exit_).

polarathene: For context with the DMS image and the [`healthcheck` in Docker Compose](https://docs.docker.com/reference/compose-file/services/#healthcheck), `CMD` will run the `dms-healthcheck` command directly, so it's shebang is used to run via bash 👍

<details>
<summary>Redundant</summary>

**EDIT:** Nevermind. Pretty sure below is irrelevant.

`CMD-SHELL` is implicit if this was just `test: dms-healthcheck`, and would then use the `Dockerfile` `SHELL` instruction configured for the image? (_default is `/bin/sh`, but [DMS customizes `SHELL`](https://github.com/docker-mailserver/docker-mailserver/blob/cb76075f25e22476e8fdb45adfbea8026d4ea898/Dockerfile#L16) for failing fast in the image build_). I haven't verified this, IIRC it's not metadata that persists into an image for container runtime (_other than when using shell-form syntax in instructions like `ENTRYPOINT`/`CMD`, which are present in image metadata_).

The `SHELL` instruction isn't acknowledged by default when building images with Podman unless using opt-in for `Dockerfile` compatibility, I'm not sure if that'd affect runtime there ([`SHELL` is only relevant for shell form syntax in a `Dockerfile`](https://docs.docker.com/reference/dockerfile/#understand-how-cmd-and-entrypoint-interact) AFAIK (_including it's `HEALTHCHECK` instruction_). Not that anyone should be using `compose.yaml` with Podman these days 😅 I was thinking more about Quadlets.

Looking over at the Podman [Quadlet](https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html#healthcmd) and [`CLI`](https://docs.podman.io/en/latest/markdown/podman-run.1.html#health-cmd-command-command-arg1) docs, they seem to use `/bin/sh` as default. Chances are Docker is doing the same for it's CLI and Compose support (_regardless of what `SHELL` is defined as_), unless using shell-form syntax with the `HEALTHCHECK` in a `Dockerfile` build.

</details>

target/bin/dms-healthcheck
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+
+# Health check for docker-mailserver.
+# Verifies that all supervisor-managed services that should be running are in
+# the RUNNING state. Returns exit code 0 if all are healthy, non-zero otherwise.
```
polarathene:
```suggestion
# the `RUNNING` state. Returns exit code `0` if all are healthy, otherwise `1`.
```

Dockerfile
```diff
@@ -296,6 +296,9 @@ ENV POSTGREY_MAX_AGE=35
 ENV POSTGREY_TEXT="Delayed by Postgrey"
 ENV SASLAUTHD_MECH_OPTIONS=""
 
+HEALTHCHECK --interval=30s --timeout=5s --retries=3 \
+  CMD dms-healthcheck
```
polarathene:
```suggestion
# NOTE: This is not part of the OCI image spec (limited compatibility at runtime)
HEALTHCHECK --interval=30s --timeout=5s --retries=3 \
  CMD dms-healthcheck
```

polarathene (changes requested):
Thanks for taking the time to put this PR together! ❤️ 

I've provided some revisions that will improve the correctness of the check, but there's a related issue that'll need to be resolved separately and won't be identified by this healthcheck script into fixed.

---

## Undetected failure - Infinite service restarts due to `startsecs=0`

**UPDATE:** [Raised as a separate issue](https://github.com/docker-mailserver/docker-mailserver/issues/4708).

I'd like to also point out that presently services that immediately fail will result in [log spam with repeated crashes](https://github.com/docker-mailserver/docker-mailserver/issues/4579) (_this could happen with `fetchmail` parallel too but that cause was fixed for v16 of DMS_).

According to the [`supervisord` docs](https://supervisord.org/configuration.html#program-x-section-settings), these scenarios should stop restarting the services (_default retry attempts is `startretries = 3`_), but due to all our services setting the default `startsecs = 1` to `0`, such failure behaviour is skipped and won't be detected by `supervisorctl status` as a result (_like shown in the issue linked, `RUNNING` state with exit status `0` is returned_).

This `startsecs=0` bug has been around since the [introduction of the `supervisord` into DMS](https://github.com/docker-mailserver/docker-mailserver/pull/676) (Aug 2017). A separate PR should tackle that concern as increasing this back to `1` will slow down container startup by 1 sec per service since we sequentially invoke `supervisorctl start ...` manually for each service individually.

When this concern is resolved, the service goes to [`BACKOFF` state](https://supervisord.org/subprocess.html#process-states) during retries (progressively taking longer before each retry), followed by entering a `FATAL` state if the failing service is unable to achieve an uptime of `startsecs` before the `startretries` count is exhausted. Which looks like this from `supervisorctl status fetchmail-2` output:

```
fetchmail-2                      FATAL     Exited too quickly (process log may have details)
```

compose.yaml
casperklein: As polarathene pointed out, function wise this would be a regression.
Checking if `postfix` opened a port is more advanced than just checking if `supervisord` thinks a service is running.

+1 for keeping the check inside `compose.yaml`.

radicand:
[Agent comment] I've updated the PR with all your feedback incorporated:

## Changes Made

✅ **Exit code semantics** — Now uses `test 'RUNNING' = "$(supervisorctl status "${SERVICES[@]}" | awk '{print $2}' | sort -u)"` to ensure only exit codes 0 (healthy) or 1 (unhealthy) per Docker HEALTHCHECK spec. This also correctly detects STARTING/BACKOFF/FATAL states as unhealthy.

✅ **Suppressed stdout** — No output during healthy checks. Docker's health log buffer stays clean.

✅ **Saslauthd alignment** — Folded to one-liner with proper formatting: `[[ ${ENABLE_SASLAUTHD} -eq 1 ]] && SERVICES+=("saslauthd_${SASLAUTHD_MECHANISMS}")`

✅ **Parallel service helper** — Extracted `_get_services_for()` function exactly as suggested. Queries `supervisorctl status` output for `SERVICE-N` patterns instead of walking the filesystem.

✅ **Future-proof for GETMAIL_PARALLEL** — Already included support for getmail parallel instances (PR #4675). Will work immediately when that lands.

✅ **ENABLE_MTA_STS fallback removed** — Rebased on latest master which includes PR #4707. Now directly reads `ENABLE_MTA_STS` from `/etc/dms-settings`.

✅ **OCI compatibility note** — Added comment to HEALTHCHECK in Dockerfile.

✅ **compose.yaml kept as-is** — Reverted to keep the original ss-based SMTP port check. Acknowledges that checking the port is more advanced than supervisor status alone.

## Tested

I tested this locally on my own running instance (in k8s).

Dockerfile
radicand: improved per suggestion

target/bin/dms-healthcheck
radicand: rebased against latest main with the patch; made it work the same as the rest

target/bin/dms-healthcheck
radicand: formatting error, fixed

target/bin/dms-healthcheck
radicand: good call, implemented

target/bin/dms-healthcheck
radicand: applied the suggestion

compose.yaml
radicand: i reverted this change to how it was originally (`"ss --listening --ipv4 --tcp | grep --silent ':smtp' || exit 1"`)

target/bin/dms-healthcheck
radicand: applied formatting

polarathene:
Please disclose the extent to your use of AI. I assume it's Claude since one interaction only really stood out as highly unlikely for a human 🤔 (_other signals I can give the benefit of the doubt_)

[You use Copilot](https://github.com/radicand/forwardauth-rs), Claude Opus is a [supported model](https://docs.github.com/en/enterprise-cloud@latest/copilot/how-tos/use-copilot-agents/cloud-agent/changing-the-ai-model#supported-models).

---

Attribution was [retained in PR commit](https://github.com/docker-mailserver/docker-mailserver/pull/4706/changes/5440ce40c0b5a3d558634649c6c2d6da9cd7b467), confirming copilot use:

```
Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>
```

Please clarify how much copilot was involved in the PR description and subsequent comments, how much interaction am I having with an actual human during review?

target/bin/dms-healthcheck
```diff
@@ -0,0 +1,70 @@
...
+[[ ${ENABLE_GETMAIL}      -eq 1 ]] && SERVICES+=(
+  $(_get_services_for 'getmail' "${GETMAIL_PARALLEL:-0}")
+)
+
+# --- Verify all expected services are RUNNING ---
+
+# `supervisorctl status` reports RUNNING for STARTING/BACKOFF states too.
+# Extract the actual state of each service and verify they all reduce to
+# a single unique value: RUNNING.
+test 'RUNNING' = "$(supervisorctl status "${SERVICES[@]}" | awk '{print $2}' | sort -u)"
```
polarathene: It doesn't report `RUNNING` for those states, it shares the exit status of `0`. The service has not entered a failed state at that point, so it's understandable why the command returns `0`. However we only want to consider `RUNNING` as healthy in our check to my knowledge?

Correction these inline docs:

```suggestion
# Ensure all enabled services report a 'RUNNING' status.
# - We cannot rely upon the exit status of `supervisorctl status` (due to `STARTING`/`BACKOFF`).
# - Returns an exit status of `0` only when the 2nd column reduces to `RUNNING`.
test 'RUNNING' = "$(supervisorctl status "${SERVICES[@]}" | awk '{print $2}' | sort -u)"
```

target/bin/dms-healthcheck
```diff
@@ -0,0 +1,70 @@
...
+[[ ${ENABLE_SASLAUTHD}    -eq 1 ]] && SERVICES+=("saslauthd_${SASLAUTHD_MECHANISMS}")
+
+[[ ${ENABLE_FETCHMAIL}    -eq 1 ]] && SERVICES+=(
+  $(_get_services_for 'fetchmail' "${FETCHMAIL_PARALLEL}")
+)
+[[ ${ENABLE_GETMAIL}      -eq 1 ]] && SERVICES+=(
+  $(_get_services_for 'getmail' "${GETMAIL_PARALLEL:-0}")
+)
+
+# --- Verify all expected services are RUNNING ---
```
polarathene: Redundant noise

```suggestion
```

target/bin/dms-healthcheck
```diff
@@ -0,0 +1,70 @@
...
+  local IS_PARALLEL="${2:-0}"
+
+  if [[ ${IS_PARALLEL} -eq 1 ]]; then
+    grep -E "^${SERVICE_NAME}-[0-9]+$" <<< "${SERVICE_LIST}"
+  else
+    echo "${SERVICE_NAME}"
+  fi
+}
+
+# --- Build expected service list ---
```
polarathene: This can be inferred by the reader. I'm assuming this was noise added by the AI agent without any human direction to add it?

If I'm mistaken, you can add clarity to the bounds of the logic by wrapping it into a function call. This is not necessary though given the script itself is quite simple.


```suggestion
```

target/bin/dms-healthcheck
```diff
@@ -0,0 +1,70 @@
...
+
+# shellcheck source=/dev/null
+source "${DMS_SETTINGS}"
+
+# When FETCHMAIL_PARALLEL or GETMAIL_PARALLEL is enabled, numbered service
+# instances (e.g. fetchmail-1, fetchmail-2) are registered with supervisor
+# at runtime. Resolve a service name to the actual instance name(s).
+SERVICE_LIST=$(supervisorctl status | awk '{print $1}')
+
+function _get_services_for() {
```
polarathene: The comment you've added is misplaced. I'm fine with some inline documentation if it's helpful but it should be co-located by the actual function.

The `SERVICE_LIST` variable is used by the helper method, but the comment is providing context for the helper method. The variable name `SERVICE_LIST` should be rather self-explanatory for the associated assignment commands involved.

Here's a revised version of the comment in the right place:

```suggestion
SERVICE_LIST=$(supervisorctl status | awk '{print $1}')

# Support for `IMAP IDLE` (via ENV `FETCHMAIL_PARALLEL` / `GETMAIL_PARALLEL`):
# This feature requires multiple instances of a service managed by supervisord.
# Filter the service list for a numbered suffix of the service name.
function _get_services_for() {
```

target/bin/dms-healthcheck
```diff
@@ -0,0 +1,70 @@
...
+# Verifies that all supervisor-managed services that should be running are in
+# the `RUNNING` state. Returns exit code `0` if all are healthy, otherwise `1`.
+
+readonly DMS_SETTINGS='/etc/dms-settings'
+
+# The settings file is written near the end of container startup.
+# Until it exists the container is still initializing.
+if [[ ! -f ${DMS_SETTINGS} ]]; then
+  exit 1
+fi
```
polarathene: The comment is a bit misleading actually 😅 

The file is created when this method is called:

https://github.com/docker-mailserver/docker-mailserver/blob/dacec800f7c8db91683bbf5fa68fbb55300a05d1/target/scripts/startup/variables-stack.sh#L6-L18

As per the function name, it's occurring rather early in `start-mailserver.sh`:

https://github.com/docker-mailserver/docker-mailserver/blob/dacec800f7c8db91683bbf5fa68fbb55300a05d1/target/scripts/start-mailserver.sh#L169-L175

As such the services like fetchmail/getmail parallel services won't have been configured yet, so they won't be picked up in `SERVICE_LIST` below.

This is the actual near end of container startup. This is where supervisord services are started:

https://github.com/docker-mailserver/docker-mailserver/blob/dacec800f7c8db91683bbf5fa68fbb55300a05d1/target/scripts/start-mailserver.sh#L203-L207

Slightly earlier we do have another file as an indicator of progress, much later than `/etc/dms-settings`:

https://github.com/docker-mailserver/docker-mailserver/blob/dacec800f7c8db91683bbf5fa68fbb55300a05d1/target/scripts/start-mailserver.sh#L192-L193

However if the container has restarted, both of these files will be present within the containers own persistence layer until it's destroyed.

---

Since we are only concerned about `/etc/dms-settings` being available to proceed further, and knowing that the final check (`test` command) will fail anyway due to other services not running in the scenario of getmail/fetchmail parallel services not being present in `SERVICE_LIST`, we can dismiss all these concerns.

```suggestion
# Required to proceed (ensures awareness default enabled services):
[[ ! -f ${DMS_SETTINGS} ]] && exit 1
```

That is more accurate, since we are depending upon it for the default values typically not specified which are set here:

https://github.com/docker-mailserver/docker-mailserver/blob/dacec800f7c8db91683bbf5fa68fbb55300a05d1/target/scripts/startup/variables-stack.sh#L94-L117

target/bin/dms-healthcheck
```diff
@@ -0,0 +1,70 @@
...
+[[ ${ENABLE_RSPAMD_REDIS} -eq 1 ]] && SERVICES+=('rspamd-redis')
+[[ ${ENABLE_RSPAMD}       -eq 1 ]] && SERVICES+=('rspamd')
+[[ ${ENABLE_MTA_STS}      -eq 1 ]] && SERVICES+=('mta-sts-daemon')
+[[ ${ENABLE_SASLAUTHD}    -eq 1 ]] && SERVICES+=("saslauthd_${SASLAUTHD_MECHANISMS}")
+
+[[ ${ENABLE_FETCHMAIL}    -eq 1 ]] && SERVICES+=(
+  $(_get_services_for 'fetchmail' "${FETCHMAIL_PARALLEL}")
+)
+[[ ${ENABLE_GETMAIL}      -eq 1 ]] && SERVICES+=(
+  $(_get_services_for 'getmail' "${GETMAIL_PARALLEL:-0}")
```
polarathene: No need to set a fallback, that's already handled in the helper function called:

```suggestion
  $(_get_services_for 'getmail' "${GETMAIL_PARALLEL}")
```

polarathene (changes requested):
Please add an entry to the changelog.

---

Please resolve the lint failure without having your AI agent over-engineer the solution.

It may tend to prefer applying best practices if it is unable to determine relevant context to why it could avoid adding complexity. I'd prefer a simple script that is easy to grok and maintain by humans that can grasp why this logic is safe.

If I didn't get the impression I was engaging with AI for review, I'd provide a clearer suggestion here that could be applied. AI disclosure should have been provided earlier in your PR description tbh, but I see we haven't required that in the PR template.

---

> ✅ **compose.yaml kept as-is** — Reverted to keep the original ss-based SMTP port check. Acknowledges that checking the port is more advanced than supervisor status alone.

It is not, it would be complimentary.
- It would be better to incorporate the listening state of the port into the healthcheck script was my input **but** only if reports were received where it made a meaningful difference so that we'd have context of a false positive (_Postfix binds to port 25 quite quickly once started that for it to be deemed `RUNNING` by supervisord but not have binded port 25, I'd like to know the cause of failure_).
- My advice was to leave the healthcheck example in `compose.yaml`, but make users aware of the built-in `dms-healthcheck` command.

> ✅ **Suppressed stdout** — No output during healthy checks. Docker's health log buffer stays clean.

Meaningful logs were still relevant, such as the one to communicate the cause of an unhealthy state (_during container startup when `/etc/dms-settings` doesn't exist yet_).

**EDIT:** As per my review feedback on that early exit from the script, it's less likely to be meaningful (_`/etc/dms-settings` is created quite early in startup, or if the container restarted it'd already be present and potentially outdated_). So no need to revert.

radicand:
> Please add an entry to the changelog.
> 
> Please resolve the lint failure without having your AI agent over-engineer the solution.

@polarathene thanks for taking the time to review. I admit I was in a bit of a rush this morning and it was a case of "looks good to me, runs and functions". The feedback is appropriate. I'll work out the rest and follow up again.

radicand:
> Please disclose the extent to your use of AI. I assume it's Claude since one interaction only really stood out as highly unlikely for a human 🤔 (_other signals I can give the benefit of the doubt_)
> 
> [You use Copilot](https://github.com/radicand/forwardauth-rs), Claude Opus is a [supported model](https://docs.github.com/en/enterprise-cloud@latest/copilot/how-tos/use-copilot-agents/cloud-agent/changing-the-ai-model#supported-models).
> 
> Attribution was [retained in PR commit](https://github.com/docker-mailserver/docker-mailserver/pull/4706/changes/5440ce40c0b5a3d558634649c6c2d6da9cd7b467), confirming copilot use:
> 
> ```
> Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>
> ```
> 
> Please clarify how much copilot was involved in the PR description and subsequent comments, how much interaction am I having with an actual human during review?

Hey there - I use AI in drafting, and yes, with Claude generally. I don't blind commit. You're interacting with me here in this PR - I do want to close my PR from the sister helm repo and this HC script is not incredibly complicated as a concept. Hope that helps clarify.

polarathene:
> Hope that helps clarify.

Thanks! All good, I just needed to verify if AI was being used and to what extent. I don't mean any disrespect by it 😅 

---

> I admit I was in a bit of a rush this morning and it was a case of "looks good to me, runs and functions".

No worries I understand, it happens.

> I use AI in drafting, and yes, with Claude generally.

> I don't blind commit. You're interacting with me here in this PR

I'm not against using AI, but without upfront disclosure it can sometimes be difficult to discern how much of the interactions (_PR description, review comments, [update comment](https://github.com/docker-mailserver/docker-mailserver/pull/4706#issuecomment-4528952042)_) are delegated to tooling.

The PR description and notably the update comment had signals of AI use. The review comments were terse but sometimes seemed unnecessary to individually reply to like:
- "[applied formatting](https://github.com/docker-mailserver/docker-mailserver/pull/4706#discussion_r3295261124)"
- "[applied the suggestion](https://github.com/docker-mailserver/docker-mailserver/pull/4706#discussion_r3295259793)"
- etc

It was becoming unclear if I was interacting with AI only (has happened in reviews before). When decisions were made / committed that didn't seem well thought out, that's when I needed to sanity check how AI was involved.

---

> this HC script is not incredibly complicated as a concept.

While that is true, there were bugs identified along with questionable commentary. If the use of AI is at fault for that, it eats away into my time as a reviewer which I'm not fond of.

compose.yaml
radicand: Apologies for the back and forth on this; I'd misinterpreted the outcome of the above conversation to mean you'd preferred to revert/keep the earlier version. I'm re-committing the version with the healthcheck script in it, with a note/TODO that checking for the listening port or similar is an improvement we could do it needed later.

radicand:
@polarathene I think I've got the bulk of the feedback addressed at this point; I need to resolve the lint still and will ping you when that's done.

polarathene:
> I need to resolve the lint still

You may just annotate them to ignore the lint if you like.

My earlier feedback was `mapfile`/`read` suggested from the lint are not necessary. The script is straight-forward, the service names don't contain white-space, splitting is intentional and safe. Furthermore the grep regex filter ensures that nothing else should get through unintentionally.

If another maintainer chimes in later to disagree, then we'll address that, but my opinion is there is no need to go to extra lengths.

target/bin/dms-healthcheck
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
+
+# Verifies that all supervisor-managed services that should be running are in
+# the `RUNNING` state. Returns exit code `0` if all are healthy, otherwise `1`.
+
+# NOTE: There is an improvement opportunity, if identified in the future, to verify
+# service ports are also listening and operational. An example of this might look like:
+# `ss --listening --ipv4 --tcp | grep --silent ':smtp' || exit 1` or similar.
```
polarathene:
```suggestion
# TODO: If beneficial, the script could additionally verify service ports have
# listeners bound (`ss --listening --ipv4 --tcp | grep --silent ':smtp' || exit 1`)
```

target/bin/dms-healthcheck
```diff
@@ -0,0 +1,66 @@
...
+
+readonly DMS_SETTINGS='/etc/dms-settings'
+
+# Required to proceed (ensures awareness default enabled services):
```
polarathene:
```suggestion
# Required to proceed (ensures awareness of default enabled services):
```

target/bin/dms-healthcheck
```diff
@@ -0,0 +1,66 @@
...
+[[ ${ENABLE_RSPAMD}       -eq 1 ]] && SERVICES+=('rspamd')
+[[ ${ENABLE_MTA_STS}      -eq 1 ]] && SERVICES+=('mta-sts-daemon')
+[[ ${ENABLE_SASLAUTHD}    -eq 1 ]] && SERVICES+=("saslauthd_${SASLAUTHD_MECHANISMS}")
+
+[[ ${ENABLE_FETCHMAIL}    -eq 1 ]] && SERVICES+=(
+  $(_get_services_for 'fetchmail' "${FETCHMAIL_PARALLEL}")
+)
+[[ ${ENABLE_GETMAIL}      -eq 1 ]] && SERVICES+=(
+  $(_get_services_for 'getmail' "${GETMAIL_PARALLEL}")
+)
```
polarathene, line 64: **EDIT:** Resolving as already applied by PR author.

```suggestion
# shellcheck disable=SC2207 # Values returned are safe to split in this context
[[ ${ENABLE_FETCHMAIL}    -eq 1 ]] && SERVICES+=(
  $(_get_services_for 'fetchmail' "${FETCHMAIL_PARALLEL}")
)
# shellcheck disable=SC2207 # Values returned are safe to split in this context
[[ ${ENABLE_GETMAIL}      -eq 1 ]] && SERVICES+=(
  $(_get_services_for 'getmail' "${GETMAIL_PARALLEL}")
)
```

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ All notable changes to this project will be documented in this file. The format
   - `ENABLE_QUOTAS=1` - When an alias has multiple addresses, the first local mailbox address found will be used for the Dovecot dummy account workaround ([#4581](https://github.com/docker-mailserver/docker-mailserver/pull/4581))
   - Change Detection service - Added support for responding to updated DMS config (_Rspamd and TLS certificates_) to `ACCOUNT_PROVISIONER=LDAP` ([#4627](https://github.com/docker-mailserver/docker-mailserver/pull/4627))
   - The `Dockerfile` has changed the source of `dovecot-fts-xapian` package from the [upstream Github repo source](https://github.com/grosjo/fts-xapian) (_which is presently unaccessible_) to instead use the [Debian package source files](https://deb.debian.org/debian/pool/main/d/dovecot-fts-xapian/) (#4700)
+  - Improved service healthcheck capabilities to validate against the services the running instance has actually enabled. Pre-requiste to resolve an outstanding edge case from [docker-mailserver-helm](https://github.com/docker-mailserver/docker-mailserver-helm/pull/135) ([#4706](https://github.com/docker-mailserver/docker-mailserver/pull/4706))
```
polarathene: Reworded it for better clarity to our users reading the changelog:

```suggestion
  - Added the command `dms-healthcheck` to ease implementing a healthcheck for services enabled in DMS. ([#4706](https://github.com/docker-mailserver/docker-mailserver/pull/4706))
    - Includes a default [`HEALTHCHECK`](https://docs.docker.com/reference/dockerfile/#healthcheck) in the `Dockerfile` for container runtimes that support it.
    - Runtimes like k8s or podman must configure a healthcheck explicitly. The DMS Helm chart (`docker-mailserver-helm`) will be [updated to use `dms-healthcheck`](https://github.com/docker-mailserver/docker-mailserver-helm/pull/135).
```

FWIW the edge case you referenced is just that the current helm check runs `grep` on the status output, only to verify if any single service is `RUNNING` (_thus a bug rather than an edge-case_) 😅

compose.yaml
```diff
@@ -25,6 +25,5 @@ services:
     # cap_add:
     #   - NET_ADMIN
     healthcheck:
-      test: "ss --listening --ipv4 --tcp | grep --silent ':smtp' || exit 1"
-      timeout: 3s
-      retries: 0
+      # Note that the docker image contains a HEALTHCHECK directive in the Dockerfile; the below is in case your runtime does not support it.
```
polarathene: ⚠️ This suggestion seems a bit bugged out and would need to be applied manually.

```suggestion
    # NOTE: The container image configures a `HEALTHCHECK` internally.
    # Docker Compose should default to using `dms-healthcheck`.
    #healthcheck:
    #  test: "ss --listening --ipv4 --tcp | grep --silent ':smtp' || exit 1"
```

- Dropping the `timeout: 3s` (default: 10s) + `retries: 0` (default: 3) seems fine 👍
- I think @casperklein wanted to retain the port example (_which is only monitoring Postfix SMTP as opposed to considering unhealthy status when any enabled service has crashed and failed to recover_).
- I cared about the reader being aware of `dms-healthcheck` now that we have `HEALTHCHECK` added into the `Dockerfile`.

polarathene (changes requested):
Apart from the `compose.yaml`, you should be able to switch to the "Files changed" tab of this PR and click "Add to batch" on each suggestion to apply them all via a single commit.

The `compose.yaml` suggestion looks borked for what it'd modify, you'd need to manually apply that one (and may want to wait until @casperklein provides feedback).

compose.yaml
radicand: Ah, thanks for clarifying re the compose file. I admit I was confused on where the discussion landed as to *what* to keep and intrepreted it in the way I edited it just now. With that understanding I'll re-apply manually.

CHANGELOG.md
radicand: Agreed re: the checking of service statuses, and that this is an improvement overall against that. What sent me down this rabbit hole in the first place was a weird case (my edge case) where the double grep syntax used was hanging in my production deployment and causing the pod to fail out at random intervals. We're solving for both here :)

radicand:
> Apart from the `compose.yaml`, you should be able to switch to the "Files changed" tab of this PR and click "Add to batch" on each suggestion to apply them all via a single commit.
> 
> The `compose.yaml` suggestion looks borked for what it'd modify, you'd need to manually apply that one (and may want to wait until @casperklein provides feedback).

Cheers, I'd missed that feature -- appreciate you calling it out. I'm good with this now if/when @casperklein weighs in. Really appreciate you taking the time and being thorough here, and my apologies for my sloppiness earlier today.

polarathene (dismissed):
LGTM 👍 Thanks!

target/bin/dms-healthcheck
```diff
@@ -0,0 +1,68 @@
...
+
+# shellcheck disable=SC2207
+[[ ${ENABLE_GETMAIL}      -eq 1 ]] && SERVICES+=(
+  $(_get_services_for 'getmail' "${GETMAIL_PARALLEL}")
+)
+
+# Ensure all enabled services report a 'RUNNING' status.
+# - We cannot rely upon the exit status of `supervisorctl status` (due to `STARTING`/`BACKOFF`).
+# - Returns an exit status of `0` only when the 2nd column reduces to `RUNNING`.
+test 'RUNNING' = "$(supervisorctl status "${SERVICES[@]}" | awk '{print $2}' | sort -u)"
```
casperklein, line 69:
```suggestion
if ! [[ "RUNNING" == "$(supervisorctl status "${SERVICES[@]}" | awk '{print $2}' | sort -u)" ]]; then
  exit 1
fi

# Check if postfix is listening on port tcp/25
if ! ss -l -4 -n | grep -qF '0.0.0.0:25'; then
  exit 1
fi

# DMS is healthy
exit 0
```

To have all health checks in one place, we could also add the port check here.

Dockerfile
casperklein: The line break seems unnecessary to me.

---

Do we really want/need 3 retries?

target/bin/dms-healthcheck
polarathene: The port 25 check was left in a comment as a TODO, should we actually find a need/benefit in having that check. Presently AFAIK it's just an assumption, and it'd be nicer to know a scenario where DMS has Postfix reporting `RUNNING` but isn't listening on port 25 when it's expected to.

In that same vein we could also query other ports on other services, but even then a bound port for a service to listen on doesn't equate to a working service port 😅

If you do insist on including it, I won't challenge that but we could express the conditions in a shell group `{ ... }` with a single `|| exit 1` instead? (_to ensure any non-zero status is coalesced into `1`_):

```suggestion
{
  # Verify all enabled services are ready:
  test 'RUNNING' == "$(supervisorctl status "${SERVICES[@]}" | awk '{print $2}' | sort -u)" && \
  # Verify port 25 (Postfix) is bound:
  ( ss --listening --ipv4 --tcp | grep --silent ':smtp' ) 
} || exit 1
```

For me at least, stacking a bunch of if statements and inverting the condition to trigger `exit 1` in each seemed like a code smell 🤔

Dockerfile
polarathene: @casperklein

> The line break seems unnecessary to me.

Personally I prefer the distinction/separation if there's all those `HEALTHCHECK` options defined, and that's similar to what is shown in the [`HEALTHCHECK` docs example](https://docs.docker.com/reference/dockerfile/#healthcheck).

Given feedback with suggestion below, I do agree that it's not worthwhile and we can collapse it back to a single line 👍

---

> Do we really want/need 3 retries?

The [default retries is already 3](https://docs.docker.com/reference/dockerfile/#healthcheck), as is the interval of 30s.

If anything I'd raise more concern with `--interval=30s` as that is how long we're delaying the first health check which affects any containers using `depends_on` with the health status of DMS.

Retries being more than 1 is absolutely valid, if the health check is successful at the first attempt, the container is marked healthy, but if it failed, you may still want to permit checking again after a moment in the event the healthcheck was run when a service was restarted (and recovered), not immediately assume failure.

The `--timeout` shouldn't need to be lowered (default 30s), on the basis that we don't expect any part of the script to hang, that SIGTERM should be sent any earlier. I'd rather introduce a lower timeout as a response to a real report where investigation is unable to resolve what causes the hang to occur.

DMS is up and ready by default in less than 5s on my system, a bit longer on a budget VPS instance, and some other services when enabled (_like Rspamd IIRC which is intended to become default in future_) can slow that down (_Amavis adds some delay too IIRC_).

If we want to assume that DMS should be up within 10-15s we could use an interval of 5s, or with other services enabled, give some additional slack. As this is the `Dockerfile` healthcheck (_also [applies for Docker Compose](https://docs.docker.com/reference/compose-file/services/#healthcheck)_), we have [additional options to configure](https://docs.docker.com/reference/dockerfile/#healthcheck) (_defaults: `--start-period=0s` + `--start-interval=5s`_) so that the polling interval is shorter only during the initial container startup (_duration window defined by `--start-period`_). Retry failures do not count during the start period, but that duration is cut short as soon as as a healthcheck is successful (_such that if a failure then occurs and subsequent healthchecks fail, even if they were within the declared start period window, they'd no longer be ignored_).

The 5s `--start-interval` default is sufficient for DMS, so just raise `--start-period`

```suggestion
HEALTHCHECK --start-period=30s CMD dms-healthcheck
```

polarathene:
Posting this comment for my benefit of tracking (_since some feedback is not being displayed linearly for me at least, it's a bit awkward to follow/track discussion updates_):

- [x] [`HEALTHCHECK` suggestion](https://github.com/docker-mailserver/docker-mailserver/pull/4706#discussion_r3352668761) (_drop redundant options, fold into a single line, add `--start-period=30s`_)
- [x] [Suggestion for handling additional checks](https://github.com/docker-mailserver/docker-mailserver/pull/4706#discussion_r3352558370) (_this would remove the associated TODO comment, do not apply this suggestion unless either @casperklein or @georglauterbach 👍 to approve it_)

target/bin/dms-healthcheck
casperklein: IMO: A service check (e.g. sending a mail) > port check > `supervisor status`

Replacing the port check with `supervisor status` seems like a little regression to me.

> In that same vein we could also query other ports on other services

Absolutely. But my take on this PR was to improve things a bit and not to make them 100% bullet proof and give OP additional homework 😉 

> For me at least, stacking a bunch of if statements and inverting the condition to trigger exit 1 in each seemed like a code smell 🤔

I agree that we disagree 😆
Compared to multiple conditions spread across multiple lines, that are glued together with `&& \` and using an additional sub-shell, I find my proposal in that case much more readable and explicit.

---

That said, I am fine with the current state of this PR, weather the port check is included or not.
And if it's included, I am fine with both variants 👍

Dockerfile
casperklein: I couldn't imagine a case, where the health check only succeed on the 3rd attempt. That's why I questioned the retry amount. We can leave it as it is..

`--start-period=30` seems like a good addition 👍 

Edit: 

> HEALTHCHECK --interval=30s --timeout=5s --retries=3

This gives the container 1:30min to start, before becoming unhealthy.

polarathene: > I couldn't imagine a case, where the health check only succeed on the 3rd attempt.

First time starting the container and setting up a user? For someone new to it, that can take more than 30s depending how they go about it.

For me it was when I was setting up automated test examples with multiple containers, setting a lower interval that is too aggressive might fail the first try depending on deployment environment. Thankfully I came across `--start-period` which resolved that concern completely.

---

> > ```Dockerfile
> > HEALTHCHECK --interval=30s --timeout=5s --retries=3
> > ```
> 
> This gives the container 1:30min to start, before becoming unhealthy.

Technically during a container startup the status would be `starting` (_which we can still consider as `unhealthy`, at least a `depends_on` would_), and if no check passes, then we stay in that unhealthy state until officially switching to an `unhealthy` status at 90sec uptime, yes.

We resolved that concern for startup healthchecks via a start period and shorter start interval (_which is immune to `--retries` within the start period_) 👍

Thus the `starting` status can switch to a `healthy` status much sooner.
- While a `unhealthy` status still takes 90s (_30s start period + 30s x2 interval checks. Exhausting the start period seems to count as 1 failure, we then wait `--interval` until the next check is performed which fails and so on_).
- If you need to assume failure earlier, you can override these healthcheck params, or have your own logic in place when `healthy` has not been reported after X time.

For actual runtime, the first healthcheck could fail for various reasons (_a single check on a port being bound, not so much, but this PR covers a variety of services_), such as:
- A service may not report a `RUNNING` status if it just crashed for example but it could recover shortly after.
- Intermittent connectivity (_perhaps using Fetchmail/Getmail and that fails to connect to an external mail server? It'll try again and the connection may succeed_).

30s or 90s before your monitoring tools can signal the unhealthy status isn't that much of a difference for your ability to respond, but it does help ensure we avoid false positives wasting your time to respond. Retries are important, and given the types of failures, it's also important that the retries are given enough padding.

Container startup time is a totally different scenario and generally more predictable (_but as mentioned depending on context, there can be some where this will take longer_).

target/bin/dms-healthcheck
polarathene: > IMO: A service check (e.g. sending a mail) > port check > `supervisor status`
> 
> Replacing the port check with `supervisor status` seems like a little regression to me.

I think that depends on perspective?

For example your checking port 25 as your healthcheck, but sending a mail from DMS would use an outbound port 25, which has nothing to do with that, whilst the listening port for mail submission is also 587 or 465.

You may also just want to use DMS with Fetchmail/Getmail and Dovecot for a test, where mail can be retrieved from a remote mail server, and then delivered via LMTP, or other means that do not depend on Postfix listening to port 25 (_or Postfix at all for that matter_). A scenario like has different expectations for what a healthy/ready DMS container means.

If Postfix is healthy, there's a good chance it's already listening on port 25. I've asked for what scenarios that is not the case, but haven't yet had a clear example cited?

**NOTE:** Presently, you would be correct however that `supervisor status` here is a regression. Until a follow-up PR addresses the `startsecs=0` concern I raised with Supervisord (_services instantly are treated as `RUNNING` state due to this, even when they're in a failure loop, it also prevents a service being marked as unhealthy as retry limits only apply to failure transitions from states that are not `RUNNING` which resets the retry count_).

---


> > In that same vein we could also query other ports on other services
> 
> Absolutely. But my take on this PR was to improve things a bit and not to make them 100% bullet proof

I think there's a misunderstanding here. I was trying to highlight that it'd be adding complexity for little to no gain for our project and users.

I would rather we add port checks only when we have evidence of scenarios that cause that, so that we don't just carry redundant checks for the sake of it.

If we carry the port checks, extra work should be done for communicating that kind of failure, rather than just reporting an unhealthy status, as that may not lead to maintainers being made aware of a problem causing it, especially if the fix is just to have a restart policy when the container is `unhealthy`?

If instead it were marked `healthy` but had a problem with a port, it's far more likely a user would report this to us, or PR a port check into the healthcheck script, which comes with valuable context that it's a known issue, and presumably one we've not been able to troubleshoot/resolve the cause of it.

---

> and give OP additional homework 😉

I don't think that is necessary 🤨 

---

> > For me at least, stacking a bunch of if statements and inverting the condition to trigger exit 1 in each seemed like a code smell 🤔
> 
> I agree that we disagree 😆
> Compared to multiple conditions spread across multiple lines, that are glued together with `&& \` and using an additional sub-shell, I find my proposal in that case much more readable and explicit.

That's fair, but the proposal I gave still resulted in the same conditions that you'd have wrapped with `if [[ ... ]];then` + `exit 1` + `fi`. If that's all that is done, it's just extra noise to communicate "N conditions (any true) => `exit 1`", which can really add up as more conditions are accumulated.

You're still going to see those same condition regardless if they're wrapped in several lines of syntax noise, or if each check is co-located right next to each other and scoped to individual lines (_which the subshell was intended to help convey, pragmatically speaking in this context there's no notably overhead in doing so?_).

You know me, I tend to prefer verbosity (_long options, detailed commentary inline docs for context, etc_), but in this case my mental model (_monotropic / gestalt thinking_) is biased towards expressing the logic with a clear encapsulated pattern 😅 (_which I appreciate as an explicit flow to grok for readability_)

That said, abstractions can add complexity, so as the author perhaps I'm blinded by bias 😓 (_I did spend some time trying to see it your way too... but if the conditions themselves were replaced with `_all_enabled_services_are_running && _is_port_25_listening`, I don't see an advantage with not being DRY here?_)

---

> That said, I am fine with the current state of this PR, weather the port check is included or not.
> And if it's included, I am fine with both variants 👍

Sweet!

I'd prefer to avoid bike shedding and just defer implementation to the existing TODO comment which requests that we first have proper justification for the port 25 health check (_or any other service ports_).

polarathene:
@radicand I don't have permission to apply edits to your branch myself, so could you please do the following:

```Dockerfile
HEALTHCHECK --start-period=30s CMD dms-healthcheck
```

```bash
# TODO: The script could also verify that service ports have listeners bound (if justified).
# Implementation approach was discussed with two examples for checking port 25 here:
# https://github.com/docker-mailserver/docker-mailserver/pull/4706#discussion_r3344258001
```

These two discussions were referenced by my previous comment. After these changes are applied, PR is good for merging 👍

radicand:
> @radicand I don't have permission to apply edits to your branch myself, so could you please do the following:
> 
> ```dockerfile
> HEALTHCHECK --start-period=30s CMD dms-healthcheck
> ```
> 
> ```shell
> # TODO: The script could also verify that service ports have listeners bound (if justified).
> # Implementation approach was discussed with two examples for checking port 25 here:
> # https://github.com/docker-mailserver/docker-mailserver/pull/4706#discussion_r3344258001
> ```
> 
> These two discussions were referenced by my previous comment. After these changes are applied, PR is good for merging 👍

Sounds good - I've made the edits in https://github.com/docker-mailserver/docker-mailserver/pull/4706/changes/5ebd34bc6be5cddf5025f3d39a284f0dbe87c8d5 . Thanks again both for the thoughtful review.

polarathene (approved):

casperklein (approved):

target/bin/dms-healthcheck
casperklein: > If Postfix is healthy, there's a good chance it's already listening on port 25. 
> I've asked for what scenarios that is not the case, but haven't yet had a clear example cited?

I can't imagine a scenario too. This was just to be safe in the future, when something unexpected happens.

PR #4786, opened 2026-09-04
feat: remove Xapian

### LINKED ISSUE #4720: feat: replace Dovecot FTS Xapian with Flatcurve
Some considerations were discussed in [this issue description](https://github.com/docker-mailserver/docker-mailserver/issues/4719):
- If we keep the plugin, perhaps considering reverting back to upstream repo for compilation/updates? (_which has since been restored_) Although presently I don't think there is a benefit (_no new commits for 8 months_).
- Compilation could also be restricted to only when enabling image builds to use the Dovecot CE repo, with fallback to Debian package otherwise?
- With Dovecot 2.4, we may want to assess if Flatcurve is a better alternative. This is an FTS option that is bundled with Dovecot 2.4 IIRC? We have some past issue/discussion about this.

--

Related to this, the Dovecot Solr docs page may need to be revised. I know it refers to an upstream issue which has recently been resolved for Solr config, where our docs had to raise some extra awareness for XML related changes.

Existing Xapian docs could be migrated to the Examples section as community maintained advice, assuming we no longer maintain compilation support if Flatcurve can support FTS for users without any justification for Xapian (_generally our users have been vocal about Solr being a better FTS solution, stating various issues with Xapian_).

georglauterbach:
## Description

Remove FTS Xapian. Use Flatcurve!

Fixes #4720

> [!CAUTION]
>
> This is a breaking change

## Type of change

- Update

---

> [!IMPORTANT]
>
> I understand and have ensured that
>
> 1. my code follows the style guidelines of this project,
> 2. I have commented my code, particularly in hard-to-understand areas,
> 3. new and existing unit tests pass locally with my changes,
> 4. if necessary, I added tests that prove my fix is effective or that my feature works,
> 5. if necessary, I added information about changes made in this PR to the changelog.


georglauterbach:
@casperklein Please drop me a short note whether I should wait on your review before merging this.

casperklein (approved):

PR #4787, opened 2026-09-04
docs: update Solr image and clean up FTS configuration

beertje44:
Updated Solr image version,  fixed SOLR_MODULES format and explictly start solr in user-managed mode. Removed commented-out FTS decoder configuration: it is not advised and incorrect for the debian provided dovecot packages. Tika is a better alternative and easy to implement but this is a dovecot fts feature and not fts solr specific so that doesn't belong in the solr tutorial.

## Description

Small iteration: 
 - current solr is 10.0, higher version probably will just work but there is no garantee so don't just use 10.x
 - Starting with solr 10, solr-cloud is the default. In this example a single stand alone solr server is assumed, which was the previous default.
 - Upon further inspection: script based text extraction is not included with dovecot on debian 13, so the example was useless. Tika extraction is better and easy to implement, but has nothing to do with the fts solr plugin for dovecot: it is a dovecot fts feature and should be available to other forms of fts searching also.

Fixes #

## Type of change

- Bug Fix
- Refactoring
- Update

---

> [!IMPORTANT]
>
> I understand and have ensured that
>
> 1. my code follows the style guidelines of this project,
> 2. I have commented my code, particularly in hard-to-understand areas,
> 3. new and existing unit tests pass locally with my changes,
> 4. if necessary, I added tests that prove my fix is effective or that my feature works,
> 5. if necessary, I added information about changes made in this PR to the changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,11 @@ All notable changes to this project will be documented in this file. The format 
 - Moved `setup_cli.bats` to the serial test suite to prevent `setup.sh` from racing with teardown of another DMS test container ([#4701](https://github.com/docker-mailserver/docker-mailserver/issues/4701))
 - Excluded `saslauthd` from the process kill/restart assertion because its forked parent process is not reliably reaped by the test
 
+### Fixed
+
+- **Dovecot**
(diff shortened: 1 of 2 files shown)


georglauterbach (approved):

georglauterbach:
This will land in the `:edge` docs for now until `:17.0.0` is released.

